PR #305, opened 2014-08-23
Hierarchy widget flexibility improvements

zachmullen:
- "routing": whether to change the route when navigating through widget
- "showItems": whether to show items or just folders
- "showActions": whether to show the action bar
- "checkboxes": whether to add checkboxes and checked actions to the widget


zachmullen:
Finally got around to adding the tests for this branch, it's now ready for review.

clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -681,6 +707,21 @@ girder.views.HierarchyWidget = girder.View.extend({
         }).on('g:saved', function (folder) {
             // need to do anything?
         }, this);
+    },
+
+    /**
+     * Select (highlight) an item in the list.
+     * @param item An ItemModel instance representing the item to select.
+     */
+    selectItem: function (item) {
```
manthey: What calls `selectItem`?   Is it something that other javascript would call?  If so, can you add a test doing so?  Similarly with `getSelectedItem`.

zachmullen: I forgot that was in there, thanks for reminding me. I will add a test for this.

zachmullen:
@manthey ping

manthey:
LGTM.

Issue #573, opened 2014-12-15
push python cli client, add recursive upload of a directory


mgrauer:
* add a main and arg parsing to the python client
 * add a recursive upload to the python client given a locally accessible filesystem directory and a target girder folder

zachmullen:
Fixed in #582

PR #582, opened 2014-12-19
initial work on recursive upload

mgrauer:
work on #573 .


zachmullen:
@mgrauer We have an external user trying to use the girder python client against a tangelo installation, and wants to make use of some features of this topic branch. I'm going to clean and merge it in its current state, and move the outstanding stuff into a separate issue.

Issue #577, opened 2014-12-16
System health and usage


manthey:
We discussed having some mechanism to check on system 'health' and memory usage.  I could see a few things that would be interesting: process memory and cpu usage, machine available memory, swap, and cpus, disk availability wherever the girder database is located and wherever assetstores are located, up time.

Much of this information can be gleaned from psutil, which we can add to the requirements.

We have another issue (#496) for adding a system consistency check.  I could see tying in to that as well.  For instance, a thorough health check would tell if a consistency check would need to do anything.

Some health queries are expensive in terms of cpu or IO, so this either should cache the results for a short duration, have a parameter to get only cheap statistics, or some other method to avoid overloading the system.

When we eventually add email alerts for things like low disk space, we can use this to check periodically and send alerts for low memory conditions, too.

manthey:
Added in PR #588.

Issue #496, opened 2014-11-13
System consistency check


manthey:
We should add a system consistency check.  This would perform a variety of tasks:

For each model, ensure documents within that model meet basic requirements.  For instance, items would have a folderId and that folderId exists in the system.  To be pedantic, we could check that each document can validate for that model.

For files, items, folders, and base parents, ensure that the size is correct.

For assetstores, ensure that all files (or chunks) are used.  Check for files and chunks that are not tracked.

I see having a progress context for this, as for any collection of significant size, it will be slow.  We should probably prevent two consistency checks from running concurrently.

Questions:

If a document fails to have the correct references or doesn't validate, what do we do with it?  Do we automatically delete it?  Move it to a lost+found area?

If an assetstore has files that aren't used, do we delete them or add them to a lost+found item?  Similarly, if there is a file that isn't in the database, do we add it to the database, ignore it, or delete it?

danlamanna:
@manthey Did either of the referenced PRs resolve this?

manthey:
@danlamanna I think this was still open as we never decided on the answer to the questions I posed.  I'll close.  We can open new issues if we ever care.

PR #588, opened 2014-12-22
System check.

manthey:
This adds a GET system/check endpoint to report the overall system status.  This includes showing log locations and how much space is available on the logs' drive(s).  This adds some tracking and reporting of cherrypy threads.

This adds a PUT system/check endpoint to perform a consistancy check.  This checks to make sure all items have valid sizes and valid folderIds.  This should eventually be extended to check other models (see TODO list).

Fixed the plugin enable/disable test (it was expecting toggling one plugin would only affect one plugin, but because plugins can have dependencies, this isn't necessarily true).

Order json output by default.

Convert to using psutil for disk space on filesystem assetstores.  This should make the capacity work more generically (not just on linux).

Increased item test coverage.

This addresses issue #577 and starts to add features for issue #496.


cpatrick:
The better use of caching should be an issue that we consider, but it is way out of scope for this PR. LGTM! :ship:

PR #590, opened 2014-12-25
Anon group view bug

zachmullen:


cpatrick:
@zachmullen Should there be a Python test for this?

PR #593, opened 2014-12-30
Let CORS accept an X-Forwarded-Host.

manthey:
A common use case is to have girder behind a proxy.  In this case, X-Forwarded-Host is probably set.  If so, and it matches the Origin, then accept the data as from that origin, rather than triggering a cross-origin response.  Added some additional logging to help debug such issues.


zachmullen:
LGTM!

Issue #594, opened 2014-12-30
Issues when the primary on a mongo replica set changes


manthey:
If girder is using a mongo replica set for its database, and the primary changes within the replica set, girder stops communicating with the mongo database and doesn't recovery in a timely fashion.

I have made a vagrant file that will create a replica set within docker containers.  Asking the primary to stop exhibits this behavior.  We should be able to replicate this with multiple mongo servers on the test platform.  Perhaps adding a connection proxy (for instance https://github.com/arngarden/MongoDBProxy) would help.

manthey:
Addressed in PR #595.

Issue #438, opened 2014-10-25
GridFS assetstore adapter should have configurable URL


zachmullen:
Right now it's assumed to live at the same URL as the main girder database. In production, we may actually want it to live in a different mongo instance.

manthey:
Added in PR #595.

PR #595, opened 2015-01-01
Gridfs alternate host and mongo hardening

manthey:
Added mongo connection proxy from https://github.com/arngarden/MongoDBProxy to autoreconnect when the mongo connection is interrupted.  Added default socket timeouts to our mongo connection, as otherwise when a replica set member goes down we never get a chance to reconnect.

Excluded girder/lib/* from static analysis and from coverage.  The lib directory should only be used for external modules we can't install via pip.

Added support for GridFS assetstores to use a different mongo database than girder.  This adds a guard against connecting to an offline database.

Added a python file to create a mongo replica set for testing, and added replica set testing.  This includes testing reconnecting to a mongo server when it is disconnected.

This addresses issues #438 and #594.


manthey:
One aspect of this I'm not completely happy with is the socketTimeoutMS parameter used with mongo.  Without this parameter, we will never disconnect from a missing mongo server and try to reconnect to one that is working.  With this parameter, long, slow queries can timeout and thus fail.

I've picked a default of 60 seconds.  When it times out, the reconnect can take up to twice this long.

manthey:
If it is of use for anyone while testing, I've made two Vagrantfiles; one creates a mongo replica set, much like the mongo_replicaset.py configuration, but in docker containers.  The second creates a mongo sharded server with replica sets.  I don't know how useful the first (replica set) is unless you are accessing it from another virtulabox.  The sharded server should be usable from anywhere.  I can commit these somewhere, though I am not sure where they should go.

girder/lib/mongodb_proxy.py
```diff
@@ -0,0 +1,145 @@
+"""
```
zachmullen: What do you think about changing the "lib" dir to "ext" (or "external" or "tpl") instead? Seeing "lib" in that location might make people assume it's girder library code in there.

manthey: Sounds good.  I think I like 'external' best of those choices, as it is explicit.

zachmullen:
> If it is of use for anyone while testing, I've made two Vagrantfiles; one creates a mongo replica set, much like the mongo_replicaset.py configuration, but in docker containers. The second creates a mongo sharded server with replica sets. I don't know how useful the first (replica set) is unless you are accessing it from another virtulabox. The sharded server should be usable from anywhere. I can commit these somewhere, though I am not sure where they should go.

That's awesome, we should put those underneath `devops` somewhere.

zachmullen:
LGTM!

cpatrick:
:clap: :clap: :clap:

jeffbaumes:
Nice!

PR #596, opened 2015-01-02
Alphabetize the plugins on the admin page.

manthey:
Plugins were in no consistent order, and they could get shuffled by the addition of an additional plugin.

### DIFF
--- a/clients/web/src/templates/body/plugins.jade
+++ b/clients/web/src/templates/body/plugins.jade
@@ -16,7 +16,7 @@ div.g-plugin-restart
         | #{plugin.value.name}
         if (plugin.value.version)
           span.g-plugin-version
-            | Version #{plugin.value.version)
+            | Version #{plugin.value.version}
         if(plugin.value.configRoute)
           a.g-plugin-config-link(g-route="#{plugin.value.configRoute}",
                                  title="Configure Plugin")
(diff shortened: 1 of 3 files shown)


clients/web/src/templates/body/plugins.jade
```diff
@@ -8,17 +8,17 @@ div.g-plugin-restart
...
       .g-plugin-name
-        | #{plugin.name}
-        if (plugin.version)
+        | #{plugin.value.name}
+        if (plugin.value.version)
           span.g-plugin-version
-            | Version #{plugin.version)
-        if(plugin.configRoute)
-          a.g-plugin-config-link(g-route="#{plugin.configRoute}",
+            | Version #{plugin.value.version)
```
zachmullen: This `#{...)` notation seems to work, but looks wonky to me, can we use a closing curly brace?

manthey: That IS wonky.  A typo from when I first added it that peculiarly did no harm.  I'll fix it.

clients/web/src/views/body/PluginsView.js
```diff
@@ -79,6 +79,23 @@ girder.views.PluginsView = girder.View.extend({
         return this;
     },
 
+    _sortPlugins: function (plugins) {
+        /* Sort a dictionary of plugins alphabetically so that the appear in a
+         * predictable order to the user.
+         *
+         * :param plugins: a dictionary to sort.  Each entry has a .name
+         *                 attribute used for sorting.
+         * :returns sortedPlugins: the sorted list. */
```
zachmullen: Let's use jsdoc conventions in the javascript (@param, @returns)

zachmullen:
LGTM

PR #597, opened 2015-01-05
Move enable upload functionality into public API for upload widget

zachmullen:
This is used by COVALIC when we need to validate the input file set.

### DIFF
--- a/clients/web/src/views/widgets/UploadWidget.js
+++ b/clients/web/src/views/widgets/UploadWidget.js
@@ -144,7 +144,7 @@ girder.views.UploadWidget = girder.View.extend({
     filesChanged: function () {
         if (this.files.length === 0) {
             this.$('.g-overall-progress-message').text('No files selected');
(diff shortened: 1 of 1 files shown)


clients/web/src/views/widgets/UploadWidget.js
```diff
@@ -173,8 +173,8 @@ girder.views.UploadWidget = girder.View.extend({
     startUpload: function (e) {
         e.preventDefault();
 
+        this.setUploadEnabled(false);
         this.$('.g-drop-zone').addClass('hide');
-        this.$('.g-start-upload').addClass('disabled');
```
cpatrick: Why the change in order? I'm assuming it's harmless.

cpatrick:
LGTM.

clients/web/src/views/widgets/UploadWidget.js
zachmullen: I like to group together lines that look similar. Don't judge me. :stuck_out_tongue:

PR #600, opened 2015-01-06
Fixes #599. Implement recursive downloading in python client

### LINKED ISSUE #599: Girder client: implement folder download

zachmullen:


zachmullen:
@girder/developers @girder/owners Can anyone do a quick QA on this so I can get it turned around for our external user? Test script I used:

```
from girder_client import GirderClient

c = GirderClient('localhost', port=8080, apiRoot='/api/v1')
c.authenticate(username='zachmullen', interactive=True)
c.downloadFolderRecursive('54ac218a9e462c2c4a8f2cc9', dest='.')
```

Thanks!

clients/python/girder_client.py
```diff
@@ -423,3 +423,103 @@ def addMetadataToFolder(self, folderId, metadata):
...
+        :param name: If the item name is known in advance, you may pass it here
+            which will save a lookup to the server.
+        """
+        if name is None:
+            item = self.get('item/' + itemId)
+            name = item['name']
+
+        offset = 0
+        first = True
+        while 1:
```
manthey: I'd rather see `while True` than `while 1`.

clients/python/girder_client.py
```diff
@@ -423,3 +423,103 @@ def addMetadataToFolder(self, folderId, metadata):
...
+            files = self.get('item/%s/files' % itemId, parameters={
+                'limit': 50,
+                'offset': offset
+            })
+
+            if first:
+                if len(files) == 1 and files[0]['name'] == name:
+                    self.downloadFile(
+                        files[0]['_id'], os.path.join(dest, name))
```
manthey: The `name` (and the `file['name']` below) may not be valid path elements on the current os.  These should be sanitized.

clients/python/girder_client.py
```diff
@@ -423,3 +423,103 @@ def addMetadataToFolder(self, folderId, metadata):
...
+    def downloadItem(self, itemId, dest, name=None):
+        """
+        Download an item from Girder into a local folder. Each file in the
+        item will be placed into the directory specified by the dest parameter.
+        If the item contains multiple files or a single file with a different
+        name than the item, the item will be created as a directory under dest
+        and the files will become files within that directory.
+
+        :param itemId: The Id of the Girder item to download.
+        :param dest: The destination to write the item into.
```
manthey: I'd change the description of `dest` to state that it is the destination *directory*.

clients/python/girder_client.py
zachmullen: This brings up an interesting security question. Do you think we should disallow "." and ".." as resource names in the Girder hierarchy?

zachmullen: I plan to borrow [werkzeug's implementation of this sanitization](https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug/utils.py#L247) but I'm wondering if we should disallow this sort of thing on the server or not.

manthey: I don't it is necessary to prohibit those, as we need to sanitize resource names anyway.  Otherwise, we end up trying to prohibit other nonsense paths, which is OS dependent.  For instance, you can't have `%PATH%` inside the name of a windows file, but linux doesn't care.

zachmullen:
@manthey I added an implementation of transformFilename that isn't complete (i.e. not robust for Windows), but will at least prevent security problems related to absolute paths and `.` / `..`.

zachmullen:
I'll leave it to another issue to make that more useful on Win, since I don't have a machine to test on at this point.

manthey:
LGTM (when travis is happy).  It looks like the travis failure is just one of its intermittent memory issues.

PR #601, opened 2015-01-07
Use robust makedirs logic

zachmullen:
Also precompile the regex for name cleaning since we use it for
every file and directory.

### DIFF
--- a/clients/python/girder_client.py
+++ b/clients/python/girder_client.py
@@ -25,6 +25,9 @@ import re
 import requests
 
 
+_safeNameRegex = re.compile(r'^[/\\]+')
+
+
 class AuthenticationError(RuntimeError):
     pass
 
@@ -434,7 +437,7 @@ class GirderClient(object):
         name = name.replace(os.path.sep, ' ')
         if os.path.altsep:
             name = name.replace(os.path.altsep, ' ')
-        return re.sub(r'^[/\\]+', '', name)
+        return _safeNameRegex.sub('', name)
 
     def downloadFile(self, fileId, path):
         """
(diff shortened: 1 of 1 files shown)


clients/python/girder_client.py
```diff
@@ -434,7 +437,7 @@ def _transformFilename(self, name):
         name = name.replace(os.path.sep, ' ')
         if os.path.altsep:
             name = name.replace(os.path.altsep, ' ')
-        return re.sub(r'^[/\\]+', '', name)
+        return _safeNameRegex.sub('', name)
```
manthey: Do we really want to substitute the empty string (I realize that you didn't just add that)?  What if the name is just the restricted characters?  In the lines above, we substituted a space.

zachmullen: That's a good point, I'll go ahead and make it the underscore, since I'm not sure about having a filename that is entirely whitespace. :)

clients/python/girder_client.py
```diff
@@ -482,7 +485,11 @@ def downloadItem(self, itemId, dest, name=None):
                     break
                 else:
                     dest = os.path.join(dest, self._transformFilename(name))
-                    os.makedirs(dest)
+                    try:
+                        os.makedirs(dest)
+                    except:
```
manthey: Can you make this `except OSError as exception: if exception.errno != errno.EEXIST:` so we don't catch other exceptions? (see girder/__init__.py)

manthey:
LGTM.

I'd like to see tests added for the python client before it grows too much more.

zachmullen:
Agreed.

PR #603, opened 2015-01-07
Quota and assetstore policies for users and collections.

manthey:
Added support for preferred and fallback assetstores on users and collections.

Added user interface for quota and assetstore policies for users and collections.  This includes pie charts for capacity.  Non admins can check their own user and collection capacities.

Implemented quotas for total file size for users and collections.

Added a GirderException so that errors that aren't strictly RestExceptions, are failures, but aren't catastrophic, can be handled distinctly.  For instance, if the assetstore for a particular user is unavailable, there is not necessarily anything wrong with the system (just with the assetstore).  These have an optional 'identifier' that can get passed through to a rest consumer or other error handler.  This obviates the need to parse the error text to determine the type of error.

Updated documentation for new events.

Parallelized web_client tests.  Most web_client tests can now be run in parallel.  Those involving monog replicasets cannot.  Those that upload via the phantomjs hook in the specRunner cannot.  These ones could be parallelized by using distinct file names so as not to clobber one another.

Adjusted some doc strings to fix some sphinx warnings.

Added tests for quota and assetstore policies for both server and web client.

Format sizes the same way in javascipt and python.

Offer to repick files from the upload widget if it fails due to space.

Moved testUpload to the general javascript testUtils so that more than one test can use it.

### DIFF
--- a/clients/web/src/init.js
+++ b/clients/web/src/init.js
@@ -17,8 +17,10 @@ _.extend(girder, {
     models: {},
     collections: {},
     views: {},
-    apiRoot: $('#g-global-info-apiroot').text(),
-    staticRoot: $('#g-global-info-staticroot').text(),
+    apiRoot: $('#g-global-info-apiroot').text().replace(
+        'TESTURL', 'http://' + window.location.host),
+    staticRoot: $('#g-global-info-staticroot').text().replace(
+        'TESTURL', 'http://' + window.location.host),
     currentUser: null,
     events: _.clone(Backbone.Events),
     uploadHandlers: {},
@@ -89,6 +91,14 @@ _.extend(girder, {
                 } else if (error.status === 0 && error.statusText === 'abort') {
                     /* We expected this abort, so do nothing. */
                     return;
+                } else if (error.status === 500 && error.responseJSON &&
+                           error.responseJSON.type === 'girder') {
+                    info = {
+                        text: error.responseJSON.message,
+                        type: 'warning',
+                        timeout: 5000,
+                        icon: 'info'
+                    };
                 } else {
                     info = {
                         text: 'An error occurred while communicating with the ' +
--- a/clients/web/src/init.js
+++ b/clients/web/src/init.js
@@ -18,9 +18,9 @@ _.extend(girder, {
     collections: {},
     views: {},
     apiRoot: $('#g-global-info-apiroot').text().replace(
-        'TESTURL', 'http://' + window.location.host),
+        '%HOST%', 'http://' + window.location.host),
     staticRoot: $('#g-global-info-staticroot').text().replace(
-        'TESTURL', 'http://' + window.location.host),
+        '%HOST%', 'http://' + window.location.host),
     currentUser: null,
     events: _.clone(Backbone.Events),
     uploadHandlers: {},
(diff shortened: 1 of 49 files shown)


clients/web/src/utilities/MiscFunctions.js
```diff
@@ -36,24 +36,22 @@ girder.formatDate = function (datestr, resolution) {
 };
 
 /**
- * Format a size in bytes into a human-readable string with metric unit prefixes.
+ * Format a size in bytes into a human-readable string with metric unit
+ * prefixes.
  */
 girder.formatSize = function (sizeBytes) {
-    // If it's > 1GB, report to two decimal places, otherwise just one.
-    var precision = sizeBytes > 1073741824 ? 2 : 1;
```
zachmullen: The reason for this logic was that it made it a bit easier to see progress of uploads on files on the order of GB. In that case, if we report to only 1 decimal place, the number only changed every ~100MB, which can take a very long time while uploading.  Maybe this is OK though.

clients/web/src/init.js
```diff
@@ -17,8 +17,10 @@ _.extend(girder, {
     models: {},
     collections: {},
     views: {},
-    apiRoot: $('#g-global-info-apiroot').text(),
-    staticRoot: $('#g-global-info-staticroot').text(),
+    apiRoot: $('#g-global-info-apiroot').text().replace(
+        'TESTURL', 'http://' + window.location.host),
+    staticRoot: $('#g-global-info-staticroot').text().replace(
+        'TESTURL', 'http://' + window.location.host),
```
zachmullen: If we must do this replacement, I think we should indicate that this is a variable, and also decouple it from the notion of testing since it will affect non-testing environments as well. Something like `%HOST%` or `$HOST`.

clients/web/src/utilities/MiscFunctions.js
manthey: I see the point.

I think that I'd rather see a minimum number of digits of significance (say 4), so that until you cross the 100 Gb threshold, you'd have two decimal places.  That format looks poor with small values, since 2345 bytes ends up being '2.290 kB', so then you need further logic not to apply the reduction to values below 10000.

clients/web/src/init.js
manthey: Okay.  It gained me the power to parallelize the web client tests.

zachmullen: That is definitely worth it then :)

zachmullen:
This is looking very nice, one improvement I think we should make is to let the user specify the quota in two parts, one numeric and the other a unit. Most of the time these things will be on the order of GB or TB, so typing out the full byte value is a lot more work than typing "10" and selecting "GB" from a dropdown.

manthey:
Sure.  We also display the quota setting with our size formatting then.

Do we care about 1000-based prefixes versus 1024-based prefixes?  We are currently using 1024-based prefixes, which I prefer in this context.

zachmullen:
1024 all the way :)

jeffbaumes:
+1

manthey:
I made it so the dialog will show however much precision as it needs to (that is, 1073741824 would show as 1 Gb, whereas 1072693248 would show as 1023 Mb, and 1073741823 would show as 1073741823 bytes).  You can type it expressions like '1.5 Gb' and it will figure out what you meant (showing up as 1536 Mb when you go back into the dialog).  A typed unit suffix overrides the select dropdown.

manthey:
I've addressed issue #605 with the latest commit.

zachmullen:
LGTM. Great work.

PR #606, opened 2015-01-08
Job notifications

zachmullen:


zachmullen:
Ready for review.

girder/api/v1/notification.py
```diff
@@ -78,6 +77,9 @@ def streamGen():
                     wait = MIN_POLL_INTERVAL
                     start = time.time()
                     yield sseMessage(event)
+                if (cherrypy.engine.state != cherrypy.engine.states.STARTED or
+                        time.time() - start > timeout):
+                    break
```
manthey: Should the conditional about `cherrypy.engine.states.STARTED` still be in the `while` condition?  That is, if cherrypy isn't started, do we ever want to yield an sseMessage?

zachmullen: Fair point, I can move it there.

manthey:
Does anything use the JobDetailsWidget other than the test script?

Or, more generally, aside from the test functions in the plugin itself, is there any easy way to see this in action?

zachmullen:
For the moment, it's only used within the covalic plugin in cases where jobs failed. I intend for it to also be used when we add a UI for job management to the girder web client. It looks like this.

screen shot 2015-01-14 at 11 45 30 am

manthey:
I invoked it manually and fooled with it a bit.

This LGTM.

Issue #604, opened 2015-01-08
Create collection locks the page


zachmullen:
When the user is redirected to the new page, the page is obscured by the modal container and nothing can be clicked, even though the modal is not shown.

zachmullen:
Same thing happens when creating a group.

manthey:
Fixed in PR #607.

PR #607, opened 2015-01-08
Dialog container not always removed.

manthey:
When we dispose of certain dialogs, we weren't giving them a chance to fully clean up.  This could lead to the dialog container remaining present and consuming all clicks.  I've updated the tests so that they fail in this condition, and fixed the problem.  This fixes issue #604.


zachmullen:
LGTM, thanks for fixing what I broke. :smile:

manthey:
No problem.

PR #609, opened 2015-01-13
Some changes to improve builds.

manthey:
phantomjs occasionally fails to load javascript files.  This appears to be a known issue: https://github.com/ariya/phantomjs/issues/10652.  Retry several times if it looks like this has occurred.

This would occasionally show up in Travis.  I could repeat it about 1 time in 8 by have 'ctest -j 5' in an endless loop on an ubuntu test box.

Changed how the mock smtp loop is performed so as to not be using up 100% CPU.

### DIFF
--- a/clients/web/test/specRunner.js
+++ b/clients/web/test/specRunner.js
@@ -70,8 +70,7 @@ page.onConsoleMessage = function (msg) {
         } catch (e) {
             console.log('Exception writing coverage results: ', e);
         }
-    }
-    else {
+    } else {
         console.log(msg);
     }
     if (msg === 'ConsoleReporter finished') {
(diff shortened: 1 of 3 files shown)


zachmullen:
I was seeing the error message from blanket and assumed that the blanket default timeout of 180ms was simply not long enough, so I increased it to 2k on my other branch. I'm guessing this is the underlying cause?

manthey:
That wasn't the problem I actually saw (so increasing that timeout may help, too).

What I would see is that a script would fail to load.  cherrypy would send the script, but phantom would never receive it, regardless of timeout.  By setting a phantom resource timeout to 15 seconds and then logging an error, I catch the log message and try again.

zachmullen:
LGTM.

PR #610, opened 2015-01-14
Fix gitignore for plugin filtering

zachmullen:
The previous incarnation was overly heavy-handed such that new
files in packaged plugins did not show up in `git status`. This
version now makes all files in known plugins show correctly, and
still hides all untracked plugins.


cpatrick:
LGTM.

PR #611, opened 2015-01-14
Ensure boolean return value from _hasGroupAccess

zachmullen:
This isn't very impactful since it's just a private method, but it still can't hurt to make it consistent.


manthey:
LGTM.

Issue #427, opened 2014-10-23
On password reset, make old password still work


zachmullen:
And make the temporary expire after some smallish duration.

manthey:
Added in PR #612.

PR #612, opened 2015-01-14
Added a temporary access token as an alternative to resetting a password.

manthey:
To support testing this, we can now check email from the mock smtp server in the web client tests.

I did not remove the reset password option, but it is no longer accessible via the web client.

This addresses issue #427.

### DIFF
--- a/clients/web/src/app.js
+++ b/clients/web/src/app.js
@@ -181,6 +181,7 @@ girder.App = girder.View.extend({
         girder.router.navigate(route, {trigger: true});
 
         if (girder.currentUser) {
+            girder.eventStream.close();
             girder.eventStream.open();
         } else {
             girder.eventStream.close();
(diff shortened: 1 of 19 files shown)


zachmullen:
This works. I opened #615 which we can fix on a different topic, to account for redirects to the web app in cases where girder is not located under its normal server path. Mainly we just need some reusable utilities for querying:

1) The URL for the web API root,
2) The URL for the web client root

zachmullen:
The only bug I see is that when I click the link sent in the email, the top right part of the header does not indicate that I'm logged in. `girder.events.trigger('g:userChanged');` will hopefully fix that.

manthey:
I'm failing to reproduce the issue you mentioned.  That is, the top right part of the header properly changes to the appropriate user account when I follow the email link.

The code sends `girder.events.trigger('g:login-changed');`, which should rerender the user portion of the header.

What is different between our configurations?  What browser are you using?

zachmullen:
It's because I was in the covalic app served out of / instead of the girder app served out of girder. LGTM.

PR #617, opened 2015-01-15
Prevent an error when assetstores are mounted on different filesystems

brianhelba:


zachmullen:
Nice fix, did you run into this bug in the wild?

zachmullen:
LGTM.

manthey:
Nice.  We should always use `shutil.move` rather than `os.rename` is there is a chance of crossing file systems (pretty much if the directories might be different).

Issue #426, opened 2014-10-23
Allow keyboard selection on search drop downs


cjh1:
For example when adding user the group, when the user list gets populated I would like to be able to press the down arrow and select using enter ...

manthey:
Added in PR #619.

PR #619, opened 2015-01-15
Add keyboard selection to the search widget.

manthey:
This addresses issue #426.


zachmullen:
Very nice. Why the extra step required between the last element and going back to the first?

manthey:
So that way if we ever want enter to do something on the search field directly, you can get back to it.  Probably not that useful on a search field, but if we ever add an edit field with suggestions, it provides a way not to pick any of the suggestions, and when we add such a field the UI will be consistent.

zachmullen:
Ok. LGTM

Issue #614, opened 2015-01-14
Emails show incorrect URL


zachmullen:
I think this is due to being behind a proxy, I saw this from data.kitware.com. The text of the email was:

Your password has been reset. Your new password is: ...

If you did not initiate this password reset, please contact a site administrator.

This auto-generated email sent from Girder @ http://0.0.0.0:8080.

zachmullen:
We should fix #615 and use those utilities to fix this issue.

manthey:
I see two parts to this: (a) the host and port that the email is generated from, and (b) the url that is provided for the user to get back to girder.

The url (b) is generated from the host (or X-Forwarded-Host) and the base path for girder.  I think that is generated appropriately, but we could override it (which is what I interpreted issue #615 to be).  In the event that girder isn't served in the normal spot, this should move appropriately as well.  The only issue I think it would have is that if a plugin wants that callback to route somewhere other than the standard girder web client.

The host (a) is currently taken from the cherrypy local name.  The local name doesn't really look like it will ever be useful.  Since the goal of this field is to let the user know where the email originated, I think this has to be a setting, much like the email-from-address.

zachmullen:
As much as I prefer automated detection, I think you're right and we'll probably save ourselves headaches by simply making it a system setting (two system settings, to address #615).

cpatrick:
Cool! Now that we've decided to make it configurable, we can argue over the default. :smile_cat:

manthey:
The cherrypy local host is ([global] server.socket_host):([global] server.socket_port).  If unspecified, this can be the default (as it is currently).  There, I've started the argument. :crocodile:

manthey:
How much validation should we enforce on this value?  It is currently just shown as a link at the bottom of emails, so the value doesn't have to really be anything sensible, but we could choose to make sure it is of the form (schema)://(host).  We might want to be more liberal that that, too, since it could be used as a link back to an about or help page.

manthey:
Added is PR #620.

Issue #615, opened 2015-01-14
Uniform way to query for front-end URL


zachmullen:
Some of our instances will modify the path of the main girder web client to be something other than the webroot. When we send emails, for instance, we will need to know what that root is rather than assuming it's at /. I think this should be an app config property that a plugin will set whenever it moves the girder web app service path.

manthey:
Do we need to expose this further than adding a value to the cfg file (and using it appropriately for email)?  Would server.client_root be an appropriate key?

I note that we don't have fall-back values for api_root and static_root.  I would think that the utility/server.py script should have default values in case the cfg file si missing them.

zachmullen:
We need a way for plugins to dynamically update the property in cases where they relocate the core web app. The static specification of that root as a config parameter is somewhat orthogonal to that concern, though I could see it being useful. I need to think a bit more about how to do this right.

manthey:
We should probably have a test that relocates the core web app.  I know there are external projects that do so, but perhaps we should have a test plugin that does so, too.

zachmullen:
:+1:

zachmullen:
@manthey I forgot about it, but we already have such a test. https://github.com/girder/girder/blob/master/tests/cases/custom_root_test.py

manthey:
I had forgotten about that, too.  Interestingly, it is the one test that wasn't clearing its mongo database as it started.  I don't think this was causing any problems.

manthey:
Should we be using the Referer header to try to glean this information instead of using a setting for it (or at least as a fallback if the setting hasn't been changed)?

zachmullen:
That might be a good choice for the fallback.

PR #620, opened 2015-01-15
Added an EMAIL_HOST setting.

manthey:
This defaults to the cherrypy server.socket_host.  I've added the concept of a non-constant setting default.

This addresses issue #614.

### DIFF
--- a/clients/web/src/templates/body/systemConfiguration.jade
+++ b/clients/web/src/templates/body/systemConfiguration.jade
@@ -20,6 +20,12 @@ form.g-settings-form(role="form")
       placeholder="Default: #{defaults['core.email_from_address'] || ''}",
       title="The email address the system will use when sending emails.")
   .form-group
+    label(for="g-core-email-host") Email host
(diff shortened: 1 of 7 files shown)


zachmullen:
I think this same thing can be used to address part of #615 if we generalize it slightly. Rather than "email host", I think this setting should represent the root of the girder web client. We should use that as the value in our email footer, but can also use it in other places where we need to create URLs pointing to our web app.

manthey:
I'm not sure about that.  Here, it has some value as a distinct link that can direct an email recipient to somewhere other than girder.

For the URLs pointing to the web app, we can use the Host/X--Forwarded-Host as the host portion of the URL.  That way users in the LAN see a local network url and users in the WAN see an external url.  The root of the URL to the web app is needed as part of this, but doesn't have to be involved with the link shown in the email footer.

manthey:
Of course, we can merge the concepts if you think that these shouldn't be decoupled.

zachmullen:
Guess extra flexibility won't hurt. LGTM.

Issue #608, opened 2015-01-12
Folder level metadata displays on the wrong folder


mgrauer:
I observed this in 1.1.0, master 5996e43838de8d09a3dfe481f199fabc666ce56a.

Say you have a hierarchy like

-top
--mid (with metadata)
---bottom

Only `mid` has metadata. As I navigate through the hierarchy widget, I view `top`, which correctly shows no metadata, then click on `mid` which *incorrectly* shows no metadata.  If I refresh here, the metadata is correctly shown.  I click on `bottom`, the metadata is *incorrectly* shown.  If I refresh here, the metadata is correctly not shown.  

Seems like the metadata view in the hierarchy widget is not getting refreshed.

mgrauer:
I can look into this as I'm working with metadata lately, and will also look into item metadata for similar issues.

zachmullen:
Interesting, when I set up a three level hierarchy and attach metadata to the second level folder, it doesn't show up anywhere when I refresh the page. Probably a different manifestation of the same bug.

zachmullen:
In fact it seems the metadata doesn't get saved to the server, even though it appears in the client when I hit the check button.

mgrauer:
Aha, I was saving metadata on the folder via an API call.

zachmullen:
This looks like a server-side bug actually, the client made a correct-looking request when saving the metadata, it just didn't get saved persistently.

mgrauer:
In my case I saved metadata on the folder through the API, which correctly saved the metadata server-side and correctly returns the metadata when I GET the folder/id endpoint.

manthey:
Fixed in PR #621.

PR #621, opened 2015-01-15
Folders showed wrong metadata.

manthey:
We didn't change the resource associated with the metadata widget when changing folders.  I've fixed the bug and added a test that would have caught it.

This fixes issue #608.


zachmullen:
Fixed the issue I was seeing, LGTM.

PR #622, opened 2015-01-15
Fix a Grunt failure when no "server" section is defined in the local config

brianhelba:
The Grunt "shell:readServerConfig" task was failing with the error "TypeError: Cannot read property 'api_root' of undefined" when a local config file was provided without a "server" section. This fixes the issue.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -33,8 +33,8 @@ module.exports = function (grunt) {
         }
         try {
             var cfg = JSON.parse(stdout);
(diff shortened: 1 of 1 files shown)


brianhelba:
I've rebased this commit onto the lastest master, and pushed the branch directly to the upstream girder/girder repo.

Hopefully this will make Travis run.

I'll close the other PR.

zachmullen:
LGTM.

PR #626, opened 2015-01-16
README.rst: Fix title underline too short

msabramo:
This prevents the description from rendering nicely on PyPI.

Merge this and then do `python setup.py register` to update the description.

See
https://bitbucket.org/pypa/pypi/pull-request/60/fix-bb-214-make-rst-rendering-not-fail-for/diff


zachmullen:
Thanks Marc!

PR #631, opened 2015-01-16
Auto-detect the bad memory condition on travis so we can die early

zachmullen:


zachmullen:
@manthey PTAL

failure condition: https://travis-ci.org/girder/girder/builds/47300658
success condition: https://travis-ci.org/girder/girder/builds/47300595

manthey:
I had already done this on tangelohub:
```
    - vmstat 1 1
    # Terminate if in the bad memory condition
    - python -c "import sys;sys.exit(int(open('/proc/meminfo').read().split('Cached:')[1].strip().split()[0])<32)"
```

I don't know if you like this any better or worse than what you just did.  It has the virtue (or drawback) of not using an external file.

zachmullen:
I'm OK with either, which do you prefer?

manthey:
I don't have a strong preference.  Yours has a nicer error message and mine is more compact.  Either one is fine by me.  If you prefer your own, LGTM.

PR #633, opened 2015-01-19
Static service util

zachmullen:
@brianhelba This simplifies the code required to serve a static file from a cherrypy route. Now you can simply do

```
from girder.utility.server import staticFile


def load(info):
    info['serverRoot'].test_route2 = staticFile('/Users/zach/Desktop/Screen Shot 2015-01-14 at 11.45.30 AM.png')
```

### DIFF
--- a/clients/python/girder_client.py
+++ b/clients/python/girder_client.py
@@ -327,10 +327,10 @@ class GirderClient(object):
         Tests whether the passed in filepath exists in the item with itemId,
         with a name of filename, and with the same contents as the file at
         filepath.  Returns a tuple (file_id, current) where
-            file_id: id of the file with that filename under the item, or
-            None if no such file exists under the item.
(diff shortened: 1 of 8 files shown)


zachmullen:
From @brianhelba 
> Also note that "cherrypy.lib.static.serve_file" requires an absolute path to the file ( https://github.com/cherrypy/cherrypy/blob/3.3.0/cherrypy/lib/static.py#L44 ); so obviously I'll join the relative path to "girder.constants.ROOT_DIR" or "girder.constants.STATIC_ROOT_DIR" as appropriate, since one wouldn't hardcode a full absolute path in practice. Perhaps you might want to include this path joining as a convenience inside "serveStaticFile" and let callers just specify a relative path.

zachmullen:
ping @jbeezley

jbeezley:
:+1:

PR #634, opened 2015-01-20
Prevent our textareas from overflowing their containers

zachmullen:


cpatrick:
LGTM.

manthey:
You may want to check the style of, for instance, a collection with a description which consists of a very long string of non-white space characters.  It looks bad on Chrome even with this change.  I'll post a screenshot in a moment.

zachmullen:
I see the problem you mention, this change was merely meant to make it impossible to drag the textarea bounds beyond the modal dialog container. We should put in a separate fix for the wrapping issue.

manthey:
Sounds good.

PR #636, opened 2015-01-20
Harden build and js coverage

manthey:
If you wish to disable blanket (javascript) coverage and remove all effects of the blanket module, comment out the line in the Gruntfile that includes 'blanket_jasmine.js'.  You will have to rerun 'grunt init'.  In the tests/JavascriptTests.cmake, change the line that includes 'COVERAGE_FILE' to specify a blank entry for the coverage file.  You should also change the 'combine_report' lin in tests/JavascriptTests.cmake to 'combine_report_skip' to disable the threshold test, since, otherwise, that test will fail.

Added the ability to set the jasmine default timeout from the environment variable JASMINE_TIMEOUT (in milliseconds).

I shifted the ports used in Girder tests to ports below 32768.  By default, linux uses the range of ports shown in /proc/sys/net/ipv4/ip_local_port_range for ephemeral ports, and when an emphemeral port happens to line up with a test port, that test will fail.  Ubuntu (at least) defaults this port range to 32768-61000.  IANA feels that this port range should be 49152-65535.  We could have, instead, added our port range to /proc/sys/net/ipv4/ip_local_reserved_ports.

Another cause of build failures is when two tests start nearly concurrently and the mock SMTP or mock S3 servers attempt to allocate a port.  If they start staggered, we properly avoid a port conflict.  However, there is a brief opportunity for a race condition when two tests will attempt to reserve the same port.  I reduce port selection conflicts in the mock services by biasing the selection based on the test process's pid.

When we transitioned from one modal dialog to another, the old one still has some effect on the javascript up until the transition time ends (300 ms by default).  If we dismissed a second modal dialog within that time, sometimes bootstrap doesn't actually dismiss the second dialog properly (the backdrop was removed but not the dialog).  In theory, this bug had exposure beyond the tests.

Don't fetch the last chunk of an upload just to learn what the current chunk number or offset is.  This reduces the memory and database load.

### DIFF
--- a/clients/web/src/utilities/JQuery.js
+++ b/clients/web/src/utilities/JQuery.js
@@ -19,11 +19,16 @@
          * problem switching from one modal dialog to another. */
         if ($(this).is('.modal')) {
             /* We have to reach into the backbone modal object a little to see
-             * if we need to do anything. */
+             * if we need to do anything.  By turning off the fade as we
+             * remove the old dialog, the removal is synchronous and we are
+             * sure it is gone before we add the new dialog. */
             if ($(this).data('bs.modal') && $(this).data('bs.modal').isShown) {
+                var elem = $($(this).data('bs.modal').$element);
+                var hasFade = elem.hasClass('fade');
+                elem.removeClass('fade');
                 $(this).modal('hide');
+                elem.toggleClass('fade', hasFade);
             }
-            $(this).modal('removeBackdrop');
         }
         if (view !== 'close') {
             this.off().modal().find('[data-dismiss="modal"]')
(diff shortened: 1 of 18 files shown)


manthey:
@zachmullen when you get a chance, can you review this?  I think I'm done (at least for now).

My goal is to eliminate intermittent build errors (excepting, of course, the Travis memory problem).  In my overnight testing (`while true; do if ctest -j 5; then echo 'okay'; else echo 'failed'; fi; done`) on a 4-core ubuntu virtualbox, this went from a failure rate of about 1 in 8 to a failure rate of about 1 in 100.

The remaining failures seem to occur when the host computer's system load is high (not a load on the virtualbox, so I'm assuming the virtualbox becomes slightly starved for resources), and are not particularly consistent -- they seem to be phantom errors and some very obscure timing issues.  I'd love to fix them, but they are rare enough that debugging them is tedious.

zachmullen:
Wow, that's awesome. This LGTM.

PR #637, opened 2015-01-20
In general, default to break-word wrapping behavior

zachmullen:
This fixes the issue described by @manthey earlier. Now all elements
will default to word-wrap break-word behavior.


manthey:
LGTM.

PR #638, opened 2015-01-21
Added a markdown library

zachmullen:
This is currently only used in the COVALIC plugin, but so far it's working quite nicely there.

### DIFF
--- a/clients/web/src/stylesheets/layout/global.styl
+++ b/clients/web/src/stylesheets/layout/global.styl
@@ -88,6 +88,9 @@ ul.dropdown-menu
 .modal-backdrop.in
   opacity .35
 
+blockquote p
+  font-size inherit
+
 /* Don't show the vertical scroll bar when we show a dialog unless we must */
 .modal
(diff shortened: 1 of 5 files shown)


clients/web/src/stylesheets/layout/global.styl
```diff
@@ -88,6 +88,9 @@ ul.dropdown-menu
 .modal-backdrop.in
   opacity .35
 
+blockquote p
+  font-size inherit
```
cpatrick: I'm not sure how this relates?

cpatrick:
Other than the question about the additional stylus, this LGTM.

clients/web/src/stylesheets/layout/global.styl
zachmullen: Bootstrap has a whole bunch of default element styling, among which is setting `blockquote p` to have a font size of 17.5px, which is much larger than the other text rendered in the markdown. This overrides that because it's silly.

PR #641, opened 2015-01-21
Minor text bugfix

zachmullen:


manthey:
LGTM.  Strangely we didn't make the same error in other models.

PR #642, opened 2015-01-21
Fixing testing problems that occur in asyncore loop

jbeezley:
While running tests locally, I get messages like this popping up:
```python
15: Exception in thread Thread-3:
15: Traceback (most recent call last):
15:   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 810, in __bootstrap_inner
15:     self.run()
15:   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 763, in run
15:     self.__target(*self.__args, **self.__kwargs)
15:   File "tests/mock_smtp.py", line 69, in loop
15:     asyncore.loop(timeout=1, use_poll=True)
15:   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/asyncore.py", line 216, in loop
15:     poll_fun(timeout, map)
15:   File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/asyncore.py", line 145, in poll
15:     r, w, e = select.select(r, w, e, timeout)
15: error: (9, 'Bad file descriptor')
```
It appears to be the same issue mentioned [here](http://bugs.python.org/issue10878).

The tests pass, but file descriptors aren't closed causing the ctest process to hang eventually.  This seems to fix the problem on my end.

### DIFF
--- a/tests/mock_smtp.py
+++ b/tests/mock_smtp.py
@@ -64,7 +64,12 @@ class MockSmtpReceiver(object):
         exit.
         """
         while len(asyncore.socket_map):
-            asyncore.loop(timeout=1, use_poll=True)
+            try:
+                asyncore.loop(timeout=1, use_poll=True)
+            except asyncore.select.error as e:
+                if getattr(e, 'args', [None])[0] == 9:
+                    return
+                raise e
 
     def stop(self):
         """Stop the mock STMP server"""


tests/mock_smtp.py
```diff
@@ -64,7 +64,12 @@ def loop(self):
         exit.
         """
         while len(asyncore.socket_map):
-            asyncore.loop(timeout=1, use_poll=True)
+            try:
+                asyncore.loop(timeout=1, use_poll=True)
+            except asyncore.select.error as e:
+                if getattr(e, 'args', [None])[0] == 9:
+                    return
+                raise e
```
ronichoudhury: I think this can just be

    raise

jbeezley:
Now that I tried running the full test suite again, ctest again hangs, but it lasts longer.  So, I don't think this fixed my issue and isn't worth merging.  I am curious if anyone else has run into the errors above.  My setup is OSX 10.10, homebrew python 2.7.9.

mgrauer:
@jbeezley I also run into those errors, on OS X 10.8.5, virtual env, Python 2.7.2 .

zachmullen:
I consistently see them as well; in my fiddling I was able to eat the exception, but have no idea whether that was a good idea or not. :fried_shrimp:

jbeezley:
Probably an OSX specific issue.  As I mentioned, I can prevent the errors from appearing, but whatever file handles are opened aren't closing.  Eventually the maximum file handles per process is overrun and the ctest process hangs.  @manthey I can create an account on my mac if you want to look into it.

manthey:
I have occasionally seen these errors in Ubuntu.  My understanding is that they occur because the asyncore thread is torn down after the smtp thread, and it shouldn't be causing any direct problem.  Eating the exception (as @jbeezley had done) cleans up the output.

I don't know how to measure it on OSX, but in ubuntu the full ctest run tops out with a maximum of around 600 or 700 open files (the delta of the max /proc/sys/fs/file-nr and the value it has when not running the tests).  Perhaps an equivalent measure can be made on the Mac.

zachmullen:
Maybe I need some schooling on how Linux manages file handles wrt subprocesses... CTest ``forks`` a separate subprocess for each individual test, so I assumed that when the subprocesses exited, the system would clean up file handles owned by the subprocess even if the parent process was still running. Is that not the case?

jbeezley:
To be honest, I have no idea why any of this is happening.  I'm just reporting what is happening.

zachmullen:
https://github.com/Kitware/CMake/blob/master/Source/CTest/cmProcess.cxx
https://github.com/Kitware/CMake/blob/master/Source/kwsys/ProcessUNIX.c

ProcessUNIX.c is pretty dense but from what I can gather it's a pretty typical `fork()` -> `execvp()` situation.

PR #643, opened 2015-01-21
The event stream was eating exceptions from the callback handlers

zachmullen:


manthey:
Should we have a test to make sure callback exceptions show up the way we expect?

zachmullen:
I might punt on that since this mostly affects development, and I think it would involve adding some infrastructure for introspecting exceptions in our test environment. Unless you know of some straightforward way to do that already?

manthey:
I think there is a way to test it without too much trouble.  I'll take a look.

manthey:
I just pushed a test that properly passes with this fix and fails without it.  It involves reaching into the girder object and doing somewhat obtuse things with jasmine.

zachmullen:
LGTM if the rest looks good to you.

manthey:
Your part LGTM.  Merge away when you think travis is happy enough.

clients/web/test/spec/widgetsSpec.js
```diff
@@ -52,6 +54,35 @@ describe('Test widgets that are not covered elsewhere', function () {
...
+            var stream = girder.events._events['g:navigateTo'][0].ctx.progressListView.eventStream;
+            stream.on('g:error', function () { errorCalled += 1; });
+            stream.on('g:event.progress', function () {
+                throw 'intentional error';
+            });
+            _setProgress('success', 0);
+        });
+        waitsFor(function () {
+            return onMessageError == 1;
+        }, 'bad progress callbcak to be tried');
```
jbeezley: I'll be the designated spell checker.  :smirk_cat: 

callbcak here and at line 80

manthey: Thanks.

PR #644, opened 2015-01-22
Cookie-based authentication options for route handlers

### LINKED ISSUE #639: Support cookie-authenticated downloads
We should make an exception to our normal authentication requirements for the download endpoints by allowing cookie-based auth so that it's easy for people to provide direct download links to private data that are consumable by web browsers. Since they're read-only operations, this won't be problematic wrt CSRF.

My use case is for COVALIC, where users who have access to a phase will be able to see user-created text containing download links to data that may be private.

zachmullen:
By default, the token value in the cookie is not sufficient to
authenticate to Girder (to protect against CSRF). There are a few
read-only endpoints (i.e. not vulnerable to CSRF attacks) that
make sense to open up to cookie-based authentication to support
user agent (browser) tools that make it infeasible to pass the normal
Girder-Token custom HTTP header since they are not requested with
normal XHR mechanisms. Namely:

* The notification stream endpoint, since in our web client it is
requested with an EventSource that does not allow specification of
headers in its API at this time.
* The download endpoints, since we want those links to download
as attachments to the user's filesystem and therefore have no
good way of specifying headers.

By allowing cookie auth for these endpoints, we also avoid placing
secure token values into query strings that might end up in
access logs on the server. Additionally, this allows us to simply
use the href property for download direct links, making it easy
for users to copy and disseminate direct download links, even to
private data.

### DIFF
--- a/girder/api/v1/item.py
+++ b/girder/api/v1/item.py
@@ -242,8 +242,7 @@ class Item(Resource):
     @loadmodel(model='item', level=AccessType.READ)
     def download(self, item, params):
         """
-        Defers to the underlying assetstore adapter to stream the file or
-        file out.
+        Defers to the underlying assetstore adapter to stream the file out.
         """
         offset = int(params.get('offset', 0))
         user = self.getCurrentUser()
@@ -256,6 +255,7 @@ class Item(Resource):
             return self.model('file').download(files[0], offset)
         else:
             return self._downloadMultifileItem(item, user)
+    download.cookieAuth = True
     download.description = (
         Description('Download the contents of an item.')
         .param('id', 'The ID of the item.', paramType='path')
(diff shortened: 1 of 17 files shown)


girder/api/v1/item.py
```diff
@@ -242,8 +242,7 @@ def getFiles(self, item, params):
     @loadmodel(model='item', level=AccessType.READ)
     def download(self, item, params):
         """
-        Defers to the underlying assetstore adapter to stream the file or
-        file out.
+        Defers to the underlying assetstore adapter to stream the file out.
```
manthey: I suspect the original comment was meant to read "the file or files out".

manthey:
LGTM.

PR #646, opened 2015-01-23
Improvements to model methods

brianhelba:
This contains 8 commits, all improving the implementations of Girder models. I can split some of this up into seperate PRs if requested.

### DIFF
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -542,8 +542,7 @@ class Folder(AccessControlledModel):
             'parentId': folder['_id'],
             'parentCollection': 'folder'
         }, fields=(), limit=0, timeout=False)
-        for subfolder in folders:
-            count += self.subtreeCount(subfolder)
+        count += sum(self.subtreeCount(subfolder) for subfolder in folders)
 
         return count
 
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -356,10 +356,8 @@ class Folder(AccessControlledModel):
         }
         q.update(filters)
 
-        cursor = self.model('item').find(
+        return self.model('item').find(
             q, limit=limit, offset=offset, sort=sort, **kwargs)
-        for item in cursor:
-            yield item
 
     def childFolders(self, parent, parentType, user=None, limit=50, offset=0,
                      sort=None, filters=None, **kwargs):
@@ -397,10 +395,9 @@ class Folder(AccessControlledModel):
         # afterward.
         cursor = self.find(q, limit=0, sort=sort, **kwargs)
 
-        for r in self.filterResultsByPermission(cursor=cursor, user=user,
-                                                level=AccessType.READ,
-                                                limit=limit, offset=offset):
-            yield r
+        return self.filterResultsByPermission(
+            cursor=cursor, user=user, level=AccessType.READ, limit=limit,
+            offset=offset)
 
     def createFolder(self, parent, name, description='', parentType='folder',
                      public=None, creator=None, allowRename=False):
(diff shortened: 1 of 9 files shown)


girder/models/item.py
```diff
@@ -297,26 +268,43 @@ def filterResultsByPermission(self, cursor, user, level, limit, offset,
...
+        :param removeKeys: List of keys that should be removed from each
+                           matching document.
+        :type removeKeys: list
+
         """
         # Cache mapping folderIds -> access granted (bool)
-        folderCache = {}
-        count = skipped = 0
-        for result in cursor:
+        folderAccessCache = {}
```
zachmullen: Need a blank line here to satisfy pep8.

brianhelba:
It looks like there are a few problems. I'll fix and resubmit.

@zachmullen if I amend some of the commits in this branch, then re-push to the same branch name, will that trigger a new Travis build?

cpatrick:
@brianhelba it will.

zachmullen:
I can't figure out why it's saying it received multiple spec kwargs, but it's happening on my machine too. Wondering if it's a bug within pymongo, going to dig into it a little to see.

brianhelba:
@zachmullen it's my mistake: it should be named "spec_or_id". I'm going to just call find_one with a positional argument instead of a keyword argument. I'll resubmit soon.

brianhelba:
Pushed again. New build in progress...

brianhelba:
Travis comes back with: "!!! ABORTING, LOW CACHE DETECTED: 0 !!!"

What's the proper way to force a new build?

brianhelba:
I've removed the more impactful commits that were causing failures. I'll fix and resubmit them in separate PRs for closer review.

This PR is now ready for review / merge.

girder/models/folder.py
```diff
@@ -309,7 +309,6 @@ def remove(self, folder, progress=None, **kwargs):
             if progress:
                 progress.update(increment=1, message='Deleted item ' +
                                 item['name'])
-        items.close()
```
zachmullen: Why remove the close here? I think we were trying to free the cursor early since the impending operations might take a while.

girder/models/user.py
```diff
@@ -187,10 +187,9 @@ def search(self, text=None, user=None, limit=50, offset=0, sort=None):
         else:
             cursor = self.find({}, limit=0, sort=sort)
 
-        for r in self.filterResultsByPermission(cursor=cursor, user=user,
-                                                level=AccessType.READ,
-                                                limit=limit, offset=offset):
-            yield r
+        return self.filterResultsByPermission(
```
manthey: In the function documentation, we claim that we return a "List of users", and now we return a cursor with the result set of users.  We probably should change that.  Strangely, this is the only function that was so modified that even has documentation on the return parameter.

zachmullen: Yeah, we should make them all say returns an iterable of documents.

brianhelba: @manthey @zachmullen  Good point, I will fix this.

girder/models/folder.py
brianhelba: That's understandable. I'll change this to either comment on the reason for the close, or refactor to make it automatically close sooner (I need to see if this is possible).

zachmullen: In cpython it will be closed via GC after going out of scope, so you'd have to put each in its own function.

manthey: The cursors will each hold a read lock on the accessed elements of their collections.  This will block a write operation on those collections from completing until the function ends and the cursors are released.  Closing them is better.  Any time we do a significant amount of processing after finishing with a cursor, we should close it rather than waiting for it to be closed when it goes out of scope.

brianhelba:
@manthey @zachmullen I've updated this PR to incorporate your feedback. I've also rebased it on the latest master. The tests are all passing, so it's ready for review.

manthey:
LGTM.  @zachmullen may have comments.

zachmullen:
LGTM.

PR #647, opened 2015-01-23
Harden phantom timeouts

manthey:
Improved restarting tests when phantomjs times out.  Phantomjs times out occasionally due to a bug.  We can retry tests when this happens, but we need to reset the databases and other services between tests.

When we added a test to check on callbacks throwing an exception, we needed to adjust one of the tests for progress notifications going away.


manthey:
@zachmullen I noticed that the test I added in your PR #643 made one of the web client tests more likely to fail.  This fixes that.  It also restarts phantom timeouts better, so maybe we'll get a good cdash result.

zachmullen:
LGTM once travis is happy.

PR #649, opened 2015-01-26
Fix javascript coverage reporting in aiur.

manthey:
Whenever we add coverage of the javascript of a plugin, we needed to copy the templates.js file to a particular location.  I've attempted to automate and generalize this so that next time adding coverage won't break the js coverage report.

I'm not sure I can truly test this outside of aiur.


zachmullen:
LGTM.

PR #650, opened 2015-01-27
Add a model removal event that also passes kwargs down

zachmullen:
This supports callbacks that need to handle progress, or any other
special arguments to remove(). The old event is still triggered
to preserve backward compatibility.

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -274,7 +274,12 @@ class Model(ModelImporter):
 
         event = events.trigger('.'.join(('model', self.name, 'remove')),
                                document)
-        if not event.defaultPrevented:
+        kwargsEvent = events.trigger(
+            '.'.join(('model', self.name, 'remove_with_kwargs')), {
+                'document': document,
+                'kwargs': kwargs
+            })
+        if not event.defaultPrevented and not kwargsEvent.defaultPrevented:
             return self.collection.remove({'_id': document['_id']})
 
     def removeWithQuery(self, query):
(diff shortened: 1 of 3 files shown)


girder/models/model_base.py
```diff
@@ -274,7 +274,12 @@ def remove(self, document, **kwargs):
 
         event = events.trigger('.'.join(('model', self.name, 'remove')),
                                document)
-        if not event.defaultPrevented:
+        kwargsEvent = events.trigger(
+            '.'.join(('model', self.name, 'remove_with_kwargs')), {
+                'document': document,
+                'kwargs': kwargs
+            })
+        if not event.defaultPrevented and not kwargsEvent.defaultPrevented:
```
manthey: if the model.(model).remove event prevents the default, should the model.(model).remove_with_kwargs event be triggered?

Is there any particular rationale for remove coming before remove_with_kwargs rather than the other way around?

zachmullen: > if the model.(model).remove event prevents the default, should the model.(model).remove_with_kwargs event be triggered?

Yes -- the default is the line below: `self.collection.remove({'_id': document['_id']})`.

> Is there any particular rationale for remove coming before remove_with_kwargs rather than the other way around?

The order is arbitrary, they are simply two variants of the same event. The new one simply accommodates the lack of forward compatibility in the old signature. Plugins needing to handle model removal events would bind to one or the other but not both.

zachmullen: It's more a question of the behavior of stopPropagation, but it is my sincere hope that nobody ever calls that on a model removal event since I can't imagine that would ever be a good idea.

manthey:
LGTM.

Issue #659, opened 2015-01-28
Design and implement private user behavior


zachmullen:
Currently we don't even expose the creation of private users or editing of user ACLs via the REST API, it can only be done programmatically at this time. We now have plugin developers attempting to use private users, which means we need to answer a few questions about how they should behave.

* Should users be able to have their own ACL controlled through a widget just like other resource types, or should web API-level control be more specialized?
* If a user is private but generates data that is visible to people who do not have read access to the user, what appears in the "created by" field or any other field referencing the user? If the user is in an ACL for a resource, should admins over that resource be able to see them in the list even if they don't have read access to the user?
* If a private user is in a group, should all members (or moderators or admins) of the group automatically be able to see that user? How will this be managed by the backing ACL for the user?

mgrauer:
What are the driving use cases for private users?  Who do they need to be private from?

zachmullen:
@brianhelba can speak to his requirements for private users. We also need to consider how they might be useful (or at least sensible) in general.

brianhelba:
Currently, we're using private users in 2 ways:
1. As pre-configured test / demonstraton users. When first enabled, plugin creates a number of test users, with each corresponding to a different user role. Since the test users will be operating on the live database / instance, it's useful to hide them from the main user pool to prevent confusion. These users should have first-class access to all of the normal web API when a user is logged in as one of them, but other public users generally have no need to know about them. In theory, site admins may want to add / remove them from groups or grant / revoke their access to resources, but even that's not currently required.

2. As the internal owner / admin of most of our resources. Normal user interaction with Girder resources is very limited: some may upload to a specificed collection and some may download from other collections, depending on their group membership. Our resource organization is extremely sensitive to state, so ordinary users are forbidden from editing any resources via normal mechanisms. Users modify the site via a set of seperate endpoints which our plugin provides. Internally, these endpoints do extensive validation of user input, then they directly use models to make all modifications. Having this admin user be private makes the user list more friendly to ordinary users, as all of the other users that they see are real people. However, since all resources are owned / created by the admin user, it would be acceptible either if the admin user's name was displayed in the item metadata or if those metadata fields were not displayed at all (in the web client) to ordinary users.

Neither of these cases strictly requires a private user for security / privacy purposes, but since it's nice to hide some of the gory details from our users (who are  very non-technical), private users seemed like a nice convenience.

brianhelba:
[...]

PR #660, opened 2015-01-28
Private user upload support

zachmullen:


zachmullen:
@brianhelba ping.

You will probably still run into a bunch of undefined behavior wrt private users until #659 is addressed.

brianhelba:
LGTM.

The always-forced behavior will be very helpful for my use case, as I'm doing some storage / fetching of files entirely interally via direct use of the models. I'm happy to have authorization checks only occur in the HTTP API or when I expicitly call "Item.hasAcccess" internally.

PR #663, opened 2015-01-29
Fix some style and spelling issues

brianhelba:
* Docstrings should use double quotes, per PEP 257
* Adjacent string literals don't need to be concatinated with "+"
* Fix spelling errors

### DIFF
--- a/clients/web/src/app.js
+++ b/clients/web/src/app.js
@@ -90,7 +90,7 @@ girder.App = girder.View.extend({
 
     /**
      * Close any open dialog if we are already logged in.
-     * :returns loggedIn: true if we have a current user.
+     * :returns: true if we have a current user.
      */
     closeDialogIfUser: function () {
         if (girder.currentUser) {
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -60,7 +60,7 @@ and/or ``removeFields`` keys, e.g.
 Geospatial
 ----------
 
-The geospatial plugin enables the storage and querying of `GeoJSON <http://geojson.org>`_
+The geospatial plugin enables the storage and querying of `GeoJSON <http://geojson.org>`__
 formatted geospatial data. It uses the underlying MongoDB support of geospatial
 indexes and query operators to create an API for the querying of items that
 either intersect a GeoJSON point, line, or polygon; are in proximity to a
@@ -68,7 +68,7 @@ GeoJSON point; or are entirely within a GeoJSON polygon or circular region. In
 addition, new items may be created from GeoJSON features or feature collections.
 GeoJSON properties of the features are added to the created items as metadata.
 
-The plugin requires the `geojson <https://pypi.python.org/pypi/geojson/>`_
+The plugin requires the `geojson <https://pypi.python.org/pypi/geojson/>`__
 Python package, which may be installed using **pip**: ::
 
     pip install -r plugins/geospatial/requirements.txt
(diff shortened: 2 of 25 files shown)


manthey:
Should we add flake8-docstrings into the requirements-dev.txt file?  It claims to use PEP257.  My concern is that doing so would complain for missing docstrings, long docstrings, and lots of other conventions we haven't used.  Alternately, is there a different way to add our desired formatting into the flake8 check?

cpatrick:
This calls down to pep257 and that supports ignoring specific things. I vote we switch it on and see how we do.

https://github.com/GreenSteam/pep257

manthey:
We will do very badly.  We will throw errors D10[012], D20[0345], D40[01].

cpatrick:
I would tend toward adding it with those exceptions and then creating
issues for removing those exceptions. We can then decide at some point if
we want to deal with them or explicitly not deal with them.

-Patrick

manthey:
@brianhelba Would you like me to push a commit that checks some of PEP257 as part of the flake8 tests?

brianhelba:
@manthey that sounds good.

brianhelba:
Since this still an open branch, I've rebased it on the latest master and pushed more style fixes on top.

brianhelba:
@manthey Good points about consistancy. I think we should add an example function with a exemplar docstring to the developer documentation.

As for nameed return values, the most important issue is that they're not parsing correctly. See:
* https://github.com/girder/girder/blob/v1.2.1/girder/models/setting.py#L247
* http://girder.readthedocs.org/en/v1.2.1/api-docs.html#girder.models.setting.Setting.corsSettingsDict
Where the return description is not included in the same line as the "Returns:", as it should be. 

In principle, I'd still disagree with the use of named return values. While some languages have them, it's not really a "thing" in Python. Also, Girder tends towards abbreviated less-verbose variable names (you kind of have to when you're using 4-space indents and an 80-column limit). So any pretend variable names that we did create for function return values would probably always be less useful than a short text description of it's content and formal type information.

I think the best plan going forward is to make an examplar docstring that covers all the cases we can think of, do our bikeshedding on that, then just attempt to follow it for new development.

manthey:
I see the issue with a returns named value.  I'm fine not having them.

@zachmullen `:param (type) (parameter)` does work.  We don't need consistency in the functions; the Sphinx result is the same either way (I just checked).

I agree, we should have a examplar docstring.

manthey:
I saw a few other minor issues in the rst files, plus I noticed that the branch would no longer auto-merge.  @brianhelba Sorry for pushing to your PR.

brianhelba:
@manthey is this ready to merge?

manthey:
Yes.  LGTM (assuming you like the changes I added).

brianhelba:
@manthey Changes are fine with me in theory, though I didn't verify that the RST is actually fixed or that the flake8-docstrings module actually works.

Anyone can feel free to merge.

manthey:
@brainhelba If you pull my changes, make sure you've run `pip install -r requirements-dev.txt` in your build environment, then when you `ctest` the flake8-docstrings module will be included.  Typing `flake8 --version`, you should see `pep257` included in the version information.

`grunt docs` should show no errors and only a warning about a static path.

brianhelba:
@manthey I probably won''t have time today to test this locally. If we trust Travis, or if someone else wants to provide a code review / verification, that will probably get this PR resolved sooner.

brianhelba:
@manthey I probably won't have time *today*. If it's still around tomorrow, I can test it. (But I'd really be fine if we just merged it now).

PR #664, opened 2015-01-29
Jasmine timeouts and retry

manthey:
For automated builds (travis, dashboard), set the Jasmine timeout to 15 seconds.  I've noticed that when the build computer is loaded or has none of the necessary files in the file system cache, that builds succeed with larger timeouts.

Also, if we don't get a report that Jasmine tests were finished, retry the test.  This appears to be a Jasmine bug (at least, the failure usually occurs within the Jasmine javascript).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ script:
     - mkdir _build
     - cd _build
     - cmake ..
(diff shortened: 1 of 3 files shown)


cpatrick:
LGTM!

PR #665, opened 2015-01-29
Gravatar plugin & model filtering utilities

zachmullen:


girder/models/item.py
```diff
@@ -45,17 +45,13 @@ def initialize(self):
...
-                'baseParentId']
-
-        filtered = self.filterDocument(item, allow=keys)
+        self.exposeFields(level=AccessType.READ, fields=(
+            '_id', 'size', 'updated', 'description', 'created', 'meta',
+            'creatorId', 'folderId', 'name', 'baseParentType', 'baseParentId'))
 
-        return filtered
+    def filter(self, item, user=None):
+        """Filter an item document for display to the user."""
```
cpatrick: Could you add the backward compatibility note here as well?

cpatrick:
This is looking good so far. The addition of `SITE_ADMIN` is cool and the gravitar plugin is fantastic. Tests and docs and this will be great :).

zachmullen:
This is now ready for a review. It doesn't actually show anything in the core girder web client at this point, but does extend the javascript user model to add `getGravatarUrl()` so clients can easily add gravatars.

cpatrick:
Add a recipe to the developer cookbook for using this in a plugin or external app?

zachmullen:
Added some docs.

cpatrick:
[LGTM](http://www.lgtm.in/i/2FGiyeuSp)

PR #666, opened 2015-01-29
Adding more security documentation.

cpatrick:


docs/development.rst
```diff
@@ -129,6 +129,26 @@ you can find the HTML output from the coverage tool in the source directory
...
+accepting a PR by saying it "Looks good to me" or LGTM. At this point, the
+topic branch can be merged to master. This is meant to be a simple,
+low-friction process; however, code review is very important. It should be done
+carefully and not taken lightly. Thorough code review is a crucial part of
+developing quality software. When performing a code review, ask the following:
+
+1.  Is the continuous integration server happy with this?
+2.  Are there tests for this feature or bug fix?
+3.  Is this documented (for users and/or developers)?
+4.  Are the commits modular with good notes.
```
zachmullen: Should be  question

docs/security.rst
```diff
@@ -60,3 +60,70 @@ Girder returns an error when a Cross-Origin request is made (one with the
 ``Origin`` header) that does not match the system configuration settings.
 Although most modern web browsers also enforce this, some additional security
 is added by enforcing it at the request level.
+
+
+Database Injection Attacks
+--------------------------
+
+Girder defends against database injection attacks by using PyMongo as the only
```
brianhelba: I don't think this paragraph is accurate. It is indeed possible to write injection-vulnerable code with PyMongo. See something like:
def exploitMe(operation, database, collection, data):
    """ This is a super-helpful general-purpose database interaction function. What could possibly go wrong?"""
    getattr(pymongo.MongoClient()[database][collection], operation)(data)

This should read something like (feel free to rephrase):

Girder is defends against database injection attacks by using MongoDB, which is immune to traditional SQL injection attacks. See the <link to Mongo FAQ> for more information on this. Additionally, Girder uses a database model layer to mediate and validate all interaction with the database. This ensures that for all database operations, structural attributes (collection name, operation type) are hardcoded and not modifiable by the client, while data attributes (stored content) are validated for proper form before being accepted from a client.

cpatrick: Good point. I'll add a bit of language that clarifies this.

docs/security.rst
```diff
@@ -60,3 +60,70 @@ Girder returns an error when a Cross-Origin request is made (one with the
...
+pathway between the application server and the database server. PyMongo--like
+most MongoDB client libraries build up a BSON object before sending the query
+to the database. This prevents most injection attacks. See the
+`MongoDB Documentation
+<http://docs.mongodb.org/manual/faq/developers/#how-does-mongodb-address-sql-or-query-injection>`_
+for more information.
+
+Additionally, we strongly recommend running MongoDB with JavaScript disabled
```
brianhelba: Change to: "...strongly recommend configuring your MongoDB server with JavaScript..." to make clear that this is not an option within Girder, but rather must be done by the server admin.

docs/security.rst
```diff
@@ -60,3 +60,70 @@ Girder returns an error when a Cross-Origin request is made (one with the
...
+see the `MongoDB Documentation
+<http://docs.mongodb.org/manual/faq/developers/#javascript>`_
+for more information.
+
+
+Session Management
+------------------
+
+Girder uses session management performed through the Girder-Token header or
+through a token passed through a GET parameter. This token is passed through
```
brianhelba: Change to: "This token is provided to the client through" to be more clear about which direction accepts cookies.

docs/security.rst
```diff
@@ -60,3 +60,70 @@ Girder returns an error when a Cross-Origin request is made (one with the
...
+then passed in the request as part of the Girder one-page application and other
+clients such that the cookie alone is not enough to form a valid request. A
+sensible CORS policy (discussed above) also helps mitigate this attack vector.
+
+
+Dependent Libraries
+------------------
+
+Another common attack vector is through libraries upon which girder depends
+such as CherryPy, Jade, PyMongo, etc. During development, and before releases
```
brianhelba: Change to / insert:

...Cherrypy, Jade, PyMongo, etc. Girder's library dependencies reference specific versions, ensuring that arbitrary upstream changes to libraries are not automatically accepted into Girder's environment. Conversely, during development, and before releases...

docs/security.rst
```diff
@@ -60,3 +60,70 @@ Girder returns an error when a Cross-Origin request is made (one with the
...
+
+Notes on Secure Deployment
+--------------------------
+It is recommended that Girder be deployed using HTTPS as the only access
+method. Additionally, we recommend encrypting the volume where the Mongo
+database is stored as well as always connecting to Mongo using authenticated
+access. The volume containing any on-disk assetstores should also be encrypted
+to provide encryption of data at rest. We also recommend using a tool such as
+logrotate to enable the audit of girder logs in the event of a data breach.
+Finally, we recommend a regular backup of the Girder database, configuration,
```
brianhelba: Also reccomend that users test the recovery process periodically. You don't want to first learn how to recover a backup or discover problems with the backups when you actually need to do a recovery.

docs/security.rst
cpatrick: That seems superfluous in terms of security.

docs/security.rst
```diff
@@ -60,3 +60,75 @@ Girder returns an error when a Cross-Origin request is made (one with the
...
+against many injection vulnerabilities as described in the
+`MongoDB Documentation
+<http://docs.mongodb.org/manual/faq/developers/#how-does-mongodb-address-sql-or-query-injection>`_.
+Girder also uses a model layer to mediate and validate all interaction with
+the database. This ensures that for all database operations and structural
+attributes (collection name, operation type) are hardcoded and not modifiable
+by the client, while data attributes (stored content) are validated for proper
+form before being accepted from a client.
+
+Additionally, we strongly recommend configuring your MongoDB server with
```
brianhelba: Change to: "...strongly recommend configuring your MongoDB server with JavaScript..." to make clear that this is not an option within Girder, but rather must be done by the server admin.

docs/security.rst
brianhelba: I disagree. An upstream change doesn't have to malicious to break things, since dependant library code is always designed to be used in conjunction with your own. Blindly accepting updates for your OS may be ok, because it's designed to work as a self-contained system. Blindly accepting new versions of libraries is a great way to break your own software due to API changes and such. Security bugs can be part of such breakages.

cpatrick: I already changed my mind and fixed it :smile_cat:.

brianhelba:
It looked like some of my inline comments are being eaten up as part of outdated diffs. By the time I finish writing them, you've already pushed another version.

Still, I think you got everything.

cpatrick:
@brianhelba Thanks for the review. Great points.

docs/security.rst
```diff
@@ -60,3 +60,77 @@ Girder returns an error when a Cross-Origin request is made (one with the
...
+
+
+Dependent Libraries
+------------------
+
+Another common attack vector is through libraries upon which girder depends
+such as Cherrypy, Jade, PyMongo, etc. Girder's library dependencies reference
+specific versions, ensuring that arbitrary upstream changes to libraries are
+not automatically accepted into Girder's environment. Conversely, during 
+development, and before releases we work to ensure our dependencies are up to
```
brianhelba: No comma between "deployment and before".

docs/security.rst
```diff
@@ -60,3 +60,77 @@ Girder returns an error when a Cross-Origin request is made (one with the
...
+Database Injection Attacks
+--------------------------
+
+Girder defends against database injection attacks by using PyMongo as the only
+pathway between the application server and the database server. This protect
+against many injection vulnerabilities as described in the
+`MongoDB Documentation
+<http://docs.mongodb.org/manual/faq/developers/#how-does-mongodb-address-sql-or-query-injection>`_.
+Girder also uses a model layer to mediate and validate all interaction with
+the database. This ensures that for all database operations and structural
```
brianhelba: No and in "database operations, structural"

docs/security.rst
```diff
@@ -60,3 +60,77 @@ Girder returns an error when a Cross-Origin request is made (one with the
 ``Origin`` header) that does not match the system configuration settings.
 Although most modern web browsers also enforce this, some additional security
 is added by enforcing it at the request level.
+
+
+Database Injection Attacks
+--------------------------
+
+Girder defends against database injection attacks by using PyMongo as the only
+pathway between the application server and the database server. This protect
```
brianhelba: "This protects" should be plural.

brianhelba:
alt text

Otherwise, LGTM

PR #667, opened 2015-01-30
Bug fixes and improvements to static file serving

brianhelba:
* Allow static files to be served within Girder "Resource" classes
* Allow the content type of a static file to be explicity provided

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -268,6 +268,9 @@ def endpoint(fun):
                 # generator function for a streaming response.
                 cherrypy.response.stream = True
                 return val()
+            if isinstance(val, cherrypy.lib.file_generator):
+                # Don't do any post-processing of static files
+                return val
 
         except RestException as e:
             # Handle all user-error exceptions from the rest layer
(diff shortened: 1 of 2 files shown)


girder/api/rest.py
```diff
@@ -18,6 +18,7 @@
 ###############################################################################
 
 import cherrypy
+import cherrypy.lib
```
zachmullen: Since we already have `cherrypy` top level module imported, do we need to explicitly import `cherrypy.lib`?

brianhelba: Yes, we do need the explicit import. By design, CherryPy's root __init__.py file [1] does not do an "import lib" anywhere, so it's not available in the top "cherrypy" namespace.

Also, note that CherryPy's careful use of modules and imports presents a very clean public namespace. This is something Girder should eventually start to consider.

[1] https://github.com/cherrypy/cherrypy/blob/3.3.0/cherrypy/__init__.py

brianhelba: Oops, I'm totally wrong. I was using it with a different import style during my development, and that wasn't working. The extra explicit import of "cherrypy.lib" is not necessary.

I'll fix this.

brianhelba:
Pushed an updated branch in response to @zachmullen feedback.

zachmullen:
LGTM

PR #669, opened 2015-01-30
Specify the version of the McCabe pip module.

manthey:
Specify the version of the complexity module used in flake8.  Otherwise, different build platforms may fail complexity tests because they are using a different version of the test.


zachmullen:
LGTM

PR #672, opened 2015-02-03
UI tweaks for progress list widget

zachmullen:
Just noticed we had wonky looking progress UI for jobs with long titles.


cpatrick:
This LGTM.

Issue #671, opened 2015-02-03
parseQueryString doesn't work as expected


manthey:
I was expecting parseQueryString() to be the functional inverse of `jQuery's $.param()` function.  It is not, as it uses `decodeURI` on the *entire* string before splitting on `&` and subsplitting on `=`.  It should split on `&` before decoding, the split on `=` should use everything after the first `=` as the value, and the decode should be on the individual key and value using `decodeURIComponent`, not `decodeURI`.

Am I wrong about the intent of this function?  If so, it should be commented.  If not, it should be fixed.

cpatrick:
@manthey I'm all for a fix as long as it doesn't break dialogs or routing. :smiley:

manthey:
I'll submit a fix later today or tomorrow.  I have written a replacement function for another project I'm working on since the girder function didn't work.  I just have to consider how it should be tested.

manthey:
Fixed in PR #676.

PR #676, opened 2015-02-05
Fix parseQueryString

manthey:
parseQueryString didn't work as expected.  It should properly handle query parameters that include escaped characters, and it did not.  Specifically, it was decoding the string with decodeURI instead of decoding the query keys and values with decodeURIComponent.  Also, if a value contained an equals sign, it would not work.

This fixes issue #671.


cpatrick:
This LGTM.

PR #679, opened 2015-02-06
Support streaming zip downloads for files in S3 assetstores

zachmullen:


zachmullen:
This works, I just need to write some tests.

zachmullen:
Ready for review.

mgrauer:
LGTM.  Tested live in the field by Covalic.

PR #682, opened 2015-02-07
Fixes #562. Workable controls for disabling core router

### LINKED ISSUE #562: Make it easier to disable girder's router
We have the `girder.handleRouting` property, but unless downstream code manages to set the property after girder has been defined but before router.js is executed (which would be difficult since they're minified into the same file), setting the property will have no effect. What we need is a control that can be called any time at runtime that will disable the routing, like:

    girder.router.disable();

@jeffbaumes Are you using `girder.handleRouting`? How do you have it working?

zachmullen:


zachmullen:
As mentioned in the bug, there was no straightforward way to disable the girder router in downstream js apps, as users would need to set the value of ``handleRouting`` in between ``init.js`` and ``router.js``, which was difficult since they were minified into the same file. Now the value can be set anytime at runtime via ``girder.router.enabled(false)``.

The static analysis tests are all failing because there was a pep8 package release last night that contained [a bug, which has been fixed but not yet pushed to pypi](https://github.com/jcrocholl/pep8/issues/377).

cpatrick:
This LGTM!

PR #683, opened 2015-02-09
Acl widget issues

zachmullen:


zachmullen:
Depends on #682 (only the last commit here is this topic)

cpatrick:
Does the ACL Widget have testing coverage? This LGTM if we add an issue for that (or better yet, a test).

zachmullen:
[It has decent coverage currently](http://my.cdash.org/viewCoverageFile.php?buildid=724278&fileid=1274687).

cpatrick:
In that case, this LGTM--without qualification.

Issue #680, opened 2015-02-07
Unexpected behavior in size quota dialog


zachmullen:
**Steps to reproduce**:

1. Set default user quota to something large (I used 5 GB).
2. Set a specific user quota to something else (I used 50 B), save settings.
3. Refresh browser, open quota dialog, and choose to use the default quota and save settings.
4. Refresh browser, open quota dialog.

**Symptom**: correct radio button is selected but the status shows as if we were still using the user-specific value, as shown below:

![screen shot 2015-02-06 at 8 26 34 pm](https://cloud.githubusercontent.com/assets/555959/6090186/f5f05442-ae3e-11e4-8d3a-99106af228bc.png)

ping @manthey I looked through the code but it wasn't immediately clear to me why this was happening, figured you'd be able to track it down faster.

manthey:
Fixed in PR #684.

PR #684, opened 2015-02-09
Quota display is wrong with default quota.

manthey:
When the default quota is used (rather than a specific Group or User quota), the reported capacity was still based on the specific quota.  This affected both the text and the graph.

This fixes issue #680.


zachmullen:
LGTM!

PR #685, opened 2015-02-09
Python client packaging

zachmullen:
https://pypi.python.org/pypi/girder-client/


jeffbaumes:
Will this be installed with a separate pip install or as part of the main girder package? I just ask since it may be more difficult to maintain two pip packages with version (in)consistencies between the two (like client only working with certain server versions).

zachmullen:
At the moment it's a separate install, since we want people to be able to install the client without having to install the whole server on their machine. Not sure of the best way to handle the versioning other than perhaps trying to keep the client major version in sync with the server's. Open to suggestions on that.

clients/python/setup.py
```diff
@@ -0,0 +1,64 @@
...
+        'Development Status :: 3 - Alpha',
+        'Environment :: Console',
+        'License :: OSI Approved :: Apache Software License',
+        'Operating System :: OS Independent',
+        'Programming Language :: Python :: 2'
+    ],
+    packages=find_packages(exclude=('tests.*', 'tests')),
+    install_requires=reqs,
+    zip_safe=False,
+    scripts=['girder-cli']
```
jbeezley: You might consider using [`entry_points`](https://pythonhosted.org/setuptools/setuptools.html#automatic-script-creation) so that it works on Windows.  It is probably more important here than with the `girder-install` script because the client could realistically be supported on Windows.

Also see @ronichoudhury [internal blog post](http://kwdocs/blog/kwblog/home/post/1622).

zachmullen: Good thought, I will do that.

jeffbaumes:
Yeah I definitely see that reasoning @zachmullen. The dependencies for girder_client are way less than the girder server. Perhaps the pip versioning track the main girder package and should we make sure we release each as part of the official release procedure?

brianhelba:
+1 to @zachmullen 's reasoning. Having the client not depend on the server is a definite requirement for my own anticipated use of the Girder Python client.

cpatrick:
:+1: to adding this to the official release procedures.

zachmullen:
Roger that.

zachmullen:
The packaging process is in place and we have 0.1.0 on pypi. We still need to add a testing framework for the python client before we move to 1.0. This topic is pretty large though so maybe we should do that on another branch.

jeffbaumes:
The release instructions LGTM - but I have not tested the client code.

zachmullen:
Merging this, I'll make adding tests for this a priority in the next few days.

PR #686, opened 2015-02-09
Downloading zip files fail with >= 2Gb of data.

manthey:
When downloading equal to or more than 2Gb of data, zip files would fail to write a footer because they were trying to fit an unsigned 32 bit value into a signed field.

This is an awkward issue to write a test for, as it requires a massive download.  I may add an artificial test if I can concoct a quick one.

### DIFF
--- a/girder/utility/ziputil.py
+++ b/girder/utility/ziputil.py
@@ -251,7 +251,7 @@ class ZipGenerator(object):
             size = min(size, 0xFFFFFFFF)
             offsetVal = min(offsetVal, 0xFFFFFFFF)
 
-        endrec = struct.pack(b'<4s4H2lH', b'PK\005\006', 0, 0, count, count,
+        endrec = struct.pack(b'<4s4H2LH', b'PK\005\006', 0, 0, count, count,
                              size, offsetVal, 0)
         data.append(self._advanceOffset(endrec))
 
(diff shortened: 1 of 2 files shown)


zachmullen:
Hm, it's gone from a long to an unsigned long, which leads me to believe it will only support twice as much range. I notice right above this in the condition of large zip we are calling `size = min(size, 0xFFFFFFFF)`, truncating to 4GB. Does this work for stuff over 4GB?

zachmullen:
Nevermind, had my symbols backward, it's only doing the truncation for *small* zips. Still, intuitively it seems like `l` vs. `L` would only have one extra bit of storage, so we should test the >4GB case.

manthey:
When you have a large zip file (one with >= 2Gb of uncompressed data or more than 65535 files), it writes TWO footers.  One with 64-bit fields with the actual values, and one with 32-bit fields with 'max' values.  The problem was that those 32-bit fields were marked as signed but the 'max' values were unsigned and wouldn't fit.

Changing from `l` to `L` eliminates throwing an error.

I'll have a test of this in a few minutes that can run quickly.

Chris is testing his real-world case right now, too.

zachmullen:
You're right, I was confused but see what it's doing now. Thanks for clarifying.

zachmullen:
LGTM

PR #688, opened 2015-02-09
Make some model methods into generators, instead of returning lists

brianhelba:
Also ensure that API endpoints do return lists, instead of generators, and do so using the simpler list constructor syntax when possible.


brianhelba:
Marking this as WIP for now. Some of the test code needs to be updated to not call len() on generators.

brianhelba:
PR is ready for review.

brianhelba:
@zachmullen @manthey No pressure, but this is ready for review.

zachmullen:
LGTM!

PR #691, opened 2015-02-10
Web client UI for uploading files into an item

zachmullen:


zachmullen:
Ready for review.

manthey:
I was sort of expecting the interface to upload another file to be the a upload button (like in the g-hierarchy-actions-header) and located in the g-item-files-header, rather than an entry in the item Actions dialog.  All of the other upload points are independent icons.

zachmullen:
Good point, I will move it to its own button.

zachmullen:
ping @manthey

manthey:
LGTM.

PR #692, opened 2015-02-11
Add a girder-server executable for starting the girder server

zachmullen:


zachmullen:
ping @jbeezley 

Thanks @ronichoudhury for the blog post about entry_points :)

jbeezley:
Nice.

As we are adding executables, I'm wondering if we should combine them somehow into a real cli interface.  At first, I was thinking at first about just `girder` with subcommands, i.e. `girder server [args]`, `girder install [args]`, etc.  But we do have to differentiate between client and server so maybe `girder-client` and `girder-server` is the way to go.  Anyway, just a thought.

zachmullen:
I don't know if there's One True Right Answer to how many executables to expose. Plenty of tools that involve daemons end up installing a bunch of different executables (e.g. mongodb, mysql server, rabbitmq), so I don't think we should feel like we have to combine them all. Having one for installing optional components and one for running the daemon seems like a reasonable separation to me.

jbeezley:
Yes, certainly it isn't an eminent concern.  In any case, I can (should) add an entry point for girder-install in a future PR.

zachmullen:
Does this look good to merge?

jbeezley:
+1 from me.

cpatrick:
LGTM

-Patrick

On Thu, Feb 12, 2015 at 10:52 AM, Jonathan Beezley <notifications@github.com

PR #694, opened 2015-02-11
Email admins

zachmullen:


girder/utility/mail_utils.py
```diff
@@ -53,25 +53,40 @@ def renderTemplate(name, params=None):
...
+
+    if not isinstance(to, (list, tuple)):
+        raise Exception('You must specify a "to" address or list of addresses '
+                        'or set toAdmins=True when calling sendEmail.')
+
     msg = MIMEText(text, 'html')
-    msg['Subject'] = subject
-    msg['To'] = to
-    msg['From'] = ModelImporter().model('setting').get(
+    msg['Subject'] = subject or '[no subject]'
```
brianhelba: If a user wants to send an empty string as the Subject, as opposed to just omitting the subject keyword argument, this makes it impossible. RFC2822 3.6 does actually allow empty subjects.

I don't really care one way or another, I just wanted to point out the logic.

brianhelba:
LGTM.

PR #696, opened 2015-02-12
Move computation of email link URL to its own function

zachmullen:
I needed this for COVALIC since it needs to compute an alternative host value for its emails based on the core one.


manthey:
LGTM.

PR #697, opened 2015-02-14
Add testing infrastructure for python client

zachmullen:


zachmullen:
This is now ready for review. I've added the infrastructure necessary to make testing of the client library and CLI straightforward. Additionally I got the coverage to 88% for the library and 83% for the CLI-related code -- I leave it to other devs to fill in the remaining gaps.

cpatrick:
It looks like the callback infrastructure is no longer present. I'm assuming you and @mgrauer discussed that. Could I ask why?

zachmullen:
That was removed for the moment as part of the refactor. The CLI module had no actual command-line hook into adding callbacks, which is reasonable; it doesn't really make sense to have python callback functions for something that is a command-line interface. I would have moved them down to the library base class (GirderClient) except that the functions that made use of them only exist in the CLI subclass. TLDR: we need further refactoring to move the upload and download functions into the base library, where it will then be possible to test callbacks. I'd rather do that refactoring on a separate topic though.

cpatrick:
As long as @mgrauer  is happy, this LGTM! Yay testing. :dango:

PR #698, opened 2015-02-16
Improve documentation and implementation of "filterResultsByPermission"

brianhelba:
This fixes a bug where "Item.filterResultsByPermission" did not apply its "removeKeys" argument. This also adds parameter documentation to "Item.filterResultsByPermission", which was missing.

This improves the implementation of both "filterResultsByPermission" methods to use "itertools.ifilter", which is faster and simpler than manual iteration and testing.

### DIFF
--- a/girder/models/item.py
+++ b/girder/models/item.py
@@ -296,7 +296,8 @@ class Item(Model):
         filteredCursor = itertools.ifilter(hasAccess, cursor)
         for result in itertools.islice(filteredCursor, offset, endIndex):
             for key in removeKeys:
-                del result[key]
+                if key in result:
+                    del result[key]
             yield result
 
     def createItem(self, name, creator, folder, description=''):
--- a/girder/models/item.py
+++ b/girder/models/item.py
@@ -19,6 +19,7 @@
 
 import copy
 import datetime
+import itertools
 import json
 import os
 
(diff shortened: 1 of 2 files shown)


zachmullen:
@brianhelba I just ran into an issue related to `filterResultsByPermission` that wasn't uncovered until your recent change to setting the default limit to 0; the function does not handle `limit=0` or `limit=None` as we would like it to (which would be to have no limit). Your new implementation also has the same behavior, so it seems like this PR would be the best place to fix this.

zachmullen:
Looking deeper into the differences, the current master implementation actually does handle the limit=0 case in a subtle (accidental?) way. Because it compared with exact equality rather than an inequality, if the limit was zero we would return unlimited results.

    for result in cursor:
            if self.hasAccess(result, user=user, level=level):
                if skipped < offset:
                    skipped += 1
                else:
                    for key in removeKeys:
                        del result[key]
                    yield result
                    count += 1
            if count == limit:
                break

As long as the user had access to the first result, the count would be 1 and the results would be unlimited. If access was denied on the first resource in the set, no results would be returned. This error was presumably not covered by our tests. Needless to say, that's a bug so I'm glad you've reimplemented this. The fix for your implementation should be as simple as changing the following line:

    if index >= offset + limit:

to

    if limit and index >= offset + limit:

zachmullen:
Looks like you've sorted it out :)

brianhelba:
@zachmullen Sorry, I didn't have time to work on this until now. Here's a new version (rebased on master).

zachmullen:
Very functional and elegant, glad we will get another important bugfix into this point release.

The `server_mongo_search.search` test is failing on both builds and also locally for me. I will push a fix for that test to this topic.

zachmullen:
If that last commit looks good to you, I will merge this.

brianhelba:
LGTM.

PR #699, opened 2015-02-16
Remove arbitrary "limit=50" defaults from internal model API

brianhelba:
Internal model methods should return iterators, so there is no reason to place an arbitrary limit of 50 on such methods. Changing all relevant model methods to use "limit=0" provides for the full result set by default, while retaining the ability to limit results if explicitly desired.

The HTTP API remains unchanged. All relevant HTTP API methods use "Resource.getPagingParameters" to extract the "limit" parameter, which returns "50" by default. This also has the benefit of placing the magic value of "50" into a single place in the code.

### DIFF
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -233,7 +233,7 @@ class System(Resource):
 
     @access.admin
     def discardPartialUploads(self, params):
-        uploadList = list(self.model('upload').list(filters=params, limit=0))
+        uploadList = list(self.model('upload').list(filters=params))
         # Move the results to list that isn't a cursor so we don't have to have
         # the cursor sitting around while we work on the data.
         for upload in uploadList:
--- a/girder/models/assetstore.py
+++ b/girder/models/assetstore.py
@@ -102,7 +102,7 @@ class Assetstore(Model):
                 first['current'] = True
                 self.save(first)
 
-    def list(self, limit=50, offset=0, sort=None):
+    def list(self, limit=0, offset=0, sort=None):
         """
         List all assetstores.
 
--- a/girder/models/collection.py
+++ b/girder/models/collection.py
@@ -82,7 +82,7 @@ class Collection(AccessControlledModel):
         folders = self.model('folder').find({
             'parentId': collection['_id'],
             'parentCollection': 'collection'
-        }, limit=0)
+        })
         for folder in folders:
             self.model('folder').remove(folder, progress=progress, **kwargs)
 
@@ -92,11 +92,11 @@ class Collection(AccessControlledModel):
             progress.update(increment=1, message='Deleted collection ' +
                             collection['name'])
 
-    def list(self, user=None, limit=50, offset=0, sort=None):
+    def list(self, user=None, limit=0, offset=0, sort=None):
         """
         Search for collections with full text search.
         """
-        cursor = self.find({}, limit=0, sort=sort)
+        cursor = self.find({}, sort=sort)
         return self.filterResultsByPermission(
             cursor=cursor, user=user, level=AccessType.READ, limit=limit,
             offset=offset)
@@ -185,7 +185,7 @@ class Collection(AccessControlledModel):
         folders = self.model('folder').find({
             'parentId': doc['_id'],
             'parentCollection': 'collection'
-        }, limit=0, timeout=False)
+        }, timeout=False)
         for folder in folders:
             for (filepath, file) in self.model('folder').fileList(
                     folder, user, path, includeMetadata, subpath=True):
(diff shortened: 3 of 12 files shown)


brianhelba:
Ready for review / merge.

manthey:
Mongo has an arbitrary timeout on cursors of 10 minutes, unless you ask for no timeout (this is separate from the connection timeout and query timeout).  In some places, we turn off the timeout on the cursor because the query could take a while.  Do we need to do that in any new locations because of this change?  Or, is it sufficient to let the calling client add the timeout parameter when necessary?

brianhelba:
@manthey Interesting consideration. Any synchronous operation from an API call obviously shouldn't take longer than 10 minutes to complete, but a batch job could. However, I would think that this sort of use case would almost certainly already have been passing "limit=0". I see my change as just a usability improvement to the model API.

That said, it looks like "timeout=False" has been added to many of the maintaince calls, but would there be any harm from just adding it everywhere? That is, add it directly in "ModelBase.find" so it gets added to all calls? Any cursor that's been exhausted or goes out of scope will still be automatically closed.

zachmullen:
From what I can tell, this looks like simply changing a default from limited to unlimited. Paging result sets by default is a design decision for Girder that I still think is a good idea; I want developers to always have to consider the amount of data they might be getting back, so I think having a non-zero default should be preserved. But you're right that we shouldn't have "50" written everywhere -- replacing it with a constant would be my preferred solution. ping @girder/developers @girder/owners to weigh in on this.

brianhelba:
@zachmullen I agree that default pagination of HTTP API result sets is a great idea. Accidentally sending a JSON document with thousands of entries is an easy source for problems. This doesn't change the HTTP API.

However, the internal model API now uses lazy iterators in every place that "limit=50" was removed (which wasn't the case when the model API was initally written). There is no difference in memory usage between "limit=0" and "limit=50". If a plugin developer writing Python code doesn't understand or consider basic complexity analysis ("loops can be expensive and cost O(n)"), I don't think there's much we can do to save them from themselves (Python is a language for consenting adults).

I cannot honestly think of a use case where a plugin developer would say "I want to internally interate through up to just 50 folders that match my criteria, then stop and do nothing with the rest". Even if such a case does arise, it's likely that the developer will want a value other than 50, and explicitly pass the limit argument. What is quite likely for a new developer (and happened to me at first), is that someone accustomed to PyMongo makes a simple call to "model.find" and then is confused because the iteration magically stops after 50 results.

I don't see a performance issue either way, and it greatly improves usability. I'm certainly open to other opinions @girder/developers @girder/owners , of course.

zachmullen:
If your assertion that there is no memory difference in the pymongo cursor layer is true, I would be OK with this. I have never looked at the implementation details of the cursor -- I thought there was a chance it might buffer results on the client (pymongo) side. If it doesn't do that, though, I suppose this is fine.

brianhelba:
@zachmullen The "batch_size" method can be called on a Cursor to change the default amount of client-side buffering (see http://docs.mongodb.org/manual/reference/method/cursor.batchSize/#cursor.batchSize and http://api.mongodb.org/python/current/api/pymongo/cursor.html#pymongo.cursor.Cursor.batch_size ). As with @manthey 's point about "timeout", this can be set once in "ModelBase.find". Note that if a cursor is already called with the "limit" parameter, "batch_size" only has an effect if it's smaller than "limit".

If we want to do low-level optimization of PyMongo's client side buffering, that's fine, but it's a completely seperate issue from pagination. Pagination implies that the caller might later go back for additional pages with a different "offset". Internal calls to the model API wouldn't do this: they'd iterate through some or all of the cursor in one loop. 

If we do add an explicit "batch_size", we should consider the typical usage of either "limit=50" (in which case the "batch_size" would have to less than 50 to have an effect) or "limit=0" (which is quite common, based on all the explicit usages this patch removed).

zachmullen:
Thanks for the links. It seems pymongo sets the default batch_size to 0 when a cursor is instantiated, and it's unclear from the mongo-level docs exactly what behavior that causes but it appears from a quick googling that the first batch will be 101 documents in that case, which is a fine default value for me.

So this LGTM once you merge origin/master or rebase.

brianhelba:
@zachmullen I'll rebase and drop the 2nd commit which fixes style issues.

@manthey I'll create a 2nd PR to do the global "timeout=False" change. If it turns out to be a bad idea, or requires setting on a more granular basis, we can discuss it there.

brianhelba:
This is now rebased, and all tests pass. Merge at your convenience, @zachmullen .

I have another PR to address @manthey 's point about timeout.

PR #700, opened 2015-02-16
Pin version of pep8 module so that analysis test won't break unexpectedly.

manthey:


zachmullen:
Did they break it again, or is this just a defensive move?

manthey:
The pep8 version changed, and so some of the commits made since Friday are failing style tests.  This pins the version to the current most recent and fixes the two files that were failing.  And, by pinning the version, this won't happen again.

zachmullen:
LGTM

Issue #91, opened 2014-02-18
indicate pending group management requests in UI


mgrauer:
From the review of #78, when a group management has tasks, have some visual indication in the pending "inbox" tray.  Perhaps a number of open tasks, or some paper to appear in the tray or ...

manthey:
Added in PR #701.

Issue #673, opened 2015-02-04
Group Moderators / Administrators are also listed as Group Members


brianhelba:
A user who is a Moderator or Administrator of a Group also shows up as a Member, which is confusing. The user should be listed only under the Moderator / Administraor section, and have the necessary promote / demote / remove icons as part of that single listing.

manthey:
Should group moderators be able to ADD new members to a group, or only INVITE new members?

Also, why can moderators edit group properties?  It seems like that should be an admin power (to change, for instance, if a group is public or private).

zachmullen:
Just invite. I think only site admins should be able to force a user into a group. I could be persuaded otherwise though if we have a good story.

I agree that only admin access should confer the ability to change properties like the public flag.

manthey:
Also, a moderator can currently invite a new member to be a moderator, but doesn't have the power to raise (or demote) a member to moderator status.  It seems like either a moderator shouldn't be able to invite a user to be a moderator or should be able to promote (but then, would they be allowed to demote?).

zachmullen:
We can change that policy to be more consistent, maybe just by restricting moderators to only inviting users as members.

brianhelba:
@zachmullen @manthey This issue is a straightforward bug that should be fixed. However, group moderators being able to ADD members is something I have thoughts on as well. I'm going to create a seperate issue to discuss that.

brianhelba:
See https://github.com/girder/girder/issues/690 for the ADD / INVITE issue.

brianhelba:
@manthey @zachmullen While this issue is being fixed, it would be nice if some tooltips explaining what group "moderator" and "administrator" meant, outside of just the user invite dialog.

zachmullen:
:+1:

manthey:
Fixed in PR #701.

Issue #92, opened 2014-02-18
rework ui for group management promotion


mgrauer:
The layout in group management is a bit confusing, promotions are accomplished with an up arrow, so members are moved up to moderators and moderators up to admins, but moderators are displayed below members and admins below moderators, so there is a bit of a disconnect here.

This came out of a review of #78.

manthey:
Fixed in PR #701.

PR #701, opened 2015-02-16
Group interface changes

manthey:
The members of a group are now only listed in one place: admin, moderator, or member.  The admin and moderator lines have promotion, demotion, and removal controls.

The Pending tab lists how many actions are pending.

Moderators can no longer edit a group.

Updated and added tests for all changes.  Groups are still under-tested, but they have greater coverage now than before.  Also, I'm working on gradually getting some of the test code to match our main code's style guidelines, even though this isn't currently enforced.

This fixes issues #91, #92, and #673.

### DIFF
--- a/clients/web/src/stylesheets/body/groupPage.styl
+++ b/clients/web/src/stylesheets/body/groupPage.styl
@@ -34,7 +34,7 @@
   > a
     font-weight bold
 
-.g-group-members-container,.g-group-requests-container
+.g-group-members-container,.g-group-requests-container,.g-group-admins-container,.g-group-mods-container
   margin-top 15px
 
 .g-group-members,.g-group-mods,.g-group-admins,.g-group-requests,.g-group-invites
@@ -103,3 +103,6 @@
 #g-invite-role-container
   .g-add-user-button
     margin-left 10px
+
+#g-group-tab-pending-status
+  display inline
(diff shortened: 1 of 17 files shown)


brianhelba:
@manthey I've tested it locally, and it works fine. Good improvement, thank you.

I do question whether the whole "g-group-invite-form" ("Invite a user to join..." box) should logically be positioned outside of the "g-group-members-header", since it can be used to directly invite / add people as Moderators or Administrators, not just members.

Finally, note that, the "Prev", "Page 1", "Next" pagination buttons always appear under the Members group, even if there are no members. The Moderators and Administrators have no such pagination buttons. This will probably be fine for most cases, where there are many members and few Moderators / Administrators, but more robust logic for displaying the pagination buttons would probably be good. However, I think that's outside the scope of this PR.

I don't feel strongly about either issue, so LGTM if you'd like to merge as-is.

manthey:
I think that either of those changes should be a separate issue.

One minor other issue I have with pagination is that we aren't consistent in how we handle pagination here versus in a folder with lots of items.  In a folder, when you have a lot of items, there is a link to "show more items", which extends the list.  In the list of members in a group, the data is paged, and you only see one page of users.  Both have virtues (and, in a paged view, being able to jump to page X can be a good thing).

PR #703, opened 2015-02-16
Make all PyMongo "find" queries not timeout

brianhelba:
A PyMongo cursor will always be deallocated by the server when it is exhausted or when it is closed by the client (typically by going out of scope). This change will ensure that cursors held open longer than 10 minutes will not be automatically closed by the server prematurely, which could affect long-running batch jobs.

See: http://docs.mongodb.org/manual/core/cursors/#closure-of-inactive-cursors

Developers are still advised to manually close cursors that will not quickly go out of scope, to ensure that both client and server-side resources are freed as soon as possible.

Thanks to @manthey for bringing up this topic.

### DIFF
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -307,7 +307,7 @@ class User(Resource):
         self.requireParams('email', params)
         email = params['email'].lower().strip()
 
-        users = self.model('user').find({'email': email}, timeout=False)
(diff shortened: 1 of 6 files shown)


brianhelba:
@manthey What do you think?

brianhelba:
All tests passing, ready for review.

manthey:
LGTM.

PR #704, opened 2015-02-16
Fixes #702. Package mail_templates directory for pip

### LINKED ISSUE #702: Install mail_templates dir

zachmullen:
This should fix the issue @ronichoudhury ran into.

### DIFF
--- a/tests/packaging/test_package_contents.py
+++ b/tests/packaging/test_package_contents.py
@@ -1,0 +1,20 @@
+import glob
+import sys
+import tarfile
+
+
+package = glob.glob('girder-[0-9].*.tar.gz')
+if not len(package):
+    print('Could not find package archive')
+    sys.exit(1)
+if len(package) > 1:
+    print('Error: multiple package versions found')
+    sys.exit(1)
+
+basename = package[0][:-7]  # remove '.tar.gz'
+
+with tarfile.open(package[0], 'r:gz') as tar:
+    headerMako = tar.getmember(basename + '/girder/mail_templates/_header.mako')
+    if headerMako.size <= 0:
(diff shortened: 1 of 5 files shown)


jbeezley:
I don't think that is sufficient to make pip actually install the templates.  You will probably have to add an entry to `package_data` in `setup.py`.  A better test would be to see if the template exists in the virtualenv after [installing](https://github.com/girder/girder/blob/master/tests/packaging/pip_install_girder.sh).

zachmullen:
You were right, I added a check for the installed package as well.

zachmullen:
ping @jbeezley

tests/packaging/test_package_contents.py
```diff
@@ -0,0 +1,20 @@
...
+    print('Error: multiple package versions found')
+    sys.exit(1)
+
+basename = package[0][:-7]  # remove '.tar.gz'
+
+with tarfile.open(package[0], 'r:gz') as tar:
+    headerMako = tar.getmember(basename + '/girder/mail_templates/_header.mako')
+    if headerMako.size <= 0:
+        print('Mail templates were not packaged properly')
+        sys.exit(1)
```
jbeezley: Is this test necessary any more?

zachmullen: While it is superseded by the new test in terms of functionality, it adds an additional layer that could make it easier to detect where an error is occurring (sdist step vs. install step) so I think we may as well keep it around since it's already written and runs in a tenth of a second.

jbeezley: Sounds reasonable.

jbeezley:
LGTM

Issue #690, opened 2015-02-10
Group Moderators / Administrators should be able to add (not just invite) users


brianhelba:
Our use of Girder would benefit from allowing group moderators / administrators to directly add users to groups. If Girder can't / won't support this, I'll eventually want to add the functionality as part of our plugin. 
 
Our use case is that we have several "phases" in a data annotation pipeline. Each phase is administered and used by a separate group of people (some are clinicians, some are CV experts, etc.). There is also a clear hierarchy between phase administrators, who are trusted and appointed by us, and phase users, who have no site powers beyond group membership + read access. The phase administrators should be able to independently decide who needs to be a member of their group, without having to burden the ordinary users with all of the "invite+accept" stuff, which many of them have a hard time figuring out, due to their non-technical background.

brianhelba:
I also totally see the use case where the group invite functionality is useful. In a more flat community, you wouldn't want a rogue group moderator to start adding random people from all over the site to the group.

However, maybe we could have a site configuration option to toggle this ability?

brianhelba:
@zachmullen @manthey I'd be interested in your feedback here.

zachmullen:
+1 for site config option or a small plugin that exposes this option. Would be nice to be able to set it on a per group basis as a site admin, but we could add that down the road.

manthey:
I'm working on this code right now, so I can add this, too.

This is what I think the desired features and powers are:

Group admins can ADD or INVITE users to to be members, moderators, or admins.  Group admins can edit group properties and delete groups.

Group moderators can INVITE users to be members.  Group mods can NOT edit group properties.  Groups mods can NOT INVITE users to be moderators.

If a (new) group property of 'allow moderators to directly add members' is selected, then group moderators can ADD users to be members.

A (new) system config setting could determine the default of the group property for allowing mods to directly add members.

brianhelba:
@manthey I really like this system. Not only is it a good compromise of the ADD / INVITE dilemma that doesn't involve a special config option, but it further delineates and clarifies the powers and roles of moderators and admins, which were a bit muddled before.

zachmullen:
I still think the default should be that only site-level admins can add a user to a group directly (since technically that's changing a property of the user without their permission), and that the system-level setting can add that power to group admins and/or moderators. ping @girder/developers @girder/owners for consensus.

jcfr:
+1

On Tue, Feb 10, 2015 at 10:58 AM, Zach Mullen <notifications@github.com>
[...]

PR #707, opened 2015-02-18
Allow group admins and mods to directly add users to groups.

manthey:
Added a setting to pick whether group admins and mods are allowed to directly add users to groups.  If enable, individual group's have a property that further modifies this option.

Updated system tests and javascript tests.

This completes issue #690.

### DIFF
--- a/girder/api/v1/group.py
+++ b/girder/api/v1/group.py
@@ -20,7 +20,7 @@
 from ..describe import Description
 from ..rest import Resource, loadmodel
 from girder.models.model_base import AccessException
-from girder.constants import AccessType
+from girder.constants import AccessType, SettingKey
 from girder.utility import mail_utils
 from girder.api import access
 
@@ -122,7 +122,11 @@ class Group(Resource):
     @loadmodel(model='group', level=AccessType.READ)
     def getGroup(self, group, params):
         user = self.getCurrentUser()
-        return self.model('group').filter(group, user)
+        group = self.model('group').filter(group, user)
+        # Add in the current setting for adding to groups
+        group['_addToGroupPolicy'] = self.model('setting').get(
+            SettingKey.ADD_TO_GROUP_POLICY)
+        return group
     getGroup.description = (
         Description('Get a group by ID.')
         .responseClass('Group')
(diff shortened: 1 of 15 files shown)


girder/api/v1/group.py
```diff
@@ -251,7 +263,25 @@ def inviteToGroup(self, group, params):
             id=params['userId'], user=user, level=AccessType.READ, exc=True)
 
         if force:
-            self.requireAdmin(user)
+            if not user.get('admin', False):
+                mustBeAdmin = True
+                addPolicy = self.model('setting').get(
+                    SettingKey.ADD_TO_GROUP_POLICY)
+                addGroup = group.get('addAllowed', 'default')
+                if addGroup not in ['no', 'yesadmin', 'yesmod']:
```
zachmullen: Should we turn these strings into constants? I don't have an opinion one way or the other, just bringing up the dogmatic position.

tests/cases/group_test.py
```diff
@@ -33,10 +33,11 @@ class GroupTestCase(base.TestCase):
     def setUp(self):
         base.TestCase.setUp(self)
 
-        # Create a set of users so we can work with these groups
+        # Create a set of users so we can work with these groups.  User 0 is
+        # an admin
         self.users = [self.model('user').createUser(
             'usr%s' % num, 'passwd', 'tst', 'usr', 'u%s@u.com' % num)
-            for num in [0, 1, 2]]
+            for num in [0, 1, 2, 3, 4, 5]]
```
zachmullen: I think this has reached a point where `range(6)` makes more sense :)

zachmullen:
Manually tested a good deal of the configuration space, this LGTM.

tests/cases/group_test.py
manthey: @zachmullen, did you want me to change this, or should I merge?  (And, I agree, range is looking more appropriate.)

zachmullen: Let's change this to range and then merge it once travis is happy.

manthey: Right-o.

Issue #632, opened 2015-01-19
Automatic parsing of JSON request bodies


brianhelba:
For HTTP requests with bodies (currently POST, PUT, PATCH), the body of any request with "Content-Type: application/x-www-form-urlencoded" (aka a POST form) is automatically parsed and passed to request handlers within the "params" argument. ( https://github.com/cherrypy/cherrypy/blob/3.3.0/cherrypy/_cpreqbody.py#L175 )

We should enable the same behavior for bodies of "Content-Type: application/json", but there are several decisions to be made:
- Do we enable this behavior for all routes in Girder, or only those under "/api/v1"? I'd suggest all routes, as it's consistant with the other request body processors, and it's likely a useful behavior for any other non-API endpoinds that plugins may create.
- Where do we store the parsed JSON data? There are several options:
  1) Put it in the magic "params" argument. This has several issues:
  - While a POST form is always nicely represented directly as a Python dict (with each key-value item equivalent to a name-value field), a JSON body could be an array. Where would we place this? Something like "params['json']" is even more magicical.
  - The "params" argument already stores both HTTP query string arguments and HTTP request bodies. In my opinion, this is a bit sloppy, as a request handler has no way to tell which source a "params" item came from. Additionally, if there are any repeated key names between the query string and the POST form, each corresponing "params" value becomes a list containing both values (something that Girder API handlers currently do not check for at all!). If anything, I think we should change this behavior for parsed POST forms too and provide the data via some other mechanism.
  
  2) Also solve ( https://github.com/girder/girder/issues/442 ) by making a new Resource method (say "getBody()") that returns the parsed JSON / parsed POST form / raw body, depending on the "Content-Type" of the request. However, this too has problems:
  - For request handlers that are not members of a class inheriting Resource, this method becomes inaccessable (unless we make it static, which is inconistant with how "Resource.getCurrentUser" et. al. are handled).
  - A request handler that expected "Resource.getBody()" to return some parsed Python datatype could instead get a raw string if the request wasn't of the proper "Content-Type". This misses a great oppertunity to move all of the JSON validation boilerplate (like https://github.com/girder/girder/blob/v1.2.0/girder/api/v1/item.py#L179 ) to a central place, as handlers still have to do an all-or-nothing check of whether the body was parsed at all and raise an exception if it wasn't.
  
[...]

PR #708, opened 2015-02-19
Refs #632. getBodyJson method

zachmullen:


cpatrick:
Once Travis is happy, LGTM.

PR #711, opened 2015-02-27
Prepend mail templates from plugins to the search path

zachmullen:
This will make it easy for plugins to override core mail templates.
Many custom apps built on top of Girder will probably want to show
something other than the core header and footer, for instance.


manthey:
LGTM.

PR #713, opened 2015-03-01
Item merging.

cpatrick:
Just the backend at this point


girder/api/v1/item.py
```diff
@@ -319,3 +321,48 @@ def copyItem(self, item, params):
...
+        doc = self.getBodyJson()
+        ids = doc['ids']
+        firstItem = self.model('item').load(id=ids[0], user=user)
+        if firstItem is None:
+            raise RestException('One or more of the items to merge was '
+                                'not found.')
+        name = doc.get('name', firstItem['name'])
+        folderId = doc.get('folderId', firstItem['folderId'])
+        folder = self.model('folder').load(id=folderId, user=user,
+                                           level=AccessType.WRITE, exc=True)
```
zachmullen: Why do we have all this special behavior for the first item?

cpatrick: If a name and/or description is not specified, we pull the name and description from the first in the list and use that.

girder/api/v1/item.py
```diff
@@ -319,3 +321,48 @@ def copyItem(self, item, params):
...
+                                                    description=description,
+                                                    progress=None)
+        return newItem
+    mergeItems.description = (
+        Description('Merge multiple items into a single item')
+        .responseClass('Item')
+        .param('body', 'A JSON object containing a list of the item ids to '
+               'merge as well as optionally a name and description for the '
+               'new item. A progress field can aslo be passed as part of the '
+               'object.', paramType='body')
```
zachmullen: I think we should change this to be consistent with our other actions that operate on a set of resources; see the methods like download, delete, move, and copy in https://github.com/girder/girder/blob/master/girder/api/v1/resource.py. That will also make the parameters able to be more swagger-friendly.

cpatrick: Yep. You're totally right. I'll rework this into that form, and add support for folders as well. It should have roughly the same special cases as moving anyway.

cpatrick: @zachmullen This does prompt the question: In what model should a general folder/item merging function live. I'm going place it in item.

zachmullen: Item sounds good.

girder/api/v1/item.py
zachmullen: Since we already have the same check in the model layer that sets the default name if name is `None`, could we just use that instead of having it in both places?

cpatrick:
I'm reworking this under resource and adding folder support. This is going to be much more consistent with move/copy. Thanks for the great review!

zachmullen:
Does that mean a user could perform a merge containing both folders and items? What would that look like?

zachmullen:
I ask because it feels like there may be a lot of overlap with the move functionality once we start talking about folders.

cpatrick:
So, a merge containing only items would result in a single item containing the files and metadata of all of the items selected. A merge containing folders and items would result in a single item containing all of the metadata of the selected folders and items. This will apply recursively on each folder.

I'm working on the test, but I'm going to go ahead and push the new endpoint.

cpatrick:
Alright. I fixed the new test, but certainly more testing is needed. Let me know if this is more along the lines of what you are thinking. I wish I had checked the resource endpoint first. It was very helpful for a lot of the checking that needed to happen. :tanabata_tree:

girder/api/v1/resource.py
```diff
@@ -369,3 +374,55 @@ def copyResources(self, params):
...
+                                            description, ctx)
+
+    mergeResources.description = (
+        Description('Merge a set of items and folders.')
+        .param('resources', 'A JSON-encoded list of types to move.  Each type '
+               'is a list of ids.  Only folders and items may be specified.  '
+               'For example: {"item": [(item id 1), (item id2)], "folder": '
+               '[(folder id 1)]}.')
+        .param('parentType', 'Parent type for the new parent of these'
+               'resources.')
```
zachmullen: Need a space between these words.

girder/api/v1/resource.py
```diff
@@ -369,3 +374,55 @@ def copyResources(self, params):
...
+        .param('parentId', 'Parent ID for the new parent of these resources.')
+        .param('name', 'The name of the new item to be created')
+        .param('description',
+               'The description of the new item to be created.',
+               required=False)
+        .param('progress', 'Whether to record progress on this task. Default '
+               'is false.', required=False, dataType='boolean')
+        .errorResponse('Unsupport or unknown resource type.')
+        .errorResponse('Invalid resources format.')
+        .errorResponse('Resource type not supported.')
```
zachmullen: "Resource type not supported" and "Unsupported or unknown resource type" should probably just be one thing, or is there a difference?

girder/models/item.py
```diff
@@ -540,3 +540,83 @@ def checkConsistency(self, stage, progress=None):
...
+            self._mergeHelperFolder(cur_folder, newItem)
+
+    def merge(self, toBeMerged, creator, folder, name, description='',
+              progress=None):
+        """
+        Merge several items into one containing the union of files and
+        metadata.
+
+        :param ids: the ids of items to merge.
+        :type ids: list
```
zachmullen: Looks like the `ids` parameter is now covered under `toBeMerged`.

girder/models/item.py
```diff
@@ -540,3 +540,83 @@ def checkConsistency(self, stage, progress=None):
                     progress.update(increment=1, message='Checking items (%d '
                                     'left)' % itemsLeft)
             return numItems, itemsCorrected
+
+    def _mergeMetadataHelper(self, oldDoc, newDoc, type='item'):
+        # metadata
+        if 'meta' in oldDoc:
+            if 'meta' in newDoc:
+                newDoc['meta'] = dict(
+                    newDoc['meta'].items() + oldDoc['meta'].items())
```
zachmullen: Really unimportant, but another option for lines 548-9 would just be `newDoc['meta'].update(oldDoc['meta'])`.

girder/models/item.py
```diff
@@ -540,3 +540,83 @@ def checkConsistency(self, stage, progress=None):
...
+            else:
+                newDoc['meta'] = oldDoc['meta']
+        # other fields
+        filteredDoc = self.model(type).filter(oldDoc)
+        for key in oldDoc:
+            if key not in filteredDoc and key not in oldDoc:
+                newDoc[key] = copy.deepcopy(oldDoc[key])
```
zachmullen: Not sure what this loop is intended to do, but it appears that it will always be a no-op:

```
for key in oldDoc:
   if ... and key not in oldDoc:
      <never reached>
```

girder/api/v1/resource.py
```diff
@@ -369,3 +374,55 @@ def copyResources(self, params):
...
+                             message='Calculating requirements...',
+                             total=total) as ctx:
+            for kind in resources:
+                model = self._getResourceModel(kind)
+                toBeMerged = {
+                    'item': [],
+                    'folder': []
+                }
+                for id in resources[kind]:
+                    doc = model.load(id=id, user=user, level=AccessType.WRITE)
```
zachmullen: If we use `exc=True` in the load call, it will take care of throwing the `RestException` for you.

zachmullen: Actually it's a `ValidationException` but it's a 400 either way.

zachmullen:
Any status on this PR? Should we close it for the moment?

cpatrick:
This is just blocked on my free time. We can close it for now.

-Patrick

PR #714, opened 2015-03-02
Fixed header and nav

zachmullen:
@mgrauer was demoing monkey brains to me earlier and had a page where the body scrolled a very long way down, and we both agreed that it would be better for usability if the navbar and header were fixed in the viewport.

Even if we decide not to go through with this change, we should still merge the first commit on this topic, which fixes what I consider buggy behavior in bootstrap with inline forms.


mgrauer:
LGTM, thanks for the fix.

PR #716, opened 2015-03-03
Replace markedown with marked.  Fixes #715.

### LINKED ISSUE #715: change markdown library
The current library has xss issues.

https://github.com/evilstreak/markdown-js/issues/235

mgrauer:
This branch adds support for marked as a markdown parser.  I'm not saying we should go with marked, but you can play with it here.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -231,7 +231,7 @@ module.exports = function (grunt) {
                         'node_modules/jade/runtime.js',
                         'node_modules/underscore/underscore.js',
                         'node_modules/backbone/backbone.js',
-                        'node_modules/markdown/lib/markdown.js',
+                        'node_modules/marked/lib/marked.js',
                         'clients/web/lib/js/d3.js',
                         'clients/web/lib/js/bootstrap.js',
                         'clients/web/lib/js/bootstrap-switch.js',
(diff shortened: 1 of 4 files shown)


zachmullen:
There's something to be said for having a drop-in replacement that only involves changing 4 lines :)

zachmullen:
This LGTM once travis is happy. Did you try it out with tables and image sizing?

zachmullen:
Actually, hold off on merging this, I want to add one small change that will prevent us from breaking backward compatibility on this later if we have to change the library in the future.

zachmullen:
Ok, if that looks good to you feel free to merge it. See https://github.com/girder/covalic/pull/84

mgrauer:
I was able to get a subset of tables working: it seems like you need headers for the tables to render, but the headers can be blank.

On image sizing there are three routes I've been able to make work

  1. resize the image before uploading :sunglasses: 
  2. use HTML with an `img` tag that has width and height properties, but obviously this requires HTML which we will probably want to turn off
  3. target the alt text of the image and style that.  this would only work for a developer, not someone who wants enter text purely from the client side

markdown

    [ccaxial](http://localhost:8080/api/v1/file/54f77eee89264833b2478bf2/download)

style file

    img[alt=ccaxial]
      width 200px
      height 200px

zachmullen:
I think the `[](url =wxh)` syntax is not part of the core markdown syntax, I guess that's why few renderers support it. This LGTM if you are happy with it.

mgrauer:
Great, this seems good enough for now, and we can keep an eye on [markdown-it](https://github.com/markdown-it/markdown-it).

BTW, the `[](url =wxh)` did not work with marked.

PR #718, opened 2015-03-04
Hdfs assetstore

zachmullen:


zachmullen:
@girder/developers @girder/owners so this topic branch has gotten to be quite a beast, and I think it's time to give it some QA. I got most of the way through testing the new plugin, and I'd like to defer the remaining things on the checklist to a future topic before this one gets any larger than it already is. I had to mock pretty much the entire interface to HDFS, so any significant changes to that layer (e.g. new snakebite version) will need to be manually tested in the future.

Testing checklist:
- [x] Creation/validation of HDFS assetstores
- [x] Updating of HDFS assetstore info
- [x] Importing of pre-existing files
- [x] Downloading of imported files
- [x] Deletion of imported files
- [x] Uploading (managed) files
- [x] Multi-chunk uploading and resume
- [x] Downloading managed files
- [x] Deletion of managed files
- [ ] Cancel an upload
- [ ] Various error conditions
- [ ] Javascript UI tests

girder/api/v1/file.py
```diff
@@ -142,7 +142,7 @@ def requestOffset(self, params):
         upload = self.model('upload').load(params['uploadId'], exc=True)
         offset = self.model('upload').requestOffset(upload)
 
-        if type(offset) is int:
+        if type(offset) in (int, long):
```
manthey: For even more flexibility, should we be using `isinstance(offset, (int, long))` rather than `type`?  It doesn't matter here, but it would allow subclasses of int and long.  In general, places we use `type` probably should be `isinstance`.

girder/utility/assetstore_utilities.py
```diff
@@ -24,28 +24,37 @@
...
-def getAssetstoreAdapter(assetstore):
+def getAssetstoreAdapter(assetstore, instance=True):
     """
     This is a factory method that will return the appropriate assetstore adapter
     for the specified assetstore. The returned object will conform to
     the interface of the AbstractAssetstoreAdapter.
+
     :param assetstore: The assetstore document used to instantiate the adapter.
     :type assetstore: dict
+
```
manthey: Add instance to param documentation.

tests/mock_smtp.py
```diff
@@ -70,7 +70,7 @@ def loop(self):
         exit.
         """
         while len(asyncore.socket_map):
-            asyncore.loop(timeout=1, use_poll=True)
+            asyncore.loop(timeout=0, use_poll=True)
```
manthey: Was there a reason for this change?  When I last touched this file it was to prevent it from using 100% cpu because it was in a tight loop.  Does this put it back into that state?

zachmullen: It made the "bad file descriptor" errors appear to go away on my machine, I will revert it back to 1 (I was not aware of the reason for it being at 1).

manthey: There was another way to squelch those errors (since they are just about having threads stopped in a sub-optimal order).  I'll have to check what I was doing several months ago...

manthey: Can you try changing the timeout something like 0.25 or 0.1 and see if that prevents you from getting the "bad file descriptor" error?  I think any timeout is better than a 0 timeout.  Ideally, it would be as large as possible and still allow a graceful exit.

zachmullen:
@manthey I fixed the timeout (set it to 0.5), I also added a more significant change in my last commit that needs to be looked at. @mgrauer and I were discussing this morning about the "rest.resource.<path>.before" callbacks, and how the handlers could accidentally circumvent the route's access level since the handler is executed before the default handler (where the access level is enforced). The change I made simply requires handlers bound to those sorts of events to also declare an access level or fall back to `@access.admin` as the default.

plugins/hdfs_assetstore/plugin_tests/assetstore_test.py
```diff
@@ -0,0 +1,387 @@
...
+        resp = self.request('/file/{}/download'.format(file['_id']),
+                            isJson=False, user=self.admin)
+        self.assertStatusOk(resp)
+        self.assertEqual(resp.collapse_body(), chunk1 + chunk2)
+
+        # Delete the file, should remove the underlying file on disk
+        resp = self.request('/file/' + str(file['_id']), method='DELETE',
+                            user=self.admin)
+        self.assertEqual(None, self.model('file').load(file['_id']))
+        self.assertFalse(os.path.exists(absPath))
```
manthey: Is there a convenient way to run the mock server so that I can use it from the user interface?  For instance, the mock_s3.py file can be run in a stand-alone mode to facilitate human testing.

zachmullen: I'm assuming you mean the mock HDFS client; the way that gets mocked is just by the above line:

    sys.modules['snakebite.client'].Client = MockSnakebiteClient

So theoretically you could put that anywhere and make use of the same mocking. We can move the class and a helper  method for the mocking line into their own file if we feel it's generally useful.

manthey: I do mean the mock HDFS client.  I was hoping to run girder normally, enable the HDFS plugin, and then run a separate process to provide a mock HDFS service.

zachmullen: Ah, we don't actually have a mock HDFS service, we just mock the client to make it look like we are talking to a real one.

plugins/hdfs_assetstore/web_client/templates/hdfs_assetstore_import.jade
```diff
@@ -0,0 +1,24 @@
...
+    input.form-control.input-sm#g-hdfs-import-path(type="text")
+  .form-group
+    label(for="g-hdfs-import-dest-id") Destination ID
+    input.form-control.input-sm#g-hdfs-import-dest-id(type="text")
+  .form-group
+    label(for="g-hdfs-import-dest-type") Destination type
+    select.form-control#g-hdfs-import-dest-type
+      option(value="folder") Folder
+      option(value="user") User
+      option(value="collection") Collection
```
manthey: I recommend swapping the order of `Destination type` and `Destination ID`, as `type` is the more general of the two. I'd also recommend adding some default text to the ID that reads something like 'An existing Girder folder, user, or collection ID to use as the destination.'

manthey:
Setting an already created assetstore to current doesn't work for me.  It looks like the set-as-current button isn't passing back the required fields, so it fails validation.

plugins/hdfs_assetstore/server/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    events.bind('rest.post.assetstore.before', 'hdfs_assetstore',
+                createAssetstore)
+
+    (Assetstore.createAssetstore.description
+        .param('host', 'The namenode host (for HDFS type).', required=False)
+        .param('port', 'The namenode RPC port (for HDFS type).', required=False)
+        .param('path', 'Absolute path under which new files will be stored ('
+               'for HDFS type).', required=False)
+        .param('user', 'The effective user to use when calling HDFS RPCs (for '
+               'HDFS type). Default is current user.', required=False)
```
manthey: Is this the current OS user, or the current girder user?  It looks like it passed my Ubuntu user name to the url, not my girder user name.  If this is the desired behavior, it should be explained (both here and in the appropriate jade template).

zachmullen:
ping @manthey I think I addressed everything you mentioned

plugins/hdfs_assetstore/web_client/templates/hdfs_assetstore_editFields.jade
```diff
@@ -0,0 +1,16 @@
...
+      placeholder="RPC port")
+.form-group
+  label.control-label(for="g-edit-webhdfs-port") WebHDFS port (default is 50700)
+  input.input-sm#g-edit-webhdfs-port.form-control(type="text" placeholder="WebHDFS port")
+.form-group
+  label.control-label(for="g-edit-hdfs-path") Root path
+  input.input-sm#g-edit-hdfs-path.form-control(type="text" placeholder="Root path")
+.form-group
+  label.control-label(for="g-edit-hdfs-user") Effective user (default is current user)
+  input.input-sm#g-edit-hdfs-user.form-control(type="text" placeholder="Username")
```
manthey: I think these fields should be in the same order as the fields in `hdfs_assetstore_create.jade`.  For instance, Root Path is after host here but before it in the create area.

manthey:
Has the code been tested against a real HDFS system?

zachmullen:
Yeah, I have a pseudo-distributed Hadoop installed on my local machine and everything works.

plugins/hdfs_assetstore/web_client/templates/hdfs_assetstore_editFields.jade
zachmullen: Good point, I'll fix it.

zachmullen:
ping @manthey

manthey:
LGTM.  Nice work.

zachmullen:
Thanks for the thorough QA!

PR #720, opened 2015-03-05
Static functions for uploading standalone data

ronichoudhury:
Includes a pair of functions for uploading free-standing data into a new file in
an Item or Folder.


ronichoudhury:
I am not sure how/where to write tests for this.  Can someone point me the right way?

zachmullen:
So, since this still depends on girder.app.js, I think we should turn these into model methods of `FileModel` and make them flexible in two ways:

1. The first argument to `uploadToFoo` should be either a string (the id) or a model object of that type, e.g.

```
uploadToFolder: function (modelOrId, data) {
    if (typeof(modelOrId) === 'string') {
        modelOrId = new girder.models.FolderModel({_id: modelOrId});
    }
    this.upload(modelOrId, data);
}
```

2. The second argument should be allowed to be a `Blob` like we support now, as well as a string of already-in-memory data, e.g.

```
upload: function (model, data) {
    if (typeof(data) === 'string') {
        data = new Blob([data]);
    }
   ...
}
```

Then we should make the [cookbook section](http://girder.readthedocs.org/en/latest/developer-cookbook.html#upload-a-file) about this a lot better by describing these options.

ronichoudhury:
Sounds good.  What do we do about uploading to folders vs. items?

zachmullen:
```
uploadToFolder: function (modelOrId, data) {
    if (typeof(modelOrId) === 'string') {
        modelOrId = new girder.models.FolderModel({_id: modelOrId});
    }
    this.upload(modelOrId, data);
},

uploadToItem: function (modelOrId, data) {
    if (typeof(modelOrId) === 'string') {
        modelOrId = new girder.models.ItemModel({_id: modelOrId});
    }
    this.upload(modelOrId, data);
}
```

Just two functions like you initially suggested. I think that's a pretty grokable API, but if someone can think of a better way I'm all ears.

zachmullen:
Feel free to still use underscore's handy partial function application to reduce duplication from my example.

ronichoudhury:
I do love to slip in functional programming whenever no one's looking :neckbeard:

Thanks for the guidance.  I'll merge these ideas into the models in a bit.

ronichoudhury:
Quick question:  if a ``FileModel`` already has an ``_id`` (i.e., isn't "new" in the Backbone sense), what would happen when you call ``uploadToFolder`` or ``uploadToItem`` on it?

I ask because eventually I will also want a utility function that can do contents replacement on an existing file as well.

ronichoudhury:
In answer to my last question, I would create a ``FileModel`` with a given id, then call ``updateContents()`` on it, but as part of this effort we'll want to generalize the ``file`` parameter of that function as well.  Is that right?

zachmullen:
That's right :)

ronichoudhury:
@zachmullen let me know if this is the right idea.

zachmullen:
Looks great

ronichoudhury:
What should I do to move this forward?

zachmullen:
We just need a test that calls this with a string. We have existing test utilities that perform uploads, see https://github.com/girder/girder/blob/master/clients/web/test/testUtils.js#L703

We can add a new part to the `dataSpec.js` that uploads another item using the string argument.

zachmullen:
And we could tweak something in the test code or the application code to make sure we call either `uploadToItem` or `uploadToFolder` instead of just `upload`, so that we can cover the new wrapping methods.

ronichoudhury:
Thanks for the pointers.  I'll take a look soon when some of my immediate fires are out.

ronichoudhury:
@zachmullen let me know if the changes to the S3 upload handler look right.

zachmullen:
So I was wrong, I just tested a multichunk s3 upload, it is simply

    handler.trigger('g:upload.complete', resp);

The XHR is the request after that to the S3 service, the resp is the file response from Girder.

ronichoudhury:
Do we need to ``delete resp.s3FinalizeRequest`` or anything like that?

zachmullen:
Yeah probably makes sense to remove that.

clients/web/src/model.js
```diff
@@ -4,6 +4,7 @@
  */
 girder.Model = Backbone.Model.extend({
     resourceName: null,
+    idAttribute: "_id",
```
zachmullen: Crockford quotes

zachmullen:
I'm getting the same failure locally that travis is showing, looks like a legit failure.

ronichoudhury:
Is it an S3 error?  I'm not sure I am equipped to diagnose/fix that :(

zachmullen:
I will investigate tonight.

zachmullen:
Can you push this topic to girder/girder and reopen the PR? I will probably make a commit onto the branch to fix that test.

zachmullen:
Closing in favor of https://github.com/girder/girder/pull/755

PR #721, opened 2015-03-06
Sanitize marked markdown output

mgrauer:
@zachmullen turning on the `sanitize` option prevents inline HTML or javascript links like this one that we saw

    [link](javascript:alert(alert))


zachmullen:
LGTM

PR #722, opened 2015-03-06
Raise http error if authentication returns with 404

ronichoudhury:
If, e.g., the ``apiRoot`` is bad, attempting authentication would fail with an error about decoding JSON; now, the code explicitly checks for a 404 response (which, of course, carries no JSON data with it, resulting in the original error) and raises an ``HttpError`` as appropriate.

This results in a better error message, as well as a way to guard against things like bad API roots in client code.


zachmullen:
LGTM

PR #723, opened 2015-03-07
Improve frontpage documentation

zachmullen:


zachmullen:
@girder/developers @girder/owners I'd love to get some feedback on this as I go along if anyone has any good thoughts on what should be in this section. Thanks!

jcfr:
:+1:  I really like where this is going.

docs/index.rst
```diff
@@ -1,11 +1,75 @@
...
+retrieve persistent records. The models contain methods for creating, changing,
+retrieving, and deleting those records.
+
+The primary method of customizing and extending Girder is via the development of
+**plugins**, the process of which is described in the :doc:`plugin-development`
+section of this documentation. Plugins can be turned on or off on a single
+Girder instance to control its functionality dynamically. The extent to which plugins are
+allowed to modify and extend the core system behavior is nearly limitless; they
+can add new REST routes, modify or remove existing ones, serve up a different
+web application from the server root, hook into various events related to models
```
mgrauer: hook into various events related to models => hook into model lifecycle events ?

docs/index.rst
```diff
@@ -1,11 +1,75 @@
...
+or specific API calls, override authentication behavior to support new authentication
+services or protocols, add a new backend storage engine for file storage, interact
+with a completely different DBMS to persist system records -- really almost anything
+can be added or changed with plugins.
```
mgrauer: really almost anything...   => this is a bit redundant with the earlier "extent to which ... nearly limitless", probably can be dropped without losing anything.

docs/index.rst
```diff
@@ -1,11 +1,75 @@
...
+
+Plugins are self-contained in their own directory, therefore they can live in
```
mgrauer: Tried to massage this text a bit:

Plugins are self-contained in their own directory within the Girder source tree.  Therefore they can live in
their own separate source repository, and are installed by simply copying the plugin source tree under an existing Girder installation's `plugins` directory.  The Girder repository contains several generally
useful plugins out of the box, which are described in the :doc:`plugins` section.

docs/index.rst
zachmullen: much better, thanks

docs/index.rst
zachmullen: I agree it's redundant, but I'm having trouble finding a good ending for that long list sentence, any suggestions?

docs/index.rst
```diff
@@ -1,11 +1,75 @@
...
+Girder: A data management platform
+==================================
 
-For a high-level overview of the system, we recommend checking out the :doc:`user-guide`.
+What is Girder?
+---------------
+
+Girder is a free and open source web-based **data management platform**. What does
+that mean? Girder is both a standalone application as well as a platform on top of
+which new web services can be quickly and easily constructed. It's meant to be a
```
mgrauer: I'm a bit torn here--I really like the overall clarity and tone of this approach, as it seems fact-based and relatively cruft-free.  This sentence about "quickly and easily constructed" seems the most like marketing text to me.  Perhaps even changing/merging it with the following sentence could help?  The word "meant" in the following sentence shows the intent for Girder, so applying something like that to "quickly and easily" shows our goals for the project and makes this seem less like marketing.

This is a fairly small quibble (4 nits at most).

zachmullen: Good point, those two sentences probably can be combined. Suggestions on the language?

mgrauer:
I see your point.  How about:

Plugins can add new REST routes, modify or remove existing ones, serve up a different web application from the server root, hook into various events related to models or specific API calls, override authentication behavior to support new authentication services or protocols, add a new backend storage engine for file storage, or even interact with a completely different DBMS to persist system records -- the extent to which plugins are allowed to modify and extend the core system behavior is nearly limitless.

docs/index.rst
mgrauer: Girder is both a standalone application and a platform for building new web services.  It's meant to allow quick and easy construction of web-based applications that have some or all of the following requirements:

mgrauer:
LeGiTiMate

zachmullen:
@mgrauer I actually added one more small addition, let me know if that looks good to you.

cpatrick:
Late LGTM.

[LGTM](http://www.lgtm.in/i/cURPzX9gq)

jeffbaumes:
This looks very nice. Thanks!

PR #728, opened 2015-03-10
Add change directory command in "Install from Git Checkout" admin documentation

jcfr:


cpatrick:
LGTM! Merging.

zachmullen:
@jcfr you now have write access to the repo so you should be able to push branches directly if you prefer, and can also hit the merge button once you have LGTM. Thanks!

PR #734, opened 2015-03-11
Update web external docs example match repo code

chdoig:
Fixed the example in the docs: http://girder.readthedocs.org/en/latest/external-web-clients.html. 

Used the code in the repository example: https://github.com/girder/girder/blob/master/clients/web-external/main.js.


zachmullen:
Good catch, I was looking at the js file not the docs. Thanks for the fix!

PR #737, opened 2015-03-11
Fixes #735. Passing a timeout param to Model.find broke stuff

### LINKED ISSUE #735: Do not fail if multiple identical keyword are passed to girder model method
To provide some background information, following f34130b (`Make all PyMongo "find" queries not timeout`), the parameter `Timeout` is now explictily specified in [girder/models/model_base.py#L214](https://github.com/girder/girder/blob/935c7c5b28a193408285f15d05c28c2c9b0aaf4f/girder/models/model_base.py#L214).

As detailed in https://github.com/girder/covalic/issues/88, this is now causing problem as it is breaking backward compatiblity.

zachmullen:
@jcfr PTAL

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -210,8 +210,11 @@ class Model(ModelImporter):
         if not query:
             query = {}
 
+        if 'timeout' not in kwargs:
+            kwargs['timeout'] = False
+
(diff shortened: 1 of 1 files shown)


jcfr:
:+1:

PR #738, opened 2015-03-12
Fix typos in security.rst

msmolens:


cpatrick:
Thanks, @msmolens LGTM. Merging.

PR #739, opened 2015-03-12
Python client upload callbacks

mgrauer:


zachmullen:
I think since these callbacks can really only be accessed programmatically within python, they belong in `girder_client.GirderClient` rather than within the CLI. As it stands, the example code you wrote in the documentation won't work since these methods are not in the core client class.

mgrauer:
Good catch.  I've updated the docs so the example code would work.

I see your point, but it wouldn't be a matter of moving the callback additions and callback() calls to the GirderClient--since there they wouldn't have context to know when to actually call the callback, for instance on a multi-File Item.  I could move all of [this](https://github.com/girder/girder/blob/python_client_upload_callbacks/clients/python/girder_client/cli.py#L55-L298) from the Cli to the Client, as this code is useful programmatically.  This would leave the Cli as a thin wrapper totally devoted to Cli stuff alone.

zachmullen:
> This would leave the Cli as a thin wrapper totally devoted to Cli stuff alone.

I think that would make for better organization of this code, if it's not too much trouble to move the other stuff in.

mgrauer:
Travis failed a test, somewhat of a race condition.  I'll have the uploadCallback test isolate the directory it is uploading from.

mgrauer:
@zachmullen Travis is finally happy, after several restarts.

mgrauer:
@zachmullen

zachmullen:
LGTM once travis passes

PR #740, opened 2015-03-15
Configurable timeout for our SSE stream

zachmullen:
We set it to 10 seconds in our testing environment since phantom
applies its normal timeout even to SSE connections.

@ronichoudhury If you cherry-pick or merge this onto your topic, it will hopefully fix the issue you encountered with the stream timeout error.


ronichoudhury:
Thanks!  This does seem to fix the event stream timeout error, but leaves me with a different error (which I'll describe in the email thread).

zachmullen:
@ronichoudhury is this good to merge?

ronichoudhury:
Yep.

PR #741, opened 2015-03-16
Adding a logic path to catch "parsererror" status

ronichoudhury:
This occurs when, e.g., you retrieve general text data from an endpoint (say,
``/file/<id>/download``), but forget to set ``dataType: 'text'`` in the request.

### DIFF
--- a/clients/web/src/init.js
+++ b/clients/web/src/init.js
@@ -65,7 +65,7 @@ _.extend(girder, {
             dataType: 'json',
             type: 'GET',
 
-            error: function (error) {
+            error: function (error, status) {
                 var info;
                 if (error.status === 401) {
                     girder.events.trigger('g:loginUi');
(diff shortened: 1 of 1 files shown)


ronichoudhury:
I don't know if the error message I included is appropriate, but thought the mention of ``dataType`` might be good because it's probably a common cause of this error.

ronichoudhury:
Thanks @manthey for helping me track this down.

zachmullen:
Apparently our style tool doesn't like the double quotes: https://travis-ci.org/girder/girder/builds/54618738#L3082

Not sure why that's a thing, probably a Crockford thing.

ronichoudhury:
I am enjoying the irony of "invalid quote mark" followed by a big arrow pointing at a literal quotation mark :smiling_imp:

I too am confused why that's a thing - I actually would expect Crockford to shun single quotation marks - but I made Travis happy.

zachmullen:
LGTM

PR #742, opened 2015-03-16
Using find_package() for Python rather than find_program

ronichoudhury:
This allows the build to work properly for OSes with nonstandard names for, e.g., the Python 2.7 executable, such as Arch Linux (``python2`` instead of ``python``).

CMake documentation for this module: http://www.cmake.org/cmake/help/v3.0/module/FindPythonInterp.html


zachmullen:
LGTM, nice fix.

PR #744, opened 2015-03-17
Display all collection actions after edit

### LINKED ISSUE #743: Incomplete list of collection actions available after collection editing
With Admin permissions, when I first look at a collection I get these actions.

screen shot 2015-03-17 at 1 23 10 pm

After editing and saving the collection, I get a reduced set.

screen shot 2015-03-17 at 1 23 25 pm

mgrauer:
Fixes #743.

~~I didn't have enough time to add a client side test for this.~~

### DIFF
--- a/clients/web/test/spec/collectionSpec.js
+++ b/clients/web/test/spec/collectionSpec.js
@@ -58,7 +58,78 @@ describe('Test collection actions', function () {
     it('create another collection',
         girderTest.createCollection('collName1', 'coll Desc 1'));
 
+    it('change collection description', function () {
+
+        waitsFor(function () {
+            return $('.g-collection-actions-button:visible').is(':enabled');
+        }, 'collection actions link to appear');
+
+        runs(function () {
+            $('.g-collection-actions-button').click();
+        });
+
+        waitsFor(function () {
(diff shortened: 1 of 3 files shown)


cpatrick:
Open an issue to add a test for this and it LGTM.

clients/web/test/spec/collectionSpec.js
```diff
@@ -58,8 +58,79 @@ describe('Test collection actions', function () {
...
+            return girder.numberOutstandingRestRequests() === 0;
+        }, 'dialog rest requests to finish');
+
+        waitsFor(function () {
+            return $('.g-collection-description').text().match('New Description');
+        }, 'description to be updated to new text');
+
+        waitsFor(function () {
+            return $('.g-collection-actions-button:visible').is(':enabled');
+        }, 'collection actions link to appear');
```
zachmullen: In theory multiple consecutive `waitsFor` blocks could be combined into one.

zachmullen: That said, if you think it's cleaner this way, LGTM.

mgrauer:
I think it is easier to see the intent this way, rather than combining.

PR #745, opened 2015-03-17
Write to std err if GIRDER_PORT already bound during tests

mgrauer:
This is a starting point to give better notification when client tests fail silently.

### DIFF
--- a/tests/web_client_test.py
+++ b/tests/web_client_test.py
@@ -18,6 +18,7 @@
 ###############################################################################
 
 import os
+import socket
 import subprocess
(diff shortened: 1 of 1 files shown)


zachmullen:
@mgrauer thanks for checking into this. I've just put in a different PR #746  that addresses this issue, I propose that we go with my solution for now, which is simply not to suppress cherrypy output while it is starting up, which has the benefit of showing us any errors that happen during the startup process instead of just this one error condition.

mgrauer:
sounds good.

PR #747, opened 2015-03-17
Utility functions for UserModel

ronichoudhury:
Added three functions to ``UserModel``:
- ``current()`` - retrieves the logged in user from the ``/user/me`` endpoint and, if the response is not ``null``, fills in the model with the user model data
- ``login()`` - attempts to log a user in by providing the given username and password, fills the model in with user model data on success
- ``logout()`` - logs out the currently logged in user

The purpose of these functions is to provide a bit of abstraction over the low-level calls to client programs wanting to log a user in or out, etc (in similar fashion as the ``FileModel`` upload methods).

Some questions/concerns:
1. These functions have explicit arguments for the success and failure callbacks.  Should they be delivered via an ``options`` arguments object, as is the case for most Backbone interfaces?
2. The ``logout()`` function is kind of weird in that it can cause a global effect no matter which particular ``UserModel`` object calls it.  Should the function first check to see that the logged in user is the same as the user represented by the model?  In practical terms this would mean that in order to call ``logout()`` you would first simply need to call ``current()``.  That seems kind of silly.  Not sure what the right semantics are here.
3. Similarly for ``login()`` - that function allows any ``UserModel`` object whatsoever to override the currently logged in user.  Of course, this is how the system works and anyone with the power to call this function can also issue the underlying request to to ``/user/authentication`` endpoint.
4. Right now I am passing the model as the first argument to the success/error callbacks.  Is it better to call them with the model set as the ``this`` context?  Is there any need to pass along the server response when it has already been set on the model itself?


zachmullen:
I do like having convenience methods for `login` and `logout`; we currently don't have those and we need them. However I think they might be better placed as global methods directly under the girder namespace, i.e. `girder.login()` and `girder.logout()` rather than on the `UserModel`, because they affect the global state of the application. As for `current()`, we actually already have `girder.currentUser` that stores that value, though it might not hurt to have a method that will re-check the server to see if the token is still valid. Again I think that belongs outside of the user model since it isn't manipulation of a user resource so much as querying of an authentication state in the runtime context.

For the callbacks, in general we prefer to use the observer semantics provided by the Backbone events system; the model can simply call `this.trigger('g:success')` or `this.trigger('g:failure')` and it can be left up to the caller to bind to that event.

zachmullen:
Once done you can replace the last two locations where we call `girder.restRequest` outside of the model layer (which is bad practice) with calls to your new convenience methods:

https://github.com/girder/girder/blob/master/clients/web/src/views/layout/LoginView.js#L9
https://github.com/girder/girder/blob/master/clients/web/src/views/layout/HeaderUserView.js#L16

ronichoudhury:
>
> I do like having convenience methods for login and logout; we currently
> don't have those and we need them. However I think they might be better
> placed as global methods directly under the girder namespace, i.e.
> girder.login() and girder.logout() rather than on the UserModel, because
> they affect the global state of the application.
>
​Great, I tend to agree.​

> As for current(), we actually already have girder.currentUser that stores
> that value,
>
​I saw this when I was writing tests the last couple of days, so I thought
I'd see if it works under Sitar.  However, in Sitar, regardless of whether
any user is logged in or out, ``girder.currentUser`` is null when I print
it out in the console.  How is that variable supposed to​ get set
during/after hitting the ``/user/authentication`` endpoint to log in?

> though it might not hurt to have a method that will re-check the server to
> see if the token is still valid. Again I think that belongs outside of the
> user model since it isn't manipulation of a user resource so much as
> querying of an authentication state in the runtime context.
>
​I actually think having that ``current()`` function in the global scope
(probably renamed to ``currentUser()``) actually makes more sense than
making Girder maintain a global variable that has the user model in it.
Then, whenever I call that function, I would get either ``null`` or a valid
``UserModel``, ​and the function of checking the server for a valid token
would also be solved (as far as I understand - especially since
``girder.currentUser`` doesn't seem to be working for me).

One argument against this mechanism is that presumably it takes many fewer
bytes of network traffic to confirm or disconfirm the validity of a token
than it does to return a representation of the current user.  However, I
also am not sure that conserving bandwidth in this instance is a
make-or-break type proposition.

Thoughts?

> For the callbacks, in general we prefer to use the observer semantics
> provided by the Backbone events system; the model can simply call
> this.trigger('g:success') or this.trigger('g:failure') and it can be left
> up to the caller to bind to that event.
>
​Right - I should have absorbed this bit of wisdom by now.  I'll recast the
login/logout functions to conform to this behavior.  One question - if
those functions become globals, I assume I would just invent global events
to go along with success/failure of login/logout?  Something like
``g:login.success``, ``g:logout.failure``, etc.?​

ronichoudhury:
>
> Once done you can replace the last two locations where we call
> girder.restRequest outside of the model layer (which is bad practice)
> with calls to your new convenience methods:
>
>
> https://github.com/girder/girder/blob/master/clients/web/src/views/layout/LoginView.js#L9
>
> https://github.com/girder/girder/blob/master/clients/web/src/views/layout/HeaderUserView.js#L16
>
​Ahhh, sweet sweet refactoring!​

ronichoudhury:
Ugh, I will also fix the double quotes.  I'm going to get deported from quotelandia at this rate.

ronichoudhury:
We talked a bit about how ``g:login`` is maybe misnamed.  I went ahead and invented new event names: ``g:login.success``, ``g:login.error``, ``g:logout.success``, and ``g:logout.error``.  I really think these names are useful to people writing clients against Girder.

For backward compatibility, I still emit ``g:login`` with ``null`` passed as the data on logout, but I think we could really think about changing the name of that event.  I realize that breaks backward compatibility, but... at worst we could deprecate it and abandon it in the eventual 2.x line.  Just something to consider.

ronichoudhury:
@zachmullen bam, travis is green :smile_cat:

clients/web/src/models/UserModel.js
```diff
@@ -1,6 +1,26 @@
 girder.models.UserModel = girder.Model.extend({
     resourceName: 'user',
 
+    // Sets this model to the currently logged in user; has no effect is no user
+    // is logged in.
+    current: function (success, error) {
```
zachmullen: For consistency, can we have this trigger events rather than accepting callbacks as arguments?

clients/web/src/init.js
```diff
@@ -130,6 +130,43 @@ _.extend(girder, {
...
+        var auth = 'Basic ' + window.btoa(username + ':' + password);
+
+        return girder.restRequest({
+            method: 'GET',
+            path: '/user/authentication',
+            headers: {
+                Authorization: auth
+            }
+        }).then(function (response) {
+            girder.events.trigger('g:login.success', response.user);
```
zachmullen: For  backward compatibility let's make sure this still triggers `g:login` on success.

zachmullen:
Nice model layer functions, now we should make our app use them in the following places:

https://github.com/girder/girder/blob/master/clients/web/src/app.js#L3-L5
https://github.com/girder/girder/blob/master/clients/web/src/views/layout/LoginView.js#L9-L18
https://github.com/girder/girder/blob/master/clients/web/src/views/layout/HeaderUserView.js#L16-L22

Converting those will get automatic testing coverage for your new code. :+1:

clients/web/src/models/UserModel.js
```diff
@@ -1,6 +1,26 @@
...
+        }).then(_.bind(function (user) {
+            if (user) {
+                this.set(user);
+                if (success) {
+                    success(this);
+                }
+            } else {
+                if (error) {
+                    error(this);
+                }
```
zachmullen: I don't think we want to treat this as an error condition since `/user/me` is meant as a way to test whether or not you're logged in. It makes it a bit tricky since you put this in the model layer; maybe it makes more sense to move it out under the `girder` object like the other two functions you added?

ronichoudhury: I think there are two possibilities:

1. Instead of success/error, we call either ``this.set()`` or ``this.clear()`` depending on whether ``user`` is ``null``.  Clients can react to the model's ``change`` event and then call ``isNew()`` etc. to see the result of the call to ``current()``.

2. We just dump this function altogether, since the same effect can be achieved by calling ``girder.fetchCurrentUser()`` followed by ``userModel.set()``/``userModel.clear()``.

The latter is obviously simpler, but my whole intent was to provide some abstraction to client writers, even if it's only a small amount - this instance would eliminate the need to check the result of ``fetchCurrentUser()`` by hand, for example.  I think that's useful but obviously I'll defer to your judgment.

clients/web/src/models/UserModel.js
ronichoudhury: Whatever we decide on the [other question](https://github.com/girder/girder/pull/747#discussion_r27307636) will solve this as well.

clients/web/src/init.js
ronichoudhury: Done in 53e568f9252ce5b024ffefe0dda2e95cb6bf1986

clients/web/src/models/UserModel.js
zachmullen: I'm fine with either of those solutions. If we go with 1, let's just document that in the docstring and maybe also mention `girder.fetchCurrentUser` in the docstring.

ronichoudhury: Opting for 1 (via 3803da502d3ef8e5179598b87d0f65d0c2f9fa71).  Let me know if the documentation is ok.

ronichoudhury:
I think I've addressed all the points - let me know if I missed anything.

clients/web/src/views/layout/HeaderUserView.js
```diff
@@ -13,10 +13,9 @@ girder.views.LayoutHeaderUserView = girder.View.extend({
         },
 
         'click a.g-logout': function () {
-            girder.restRequest({
-                path: 'user/authentication',
-                type: 'DELETE'
-            }).done(_.bind(function () {
+            girder.logout();
+
+            girder.events.once('g:logout.success', _.bind(function () {
```
zachmullen: I think this will make it so `g:login` gets triggered twice, once here and once by `girder.logout`. I think it might make sense to remove this binding altogether and simply have `girder.login` and `girder.logout` update the value of `girder.currentUser` internally.

ronichoudhury: I'm not completely sure what you mean, but I think you're saying that there's no reason to set ``girder.currentUser`` here specifically - it can just be a general thing that Girder does by listening to the login/logout events.  Is that right?

ronichoudhury: Ah, I see the duplicated triggering of ``g:login`` but that seems like an orthogonal issue.  I'll move the ``girder.currentUser`` stuff to the ``login()`` and ``logout()`` functions.

zachmullen: Cool, that's what I was getting at :)

clients/web/src/init.js
```diff
@@ -130,6 +130,51 @@ _.extend(girder, {
...
+
+    logout: function () {
+        return girder.restRequest({
+            method: 'DELETE',
+            path: '/user/authentication'
+        }).then(function () {
+            girder.events.trigger('g:login', null);
+            girder.events.trigger('g:logout.success');
+
+            girder.currentUser = null;
```
zachmullen: This needs to go above line 164 so that handlers of `g:login` that inspect the value of `girder.currentUser` will see the new value. (Currently this breaks the core web client logout button behavior.)

ronichoudhury: Oops.  Fixed.

zachmullen:
LGTM

PR #748, opened 2015-03-18
Also output messages for plugins with custom grunt targets

zachmullen:


zachmullen:
@mgrauer PTAL

mgrauer:
LGTM, tested with Covalic.

PR #749, opened 2015-03-18
javascript_tests_init: Update CMake function to check if directory exists

jcfr:
If the input parameter is a relative directory, it is prepended with
the expected working directory.

See girder/covalic#93


jcfr:
Closing this PR. An other one will be submitted shortly with a more meaningful name.

PR #750, opened 2015-03-18
Fixes error when user logs in when core router disabled

zachmullen:
This is a bug in the COVALIC app and presents in any custom app
that uses the core login widget.

@cpatrick PTAL


cpatrick:
When Travis is happy, this LGTM.

PR #752, opened 2015-03-18
Configurable config file location for js static analysis tests

zachmullen:
@jcfr


zachmullen:
https://github.com/jshint/jshint/issues/833

PR #753, opened 2015-03-18
Make marked optional

ronichoudhury:
This makes marked optional by performing a check for ``window.marked`` and constructing an error function if it is not available.


zachmullen:
LGTM once travis passes

PR #758, opened 2015-03-23
Use Travis container-based builds.

manthey:
Our Travis builds were erroring-out because of intermittent problems with Travis's virtualization method.  The people at Travis recommended switching to their container-based builds.  This requires not using sudo, though you can still use apt-get from the default repositories.  I had to install Mongo manually, and, that change required a change to how testing for bad assetstores was done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: cpp
 
 sudo: false
 
+# Build matrix
+env:
+    - MONGO_SOURCE=mongodb-linux-x86_64-2.6.8.tgz
+    - MONGO_SOURCE=mongodb-linux-x86_64-3.0.1.tgz
+
 compiler:
     - gcc
 before_install:
@@ -11,9 +16,7 @@ before_install:
     #- python scripts/vmtest.py
 
     - pushd "${HOME}"
-    # We can test with Mongo 2.6.8 or Mongo 3.0.1
-    #- curl -O https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-3.0.1.tgz | gunzip -c | tar x
-    - curl https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-2.6.8.tgz | gunzip -c | tar x
+    - curl "https://fastdl.mongodb.org/linux/${MONGO_SOURCE}" | gunzip -c | tar x
     - cd mongodb-*/bin && export PATH="${PWD}:${PATH}"
     - popd
     - mkdir /tmp/db
@@ -32,7 +35,6 @@ install:
     - pip install --user -r plugins/celery_jobs/requirements.txt -r plugins/hdfs_assetstore/requirements.txt
     # We have to upgrade six or requiring moto can cause other modules to fail
     - pip install --user -U six
-    - python -c "import bcrypt"
     - npm install grunt grunt-cli
     - npm install
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,24 @@
 language: cpp
+
+sudo: false
+
 compiler:
     - gcc
 before_install:
     # Show the memory state; this lets us more quickly determine when the
     # travis environment is bad
     - vmstat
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -18,12 +26,14 @@ before_install:
...
-    - sudo pip install -r plugins/celery_jobs/requirements.txt -r plugins/hdfs_assetstore/requirements.txt
+    - mkdir -p "${HOME}/.local/bin"
+    - export PATH="${HOME}/.local/bin:${PATH}"
+    - pip install --user -r requirements.txt -r requirements-dev.txt -r plugins/geospatial/requirements.txt -r plugins/metadata_extractor/requirements.txt
+    - pip install --user -r plugins/celery_jobs/requirements.txt -r plugins/hdfs_assetstore/requirements.txt
     # We have to upgrade six or requiring moto can cause other modules to fail
-    - sudo pip install -U six
-    - sudo python -c "import bcrypt"
-    - npm install -g grunt grunt-cli
+    - pip install --user -U six
```
jbeezley: I don't think the `--user` is strictly necessary because travis builds are always done in a userspace virtualenv.  Getting rid of the sudo probably make the initial `import bcrypt` below unnecessary as well.

manthey: It didn't work without the --user (and the Travis folks told me to use it).

I agree, the bcrypt is unnecessary.

On Mon, Mar 23, 2015 at 11:00 AM, Jonathan Beezley <notifications@github.com
> wrote:

> In .travis.yml
> <https://github.com/girder/girder/pull/758#discussion_r26943979>:
>
> >      # We have to upgrade six or requiring moto can cause other modules to fail
> > -    - sudo pip install -U six
> > -    - sudo python -c "import bcrypt"
> > -    - npm install -g grunt grunt-cli
> > +    - pip install --user -U six
>
> I don't think the --user is strictly necessary because travis builds are
> always done in a userspace virtualenv. Getting rid of the sudo probably
> make the initial import bcrypt below unnecessary as well.
>
> —
> <https://github.com/girder/girder/pull/758/files#r26943979>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

zachmullen:
Wow, so our CI builds now run much faster and work the first time. You rock @manthey !

LGTM

zachmullen:
(just saw @jbeezley 's comments, sorry for premature LGTM)

.travis.yml
jbeezley: Right, I was thinking of readthedocs...

manthey:
It is pretty easy to use a build matrix.  Is there any reason NOT to test with Mongo 2.6.8 and Mongo 3.0.1?

zachmullen:
Now that builds are no longer failing most of the time, I can't think of any reason not to do that.

manthey:
When using a build matrix, you obviously have to wait for all of the items in the matrix to complete.  In my most recent submission, the PR build finished in a timely manner (though the reported build time is the sum of all matrix options).  The PUSH build finished one of its two halves just fine, but the other half had a spurious build error (which I hadn't personally seen in a while).  I've restarted it to see what happens.

manthey:
@jbeezley Any other issues?

jbeezley:
From my experience with gaia, the different builds in the build matrix will actually run in parallel depending on available resources.  One small thing that would be nice to add is to add the env information to the cdash submission like I did with python versions [here](http://my.cdash.org/index.php?project=Gaia&date=2015-03-21).  Otherwise, you can't easily tell which is which from cdash.

manthey:
Also, I should only have one of the matrix elements do the deployment.

manthey:
@jbeezley I added Mongo version information to the cdash submission (see [here] (http://my.cdash.org/index.php?project=girder)).  Hopefully my code to only deploy once will work, but I'm not sure how to test that without merging the PR.

jbeezley:
Nice!  LGTM.

PR #759, opened 2015-03-23
Add a model type field to models when filtering them

zachmullen:
I'm putting this up as a straw man; I could see this field being helpful when objects are returned via the REST API to identify what types are being dealt with. Can anyone come up with a good reason *not* to add this field?


manthey:
I like it.  I'd go so far as to suggest that there even should be a function of some sort that takes an ID and returns the model type of that ID if it exists at all.  I'm not sure what the direct usage of this would be, but when trying to debug something it would be useful.

manthey:
LGTM.

PR #762, opened 2015-03-23
Bug squashing ascii-only pitfalls

zachmullen:
@cpatrick this fixes the bug described in your email.

@manthey ping


cpatrick:
Nice test, and LGTM. It might be good to add a not about the `.format` vs `%` discussion that we constantly have in the docs :smile_cat:

PR #764, opened 2015-03-25
Show size next to files in file list view

zachmullen:
This is especially helpful for items containing many files


manthey:
LGTM.

PR #765, opened 2015-03-25
Mongo version wasn't being used properly.

manthey:
I failed to convert a variable name, so we aren't currently building two mongo version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
     #- python scripts/vmtest.py
 
     - pushd "${HOME}"
-    - curl "https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${MONGO_SOURCE}.tgz" | gunzip -c | tar x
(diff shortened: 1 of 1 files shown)


zachmullen:
Does that mean `MONGO_SOURCE` was magically defined somewhere?

zachmullen:
To answer my own question: apparently somebody has thought of this because `https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-.tgz` is an actual download link :)

manthey:
No, we were downloading a file from Mongo that turned out to be version 2.6.2.rc0.  I'm surprised it worked.  Thanks to @brianhelba for pointing out this problem.

brianhelba:
@zachmullen No, it was evaluating to an empty string, but serendipitously, the resulting URL did actually serve a file with some arbitrary version of Mongo (v2.6.2-rc0).

brianhelba:
LGTM.

brianhelba:
Also, I can confirm that the Travis build logs are showing both correct versions of MongoDB are now running.

PR #766, opened 2015-03-25
Update PyMongo requirement to v2.8

brianhelba:
Version 2.8 of PyMongo is necessary for full support of MongoDB 3.0.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 cffi==0.8.6
 CherryPy==3.3.0
-pymongo==2.7.1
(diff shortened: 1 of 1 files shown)


brianhelba:
FYI: https://api.mongodb.org/python/2.8/changelog.html

manthey:
LGTM as soon as travis is happy.

Incidentally, I've been using Mongo 3 with pymongo 2.7.1 without any issues (though I wasn't using the new authentication method).  pymongo 2.8 should only be an improvement.

PR #772, opened 2015-03-25
Use prefix-based checking for CORS Content-Type validation

zachmullen:


zachmullen:
@jcfr I just force pushed a more robust fix.

zachmullen:
@jcfr Thanks, that is actually what I meant to do :)

brianhelba:
LGTM. According to RFC2046, only "text/plain" can actually have a "charset", but this is a fine way to deal with it generally.

zachmullen:
In general, [RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7) dictates that media types may contain 0 or more semicolon-delineated parameters after the type and subtype. We also see parameters in practice with the multipart/form-data declaration, which looks like:

    multipart/form-data; boundary=----WebKitFormBoundary85YOGRliKRJPFZM2

manthey:
@zachmullen Thanks for hunting this down and fixing it.

PR #773, opened 2015-03-26
Fix race condition in jobs plugin phantom test

zachmullen:
I couldn't recreate this failure locally, but I've seen it several times on travis and the [screenshots](http://my.cdash.org/testDetails.php?test=18817932&build=743273) from the failure case make it fairly clear what's happening. The app has not yet loaded due to waiting on the `/user/me` xhr to finish.


manthey:
LGTM.

PR #774, opened 2015-03-26
Double the timeout when waiting for replica sets

zachmullen:
This only fails on mongo 3.0.1 on travis, and only occasionally.
It seems sometimes the full replica set can take longer than 60s
to start up in the parallel environment.

@manthey ping

### DIFF
--- a/tests/mongo_replicaset.py
+++ b/tests/mongo_replicaset.py
@@ -111,7 +111,7 @@ def pauseMongoReplicaSet(pauseList, verbose=0):
             pprint.pprint(stat)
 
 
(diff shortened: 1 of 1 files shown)


manthey:
@zachmullen It isn't just a timing issue.  Specifically, I expected the first replica set to be the primary, but in some instances the second one becomes the primary (the first never does).

Specifically, I'm checking if the replica sets are returning statuses of [1, 2, 2] (primary, secondary, secondary), but they end up as [2, 1, 2].  I'll work on making this not care which one comes up first.

zachmullen:
Sounds good, closing this.

PR #776, opened 2015-03-26
Add 2015 to docs Copyright

mgrauer:


zachmullen:
I remember a while back someone claimed that we didn't have to put a year on these things at all, at least not in our headers. Is that still the case? If so, does it also apply to these tags?

manthey:
Copyright law actually has shifted around a bit.  Strictly speaking, after
the Berne Convention in the 1989 you don't need a copyright message at all
-- any content you create is automatically copyright to the creator, but it
is still recommended.  According to the US Copyright Office, failure to
include a date is the same as NOT have a notice at all. [1]

In my opinion, we should have a date (the date of first publication).
Having a list of dates that include first publication and last modification
would be fine, too.

We could always ask someone with real legal knowledge, though.

- David

[1] http://www.copyright.gov/circs/circ03.pdf  See section on Omission of
Notice.

On Thu, Mar 26, 2015 at 2:12 PM, Zach Mullen <notifications@github.com>
wrote:

> I remember a while back someone claimed that we didn't have to put a year
> on these things at all, at least not in our headers. Is that still the
> case? If so, does it also apply to these tags?
>
> —
> <https://github.com/girder/girder/pull/776#issuecomment-86652585>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

manthey:
Anyway, LGTM.

mgrauer:
I got this sudden urge to update it, is all I can say.  I'm fine with closing this PR and leaving it with just 2014, as @manthey says that covers us, and we won't need to update it in future years.

zachmullen:
Doesn't much matter to me, I never think about the copyright stuff much anyway since our source code is distributed free and with no strings attached for use.

cpatrick:
:hammer: MERGE HAMMER

PR #777, opened 2015-03-26
Replica sets don't always come online in order.

manthey:
Usually, in Mongo 2.6, replica sets come up in the order started, with the first one being the primary.  Sometimes in Mongo 3 (and possibly more rarely in Mongo 2.6), this is not the case, and the first replica set starts as a secondary to one of the others.  Our tests assumed that the first RS would be the initial primary, and would therefore sometimes fail.

This doesn't care which starts out as the primary, and we reorder our internal list so that when we pause an RS, we pause the primary (not necessarily the first).

I've verified that this cures the web_client_data_gridfsrs and server_upload tests.


zachmullen:
Lgtm, thanks for tracking this down

PR #782, opened 2015-03-29
Plugin override layout

mgrauer:
Allows a route to specify an empty layout (only the `g-app-body-container` will be shown, other containers are hidden) via an option passed to `navigateTo`.


clients/web/src/app.js
```diff
@@ -44,6 +45,45 @@ girder.App = girder.View.extend({
         girder.events.on('g:login', this.login, this);
     },
 
+    _initLayoutRenderMap: function () {
+        var hide_on_empty = ['#g-app-header-container', '#g-global-nav-container', '#g-app-footer-container'],
```
zachmullen: A minor nit, I suggest we stick to camel case variable names for consistency.

clients/web/src/app.js
```diff
@@ -59,14 +99,31 @@ girder.App = girder.View.extend({
...
      * @param [settings={}] Settings to pass to the view initialize() method.
+     * @param opts Additional options for navigateTo, if any.
      */
-    navigateTo: function (view, settings) {
+    navigateTo: function (view, settings, opts) {
         var container = this.$('#g-app-body-container');
 
         this.globalNavView.deactivateAll();
 
         settings = settings || {};
```
zachmullen: Let's add `opts = opts || {}` to make this easier to extend.

clients/web/src/app.js
```diff
@@ -59,14 +99,31 @@ girder.App = girder.View.extend({
...
 
+        if (opts && opts.layout) {
+            if (girder.layout !== opts.layout) {
+                if (_.has(this.layoutRenderMap, opts.layout)) {
+                    // set a layout if opts specifies one different from current
+                    girder.layout = opts.layout;
+                    this.layoutRenderMap[girder.layout]();
```
zachmullen: Let's change these lines to `this.layoutRenderMap[girder.layout].call(this, opts);`. That will let us remove the `_this` bindings above and just use regular `this` instead, and also will allow any special options for layout rendering to be passed through to the function.

zachmullen:
Nice job, this will be a nice reusable feature for plugins that wish to set custom layouts for specific pages. Other than the minor issues I noted inline, this just needs some developer documentation (maybe in plugin development section) and a web client test.

mgrauer:
Thanks for the QA.  I'll have to pick this back up April 1.  :foggy:

mgrauer:
@zachmullen QA points addressed, tests and docs added.

clients/web/src/app.js
```diff
@@ -43,6 +44,44 @@ girder.App = girder.View.extend({
         girder.events.on('g:login', this.login, this);
     },
 
+    _initLayoutRenderMap: function () {
+        var hideOnEmpty = ['#g-app-header-container', '#g-global-nav-container', '#g-app-footer-container'],
+            css_empty = {
```
zachmullen: camel case

docs/development.rst
```diff
@@ -135,6 +135,15 @@ There are many ways to filter tests when running CTest, or run the tests in
...
 `here <http://www.cmake.org/cmake/help/v3.0/manual/ctest.1.html>`_.
 
+If you run into errors on any of the packaging tests, two possible fixes are
+
+1) run ``make`` inside your ``girder-build`` directory, which will create a special
+virtualenv needed to build the packages.
+
+2) delete any of the files generated by the packaging tests, which will be in your
+source dir ``girder`` and could include ``girder-1.2.4.tar.gz``, ``girder-web-1.2.4.tar.gz``,
+and ``girder-plugins-1.2.4.tar.gz``.
```
zachmullen: We should probably just list these as girder-<version>.tar.gz instead of hard-coding the version, so we don't have to worry about keeping these docs up to date with each release.

docs/development.rst
```diff
@@ -143,6 +152,40 @@ The coverage data will be automatically generated. After the tests are run,
...
+Using the same setup as above for the Server Side Tests, your environment will be set up
+to run client side tests. Running ::
+
+    cd girder-build
+    ctest
+
+will run all of the tests, which include the client side tests.  Our client tests use the
+Jasmine JS testing framework.
+
+When running client side test, if you try to SIGINT (ctrl+c) the CTest process, CTest
```
zachmullen: tests*

zachmullen:
@mgrauer I like what has come out of this but I am considering making some changes to the extension API that will allow this to be more flexible for the addition of other dynamic layouts. Namely, I think I might change the map from `layout -> function` to `layout -> {show: function, hide: function}`. The API for rendering will remain the same, and I will be able to reuse your test and docs and will go ahead and do it on this branch.

zachmullen:
@mgrauer as promised, I finally got around to implementing what I mentioned, PTAL at that last commit.

mgrauer:
Why the changes in main.js here?  They don't appear to be used.

zachmullen:
That's not to do with this topic, I just piggybacked it off of this commit. It provides a nice way for plugins to get a handle on the main app view if they need to.

mgrauer:
@zachmullen Looks good from my end.  I fixed up the test and added a small change.  I'm passing the QA baton back to you.

I like the direction of these changes, both applying the style and having a layout that knows how to show/hide itself.

zachmullen:
LGTM!

PR #784, opened 2015-03-30
Add d3 to jshint global predefs

mgrauer:
d3 was already included as a downloaded library in girder core.  this change adds it to the jshint globals predef.


zachmullen:
LGTM

PR #787, opened 2015-04-01
Move to an official release of cmake.

manthey:
We were on a release candidate.  This is the first official release that supports everything we want.  The only relatively new feature we need from cmake is the ability to report our coverage files to cdash.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_install:
 
     - mongod --version
     - pushd "${HOME}"
-    - curl "http://www.cmake.org/files/v3.1/cmake-3.1.0-rc3-Linux-x86_64.tar.gz" | gunzip -c | tar x
+    - curl "http://www.cmake.org/files/v3.1/cmake-3.1.0-Linux-x86_64.tar.gz" | gunzip -c | tar x
     - cd cmake-*/bin && export PATH="${PWD}:${PATH}"
     - popd
     - cmake --version


zachmullen:
LGTM

aashish24:
CMake 3.2.1 is out. Why use a older version?

manthey:
As per a conversation on girder-developers, we should should use the oldest official version that supports what we need, so as to illustrate the minimum dependencies.  We should occasionally test separately with the latest cmake (and with release candidates).

aashish24:
Sure but this is just for travis right? But please do what we all agree.

manthey:
Yes, just for Travis.  The nightly builds can use a different cmake version.

aashish24:
oops.. hit send too soon. I guess we want users to use 3.1 as well so using a older version probably make sense.

zachmullen:
There is still the question of what to put as our `
cmake_minimum_required(VERSION`. Right now it's set to 2.8.6, since that works fine for everything except coverage reporting to CDash. I'm fine with keeping it at that since the latest version in APT is still only 2.8.12. We can rethink that value perhaps once the APT package gets updated, but I have no idea on a timeline for that.

jbeezley:
The next release is set to get 3.0.2.

https://launchpad.net/ubuntu/vivid/amd64/cmake/3.0.2-1ubuntu2

Issue #446, opened 2014-10-28
Swagger throbber image is broken


manthey:
When you execute a task in the swagger rest api area, the 'throbber' image is supposed to appear for the duration of the execution.  The link to this image is broken (currently it points to /api/images/throbber.gif).

manthey:
Fixed in PR #788.

zachmullen:
@manthey I still see this error on latest master with #788 merged in

`GET http://localhost:9080/girder/api/images/throbber.gif 404 (Not Found)`

Is there something I have to do besides running `npm install` to get it working?

manthey:
It first tries to load the throbber image from the wrong place, then the css makes it load it from the correct place.  This means that you get a 404 in the logs, but the throbber image appears to the user.

zachmullen:
Nevermind my last comment, my server simply hadn't updated. I'm running into other issues now when I'm behind a proxy, but the throbber thing was not a real error.

Issue #320, opened 2014-09-10
Add client-side tests of Swagger docs.


cpatrick:

manthey:
Added in PR #792.

PR #788, opened 2015-04-03
Several changes to make deployment easier.

manthey:
The swagger api pages would only work if the static_root was absolute, because this was otherwise treated as relative to the api path.  I've added a new api_static_root parameter, which now allows all of the paths to be relative.  This makes it easier, for instance, for proxies to get to these locations.

The api_root is placed in the swagger page as well, as this also also the sub-describe calls to use a relative path instead of the entire query path.

Fixed the swagger throbber using css to override its path.  This still throws a 404 in the browser's console, when the original location is requested, but the throbber is properly displayed.  This fixes issue #446.

Add documentation that you don't need to specify a tools.proxy.base or tools.proxy.local if the proxy server adds X-Forwarded-Host.

Expand the user path element when parsing the log root, not just in the constants.  This allows a conf file to use ~ in the log root.

### DIFF
--- a/clients/web/static/girder-swagger.js
+++ b/clients/web/static/girder-swagger.js
@@ -1,6 +1,11 @@
 $(function () {
+    var apiRoot = $('#g-global-info-apiroot').text().replace(
+        '%HOST%', 'http://' + window.location.host);
+    if (!apiRoot) {
+        apiRoot = window.location.origin + window.location.pathname;
+    }
     window.swaggerUi = new SwaggerUi({
-        url: window.location.origin + window.location.pathname + '/describe',
+        url: apiRoot + '/describe',
         dom_id: 'swagger-ui-container',
         supportHeaderParams: false,
         supportedSubmitMethods: ['get', 'post', 'put', 'delete'],
--- a/clients/web/static/girder-swagger.js
+++ b/clients/web/static/girder-swagger.js
@@ -1,6 +1,6 @@
 $(function () {
     var apiRoot = $('#g-global-info-apiroot').text().replace(
-        '%HOST%', 'http://' + window.location.host);
+        '%HOST%', window.location.origin);
     if (!apiRoot) {
         apiRoot = window.location.origin + window.location.pathname;
     }
--- a/girder/api/describe.py
+++ b/girder/api/describe.py
@@ -172,7 +172,7 @@ class ApiDocs(object):
     <div class="docs-header">
       <span>Girder REST API Documentation</span>
       <i class="icon-book-alt right"></i>
-      <div id="g-global-info-apiroot" class="hide">${apiRoot}</div>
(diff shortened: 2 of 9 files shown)


clients/web/static/girder-swagger.js
```diff
@@ -1,6 +1,11 @@
 $(function () {
+    var apiRoot = $('#g-global-info-apiroot').text().replace(
+        '%HOST%', 'http://' + window.location.host);
```
cpatrick: What happens if we're https only?

manthey: This replace only occurs in tests, not in actual deployment.  We don't
currently test the swagger docs (but we should -- see issue #320), so this
doesn't matter yet.

On Fri, Apr 3, 2015 at 10:25 AM, Patrick Reynolds <notifications@github.com>
wrote:

> In clients/web/static/girder-swagger.js
> <https://github.com/girder/girder/pull/788#discussion_r27731665>:
>
> > @@ -1,6 +1,11 @@
> >  $(function () {
> > +    var apiRoot = $('#g-global-info-apiroot').text().replace(
> > +        '%HOST%', 'http://' + window.location.host);
>
> What happens if we're https only?
>
> —
> <https://github.com/girder/girder/pull/788/files#r27731665>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

ronichoudhury: I assume "window.location.protocol" could replace the hard-coded http://?

roni

manthey: That would be nicer (this same replace is also in clients/web/src/init.js).

zachmullen: Doesn't `window.location.origin` capture the protocol, host, and port?

manthey: Does IE support `window.location.origin` yet?  Do we care?

manthey: In answer to my last question, IE 11 supports it, and we don't care.

girder/api/describe.py
```diff
@@ -139,76 +139,83 @@ class ApiDocs(object):
...
+.response_throbber {
+  content:url("${staticRoot}/built/swagger/images/throbber.gif");
+}
+    </style>
+  </head>
+  <body>
+    <div class="docs-header">
+      <span>Girder REST API Documentation</span>
+      <i class="icon-book-alt right"></i>
+      <div id="g-global-info-apiroot" class="hide">${apiRoot}</div>
```
zachmullen: This doesn't appear to be hidden correctly, maybe need to just directly set `display: none`.

screen shot 2015-04-05 at 8 04 06 pm

zachmullen:
LGTM

PR #790, opened 2015-04-06
CORS simple requests should be restricted to the CORS policy.

manthey:
I had misinterpretted the documentation on CORS that simple requests were always allowed.  Rather, they just don't require "preflighting" with an OPTIONS request.  This change subjects them to the same restriction as other requests.

For maximum compliance, we should cache approved OPTIONS requests, and non-simple requests should be denied if they haven't sent a corresponding OPTIONS request recently.

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -376,13 +376,16 @@ def _setCommonCORSHeaders(isOptions=False):
     origin = '%s://%s' % (originparse.scheme, originparse.netloc)
     cherrypy.response.headers['Access-Control-Allow-Origin'] = origin
     cherrypy.response.headers['Vary'] = 'Origin'
-    # Some requests do not require further checking
-    if (cherrypy.request.method in ('GET', 'HEAD') or (
-            cherrypy.request.method == 'POST' and cherrypy.request.headers.get(
-            'Content-Type', '').split(';', 1)[0].strip() in (
-                'application/x-www-form-urlencoded',
-                'multipart/form-data', 'text/plain'))):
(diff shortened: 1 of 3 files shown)


girder/api/rest.py
```diff
@@ -376,13 +376,16 @@ def _setCommonCORSHeaders(isOptions=False):
...
-    if (cherrypy.request.method in ('GET', 'HEAD') or (
-            cherrypy.request.method == 'POST' and cherrypy.request.headers.get(
-            'Content-Type', '').split(';', 1)[0].strip() in (
-                'application/x-www-form-urlencoded',
-                'multipart/form-data', 'text/plain'))):
-        return
+    # Some requests are "simple", and don't require the browser to have sent an
+    # OPTIONS request before them.  If we want to be stricter, we would log
+    # what OPTIONS requests have occurered, and only respond to non-simple
+    # requests that had been preflighted.
```
zachmullen: Is this true? I thought we could simply rely on the browser to enforce the preflighting rules.

manthey: We *can* rely on the browser, since a non-compliant browser could always just not send an `ORIGIN` header.  To be the most pedantic, though, we would enforce getting an `OPTIONS` preflight before another method.  I don't think it is really necessary as only non-complaint browsers would be affected.

zachmullen: Agreed, my vote would be to remove the comment block because I think what we have is sufficient.

zachmullen:
LGTM, thanks for fixing

PR #791, opened 2015-04-06
We were erroneously setting the "consumes" field when it was empty.

zachmullen:


manthey:
How did this manifest itself?

zachmullen:
If you open the console on the swagger page and make a request using swagger, you'd see a console message saying "server doesn't consume application/json, try []"

manthey:
LGTM,

PR #792, opened 2015-04-06
Fixed some swagger issues with proxies.

manthey:
Update proxy documentation.

Don't use the apiRoot within the swagger base.

Added a test for swagger docs.

Loading the actual swagger page doesn't play well with the blanket js coverage tests, so a testing-no-coverage.min.js library is built.

Cleaned up some style issues in various files.

The EXTRADEBUG=cherrypy option to show all cherrypy calls had been disabled when the startup log output was changed.

Swagger doesn't work with certain basePath values behind a proxy.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -150,7 +150,7 @@ module.exports = function (grunt) {
                         callback();
                     }
                 }
-            },
+            }
         },
 
         compress: {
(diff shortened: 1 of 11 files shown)


manthey:
@zachmullen I've tested this behind an nginx proxy.  I would expect it to work behind the Apache proxy, too, but I haven't tested that.

zachmullen:
LGTM!

PR #793, opened 2015-04-09
HDFS assetstore was breaking assetstore page

zachmullen:
when server was not reachable.

### DIFF
--- a/plugins/hdfs_assetstore/server/assetstore.py
+++ b/plugins/hdfs_assetstore/server/assetstore.py
@@ -101,11 +101,17 @@ class HdfsAssetstoreAdapter(AbstractAssetstoreAdapter):
         return doc
 
     def capacityInfo(self):
-        info = self.client.df()
-        return {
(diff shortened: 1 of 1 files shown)


plugins/hdfs_assetstore/server/assetstore.py
```diff
@@ -101,11 +101,17 @@ def validateInfo(doc):
...
-            'free': info['capacity'] - info['used'],
-            'total': info['capacity']
-        }
+        try:
+            info = self.client.df()
+            return {
+                'free': info['capacity'] - info['used'],
+                'total': info['capacity']
+            }
+        except:
```
manthey: Can you change this to a more specific exception?  Even just `except Exception:` would be better (so it doesn't catch base exceptions).

manthey: I've added a PR with a flake8 test for this condition.

zachmullen:
ping @manthey

manthey:
LGTM.

PR #794, opened 2015-04-09
Don't use except: without specifying the exception.

manthey:
By itself, `except:` will catch all exceptions, even, for instance, SIGINT (Control-C).  This isn't usually desired, and if it is, a comment should be added to explain why.  Rather, the default exception that should be used when you want to catch nearly all exceptions is `except Exception:`.  This lets `BaseException`s through (which includes SIGINT) while stopping all subclassed `Exception`.

### DIFF
--- a/docs/developer-cookbook.rst
+++ b/docs/developer-cookbook.rst
@@ -301,7 +301,7 @@ behaviors, use the following examples:
 
     try:
         ...
-    except:
+    except Exception:
         # Will log the most recent exception, including a traceback, request URL,
         # and remote IP address. Should only be called from within an exception handler.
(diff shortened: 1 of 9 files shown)


jbeezley:
:+1: blocking SIGINT is evil

zachmullen:
This sounds like a good idea. Besides SIGINT, what else falls under `BaseException` and not `Exception`?

manthey:
Besides SIGINT, `SystemExit` and `GeneratorExit`.

See https://docs.python.org/2/library/exceptions.html.

zachmullen:
Ah, thanks. LGTM

PR #795, opened 2015-04-14
Only do what is required for CORS

zachmullen:
@mgrauer PTAL at the initial changes

### DIFF
--- a/clients/web/src/templates/body/systemConfiguration.jade
+++ b/clients/web/src/templates/body/systemConfiguration.jade
@@ -51,7 +51,7 @@ form.g-settings-form(role="form")
       .panel-heading
         .panel-title
           a(data-toggle="collapse", data-parent="#g-configuration-accordion",
-            href="#g-advanced-settings-tab") 
+            href="#g-advanced-settings-tab")
             b Advanced Settings
       #g-advanced-settings-tab.panel-collapse.collapse
         .panel-body
@@ -67,7 +67,9 @@ form.g-settings-form(role="form")
               placeholder="Default: #{defaults['core.upload_minimum_chunk_size'] || ''}",
               title="For large files, the minimum size of all but the last chunk.")
           | CORS (Cross-origin resource sharing) allows access from other
-          | websites.
+          | websites. These settings are described in detail
+          a(href="http://girder.readthedocs.org/en/latest/security.html#cors-cross-origin-resource-sharing"
+            target="_blank")  here.
           .form-group
             label(for="g-core-cors-allow-origin") CORS Allowed Origins
             input#g-core-cors-allow-origin.form-control.input-sm(
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -17,48 +17,29 @@ endpoint code is executed.
 CORS (Cross-Origin Resource Sharing)
 ------------------------------------
 
-When a request is sent from a web browser that could modify the data in Girder,
-the web browser sends an ``Origin`` header.  If this is not the same origin as
-where a user's sessions was initiated, it is a Cross-Origin request, and is
-restricted based on the Girder CORS settings.
+In an out-of-the-box Girder deployment, `CORS <http://en.wikipedia.org/wiki/Cross-origin_resource_sharing>`__
+is disabled for API calls. If you want your server to support API calls that are cross-origin
+requests from web browsers, you'll need to modify some configuration settings.
 
-By default, all cross-origin requests that could modify data are refused.
-Different origins may be allowed via the System Configuration.  For best
-security it is highly recommended that only a specific list of origins be
-allowed, and not all origins using the ``*`` token.  When responding to a valid
-Cross-Origin request, Girder only responds that the specific origin is allowed,
-and does not reveal what other origins can be accessed.
+As an administrator, go to the **Admin console**, then to **Server configuration**.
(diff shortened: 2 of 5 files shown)


mgrauer:
I tested this branch by resetting the girder `CORS Allowed Headers` to the default set, which is `Accept-Encoding, Authorization, Content-Disposition, Content-Type, Cookie, Girder-Token` and a `CORS Allowed Origins` setting of `*` unless otherwise specified.

tl; dr

It worked in all 4 cases that I tested, and I did not need to add additional accepted CORS headers beyond the default set to girder.



### Simple CORS request

#### Simple CORS request success case

I had success with a simple CORS request, the server correctly responded with

    Access-Control-Allow-Credentials:true
    Access-Control-Allow-Origin:*

Before this branch on `master`, I would have needed to specifically add `Pragma, Cache-Control` to the accepted header list.  On this branch I did not need to add them, which is correct as these are part of the  simple CORS request header list.  

#### Simple CORS request failure case

I removed the `CORS Allowed Origins` setting of `*` to test this case.  The server correctly responded without any response data, with a `200`, without any CORS response headers, and with a console error message of

> XMLHttpRequest cannot load http://localhost:8083/api/v1/file/552c1e050640fd09ee545395/download. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:8882' is therefore not allowed access.

*Slightly strange* The only thing that seemed a little off here was that `Content-Length` response was reported as the correct size of the content if it would have been sent, though again, the content was not sent in the response.

### Not-So-Simple CORS request

#### Not-So-Simple CORS request success case

The OPTIONS request succeeded, with the CORS response headers of 

    Access-Control-Allow-Credentials:true
    Access-Control-Allow-Headers:Accept-Encoding, Authorization, Content-Disposition, Content-Type, Cookie, Girder-Token
    Access-Control-Allow-Methods:GET, POST, PUT, HEAD, DELETE
    Access-Control-Allow-Origin:*

The GET was issued and responded correctly with the CORS response headers of

    Access-Control-Allow-Credentials:true
    Access-Control-Allow-Origin:*

Before this branch on `master`, I would have needed to specifically add `Pragma, Cache-Control` to the accepted header list.  On this branch I did not need to add them, which is correct as these are part of the  simple CORS request header list.

#### Not-So-Simple CORS request failure case

I removed the `CORS Allowed Origins` setting of `*` to test this case.  The server correctly responded to the OPTIONS request without any response data, with a `200`, with the following CORS response headers

    Access-Control-Allow-Headers:Accept-Encoding, Authorization, Content-Disposition, Content-Type, Cookie, Girder-Token
    Access-Control-Allow-Methods:GET, POST, PUT, HEAD, DELETE

and with a console error message of

> XMLHttpRequest cannot load http://localhost:8083/api/v1/file/552c1e050640fd09ee545395/download. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:8882' is therefore not allowed access.

and **did not** issue a GET as is correct.

Again with the strange `Content-Length` set to the correct but hypothetical value.

zachmullen:
> Again with the strange Content-Length set to the correct but hypothetical value.

Were you able to access that programmatically in javascript after making the XHR? I'm wondering if that's just something you can see in your browser network console or if it's legitimately visible to the runtime environment.

mgrauer:
Only in the network console.  And I revise the Content-Length of the OPTIONS to 0.  It was only in the case of the Simple request.

zachmullen:
@manthey This is ready for review; I removed most of our server-side logic for trying to enforce CORS and instead kept it to just what is needed to make the browser enforce it for us. This is pretty much a bare-minimum approach, but is also correct and will make it so we don't run into the false negative cases we've been seeing lately when we try to duplicate the browser's logic on the server.

clients/web/src/templates/body/systemConfiguration.jade
```diff
@@ -67,7 +67,9 @@ form.g-settings-form(role="form")
               placeholder="Default: #{defaults['core.upload_minimum_chunk_size'] || ''}",
               title="For large files, the minimum size of all but the last chunk.")
           | CORS (Cross-origin resource sharing) allows access from other
-          | websites.
+          | websites. These settings are described in detail
+          a(href="http://girder.readthedocs.org/en/latest/security.html#cors-cross-origin-resource-sharing"
```
manthey: Can we make this a relative link?  That way if we have multiple versions of the docs on readthedocs (or elsewhere), the versions will stay within themselves.

zachmullen: What do you mean by relative link?

manthey: Will `href="security.html#cors-cross-origin-resource-sharing"` link properly, rather than using the full link which includes `http://girder.readthedocs.org`?

zachmullen: Something like that would work if we were serving the docs statically from each instance, but we aren't doing that at this time.

manthey: My real desire is that this would go to `girder.readthedocs.org/en/(version)` where `(version)` is latest if this is a build that is after the latest version, and is otherwise the appropriate `v1.2.4` or other string.  I don't see how to do that conveniently (and that would be a whole different issue).

LGTM.

PR #797, opened 2015-04-16
Document nginx proxy config to support SSE

zachmullen:
I noticed on one of our nginx-proxied deployments that
the EventStream notifications were not working, and
eventually tracked it down to the fact that they were
being eaten by a combination of nginx's default buffering
and timeout fields.

Also includes a couple of inconsequential fixes

### DIFF
--- a/clients/web/src/app.js
+++ b/clients/web/src/app.js
@@ -180,11 +180,10 @@ girder.App = girder.View.extend({
         var route = girder.dialogs.splitRoute(Backbone.history.fragment).base;
         Backbone.history.fragment = null;
         girder.router.navigate(route, {trigger: true});
+        girder.eventStream.close();
 
         if (girder.currentUser) {
             girder.eventStream.open();
-        } else {
-            girder.eventStream.close();
         }
     }
 });
--- a/clients/web/src/app.js
+++ b/clients/web/src/app.js
@@ -182,7 +182,6 @@ girder.App = girder.View.extend({
         girder.router.navigate(route, {trigger: true});
 
         if (girder.currentUser) {
-            girder.eventStream.close();
             girder.eventStream.open();
         } else {
             girder.eventStream.close();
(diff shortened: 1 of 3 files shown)


clients/web/src/app.js
```diff
@@ -182,7 +182,6 @@ girder.App = girder.View.extend({
         girder.router.navigate(route, {trigger: true});
 
         if (girder.currentUser) {
-            girder.eventStream.close();
```
manthey: If I recall correctly, this `close` seemed necessary to release the request in cherrypy in a timely fashion.  We had originally configured cherrypy to have only 10 workers, and, since the event stream could hang on for minutes, it was possible to exhaust the available workers if this wasn't closed.  Are we now just relying on having enough workers?  Was there a problem with closing the stream?

docs/deploy.rst
```diff
@@ -63,6 +63,9 @@ Nginx can be used by adding a block such as:
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto $scheme;
         proxy_pass http://localhost:9000/;
+        # Must turn off response buffering for SSE to work
+        proxy_buffering off;
+        proxy_read_timeout 600s;
```
manthey: Do we need to also specify the `proxy_write_timeout` or `proxy_connect_timeout`?

zachmullen: We shouldn't need `proxy_connect_timeout` since the documentation says that's only for how long it takes to make the initial connection. I don't see `proxy_write_timeout` in the documentation (I am looking at http://nginx.org/en/docs/http/ngx_http_proxy_module.html), is that actually an option?

manthey: It's `proxy_send_timeout`.

clients/web/src/app.js
zachmullen: I will add this back in to be safe, I figured this wouldn't ever be called if the event stream was already open, but that may happen in some corner cases.

manthey: Wouldn't it be invoked when you hit back in the browser and are logged in?

docs/deploy.rst
zachmullen: Ah, yes, I will add that as we probably do need it since no data will be sent over the channel.

clients/web/src/app.js
zachmullen: No, this only gets called when the user logs in or out explicitly.

manthey:
LGTM.

PR #798, opened 2015-04-16
WIP Support local fonts

jeffbaumes:
This is WIP but allows for entirely offline use. Used the following site to download font files. Works for modern browsers only which I think should be ok.

https://google-webfonts-helper.herokuapp.com/fonts/droid-sans?subsets=latin

Not sure how this will work or not work with other full-app plugins like Covalic which include the google font directly. This seems to be nice in that girder plugins would no longer need to do that.


zachmullen:
@jeffbaumes I definitely like the offline option. On the other hand, I also like offloading a good chunk of our static bandwidth to Google for actual deployments -- how about we achieve the best of both worlds and make this a config file option that controls the webroot template?

jeffbaumes:
Sounds like a plan. Not sure how to do that - but sounds like a plan :) In
the near term I need this commit SHA public since I'm using it for a Docker
deployment so please don't force push on this.

zachmullen:
Roger that.

PR #807, opened 2015-04-24
Querying access on object with no ACL raised an exception

zachmullen:
We should simply return an empty ACL in this case.


manthey:
Should we have a test for this?

zachmullen:
I didn't think so, but if someone does think so I can write one.

manthey:
Okay.  LGTM.

PR #810, opened 2015-04-25
Delete pending group invites when users are deleted

brianhelba:
The invalid database references left behind by this bug were causing large parts of the groups API to break. Some manual cleanup may be necessary on instances affected by this issue.


brianhelba:
I'll push the 2nd commit as soon as I verify that the tests properly fail on the 1st.

brianhelba:
Ready for review and merge.

cpatrick:
LGTM!

PR #811, opened 2015-04-25
Recursive permission application

zachmullen:


zachmullen:
This is ready for review.

clients/web/src/views/body/CollectionView.js
```diff
@@ -133,8 +133,11 @@
                 modelType: 'collection',
                 model: this.model,
                 parentView: this
-            }).on('g:saved', function (collection) {
-                // need to do anything?
+            }).on('g:accessListSaved', function (params) {
+                if (params.recurse) {
+                    console.log('x');
```
mgrauer: debug cruft

girder/api/v1/collection.py
```diff
@@ -141,22 +142,37 @@ def getCollectionAccess(self, collection, params):
...
     updateCollectionAccess.description = (
         Description('Set the access control list for a collection.')
         .param('id', 'The ID of the collection.', paramType='path')
         .param('access', 'The access control list as JSON.')
         .param('public', 'Whether the collection should be publicly visible.',
                dataType='boolean')
+        .param('recurse', 'Whether the policies should be applied to all '
+               'subfolders under the collection as well (default=False).',
+               dataType='boolean')
+        .param('progress', 'If recurse is set to True, this controls whether '
```
mgrauer: can you add the default value to the swagger docs for `progress` and `public` ?

girder/api/v1/folder.py
```diff
@@ -179,22 +179,38 @@ def updateFolder(self, folder, params):
...
     updateFolderAccess.description = (
         Description('Update the access control list for a folder.')
         .param('id', 'The ID of the folder.', paramType='path')
         .param('access', 'The JSON-encoded access control list.')
         .param('public', "Whether the folder should be publicly visible.",
                dataType='boolean')
+        .param('recurse', 'Whether the policies should be applied to all '
+               'subfolders under this folder as well (default=False).',
+               dataType='boolean')
+        .param('progress', 'If recurse is set to True, this controls whether '
```
mgrauer: also can you add the default values here to progress and public?

2 additional notes that don't require any action.

1) This method is quite similar to the same one in the collection API.  They *could* be joined up, but that would take more work, and prevent us from having separate behavior in the future for the two separate resources.

2) Wouldn't it be cool to be able to set the default value for a param in the description **xor** the code, and have the one read from the other?

girder/models/collection.py
```diff
@@ -191,18 +192,57 @@ def fileList(self, doc, user=None, path='', includeMetadata=False,
...
+        count += sum(self.model('folder').subtreeCount(
+            folder, includeItems=includeItems, user=user, level=level)
+            for folder in folders)
         return count
+
+    def setAccessList(self, doc, access, save=False, recurse=False, user=None,
+                      progress=noProgress, setPublic=None):
+        """
+        Overrides AccessControlledModel.setAccessList to add a recursive
+        option. When true, this will set the access list on all subfolders to
```
mgrauer: This could be a bit clearer by saying "When recurse is true" or similar.

girder/models/folder.py
```diff
@@ -681,3 +694,32 @@ def copyFolderComponents(self, srcFolder, newFolder, creator, progress,
             progress.update(increment=1, message='Copied folder ' +
                             newFolder['name'])
         return newFolder
+
+    def setAccessList(self, doc, access, save=False, recurse=False, user=None,
+                      progress=noProgress, setPublic=None):
+        """
+        Overrides AccessControlledModel.setAccessList to add a recursive
+        option. When true, this will set the access list on all subfolders to
```
mgrauer: This could be a bit clearer by saying "When recurse is true" or similar.

girder/models/collection.py
```diff
@@ -191,18 +192,57 @@ def fileList(self, doc, user=None, path='', includeMetadata=False,
...
+        which the given user has ADMIN access level. Any subfolders that the
+        given user does not have ADMIN access on will be skipped.
+        """
+        progress.update(increment=1, message='Updating ' + doc['name'])
+        if setPublic is not None:
+            self.setPublic(doc, setPublic, save=False)
+        doc = AccessControlledModel.setAccessList(self, doc, access, save=save)
+
+        if recurse and save:
```
mgrauer: I'm slightly torn about this. 

If you pass in recurse=true and save=false, this block will not be executed, which makes sense since you are not returning the set of changed resources and you are not saving.  The only time I could see this not doing what you expect is if you want to load, modify, and save the same set of changed resources externally to this method, and also want them to have the updated accessList, but that seems a bit contorted.

After consideration, I think it makes sense as it is, but I wanted to list my concerns. :flags:

girder/api/v1/folder.py
zachmullen: 1) Agreed. I think we can leave it for right now.
2) Yes :) Now you've got me thinking about how I would implement that. It definitely could be done by specifying it in the docs and using that to set the value in the function, but would be much much harder the other direction.

girder/models/collection.py
```diff
@@ -191,18 +192,74 @@ def fileList(self, doc, user=None, path='', includeMetadata=False,
...
+                'parentId': doc['_id'],
+                'parentCollection': 'collection'
+            })
+
+            folders = self.filterResultsByPermission(
+                cursor=cursor, user=user, level=AccessType.ADMIN, limit=None)
+
+            for folder in folders:
+                self.model('folder').setAccessList(
+                    folder, access, save=True, recurse=True, user=user,
```
mgrauer: At this point you are saving the subtree objects regardless of the value of the method param `save`.  

This is probably better than setting `save=save` as it prevents some accidental behavior of not saving the subtree objects externally, at the possible cost of performance.  Most likely this would be called with `save=True` anyway (as it is in the only places it is currently called).

:heavy_check_mark:

girder/models/folder.py
```diff
@@ -514,29 +514,42 @@ def parentsToRoot(self, folder, curPath=None, user=None, force=False,
             return self.parentsToRoot(curParentObject, curPath, user=user,
                                       force=force)
 
-    def subtreeCount(self, folder):
+    def subtreeCount(self, folder, includeItems=True, user=None, level=None):
```
mgrauer: I think it would make sense for the [user model subtreeCount method](https://github.com/girder/girder/blob/recursive-permission-setting/girder/models/user.py#L286-L299) to expose these additional parameters and pass them on to folder model subtreeCount call.

girder/models/collection.py
```diff
@@ -191,18 +192,74 @@ def fileList(self, doc, user=None, path='', includeMetadata=False,
...
+        :type includeItems: bool
+        :param user: If filtering by permission, the user to filter against.
+        :param level: If filtering by permission, the required permission level.
+        :type level: AccessLevel
         """
         count = 1
         folders = self.model('folder').find({
             'parentId': doc['_id'],
             'parentCollection': 'collection'
         })
```
mgrauer: add access in fields please to match subtreeCount in folder.

    }, fields=('access',))

girder/models/folder.py
```diff
@@ -514,29 +514,42 @@ def parentsToRoot(self, folder, curPath=None, user=None, force=False,
...
         count = 1
 
-        items = self.model('item').find({
-            'folderId': folder['_id']
-        }, fields=())
-        count += items.count()
-        # subsequent operations take a long time, so free the cursor's resources
-        items.close()
+        if includeItems:
+            items = self.model('item').find({
```
mgrauer: If you wanted to be premature, you could call count instead of find on the items, since we don't filter them by permission.

see http://api.mongodb.org/python/current/api/pymongo/collection.html#pymongo.collection.Collection.count

mgrauer:
There is a slight disconnect between the way the public value is set on a Collection or Folder from the api and the model.

The api forces a boolean value for `public`.  e.g. [for folder](https://github.com/girder/girder/blob/recursive-permission-setting/girder/api/v1/folder.py#L184-L201)

whereas the [folder model allows a value of `None` for `public`](https://github.com/girder/girder/blob/recursive-permission-setting/girder/models/folder.py#L724-L725), and will not update the folder resource in this case. 

Just wanted to make sure this is intentional.

tests/cases/collection_test.py
```diff
@@ -172,3 +180,33 @@ def testCollectionAccess(self):
...
+        self.assertEqual(folder2['public'], True)
+        self.assertEqual(folder1['access'], coll['access'])
+        self.assertEqual(folder1['access'], folder2['access'])
+        self.assertEqual(folder2['access'], {
+            'users': [{
+                'id': self.user['_id'],
+                'level': AccessType.WRITE
+            }],
+            'groups': []
+        })
```
mgrauer: On something as important as setting access control I'd like a bit more testing.

I'd like to see a test case where there is a tree that is public (extending this existing case would work fine) and gets set to private, and then testing that multiple levels in the tree inherit the private access.  I'd also like to see in that same case a deletion of access, this could be as simple as taking the existing test access object and removing the user write access, setting this access on the tree recursively, then checking that this has been removed at multiple levels.

zachmullen:
@mgrauer I think I addressed all your points, can you take another look?

mgrauer:
I should have gathered these up for you in one place, I needed to do it myself to sign off.

 - [ ] remove console.log
 - [ ] vastly improve default values for api endpoint descriptions
 - [ ] when `recurse=True` in docs for folder and collection
 - [ ] did you want to call `count` instead of `find` on items, or leave it as is?
 - [ ] always recurse if `recurse=True` is passed to `setAccessList` in folder and collection
 - [ ] expand interface on user subTreeCount
 - [ ] expand message in dialog and keep text consistent
 - [ ] keep model and rest APIs consistent regarding passing `public` value (allowing for no-op)
 - [ ] expand testing of recursive permission setting, including removal and omitting `public` value

Other than deciding on count vs find, and I'm fine with it as is, LGTM.

Nice work.  :boar:

zachmullen:
> did you want to call count instead of find on items, or leave it as is?

We should add that, however it will involve adding a new method to `model_base.Model` to expose it on the mongo collection, so I'll relegate that to another topic.

PR #813, opened 2015-04-27
Doc python client on setting host, https, and S3 support

mgrauer:
Fixes a few small holes in our python client documentation.


zachmullen:
LGTM

PR #823, opened 2015-05-06
Add leading slash to apiRoot if necessary

cjh1:
@zachmullen Not sure this is the right fix. However, when I use the OAuth plugin Authorized redirect URI is missing a slash before the API root part. So I track it down this config file. Obviously, I could check for a slash in the client and fix it up there, thoughts?

### DIFF
--- a/plugins/oauth/web_client/js/ConfigView.js
+++ b/plugins/oauth/web_client/js/ConfigView.js
@@ -37,11 +37,17 @@ girder.views.oauth_ConfigView = girder.View.extend({
     },
 
     render: function () {
-        var origin = window.location.protocol + '//' + window.location.host;
+        var origin = window.location.protocol + '//' + window.location.host,
+            apiRoot = girder.apiRoot;
+
+        if (apiRoot.substring(0, 1) !== '/') {
+            apiRoot = '/' + apiRoot;
+        }
+
         this.$el.html(girder.templates.oauth_config({
             google: {
                 jsOrigin: origin,
-                redirectUri: origin + girder.apiRoot + '/oauth/google/callback'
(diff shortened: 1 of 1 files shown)


manthey:
I'm against making this a non-relative path.  It makes it harder to set up
proxies.  I specifically changed this to a relative path in PR #788.

On Wed, May 6, 2015 at 2:39 PM, Chris Harris <notifications@github.com>
wrote:

> @zachmullen <https://github.com/zachmullen> Not sure this is the right
> fix. However, when I use the OAuth plugin Authorized redirect URI is
> missing a slash before the API root part. So I track it down this config
> file. Obviously, I could check for a slash in the client and fix it up
> there, thoughts?
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/girder/girder/pull/823
> Commit Summary
>
>    - Add leading slash to apiRoot
>
> File Changes
>
>    - *M* girder/conf/girder.dist.cfg
>    <https://github.com/girder/girder/pull/823/files#diff-0> (2)
>
> Patch Links:
>
>    - https://github.com/girder/girder/pull/823.patch
>    - https://github.com/girder/girder/pull/823.diff
>
> —
> <https://github.com/girder/girder/pull/823>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

zachmullen:
I think we should leave this as-is since that changes the api root from a relative path to an absolute one. Is it simple enough to just fix this in the oauth plugin?

cjh1:
@zachmullen Sure, I fix it in the plugin

zachmullen:
LGTM

PR #824, opened 2015-05-07
Upgrade npm to the latest version.

manthey:
Occasionally npm install would fail with an EEXIST error message.  This was caused by a race condition in npm (see https://github.com/npm/npm/issues/6309).  Upgrading to the latest npm version avoids this problem.

I've occasionally seen this failure in a travis build, but it is fairly rare.


zachmullen:
LGTM

PR #828, opened 2015-05-08
Fix test failure from undefined dict serialization order

zachmullen:
This is the reason we have test failures in master currently


manthey:
LGTM.

Issue #677, opened 2015-02-05
add ability to upload to S3 assetstore to Python client


mgrauer:
This can largely follow the javascript implementation in the web client.

mgrauer:
This exists when using the python client with a Girder server version 1.3.0 or later.

danlamanna:
@mgrauer so is this resolved?

PR #830, opened 2015-05-11
Support proxied uploads to S3 assetstores

zachmullen:
This will support clients that do not implement the direct-to-S3
upload behavior.

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -23,10 +23,10 @@ import cherrypy
 import json
 import re
 import requests
+import six
 import uuid
 
 from .abstract_assetstore_adapter import AbstractAssetstoreAdapter
-from .model_importer import ModelImporter
 from girder.models.model_base import ValidationException
 from girder import logger, events
 
@@ -105,7 +105,7 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
         self.assetstore = assetstore
 
     def _getRequestHeaders(self, upload):
-        headers = {
+        return {
             'Content-Disposition': 'attachment; filename="{}"'
                                    .format(upload['name']),
             'Content-Type': upload.get('mimeType', ''),
@@ -114,7 +114,6 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
             'x-amz-meta-uploader-id': str(upload['userId']),
             'x-amz-meta-uploader-ip': str(cherrypy.request.remote.ip)
         }
-        return headers
 
     def initUpload(self, upload):
         """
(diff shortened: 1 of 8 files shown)


zachmullen:
Fixes #677

zachmullen:
This is now ready for review.

girder/utility/s3_assetstore_adapter.py
```diff
@@ -155,10 +154,25 @@ def initUpload(self, upload):
...
         :param chunk: This should be a JSON string containing the chunk number
-        and S3 upload ID.
+            and S3 upload ID. If a normal chunk file-like object is passed,
+            we will send the data to S3.
+        """
+        if isinstance(chunk, six.string_types):
+            return self._clientUploadChunk(upload, chunk)
+        else:
+            return self._proxiedUploadChunk(upload, chunk)
+
```
mgrauer: Something more explicit than checking the type of the chunk would be ideal.  @zachmullen said something like #837 would help.

girder/utility/s3_assetstore_adapter.py
```diff
@@ -172,12 +186,65 @@ def uploadChunk(self, upload, chunk):
...
+            key = mp.upload_part_from_file(
+                chunk, upload['s3']['partNumber'],
+                headers=self._getRequestHeaders(upload))
+            upload['received'] += key.size
+        else:
+            key = bucket.new_key(upload['s3']['key'])
+            key.set_contents_from_file(chunk,
+                                       headers=self._getRequestHeaders(upload))
+
+            if key.size < upload['size']:
```
mgrauer: how would this condition (being true) arise?

zachmullen: The nature of the proxied uploads is such that we decide at init time whether this is a chunked upload or a single-request upload to S3 based on the CHUNK_LEN property of the adapter. If the upload is less than that, we automatically assume it will be sent in one chunk to girder. If it is not, we let the client know that they should send it all in one chunk.

Given that our chunked upload threshold is 32MB for S3 and our default/recommended chunk size for clients is 64MB, this condition will probably not be reached by regular clients.

zachmullen: For instance, our testing code exercises this error by initiating an upload of a few bytes, and then trying to send it in two chunks. That would work for normal assetstores, but not for S3. In fact, S3 itself has a minimum chunk length of 5MB, which we fake to be much lower when we use moto in our testing.

girder/utility/s3_assetstore_adapter.py
```diff
@@ -185,7 +252,8 @@ def requestOffset(self, upload):
         return {
             'method': 'PUT',
             'url': url,
-            'headers': headers
+            'headers': headers,
+            'offset': upload['received']
```
mgrauer: will `upload['received']` always be 0 here?

zachmullen: Yep. This is to accommodate the case of something failing during the first chunk in a proxied upload condition; "dumb" clients expect an `offset` field in the response object, and if it's not there they will not be able to resume an upload unless they choose 0 as the default in that case.

If you would rather I changed it to a literal 0, I am happy to do so.

mgrauer: Please do, that will read more clearly to me.

mgrauer:
LGTM.  Awesome work.  I tested with both the standard web client and the python client.  The python client now magically has the ability to upload to S3.

:fireworks:

PR #831, opened 2015-05-11
Set Blob type correctly via constructor

### LINKED ISSUE #829: Girder attempting to set read-only type property on Blob objects
According to

https://developer.mozilla.org/en-US/docs/Web/API/Blob

the `type` field is read only, but we are setting it here:

https://github.com/girder/girder/blame/3cd83a3/clients/web/src/models/FileModel.js#L55

I cannot reproduce the error being thrown on my browser/versions of Chrome, Firefox, and Safari but @curtislisle is hitting this in a client app Kitware/tangelohub#145. But even on all my browsers I can see that `type` is unaffected when trying to set it, so it is reproducible in that way everywhere. His browser seems to be using a getter for a read-only property:

http://javascriptplayground.com/blog/2013/12/es5-getters-setters/

I did not even know such things existed in JavaScript.

ronichoudhury:
Closes #829

### DIFF
--- a/clients/web/src/models/FileModel.js
+++ b/clients/web/src/models/FileModel.js
@@ -44,8 +44,6 @@ girder.models.FileModel = girder.Model.extend({
      * @param type The mime type of the file (optional).
      */
     _uploadToContainer: function (Model, model, data, name, type) {
-        var datatype = data.toString();
-
         if (_.isString(model)) {
             model = new Model({
                 _id: model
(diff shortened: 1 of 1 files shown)


mgrauer:
LGTM if Travis is happy.  A file that I created and uploaded via [uploadToItem](https://github.com/girder/girder/blob/master/clients/web/src/models/FileModel.js#L78) now has the correct MIME type with this fix.

screen shot 2015-05-11 at 5 56 19 pm

ronichoudhury:
Should I merge this?  Or @mgrauer / @zachmullen can do it.

zachmullen:
@ronichoudhury you get to push the fancy button

Issue #207, opened 2014-06-12
Consider redirect to root on login/logout


jeffbaumes:
I was often confused when switching between users that the page does not change. For example, I stayed on the admin page when logged out, and stayed on item/folder pages that I did not have access to when logged out or logged in as a user without privileges to that resource. When I'm on a page that doesn't make sense, I get a message in HTML stating this (like on the admin page) or I get a popup on a failed request. The latter seems especially disrupting so that could at least be changed to an HTML message on the page.

The only plus of keeping the URL the same is that there could be a "user switching" pattern which could be interesting that enables looking at the same thing, like an item or group from different roles.

mgrauer:
Closed by #834 .

PR #834, opened 2015-05-12
Fetch no longer ignores default error handler

mgrauer:
Now fetch will call the default error handler.  This will report the error in the error console and send a notification through girder.

In some cases where the user isn't authenticated and attempts to read access controlled resources, the default error handler will open the login dialog.  Sometimes the login dialog won't actually appear, e.g. the [CollectionView will send the user back to the Collections List](https://github.com/girder/girder/blob/master/clients/web/src/views/body/CollectionView.js#L157).

### DIFF
--- a/clients/web/src/init.js
+++ b/clients/web/src/init.js
@@ -168,7 +168,7 @@ _.extend(girder, {
         }).then(function () {
             girder.currentUser = null;
 
-            girder.events.trigger('g:logout');
+            girder.events.trigger('g:login', null);
             girder.events.trigger('g:logout.success');
         }, function (jqxhr) {
             girder.events.trigger('g:logout.error', jqxhr.status, jqxhr);
--- a/clients/web/src/init.js
+++ b/clients/web/src/init.js
@@ -168,7 +168,7 @@ _.extend(girder, {
         }).then(function () {
             girder.currentUser = null;
 
-            girder.events.trigger('g:login', null);
+            girder.events.trigger('g:logout');
             girder.events.trigger('g:logout.success');
         }, function (jqxhr) {
             girder.events.trigger('g:logout.error', jqxhr.status, jqxhr);
--- a/clients/web/test/spec/adminSpec.js
+++ b/clients/web/test/spec/adminSpec.js
@@ -91,6 +91,13 @@ describe('Test the settings page', function () {
             return $('#g-settings-error-message').text() === '';
         }, 'error message to be cleared');
     });
+    it('logout and check for redirect to front page from settings page', function () {
+        girderTest.logout()();
+
+        waitsFor(function () {
+            return $('.g-frontpage-title:visible').length > 0;
+        }, 'front page to display');
+    });
 });
 
 describe('Test the assetstore page', function () {
@@ -233,6 +240,12 @@ describe('Test the assetstore page', function () {
     }
 
     it('Go to assetstore page', function () {
+        girderTest.login('admin', 'Admin', 'Admin', 'adminpassword!')();
+
+        runs(function () {
+            $("a.g-nav-link[g-target='admin']").click();
+        });
+
         runs(function () {
             $("a.g-nav-link[g-target='admin']").click();
         });
@@ -284,10 +297,10 @@ describe('Test the assetstore page', function () {
 
     /* Logout to make sure we don't see the assetstores any more */
     it('logout from admin account', girderTest.logout('logout to no longer view asset stores'));
-    it('check logged out state', function() {
-        runs(function () {
-            expect($('#g-app-body-container').text()).toEqual('Must be logged in as admin to view this page.');
-        });
+    it('logout and check for redirect to front page from assetstore page', function () {
+        waitsFor(function () {
+            return $('.g-frontpage-title:visible').length > 0;
+        }, 'front page to display');
     });
 });
 
@@ -372,7 +385,7 @@ describe('Test the plugins page', function () {
         });
     });
     /* Logout to make sure we don't see the plugins any more */
-    it('log out and check state', function() {
+    it('log out and check for redirect to front page from plugins page', function() {
         waitsFor(function () {
             return $('.g-logout').length > 0;
         }, 'logout link to render');
@@ -390,5 +403,9 @@ describe('Test the plugins page', function () {
             $('#g-dialog-container').click();
         });
         girderTest.waitForLoad();
+
+        waitsFor(function () {
+            return $('.g-frontpage-title:visible').length > 0;
+        }, 'front page to display');
     });
 });
--- a/clients/web/test/spec/adminSpec.js
+++ b/clients/web/test/spec/adminSpec.js
@@ -33,6 +33,10 @@ describe('Create an admin and non-admin user', function () {
         expect($('.g-global-nav-li span').text()).not.toContain('Admin console');
     });
 
+    waitsFor(function () {
+        return $('.g-frontpage-title:visible').length > 0;
+    }, 'front page to display');
+
     it('register a (normal user)',
         girderTest.createUser('johndoe',
                               'john.doe@email.com',
@@ -47,6 +51,11 @@ describe('Create an admin and non-admin user', function () {
 
 describe('Test the settings page', function () {
     it('Logout', girderTest.logout());
+
+    it('Test that anonymous loading settings page prompts login', function () {
+        girderTest.anonymousLoadPage(false, 'settings', true);
+    });
+
     it('Login as admin', girderTest.login('admin', 'Admin', 'Admin', 'adminpassword!'));
     it('Go to settings page', function () {
         runs(function () {
@@ -239,6 +248,10 @@ describe('Test the assetstore page', function () {
         });
     }
 
+    it('Test that anonymous loading assetstore page prompts login', function () {
+        girderTest.anonymousLoadPage(false, 'assetstores', true);
+    });
+
     it('Go to assetstore page', function () {
         girderTest.login('admin', 'Admin', 'Admin', 'adminpassword!')();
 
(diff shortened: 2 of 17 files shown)


mgrauer:
Feels like timing issues are at play, as failures come and go.  I'm going to set this one aside for a bit, so it shouldn't go into the release.

The superset of tests I've seen fail so far is

  * web_client_data_filesystem
  * web_client_data_s3
  * web_client_group
  * web_client_data_gridfsrs
  * web_client_data_gridfs
  * web_client_user_quota.user_quota

clients/web/test/spec/dataSpec.js
```diff
@@ -861,5 +861,13 @@ describe('Test FileModel static upload functions', function () {
         });
     });
 
-    it('logout from test account', girderTest.logout('logout from test account'));
+    it('logout from test account', function () {
```
manthey: I can help run down the timing issues on the web client tests.  Superficially, this change looks like it is the cause of intermittent web_client_data* failures, because there isn't an appropriate wait after the click (and possibly before it).

The web_client_group test failure is unrelated to this PR, as I've seen it occasionally anyway.

cpatrick:
ref #207

mgrauer:
In [this comment](https://github.com/girder/girder/issues/207#issuecomment-46229473) @zachmullen says we should default to send users to the front page after logging out.  I don't know that this would solve the timing issue (probably not), but considering that this behavior was the fix I was putting into the testing code in the vain attempt to get tests to pass, it might be worth thinking about again.

cpatrick:
I second sending users to the front page after logging out.

zachmullen:
Thirded

mgrauer:
@manthey I can do the logout redirect and rebuild this branch, before you look at the tests, or let you do that change.  Let me know what you'd like.

manthey:
Go ahead.  I won't get to it until at least tomorrow anyway.

On Thu, May 14, 2015 at 2:38 PM, mgrauer <notifications@github.com> wrote:

> @manthey <https://github.com/manthey> I can do the logout redirect and
> rebuild this branch, before you look at the tests, or let you do that
> change. Let me know what you'd like.
>
> —
> <https://github.com/girder/girder/pull/834#issuecomment-102130885>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

manthey:
V1.3.1 is released.

On Thu, May 14, 2015 at 2:54 PM, David Manthey <david.manthey@kitware.com>
wrote:

> Go ahead.  I won't get to it until at least tomorrow anyway.
>
> On Thu, May 14, 2015 at 2:38 PM, mgrauer <notifications@github.com> wrote:
>
>> @manthey <https://github.com/manthey> I can do the logout redirect and
>> rebuild this branch, before you look at the tests, or let you do that
>> change. Let me know what you'd like.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/girder/girder/pull/834#issuecomment-102130885>.
>>
>
>
>
> --
> David Manthey
> R&D Engineer
> Kitware Inc.
> (518) 881-4439
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

mgrauer:
At least the same tests are failing for all 4 runs of Travis.  I'll investigate.

mgrauer:
Only 1 of the 4 Travis combos failed, (Python 3.4, push), and that failure was a timeout on a single test.  Slightly disconcerting that it was one of the failing tests from earlier on this branch `web_client_data_filesystem`.  I'll restart Travis for push.

I still want to test that the login dialog appears when anony:mouse: requests access protected resources.

mgrauer:
And now (Python 2.7, push, web_client_admin, failure).  Restarting again.

zachmullen:
@mgrauer The following looks like a legitimate race condition failure:

```
63: . Create an admin and non-admin user -> check redirect to front page after logout from user page
60: F Test the plugins page -> log out and check for redirect to front page from plugins page
60: Created screenshot: phantom-2015-05-16T00.48.33.896Z.png
60: <DartMeasurementFile name="PhantomScreenshot" type="image/png">/home/travis/build/girder/girder/phantom-2015-05-16T00.48.33.896Z.png</DartMeasurementFile>
60: 
60:  Error: timeout: timed out after 15000 msec waiting for a dialog to fully render
```

mgrauer:
The last failure was (Python 3.4, pr, web_client_admin, failure), a timeout waiting on the modal to be hidden.

At this point the features are largely done.  I couldn't get the login to appear on the `/#admin` page due to some seeming race conditions with the modal dialog.  The subpages of the admin console all work, and I didn't want to spend more time on it now.

There are definitely still some lurking race conditions in the tests, that seem to show up on Travis.

@zachmullen @manthey  How do you recommend that we proceed?

zachmullen:
@manthey have time to take a look at the failing test?

manthey:
Will do, though perhaps not until tomorrow.

On Tue, May 19, 2015 at 2:44 PM, Zach Mullen <notifications@github.com>
wrote:

> @manthey <https://github.com/manthey> have time to take a look at the
> failing test?
>
> —
> <https://github.com/girder/girder/pull/834#issuecomment-103627857>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

manthey:
This error occurs when the PUT command to turn off a plugin is on a slow cherrypy thread that ends up being served AFTER the logout command.  I can occasionally get other intermittent failures, which may be similar in cause.

I'll push a fix as soon as I decide which way we should test for completion.

manthey:
@zachmullen @mgrauer I've fixed the adminSpec test.  I noted while testing this that we have a few other brittle web tests.  I'll try to hunt those down an submit improvements on a new PR.

clients/web/test/spec/adminSpec.js
```diff
@@ -370,25 +401,22 @@ describe('Test the plugins page', function () {
...
-    it('log out and check state', function() {
         waitsFor(function () {
-            return $('.g-logout').length > 0;
-        }, 'logout link to render');
-        runs(function () {
-            $('.g-logout').click();
+            var resp = girder.restRequest({
+                path: 'system/plugins',
+                type: 'GET',
+                async: false
```
zachmullen: `async: false` isn't necessary.

manthey:
I thought it defaulted to `async: true` (and without it, it doesn't work on my test box).

mgrauer:
Thanks for the fix @manthey .

This branch still needs QA otherwise.

zachmullen:
@manthey sorry, I meant that we don't have to block because we have nice async wait utilities available. e.g.


    it('Disable a plugin', function () {
        var resp = null;
        
        runs(function () {
            $('.g-plugin-list-item:contains(Metadata extractor) .bootstrap-switch-label').click();
        });
        runs(function () {
            expect($('.g-plugin-list-item input[type=checkbox]:checked').length).toBe(0);
            girder.restRequest({
                path: 'system/plugins'
            }).done(function (resp) {
                resp = resp;
            });
        });
        waitsFor(function () {
            return (resp && resp.responseJSON && resp.responseJSON.enabled &&
                resp.responseJSON.enabled.length === 0);
        });
    });

Haven't tested, was just suggesting something along those lines

manthey:
The `async: false` method seemed cleaner to me (as much as it is anathema in ordinate javascript code), because I want to call the GET system/plugins during the wait process (not just once after the expects).  If it returns that there are still plugins enabled, I'd have to call it again.  The `waitsFor` code would end up like:

    waitsFor(function () {
        if (!resp) {
            return false;
        }
        if (resp.responseJSON && resp.responseJSON.enabled && 
                resp.responseJSON.enabled.length === 0) {
            return true;
        }
        resp = null;
        girder.restRequest({
            path: 'system/plugins'
        }).done(function (resp) {
            resp = resp;
        });
    }

I can switch to this, but it seemed clumsier.

zachmullen:
Ah, didn't realize it was actually important for that to be in the wait condition. That actually seems strange to me; are we not appropriately waiting on some previous request to complete? Or is there actually something asynchronous happening on the server side?

manthey:
Cherrypy supports multiple connections.  The connection to PUT system/plugins was sent, and the javascript shows the user that the action has occurred (we don't wait for the PUT to return).  The existing `expects` just checked that the UI was updated.  We then logged out.  If the logout occurred (on another cherrypy connection) *before* the PUT was processed, the PUT would fail because the token was now cancelled (and therefore didn't have permission to do the PUT).  The error message from that caused the test to fail.

Either we'd have to modify the code to store the response from the PUT somewhere, so we could test that, OR we need a mechanism to test if there are any outstanding requests (other than stream related requests), OR we can poll the system to see if the effects of our PUT occurred.  I chose the latter of these as this simplest solution for now.

For testing simplicity, it would be nice if there was a way to ask cherrpy how many outstanding requests it was processing, but since that should *not* count queries about outstanding requests and stream requests, and *should* count requests that are queued but not started processing, it isn't obvious to me how to implement it.

zachmullen:
Ah, I see. We can leave the async: false.

clients/web/src/init.js
```diff
@@ -168,7 +168,7 @@ _.extend(girder, {
         }).then(function () {
             girder.currentUser = null;
 
-            girder.events.trigger('g:login', null);
+            girder.events.trigger('g:logout');
```
zachmullen: @mgrauer this change would break backward compatibility; the `g:login` event is triggered on both login and logout (it's poorly named). That's by design so that pages that need to update when the user changes only have to bind to one event rather than two.

mgrauer:
@zachmullen I've put the login event back in.

zachmullen:
This LGTM, thanks @mgrauer

mgrauer:
After the merge to master, Travis failed on `web_client_group`, but it ran correctly upon a build restart.  Looks like there are some timing issues still lurking.

manthey:
Yes.  I noted that we have other brittle tests (unrelated to these
changes), and someone (me) should hunt them down and fix them on a separate
PR.

On Thu, May 28, 2015 at 2:10 PM, mgrauer <notifications@github.com> wrote:

> After the merge to master, Travis failed on web_client_group, but it ran
> correctly upon a build restart. Looks like there are some timing issues
> still lurking.
>
> —
> <https://github.com/girder/girder/pull/834#issuecomment-106541797>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

PR #850, opened 2015-05-18
S3 import

zachmullen:


zachmullen:
@aashish24 the S3 import functionality now works on this topic branch. Anybody available for a QA on this?

aashish24:
Cool! I can test it out today using an existing bucket. I will report back.

aashish24:
Sorry for the delay. I am back to testing this branch now.

aashish24:
As I mentioned to @zachmullen, currently we assume that we have write access to s3 bucket which in my use-case I don't. Having a option would be nice so that Girder won't test the write access and fail.

zachmullen:
@aashish24 I just made it so you can set a read-only property on S3 assetstores. You will still need to supply credentials (simply due to a limitation of boto), but those credentials do not need to have write access to the bucket.

aashish24:
Got further along but when importing got this error: 

500 Internal Server Error {"message": "AttributeError: AttributeError(\"'NoneType' object has no attribute 'list'\",)", "trace": [["girder/api/rest.py", 279, "endpointDecorator", "val = fun(self, args, kwargs)"], ["girder/api/rest.py", 773, "POST", "return self.handleRoute(method, path, params)"], ["girder/api/rest.py", 561, "handleRoute", "val = handler(**kwargs)"], ["girder/api/access.py", 34, "accessDecorator", "return fun(*args, **kwargs)"], ["girder/api/rest.py", 227, "wrapped", "return fun(*args, **kwargs)"], ["girder/api/v1/assetstore.py", 155, "importData", "progress=ctx, user=user)"], ["girder/models/assetstore.py", 177, "importData", "progress=progress, user=user)"], ["girder/utility/s3_assetstore_adapter.py", 339, "importData", "for obj in bucket.list(importPath, '/'):"]], "type": "internal"}

aashish24:
This happened when I actually hit import, looks like authentication and everything else before that was good.

zachmullen:
Interesting, I will take a look, thanks.

zachmullen:
@aashish24 You uncovered a bug -- our validation for read only S3 assetstores was in fact not doing the verification that it could connect to the bucket. The latest commit fixes that. It seems with your credentials that you do not have read access to the bucket in question.

zachmullen:
@aashish24 That last commit now allows anonymous S3 access. When creating/editing the assetstore, simply leave the access key and secret key fields empty, and check the read-only box, and it should work. I was able to import data from the nasanex bucket this way.

aashish24:
awesome@ .. I will try again..

aashish24:
Latest changes worked for me!! 

LGTM :+1:

PR #851, opened 2015-05-20
Fixes #809. Safer checkbox menu behavior

### LINKED ISSUE #809: make it harder to delete all your data
This might be a UI issue, or an issue with confirm vs redo, or maybe we should have a dry-run of notifications on delete in the progress area, I'm open to ideas.

I was in a hierarchy view, and selected a folder, then instead of clicking delete directly above that checkbox for the folder I had selected, I clicked the delete action on the right and clicked to confirm without even reading the confirmation, and then this deleted the parent folder.

zachmullen:
When a user checks resources in the hierarchy widget, we now disable
the folder actions widget to make it hard to accidentally apply
actions to the parent folder rather than the checked resources.

### DIFF
--- a/clients/web/src/views/widgets/HierarchyWidget.js
+++ b/clients/web/src/views/widgets/HierarchyWidget.js
@@ -470,6 +470,14 @@ girder.views.HierarchyWidget = girder.View.extend({
                 minItemLevel = Math.min(minItemLevel, this.parentModel.getAccessLevel());
             }
         }
+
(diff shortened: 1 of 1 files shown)


zachmullen:
@mgrauer PTAL

manthey:
That's a nice solution to the problem.

mgrauer:
nice.  LGTM.  :fountain:

PR #858, opened 2015-06-09
Changing default auth header to Girder-Authentication

jeffbaumes:
The Authentication header conflicts with htaccess authentication in Apache, so changing the default to Girder-Authentication with a fallback to Authentication.


zachmullen:
LGTM, nice fix.

PR #859, opened 2015-06-10
Fix CSS precedence issue that had broken the low-width layout

zachmullen:


cpatrick:
I'm not sure what this is fixing. I'm still having some issues (e.g. name overlap) in low-width settings.

zachmullen:
Without this change, when going to the low width mode the main body view container still had the same margin-left rather than filling the whole width. Can you post a screenshot of the issue you are seeing?

cpatrick:
image

zachmullen:
Hm, you are still seeing the left margin bug that this was meant to fix... you re-ran grunt and got a non-cached version of the css?

cpatrick:
I did those things.

zachmullen:
@cpatrick I'm hoping that last commit fixes both issues you were seeing.

cpatrick:
This is much better on desktop, so lgtm.

We should explore how this looks on mobile, however, it's almost certainly not what we want (see below from Chrome on a Galaxy S6):

screenshot_2015-06-11-10-38-03

zachmullen:
I have also seen really odd behavior on mobile -- the devices tend to lie about their screen sizes, which makes it quite hard to do media queries based on width. Also they tend to resize fonts in ways that are completely puzzling and inconsistent.

PR #860, opened 2015-06-10
Add top level view in web client for job details

zachmullen:


zachmullen:
@cjh1 I created this feature to make it easier to look at the output of romanesco jobs during development. With this you can run romanesco jobs via girder using the test endpoint in https://github.com/zachmullen/sparktest and then view the job output much more easily by navigating to localhost:8080/#job/\<job_id\> in the web client.

cjh1:
@zachmullen Nice!

cjh1:
localhost:8080/#job/<job_invalid_id> produces error and redirect to collections route, I am guess that is the desired behavior?

LGTM

zachmullen:
It's not necessarily desired, but it is expected for the moment until we have a better fallback view. That should eventually be the job list view, but one does not exist yet. Thanks for the review.

PR #862, opened 2015-06-16
Avoid duplicate route documentation

zachmullen:
If plugins extended existing rest resource classes to add new
routes, they would end up duplicating all of the existing routes
in the docs. This prevents that from happening, and therefore
allows normal class extension for rest resources.


cpatrick:
LGTM.

PR #863, opened 2015-06-16
Add a convenience method for extracing the root API URL

zachmullen:


zachmullen:
This is a bit hackish, but it should work for us. We can always change the implementation later if we come up with something that feels better.

girder/api/rest.py
```diff
@@ -38,6 +38,23 @@
...
+    The returned path does *not* end in a forward slash.
+
+    :param url: URL from which to extract the base URL. If not specified, uses
+        `cherrypy.url()`
+    """
+    url = url or cherrypy.url()
+    idx = url.find('/api/v1')
+
+    if idx < 0:
+        raise Exception('Could not determine API root in %s.' % url)
```
cpatrick: This should be a subclass of Exception.

zachmullen:
@cpatrick

cpatrick:
LGTM :shipit:

PR #866, opened 2015-06-17
Update install docs to clarify install steps for developers vs. users

brianhelba:
The Grunt CLI does not directly depend on the Grunt task runner.

See http://gruntjs.com/getting-started#installing-the-cli

### DIFF
--- a/docs/deploy.rst
+++ b/docs/deploy.rst
@@ -104,10 +104,9 @@ example, we have the following:
        tools.proxy.local: ""
 
 After modifying the configuration, always remember to rebuild Girder by
-changing to the main Girder directory and issuing the following command: ::
+changing to the Girder directory and issuing the following command: ::
 
-    $ grunt init && grunt
-
+    $ npm install
 
 Docker Container
 ----------------
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -10,10 +10,31 @@ the tools used to develop Girder.
 Configuring Your Development Environment
 ----------------------------------------
 
-In order to develop Girder, you can refer to the :doc:`prerequisites` and
-:doc:`installation` sections to setup a local development environment. Once
-Girder is started via ``python -m girder``, the server will reload itself
-whenever a Python file is modified.
+In order to develop Girder, you should first refer to the :doc:`prerequisites`
+and :doc:`installation` sections to setup a basic local environment.
+
+Next, you should install the `Grunt <http://gruntjs.com>`_ build tool globally,
+to allow it to be run directly from the command line: ::
+
+    npm install -g grunt
+    npm install -g grunt-cli
+
+Finally, you should also install the Python development dependencies, to
+provide helpful development tools and to allow the test suite to run: ::
+
+    pip install -r requirements-dev.txt
+
+.. note:: One of the development requirements, `httpretty`, can fail to install
+   if under certain system locales (see the `error report
+   <https://github.com/gabrielfalcao/HTTPretty/issues/108>`_).  Changing to any
+   UTF8 locale works around this problem.  For instance, on Ubuntu, you can
+   change your system locale using the commands: ::
+
+        locale-gen en_US.UTF-8
+        export LANG=en_US.utf8
+
+During development, once Girder is started via ``python -m girder``, the server
+will reload itself whenever a Python file is modified.
 
 To get the same auto-building behavior for JavaScript, we use ``grunt-watch``.
 Thus, running ``grunt watch`` in the root of the repository will watch for
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -47,20 +47,6 @@ dependencies: ::
 
     pip install -r requirements.txt
 
-.. note:: If you intend to develop Girder or want to run the test suite, you should also
-   install the development dependencies: ::
-
-        pip install -r requirements-dev.txt
-
-   One of the development requirements, `httpretty`, can fail to install if
-   under certain system locales (see the `error report
-   <https://github.com/gabrielfalcao/HTTPretty/issues/108>`_).  Changing to any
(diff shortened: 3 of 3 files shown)


zachmullen:
@brianhelba we install the grunt worker locally via package.json, is this actually necessary?

jbeezley:
I think when you install locally then you have to run grunt with the local path `./node_modules/.bin/grunt`.

zachmullen:
The CLI is installed globally, but the linked docs say that the CLI uses the grunt installed next to the Gruntfile being executed.

zachmullen:
```
[zach girder]$ rm -rf node_modules/grunt
[zach girder]$ grunt
grunt-cli: The grunt command line interface. (v0.1.13)

Fatal error: Unable to find local grunt.

If you're seeing this message, either a Gruntfile wasn't found or grunt
hasn't been installed locally to your project. For more information about
installing and configuring grunt, please see the Getting Started guide:

http://gruntjs.com/getting-started
```

jbeezley:
I'm not sure what the issue is here, but as I recall, you **do** need to install `grunt` locally either way.  You **don't** need `grunt-cli` installed locally though.

brianhelba:
@zachmullen You're right, I've looked closer in the "package.json", and since it lists Grunt / Grunt CLI as a dependency and as a post-install step, we can just remove all direct references to Grunt as an explicit package to install and as an explicit build step to run.

That is, a user just needs to run "npm install" from inside the root "girder/" directory. All other Grunt-related steps should happen automatically at that point.

Not only is this simpler for users (removing 2 installation steps, and making the build always be just 1 step), but it makes Grunt / Grunt CLI both get installed locally, which is the best practice http://blog.nodejs.org/2011/03/23/npm-1-0-global-vs-local-installation .

zachmullen:
Agreed, though I'm hesitant to remove the instruction about installing the CLI globally, since elsewhere in the docs we frequently assume "grunt" is on the PATH, and we have existing users who also are used to that being the case. What do others think? @girder/developers @girder/owners

jbeezley:
:+1: To installing `grunt-cli` globally.

brianhelba:
@zachmullen @jbeezley So, I've updated my PR to remove this entirely, because I think it's probably the "right" solution. I'm fine to not merge this yet, and continue discussion. We can go in another direction if that's the consensus.

cpatrick:
Part of me is okay with continuing the grunt-cli global install. The other part of me thinks that we should add additional npm targets for the things that need the grunt cli globally later.

brianhelba:
So here's an interesting compromise, the "npm install" command can be run with "-g", to install required packages globally (though that would include everything, not just grunt). https://docs.npmjs.com/cli/install

brianhelba:
Additionally, note that http://blog.nodejs.org/2011/03/23/npm-1-0-global-vs-local-installation reccomends that we install Grunt twice, once globally for developers, and once locally to run the app with.

We could update the docs so that the standard install procedure does not mention Grunt (as my PR does). Then add a "Development Setup" section, instructing users to install Grunt globally, install "requirements-dev.txt" with pip, install CMake, etc.

jbeezley:
@cpatrick npm targets?

@brianhelba `npm install -g grunt-cli` is okay, but `npm install -g` for all dependencies is a bad idea.  I have seriously messed up my global package repository doing that before.

zachmullen:
> npm install -g grunt-cli is okay, but npm install -g for all dependencies is a bad idea. I have seriously messed up my global package repository doing that before.

Agreed. Then we are potentially clashing with any other projects using npm packages on the system, and I don't want to be a bad citizen by polluting global scope.

> We could update the docs so that the standard install procedure does not mention Grunt (as my PR does). Then add a "Development Setup" section, instructing users to install Grunt globally, install "requirements-dev.txt" with pip, install CMake, etc.

I'd be OK with that if others are.

jbeezley:
Sounds good to me.

brianhelba:
@jbeezley I now agree, "npm install -g" is a bad idea, as I just tested it, and it does some wacky things.

For everyone, I'm now more convinced that all references to Grunt and installing Grunt globally should be moved to the development documentation page.

I see the biggest advantage here in simplifying the installation / build process for new users, particularly those who are naive to a JS-based build environment. I can testify first-hand that when I first installed Girder, I was not familiar with the details of NPM or Grunt, and I was a little overwhelmed at how many new technologies there were. The more we can abstract things away, the better.

I'm updating my PR now.

cpatrick:
@jbeezley I assumed that [scripts](https://docs.npmjs.com/files/package.json#scripts) was more versatile than it actually is...

brianhelba:
@zachmullen @jbeezley @cpatrick 
I've rebased this and added another commit to address the issues that have been discussed so far.

This is still WIP, as I want to add a bit more cleanup:
* clarify the prerequisites for building documentation
* add install instructions for CMake; none of the global package install instructions in "prerequisites.rst" include it; that said, some slower distros may not be able to use the package manager to get CMake, as the packaged version might be too old
* note that MongoDB does not *have* to be installed locally; users can use connect to a MongoDB server running somewhere else
* if they are any different, separate the system prerequisites for running the server vs. development

zachmullen:
:+1:

jbeezley:
Looking good.

jcfr:
:+1:

zachmullen:
@brianhelba is this still WIP?

zachmullen:
@brianhelba I'd like to merge this since it contains good changes, but it's still labeled as WIP. Can we get this merged and defer your other proposed changes to a future branch?

brianhelba:
@zachmullen Yeah, lets do that.

I'd hoped to have a little time to update the rest of the items, but I've been so busy.

brianhelba:
Rebased. I'll merge when the tests come back.

PR #867, opened 2015-06-17
[WIP] ACL resolver for non ACL models

danlamanna:
Very much a WIP, trying to get feedback on the general approach.

- Still fails ```server_thumbnails.thumbnail``` test which causes issues with the web client, etc.
- Commenting far from complete, multiple pep8 errors still exist


plugins/provenance/server/resource.py
```diff
@@ -329,7 +329,9 @@ def fileSaveHandler(self, event):
                                        level=AccessType.READ, force=True)
         if not item:
             return
-        prevFile = self.model('file').load(curFile['_id'])
+        # should all of these sort of events be forced, since a save has to be
+        # granted first?
+        prevFile = self.model('file').load(curFile['_id'], force=True)
```
zachmullen: To answer the question in the comment above this line: we can use `force=True` anytime we want to load the resource without performing an access check for a given user, so really it depends on what we're doing. Your usage here is correct. In some old code a few lines above this is a misleading usage (which should be changed):

```
item = self.model('item').load(id=curFile['itemId'], user=user,
                               level=AccessType.READ, force=True)
```

We should only ever pass a user/level OR force=True, but not both, since force=True simply circumvents access checking altogether.

girder/utility/acl_mixin.py
```diff
@@ -0,0 +1,83 @@
...
+
+    resource_coll corresponds to the resource collection that needs to be used
+    for resolution, for example an item would resolve its resource_coll as
+    folder (if it weren't already access controlled).
+
+    resource_parent corresponds to the field in which the parent resource
+    belongs, so for an item it would be the folderId.
+    """
+    resource_coll = None
+    resource_parent = None
```
zachmullen: For consistency, let's use camelcase for these. I know one of the style PEPs discourages that, but all of our other variable names are that way so we are kind of stuck now :)

plugins/provenance/server/resource.py
danlamanna: Sure.

In regards to the code a few lines above, shouldn't ```Item``` model loads that have ```user``` AND ```force``` be retained due to this call:
https://github.com/girder/girder/blob/acl-resolver-for-non-acl-models/girder/models/item.py#L130
which eventually trickles up to
https://github.com/girder/girder/blob/acl-resolver-for-non-acl-models/girder/models/model_base.py#L486

In which case only ```level``` is extraneous if ```force=True```.

zachmullen: Hm, you're right. Although, it's counterintuitive and inconsistent -- I think a good fix for that would be to change `parentsToRoot` such that it doesn't call `filter` when `force=True`.

girder/models/item.py
```diff
@@ -120,12 +122,8 @@ def load(self, id, level=AccessType.ADMIN, user=None, objectId=True,
...
         :type force: bool
         """
-        doc = Model.load(self, id=id, objectId=objectId, fields=fields,
-                         exc=exc)
-
-        if not force and doc is not None:
-            self.model('folder').load(doc['folderId'], level, user, objectId,
-                                      force, fields)
+        doc = super(Item, self).load(id, level, user, objectId, force, fields,
+                                     exc)
```
zachmullen: I personally would prefer the explicitness of using `AccessControlMixin.load(self, ...)` instead of `super`, just to make it easier for devs looking at this for the first time to know where to check for that implementation. I could be convinced otherwise if you think there is a reason `super` is better, though.

danlamanna: I'm not very familiar with where/how Girder is currently being used, but my line of thinking was that this would allow someone to create any number of layers of mixins and ```load``` would "just work"™ assuming the inheritance was declared correctly. That might be a far cry from it's actual usage however.

zachmullen: That is fine, I guess this will support people mucking with the __mro__ if it ever comes to that :)

jbeezley: I agree in general with using `super` for this kind of thing.  It is recommended practice even though it rarely makes a difference.  For the seven people in the world that really understand python's mro semantics, this will do what they expect.

zachmullen:
@danlamanna I discovered why the thumbnail tests are failing. Thumbnails are a special case of File wherein the `itemId` is set to `None` -- thumbnail files correspond to a file, but they are not a proper child item like all the rest. Instead, they are attached to the item via the fields `attachedToType` and `attachedToId`, which can be seen in the old access checking helper method that you removed. I would suggest, to accommodate this in the most general way, that in the ACL mixin's `load` we check if the `self.resourceParent` key is nonexistent in the document, or set to `None`, and if so we use `attachedToType` and `attachedToId` to load the parent instead.

zachmullen:
Additionally, we can remove [these lines](https://github.com/girder/girder/blob/master/plugins/thumbnails/server/rest.py#L37-L38) and we must add `force=True` to [this line](https://github.com/girder/girder/blob/master/plugins/thumbnails/server/worker.py#L52).

girder/utility/acl_mixin.py
```diff
@@ -0,0 +1,117 @@
...
+            if self.resourceParent in doc and doc[self.resourceParent]:
+                loadType = self.resourceColl
+                loadId = doc[self.resourceParent]
+            else:
+                loadType = doc['attachedToType']
+                loadId = doc['attachedToId']
+
+        if not force and doc is not None:
+            self.model(loadType).load(loadId, level, user, objectId, force,
+                                      fields, exc)
```
zachmullen: I think we should use kwargs in the above call, and we should not pass `force`, `fields`, or `objectId` to the parent load call.

danlamanna: I addressed the excessive argument passing in afcc397 - but I'm a bit lost on what type of kwargs usage you're talking about.

zachmullen: I just meant using keyword arguments rather than positional arguments in the call, which you already fixed.

zachmullen:
LGTM. Nice work!

PR #870, opened 2015-06-22
S3 import no silent ignore

zachmullen:
@cjh1 this fixes the other issue you saw.


cjh1:
Small nit, I think we need a little more space after the error message:

screen

The S3 assertstore creation view seems to have the spacing about right if that helps.

zachmullen:
@cjh1 should be fixed now

cjh1:
LGTM

PR #871, opened 2015-06-22
Attempt to fix the web_client_group test race condition

zachmullen:


manthey:
@zachmullen Do you have a reliable way to test this?  I've had good results by running the tests repeatedly in a loop; they don't always fail, but I'll can usually see failures within a dozen or so cycles.

zachmullen:
I could actually never reproduce this locally, but I was seeing it about 8-10% of the time on travis, so I'm just going to restart these travis builds until I am convinced it is fixed.

manthey:
Okay.  I'll set up my loop and see how it fares, too.

zachmullen:
I found the issue, and this does not completely fix it. The race condition is actually due to a double render() call on the group view. I will fix that and we should no longer see this test failing at all.

zachmullen:
That should do it.

manthey:
I'll test it overnight my loop test.

On Mon, Jun 22, 2015 at 4:59 PM, Zach Mullen <notifications@github.com>
wrote:

> That should do it.
>
> —
> <https://github.com/girder/girder/pull/871#issuecomment-114260711>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

zachmullen:
This uncovered more unseemly behavior. Further fixes required.

zachmullen:
The test appears to be fixed now.

manthey:
LGTM.  I ran a loop test and it passed hundreds of times without any issue.

cpatrick:
@zachmullen @manthey Could we do something like the "loop test" on Aiur? There was actually functionality added to CTest recently to allow this grep for `--repeat-until-fail` in the [documentation](http://www.cmake.org/cmake/help/v3.3/manual/ctest.1.html). This would allow us to find these things the night they are introduced (if the probability gods are kind).

manthey:
Short answer: yes.

Long answer:

I don't know if the `--repeat-until-fail` will necessarily do what we want (but it might, if tests can be reordered randomly when run repeatedly).

I've been running the loop-test on an Ubuntu virtualbox that has 4 cores.  IF the virtualbox is starved for resources, tests can fail due to timeouts.  Otherwise, the errors are few.  I wouldn't want to necessarily stop on the first failure, as I have found seeing repeated samples of a failure useful in detecting what went wrong.

I've been using a crude script to do the loop test, like so:

    rm /tmp/count.txt; while true; do rm /tmp/*.dmp 2>/dev/null; date; date >>/tmp/count.txt; if JASMINE_TIMEOUT=15000 ctest --schedule-random -j 5 -V --timeout 300 > /tmp/test.txt; then echo 'okay'; else echo 'failed'; cp /tmp/test.txt /tmp/fail`date "+%H%M%S"`.txt; fi; wc -l /tmp/count.txt; ls /tmp/fail*.txt 2>/dev/null | wc -l; done

A few notes on this:  Sometimes jasmine crashes; we are robust against this in general, but it leaves *.dmp files in my /tmp directory.  I crudely count the number of tests I run by appending the date of each tests start into /tmp/count.txt.  I save the failures in a file that is called /tmp/fail(time).txt, which doesn't work so well when you run the loop for more than a day.

Also, I've been using `--schedule-random`, as it exposes test interaction issues better than always running the tests in the same order.

Issue #875, opened 2015-06-26
Editor for more complex JSON metadata


jeffbaumes:
@danlamanna and I were discussing this issue today. I was sure there was already an issue for this but I could not find it. My proposed plan is twofold:

1. Pretty-print complex JSON metadata in a scrollable field to at least view it better.
2. Next to each property have an "Edit" button that will open up a full JSON editor in a dialog. I say next to each one because even for simple properties the user should be able to transform them into more complex JSON objects if desired.

The editor that @danlamanna is currently looking at is http://jsoneditoronline.org/

Another idea in place of or in addition to this is a single edit button that puts the entire metadata object into a JSON editor.

@danlamanna is currently looking into the first step. Any other thoughts? @brianhelba will this serve your use case?

zachmullen:
:+1: , it's apache licensed, npm installable, and seems well documented. (https://github.com/josdejong/jsoneditor/ for anyone looking for the repo).

As for the UI, I think it might be best not to try to fit both sections into a dialog, it might be cleaner and easier to just open it in a full-width div right below the metadata field being changed.

jeffbaumes:
By "two sections" do you mean the raw JSON and editor versions on the left and right? I was actually just thinking to use the right editor - the one that seems more user friendly - and not even allow the possibility of creating malformed JSON.

zachmullen:
I see what you mean, though I would like to have the ability, at least optionally, to edit the text raw (simply because I'm more used to it). Maybe a single view that can switch between the two modes would be OK.

jeffbaumes:
Yes I see that for some technical folks the raw JSON could be better. Please work with @danlamanna on how you envision the UI layout to work (inline vs. dialog). I just think the JSON editor has its own controls, styles, etc. and it could relieve user mental burden having to just focus on that UI while editing JSON instead of mixing in inline into the Girder UI. Sorta like how inserting/editing diagrams in a Google Doc works.

PR #879, opened 2015-06-29
JSON Metadata Editing re issue #875

danlamanna:
In reference to #875 

As it is now it's just an intermediary step until we implement an actual JSON editor.

A couple of screenshots to see how it looks like instead of having to pull it:
875-json-print-1
875-json-print-2


clients/web/src/templates/widgets/metadatumViewMixin.jade
```diff
@@ -1,6 +1,10 @@
-mixin metadatum(key, value, accessLevel, girder)
+mixin metadatum(key, value, isJson, accessLevel, girder)
```
zachmullen: Can we move this new argument to the end to preserve backward compatibility?

danlamanna:
Updated screenshot:
879-json-editor-1

clients/web/src/stylesheets/widgets/metadataWidget.styl
```diff
@@ -27,6 +27,46 @@
...
+  background-color #fbfbfb
+  border-bottom 1px solid #f0f0f0
+
+.jsoneditor .menu button
+  background #fbfbfb url("../../../lib/jsoneditor/img/jsoneditor-icons.png");
+  border 1px solid #f0f0f0
+
+.jsoneditor .menu button:disabled
+  background #fbfbfb url("../../../lib/jsoneditor/img/jsoneditor-icons.png");
+  border 1px solid #f0f0f0
```
jeffbaumes: The icons do not appear for me. This may be because I am mounting Girder not on the root level. I am able to see the icons when I change this and line 56 to

```
background #fbfbfb url("../lib/jsoneditor/img/jsoneditor-icons.png");
```

jeffbaumes:
My changes after using the JSON widget to not seem to persist. Is this supposed to be working yet?

clients/web/src/stylesheets/widgets/metadataWidget.styl
zachmullen: @jeffbaumes that is the correct relative path, since the style is in `static/built/app.min.css` and the icons are in `static/lib/jsoneditor/img/jsoneditor-icons.png`.

danlamanna:
Of the list of things I still have to work on:
- [ ] JSON edits are persisting **most** of the time for me, with an occasional failure. @jeffbaumes 
- [ ] Many of the existing UI tests are failing
- [ ] Creating new tests (and familiarizing with phantom/existing JS testing)
- [ ] Miscellaneous style fixes (a lot of the hover states are inconsistent, icons not quite right, etc)
- [ ] Probably talk again with @jeffbaumes and @zachmullen about whether or not we want the current sort of introspection of JSON objects versus explicitly toggling a metadata item as JSON which I know @zachmullen wanted, and how we would store that information (if at all).
- [ ] Maintaining state of the expansion of the tree when switching between various modes (this could be a potentially big change)

zachmullen:
@danlamanna The commit I just made on this topic fixes the issue you were seeing where all the web client tests were failing very early.

danlamanna:
@zachmullen Thanks for setting up the polyfill!

Everything is still a work-in-progress.

As of b603b40 everything for **newly created** metadata fields should function as expected. I haven't gotten to the toggling of fields on from simple/json which should help for integrating this into already existing projects. From a technical standpoint it should be as simple as converting the ```MetadatumEditWidget``` to a ```JsonMetadatumEditWidget``` and vice versa, and things should 'just work'.

In addition, the existing tests pass and I've created one incredibly simple test case for the pretty printing functionality. Left to do is write tests which manipulate the tree and code editor and ensure things are being persisted/rendered properly. @zachmullen Does this sound like something that should be put into it's own spec?

zachmullen:
> Left to do is write tests which manipulate the tree and code editor, converting, and ensuring things are being persisted/rendered properly. @zachmullen Does this sound like something that should be put into it's own spec?

Sure, that sounds like a good plan.

danlamanna:
@zachmullen Is this another side-effect of not using AMD: https://travis-ci.org/girder/girder/jobs/70115048#L4121

Not sure how to correct that.

danlamanna:
The convert works like this:
- Trick the editor into thinking it is the other type
- Re-render the metadata. (This is because trying to re-assign the json-metadatum editor to a simple (and vice versa) was causing issues with elements still being bound to the old objects.)

The noticeable downsides to this approach are:
- If you toggle and then cancel - nothing happens on the database side, refreshing the app will reveal how the object is actually stored.   
This is revealed even more so if you toggle -> cancel more than one metadata field, because it re-renders all of them. However I'm not so sure how common a use case that may be.

danlamanna:
A couple of notes:
- Jeff mentioned to re-work the indication of a simple/json field, removing simple and only showing an indicator if it's json.
- In regards to the above toggle -> cancel workflow, cancel should just switch the conversion in the other direction so it appears as though nothing has changed (since it hasn't)

clients/web/src/templates/widgets/jsonMetadatumEditWidget.jade
```diff
@@ -0,0 +1,13 @@
+input.input-sm.form-control.g-widget-metadata-key-input(type="text", value=key, placeholder="Key")
+
+#json-editor
```
zachmullen: Let's use a class here instead of an ID, since ID's are supposed to be globally unique and this is a widget that we may want to instantiate multiple times. Also let's prefix our classes with `g-` to avoid naming collisions.

zachmullen:
This works well, but I have one design/architectural suggestion.

The holy grail here would be constructing this so that it is easy for plugins to hook into this widget and add new editing modes and widgets for those modes, perhaps via some sort of mapping that could be dynamically altered, e.g.

```
modes: {
    simple: {
        widget: girder.views.MetadatumEditWidget,
        title: 'Simple'
    },
    json: {
        widget: girder.views.JsonMetadatumEditWidget,
        title: 'JSON'
    }
}
```

From a UI perspective, this means that the toggle button when editing should instead be more like the dropdown menu button that you made for "Add metadata". From a code perspective, it means we'll need to rethink places where we see `isJson` or other binary flags, and replace them with more flexible behavioral mappings.

We don't need to do this all right now on this topic, I'm just trying to think about the future of this functionality and record my thoughts for posterity.

clients/web/test/testUtils.js
```diff
@@ -289,19 +289,43 @@ girderTest.createGroup = function (groupName, groupDesc, pub) {
...
-     * :param action: one of 'save', 'cance', or 'delete'.  'delete' can't be
+     * :param action: one of 'save', 'cancel', or 'delete'.  'delete' can't be
      *                used with new items.  Default is 'save'.
      * :param errorMessage: if present, expect an information message with
      *                      regex.
      */
-    function _editMetadata(origKey, key, value, action, errorMessage) {
+    function _editMetadata(origKey, key, value, action, errorMessage, type) {
         var expectedNum, elem;
+        type = (type !== undefined && type === 'json') ? 'json' : 'simple';
```
zachmullen: The above line would be more readable as

    type = type || 'simple';

clients/web/test/testUtils.js
```diff
@@ -309,7 +333,12 @@ girderTest.testMetadata = function () {
             }, 'the add metadata button to appear');
             runs(function () {
                 expectedNum = $(".g-widget-metadata-row").length;
-                $('.g-widget-metadata-add-button:visible').click();
+
+                if (type === 'json') {
+                    $('a.g-add-json-metadata').click();
+                } else {
+                    $('a.g-add-simple-metadata').click();
+                }
```
zachmullen: The above blocks can be reduced to

    $('a.g-add-' + type + '-metadata').click();

danlamanna:
@zachmullen With this commit brings a good amount of test cases that cover some common patterns I believe will be used.

I agree on your architectural decision, but I think it might make sense to first solidify this case before abstracting it. Ultimately I think it may be a resources decision for @jeffbaumes.

@zachmullen The biggest issue I see at this point is the inability to render an individual metadatum. The metadata widget renders all of the metadata at once. This leads to situations like:
https://github.com/girder/girder/pull/879/files#diff-55b29b7583ffb7b88667605d3ce39818R178
https://github.com/girder/girder/pull/879/files#diff-55b29b7583ffb7b88667605d3ce39818R225

Which results in:
* Potentially bad performance (large metadata getting re-rendered often)
* Sloppy from the code perspective (just doing more work than we have to)
* Poor user experience
  * Trying to edit multiple pieces of metadatum at once
    * For example, with 2 (of any type) of editors open, trying to convert either one of them will result in lost changes in the other.

I know metadatum may not fit the REST structure of backbone models, but would having a metadatum model and collection solve this? That way each piece of metadatum would have it's own render we could call when it needs to be converted.

zachmullen:
> I know metadatum may not fit the REST structure of backbone models, but would having a metadatum model and collection solve this? That way each piece of metadatum would have it's own render we could call when it needs to be converted.

It seems to me what we actually need is to be able to re-render smaller pieces of the metadata view widget, which would just involve creating a new View. Rather than rendering them all in a single template under MetadataWidget, the MetadataWidget could contain a list of MetadatumWidgets that all could be rendered individually when one of them changed.

jeffbaumes:
This is a tricky fix which involves integrating a major library. I am ok with putting a bit more resources into it, and I think we should fix up the view rendering before this goes in. @zachmullen can you put a bit more meat on your suggestion through a follow up with @danlamanna? I'm ok if this is back-burner and doesn't go in for another week.

zachmullen:
Yep, can do.

mgrauer:
As a possible extension of this from a plugin, I'd be interested in the ability to display an arbitrary json object with checkboxes for each full path/value pair, similar to what is shown [in this earlier screenshot](https://github.com/girder/girder/pull/879#issuecomment-116838869), then allow the user to select checkboxes and finally pass back a list of full keypaths to some callback.  

An example use for this is: A user has a large collection of tweets that they want to render on a map.  By first looking at a single example json object, they would select the set of keypaths they want included in some conversion process of the json such that only the selected keypaths are included in the output. 

This might be a bit too far away from the use case of editing json metadata.

danlamanna:
Most of this is setup structurally speaking, there are just a number of todos in for my owns sake - I still have to go over it and clean some things up.

There are a [few places](https://github.com/girder/girder/commit/b4ed39cb8245a3129a4802b3c841ba391a7f8480#diff-55b29b7583ffb7b88667605d3ce39818R249) where the modal editing isn't directly utilized but could be pretty easily - this includes the ability to have an actual dropdown with all of the "modes". Fortunately the facilities are there to allow us to do this at any time, should we go ahead and implement this now?

In addition some more testing could be implemented, specifically changing one area of the tests to allow for many editors to be open at one time - though I'm not sure how necessary this is. I have a patch for @zachmullen to take a look at which partially implements it.

Again, I need to spend a little more time going over this cleaning up some of the rougher edges.

jeffbaumes:
@danlamanna and I just discussed what it will take to merge this PR. I suggested:
* There are a couple error cases his testing does not cover, but it will be a bit prohibitive to add these to testing right now so I suggested he leave testing as is.
* He explained @zachmullen's goal of having conversions be entirely self-describing and eliminate special code needed for toggling the simple and JSON editors. I think what he has looks clean enough for now and we should open another issue for cleaning that up.
* I think this should be merged *after* the next release tag so it has ample time for dev testing on people's projects before it goes into an official release.
* I think @mgrauer's suggestion seems a bit out of scope for this PR and sounds more like editing a JSON data file rather than a metadata feature. It should be captured in a separate issue.

If this sounds reasonable, @danlamanna can notify us when he feels this is ready for final approval.

danlamanna:
This is ready for review; I won't merge it until after the upcoming girder release.

clients/web/src/templates/widgets/metadataWidget.jade
```diff
@@ -1,11 +1,18 @@
-include metadatumViewMixin
 if (accessLevel >= girder.AccessType.WRITE)
-  button.btn.btn-sm.btn-primary.g-widget-metadata-add-button(title="Add Metadata")
-    i.icon-plus
+  .btn-group(style="display:block;float:right;")
```
zachmullen: Let's confine our styles to the stylus files.

clients/web/src/templates/widgets/metadataWidget.jade
```diff
@@ -1,11 +1,18 @@
...
+    button.g-widget-metadata-add-button.btn.btn-sm.btn-primary.btn-default.dropdown-toggle(data-toggle="dropdown", title="Add Metadata")
+      i.icon-plus
+    ul.dropdown-menu.pull-right(role="menu")
+      li(role="presentation")
+        a.g-add-simple-metadata(role="menuitem")
+          i.g-add-simple-metadata
+          | Simple
+      li(role="presentation")
+        a.g-add-json-metadata
+          i.g-add-json-metadata
```
zachmullen: What is the `i` element here for (and the one just above this)? Did you mean to put some icons next to these?

clients/web/src/templates/widgets/metadatumView.jade
```diff
@@ -1,2 +1,5 @@
-include metadatumViewMixin
-+metadatum(key, value, accessLevel, girder)
+if (accessLevel >= girder.AccessType.WRITE)
+  button.btn.btn-sm.btn-default.g-widget-metadata-edit-button
+    i.icon-pencil
+span.g-widget-metadata-key #{key}
+span.g-widget-metadata-value #{value}
```
zachmullen: Missing newline at EOF (should configure into your editor :)

clients/web/src/views/widgets/MetadataWidget.js
```diff
@@ -2,80 +2,418 @@
...
+            validation: {
+                from: {
+                    simple: [
+                        function (value) {
+                            try {
+                                var jsonValue = JSON.parse(value);
+                                /* This may succeed when we don't want it to (for instance with the
+                                 * value 'false' or '1234'), so check and only switch to JSON if we
+                                 * got an object back. */
+                                if (jsonValue && typeof jsonValue === 'object' && jsonValue !== null) {
```
zachmullen: The third expression `jsonValue !== null` is redundant. Also, should we really forbid scalars like `1234` and `false` since they are also valid JSON? I feel like we should just return true if JSON.parse didn't throw an exception.

clients/web/src/views/widgets/MetadataWidget.js
```diff
@@ -2,80 +2,418 @@
...
             key: '',
             value: '',
+            accessLevel: this.accessLevel,
             newDatum: true,
+            parentView: widget
+        }).render();
+    },
+
+    render: function () {
+        var _this = this;
```
zachmullen: `_this` isn't necessary here, if you pass a third argument to `_.each`, it will be used as the binding context on the internal function :smile:

clients/web/src/views/widgets/MetadataWidget.js
```diff
@@ -2,80 +2,418 @@
...
+            var mode;
+            this.key = tempKey;
+            this.value = tempValue;
+
+            this.parentView.key = this.key;
+            this.parentView.value = this.value;
+
+            if (this instanceof girder.views.JsonMetadatumEditWidget) {
+                this.parentView.mode = 'json';
+                mode = 'json';
```
zachmullen: What is the purpose of the local variable `mode` here?

zachmullen:
We should remove `clients/web/static/lib/jsoneditor/jsoneditor.min.css` from the repo since it is in the npm package under `dist/jsoneditor.min.css`. We can have a grunt step that simply copies it into `clients/web/static/built` somewhere.

danlamanna:
@zachmullen I think this is ready to be looked at again.

I think if the value was saved as a non string (false, 6, etc) it should be opened in the JSON editor. I know this disagrees with what we were talking about - but I think it only makes sense that if the user went out of their way to make "false" an object that it opens as something that would save it the same way.


This is the workflow I pictured that would break the type checking:
false (bool) is the metadata value.
The user clicks edit and sees it in a simple metadata field, changes it to true, and saves.
Now true is a string as opposed to the boolean they probably wanted.

This sort of thing could probably be fixed by adding a g-type to the dom element, but I think the code would start to get pretty convoluted at that point.

jeffbaumes:
@zachmullen all good? @danlamanna usually good to edit the PR name to remove WIP once ready to go.

clients/web/src/views/widgets/MetadataWidget.js
```diff
@@ -2,80 +2,421 @@
...
+            }
+        }
+
+        return true;
+    },
+
+    // @todo too much duplication with editMetadata
+    toggleEditor: function (event, newEditorMode, existingEditor, overrides) {
+        var fromEditorMode = (existingEditor instanceof girder.views.JsonMetadatumEditWidget) ? 'json' : 'simple';
+        if (!this._validate(fromEditorMode, newEditorMode, (overrides || {}).value || existingEditor.$el.attr('g-value'))) {
```
zachmullen: Can we break the above two lines so they aren't quite so long?

danlamanna: Done, I noticed there is no JS max line length in our jshint - is that something we should change?

zachmullen:
I played  around with this for a while and everything works pretty intuitively, nice job.

There is one thing that is sort of a change in existing behavior and I'd like input from @girder/developers @girder/owners about the "right" behavior here:

Before, if you created or edited a metadata field and wrote something like `{"foo": "bar"}` inside, it would figure out it was JSON and automatically store that field as an object instead of just a string. Now, if you specify that it is a simple type and do the same thing, it escapes it so that it remains a string, and it will be stored as `"{\"foo\": \"bar\"}"`. This is probably better behavior IMO since it's more explicit, but it is a change to how it was before, so I wanted to hear input from other stakeholders before we give this the go-ahead.

jeffbaumes:
I am good with this change from implicit to explicit behavior.

cpatrick:
I like the change in behavior. :shipit:

zachmullen:
LGTM.

jcfr:
Looks like a nice improvement. Do you have a link to local test instance to see the json editor in action ?

jcfr:
Nevermind. I didn't see the screenshot at the beginning of the thread. :+1:

PR #882, opened 2015-07-06
Improvements to plugin gruntfile handling

zachmullen:
* Allow YAML plugin files to also work
* Allow gruntfile specification without adding default targets


danlamanna:
LGTM.

PR #886, opened 2015-07-09
Use bson's JSON serialization for kwargs

cjh1:
This resolves the issue of serializing datetime objects and others.

### DIFF
--- a/plugins/jobs/server/models/job.py
+++ b/plugins/jobs/server/models/job.py
@@ -18,9 +18,9 @@
 ###############################################################################
 
 import datetime
-import json
 import pymongo
 import six
+from bson import json_util
 
 from girder import events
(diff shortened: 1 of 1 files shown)


zachmullen:
Nice change, but I think we should avoid aliasing it since that may cause confusion if someone doesn't look up at the import line.

cjh1:
@zachmullen OK, I was in two mines, just wanted to keep change to minimum ... whats the downside if a developer mistook the import for the system one?

zachmullen:
Probably none, I guess I just prefer not to alias module names to other module names, but maybe this is a good case for it. I leave it up to you.

cjh1:
@zachmullen I have made the change.

zachmullen:
Lgtm

PR #888, opened 2015-07-09
Make models always expose the '_id' field

brianhelba:
All current models that use the filter capability add the '_id' field to their list of exposed fields for all accesses. To ensure that the important '_id' field is easily availble in new models, it is now part of the default set of exposed keys.

It is still possible to call "Model.hideFields" if necessary.

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -46,7 +46,7 @@ class Model(ModelImporter):
         self._textLanguage = None
 
         self._filterKeys = {
-            AccessType.READ: set(),
+            AccessType.READ: set('_id'),
             AccessType.WRITE: set(),
             AccessType.ADMIN: set(),
             AccessType.SITE_ADMIN: set()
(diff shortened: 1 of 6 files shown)


girder/models/model_base.py
```diff
@@ -46,7 +46,7 @@ def __init__(self):
         self._textLanguage = None
 
         self._filterKeys = {
-            AccessType.READ: set(),
+            AccessType.READ: set('_id'),
```
zachmullen: Should be `set(['_id'])` or `set(('_id',))`.

zachmullen:
There are some security implications here; our token model actually uses the `_id` field as the secure token value. I don't think we ever filter it currently, but this change makes using the `_id` field in a protected capacity more dangerous as it breaks our usual paradigm of everything-is-private-by-default.

brianhelba:
Abandoning for now to keep things clean. I may revisit this.

PR #889, opened 2015-07-09
Factor out request error handling for reuse

cjh1:
This allows applications to intercept error such as 404's that may have "special" meaning within an application and then delegate the rest to the Girder client. For example if the application need to test if a resource exists then they wouldn't want he 404 to be reported to the user.

### DIFF
--- a/clients/web/src/init.js
+++ b/clients/web/src/init.js
@@ -51,6 +51,56 @@ _.extend(girder, {
         EMPTY: 'empty'
     },
     layout: 'default',
+    requestError: function (error, status) {
+        var info;
+        if (error.status === 401) {
+            girder.events.trigger('g:loginUi');
+            info = {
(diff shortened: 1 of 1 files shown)


cjh1:
I guess an alternative to this is just change my server code not to produce 404 for not found, thoughts?

zachmullen:
I'm fine with refactoring this out if it is helpful, did you choose to go a different route?

cjh1:
Yep, choose to go a different route in the end, as the browser will report 404 as an error anyway its probably not good to use it to check for the existence of a resource.

zachmullen:
FWIW, in girder we chose not to distinguish between the 404 vs. 401/403 cases for security reasons, and instead we just return the more generic 400 and say "no such resource" so that someone can't detect the existence of something to which they do not have access.

PR #890, opened 2015-07-10
Prompt for login upon 401 for item view

zachmullen:


mgrauer:
LGTM.  

@zachmullen You mind doing the Folder view also?  Should be the same fix.  I'll pre-LGTM that as long as Travis passes.

PR #891, opened 2015-07-13
new route: /resource/stat

opadron:
Adds a new stat route under `/resource`.

Accepts a `path` query parameter that is a "girder path" and responds with the `user`, `collection`, `folder`, `item`, or `file` resource identified by `path`.

### DIFF
--- a/girder/api/v1/resource.py
+++ b/girder/api/v1/resource.py
@@ -150,24 +150,26 @@ class Resource(BaseResource):
         :returns: the child resource
         """
 
-        filterObject = {'name': token, 'parentId': parent['_id']}
+        seekFolder = (parentType in ('user', 'collection', 'folder'))
+        seekItem = (parentType == 'folder')
+        seekFile = (parentType == 'item')
 
-        # look for a folder
-        candidateChild = self.model('folder').findOne(filterObject)
-        if candidateChild is not None:
-            return candidateChild, 'folder'
+        # (model name, mask, search filter)
+        searchTable = (
+            ('folder', seekFolder, {'name': token,
+                                    'parentId': parent['_id'],
+                                    'parentCollection': parentType}),
+            ('item', seekItem, {'name': token, 'folderId': parent['_id']}),
+            ('file', seekFile, {'name': token, 'itemId': parent['_id']}),
+        )
 
-        # if folder not found, look for an item
-        filterObject['folderId'] = filterObject.pop('parentId')
-        candidateChild = self.model('item').findOne(filterObject)
-        if candidateChild is not None:
-            return candidateChild, 'item'
+        for candidateModel, mask, filterObject in searchTable:
+            if not mask:
+                continue
 
-        # if item not found, look for a file
-        filterObject['itemId'] = filterObject.pop('folderId')
-        candidateChild = self.model('file').findOne(filterObject)
-        if candidateChild is not None:
-            return candidateChild, 'file'
+            candidateChild = self.model(candidateModel).findOne(filterObject)
+            if candidateChild is not None:
+                return candidateChild, candidateModel
 
         # if no folder, item, or file matches, give up
         raise RestException('Child resource not found: {}({})->{}'.format(
--- a/girder/api/v1/resource.py
+++ b/girder/api/v1/resource.py
@@ -141,43 +141,33 @@ class Resource(BaseResource):
             raise RestException('Invalid resources format.')
         return model
 
-    def _lookupToken(self, token, parentId):
+    def _lookupToken(self, token, parentType, parent):
         """
         Find a particular child resource by name or throw an exception.
         :param token: the name of the child resource to find
-        :param parentId: id of the parent to search
-        :returns: the child resource.
+        :param parentType: the type of the parent to search
+        :param parent: the parent resource
+        :returns: the child resource
         """
 
-        filterObject = {'name': token}
-        candidateParent = None
+        filterObject = {
+            'name': token,
+            'parentId': parent['_id'],
+            'baseParentId': parent.get('baseParentId'),
+            'baseParentType': parent.get('baseParentType'),
+        }
 
-        # figure out the parentId's type
-        candidateModelNames = ('collection', 'user', 'folder', 'item')
-        for modelName in candidateModelNames:
-            candidateParent = self.model(modelName).findOne({'_id': parentId})
+        if parentType in ('collection', 'user'):
+            filterObject.update(
+                baseParentId=parent['_id'],
+                baseParentType=parentType)
 
-            if candidateParent is not None:
-                if modelName in ('collection', 'user'):
-                    filterObject.update(
-                        parentId=parentId,
-                        baseParentId=parentId,
-                        baseParentType=modelName)
-                else:
-                    filterObject.update(
-                        parentId=parentId,
-                        baseParentId=candidateParent['baseParentId'],
-                        baseParentType=candidateParent['baseParentType'])
-                break
-
-        if candidateParent is None:
-            raise RestException(
-                'Parent resource not found: {}'.format(parentId))
-
+        # look for a folder
         candidateChild = self.model('folder').findOne(filterObject)
         if candidateChild is not None:
             return candidateChild, 'folder'
 
+        # if folder not found, look for an item
         filterObject['folderId'] = filterObject.pop('parentId')
         del filterObject['baseParentId']
         del filterObject['baseParentType']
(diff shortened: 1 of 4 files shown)


zachmullen:
The python static analysis tests are failing: http://my.cdash.org/viewTest.php?onlyfailed&buildid=792811

opadron:
A lot of the static analysis complaints are over camel casing variable names.  Most of the code base uses `camelCase` names and I did the same to remain consistent.  What do we want to do as far as the static analysis tests go?

zachmullen:
The failures in question are pep8 problems. The test output tells you what lines are wrong and a message about why: http://my.cdash.org/testDetails.php?test=20072653&build=792811

opadron:
Ah, I see.  I'm getting something very different when I run the test suite locally.  I'll have another pass at it.

girder/api/v1/resource.py
```diff
@@ -139,6 +140,124 @@ def _getResourceModel(self, kind, funcName=None):
...
+
+        result = self.model(model).filter(document, user)
+        return result
+
+    @access.public
+    def stat(self, params):
+        path = params.get('path')
+        currentUser = self.getCurrentUser()
+        result = self._lookupPath(path, currentUser)
+        return result
```
zachmullen: Since LOC ∝ n_bugs, I vote we rewrite this method as

```
self.requireParams('path', params)
return self._lookupPath(params['path'], self.getCurrentUser())
```

(I included the requireParams call since currently a confusing exception would bubble up in cases where someone forgot to pass the path parameter.)

girder/api/v1/resource.py
```diff
@@ -139,6 +140,124 @@ def _getResourceModel(self, kind, funcName=None):
...
+        user must have read access to all resources featured in the path.
+        :param path: the path of the object to find
+        :param user: the user to search as
+        :returns: the resource.
+        """
+
+        if not path.startswith('/'):
+            raise RestException('Invalid path format')
+
+        pathArray = path[1:].split('/')
```
zachmullen: I vote we make the leading slash optional since we are always going to be getting an "absolute" path.  So we could change this to

    pathArray = [t for t in path.split('/') if t]

opadron: Good idea.  That should also work for paths with a trailing slash, too.

girder/api/v1/resource.py
```diff
@@ -139,6 +140,116 @@ def _getResourceModel(self, kind, funcName=None):
...
+        :param parentId: id of the parent to search
+        :returns: the child resource.
+        """
+
+        filterObject = {'name': token}
+        candidateParent = None
+
+        # figure out the parentId's type
+        candidateModelNames = ('collection', 'user', 'folder', 'item')
+        for modelName in candidateModelNames:
```
zachmullen: We could make this a lot more efficient by passing the parent type and parent resource (i.e. the dict rather than just the id). Then we could remove these extra lookups, which could get pretty expensive.

girder/api/v1/resource.py
```diff
@@ -139,6 +140,96 @@ def _getResourceModel(self, kind, funcName=None):
...
+        candidateChild = self.model('item').findOne(filterObject)
+        if candidateChild is not None:
+            return candidateChild, 'item'
+
+        # if item not found, look for a file
+        filterObject['itemId'] = filterObject.pop('folderId')
+        candidateChild = self.model('file').findOne(filterObject)
+        if candidateChild is not None:
+            return candidateChild, 'file'
+
```
zachmullen: We should add some conditions to the above code so that we don't do unnecessary lookups. Namely,

1. If parentType is "user" or "collection", we only need to lookup folders.
2. If parentType is "folder", we only need to lookup folders and items.
3. If parentType is "item", we only need to lookup files.

Additionally, when looking up folders, we should include `'parentCollection': parentType` in the filter dict.

girder/api/v1/resource.py
```diff
@@ -139,6 +140,96 @@ def _getResourceModel(self, kind, funcName=None):
...
+        :param path: the path of the object to find
+        :param user: the user to search as
+        :returns: the resource.
+        """
+
+        pathArray = [token for token in path.split('/') if token]
+        model = pathArray[0]
+
+        if user is None:
+            user = self.getCurrentUser()
```
zachmullen: Let's remove the above block and simply pass the user in rather than provide a default. `None` is actually a valid user, and since this is a private method it won't be a big deal to force passing the user.

zachmullen:
@girder/developers How do others feel about the name `stat` for this route? I feel like it may be a bit too technical, but I'm having trouble coming up with a good name for such an endpoint. (Its purpose is to take in a path into the data hierarchy and spit out the corresponding resource record if it exists.)

jcfr:
What about `history` ?

cpatrick:
_inspect_, _details_?

zachmullen:
I was thinking of `resource/lookup`.

cpatrick:
We have a winner!

girder/api/v1/resource.py
```diff
@@ -139,6 +140,96 @@ def _getResourceModel(self, kind, funcName=None):
...
+        return result
+
+    @access.public
+    def stat(self, params):
+        self.requireParams('path', params)
+        return self._lookupPath(params['path'], self.getCurrentUser())
+
+    stat.description = (
+        Description('Get any resource by girder path.')
+        .param('path', 'The path of the resource.')
```
zachmullen: Let's change this description to "Lookup a resource in the data hierarchy by path." and we can add a notes section (or a longer param description) that describes the format of `path`.

ronichoudhury: Pedantic word nerd here - I think "lookup" is strictly a noun and the description should say "look up".  Sorry, for some reason lookup-vs-"look up" really bugs me :eyeglasses:

zachmullen: :+1:

jbeezley: Note to self, look up lookup in the dictionary.

girder/models/file.py
```diff
@@ -37,6 +37,10 @@ def initialize(self):
         self.resourceColl = 'item'
         self.resourceParent = 'itemId'
 
+        self.exposeFields(level=AccessType.READ, fields=(
+            "_id", "mimeType", "itemId", "exts", "name", "created",
+            "assetstoreId", "creatorId", "size"))
```
zachmullen: @opadron let's expose `assetstoreId` only under `AccessType.SITE_ADMIN`; assetstores are a detail of the system that is really only exposed to administrators.

tests/cases/resource_test.py
```diff
@@ -350,6 +350,83 @@ def testGetResourceById(self):
...
+                            method='GET', user=self.user,
+                            params={'path': '/user/goodlogin/Public'})
+        self.assertStatusOk(resp)
+        self.assertEqual(
+            str(resp.json['_id']), str(self.adminPublicFolder['_id']))
+
+        resp = self.request(path='/resource/lookup',
+                            method='GET', user=self.user,
+                            params={'path': '/user/goodlogin/Private'})
+        self.assertStatus(resp, 403)
```
zachmullen: Can we also add a test for the error condition where an invalid path is passed and we get a 400?

girder/api/v1/resource.py
```diff
@@ -139,6 +140,99 @@ def _getResourceModel(self, kind, funcName=None):
...
+    @access.public
+    def lookup(self, params):
+        self.requireParams('path', params)
+        return self._lookUpPath(params['path'], self.getCurrentUser())
+
+    lookup.description = (
+        Description('Look up a resource in the data hierarchy by path.')
+        .param('path',
+               'The path of the resource.  The path must be an absolute Unix '
+               'path starting with either "/users/[user name]", for a user\'s'
```
zachmullen: I think this should say `"/user/[user name]"`.

girder/api/v1/resource.py
```diff
@@ -139,6 +140,99 @@ def _getResourceModel(self, kind, funcName=None):
...
+            parentType, parent.get('name', parent.get('_id')), token))
+
+    def _lookUpPath(self, path, user):
+        """
+        Find a particular resource by path or throw an exception.  The given
+        user must have read access to all resources featured in the path.
+        :param path: the path of the object to find
+        :param user: the user to search as
+        :returns: the resource.
+        """
```
zachmullen: I think sphinx will throw warnings about this docstring due to not having a blank line after the description. Feel free to remove the docstring altogether; most of our REST route handlers don't have them since they are redundant with the `description` property.

zachmullen:
LGTM

opadron:
rebased from the latest master

opadron:
I'm not sure what to do about [this](http://my.cdash.org/testDetails.php?test=20855466&build=821696) failing test.

zachmullen:
All set.

PR #892, opened 2015-07-14
Add setting and UI for collection creation policies

zachmullen:


zachmullen:
This is ready for review. It allows admins to declare a list of users and groups that are allowed to create collections on the instance, or optionally allow any logged-in user to create collections.

mgrauer:
@zachmullen 

Had you thought to add the collection policy to the user object, and thus we could render the "create collection" button or not depending on current policies?  As it is now, the button will be rendered and then if the policy is disallowed, an error message will be returned.  Did you decide against this for some reason?

Everything worked well with some minor trouble in editing the policy.  This might all be improved later on by the addition of the JSON editor widget.  I pushed a couple commits in the hope of fixing these issues.  These could probably be lived with without my changes, as you can mostly get the behavior you want through a combination of saving and refreshing.  My changes do the following

  - If you hand edit the JSON, then try to update the JSON by adding something through the search widget, the JSON won't get updated.  Fixed by changing the set from `text` to `val`.
  - Based on that change, now that the hand edited JSON was getting updated, the `resultClicked` event handler from the search widget used the saved setting as its source, so that if you hand edited the JSON and then tried to use the search widget before saving, your hand edited changes would be lost.  I passed in the settings as the original source to the widget, then in the event handler I'm now getting the current value of the JSON document being edited.
  - I'm preventing duplicate adds of users or groups.  This already happens and gets cleaned up on the back-end, and upon saving it will display a `set` of users or groups.  I put this in a separate commit for ease of pitting should you not want the extra code/debt.

mgrauer:
2 :ghost: failures that were fixed by restarting Travis.

PR, Python 2.7, web_client_collection
Push, Python 2.7, web_client_item

zachmullen:
> Had you thought to add the collection policy to the user object, and thus we could render the "create collection" button or not depending on current policies?

I hadn't thought about that at all actually, but it does seem like a good alternative.

> As it is now, the button will be rendered and then if the policy is disallowed, an error message will be returned. Did you decide against this for some reason?

It's not really intended, that's just kind of a limitation of storing it in the settings as I did here -- we don't know until we try, unless we fetched the settings to the browser somehow, which would have to involve exposing some other endpoint.

I do really like your suggestion though, as it could help when extending this sort of permission model. I.e., group and user documents could contain an additional field called `privileges` that would be a list of global privileges that they are allowed. I think I will change to this and rebase.

zachmullen:
@mgrauer I thought about it a bit more, and it won't be quite as simple as I thought, mainly because we want to have the ability to add this privilege to groups. If we added that privilege on the group document rather than a system setting, we would still need to do a lookup of all of a user's groups to determine whether they were allowed to create collections, which I don't necessarily want to do anytime a user goes to the collection list page. I think it would be better to keep it as it currently is, so we only have to perform the lookup when someone is actually creating a collection.

mgrauer:
Fine by me.  Just wanted to know your thinking on the matter, and now I do.

zachmullen:
Thanks for the improvements @mgrauer . I give the "LGTM" on your changes.

PR #901, opened 2015-07-24
Support range downloading

zachmullen:
Downloading will now also support the HTTP Range header as an
alternate means of specifying byte ranges when downloading files.
This functionality is now implemented in both the filesystem and
gridfs assetstores. The others will follow shortly.

### DIFF
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -212,14 +212,32 @@ class File(Resource):
         Defers to the underlying assetstore adapter to stream a file out.
         Requires read permission on the folder that contains the file's item.
         """
-        offset = int(params.get('offset', 0))
-        return self.model('file').download(file, offset)
+        rangeHeader = cherrypy.lib.httputil.get_ranges(
+            cherrypy.request.headers.get('Range'), file.get('size', 0))
+
+        # The HTTP Range header takes precedence over query params
+        if rangeHeader and len(rangeHeader):
+            # Currently we only support a single range.
+            offset, endByte = rangeHeader[0]
+        else:
+            offset = int(params.get('offset', 0))
+            endByte = params.get('endByte')
+
+            if endByte is not None:
+                endByte = int(endByte)
+
+        return self.model('file').download(file, offset, endByte=endByte)
     download.cookieAuth = True
     download.description = (
         Description('Download a file.')
+        .notes('This endpoint also accepts the HTTP "Range" header for partial '
+               'file downloads.')
         .param('id', 'The ID of the file.', paramType='path')
         .param('offset', 'Start downloading at this offset in bytes within '
                'the file.', dataType='integer', required=False)
+        .param('endByte', 'If you only wish to download part of the file, '
+               'pass this as the index of the last byte to download.',
+               dataType='integer', required=False)
         .errorResponse('ID was invalid.')
         .errorResponse('Read access was denied on the parent folder.', 403))
 
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -236,8 +236,11 @@ class File(Resource):
         .param('offset', 'Start downloading at this offset in bytes within '
                'the file.', dataType='integer', required=False)
         .param('endByte', 'If you only wish to download part of the file, '
-               'pass this as the index of the last byte to download.',
-               dataType='integer', required=False)
+               'pass this as the index of the last byte to download. Unlike '
+               'the HTTP Range header, the endByte parameter is non-inclusive, '
+               'so you should set it to the index of the byte one past the '
+               'final byte you wish to recieve.', dataType='integer',
+               required=False)
         .errorResponse('ID was invalid.')
         .errorResponse('Read access was denied on the parent folder.', 403))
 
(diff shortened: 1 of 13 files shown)


zachmullen:
This is ready for review. The s3 assetstore gives us this essentially for free because range headers get forwarded on redirects and s3 supports them.

girder/api/v1/file.py
```diff
@@ -212,14 +212,32 @@ def download(self, file, params):
...
     download.description = (
         Description('Download a file.')
+        .notes('This endpoint also accepts the HTTP "Range" header for partial '
+               'file downloads.')
         .param('id', 'The ID of the file.', paramType='path')
         .param('offset', 'Start downloading at this offset in bytes within '
                'the file.', dataType='integer', required=False)
+        .param('endByte', 'If you only wish to download part of the file, '
+               'pass this as the index of the last byte to download.',
+               dataType='integer', required=False)
```
manthey: Perhaps we could make it clearer that the endByte is exclusive, as opposed to the http Range header where it is inclusive.  A range header of 10-19 gets 10 bytes, which is the same as offset=10, endByte=20, if I am reading this correctly.

zachmullen: That's correct. I will add that in.

plugins/hdfs_assetstore/server/assetstore.py
```diff
@@ -129,17 +128,30 @@ def downloadFile(self, file, offset=0, headers=True):
...
+                    if position + chunkLen <= offset:
+                        position += chunkLen
                     else:
-                        yield chunk[offset - skipped:]
-                        skipped = offset
+                        if position + chunkLen > endByte:
+                            chunkLen = endByte - position
+                            shouldBreak = True
+                        yield chunk[offset - position:chunkLen]
+                        position = offset
```
manthey: The preceding five lines aren't covered in the tests.  They look right to me, but I'd feel better if they were covered.

zachmullen: Good catch, I'll correct that.

tests/cases/file_test.py
```diff
@@ -228,6 +228,21 @@ def _testDownloadFile(self, file, contents):
...
+        self.assertEqual(contents[2:8], self.getBody(resp))
+        self.assertEqual(resp.headers['Accept-Ranges'], 'bytes')
+        length = len(contents)
+        begin, end = min(length, 2), min(length, 8)
+        self.assertEqual(resp.headers['Content-Length'], end - begin)
+
+        if length:
+            rangeVal = 'bytes %d-%d/%d' % (begin, end - 1, len(contents))
+            self.assertEqual(resp.headers['Content-Range'], rangeVal)
+
```
manthey: Can you also add a test that uses explicit `offset` and `endByte` parameters?

girder/utility/s3_assetstore_adapter.py
```diff
@@ -316,11 +319,7 @@ def downloadFile(self, file, offset=0, headers=True):
                 url = urlFn(key=file['s3Key'])
                 raise cherrypy.HTTPRedirect(url)
             else:
-                cherrypy.response.headers['Content-Length'] = '0'
-                cherrypy.response.headers['Content-Type'] = \
-                    'application/octet-stream'
-                cherrypy.response.headers['Content-Disposition'] = \
-                    'attachment; filename="{}"'.format(file['name'])
```
manthey: Out of curiosity, why don't we want to set the Content-Disposition any more?

manthey: Never mind, I see it was moved.

plugins/hdfs_assetstore/server/assetstore.py
```diff
@@ -129,17 +128,30 @@ def downloadFile(self, file, offset=0, headers=True):
...
+                    if position + chunkLen <= offset:
+                        position += chunkLen
                     else:
-                        yield chunk[offset - skipped:]
-                        skipped = offset
+                        if position + chunkLen > endByte:
+                            chunkLen = endByte - position
+                            shouldBreak = True
+                        yield chunk[offset - position:chunkLen]
+                        position = offset + chunkLen - 1
```
manthey: Shouldn't this line just be `position += chunkLen`?

manthey: In which case, `position += chunkLen` could be moved outside of the if statements, since all branches use it.

zachmullen: This chunk can be truncated at either the beginning or the end, so we'd have to take the beginning truncation into account in this block.

manthey: Yes, which is why you `yield [offset - position:chunkLen]`.

Isn't `position` functionally the file position (`ftell` if this were a file rather than a stream)?  In which case, `position` needs to be incremented by the amount of data read (`chunkLen`).  I'm not following why the logic would be `offset + chunkLen - 1`.

zachmullen: Ah you're right, changing.

manthey:
LGTM.

zachmullen:
Thanks for the QA @manthey

PR #903, opened 2015-07-29
Fix import capabilities for our plugins pseudo-package

zachmullen:
Before this change, it was not possible to do, e.g.:

    from girder.plugins import jobs

It is now possible and tested.


zachmullen:
ping @cjh1 I believe you were the one who ran into this problem before.

cjh1:
@zachmullen Thanks, LGTM

PR #904, opened 2015-07-30
Add an option to disable JavaScript tests

cjh1:
This adds the ability to disable JavaScript tests. I have a Girder ( server side only ) plugin and would like to run unit tests without having worry about npm javascript etc.


cjh1:
Travis may need kicking, don't think these failures are anything to do with the changes

zachmullen:
I restarted it, one of the python processes got signalled in some strange way that I've never seen before in CTest.

zachmullen:
LGTM once travis is happy

PR #905, opened 2015-08-04
geospatial plugin client test does not import the plugin client code

mgrauer:
The current geospatialSpec does not actually test that the plugin client code is loaded.  This PR first adds a test which now correctly fails.  A fix will follow.


mgrauer:
ready for review

zachmullen:
Lgtm

zachmullen:
This could be the textbook case for the importance of TDD :)

PR #906, opened 2015-08-05
Changes to allow plugin apps to test clients

mgrauer:
These changes allow a separate web app (a different web app from girder) inside a plugin to run client side tests.


Gruntfile.js
```diff
@@ -453,6 +457,7 @@ module.exports = function (grunt) {
     grunt.registerTask('test-env-html', 'Build the phantom test html page.', function () {
         var buffer = fs.readFileSync('clients/web/test/testEnv.jadehtml');
         var globs = grunt.config('uglify.app.files')['clients/web/static/built/app.min.js'];
+        var dependencies = [];
```
zachmullen: It looks like this only ever gets assigned as `[]`, so we can probably just remove the variable altogether and pass the empty array directly into `jsFilesUncovered`.

zachmullen: Actually ignore this comment as it contradicts my other comment below.

clients/web/test/testEnv.jadehtml
```diff
@@ -13,7 +13,10 @@ html(lang="en")
     #g-global-info-apiroot.hide %HOST%/#{apiRoot}
     #g-global-info-staticroot.hide %HOST%/#{staticRoot}
 
-    each js in jsFiles
+    each js in jsFilesUncovered
+      script(src="#{js}")
```
zachmullen: I would suggest we remove `libs.min.js` from line 8 and instead pass it in with `jsFilesUncovered`, just to increase the flexibility of this template slightly.

mgrauer:
@zachmullen PTAL

mgrauer:
Didn't see your comment, I agree.  I'll push a change.

zachmullen:
@mgrauer can you merge origin/master into this, it now has conflicts.

mgrauer:
@danlamanna @zachmullen 

JSON metadata editing branch #879 just went in--nice work BTW--causing conflicts on this branch.  Is `polyfill.min.js` used by the inline javascript in the jade test template, or can I move it into the Gruntfile?

zachmullen:
We can leave it hardcoded; it's tiny and gives us `Function.prototype.bind` inside the phantom environment.

mgrauer:
Travis passed, after a few restarts.

zachmullen:
LGTM

PR #907, opened 2015-08-05
Fixing heroku buildpacks.

cpatrick:
Turns out the default python buildpack supports libffi now.


zachmullen:
LGTM

zachmullen:
We should delete https://github.com/girder/heroku-buildpack-python-libffi

cpatrick:
I just did.

PR #911, opened 2015-08-06
Make debug-js beautiful.

manthey:
When the --debug-js option is specified with Grunt, beautify the results, but keep them ASCII.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -36,7 +36,10 @@ module.exports = function (grunt) {
 
     if (debugJs) {
         console.log('Building JS in debug mode'.yellow);
-        uglifyOptions.beautify = false;
+        uglifyOptions.beautify = {
+            beautify: true,
+            ascii_only: true
+        };
         uglifyOptions.mangle = false;
         uglifyOptions.compress = false;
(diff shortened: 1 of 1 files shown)


jeffbaumes:
What will this do to Girder JS code? One of the nice things is that with --debug-js the Girder JS code looks exactly as it does in your source repo - will this potentially mess that up?

manthey:
--debug-js was set to still uglify the code (beautify was set to False), so the code was minified.  Changing it to beautify: true stops that, but some parser was failing for @kotfic unless the ascii-fication was turned on.  So, the code won't look *exactly* like the source repo, but it will be similar modulo character encoding.  More work will have to be done for that, but prior to this PR, the code wasn't even close to the source.

jbeezley:
As I recall, beautify will also strip comments and join statements into a single line.  There isn't an option that I'm aware of that will just concatenate the files without processing.  In geojs, I do it with `grunt-contrib-concat`.

That said, the change here is desired.  The original setting was probably an oversight.

jeffbaumes:
Makes sense LGTM.

PR #913, opened 2015-08-10
Appends enabled plugins list to plugins variable

kotfic:
Also changes fn_ENABLEDPLIGINS to fn_ENABLEDPLUGINS


zachmullen:
Nice change, looks like this never would have worked before :)

zachmullen:
LGTM once travis passes

kotfic:
@zachmullen if you can merge this that would be great,  i don't have write access on girder.  Thanks!

zachmullen:
I have invited you to the group that has write access

PR #914, opened 2015-08-11
Allow endpoints to return any callable

brianhelba:
Now, an endpoint returning a lambda or a functools.partial object will be correctly recognised as a streaming response.

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -298,9 +298,10 @@ def endpoint(fun):
             if 'Content-Range' in cherrypy.response.headers:
                 cherrypy.response.status = 206
 
-            if isinstance(val, types.FunctionType):
-                # If the endpoint returned a function, we assume it's a
-                # generator function for a streaming response.
+            if callable(val):
+                # If the endpoint returned anything callable (function,
+                # lambda, functools.partial), we assume it's a generator
+                # function for a streaming response.
                 cherrypy.response.stream = True
                 return val()
 


girder/api/rest.py
```diff
@@ -298,9 +298,10 @@ def endpointDecorator(self, *args, **kwargs):
             if 'Content-Range' in cherrypy.response.headers:
                 cherrypy.response.status = 206
 
-            if isinstance(val, types.FunctionType):
-                # If the endpoint returned a function, we assume it's a
-                # generator function for a streaming response.
+            if callable(val):
+                # If the endpoint returned anything callable (function,
+                # lambda, functools.partial), we assume it's a generator
+                # function for a streaming response.
```
zachmullen: This change makes it so the below block will never be entered (`if isinstance(val, cherrypy.lib.file_generator)`), which would break any instance where someone used the cherrypy explicit static file methods. We can fix that by simply moving the second block above this one.

brianhelba: I don't think so. A `cherrypy.lib.file_generator` is not callable, as it does not contain an `__call__` method (not even one implicitly defined) and thus raises `TypeError: 'file_generator' object is not callable` if you try to call it. Feel free to try
```
import cherrypy
from cStringIO import StringIO
callable(cherrypy.lib.file_generator(StringIO()))
```
if you'd like.

I'm sure @zachmullen knows this next part, but I'll elaborate for the benefit of others: What may be causing some confusion is the particular way that Girder returns streaming responses. Rather than just returning a generator iterator for streaming responses, and testing whether it's a `types.GeneratorType` in this block, Girder typically returns the whole generator function. This function must then be trivially called once, whereby it returns a pristine generator iterator. `cherrypy.lib.file_generator` is an attempt to emulate this sort of generator iterator, not the generator function.

The only time you need the function itself is if you plan on iterating through it multiple times. Since that's not the case, Girder could have been designed to just have streaming responses be returned as generator iterators (or really any type of iterator). On the other hand, requiring streaming responses to return the original function is a nice protection against developers actually returning an iterator instead of a data structure, and accidentally producing a streaming response instead of getting automatic JSON-encoding.

zachmullen: You're right, I had messed up my test and accidentally tested the class and not the instance of it :)


LGTM

brianhelba:
Without objection, I'll merge this tonight, as I need it for my plugin.

PR #915, opened 2015-08-11
Make Model.load call findOne as a virtual method

brianhelba:
If a Model subclass overrides the "findOne" method, the "load" method should use this overridden method to fetch documents.


zachmullen:
LGTM

brianhelba:
Without objection, I'll merge this tonight, as I need it for my plugin.

PR #917, opened 2015-08-12
Add web client job collection

mgrauer:
that's all, nothing more.

### DIFF
--- a/plugins/jobs/plugin_tests/jobsSpec.js
+++ b/plugins/jobs/plugin_tests/jobsSpec.js
@@ -5,6 +5,7 @@ $(function () {
         '/static/built/plugins/jobs/templates.js',
         '/plugins/jobs/web_client/js/misc.js',
         '/plugins/jobs/web_client/js/models/JobModel.js',
+        '/plugins/jobs/web_client/js/collections/JobCollection.js',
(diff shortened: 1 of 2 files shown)


zachmullen:
Excellent... I've been meaning to make a job list widget but haven't had time, and this will be necessary for that. LGTM once travis passes

mgrauer:
I was going to do exactly that in Minerva.  Do you have any specific needs?  I'm on a tight timeline, so I can do a first run in Minerva and then we can generalize and bring it into Girder if that makes sense.

zachmullen:
My idea was a fairly flexible widget that would be inside the jobs plugin. It would render a list of jobs with high-level information in each row: title, progress information, and status. Clicking on a row (or a link within the row) would navigate to the existing job details route (`#job/{id}`) for more information about an individual job.

The main flexibility requirements would be for filtering the job list. I was imagining the filters would simply be passed into the widget at initialization. The most obvious use would be to show pages of jobs for a given user in reverse chronological order.

Ideally the job list widget would hook into the notification stream to update in real time for jobs that are not complete, but that feature could be added later.

zachmullen:
Once the widget was made, it was my intention to add an entry to the user dropdown menu that would be called "My jobs" that would link to the job list I described.

PR #918, opened 2015-08-12
Add a basic, flexible job list widget

zachmullen:
And add an instance of it to show reverse-chronological jobs for
the currently-logged-in user in the user dropdown menu.

### DIFF
--- a/clients/web/src/utilities/MiscFunctions.js
+++ b/clients/web/src/utilities/MiscFunctions.js
@@ -226,6 +226,35 @@ girder.restartServer._reloadWindow = function () {
 };
 
 /**
+ * Create a set of flags that can be OR'd (|) together to define a set of
+ * options.
+ *
+ * @param {Array} options An array of strings defining the option names.
+ * @param {string} allOption If you want an option that enables all options,
+ *                 pass its name as this parameter.
+ * @return {Object} An object mapping the names of options to values.
+ */
+girder.defineFlags = function (options, allOption) {
+    var i = 0,
+        obj = {};
+
+    if (allOption) {
+        obj[allOption] = 1;
+    }
+    _.each(options, function (opt) {
+        obj[opt] = 1 << i;
+
+        if (allOption) {
+            obj[allOption] |= obj[opt];
+        }
+
+        i += 1;
+    });
+
+    return obj;
+};
+
+/**
  * Transform markdown into HTML and render it into the given element. If no
  * element is provided, simply returns the HTML.
  *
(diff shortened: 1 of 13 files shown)


mgrauer:
@zachmullen I've started using this in [Minerva](https://github.com/OpenGeoscience/minerva/commit/22e739e123a321ae50ef1d9bc538d2cc5828aae1).

Thanks for such a quick creation!

A few questions/requests:

  * can you make the pagination widget optional?
  * setting the desired columns from outside the widget is awkward, is there some better way that I don't know about?
  * can you make a setting where the job detail appears in the girder modal dialog?
  * I'm getting the job update notification events, but it isn't adding anything to my jobs list for new jobs.  Is there something else I need to do to?  The `tr.length` is 0 for me in `_statusChange`.


screen shot 2015-08-12 at 7 40 33 pm

zachmullen:
> setting the desired columns from outside the widget is awkward, is there some better way that I don't know about?

Yeah, when I was experimenting with this I just changed the context:

```
with (girder.views.jobs_JobListWidget.prototype.columnEnum) {
    var cols = COLUMN_STATUS_ICON | COLUMN_TITLE;
}

new girder.views.jobs_JobListWidget({
    columns: cols,
    ...
});
```

> I'm getting the job update notification events, but it isn't adding anything to my jobs list for new jobs. Is there something else I need to do to? The tr.length is 0 for me in _statusChange.

Currently this is expected; it's only updating jobs that were already rendered. There are two ways we could achieve what you want, the easiest of which is to simply call

    jobListWidget.collection.fetch(null, true);

as soon as you know the new job has been created on the server.

As for your other requests, yes, those can certainly be done.

zachmullen:
@mgrauer The most recent commit will help you achieve what you want, namely, pass the following options to your widget:

    triggerJobClick: true

This will override the default behavior of navigating to the job details route, and instead just trigger a `g:jobClicked` event from the widget with the job model as the first argument. You can bind to that event and show the job details in a modal dialog.

    showPaging: false

This will cause the pagination widget at the bottom not to be displayed.

zachmullen:
This is now officially ready for your QA @mgrauer

mgrauer:
This works in Girder, and I can use it in Minerva.  

@kotfic noticed that on the job details page, very long longs will make that logs element huge.  Perhaps some scroll overflow there?  Although that isn't even on this PR.

LGTM.  :japanese_goblin:

zachmullen:
> Perhaps some scroll overflow there?

Good idea, feel free to open an issue for that.

Issue #916, opened 2015-08-11
vagrant up fails on npm install during provisining


chapmanbe:
Hi,

I just forked and cloned the Girder repository on github. When I try to bring up the Girder vagrantbox, it fails on the NPM install:
~~~~
TASK: [run npm install] ******************************************************* 
<127.0.0.1> ESTABLISH CONNECTION FOR USER: vagrant
<127.0.0.1> REMOTE_MODULE npm path=/vagrant
<127.0.0.1> EXEC ssh -C -tt -vvv -o UserKnownHostsFile=/dev/null -o ControlMaster=auto -o ControlPersist=60s -o ControlPath="/Users/brian/.ansible/cp/ansible-ssh-%h-%p-%r" -o StrictHostKeyChecking=no -o Port=2222 -o IdentityFile="/Users/brian/Code/girder/.vagrant/machines/default/virtualbox/private_key" -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=vagrant -o ConnectTimeout=10 127.0.0.1 /bin/sh -c 'mkdir -p $HOME/.ansible/tmp/ansible-tmp-1439320042.19-167413397287347 && chmod a+rx $HOME/.ansible/tmp/ansible-tmp-1439320042.19-167413397287347 && echo $HOME/.ansible/tmp/ansible-tmp-1439320042.19-167413397287347'
<127.0.0.1> PUT /var/folders/bn/hc7234pd2cd0gxz98dn0zq_40000gn/T/tmpyO2q46 TO /home/vagrant/.ansible/tmp/ansible-tmp-1439320042.19-167413397287347/npm
[...]

PR #919, opened 2015-08-13
Refs #916. Fixes for ansible/vagrant install

zachmullen:
Service restart handlers require `sudo`


danlamanna:
LGTM.

This took care of the restarting errors at the end of an otherwise successful ```vagrant up```.

PR #921, opened 2015-08-19
Remove double slash from file download URL in py client

zachmullen:


zachmullen:
I plan on cutting a girder_client release after this drops since some romanesco work I'm currently doing is using the pip installed version of girder_client.

mgrauer:
What was this breaking?  I'm mainly interested in why our tests weren't catching this if it was an issue.

zachmullen:
I'm using httmock to mock out girder in romanesco tests, and when comparing URLs I didn't want to have to check against "/api/v1//file/download". It doesn't really break anything that I know of, cherrypy handles it just fine.

mgrauer:
LGTM

PR #922, opened 2015-08-19
Python client v1.0.1

zachmullen:


zachmullen:
https://pypi.python.org/pypi/girder-client/1.0.1

PR #924, opened 2015-08-20
Use the env python instead of the default system python. This closes #896

### LINKED ISSUE #896: Call to python in Gruntfile.js should use env
I've noticed while running girder on DARPA memex for our Image Space application that when I use my locally installed python, it won't get picked up and config_parse fails because the system python isn't compatible with the syntax used. I think this is b/c the system python is pre 2.6 but my local python is 2.7+. In order to get it to work with my local python I keep having to change the line where it calls python in the Gruntfile.js to use `env python`. I think this is actually more correct since normally npm is installed at the system level and so that's why it picks up the system python instead of the one from my local environment. If you all agree, I'll submit a quick PR for this.

chrismattmann:


zachmullen:
@chrismattmann how does this change the behavior? I think I am missing some nuance in the behavior of `env` that I am not aware of...

chrismattmann:
hi @zachmullen - well the main way it changes the behavior is to pick up my own locally built python, rather than using the system python. The use case is described on #896 - basically I download girder on a CentOS box whose default python is still 2.6 (I have my own python 2.7 Collective.python built locally). I run Girder, it fails. That's b/c it picks up the 2.6 system python; and not the 2.7 one I have locally. /usr/bin/env fixes that, and causes it to pick up the python2.7 in my own local environment.

jbeezley:
I'm curious about this as well.  My understanding is that `env python` will run the python found first in `PATH` the same as just `python`.  I know people use `#!/usr/bin/env python` because the shebang expects a full path.  I don't see how this change effects anything.

Either way, this looks innocuous enough to me so I don't see a problem with it.

zachmullen:
The only downside I could see is that I think this will break on Windows (which we don't *really* support, but we typically try not to intentionally break Windows support if we can help it).

@chrismattmann since the behavior you're describing is somewhat unexpected, could you humor me and show the output of the following command when run in your system?

    env > e1.txt && env env > e2.txt && md5 e1.txt e2.txt

zachmullen:
(Probably need to change `md5` to `md5sum` if you are on linux.)

chrismattmann:
@zachmullen np! see:

```
[nc-128-149-250-200:~/tmp/foo] mattmann% env > e1.txt && env env > e2.txt && md5 e1.txt e2.txt
MD5 (e1.txt) = d460511213b639e81e9d331176ffdcb6
MD5 (e2.txt) = d460511213b639e81e9d331176ffdcb6
[nc-128-149-250-200:~/tmp/foo] mattmann% 
```
One thing to note. This fixed the issue I was having on my system, and I think as mentioned by @jbeezley that it's innocuous enough to be committed (though as you mentioned it will screw up on Windows). I'd be happy to test if it's *not* Windows, and then call my patch.

zachmullen:
Well, as expected the env command is not altering your environment in any way... I am baffled as to what could be going on here, but I will merge this since it fixes your issue.

PR #929, opened 2015-08-26
Small utility for getting info from the current URL

zachmullen:
This is useful for plugins that need to send jobs to romanesco that
use girder IO mode, since they have to specify the girder location
via multiple fields.


cjh1:
LGTM, I am currently extracting out what I need myself ...

PR #931, opened 2015-08-28
Allow FilesystemAssetstores to work on filesystems without POSIX permissions

brianhelba:
A CIFS-mounted filesystem does not support setting of POSIX permission bits. An NTFS filesystem may also have similar issues.

### DIFF
--- a/girder/utility/filesystem_assetstore_adapter.py
+++ b/girder/utility/filesystem_assetstore_adapter.py
@@ -196,8 +196,13 @@ class FilesystemAssetstoreAdapter(AbstractAssetstoreAdapter):
             os.remove(upload['tempFile'])
         else:
             # Move the temp file to permanent location in the assetstore.
+            # shutil.move works across filesystems
             shutil.move(upload['tempFile'], abspath)
-            os.chmod(abspath, stat.S_IRUSR | stat.S_IWUSR)
+            try:
+                os.chmod(abspath, stat.S_IRUSR | stat.S_IWUSR)
+            except OSError:
+                # some filesystems may not support POSIX permissions
+                pass
 
         file['sha512'] = hash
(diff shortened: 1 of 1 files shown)


zachmullen:
My understanding was that chmod would simply be ignored on platforms that don't support it, rather than throwing an exception. Do you have the output of the exception?

brianhelba:
From a filesystem mounted via CIFS:

[<redacted_date>] ERROR: 500 Error
  File "girder/api/rest.py", line 295, in endpointDecorator
    val = fun(self, args, kwargs)
  File "girder/api/rest.py", line 795, in POST
    return self.handleRoute(method, path, params)
  File "girder/api/rest.py", line 583, in handleRoute
    val = handler(**kwargs)
  File "girder/api/access.py", line 50, in accessDecorator
    return fun(*args, **kwargs)
  File "girder/api/v1/file.py", line 185, in readChunk
    return self.model('upload').handleChunk(upload, chunk.file)
  File "girder/models/upload.py", line 64, in handleChunk
    return self.finalizeUpload(upload, assetstore)
  File "girder/models/upload.py", line 121, in finalizeUpload
    file = adapter.finalizeUpload(upload, file)
  File "girder/utility/filesystem_assetstore_adapter.py", line 200, in finalizeUpload
    os.chmod(abspath, stat.S_IRUSR | stat.S_IWUSR)
OSError: [Errno 1] Operation not permitted: '<redacted_assetstore_root>/0c/53/0c535b4a9ff1befe280cc301ef69b7ce1fe9699fb43d51fcefa2e28f14c272a846e03345f1770b57b785577ab85039ca518397c38429fea08cd922fabbc5182e'

zachmullen:
Thanks. LGTM

PR #932, opened 2015-08-28
Manipulate Model._filterKeys using in-place set operations

brianhelba:
Using in-place set operations prevents object references from breaking and is slightly more more efficient.

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -96,7 +96,7 @@ class Model(ModelImporter):
         if isinstance(fields, six.string_types):
             fields = (fields, )
 
-        self._filterKeys[level] = self._filterKeys[level].union(fields)
+        self._filterKeys[level].update(fields)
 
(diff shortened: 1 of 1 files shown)


girder/models/model_base.py
```diff
@@ -134,15 +134,15 @@ def filter(self, doc, user=None, additionalKeys=None):
         if doc is None:
             return None
 
-        keys = self._filterKeys[AccessType.READ]
+        keys = tuple(self._filterKeys[AccessType.READ])
```
zachmullen: `tuple` is an immutable type, I think this will break...

brianhelba: @zachmullen Indeed, this should have been the set constructor. I'll fix.

brianhelba:
Fixed problem and rebased.

zachmullen:
LGTM

PR #933, opened 2015-08-28
Sign Content-Length header for S3 direct uploads

### LINKED ISSUE #928: Investigate v4 signature for S3 assetstore
http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html

See `boto.s3.connection.Connection.generate_url_sigv4`. This new signature method looks like it supports signing all HTTP headers, including Content-Length, which could get around the longstanding issue we have with unverified content length uploads when going direct to S3 from the user agent.

zachmullen:
This is now possible using AWS V4 auth signatures, though boto
does not (as of yet) support presigning easily via its API. This
will hopefully change in future version of boto.


zachmullen:
This now works for normal and multipart uploads, however it broke all our s3-related tests. Will take a bit of digging to figure out a fix to our mock_s3 stuff, unless @manthey knows of a quick fix :)

manthey:
@zachmullen Was this tested on an actual S3 instance with single-part and multi-part uploads?

zachmullen:
Yes, though it's been a while so it's worth verifying again. Let me know if you need some credentials for our bucket to test with.

manthey:
@zachmullen PTAL.

zachmullen:
@manthey You da man. LGTM

PR #936, opened 2015-09-03
414 on resource download

zachmullen:


cpatrick:
Boo to breaking purity, but yay for fixing bugs.

LGTM :shipit:

cpatrick:
(assuming travis is happy)

PR #937, opened 2015-09-03
Fixes #674. Allow admins to change other users' passwords

### LINKED ISSUE #674: Site admins have access to change password UI tab for other users
In the web clent, site admins can edit the user details of other users. However, when on the details page of another user, the "Password" tab is still visible. Using the change password functionality in this tab is broken, as all requests are sent to:
/api/v1/user/password
which assumes the password being changed is for the current user.

The simple fix is to disable showing the "Password" tab for other users in the web client.

However, allowing site admins to change / reset the password of other users is a nice functionality, but that seems like a seperate new feature.

zachmullen:


mgrauer:
@zachmullen Can you add a client side test?

zachmullen:
@mgrauer PTAL

mgrauer:
LGTM.

brianhelba:
I notice that the template logic on the client is such that an admin user is still required to enter the current password when changing their own password. This is of course only a client-side check (admin users can use the new API call to unconditionally change their password).

I don't know if this was intentional, but I just wanted to comment that I see it as a hidden feature. Obviously, if an admin user leaves their browser unattended, we can't be responsible for the havoc that could be wrought, but at least one extra obstacle that prevents the admin user from getting easily locked out of their own account might be positive. If they need to change their password without knowing it, they will have to use the reset password email functionality.

@zachmullen Was this intentional? If not, do you think it's a positive feature?

zachmullen:
It was intentional, though I hadn't thought through all the scenarios you mentioned. I think it is fine like this but I don't have strong feelings either way.

PR #938, opened 2015-09-03
If we have a JSON body add it to assert message

cjh1:
If assertStatus fails add any JSON response body to the message.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -195,16 +195,17 @@ class TestCase(unittest.TestCase, model_importer.ModelImporter):
         :type code: int or str
         """
         code = str(code)
-        msg = 'Response status was %s, not %s.' % (response.output_status, code)
 
-        if hasattr(response, 'json'):
-            msg += ' Response body was:\n%s' % json.dumps(response.json,
-                                                          sort_keys=True,
-                                                          indent=4,
-                                                          separators=(',',
-                                                                      ': '))
+        if not response.output_status.startswith(code.encode()):
+            msg = 'Response status was %s, not %s.' % (response.output_status,
+                                                       code)
 
-        self.assertTrue(response.output_status.startswith(code.encode()), msg)
+            if hasattr(response, 'json'):
+                msg += ' Response body was:\n%s' % json.dumps(
+                    response.json, sort_keys=True, indent=4,
+                    separators=(',', ': '))
+
+            self.fail(msg)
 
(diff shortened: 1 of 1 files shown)


tests/base.py
```diff
@@ -196,6 +196,10 @@ def assertStatus(self, response, code):
         """
         code = str(code)
         msg = 'Response status was %s, not %s.' % (response.output_status, code)
+
+        if response.json:
+            msg += ' Response body was %s.' % json.dumps(response.json)
```
cpatrick: @cjh1 I like this. We could go even prettier, though. `json.dumps(response.json, sort_keys=True, indent=4, separators=(',', ': '))`. :dancer:

cjh1: I am happy to make it easier for those that have trouble reading unformatted JSON :smirk_cat: 
Do you think we need a newline before printing the JSON in that case? Something like:
'''
' Response body was:\n%s.' % json.dumps(response.json, sort_keys=True, indent=4, separators=(',', ': '))
'''

cpatrick: Works for me.

zachmullen:
@cjh1 we call `assertStatus` quite a bit, but I'd rather not do all this sorting and pretty printing computation of potentially large objects unconditionally like this. Could we optimize this slightly by only computing this error message after we've tested the status and found it to be invalid? Then instead of doing the `assertTrue` we can just call `self.fail(msg)` inside that condition block.

cjh1:
Sure

zachmullen:
Also, since it looked like you were getting a little squished there: PEP8 allows expressions inside parentheses to be indented on the next line, i.e.

```
msg += ' Response body was:\n%s' % json.dumps(
    response.json, sort_keys=True, indent=4, separators=(',', ': '))
```

zachmullen:
LGTM, thanks. Just verified that this works and looks great:

```
======================================================================
FAIL: testTemporaryPassword (tests.cases.user_test.UserTestCase)
----------------------------------------------------------------------
  File "tests/cases/user_test.py", line 477, in testTemporaryPassword
    self.assertStatus(resp, 403)
  File "tests/base.py", line 208, in assertStatus
    self.fail(msg)
AssertionError: Response status was 200 OK, not 403. Response body was:
{
    "message": "Sent temporary access email."
}

----------------------------------------------------------------------
```

PR #939, opened 2015-09-04
Use strict comparison in Javascript code

brianhelba:
Coercive comparison is almost always not the desired behavior, and can be a source of bugs.

Regex: 's/([^=!])==([^=])/\1===\2/g'
       's/!=([^=])/!==\1/g'

### DIFF
--- a/clients/web/src/templates/body/assetstores.jade
+++ b/clients/web/src/templates/body/assetstores.jade
@@ -18,14 +18,14 @@
             if assetstore.get('current')
               span.g-assetstore-current-text
                 |  (Current assetstore)
-          if assetstore.get('type') == types.FILESYSTEM
+          if assetstore.get('type') === types.FILESYSTEM
             div
               b Type:
               span.g-assetstore-type  Filesystem
             div
               b Root path:
               span.g-assetstore-root  #{assetstore.get('root')}
-          else if assetstore.get('type') == types.GRIDFS
+          else if assetstore.get('type') === types.GRIDFS
             div
               b Type:
               span.g-assetstore-type  GridFS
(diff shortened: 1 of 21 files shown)


jeffbaumes:
I tend to agree with @brianhelba. Is there any style/content checking for `===`? Otherwise I can see `==` creeping back in.

cpatrick:
I'm surprised this isn't in our javascript style checks... Oh, okay. Now that I look at it, these are jade and test specs. Can we check that?

jbeezley:
Yeah, it's in our [jshint config](https://github.com/girder/girder/blob/master/tests/jshint.cfg#L32).  Testing the js embedded inside jade would be more difficult though.

jeffbaumes:
Can we add a test that performs @brianhelba's regex replace on these other files and fail if it is anything other than a no-op? I don't want to try to replicate jslint but for something easy like this it seems tractable.

danlamanna:
[jade-lint](https://github.com/benedfit/jade-lint) doesn't seem to support it yet but seems very active. Might be something to incorporate in the future? #lintAllOfTheThings

zachmullen:
Definitely looks promising. https://github.com/benedfit/jade-lint/issues/35

jbeezley:
Hmmm... I wonder if it would be crazy to allow passing javascript to different linter somehow.

brianhelba:
Amended my commit with 3 extra line changes that I missed before.

If anyone wants to augment this PR with an extra commit on top of this one with a style check / linter for template / test spec files, you're certainly welcome to.

brianhelba:
Let's followup discussion over in https://github.com/girder/girder/issues/944 .

Issue #910, opened 2015-08-06
Support connecting to SMTP servers with authentication / encryption


brianhelba:
Currently, the settings allow a custom "SMTP host" to be specified, but do not have options for authentication and encryption (SSL/STARTTLS). Adding support for these features is critical if Girder is used with any non-localhost SMTP relays.

brianhelba:
@girder/developers I have funding for work on this issue; please contact me if one of you has time to fix this.

brianhelba:
Fixed by https://github.com/girder/girder/pull/940

PR #940, opened 2015-09-04
Support connecting to SMTP servers with authentication / encryption

brianhelba:
Still needs testing, but this is difficult, since there's no standard Python library to mock an authenticating / encrypted SMTP server. We can either forgo full coverage for this functionality, or find some other way to mock.

Solves #910.


cjh1:
You could use the mock package to mock out the calls smtplib.SMTP etc. then you won't need a mock SMTP server.

clients/web/src/templates/body/systemConfiguration.jade
```diff
@@ -46,7 +74,7 @@ form.g-settings-form(role="form")
       option(value="yesmod", selected=(settings['core.add_to_group_policy'] ==
              "yesmod")) Yes by group administrators and moderators unless disabled per group
 
-  .g-collection-create-policy-container
+  .g-settings-form-container
```
zachmullen: I think this is what's breaking the test, what is the reason for this change?

zachmullen:
:+1: for what @cjh1 suggests.

clients/web/src/templates/body/systemConfiguration.jade
zachmullen: Ah, that's cool, it should work if you keep the old class also. Classes are, to some degree, part of our API surface.

brianhelba: @zachmullen Since this class name is so specific to the one place that it was used before, I'd still like to copy the content to a new "g-settings-form-container", and provide a deprecated alias (with @extend) named "g-collection-create-policy-container". That should encourage more reuse of this class as a useful grouping mechanism while maintaining compatibility.

brianhelba:
Separated out the more controversial first commit for immediate feedback in https://github.com/girder/girder/pull/951 . I'll continue to work on the testing in this PR.

brianhelba:
I'm not going to have time to finish this for the next 2 weeks. If anyone wants to finish this (and the associated PR https://github.com/girder/girder/pull/951 ), please feel free.

girder/utility/mail_utils.py
```diff
@@ -117,15 +117,48 @@ def addTemplateDirectory(dir, prepend=False):
...
+                self.connection.starttls()
+        if self.username and self.password:
+            self.connection.login(self.username, self.password)
+
+    def send(self, fromAddress, toAddresses, message):
+        self.connection.sendmail(fromAddress, toAddresses, message)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.quit()
+
```
zachmullen: I think we should make this in the public API and make `__enter__` return `self`.

clients/web/src/templates/body/systemConfiguration.jade
```diff
@@ -7,12 +7,40 @@ form.g-settings-form(role="form")
...
       placeholder="Default: #{defaults['core.cookie_lifetime'] || ''}",
       title="How long users will stay logged in.")
-  .form-group
-    label(for="g-core-smtp-host") SMTP host
-    input#g-core-smtp-host.form-control.input-sm(
-      type="text", value="#{settings['core.smtp_host'] || ''}",
-      placeholder="Default: #{defaults['core.smtp_host'] || ''}",
-      title="The address of the SMTP server used to send emails.")
+  .g-settings-form-container
+    // TODO: add proper styling to this header
```
brianhelba: Remove this.

brianhelba:
After speaking with @zachmullen , we both agree that manual testing would be sufficient. I've deployed this branch to a live instance, and successfully tested authentication with both SSL and STARTTLS against an established third-party SMTP relay (Mailgun).

I think this is now ready to merge. PTAL @girder/developers

girder/constants.py
```diff
@@ -128,6 +128,10 @@ class SettingKey:
     EMAIL_HOST = 'core.email_host'
     REGISTRATION_POLICY = 'core.registration_policy'
     SMTP_HOST = 'core.smtp_host'
+    SMTP_PORT = 'core.smtp.port'
```
zachmullen: If someone had an existing setting for SMTP_HOST that contained the port, would this change break email for them?

brianhelba: It will work fine. ``None`` will get passed through as the port, and the same [internal logic](https://github.com/python/cpython/blob/2.7/Lib/smtplib.py#L304) will take over and split the host on ``:`` to get the port.

brianhelba: To be certain, I just tested it, and it's fine.

zachmullen:
LGTM

PR #941, opened 2015-09-04
Lookup resource by name

zachmullen:
For once, I actually managed to keep my commits logically isolated.

* **1st commit**: adds new search modes for items and folders allowing them to be looked up directly by name and parent ID, which makes it much easier to externally query for the existence of a resource at a given path when the parent ID is known.

* **2nd commit**: some refactoring to reduce code duplication in our swagger descriptions.

* **3rd commit**: move some private methods in girder client into the public API and tweak them a little for consistency. These use the new lookup mode. Also added tests that exercise all this new functionality.

### DIFF
--- a/girder/api/v1/folder.py
+++ b/girder/api/v1/folder.py
@@ -77,10 +77,12 @@ class Folder(Resource):
                 exc=True)
 
             filters = {}
-            if 'text' in params:
+            if params.get('text'):
                 filters['$text'] = {
                     '$search': params['text']
                 }
+            if params.get('name'):
+                filters['name'] = params['name']
 
             return [self.model('folder').filter(folder, user) for folder in
                     self.model('folder').childFolders(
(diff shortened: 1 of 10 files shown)


girder/api/v1/folder.py
```diff
@@ -77,10 +77,12 @@ def find(self, params):
                 exc=True)
```
brianhelba: You should make a note in the docstring on line 54 about name-match search.

girder/api/v1/item.py
```diff
@@ -68,10 +68,13 @@ def find(self, params):
             folder = self.model('folder').load(id=params['folderId'], user=user,
```
brianhelba: You should make a note in the docstring on line 51 about name-match search.

brianhelba:
LGTM other than maybe adding some extra internal documentation.

+1 on the change making those Girder client methods public.

zachmullen:
> +1 on the change making those Girder client methods public.

This was actually the main purpose of this branch; the TCGA ingest script currently depends on this functionality. I'll merge once travis is happy.

PR #942, opened 2015-09-04
Fixes #670. Add utility for direct upload from file-like object

### LINKED ISSUE #670: Allow "internal" uploads directly from the local filesystem or a file-like object
The Upload model provides a great abstraction around the File creation process. If a developer has some file that's already on the local filesystem of a Girder instance, and they want to import it into Girder via server code, they have to:
* "open()" the file
* ".read()" the contents of the file into "Upload.handleChunk"
* internally, the file will be written back to disk in a temporary location
* internally, the file will be moved (or copied again, if cross-filesystem) into the FilesystemAssetstore

This is very inefficient for large files. Also for large files, the developer must remember to call ".read(SIZE)" + "Upload.handleChunk" in a buffer loop, or else memory issues may occur.

It would be a great feature to allow the Upload model (with support from AssetstoreAdapters) to accept a path to a local file or, more generally, a file-like object to efficently create a new File.

zachmullen:
@brianhelba PTAL

### DIFF
--- a/docs/api-docs.rst
+++ b/docs/api-docs.rst
@@ -117,6 +117,12 @@ Setting
 .. automodule:: girder.models.setting
    :members:
 
+Upload
+^^^^^^
+
+.. automodule:: girder.models.upload
+   :members:
+
 Python API for RESTful web API
 ------------------------------
 
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -18,6 +18,7 @@
 ###############################################################################
 
 import datetime
+import six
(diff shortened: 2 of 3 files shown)


brianhelba:
LGTM.

We could make a note that StringIO also qualifies as a "file-like" object, but most Python devs *should* already know that.

brianhelba:
This made me realize a few things that are outside the scope of this PR. I put them in https://github.com/girder/girder/issues/943 .

PR #945, opened 2015-09-05
Create User documents with an empty "groups" property

brianhelba:
This makes User documents more consistent and can simplify a number of operations in the Group model.

For now, the Group model still can't assume that a User has a "groups" property, since existing databases aren't migrated.


zachmullen:
LGTM

PR #950, opened 2015-09-07
Change all dictionary accesses to use view objects

brianhelba:
View objects are the only way to access dictionary objects in Python 3, so calling six.iterkeys/itervalues/iteritems on code run in Python 3 will wrap the result in an unnecessary iterator. Calling dict.keys/values/items on code run in Python 2 will create lists, which is inefficent for large dicts.

six.viewkeys/viewvalues/viewitems will be elided away completely for Python 3 and use an efficent data structure in Python 2.

Also includes some minor style cleanup in 2 subsequent commits.

### DIFF
--- a/girder/api/describe.py
+++ b/girder/api/describe.py
@@ -342,7 +342,7 @@ class Describe(Resource):
                 'operations': sorted(
                     op, key=functools.cmp_to_key(self._compareOperations))
                 } for route, op in sorted(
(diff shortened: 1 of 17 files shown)


brianhelba:
@girder/developers This is ready for review please.

zachmullen:
I looked at this earlier but forgot to comment. LGTM.

PR #951, opened 2015-09-07
Deprecate the ".g-collection-create-policy-container" CSS class

brianhelba:
Rename this class into a more general ".g-search-container", which can be useful for grouping similar settings in any any configuration view.


brianhelba:
Putting this change into a separate PR to get immediate feedback.

Let me know if this is a good balance of compatibility and future usability.

zachmullen:
I think moving the style rules to a more reusable class is a good idea, but I think the best answer is to simply put both classes on the element so we don't break selection of the existing element.

brianhelba:
@zachmullen Is the compatibility concern for people using the old class as a selector, people using the old class as a source of styles, or both?

If it's only one or the other, the fix is easy. To address both though, if we "@extend" the old class to have the same CSS class as the new class, then add both classes to the element, we'll repeat the addition of these same styles to the element twice. Is that a problem?

zachmullen:
I was only concerned with the selection part, but it is certainly possible someone would reuse some of our classes just for the styling. It wouldn't be a problem to have two classes provide the same style rules onto the same element, it's a logical no-op.

brianhelba:
I'm not going to have time to finish this for the next 2 weeks. If anyone wants to finish this (and the associated PR https://github.com/girder/girder/pull/940 ), please feel free.

zachmullen:
@brianhelba I implemented my suggestion, we now maintain compatibility for both styling and selection, while also adding your more general-purpose class for styling.

brianhelba:
LGTM.

PR #952, opened 2015-09-07
Simplify API description registry

brianhelba:
This significantly simplifies the registry of API descriptions.

There are 3 commits:

* **Remove useless "models" property from Swagger resource descriptions**
The "models" field in a Swagger resource description should contain a detailed listing of the properties in a model. At the moment, Girder has no mechanism to expose this information (though it might come via the Model._filterKeys attribute). Without this information, there is no reason to include a "models" field in resource descriptions. For more information, see:
https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#52-api-declaration
https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#527-model-object

 The empty "models" variable was first added by commit a54c433 , but with no mechanism to ever
set or use it.

 Later, commit cef7f4e finally started returning the "models" property with the resource description endpoint, but even assuming that the variable was set correctly, the code returns the same thing for each seperate resource type. The "addModel" function isn't tested or used anywhere, and only serves to set values in an already-misused variable.

 It's probably best at this point to consider this to be cruft, and remove it all. Future work could attempt to define Model descriptions via Swagger's specification, but this is nowhere close.

* **Remove unnecessary global variable from route doc setup**
The "discovery" variable holds the resource types that have route documentation. The exact same data can be obtained from the keys in the "routes" variable.

* **Refactor "girder.api.docs", to fix documentation and improve maintainability**
This also makes the code more maintainable, by avoiding repetition of code which must always be exactly the same for "removeRouteDocs" to work properly.

### DIFF
--- a/girder/api/docs.py
+++ b/girder/api/docs.py
@@ -20,7 +20,6 @@
 import collections
 
 
-discovery = set()
 routes = collections.defaultdict(dict)
 
 
@@ -62,7 +61,6 @@ def addRouteDocs(resource, route, method, info, handler):
     if info not in routes[resource][path]:
         routes[resource][path].append(info)
 
-    discovery.add(resource)
 
 
 def removeRouteDocs(resource, route, method, info, handler):
@@ -101,4 +99,4 @@ def removeRouteDocs(resource, route, method, info, handler):
         if not len(routes[resource][path]):
             del routes[resource][path]
             if not len(routes[resource]):
-                discovery.remove(resource)
+                del routes[resource]
--- a/girder/api/docs.py
+++ b/girder/api/docs.py
@@ -33,7 +33,7 @@ def _toRoutePath(resource, route):
     """
     # Convert wildcard tokens from :foo form to {foo} form
     convRoute = [
-        '{{{}}}'.format(token[1:]) if token[0] == ':' else token
+        '{%s}' % token[1:] if token[0] == ':' else token
         for token in route
     ]
     path = '/'.join(['', resource] + convRoute)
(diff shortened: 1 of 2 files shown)


brianhelba:
@zachmullen @cjh1 I'd welcome your review, since you've done most of the work on this code in the past.

girder/api/docs.py
```diff
@@ -18,11 +18,37 @@
...
+
+
+def _toRoutePath(resource, route):
+    """
+    Convert a base resource type and list of route components into a
+    Swagger-compatible route path.
+    """
+    # Convert wildcard tokens from :foo form to {foo} form
+    convRoute = [
+        '{{{}}}'.format(token[1:]) if token[0] == ':' else token
```
zachmullen: Since we're changing this, let's go ahead and change this to ``{%s} % token[1:]``.

brianhelba: @zachmullen Done.

Thanks. I wanted to make this change already, but I wasn't sure of the Girder conventions for using ``.format`` vs ``%``.

zachmullen: We officially prefer `%` since it handles unicode much better, but haven't squashed out all the old uses of `format`.

zachmullen:
LGTM

cjh1:
@brianhelba @zachmullen We use addModel extensively to document out request bodies. What my migration path function has been removed?

cpatrick:
@cjh1 Thanks for reminding us that removing functions breaks things :smile_cat:. I vote we revert that commit.

zachmullen:
Sorry @cjh1 , I will push a PR to revert that deletion.

PR #953, opened 2015-09-07
Simplify CherryPy tools configuration

brianhelba:
CherryPy tools can be enabled by passing any truthy value to a config parameter of "<toolName>.on". Using a Python bool instead of a string is simpler.

See: https://github.com/cherrypy/cherrypy/blob/3.3.0/cherrypy/_cptools.py#L459


zachmullen:
LGTM

PR #954, opened 2015-09-07
Fixes #948. Use proper basePath in swagger docs

### LINKED ISSUE #948: Swagger is displaying an extra "/./" component in example URLs
For example, with a default deployment, after clicking "Try it out!" on the tab for "GET /collection", the following "Request URL" is displayed:
``http://localhost:8080/api/v1/./collection?limit=50&offset=0&sort=name&sortdir=1``
Interestingly, the actual AJAX request is still made to:
``http://localhost:8080/api/v1/collection?limit=50&offset=0&sort=name&sortdir=1``

The source of this problem appears to be the "basePath" property of the individual endpoint API descriptions. As a result of [this commit](https://github.com/girder/girder/commit/237481c6d461e3830c4ce6ddd9e0a56d873e62b0#diff-bacb88e0dc16bf6978cac0125b6b79d3L296), it is now being sent as ``"."``. While this is sufficient to make Swagger functional, the display is clearly wrong.

After a bit of experimentation, I think that setting this field to the absolute API prefix (``"/api/v1"`` in the default deployment) would be best. In addition to fixing the display issue, it also makes the most semantic sense, as now result of ``/api/v1/describe/collection`` contains enough information to fully reconstruct the endpoint.

However, I'm not sure of the best way to make the actual code change, for a few reasons:
* The webroot.Webroot contains an ``apiRoot`` property, but I'm not sure of the cleanest way to pass that object all the way through to the ``Describe`` class.
* [This commit](https://github.com/girder/girder/commit/19bf39056b8dc8aba6649f98a884507667f301c8#diff-f94a7a2e0c5304caf743b4878a5f70fdL84) made the default ``apiRoot`` into a relative path, and Girder should fully support deployments into a non-top level document root anyway. I'm not sure of the current best practice for getting the absolute API root path.

zachmullen:


brianhelba:
LGTM. I've just tested this myself too.

I was originally thinking that we should pass an absolute internal URL here (e.g. ``/api/v1``), but upon re-reading the [Swagger spec](https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#52-api-declaration), it looks like external URLs are best anyway.

PR #956, opened 2015-09-08
Restore girder.api.docs.addModel method

zachmullen:
@cjh1 PTAL

### DIFF
--- a/girder/api/describe.py
+++ b/girder/api/describe.py
@@ -336,6 +336,7 @@ class Describe(Resource):
             'apiVersion': API_VERSION,
             'swaggerVersion': SWAGGER_VERSION,
             'basePath': getApiUrl(),
+            'models': docs.models,
             'apis': [{
                 'path': route,
                 'operations': sorted(
--- a/girder/api/docs.py
+++ b/girder/api/docs.py
@@ -20,7 +20,7 @@
 import collections
 import functools
 
-
+models = {}
 # A dict of dicts of lists
 routes = collections.defaultdict(
     functools.partial(collections.defaultdict, list))
@@ -110,3 +110,15 @@ def removeRouteDocs(resource, route, method, info, handler):
             del routes[resource][path]
             if not routes[resource]:
                 del routes[resource]
+
+
+def addModel(name, model):
+    """
+    This is called to add a model to the swagger documentation.
+
+    :param name: The name of the model.
+    :type name: str
+    :param model: The model to add.
+    :type model: dict
+    """
+    models[name] = model


cjh1:
Thanks, LGTM

brianhelba:
What's the rationale here?

I understand if we need to include a deprecated function stub for comparability, but the functionality as-is looks broken to me. See my comments in https://github.com/girder/girder/commit/0b39ea6434e52a4bbf3ee2a97c726f8226df008a

cjh1:
Despite its short coming this functionality currently works. It allows a model of a request body to be defined. I am happy for it to be removed if I still have away of adding a model for a request body.

zachmullen:
The `responseClass` of a route maps into these models, and it does work. When we commit to doing this properly we can add some more useful methods on top of this raw-level API while still preserving this level of the API for early adopters like @cjh1 :)

brianhelba:
@girder/owners Guys, there is too much code churn here. My first PR https://github.com/girder/girder/pull/952 should not have been merged so fast, especially with my request for @cjh1 's review in the first comment.

This PR should also not have been closed so quickly. Yes, this breaks something, but we're still in the middle of a development branch and I had outstanding questions on the PR.

I'd like to continue discussion over on an issue, so nothing gets closed until all stakeholders are satisfied.

cjh1:
I think if you don't want your PR merged a prefix of [WIP] will help.

brianhelba:
My PR was ready for final review, and ready to be merged if there were no objections. What I'm suggesting is that we wait a bit longer between the submission and merge, to give everyone a chance to chime in.

I was actually hoping that you @cjh1 would have had a chance to comment on my initial PR, and I mentioned you in the first comment asking for reviews. Let's please continue the discussion over at https://github.com/girder/girder/issues/958 , so I can understand how the system is supposed to work.

cpatrick:
@brianhelba @cjh1 @zachmullen This is a great example of the process working! It's great to see these discussions happen prior to a release.

Broadly, if we wait for buy-in from everyone on every issue, development will slow to a crawl. This should serve as a cautionary tail of looking for API breakage on every PR.

brianhelba:
@cpatrick Well, I think there's a happy middle ground between:
* waiting for buy-in from everyone on every issue
* waiting [45 minutes](https://github.com/girder/girder/pull/952) or [9 minutes](https://github.com/girder/girder/pull/956) from the time a PR is first created to when it's merged

Our communication is very good, but with an asynchronous platform like GitHub, even if we didn't ever care about consensus, you need to give people a little time to even look at things and chime in.

This isn't a huge deal, but I think it might be opportunity to improve the process a bit. I see Girder as one of our flagship uses of GitHub as a development model, so getting it Right<sup>TM</sup> benefits other projects too.

jeffbaumes:
I think it all depends on the change. I think if an obvious 15-line change can be QA'd in 9 minutes it should be ok. Our process is currently PR -> at least one LGTM and tests pass -> merge. Not sure how to incorporate "wait at least X minutes/hours" into this process, and I don't think we should require multiple LGTMs. I can definitely see "wait at least X days" from authoring something to a *release*, but not to a merge.

I think the lesson learned here is what @cpatrick stated: PRs should have an eye toward maintaining compatibility and any potential breakage should first be discussed before LGTM.

PR #963, opened 2015-09-08
Child count display

zachmullen:


brianhelba:
Thanks for this. I will review it tonight.

girder/models/folder.py
```diff
@@ -543,6 +543,39 @@ def parentsToRoot(self, folder, curPath=None, user=None, force=False,
             return self.parentsToRoot(curParentObject, curPath, user=user,
                                       force=force)
 
+    def countItems(self, folder):
+        """
+        Returns the number of items within the given folder.
+        """
+        return self.model('item').find({
```
brianhelba: It would be nice to call ``Folder.childItems`` here instead, which would allow virtual dispatch to work more cleanly.

brianhelba:
The server-side code looks good, with the exception of a small issue that I commented on. The client-side runs fine for me locally, but I'm not familiar enough with the web client to definitively review the code.

More broadly, I agree with the conclusion that including the sub-folder counts in the regular folder would hurt performance too much. We could probably add the sub-item count without performance problems, but that's inconsistent. So, a separate detail endpoint is justified to me.

On the client side, I'd like to see this information be more accessible. At the least, the hierarchyWidget should display the sub-folder and sub-item counts (or maybe just the total of the 2), and the total folder size, without having to click a button every time (creation time can stay in the detail list, which is funny, because we get that without having to call the detail endpoint). One extra API call when on a folder listing isn't a big deal. What do you think?

Ideally eventually, I'd also like to see counts for each sub-folder in an itemList. I realize that is a lot of load under our current architecture (potentially 50 extra API calls every time we visit a folder), but it would be a huge boon for usability. That feature is outside the scope of this PR, but I'm hoping we can make it a long-term goal.

zachmullen:
> On the client side, I'd like to see this information be more accessible. At the least, the hierarchyWidget should display the sub-folder and sub-item counts (or maybe just the total of the 2), and the total folder size, without having to click a button every time (creation time can stay in the detail list, which is funny, because we get that without having to call the detail endpoint). One extra API call when on a folder listing isn't a big deal. What do you think?

That sounds good. Where in the hierarchy widget do you think this information should be displayed? I am trying to make sure it still looks nice on mobile etc.

brianhelba:
Sorry, I have no ability to originally synthesize aesthetically appropriate designs. I can only mercilessly criticize bad design.

brianhelba:
@dgutman You first brought this issue to my attention. Where on the a Girder folder page would you like to see the size and number of sub-items/sub-folders in that folder?

dgutman:
I think in the upper right corner of the display bar where the folder name is shown on the left, and depending on where you are in the hierarchy, where the <UP> Arrow is located on the right side....

It should list the number of files and/or folders below it..

zachmullen:
PTAL, it now shows up in the top right:

screen shot 2015-09-14 at 5 55 22 pm

I tried to make them as compact as possible so the breadcrumbs could expand as far as possible without having to wrap.

brianhelba:
In terms of the server-side code and responsiveness to the issue, this LGTM. You'll need someone else to approve the client-side code.

The visuals look great. Nice work.

zachmullen:
@girder/developers can someone take a look at the client-side changes on this branch?

manthey:
If you have a folder with a large number of items (>= 100,000), the css is such that the item count overlaps the folder count.

image

zachmullen:
Good point, I will add a method to render those as "10K" or something, making sure it's never longer than 3 characters.

jeffbaumes:
Good idea @zachmullen

manthey:
Hmm....  I like the fact that I now know I had 118,113 items in my folder.  Can you keep the full number as a tooltip?

manthey:
At least in Chrome, it looks like 5 digits is the available space, so perhaps up to 4 digits plus a 1 character suffix would always be okay.  I think I'd rather see 1234k than 1M.

clients/web/src/views/widgets/FolderInfoWidget.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * This view allows users to see and control access on a resource.
+ */
+girder.views.FolderInfoWidget = girder.View.extend({
+    initialize: function () {
+        this.needToFetch = !this.model.has('nItems') || !this.model.has('nFolders');
+        if (this.needToFetch) {
+            this.model.fetch({extraPath: 'details'}).once('g:fetched.details', function () {
```
manthey: This branch isn't covered in the tests.  Is there a simple way to reach this code path?

manthey:
Adding an item from the folder menu doesn't increment the item count.  Deleting items and folders from the checked menu doesn't alter the item and folder counts.  Moving items and folders from the checked menu doesn't alter the item and folder counts.

zachmullen:
If it makes any difference, the exact count will still show up when you click on the info button (the new shiny blue one).

clients/web/src/views/widgets/FolderInfoWidget.js
zachmullen: Good find. It's possible, but I'd have to probably write a separate unit test for just this widget, which might not be a bad idea. (Right now the details get fetched and cached as soon as the folder is loaded as the current hierarchy widget view.)

zachmullen:
@manthey I think I addressed all the things you mentioned.

manthey:
LGTM.  But now I want this feature on Users and Collections so I know how many folders they have.

zachmullen:
:D can open an issue for that

Issue #433, opened 2014-10-24
Plugin loading path


zachmullen:
I think right now we only look for plugins in a single directory. It would be nice to be able to have an ordered list of dirs that we look in.

mgrauer:
related: allow plugins installed through pip.  a workaround is to symlink the girder/plugins/plugin dir to the pip installed plugin dir.

PR #965, opened 2015-09-09
Allow multiple plugin paths

danlamanna:
Re #433 

There are a [couple](http://stackoverflow.com/questions/335695/lists-in-configparser) of [ways](http://stackoverflow.com/questions/6759016/getting-a-list-from-a-config-file-with-configparser-in-python) to handle lists in config parsers, I just went with comma separation but it should be pretty easy to change if anyone has an issue with it. I wanted to maintain backwards compatibility so no function or config option names changed.

Things that "break":   
- Since the directory name of a plugin is it's unique identifier - ```loadPlugin``` blows up if for example ```vega``` exists in ```plugin_dir_1``` **and** ```plugin_dir_2```.
- ```config_parse.py``` sees the comma delimited string instead of an actual list - but I don't believe this is an actual issue for how it's being used now.

In regards to testing:
I created a specific ```plugin_utilities_test``` to mainly test the new function I introduced (```getPluginParentDir```). I also moved ```test_plugin``` into a separate directory from the rest of the plugins to ensure cross plugin-path dependencies still work.

@mgrauer I think this assists in what you were mentioning in #433, since you can now just add the pip installed path to the ```plugin_directory``` option.


cjh1:
I wonder if colon would be a better separator, as this is the path separator on Unix?

girder/utility/plugin_utilities.py
```diff
@@ -87,6 +87,30 @@ def loadPlugins(plugins, root, appconf, apiRoot=None, curConfig=None):
...
+    :type name: str
+    """
+    parentDirs = set()
+
+    for potentialParentDir in getPluginDir(curConfig):
+        if os.path.exists(os.path.join(potentialParentDir, name)):
+            parentDirs.add(potentialParentDir)
+
+    if len(parentDirs) > 1:
+        raise Exception('Plugin directory %s exists in multiple paths' % name)
```
zachmullen: This shouldn't be an exception, we should just use the one that is found first on the path, which is probably what most people would expect if they are familiar with path searching behavior.

jbeezley: We may want to allow the plugin to exist in multiple places and return the first (or last maybe) directory found.  I.e. something like the precedence of config file settings.

I could also be convinced that is a bad idea though...

girder/utility/plugin_utilities.py
```diff
@@ -148,39 +173,45 @@ def loadPlugin(name, root, appconf, apiRoot=None, curConfig=None):
...
+            try:
+                os.makedirs(pluginDir)
+            except OSError:
+                if not os.path.exists(pluginDir):
+                    print(TerminalColor.warning(
+                        'Could not create plugin directory %s.' % pluginDir))
+
+                    failedPluginDirs.add(pluginDir)
+
+    return pluginsDir - failedPluginDirs
```
jbeezley: My preference would be to return a list here rather than a set.  We might also want to preserve the order provided in the config.

girder/utility/plugin_utilities.py
danlamanna: Looks like you both have the same idea, and that is what I did before I pushed... so let's go with that.

I think first is best, and perhaps changing ```,``` to ```:``` as @cjh1 said.

jbeezley: Of course, @zachmullen beat me by two minutes.

girder/utility/plugin_utilities.py
zachmullen: :+1:

zachmullen:
As written, this contains API breakage that isn't strictly necessary, but may be desirable. Namely, `getPluginDir()` used to return a single path and was used to determine the install location for new plugins. With this change it returns a list of search paths, which is a breaking change, and also pushes the problem of choosing a location to install new plugins onto callers.

An alternative, non-breaking change would be to create a new method `getPluginDirs()` in addition to the old one. The old method could then be used to determine which plugin directory should be the one to write into, via the following lookup chain:

1. Another explicit setting, e.g. `plugin_install_path`.
2. If that isn't set, use the first element from the `plugin_directory` config value.
3. If that isn't set, use the old default value as before.

girder/utility/plugin_utilities.py
```diff
@@ -189,40 +220,42 @@ def findAllPlugins(curConfig=None):
...
+                            plugin))
+                        print(e)
+                        continue
+
+            allPlugins[plugin] = {
+                'name': data.get('name', plugin),
+                'description': data.get('description', ''),
+                'version': data.get('version', ''),
+                'dependencies': set(data.get('dependencies', []))
+            }
```
cjh1: I am sure I am missing something, but why don't we save the plugin directory? Rather than have to rediscover it in getPluginParentDir?

mgrauer:
@danlamanna This assists (or it will anyway) exactly what I needed, thanks!

girder/utility/plugin_utilities.py
danlamanna: I'm not against storing it, but as far as I can tell it's only used in ```loadPlugin``` which takes a name. The path would have to be discovered each time anyway.

danlamanna:
Working on fixing the Py3 json error, in the meantime the rest of the code is ready for review.

Is using ```:``` as the path separator a concern for windows systems? It looks like we don't care too much about Windows... but I'm not sure if we should knowingly break it.

girder/utility/plugin_utilities.py
cjh1: I guess my only worry is if we had a lot of directories in our search path and lots of plugins, the load time might get significant, as we have todo an extra search in the plugin path for each plugin?

danlamanna:
Rest assured @zachmullen - it was a bug in my code :).

Now this is ready for review.

girder/utility/plugin_utilities.py
jbeezley: I think I'm with @cjh1 on this.  There is a lot of touching the file system here.  I would memoize the output.

girder/utility/plugin_utilities.py
```diff
@@ -88,6 +88,22 @@ def loadPlugins(plugins, root, appconf, apiRoot=None, curConfig=None):
     return root, appconf, apiRoot
 
 
+def getPluginParentDir(name, curConfig=None):
+    """
+    Finds the directory a plugin lives in and returns it. This throws
+    an exception if it can't find a directory named name in any of the
+    set plugin_directory paths.
+
+    :params name: The name of the plugin (i.e. its directory name)
```
jbeezley: Do we want to make the equivalency here of the plugin name to the plugin directory?  The `plugin.yml` gives a name that doesn't have to be the same as the directory.  I know I have run into confusion in the past when they were named differently.  This may be an issue more generally in girder; we should make sure the usage is consistent.

danlamanna: This is consistent with existing usage, at least as far as ```plugin_utilities.py``` goes. I'm assuming the name property is the "nice name" used for UI purposes only, and the directory name is it's actual identifier to the system.

zachmullen: That's right. There is the human readable name from the plugin.yml file, but the authoritative name/id of the plugin is its directory name.

jbeezley: Okay.  I have concerns about that, but it is probably best addressed in a different issue.

girder/utility/plugin_utilities.py
danlamanna: Would having general purpose memoize functionality in core, such as [this](https://wiki.python.org/moin/PythonDecoratorLibrary#Memoize) assist in other places? Or should I just one off this instance?

danlamanna: I think @cjh1 and @jbeezley are discussing 2 separate issues, let me know if I made any wrong assumptions.

@jbeezley is simply saying that the entire function call should be memoized because it hits the filesystem a lot. I agree, seeing as how it's called in other places and even part of an API call. The only issue that arises here is ```curConfig``` is a dict, and so caching the results given certain arguments is a little bit more difficult (not sure if we want to start dealing with frozensets and so on). We could always cache it for the ```None``` case which I assume is dominant based on the rest of core's code.

@cjh1 Correct me if I'm wrong, but I think you're saying ```allPlugins[plugin]``` should contain a ```pluginParentDir``` key? If that's the case, I'm simply saying that everywhere ```allPlugins``` is used, it's handled by referring to the name of the plugin, which means it wouldn't yield any benefit for ```loadPlugin``` or anything else. That *could* be beneficial if we reworked ```loadPlugin``` to be able to handle the plugin dictionary, which might be something that @zachmullen considered?

jbeezley: @danlamanna pointed out that this method takes as an argument the cherrypy config (or uses the current config) and isn't easily memoized.  In addition, this is probably a textbook case of premature optimization.  With that in mind, I think we should move ahead with this PR without addressing it.

I don't think there was anything else holding this back so I'll pass it @cjh1 and @zachmullen to weigh in.

cjh1: I haven't got a problem with this going in without the optimization.

zachmullen: > In addition, this is probably a textbook case of premature optimization.

+1

danlamanna:
@brianhelba If you want to take a look.

zachmullen:
LGTM

PR #966, opened 2015-09-09
Make api.docs.addModel require a resource type

### LINKED ISSUE #958: girder.api.docs.addModel functionality
I'm still not understanding how girder.api.docs.addModel is supposed to work, given the way that it's implemented and the way that Swagger defines things. Some history of the variable and function is included in this commit message: https://github.com/girder/girder/commit/0b39ea6434e52a4bbf3ee2a97c726f8226df008a .

[Swagger defines the ``models`` field](https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#52-api-declaration) as:
> A list of the models available to this resource. Note that these need to be exposed separately for each API Declaration.

See the [Petstore example](http://petstore.swagger.io/) for a demo implementation that uses models. Note that while a resource can expose multiple models, you don't get every possible model on every possible resource.

The way the code is now, there is a single "models" dict, which gets returned in its entirety for every resource description. At the very least, [the line here](https://github.com/girder/girder/blob/109662e3e9cd06bf4eecc29ac554c6c3ede15f32/girder/api/describe.py#L338) should be something like ``'models': docs.models[resource],`` so that you don't get the exact same model description for each separate resource type.

For a nice clean high-level API to set models, we should build the [model objects](https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#527-model-object) up from individual parameters. However, even with just the low-level API to set models from raw objects, it looks like we need to have something to discriminate which resource type a model is being applied to. Since the existing API has no way to require that parameter, fixing this issue seems to necessarily require a breaking API change.

I have no problem with exposing this functionality at a low-level at first, but I'd like to get a more sane API figured out now. Some testing would be good too.

Of course, if I'm missing something, please let me know. Code examples from @cjh1 that show how this currently is being used would be helpful.

brianhelba:
Per the Swagger specification, each model should be applied only to a specific resource or set of resources. More rationale is provided in issue #958

This commit changes the undocumented, but public, ``addModel`` function to require each new model to specify a resource type which it will be associated with. The same model may be added to multiple resources via repeated calls to the function. This also updates the API resource description endpoint to only return models which are associated with each given resource type.

Closes #958.

### DIFF
--- a/girder/api/docs.py
+++ b/girder/api/docs.py
@@ -19,6 +19,8 @@
 
 import collections
 import functools
+import six
+from girder.constants import TerminalColor
 
 models = collections.defaultdict(dict)
 # routes is dict of dicts of lists
@@ -112,18 +114,21 @@ def removeRouteDocs(resource, route, method, info, handler):
                 del routes[resource]
 
 
-def addModel(resources, name, model):
+def addModel(name, model, resources=None, silent=False):
     """
     Add a model to the Swagger documentation.
 
     :param resources: The type(s) of resource(s) to add the model to. New
         resource types may be implicitly defined, with the expectation that
-        routes will be added for them at some point.
-    :param resources: str or tuple[str]
-    :param name: The name of the model(s).
+        routes will be added for them at some point. If no resources are
+        passed, the model will be exposed for every resource type
+    :param resources: str or tuple/list[str]
+    :param name: The name of the model.
     :type name: str
     :param model: The model to add.
     :type model: dict
+    :param silent: Set this to True to suppress warnings.
+    :type silent: bool
 
     .. warning:: This is a low-level API which does not validate the format of
        ``model``. See the `Swagger Model documentation`_ for a complete
@@ -137,7 +142,14 @@ def addModel(resources, name, model):
        swagger-spec/blob/d79c205485d702302003d4de2f2c980d1caf10f9/
        versions/1.2.md#527-model-object
     """
-    if isinstance(resources, str):
-        resources = (resources,)
-    for resource in resources:
-        models[resource][name] = model
+    if resources:
+        if isinstance(resources, six.string_types):
+            resources = (resources,)
+        for resource in resources:
+            models[resource][name] = model
+    else:
+        if not silent:
+            print(TerminalColor.warning(
+                'WARNING: adding swagger models without specifying resources '
+                'to bind to is discouraged (%s).' % name))
+        models[None][name] = model
--- a/girder/api/docs.py
+++ b/girder/api/docs.py
@@ -20,8 +20,8 @@
 import collections
 import functools
 
-models = {}
-# A dict of dicts of lists
+models = collections.defaultdict(dict)
+# routes is dict of dicts of lists
 routes = collections.defaultdict(
     functools.partial(collections.defaultdict, list))
 
@@ -112,13 +112,25 @@ def removeRouteDocs(resource, route, method, info, handler):
                 del routes[resource]
 
 
-def addModel(name, model):
+def addModel(resource, name, model):
     """
-    This is called to add a model to the swagger documentation.
+    Add a model to the Swagger documentation.
 
+    :param resource: The type of existing resource to add the model to.
+    :param resource: str
     :param name: The name of the model.
     :type name: str
     :param model: The model to add.
     :type model: dict
+
+    .. warning:: This is a low-level API which does not validate the format of
+       ``model``. See the `Swagger Model documentation`_ for a complete
+       specification of the correct format for ``model``.
+
+    .. _Swagger Model documentation: https://github.com/swagger-api/
+       swagger-spec/blob/d79c205485d702302003d4de2f2c980d1caf10f9/
+       versions/1.2.md#527-model-object
     """
-    models[name] = model
+    if resource not in routes:
+        raise Exception('"%s" is not a known resource type' % resource)
+    models[resource][name] = model
--- a/girder/api/docs.py
+++ b/girder/api/docs.py
@@ -112,13 +112,13 @@ def removeRouteDocs(resource, route, method, info, handler):
                 del routes[resource]
 
 
-def addModel(resource, name, model):
+def addModel(resources, name, model):
     """
     Add a model to the Swagger documentation.
 
-    :param resource: The type of existing resource to add the model to.
-    :param resource: str
-    :param name: The name of the model.
+    :param resources: The type(s) of existing resource(s) to add the model to.
+    :param resources: str or tuple[str]
+    :param name: The name of the model(s).
     :type name: str
     :param model: The model to add.
     :type model: dict
@@ -127,10 +127,19 @@ def addModel(resource, name, model):
        ``model``. See the `Swagger Model documentation`_ for a complete
        specification of the correct format for ``model``.
 
+    .. versionchanged:: The syntax and behavior of this function was modified
+        after v1.3.2. The previous implementation did not include a resources
+        parameter.
+
     .. _Swagger Model documentation: https://github.com/swagger-api/
        swagger-spec/blob/d79c205485d702302003d4de2f2c980d1caf10f9/
        versions/1.2.md#527-model-object
     """
-    if resource not in routes:
-        raise Exception('"%s" is not a known resource type' % resource)
-    models[resource][name] = model
+    if isinstance(resources, str):
+        resources = (resources,)
+    for resource in resources:
+        if resource not in routes:
+            raise Exception('"%s" is not a known resource type' % resource)
+    # any errors should be detected ahead of time, to make the change atomic
+    for resource in resources:
+        models[resource][name] = model
--- a/girder/api/docs.py
+++ b/girder/api/docs.py
@@ -116,7 +116,9 @@ def addModel(resources, name, model):
     """
     Add a model to the Swagger documentation.
 
-    :param resources: The type(s) of existing resource(s) to add the model to.
+    :param resources: The type(s) of resource(s) to add the model to. New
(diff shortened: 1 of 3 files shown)


brianhelba:
@cjh1 Can you review this?

Is it possible to update your own plugin's use of this function to use the new argument? I don't see a way to make addModel not contaminate every resource type and behave predictably without adding a required argument like this.

zachmullen:
@girder/developers please note, for anyone using the `girder.api.docs.addModel` function, this is a breaking change, so if people could weigh in on that it would be great.

girder/api/docs.py
```diff
@@ -112,13 +112,25 @@ def removeRouteDocs(resource, route, method, info, handler):
...
     """
-    This is called to add a model to the swagger documentation.
+    Add a model to the Swagger documentation.
 
+    :param resource: The type of existing resource to add the model to.
+    :param resource: str
     :param name: The name of the model.
     :type name: str
     :param model: The model to add.
     :type model: dict
```
cpatrick: So the way I see this, Brian is more correct in how we _should_ use Swagger for this. I think for now, resource should be the last argument and default to a value that will cause the body of this function to set it for all current and future resources on the models dict.

Does that make sense?

cjh1: This would be nice behavior if you have models that are shared across multiple endpoints, which you often do.

brianhelba: That is a possibility that I considered, but I'm worried that it will cause a lot more long-term pain. This function has no documentation that's built, and it's located in a fairly obscure place. @zachmullen and I both initially thought it was unused. I'm hoping that @cjh1 is the only person who currently uses this function (no surprising since he wrote it), and with his cooperation, we can just slip the API change in under the radar. The longer we wait, and the longer we tolerate the existing behavior, the more likely it is that others will start using the function, and then we're really stuck.

The semantics of a default argument are generally such that it behaves sensibly and sanely by default, while allowing you to optionally specify additional customization. An implementation that adds a model to every current resource is probably broken behavior (given Girder's semantics), and adding it to every future-defined resource is...yikes. I think having this behavior be the default is asking for trouble.

If the consensus is that we really can't slip in this API change, I'd propose creating a new function with the proper arguments (unfortunately, the best, most appropriate name is already taken), while slapping deprecation warnings all over this existing one. Then the behavior of the existing one can do whatever we want (including preserving the current behavior), but it's unlikely that new developers will start to rely on it. We can then safely remove the existing one in Girder 2.0 without any guilt.

brianhelba: @cjh1 I would have no problem with a convenience method to add the same model to multiple resources at once, though it's not hard to just call the function a few times, since the actual model dict will be only be shallow-copied.

I can't think of a case where you'd add a model to every single Girder resource, including the built-in ones (what will an Assetstore, Group, and Notification have in common?). I am open to a solid use case to change my mind here.

cjh1: I don't have a problem making the changes if @cpatrick is OK with the change. I think a convenience method might be nice and if you have may ends point you are going to have to make a lot of calls. I am also working on added a test case so no one will try to remove this in the future :smile_cat:, I will push to the tip of your branch.

brianhelba: +1 to @cjh1 's solution, if @cpatrick and the other @girder/owners can tolerate this minor breach of API comparability orthodoxy.

We don't even need a separate method to do the multi-resource thingy. The ``resource`` argument can just accept a ``str`` or `tuple of str` and behave appropriately. I'll push that change now.

cpatrick: @brianhelba @cjh1 @zachmullen If you're in agreement about slipping this change in (throwing semver to the wind!), I'm fine with it. But only if you add a detailed note about the behavior change in the docstring.

zachmullen: I'm totally on the fence, so I leave the decision to @cjh1.

brianhelba:
Pushed another commit to address the changes we discussed.

girder/api/docs.py
cjh1: So there is a bit of a problem with this change ...

Given the following test class, which is our current pattern for adding documentation. The addModel(...) will be called before route(...) is called that adds the resource to routes, so we get an exception say the resource type is not known .... We could move the addModel(...) call to the constructor but that feels a little icky to me.

```python

class ModelResource(Resource):
    def __init__(self):
        self.resourceName = 'model'
        self.route('POST', (), self.hasModel)


    @access.public
    def hasModel(self, params):
        pass

    addModel('model', 'Body', {
        'id': 'Body',
        'require': 'bob',
        'properties': {
            'bob': {
                'type': 'array',
                'items': {
                    'type': 'integer'
                }
            }
        }
    })

    hasModel.description = describe.Description('What a model') \
        .param('body', 'Where its at!', dataType='Body', required=True,
               paramType='body')
```

brianhelba:
@cjh1 (our existing thread is getting hidden by GitHub, so I'm just going to continue it with PR-level comments here)

We need to call ``addRouteDocs`` from within ``Resource.route`` in the ``Resource`` constructor because we don't know the full details of the route path until that point. Since models don't need to know anything about particular routes, we could just remove the check, and allow addModel to implicitly define a new resource type.

Hopefully, by the time everything is initialized, that resource will get filled in with some actual routes. I can't think of a good place check for and remove empty endpoints, unless you want to run it on every call to ``Describe.describeResource``.
* In theory, there's also a very unlikely race condition, where a plugin attempts to add a new model + route to a new resource type sometime during active server operation, and gets interrupted by a request to ``Describe.describeResource`` which deletes the new resource type before the creation is complete. This is so unlikely as to be ignorable, but I wanted to add it.

I won't make the change until I get some feedback, but I'm fine with removing the check if nobody has a better solution.

cjh1:
@brianhelba I have push a simple test case which can be used to verify an future changes, I leave towards removing the check.

brianhelba:
I've just removed the check.

girder/api/docs.py
```diff
@@ -112,13 +112,32 @@ def removeRouteDocs(resource, route, method, info, handler):
...
+    .. versionchanged:: The syntax and behavior of this function was modified
+        after v1.3.2. The previous implementation did not include a resources
+        parameter.
+
+    .. _Swagger Model documentation: https://github.com/swagger-api/
+       swagger-spec/blob/d79c205485d702302003d4de2f2c980d1caf10f9/
+       versions/1.2.md#527-model-object
     """
-    models[name] = model
+    if isinstance(resources, str):
```
zachmullen: We should check this against `six.string_types`.

brianhelba: Done. It is possible to have a unicode resource name.

zachmullen:
After reading through the conversation here, I've formulated an opinion. I think maintaining compatibility is actually more important than the small inefficiency of sending the models for every resource, and I propose that we adopt @cpatrick 's suggestion with the additional caveat that calling the method with `resource=None` will print a warning to the console to let people know that it's bad practice even though it works. Such warnings, like the ones we print if someone forgets a route description, would appear when someone starts the server, so it would be pretty obvious if it happens and people would eventually transition their code without it being a critical breakage.

We as developers of a project tend to undervalue the importance of our own projects' API stability to downstreams; I know I've been frustrated many times by breaking changes even in some of our own libraries that I'm pretty close to. I'd hate to cause a bad experience for downstream devs who aren't intimately familiar with what goes on amongst our core developers.

If we agree on such a solution, I would be happy to implement it if need be.

cpatrick:
+1 to  the tweak @zachmullen made to my suggestion. The more I think about this, the more I want to keep backward compatibility. I think the warning is really nice way to handle this and satisfies people fairly well.

@cjh1, I want to get your final sign off on this.

@brianhelba, I also want to make sure you're okay with this also. Girder is getting to the point where backward compatibility is important enough that breaking it needs greater benefit than this would get us.

cjh1:
+1 from me

brianhelba:
I don't see this as a matter of computational inefficiency at all, but rather as ensuring that our Swagger docs present usable and relevant information to developer-users. As an example, pulling up the API for ``/user`` should not present the viewer with information about an ``MyRandomPluginModel``.

As for API stability, I'm sympathetic in principle, but we should be realistic about whether anyone is actually using an undocumented (no docs built) function in an obscure, inadequately-maintained section of the code. Compatibility at all costs eventually leads to tons of unworkable, arcane, legacy code, which I'm sure we've also all experienced in other libraries.

Now that I've said my final piece, I'll accept the consensus, particularly if we promise to soon add a ``Description.model`` method as the "correct" way to set a model on a ``Resource``.

cpatrick:
@brianhelba Thanks for pushing on this. This tension is really important to point out as we push Girder forward.I love this thread.

Would you open an issue for the "correct" way so that we can track this as a technical debt item?

In the meantime:

gandalf

:smile:

cjh1:
To be fair ```MyRandomPluginModel``` would only appear in the user API doc it some used the model ...

zachmullen:
> As an example, pulling up the API for /user should not present the viewer with information about an MyRandomPluginModel.

In the swagger UI, this doesn't happen. The models would show up if someone was inspecting the raw JSON of the swagger response, but in our swagger UI, models show up only where they are referenced by endpoints. Maybe that's what you were talking about, but just wanted to make that clear.

cjh1:
Yes, that is what I meant

brianhelba:
@cpatrick https://github.com/girder/girder/issues/971

brianhelba:
@zachmullen Since you volunteered to implement some things, I'll let you amend or close this PR as you see fit.

zachmullen:
Roger that.

zachmullen:
@cjh1 @cpatrick @brianhelba PTAL

brianhelba:
LGTM.

PR #967, opened 2015-09-09
Build documentation for Assetstore, File, and Notification models

brianhelba:
The documentation for these models was previously not being built.

Also, move the documentation on Events to a more logical ordering on the page.

Finally, fix major syntax errors in File and Notification documentation. These were probably never noticed because because File and Notification documentation were not being built, but these errors caused entire docstrings for several methods to not build properly.

### DIFF
--- a/docs/api-docs.rst
+++ b/docs/api-docs.rst
@@ -67,14 +67,6 @@ Model Base
 .. automodule:: girder.models.model_base
    :members:
 
-.. _events:
-
-Events
-^^^^^^
-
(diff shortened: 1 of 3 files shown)


brianhelba:
I've tested this locally, but we should ensure that the page on ReadTheDocs looks good once the new version is published. Nobody has seen the built documentation for these models before.

zachmullen:
Will do. They *should* look roughly the same locally vs. RTD. LGTM

PR #968, opened 2015-09-09
Add method to python client for uploading from file-like object

zachmullen:


clients/python/girder_client/__init__.py
```diff
@@ -94,7 +94,7 @@ class GirderClient(object):
     }
 
     # The current maximum chunk size for uploading file chunks
-    MAX_CHUNK_SIZE = 1024 * 1024 * 64
+    MAX_CHUNK_SIZE = 1024 * 1024 * 32
```
brianhelba: @zachmullen Just curious, why did you find 32MB to be better?

clients/python/girder_client/__init__.py
```diff
@@ -488,6 +488,71 @@ def uploadFileToItem(self, itemId, filepath):
...
+                'an object with an id. Got instead: ' + json.dumps(obj))
+
+        offset = 0
+        while True:
+            data = stream.read(self.MAX_CHUNK_SIZE)
+
+            if not data:
+                break
+
+            params = {
```
brianhelba: It would eventually be nice to refactor this and logic in ``uploadFileToItem`` to a common place.

clients/python/girder_client/__init__.py
```diff
@@ -488,6 +488,71 @@ def uploadFileToItem(self, itemId, filepath):
...
+        :type stream: file-like
+        :param name: The name of the file to create.
+        :type name: str
+        :param size: The length of the file.
+        :type size: str
+        :param parentType: 'item' or 'folder'. Default assumes parent type is
+            item.
+        :type parentType: str or None
+        :param progressCallback: If passed, will be called after each chunk
+            with progress information. It passes a single parg to the callable
```
brianhelba: @manthey 's comment is about a typo, ``parg``, here.

brianhelba:
@manthey The server depends on the ``size`` to know when to call finalizeUpload: https://github.com/girder/girder/blob/master/girder/models/upload.py#L106

I'm not sure what happens if you keep uploading chunks to an upload that's been finalized, but I'm checking now. If there's not any server-side checks for this, there should be.

brianhelba:
Continuing to send chunks to a finished upload will just cause the server to return a 400 code, so it won''t cause any database breakage.
https://github.com/girder/girder/blob/master/girder/models/upload.py#L172
https://github.com/girder/girder/blob/master/girder/api/v1/file.py#L172

Additionally, there are certainly going to be circumstances where it's impractical to double-check the size inside ``uploadFile``. File-like objects are expected to have ``read``, but not always a ``seek``. The file data itself may be streamed from another external source, which could be difficult to random-access or read-through twice.

I think it could make sense to add a bit more documentation to the ``size`` parameter, warning that it must be accurate.

mgrauer:
@brianhelba Thanks for digging into that, seems like we are OK on the server side when actual file size > `size`.  But, since

>The server depends on the size to know when to call finalizeUpload: https://github.com/girder/girder/blob/master/girder/models/upload.py#L106

if the caller of `uploadFile` passes a FLO large enough for chunking, and provides fewer bytes than `size`, `finalizeUpload` will never be called.  

It would be a simple matter to check `size` vs `offset` after the end of the loop to cover this case and in general, as that might prevent some headaches.  It could be covered by documentation, as you say.

brianhelba:
+1 to a check at the end of the loop if `size != offset`, and either explicitly finalize the upload or delete the upload and raise an exception (depending on how tolerant we want the client to be)

zachmullen:
@brianhelba I agree, and this brings up another missing component -- users have no way to manually cancel an upload via the REST API, only site admins can do it currently. I'll add that endpoint and make our client call it.

zachmullen:
@mgrauer @brianhelba I think I addressed everything, PTAL

mgrauer:
I like this cleanup behavior in the client when the sizes are mismatched.

Can you add a test for the file/upload DELETE endpoint?  You are calling it in the client test, but not asserting its results that I can see.

brianhelba:
LGTM, but I'd like to hear back from @mgrauer before this is merged.

mgrauer:
LGTM.

PR #973, opened 2015-09-11
Use local install of grunt for npm install

jbeezley:
Following the conversion on the girder mailing list.  Grunt was already being installed locally, so we might as well not enforce a global grunt install in the npm install step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ before_install:
     - npm --version
 install:
     - scripts/InstallPythonRequirements.py --mode=dev --ignore-plugins=${IGNORE_PLUGINS}
-    - npm install grunt grunt-cli
     - npm install
 script:
     - mkdir _build
--- a/package.json
+++ b/package.json
@@ -43,6 +43,6 @@
     "jscs": "1.6.1"
   },
   "scripts": {
-    "postinstall": "grunt init && grunt"
+    "postinstall": "./node_modules/.bin/grunt init && ./node_modules/.bin/grunt"
   }
 }


brianhelba:
+1 as a concept, but I'm not familiar enough with this to approve

zachmullen:
@jbeezley This is essentially a revert of the following commit: https://github.com/girder/girder/commit/1434c4e1b5265aa8acc0d4c24ebff0a52b8b110e

ping @manthey as the authority here, but the commit message makes it seem like the current state of the script does not actually require these to be installed globally.

jbeezley:
Right, I had forgotten about the windows issue.  I don't personally have much of a preference here, but this seemed to be where the discussion on the mailing list was going.

If you are using a global grunt, I don't believe you need to install `grunt-cli` locally.  My understanding is that is why there exists two packages `grunt` vs. `grunt-cli`.

manthey:
I'll eventually test this on Windows again.  I haven't tested on Windows is
months.

On Fri, Sep 11, 2015 at 10:47 AM, Jonathan Beezley <notifications@github.com
> wrote:

> Right, I had forgotten about the windows issue. I don't personally have
> much of a preference here, but this seemed to be where the discussion on
> the mailing list was going.
>
> If you are using a global grunt, I don't believe you need to install
> grunt-cli locally. My understanding is that is why there exists two
> packages grunt vs. grunt-cli.
>
> —
> <https://github.com/girder/girder/pull/973#issuecomment-139565044>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

package.json
```diff
@@ -43,6 +43,6 @@
     "jscs": "1.6.1"
   },
   "scripts": {
-    "postinstall": "grunt init && grunt"
+    "postinstall": "./node_modules/.bin/grunt init && ./node_modules/.bin/grunt"
```
jcfr: This doesn't seems required. At least testing on my local workstation.

jbeezley: Okay good to know.

.travis.yml
```diff
@@ -32,7 +32,6 @@ before_install:
     - npm --version
 install:
     - scripts/InstallPythonRequirements.py --mode=dev --ignore-plugins=${IGNORE_PLUGINS}
-    - npm install grunt grunt-cli
```
jcfr: While https://github.com/girder/girder/pull/973/files#r39280145 mention there is no need to specify `./node_modules/.bin/`,  removing the explicit install makes sense. `grunt ` and `grunt-cli` are specified as dependencies in `package.json` and will be automatically installed when calling `npm install`

PR #979, opened 2015-09-14
change POST->assetstore->type type documentation from integer to string

opadron:
Changes the documentation for the `POST/assetstore` route so that the *type* of the `type`  argument correctly reflects that it may be any non-integer string.

### DIFF
--- a/girder/api/v1/assetstore.py
+++ b/girder/api/v1/assetstore.py
@@ -101,7 +101,7 @@ class Assetstore(Resource):
         .responseClass('Assetstore')
         .notes('You must be an administrator to call this.')
         .param('name', 'Unique name for the assetstore.')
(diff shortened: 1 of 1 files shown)


girder/api/v1/assetstore.py
```diff
@@ -101,7 +101,7 @@ def createAssetstore(self, params):
         .responseClass('Assetstore')
         .notes('You must be an administrator to call this.')
         .param('name', 'Unique name for the assetstore.')
-        .param('type', 'Type of the assetstore.', dataType='integer')
+        .param('type', 'Type of the assetstore.', dataType='string')
```
zachmullen: This is the default value, so we can remove the kwarg altogether.

zachmullen:
LGTM once travis is happy

PR #984, opened 2015-09-15
Login breakage with oauth users

zachmullen:

PR #986, opened 2015-09-16
Password reset security fix

### LINKED ISSUE #955: User.generateTemporaryPassword uses overly-complicated code to generate external URLs
Any references to low-level things like the ``X-Forwarded-Host`` header should be refactored out of User.generateTemporaryPassword and replaced with calls to common utility modules.

zachmullen:
This depends on #984, only the last commit here refers to this topic. I will rebase once 984 goes in, just wanted to put this up.

This contains some code reorganization that also happens to fix #955.


zachmullen:
Rebased, this is now ready for review.

zachmullen:
@girder/developers anybody want to take a crack at reviewing this?

girder/api/v1/user.py
```diff
@@ -260,28 +260,35 @@ def changeUserPassword(self, user, params):
     def changePassword(self, params):
         self.requireParams(('old', 'new'), params)
         user = self.getCurrentUser()
+        token = None
 
         if not params['old']:
             raise RestException('Old password must not be empty.')
 
         if (not self.model('password').hasPassword(user) or
```
brianhelba: If the user has no password, what would we expect ``params['old']`` to contain?

girder/api/v1/user.py
```diff
@@ -312,28 +319,24 @@ def generateTemporaryPassword(self, params):
...
-                'url': url,
-                'token': str(token['_id'])
-            })
-            mail_utils.sendEmail(to=email, subject='Girder: Temporary Access',
-                                 text=html)
+
+        token = self.model('token').createToken(
+            user, days=1, scope=TokenScope.TEMPORARY_USER_AUTH)
+
+        emailHost = self.model('setting').get(SettingKey.EMAIL_HOST)
```
brianhelba: Shouldn't you use [``getEmailUrlPrefix``](https://github.com/girder/girder/blob/master/girder/utility/mail_utils.py#L33) here?

tests/cases/user_test.py
```diff
@@ -549,17 +561,9 @@ def testTemporaryPassword(self):
         }, user=user)
         self.assertStatusOk(resp)
 
-        # Generate an email with a forwarded header
```
brianhelba: Why aren't you checking whether the email was sent anymore?

girder/api/v1/user.py
```diff
@@ -312,28 +319,24 @@ def generateTemporaryPassword(self, params):
         self.requireParams('email', params)
         email = params['email'].lower().strip()
 
-        users = self.model('user').find({'email': email})
+        user = self.model('user').findOne({'email': email})
 
-        if not users.count():
+        if not user:
             raise RestException('That email is not registered.')
-        for user in users:
```
brianhelba: Was there a legitimate reason for this code for the possibility of multiple users with the same email?

zachmullen: No, I think the code just predated `findOne`.

tests/cases/user_test.py
zachmullen: The purpose of creating the email below was to ensure the behavior of the `X-Forwarded-Host` header that was removed as part of this topic branch. That code is obsolete now that we have the email host setting. This is part of the fix for #955

girder/api/v1/user.py
zachmullen: A user with no password is currently not able to use this endpoint -- the OAuth plugin and any other plugin that allows passwordless users to exist should really hook into this and do something sensible, but as of right now that doesn't exist.

brianhelba:
LGTM.

PR #987, opened 2015-09-16
Fix large zip files

manthey:
If a zip file contained multiple files and one of them started at an offset >= 2Gb from the start of the file, the zip footer would fail to be created.

Also, zip filenames with non-ASCII characters could fail to be created due to string conversions.

The test has been updated to detect both of these conditions.

### DIFF
--- a/girder/utility/ziputil.py
+++ b/girder/utility/ziputil.py
@@ -72,13 +72,14 @@ class ZipInfo(object):
     def __init__(self, filename, timestamp):
         # Terminate the file name at the first null byte.  Null bytes in file
         # names are used as tricks by viruses in archives.
-        nullByte = filename.find(chr(0))
-        if nullByte >= 0:
-            filename = filename[0:nullByte]
         if os.sep != '/' and os.sep in filename:
             filename = filename.replace(os.sep, '/')
         if isinstance(filename, six.text_type):
             filename = filename.encode('utf8')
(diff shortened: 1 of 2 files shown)


girder/utility/ziputil.py
```diff
@@ -214,7 +216,7 @@ def footer(self):
 
             if header.headerOffset > Z64_LIMIT:
                 extra.append(header.headerOffset)
-                headerOffset = -1
+                headerOffset = 0xffffffff
```
zachmullen: So this was the actual bug?

manthey: Yes.

On Wed, Sep 16, 2015 at 1:38 PM, Zach Mullen <notifications@github.com>
wrote:

> In girder/utility/ziputil.py
> <https://github.com/girder/girder/pull/987#discussion_r39660750>:
>
> > @@ -214,7 +216,7 @@ def footer(self):
> >
> >              if header.headerOffset > Z64_LIMIT:
> >                  extra.append(header.headerOffset)
> > -                headerOffset = -1
> > +                headerOffset = 0xffffffff
>
> So this was the actual bug?
>
> —
> <https://github.com/girder/girder/pull/987/files#r39660750>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

zachmullen:
LGTM

PR #988, opened 2015-09-17
Python client version 1.1.0

zachmullen:
* Adds a new, more flexible upload method that can upload from
  a file-like object rather than a path on the filesystem.


zachmullen:
Any objections to uploading this to pypi?

PR #990, opened 2015-09-17
Ensure we load 'access' and 'public'

cjh1:
requireAccess(...) needs to use 'access' and 'public' so we need to ensure they are loaded. If they were not specified by user, remove them before returning the document.

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -798,10 +798,24 @@ class AccessControlledModel(Model):
         :raises ValidationException: If an invalid ObjectId is passed.
         :returns: The matching document, or None if no match exists.
         """
-        doc = Model.load(self, id=id, objectId=objectId, fields=fields, exc=exc)
+
+        # Ensure we load access and public, these are needed by requireAccess
+        loadFields = None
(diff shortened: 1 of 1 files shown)


girder/models/model_base.py
```diff
@@ -798,11 +798,26 @@ def load(self, id, level=AccessType.ADMIN, user=None, objectId=True,
         :raises ValidationException: If an invalid ObjectId is passed.
         :returns: The matching document, or None if no match exists.
         """
-        doc = Model.load(self, id=id, objectId=objectId, fields=fields, exc=exc)
+
+        # Ensure we load access and public, these are needed by requireAccess
+        loadFields = None
+        if not force and fields:
+            loadFields = fields[:]
+            loadFields += ['access', 'public']
```
zachmullen: It may not matter much, but to avoid duplication we could change to

```
if not force and fields:
    loadFields = list(set(fields) | {'access', 'public'})
```

cjh1: It doesn't but I can change it

zachmullen:
LGTM once travis is happy

PR #991, opened 2015-09-17
Fixes #989. Disable cherrypy autoreload unless in dev mode

### LINKED ISSUE #989: Disable cherrypy auto-reload behavior unless we are in "dev" environment
Test and prod modes should not have this behavior.

zachmullen:
@girder/developers PTAL, this is a small change but just want to make sure everyone is OK with it.

### DIFF
--- a/girder/__main__.py
+++ b/girder/__main__.py
@@ -25,7 +25,7 @@ try:  # pragma: no cover
     from girder.utility import server
 except ImportError:
     # Update python path to ensure server respawning works. See #732
-    source_root_dir = os.path.join(os.path.dirname(__file__), '..')
+    source_root_dir = os.path.dirname(os.path.dirname(__file__))
     import sys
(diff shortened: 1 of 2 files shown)


jbeezley:
Seems okay to me.  Does testing use a mode other than `development`?

zachmullen:
@jbeezley It uses a mode called ["testing"](https://github.com/girder/girder/blob/disable-autoreload/girder/utility/server.py#L80-L87).

zachmullen:
Any concerns about this, or should I merge?

jbeezley:
I'll give a LGTM.

jcfr:
Sorry for the last miniute comment. But does this mean that the server will  not restart automatically if a plugin is enabled from the admin console ?

zachmullen:
@jcfr that functionality still works, this only disables the file watcher.

jcfr:
@zachmullen  Makes sense. Thanks for clarifying.

PR #993, opened 2015-09-18
Update phantomjs version for compatability with iojs

jbeezley:
I was hitting a bug in phantomjs locally using iojs.  It looks like official iojs support wasn't add unit 1.9.15.  There doesn't appear to be any issues updating this to the latest in the 1.9 series.


zachmullen:
LGTM, I had run into this too. Thanks for looking into it.

PR #994, opened 2015-09-21
Expose log files

zachmullen:
A common pain point during deployment is that log files aren't exposed via the web app, meaning admins have to SSH into the machine, find the log file, and tail it. This exposes the same functionality via an easy-to-use REST endpoint that also shows the path to the log file it's reading from.

@mgrauer we discussed this earlier today

### DIFF
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -35,6 +35,7 @@ from ..describe import API_VERSION, Description
 from ..rest import Resource, RestException
 
 ModuleStartTime = datetime.datetime.utcnow()
+LOG_BUF_SIZE = 65536
 
 
 class System(Resource):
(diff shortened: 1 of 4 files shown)


girder/api/v1/system.py
```diff
@@ -373,3 +376,37 @@ def systemConsistencyCheck(self, params):
...
+        def stream():
+            yield '=== Last %d bytes of %s: ===\n\n' % (
+                min(length, filesize), path
+            )
+
+            with open(path, 'rb') as f:
+                if length < filesize:
+                    f.seek(-length, os.SEEK_END)
+                while True:
+                    data = f.read(65536)
```
cpatrick: image

cpatrick:
I like this. Thanks for not logging to the database. :smile_cat:

zachmullen:
@cpatrick Do you want me to move that magic number to a const variable?

cpatrick:
that works.

zachmullen:
@cpatrick PTAL

cpatrick:
That  extra `FER` just took up too much space. :smile_cat: 

Other than that sub-nit, LGTM. :ship:

zachmullen:
I could've used C naming conventions and called it `__LB_LN_` or something ;)

PR #997, opened 2015-09-23
Fix cmake binary URL on travis

jbeezley:


zachmullen:
LGTM. You'd think following redirects would be the default behavior :P

jbeezley:
I know right?

PR #999, opened 2015-09-27
Filesystem assetstore import

zachmullen:


girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -270,3 +277,61 @@ def cancelUpload(self, upload):
...
+        :param name: Name for the file. Defaults to the basename of ``path``.
+        :type name: str
+        :param mimeType: MIME type of the file if known.
+        :type mimeType: str
+        :param size: Size of the file in bytes. Will stat the file if not
+            passed.
+        :type size: int
+        :returns: The file document that was created.
+        """
+        if size is None:
```
brianhelba: What's the case for having size be a parameter at all? Why not just always stat it?

zachmullen: As a potential optimization for callers of this method. If they've already done the stat, they can save time by just passing the value in.

Issue #961, opened 2015-09-08
Add API endpoint for content-addressed downloads


mathstuf:
Midas allows one to download using `&checksum=$hash&algorithm=$algo`. It would be nice to have such content-addressed download URLs for girder as well (something like `item/content/{algo}/{hash}/download`?).

These URLs are used for VTK's data store downloading (where the `{id}` of the file is not necessarily known).

Of course, if there is a better way, I'm willing to be enlightened :) .

mathstuf:
Cc: @brad.king

jcfr:
Per discussion with @zachmullen , a possible approach will be to develop a `ContentAddressedDownload` plugin allowing to:
* enable the content-addressed download on per-collection basis
* configure which hashing algorithm should be associated with a given collection
* batch-update existing item in a given collection based on the selected hash algorithm

Notes:
* At first, since downloading data from cloud asset-store is usually not free, the plugin will be available only if a filesystem asset-store is enabled. If there is a need, cloud based asset-store support could be added later.

brianhelba:
I have a need for this now, and after discussing it with @jbeezley , I'm planning to implement this as a plugin where:
* Upon upload of a file to any non-S3 assetstore type, the plugin will hash the file with all of the algorithms that [CMake ExternalData supports](https://cmake.org/cmake/help/v3.3/module/ExternalData.html#hash-algorithms).
 * Since non-proxied uploads to S3 never give Girder direct custody of the file's binary data, we can't guarantee full support for S3 assetstores. However, if a proxied S3 upload takes case, the plugin should hash the content. Finally, note that it's possible to [get the MD5 (but not other hashes) of a file in S3 without downloading it](http://stackoverflow.com/questions/1775816/how-to-get-the-md5sum-of-a-file-on-amazons-s3)
 * Generating multiple hash types shouldn't be too much more expensive, since we can have multiple stateful hash variables, each of which is updated as the file is iterated through once.
* The file may be downloaded via a ``/api/v1/file/content/{algo}/{hash}/download`` URL.
 * I don't have a strong opinion about the syntax of this URL, so chime in if you have a better idea.
* At first, this plugin will not support access to files that were uploaded before the plugin was enabled.
* The plugin will be enabled for all files globally.
 * I don't really understand the need to only enable this on a per-collection basis. Feel free to present a use case that requires this, however.

brianhelba:
[...]

PR #1000, opened 2015-09-29
Filesystem assetstore sync

zachmullen:
Depends on #999 

Question for @girder/developers is whether we want to make this new method, which goes through and hunts down files whose underlying data is missing, should be part of the abstract assetstore API, or if it's OK to leave as a one-off method within the filesystem assetstore adapter for now.

Another question would be whether we want to expose this via the REST API. For now I chose not to since my need of this method does not require it.

### DIFF
--- a/girder/utility/abstract_assetstore_adapter.py
+++ b/girder/utility/abstract_assetstore_adapter.py
@@ -18,6 +18,7 @@ import cherrypy
 import os
 import six
 
+from girder.utility import progress
 from ..constants import SettingKey
 from .model_importer import ModelImporter
 from ..models.model_base import ValidationException
@@ -143,6 +144,24 @@ class AbstractAssetstoreAdapter(ModelImporter):
         """
         raise Exception('Must override downloadFile in %s.'
                         % self.__class__.__name__)  # pragma: no cover
+
+    def findInvalidFiles(self, progress=progress.noProgress, filters=None,
+                         checkSize=True, **kwargs):
+        """
+        Finds and yields any invalid files in the assetstore. It is left to
+        the caller to decide what to do with them.
+
+        :param progress: Pass a progress context to record progress.
+        :type progress: :py:class:`girder.utility.progress.ProgressContext`
+        :param filters: Additional query dictionary to restrict the search for
+            files. There is no need to set the ``assetstoreId`` in the filters,
+            since that is done automatically.
+        :type filters: dict or None
+        :param checkSize: Whether to make sure the size of the underlying
+            data matches the size of the file.
+        :type checkSize: bool
+        """
+        return ()
 
     def copyFile(self, srcFile, destFile):
         """
--- a/girder/utility/abstract_assetstore_adapter.py
+++ b/girder/utility/abstract_assetstore_adapter.py
@@ -85,8 +85,8 @@ class AbstractAssetstoreAdapter(ModelImporter):
         :type chunk: file
         :returns: Must return the upload document with any optional changes.
         """
-        raise Exception('Must override processChunk in %s.'
-                        % self.__class__.__name__)  # pragma: no cover
+        raise NotImplementedError('Must override processChunk in %s.' %
+                                  self.__class__.__name__)  # pragma: no cover
 
     def finalizeUpload(self, upload, file):
         """
@@ -122,8 +122,8 @@ class AbstractAssetstoreAdapter(ModelImporter):
         :param file: The File document about to be deleted.
         :type file: dict
         """
-        raise Exception('Must override deleteFile in %s.'
-                        % self.__class__.__name__)  # pragma: no cover
+        raise NotImplementedError('Must override deleteFile in %s.' %
+                                  self.__class__.__name__)  # pragma: no cover
 
     def downloadFile(self, file, offset=0, headers=True, endByte=None):
         """
@@ -142,8 +142,8 @@ class AbstractAssetstoreAdapter(ModelImporter):
             end of the file.
         :type endByte: int or None
         """
-        raise Exception('Must override downloadFile in %s.'
-                        % self.__class__.__name__)  # pragma: no cover
+        raise NotImplementedError('Must override downloadFile in %s.' %
+                                  self.__class__.__name__)  # pragma: no cover
 
     def findInvalidFiles(self, progress=progress.noProgress, filters=None,
                          checkSize=True, **kwargs):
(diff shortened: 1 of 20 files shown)


girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -70,7 +71,7 @@ def fileIndexFields():
         """
         File documents should have an index on their sha512 field.
```
jcfr: Should the docstring be updated to define the meaning of `imported` ?

jcfr:
If we are talking about the method `removeMissing`, I think it should belong to the abstract assetstore API. 

Nitpicks: 

Should we also add a method named `listMissing` ? In that case, `removeMissing` would simply be a convenience method calling `listMissing` (a generator ?) and then `self.model('file').remove(file)`

an other approach would be to add a `dry_run` parameter to the `removeMissing`, knowing what will be deleted and having a human approving it is always good.

jeffbaumes:
Happy 1000th issue Girder :birthday:

zachmullen:
> Should we also add a method named listMissing

@jcfr This is probably what this method should become. The fact that we have a callback for missing data actually makes this far more powerful than just removing the files. We can leave it to users to pass in a callback that deletes the file if that's what they wish.

jcfr:
@zachmullen I missed that. The `callback` is nice. Should we instead have callback be a mandatory value instead of hardcoding the current behavior ? 

Something like this:

```
+            if not os.path.isfile(path) and callback(info):
+                total += 1
+                logger.info('Cleaned missing file %s (%s).', file['name'], path)
```

with `callback` being set to either a staticmethod or an instance method of `FilesystemAssetstoreAdapter`

zachmullen:
PTAL, I changed the method to be more generic. Now any action taken on the invalid/missing files is left to the caller, and we use a generator instead of a callback.

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -270,3 +285,105 @@ def cancelUpload(self, upload):
...
+        if size is None:
+            size = os.path.getsize(path)
+
+        name = name or os.path.basename(path)
+
+        file = self.model('file').createFile(
+            name=name, creator=user, item=item, reuseExisting=True,
+            assetstore=self.assetstore, mimeType=mimeType, size=size)
+        file['fullPath'] = os.path.abspath(os.path.expanduser(path))
+        file['imported'] = True
```
jcfr: @zachmullen  Anticipating that we will be adding an endpoint to download data using their hash (See #961), should we systematically compute the hash of file ?

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -270,3 +285,105 @@ def cancelUpload(self, upload):
...
+        """
+        if size is None:
+            size = os.path.getsize(path)
+
+        name = name or os.path.basename(path)
+
+        file = self.model('file').createFile(
+            name=name, creator=user, item=item, reuseExisting=True,
+            assetstore=self.assetstore, mimeType=mimeType, size=size)
+        file['fullPath'] = os.path.abspath(os.path.expanduser(path))
```
jcfr: Looking at other assetstore, path are stored differently in the `file` model. Would it make sense to re-use existing convention ? `path` or `relpath`.

* imported file:

file['fullPath']

* [s3_assetstore_adapter.py#L280](https://github.com/girder/girder/blob/7760491d004510efee39e2d9f8b8174d051853fa/girder/utility/s3_assetstore_adapter.py#L280):

```
file['relpath']
```

* [filesystem_assetstore_adapter.py#L209](https://github.com/girder/girder/blob/7760491d004510efee39e2d9f8b8174d051853fa/girder/utility/filesystem_assetstore_adapter.py#L209):

```
file['path']
````

File: girder/utility/abstract_assetstore_adapter.py [hunk 4]
```diff
@@ -144,6 +145,24 @@ def downloadFile(self, file, offset=0, headers=True, endByte=None):
...
+        :type progress: :py:class:`girder.utility.progress.ProgressContext`
+        :param filters: Additional query dictionary to restrict the search for
+            files. There is no need to set the ``assetstoreId`` in the filters,
+            since that is done automatically.
+        :type filters: dict or None
+        :param checkSize: Whether to make sure the size of the underlying
+            data matches the size of the file.
+        :type checkSize: bool
+        """
+        return ()
```
brianhelba: How about ``raise NotImplementedError()`` here?

I realize that to be consistent, we'd have to change a few other methods too.

zachmullen: Fine by me.

girder/utility/filesystem_assetstore_adapter.py
zachmullen: I'm not sure if it makes sense to have such an endpoint as described in #961, as it breaks the transparency of the assetstore pattern. By design, the abstract assetstore adapter API doesn't provide any guarantees about hashing of the data, since that wasn't something we wanted to force on assetstore implementations. In fact we already have at least one adapter that does not even receive the bytes of the file, so computing a hash would be infeasible. If possible, I would prefer to keep content hashing as an internal implementation detail of our adapters.

Computing hashes of large files is also going to add a lot of time to a method like this, and I don't think it adds value -- in the case of imported files, we are not in charge of the file contents, so if they change on the underlying file, we don't care, we'll serve up whatever we're told, we just have to make sure the size is correct. For non-imported files, it shouldn't ever be the case that their content has changed unbeknownst to us.

It's my hope that whatever use case is meant to be fulfilled in #961 could also be handled using something other than hashes. If not, maybe we could create a custom plugin that will only work with certain assetstore types and exposes such an endpoint.

girder/utility/filesystem_assetstore_adapter.py
zachmullen: Sure, I will change the imported file to use `'path'` rather than `'fullPath`'.

girder/utility/filesystem_assetstore_adapter.py
jcfr: Very good point. I summarized our discussion here : https://github.com/girder/girder/issues/961#issuecomment-144836102

zachmullen:
I think I addressed all the comments, PTAL

jcfr:
:+1:  LGTM . That said I didn't QA the UI

PR #1001, opened 2015-09-29
Removing non-responsive badge

danlamanna:
Looks like pypin.in is down, [for good](https://github.com/badges/pypipins/issues/37#issuecomment-127772864)?

Do we care to look for a replacement, or can this be removed altogether?


jbeezley:
This would be an alternative if we want a replacement: [PyPI]()
```
.. image:: https://img.shields.io/pypi/l/girder.svg   :target: 
```

zachmullen:
Let's use @jbeezley 's suggestion.

PR #1002, opened 2015-09-30
Add file copy endpoint

cjh1:


cjh1:
This is ready for review

girder/api/v1/file.py
```diff
@@ -322,3 +323,23 @@ def updateFileContents(self, file, params):
...
+                                           level=AccessType.WRITE, exc=True)
+
+        newFile = fileModel.copyFile(file, user, item=item)
+
+        return fileModel.filter(newFile, user)
+
+    copy.description = (
+        Description('Copy a file.')
+        .param('id', 'The ID of the file.', paramType='path')
+        .param('itemId', 'The item to copy the file to.', required=False))
```
zachmullen: For this endpoint, I vote that we always require an itemId parameter since the behavior of copying a file into the same item is probably going to be very uncommon. It will also simplify the method, since we could use

    @loadmodel(model='file', level=AccessType.READ)
    @loadmodel(model='item', map={'itemId': 'item'}, level=AccessType.WRITE)
    def copy(self, file, item, params):

cjh1: Ok, I can do that

cjh1:
@zachmullen This should be ready again ...

zachmullen:
LGTM

Issue #1003, opened 2015-09-30
Size quotas appear to be broken


zachmullen:
I made a new non-admin user on data.kitware.com, where the default quota is 2GB. When I view my quota as that user, however, I see:

![screen shot 2015-09-30 at 3 24 47 pm](https://cloud.githubusercontent.com/assets/555959/10204019/68d30606-6787-11e5-9c2a-e4a626d8a52d.png)

But as an admin, I am told that the default 2GB limit applies:

![screen shot 2015-09-30 at 3 21 07 pm](https://cloud.githubusercontent.com/assets/555959/10204027/723d886a-6787-11e5-9ba5-13c6a293113a.png)

It also doesn't seem to be enforced; @danlamanna was able to upload well past the default quota.

ping @manthey any idea why this happens?

manthey:
Fixed in PR #1004.

PR #1004, opened 2015-10-01
Fix quota issues.

manthey:
There were two issues.  If a user or collection never had a quota set on it, default quotas were not enforced.

In the web client, if logged in as a non-admin user and viewing the quota for a user or collection that used a default quota, it would show 'Unlimited' instead of the appropriate quota.

Add tests that would have failed before and now work.

This fixes issue #1003 .

### DIFF
--- a/plugins/user_quota/web_client/js/userQuota.js
+++ b/plugins/user_quota/web_client/js/userQuota.js
@@ -206,7 +206,7 @@ girder.views.QuotaPolicies = girder.View.extend({
         }
         $(el).addClass('g-has-chart');
         used = view.model.get('size');
-        free = used < quota ? quota - used : 0;
+        free = Math.max(quota - used, 0);
         data = [
             ['Used (' + girder.formatSize(used) + ')', used],
             ['Free (' + girder.formatSize(free) + ')', free]
--- a/plugins/user_quota/web_client/js/userQuota.js
+++ b/plugins/user_quota/web_client/js/userQuota.js
@@ -3,6 +3,7 @@
 (function () {
     _.each({user: 'User', collection: 'Collection'}, function (
             modelName, modelType) {
+        var fullModelName;
         var viewName = modelName + 'View';
         girder.views[viewName] = girder.views[viewName].extend({
             events: function () {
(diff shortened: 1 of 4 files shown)


plugins/user_quota/web_client/js/userQuota.js
```diff
@@ -189,18 +192,22 @@ girder.views.QuotaPolicies = girder.View.extend({
...
         if (!quota) {
             $(el).addClass('g-no-chart');
             return;
         }
         $(el).addClass('g-has-chart');
-        var used = view.model.get('size');
-        var free = used < quota ? quota - used : 0;
-        var data = [
+        used = view.model.get('size');
+        free = used < quota ? quota - used : 0;
```
zachmullen: Looks like `Math.max(quota - used, 0)` :)

plugins/user_quota/web_client/js/userQuota.js
```diff
@@ -334,11 +346,11 @@ girder.userQuota = {
      *                    for none. */
     valueAndUnitsToSize: function (sizeValue, sizeUnits) {
```
zachmullen: Could we get rid of this function? If we changed the select element to look like the following:

```
select.form-control.input-sm.g-sizeUnits(model="#{resource.model}")
        option(value='1', selected=(resource.sizeUnits === 0)) bytes
        option(value='1024', selected=(resource.sizeUnits === 1)) kB
        option(value='#{1024*1024}', selected=(resource.sizeUnits === 2)) MB
        option(value='#{1024*1024*1024}', selected=(resource.sizeUnits === 3)) GB
        option(value='#{1024*1024*1024*1024}', selected=(resource.sizeUnits === 4)) TB
```

Then it's a simple matter of doing parseInt() and multiplying.

zachmullen: (Maybe not get rid of the function, but at least make it more readable by removing the regexp and the looping...)

manthey: Using the rexep allows a user to type '5 Gb' and ignore the pull down.

zachmullen: Fair enough, let's leave it then

zachmullen:
LGTM, thanks for fixing this.

PR #1010, opened 2015-10-01
 Fixes #1007. No default folders when creating a collection

### LINKED ISSUE #1007: Add an option to not automatically create "Private" folders

zachmullen:


cpatrick:
I hate to bring this up, but isn't this technically breaking backward compatibility... :wind_chime:

zachmullen:
@cpatrick I'm glad you brought that up since it's an important thing to consider. This behavior was an undocumented side-effect of collection creation, and while it's possible that this change could break someone's workflow, I'm hesitant to consider undocumented side effects as part of our API surface, though it's clear that [some side effects are more important than others](https://xkcd.com/1172/). 

On this particular change: this was one of the most ancient pieces of code in girder, and a replication of behavior from a previous software stack. In the history of this feature in both systems, it has almost always been regarded as an annoyance, at least in my experience.

Anyway, I'm curious to hear others' opinions on this subject, both the general point and regarding this specific feature :smile_cat:

cpatrick:
Yeah, I agree with you 96%. But I had to bring it up. Get a few more :+1:s, and I'm cool with this.

cpatrick:
image

jbeezley:
sword_bow_axe

And my :+1:.

mgrauer:
I'll add a :+1: 

I also want to say that although this is an `undocumented side effect`, since our tests exercised the old behavior, to me it pushes this a bit closer to `API surface` on the spectrum.

zachmullen:
I see four thumbs (counting the two from #1007). Is that enough thumbs to merge this?

(Also ping @brianhelba since this issue was brought up in a conversation with him yesterday)

brianhelba:
:+1: here.

zachmullen:
I'm going to go ahead and merge this then, thanks for taking a look all.

Issue #980, opened 2015-09-14
Enabling plugin throws KeyError if dependency doesn't exist


danlamanna:
At the time of enabling a plugin, Girder throws a KeyError if it has a dependency which does not exist.

[This](https://github.com/girder/girder/blob/733f987dfe1df58355574c6e8b18ad5481b2f940/girder/models/setting.py#L70) is the offending line.

For example, Romanesco requires the jobs plugin, which if not present will return something like this to the browser console:
```
[...]

PR #1011, opened 2015-10-02
Fix KeyError on non-existent plugin dependency

danlamanna:
See #980 for details of the error.

This addresses the issue by returning a 500 on the ```/system/plugins``` API call if a plugin can't be enabled. It addresses the UI issue by disabling the switch with a message if it can't enable all of the dependencies as well.

In the future, the PluginsView may need to be changed a bit - because it currently assumes that PUT calls to ```/system/plugins``` will succeed, leaving switches in an inconsistent state, restart buttons around when it doesn't need to be restarted, etc.

### DIFF
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -149,8 +149,13 @@ class System(Resource):
         Return the plugin information for the system. This includes a list of
         all of the currently enabled plugins, as well as
         """
+        allPlugins = plugin_utilities.findAllPlugins()
+        for plugin in allPlugins.values():
+            if 'dependencies' in plugin:
+                plugin['dependencies'] = list(plugin['dependencies'])
+
         return {
-            'all': plugin_utilities.findAllPlugins(),
+            'all': allPlugins,
             'enabled': self.model('setting').get(SettingKey.PLUGINS_ENABLED)
         }
     getPlugins.description = (
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -104,7 +104,8 @@ class System(Resource):
         .param('value', 'The value for this setting.', required=False)
         .param('list', 'A JSON list of objects with key and value representing '
                'a list of settings to set.', required=False)
-        .errorResponse('You are not a system administrator.', 403))
+        .errorResponse('You are not a system administrator.', 403)
+        .errorResponse('Failed to set system setting.', 500))
 
     @access.admin
     def getSetting(self, params):
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -181,6 +181,7 @@ class System(Resource):
         .responseClass('Setting')
         .notes('Must be a system administrator to call this.')
         .param('plugins', 'JSON array of plugins to enable.')
+        .errorResponse('Required dependencies do not exist.', 500)
         .errorResponse('You are not a system administrator.', 403))
 
     @access.admin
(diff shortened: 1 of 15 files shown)


zachmullen:
Let's add a test for this.

danlamanna:
Should we change the possible response statuses of ```/system/setting```? It would make things more complicated, but with this change we begin to entertain the notion that changing a setting can fail.

The JS tests will fail until I have ```has_nonexistent_deps``` enabled as a plugin.

zachmullen:
> Should we change the possible response statuses of /system/setting? It would make things more complicated, but with this change we begin to entertain the notion that changing a setting can fail.

Do you mean `PUT /system/plugins`?

danlamanna:
```PUT /system/plugins``` is already taken care of in ebe3f17. Though ```PUT /system/setting``` could trigger the required dependencies exception if trying to set the ```core_plugins_enabled``` setting to something invalid.

**Comment by zachmullen** [comment]
Oh, did you just mean in the description? Yes, we should definitely add that it can return a 400 if the value passed in is invalid (i.e. ValidationException is raised).

**Comment by danlamanna** [comment]
@zachmullen This is ready for review.

UI Screenshot:
girder-1011-plugin-pane


The only thing I couldn't figure out was how to make ```PLUGIN_DIRS``` a proper CMake multival argument while still passing all arguments to the environment variable. I made it a colon separated string for now.

Also, do you think it should be called ```PLUGIN_DIRECTORY``` to match the option we recently added to the configuration file?

File: tests/CMakeLists.txt [hunk 1]
```diff
@@ -44,7 +44,7 @@ add_web_client_test(data_filesystem "${PROJECT_SOURCE_DIR}/clients/web/test/spec
 add_web_client_test(data_gridfs "${PROJECT_SOURCE_DIR}/clients/web/test/spec/dataSpec.js" ASSETSTORE gridfs)
 add_web_client_test(data_gridfsrs "${PROJECT_SOURCE_DIR}/clients/web/test/spec/dataSpec.js" ASSETSTORE gridfsrs RESOURCE_LOCKS replicaset)
 add_web_client_test(data_s3 "${PROJECT_SOURCE_DIR}/clients/web/test/spec/dataSpec.js" ASSETSTORE s3 WEBSECURITY false)
-add_web_client_test(admin "${PROJECT_SOURCE_DIR}/clients/web/test/spec/adminSpec.js" ASSETSTORE s3 WEBSECURITY false)
+add_web_client_test(admin "${PROJECT_SOURCE_DIR}/clients/web/test/spec/adminSpec.js" ASSETSTORE s3 WEBSECURITY false PLUGIN_DIRS "${PROJECT_SOURCE_DIR}/plugins:${PROJECT_SOURCE_DIR}/tests/test_plugins")
```
zachmullen: Let's break this line up since it's gotten so long

girder/api/v1/system.py
```diff
@@ -149,8 +150,13 @@ def getPlugins(self, params):
         Return the plugin information for the system. This includes a list of
         all of the currently enabled plugins, as well as
         """
+        allPlugins = plugin_utilities.findAllPlugins()
+        for plugin in allPlugins.values():
+            if 'dependencies' in plugin:
+                plugin['dependencies'] = list(plugin['dependencies'])
```
zachmullen: @danlamanna do you mind if I add a commit on this branch to make it so the conversion of sets to lists is done automatically in our REST layer? Then I will remove this change.

danlamanna: Not at all.

zachmullen:
@danlamanna I added my suggested change 0351656. Your changes LGTM except for that one too-long line. :+1:

zachmullen:
@danlamanna Fixed my test, feel free to merge if you approve my changes.

PR #1012, opened 2015-10-08
Record the mtime for imported files in FS adapter

zachmullen:
This may be useful info for downstreams wishing to implement
synchronization behavior.


danlamanna:
LGTM.

PR #1014, opened 2015-10-08
Fixes #1013. Fix event response semantics for model.upload.assetstore

### LINKED ISSUE #1013: model.upload.assetstore should use standard event semantics
Right now it expects responses to be written to the input params object, we should make it look in `event.responses`. The size quota plugin currently uses this event.

To preserve backward compatibility, we can fall back to checking the input params before falling back on the current assetstore.

zachmullen:


girder/models/upload.py
```diff
@@ -179,6 +179,28 @@ def finalizeUpload(self, upload, assetstore=None):
...
+        Get the assetstore for a particular target resource, i.e. where new
+        data within the resource should be stored. In girder core, this is
+        always just the current assetstore, but plugins may override this
+        behavior to allow for more granular assetstore selection.
+        """
+        eventParams = {'model': modelType, 'resource': resource}
+        event = events.trigger('model.upload.assetstore', eventParams)
+
+        if event.responses:
+            assetstore = event.responses[-1]
```
brianhelba: @zachmullen To be clear, the reasoning here is that if there are multiple listeners/handlers which each add their own response, we want to use the response from the most-recently-bound handler. Correct?

If so, I'd appreciate a comment here to explain that behavior. ``[-1]`` is a little too mysterious.

Otherwise, LGTM.

zachmullen: Yep, this is a convention we have in a few places. The reason is that events have a method called `stopPropagation` so that certain listeners can declare themselves authoritative and stop any further handlers from executing. In such cases, their added response would always be the last one.

brianhelba: If this isn't the only place we use the ``[-1]`` convention, then I suppose it's fine without a comment.

zachmullen:
Additionally, this makes it so that the furthest downstream plugins can override responses from their upstreams.

PR #1017, opened 2015-10-09
Install travis npm globally not on next branch

mgrauer:
Travis was breaking.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ before_install:
     - popd
     - cmake --version
 
-    - npm install -g npm@next
+    - npm install -g npm
     - npm --version
 install:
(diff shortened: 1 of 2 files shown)


zachmullen:
Suddenly flake8 stuff is also failing, looks like it's not taking our configuration parameters into account... @girder/developers anyone have ideas why that might be happening?

zachmullen:
@mgrauer I think I found a fix, is it OK to commit it onto this branch?

mgrauer:
:goberserk:

zachmullen:
I'm guessing, with absolutely no effort towards actually figuring out if it's true, that new style edicts were just backported to older versions of pep8.

zachmullen:
@mgrauer this LGTM, merge away if you are OK with my changes

mgrauer:
:maple_leaf:

PR #1019, opened 2015-10-13
Allow UploadWidget.startUpload() to be called directly

zachmullen:
Rather than requiring triggering by an external event


zachmullen:
@girder/developers can someone do a quick QA on this?

danlamanna:
LGTM

PR #1021, opened 2015-10-19
Flexible rendering options for the access control widget

zachmullen:


zachmullen:
re: COVALIC. @cpatrick @mgrauer

clients/web/src/stylesheets/widgets/accessWidget.styl
```diff
@@ -18,16 +20,15 @@
       color #d0b822
 
 .g-ac-list-title
-  background-color #666
+  background-color #888
```
manthey: Just an idle question: I've noticed a trend to making web pages lower contrast.  It doesn't matter much here, but is there some logic to it?

jeffbaumes: Not sure, but I get the feeling that `'#000'` is like the ALL-CAPS of the 90's - just too loud. Soft and friendly is where it's at.

jbeezley: I'm patiently waiting for the geocities look to come back in style.

manthey:
LGTM

PR #1023, opened 2015-10-20
Markdown editor widget with preview

zachmullen:


zachmullen:
This is ready for review (re: COVALIC). I plan to add additional functionality to this widget for inline image uploading, but would prefer to push that to another topic branch.

manthey:
Is the folder description shown anywhere besides the Edit Folder dialog?

It feels like this should be added to the Edit Item, Edit Collection, and Edit Group dialogs.

zachmullen:
> Is the folder description shown anywhere besides the Edit Folder dialog?

No, and it was only recently added there -- it wasn't shown at all before. I'd be fine opening an issue to show it more prominently if people want.

> It feels like this should be added to the Edit Item, Edit Collection, and Edit Group dialogs.

As of right now, those descriptions are not rendered as markdown, so I didn't give them markdown editors. We'd need to rethink how we display those since they currently show up as subtitles below the names of the resources. We can rebrand that as the "short description" and also expose a new "long description" field that is markdown, but I think whatever change we decide to make there, it should probably be on a new topic.

manthey:
LGTM.

We can wait until a user or project wants markdown elsewhere or more prominently to make other issues.

PR #857, opened 2015-05-28
Thumbnails

zachmullen:

PR #1024, opened 2015-10-20
Ensure that all text files end with a newline

brianhelba:
Since this encompasses every type of text file, it may not be possible to write a style test for.

Git, however, will show warnings when a file is committed without an ending newline. We should pay more attention to this.


zachmullen:
Some of our web client tests depended on the contents of those test files not having a newline, apparently :)

zachmullen:
That is, testFile.txt and testFile2

brianhelba:
@zachmullen I'm looking at this now, I'm not seeing where we actually use the contents of the file (its hash or its size) anywhere. The usages are:
* https://github.com/girder/girder/blob/master/clients/web/test/spec/dataSpec.js#L167
* https://github.com/girder/girder/blob/master/clients/web/test/spec/dataSpec.js#L285
* https://github.com/girder/girder/blob/master/clients/web/test/spec/itemSpec.js#L201

manthey:
If I recall, there is a phantomjs issue uploading control characters when
uploading via blobs, so having the newline makes the upload fail.  Rather
than trying to figure out how to make that work in phantom, I think we just
excluded the newlines.

On Tue, Oct 20, 2015 at 1:38 PM, Brian Helba <notifications@github.com>
wrote:

> @zachmullen <https://github.com/zachmullen> I'm looking at this now, I'm
> not seeing where we actually use the contents of the file (its hash or its
> size) anywhere. The usages are:
>
>    -
>    https://github.com/girder/girder/blob/master/clients/web/test/spec/dataSpec.js#L167
>    -
>    https://github.com/girder/girder/blob/master/clients/web/test/spec/dataSpec.js#L285
>    -
>    https://github.com/girder/girder/blob/master/clients/web/test/spec/itemSpec.js#L201
>
> —
> <https://github.com/girder/girder/pull/1024#issuecomment-149642975>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

brianhelba:
@manthey Can you please file an issue describing this PhantomJS issue? I'll remove those files from my commit, but we should have the capability to test-upload files with typical content (which definitely includes newlines).

manthey:
@brianhelba We do have the ability to test binary uploads -- we do it in the server tests.  In the client tests we have a work around.  For instance, in PR #857, @zachmullen added a method to upload binary files, but it doesn't quite exercise the same path as the normal uploads, where as the `girderTest.testUpload` method exercises the normal code paths, but is subject to phantomjs's issue.  Between the various tests we get close to full coverage.

zachmullen:
LGTM. If you use atom, the **whitespace** package can be configured to handle this automatically for you.

screen shot 2015-10-21 at 10 21 21 am

brianhelba:
@manthey I'm more concerned about future developers making the same mistake I did (particularly when writing a plugin). I could very well see a plugin developer wanting to do something more sophisticated with upload data on the client site, then uploading it. Any tests for this type of functionality would run into the same issue.

If it's a known bug, it should be documented, even if we have an adequate workaround in place now.

brianhelba:
@zachmullen Indeed, most GUI editors that I'm familiar with will do it, and vim / emacs can be scripted for this too.

PR #1025, opened 2015-10-20
Make model creation an atomic database operation

brianhelba:
In several places, a partially-created model is implicitly saved before the end of the "create..." function. This creates several problems.

First, if the code terminates at a later part of the "create..." function, the new database document will be left in an incomplete state.

Second, the premature save operation causes the "model.{}.save.created" event to be fired before the "create..." function has finished. This causes callbacks to receive an incomplete model.

Most significantly, since the remainder of the "create..." code is still executed unconditionally after the "model.{}.save.created" callback completes, it may undo some of the intentional functionality of the callback. In particular, a callback that sets a new user's "public" attribute to False would have this change undone.

This also moves the creation of default user folders into an event callback. This gives developers the option of unregistering this callback if they don't want users to have default folders. A user-facing option to specify this behavior could be added in the future.

### DIFF
--- a/girder/events.py
+++ b/girder/events.py
@@ -41,6 +41,7 @@ function to be called when the task is finished. That callback function will
 receive the Event object as its only argument.
 """
 
+import contextlib
 import threading
 import types
 
@@ -206,10 +207,28 @@ def unbind(eventName, handlerName):
 
 def unbindAll():
     """
-    Clears the entire event map. Any bound listeners will be unbound.
+    Clears the entire event map. All bound listeners will be unbound.
+
+     .. warning:: This will also disable internal event listeners, which are
+       necessary for normal Girder functionality. This function should generally
+       never be called outside of testing.
     """
     global _mapping
     _mapping = {}
+
+
+@contextlib.contextmanager
+def bound(eventName, handlerName, handler):
+    """
+    A context manager to temporarily bind an event handler within its scope.
+
+    Parameters are the same as those to :py:func:`girder.events.bind`.
+    """
+    bind(eventName, handlerName, handler)
+    try:
+        yield
+    finally:
+        unbind(eventName, handlerName)
 
 
 def trigger(eventName, info=None, pre=None, async=False):
--- a/girder/models/group.py
+++ b/girder/models/group.py
@@ -20,6 +20,7 @@
 import datetime
 
 from .model_base import AccessControlledModel, ValidationException
+from girder import events
 from girder.constants import AccessType
 
 
@@ -64,6 +65,9 @@ class Group(AccessControlledModel):
         self.exposeFields(level=AccessType.READ, fields=(
             '_id', 'name', 'public', 'description', 'created', 'updated',
             'addAllowed'))
+
+        events.bind('model.group.save.created', 'grantCreatorAccess',
+                    self._grantCreatorAccess)
 
     def filter(self, group, user, accessList=False, requests=False):
         """
@@ -333,6 +337,7 @@ class Group(AccessControlledModel):
         group = {
             'name': name,
             'description': description,
+            'creatorId': creator['_id'],
             'created': now,
             'updated': now,
             'requests': []
@@ -340,14 +345,21 @@ class Group(AccessControlledModel):
 
         self.setPublic(group, public=public)
 
-        # Now validate and save the group
-        self.save(group)
+        return self.save(group)
 
-        # We make the creator a member of this group and also grant them
-        # admin access over the group.
+    def _grantCreatorAccess(self, event):
+        """
+        This callback makes the group creator an administrator member of the
+        group.
+
+        This generally should not be called or overridden directly, but it may
+        be unregistered from the `model.group.save.created` event.
+        """
+        group = event.info
+        creator = self.model('user').load(group['creatorId'], force=True,
+                                          exc=True)
+
         self.addUser(group, creator, level=AccessType.ADMIN)
-
-        return group
 
     def updateGroup(self, group):
         """
--- a/girder/models/group.py
+++ b/girder/models/group.py
@@ -21,7 +21,7 @@ import datetime
 
 from .model_base import AccessControlledModel, ValidationException
 from girder import events
-from girder.constants import AccessType
+from girder.constants import AccessType, CoreEventHandler
 
 
 class Group(AccessControlledModel):
@@ -66,7 +66,8 @@ class Group(AccessControlledModel):
             '_id', 'name', 'public', 'description', 'created', 'updated',
             'addAllowed'))
 
-        events.bind('model.group.save.created', 'grantCreatorAccess',
+        events.bind('model.group.save.created',
+                    CoreEventHandler.GROUP_CREATOR_ACCESS,
                     self._grantCreatorAccess)
 
     def filter(self, group, user, accessList=False, requests=False):
--- a/girder/models/group.py
+++ b/girder/models/group.py
@@ -343,7 +343,7 @@ class Group(AccessControlledModel):
             'requests': []
         }
 
-        self.setPublic(group, public=public)
+        self.setPublic(group, public, save=False)
 
         return self.save(group)
 
--- a/girder/models/user.py
+++ b/girder/models/user.py
@@ -23,7 +23,7 @@ import re
 
 from .model_base import AccessControlledModel, ValidationException
 from girder import events
-from girder.constants import AccessType
+from girder.constants import AccessType, CoreEventHandler
 from girder.utility import config
 
 
@@ -47,9 +47,10 @@ class User(AccessControlledModel):
         self.exposeFields(level=AccessType.ADMIN, fields=(
             'size', 'email', 'groups', 'groupInvites'))
 
-        events.bind('model.user.save.created', 'grantSelfAccess',
-                    self._grantSelfAccess)
-        events.bind('model.user.save.created', 'addDefaultFolders',
+        events.bind('model.user.save.created',
+                    CoreEventHandler.USER_SELF_ACCESS, self._grantSelfAccess)
+        events.bind('model.user.save.created',
+                    CoreEventHandler.USER_DEFAULT_FOLDERS,
                     self._addDefaultFolders)
 
     def filter(self, user, currentUser):
--- a/girder/models/user.py
+++ b/girder/models/user.py
@@ -275,9 +275,9 @@ class User(AccessControlledModel):
         This generally should not be called or overridden directly, but it may
         be unregistered from the `model.user.save.created` event.
         """
-        default_folder_setting = self.model('setting').get(
-            SettingKey.USER_DEFAULT_FOLDERS, 'public_private')
-        if default_folder_setting == 'public_private':
+        if self.model('setting').get(
+                SettingKey.USER_DEFAULT_FOLDERS, 'public_private') \
+                == 'public_private':
             user = event.info
 
             publicFolder = self.model('folder').createFolder(
--- a/girder/models/user.py
+++ b/girder/models/user.py
@@ -22,6 +22,7 @@ import os
 import re
 
 from .model_base import AccessControlledModel, ValidationException
+from girder import events
 from girder.constants import AccessType
 from girder.utility import config
 
@@ -45,6 +46,11 @@ class User(AccessControlledModel):
             'created'))
         self.exposeFields(level=AccessType.ADMIN, fields=(
             'size', 'email', 'groups', 'groupInvites'))
+
+        events.bind('model.user.save.created', 'grantSelfAccess',
+                    self._grantSelfAccess)
+        events.bind('model.user.save.created', 'addDefaultFolders',
+                    self._addDefaultFolders)
 
     def filter(self, user, currentUser):
         """Preserved override for kwarg backwards compatibility."""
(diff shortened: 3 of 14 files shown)


brianhelba:
This is WIP because I need help on how to move the granting of admin to group creators into an event callback.

We need to make this change for 2 reasons:
1. So that the "createGroup" function is database-atomic, only calling "save" once all work has been completed.
2. To permit the use case where a plugin can listen for the "model.group.save.created" event, then remove the user's admin access and possibly grant admin to another user instead.

However, since "createdBy" is not a field stored on a group, I don't see an easy way to get the creating user into the callback:
1. adding a permanent "createdBy" field would be inconsistant with existing deployments, where this value isn't set
2. adding a temporary "createdBy" to the document, which the callback then removes is messy, and has the potential to remain attached longer than intended if the callback doesn't happen for any reason
3a. it's difficult to add an additional parameter with the creator to the "model.group.save.created" event, since this event is emitted in the "Model.save" function; there is a workaround of doing the same thing as the "createUser" function (adding an "_id" and triggering the event manually), but this is still a bad idea because...
3b. any existing handlers listening for "model.group.save.created" will expect "event.info" to be exactly the group object, so we can't attach other objects (the creator id) to the event

girder/models/group.py
```diff
@@ -338,17 +342,24 @@ def createGroup(self, name, creator, description='', public=True):
...
+
+    def _addCreatorUser(self, event):
+        """
+        This callback grants admin access to the creator of a group.
 
+        This generally should not be called or overridden directly, but it may
+        be unregistered from the `model.group.save.created` event.
+        """
+        group = event.info
+        creator = None # TODO: how do we get this...
```
brianhelba: Help me here please.

zachmullen:
In all the places where you added `save=False`, it was already the default.

brianhelba:
@zachmullen Yes, but explicit is better when we require on the behavior to be that way. It's a warning flag to future developers that the parameter matters.

girder/models/user.py
```diff
@@ -244,26 +249,43 @@ def createUser(self, login, password, firstName, lastName, email,
...
         }
 
-        self.setPassword(user, password)
+        self.setPassword(user, password, save=False)
+
+        self.setPublic(user, public, save=False)
+
+        # An id is required in order to grant the user access on itself, but
+        # since the user isn't fully created yet, it should not be saved
+        user['_id'] = ObjectId()
```
zachmullen: Have you investigated to make sure this doesn't cause any weirdness as a result of pymongo conditioning behavior on the presence of an _id field? https://github.com/mher/pymongo/blob/master/pymongo/collection.py#L210-L215

girder/models/user.py
```diff
@@ -244,26 +249,43 @@ def createUser(self, login, password, firstName, lastName, email,
...
+        self.setUserAccess(user, user, level=AccessType.ADMIN, save=False)
+
+        self.save(user)
```
zachmullen: We should reassign the return value of `save` back into `user`; save is allowed to modify the values of the document, but the contract does not guarantee that it will do so in-place. It *does* guarantee that it will return the modified document, though.

girder/models/user.py
brianhelba: @zachmullen A bit, but not enough.

My initial concern was actually with ObjectId timestamps being off if the database was running on another machine, but if you go one level deeper into ``save``, the [``_id`` is actually being set](https://github.com/mongodb/mongo-python-driver/blob/2.8/pymongo/collection.py#L385) in the PyMongo client anyway. It's the same in v3.1 (the repo you linked in a really old fork of PyMongo).

The bigger problem, which I initially missed, is indeed that the ``collection.update`` will likely do nothing, since there's no matching document with that ``_id``. Reimplementing the entire ``Model.save`` functionality right here just to explicitly call ``collection.insert`` instead is bad design. I'm also not thrilled with the idea of moving the ``User.setUserAccess`` into a callback, since we don't currently support the case of a user with no access to themselves, so you wouldn't ever want that code to not run.

Any ideas?

zachmullen: Anything I can think of is pretty convoluted. I'm tempted to just not try and fix this at all since the database state would be really easy for an admin to fix via the UI in the rare event that the initial query succeeds and the second fails.

zachmullen: If you think it's important, though, I think the best idea would be to add logic to `AccessControlledModel.hasAccess` that special-cases for users and groups at runtime and automatically returns True if the user is checking their own document, or for a group to which they belong. It might be cleanest to override `hasAccess` in the group and user models.

brianhelba: What I really care about is that callbacks listening for ``model.{}.save.created`` receive a completely created object, with no extra creation work still pending to execute.

Note, what originally bit me is that I had a callback to set all newly-created users to be private. However, since the ``model.user.save.created`` was first triggered inside ``setPassword``, the work of my callback function was always undone.

The notion of atomic creation is nice, but not as practically important IMO.

girder/models/user.py
brianhelba: Will do.

girder/models/user.py
zachmullen: That's a good point. Then I vote for adding the override of `hasAccess` in the user model so we can totally avoid calling `setUserAccess` post-creation.

brianhelba:
I've just pushed a version that should now accomplish the most pressing goal of having all ``model.{}.save.created`` events fire at the "correct" time.

User and group model creation is no longer strictly database-atomic, but those models and others are still much better than they used to be on this front.

If any @girder/developers can think of an even better solution, I'm still open to one.

brianhelba:
The user's "Public" folder isn't getting created in the test. This functionality was working for me on this branch locally.

brianhelba:
I'm still not happy with this branch yet.

The final commit is just a quick hack ATM. It needs to be more carefully considered to ensure no testing coverage / effectiveness is lost.

I don't see a way to get ``createUser`` to a satisfactory state. The options are:
1. Call the first ``save`` with ``triggerEvents=False``. This means that when ``model.user.save`` eventually is triggered, handlers have no opportunity to effectively use ``preventDefault`` and prevent the new document from being inserted in the database. Worse, if a handler did do ``preventDefault``, only the call to ``setUserAccess`` would be nullified, leaving the user document in an incorrect state.
2. Call the first ``save`` with ``triggerEvents=True``. This means that when all the ``model.user.save.*`` are first called, they receive an incomplete user document. Additionally, ``model.user.save.created`` handlers have no ability to prevent or modify the work that ``setUserAccess`` will always do afterwards.
3. Reimplement all of ``Model.save`` inside ``createUser``. This is bad engineering as it, destroys the ability for users to override and virtually call ``save`` in inherited classes and is much less maintainable.

I might normally discount these issues as minor, but one of my own plugins relies on being able to reliably manipulate new users in ``model.user.save.created`` in a way that this currently still fails to allow.

@girder/developers I'd be happy to get anyone's thoughts.

tests/cases/folder_test.py
```diff
@@ -38,7 +38,7 @@ def tearDownModule():
 
 class FolderTestCase(base.TestCase):
     def setUp(self):
-        events.unbindAll()
```
brianhelba: @zachmullen Is this safe, or do we need to more selectively remove certain events here? The code was added in your commit https://github.com/girder/girder/commit/d2052b2be2767fe39d8d23f2cc023d93c9e097d3#diff-2e7bf9531d7542fbed2103c654d83bc5R40 .

brianhelba: Note, adding the comment is a quick hack. I'll actually remove the line once this is settled.

brianhelba:
I'm finally really satisfied with this PR. PTAL @girder/developers

The short version of this branch is:
* This now makes all the ``model.{}.save.created`` events fire at the appropriate time, at the end of model creation.
* Migration functions save only once, at the very end of their logic.
* Some optional model behaviors can now be be disabled by plugins, specifically:
 * new users having admin access on themselves
 * new users having default "Public" and "Private" folders created
 * new groups having their creator get membership as an admin
* Groups now store their creator
* A new event binding context manager is available

The longer version (from the commit messages) is:
* Move some non-core parts of user creation into an event callback
 * Specifically, the granting of user admin access to itself and creation of default user folders are now done immediately after "createUser" returns.
 * This gives developers the option of unregistering these callback if they don't want either behavior.
 * This also facilitates the changes in a subsequent commit.
*  Make group models store their creator
 * Also move the granting of admin access to a new group into an event callback. This gives developers the option of unregistering this callback if they don't want users to have admin membership to groups they create.
 * This also facilitates the changes in a subsequent commit.
* Make model creation an atomic database operation
 * In several places, a partially-created model is implicitly saved before the end of the "create..." function. This creates several problems.
 * First, if the code terminates at a later part of the "create..." function, the new database document will be left in an incomplete state.
 * Second, the premature save operation causes the "model.{}.save.created" event to be fired before the "create..." function has finished. This causes callbacks to receive an incomplete model.
 * Most significantly, since the remainder of the "create..." code is still executed unconditionally after the "model.{}.save.created" callback completes, it may undo some of the intentional functionality of the
callback. For example, a callback that sets a new user's "public" attribute to False would have this change undone.
* Make group models store their creator
 * Also move the granting of admin access to a new group into an event callback. This gives developers the option of unregistering this callback if they don't want users to have admin membership to groups they create.
 * This also facilitates the changes in a subsequent commit.
* Make testModelSaveHooks more rigorous
 * Now that "model.user.save.*" is only fired once per user creation, the tests can be stricter.

girder/events.py
```diff
@@ -206,12 +207,30 @@ def unbind(eventName, handlerName):
 
 def unbindAll():
     """
-    Clears the entire event map. Any bound listeners will be unbound.
+    Clears the entire event map. All bound listeners will be unbound.
+
+     .. warning:: This will also disable internal event listeners, which are
+       necessary for normal Girder functionality. This function should generally
+       never called outside of testing.
```
jcfr: `... never called ...` -> `.. never be called ..`

brianhelba: Fixed.

girder/events.py
```diff
@@ -206,12 +207,30 @@ def unbind(eventName, handlerName):
...
+@contextlib.contextmanager
+def bound(eventName, handlerName, handler):
+    """
+    A context manager to temporarily bind an event handler within its scope.
+
+    Parameters are the same as those to :py:func:`girder.events.bind`.
+    """
+    bind(eventName, handlerName, handler)
+    try:
+        yield
```
zachmullen: Should the function call be inside the `try` block?

brianhelba: No, the goal is pause on any exceptions that happen in user code, just to unbind the event.

Note that there is no ``except`` clause in this block, since all exceptions should be passed on up the call stack. If there was an exception inside ``bind`` (unlikely), there's no reason to call ``unbind`` (things would already be too screwed up at that point).

girder/models/user.py
```diff
@@ -46,6 +47,11 @@ def initialize(self):
         self.exposeFields(level=AccessType.ADMIN, fields=(
             'size', 'email', 'groups', 'groupInvites'))
 
+        events.bind('model.user.save.created', 'grantSelfAccess',
+                    self._grantSelfAccess)
+        events.bind('model.user.save.created', 'addDefaultFolders',
+                    self._addDefaultFolders)
```
zachmullen: Our convention for event binding names has always been to make them prefixed with the name of the plugin. Now that we have them in core, should we change these names to e.g. `core.grantSelfAccess`?

zachmullen:
Can we add a test that unbinds the core `addDefaultFolders` handler and creates a user and then verifies that they don't have folders? (Might necessitate an `unbound` context manager as well...)

girder/models/user.py
zachmullen: And should we consider making these names attributes under an enum in `girder.constants`?

jcfr: > an enum in `girder.constants`

Great idea. May be this could be done in a follow up PR and applies to the entire code base ?

zachmullen: If we're going to make them an enum of constants, we may as well do it now before people have the chance to use the raw strings instead :)

jcfr: :+1:  Whatever will make the review easy will work with me. Now or later.

brianhelba: @zachmullen If you have a clear idea of what you'd like to see for names in ``constants``, please feel free to add a commit on top of this PR. I don't have any real opinion on this matter.

Otherwise, I can try to do it myself, but I don't yet have a clear vision for what I'd like the particular organization to be.

girder/models/user.py
```diff
@@ -46,6 +47,11 @@ def initialize(self):
         self.exposeFields(level=AccessType.ADMIN, fields=(
             'size', 'email', 'groups', 'groupInvites'))
 
+        events.bind('model.user.save.created', 'grantSelfAccess',
```
zachmullen: We had discussed simply removing this and modifying `AccessControlledModel.hasAccess` to always return `True` if a user is checking access on themselves. Did you decide against making that change?

zachmullen:
@brianhelba be advised, I pushed a commit onto your topic. It's just the first thing I thought of, if you want to change or improve it, feel free.

girder/models/user.py
brianhelba: As I mentioned on the call today, my thinking is that it may be desirable to not always have users with admin on themselves. I could imaging a tightly managed Girder instance where users could be created via a central authority and are not allowed to change their own name or email address.

With that in mind, it no longer becomes an intrinsic precondition that users must have admin access on themselves, so we are no longer forced to do that operation inside ``createUser`` (and indeed we shouldn't, in case someone wants to unbind that event). Thus, I think moving ``setUserAccess`` to a post-create callback works quite elegantly.

brianhelba:
@zachmullen also be advices, I just responded to your question about user admin access, but it's now on an outdated diff. If you have any followup comments, lets continue the discussion on the latest revsion.

brianhelba:
@zachmullen Your commit LGTM.

girder/models/user.py
zachmullen: Sounds good.

brianhelba:
@zachmullen PTAL

zachmullen:
There's a problem with the setting UI; when I go to the page for the first time (when no setting for the new key is present in the database), the default selected is "No", when it should be set to the default value of create public and private folders.

brianhelba:
So, in [systemConfiguration.jade](https://github.com/girder/girder/blob/model-creation/clients/web/src/templates/body/systemConfiguration.jade), it looks like ``core.registration_policy`` and ``core.add_to_group_policy`` have the same bug too, where the default is not being selected. However, since the real default value is coincidentally the first item in the dropdown, it's not actually apparent.

I used ``core.registration_policy`` as a template for ``core.user_default_folders``, but I had the default value as the second item in the dropdown.

I'll fix all three now, but this code is starting to get a bit repetitive. It would be really nice if we had more sophisticated default settings for basic dropdowns and text fields, so there's not so much copy-paste boilerplate to create a very simple setting.

zachmullen:
I'm fine with just reversing the order of the dropdown so the default is the first entry, since that does technically work, if you'd rather just do that.

:+1: For reducing boilerplate load.

brianhelba:
@zachmullen Changing the order is just disguising the bug. If the default is ever changed server-side, the client should automatically use and respect it.

brianhelba:
As much as I had to open this can of worms, to be honest, this bug with settings widgets not all respecting defaults reveals a gap in our web client testing.

This PR has already had so much scope creep, I'm really hesitant to try and pack more in. Also, I'm not personally very familiar with the client-side testing, so it will delay the PR even longer until I can get a test written.

I'm going to file an issue to add a test, and reference the commit in this branch that fixes the issue.

brianhelba:
@girder/developers PTAL. I hope this is the final version.

zachmullen:
That's true.

zachmullen:
LGTM

PR #1028, opened 2015-10-21
Markdown widget image upload

zachmullen:
Depends on #1023

### DIFF
--- a/clients/web/src/stylesheets/layout/layout.styl
+++ b/clients/web/src/stylesheets/layout/layout.styl
@@ -19,6 +19,7 @@
   top 65px
   right 10px
   width 320px
+  z-index 11000
 
   .alert
     display none
--- a/clients/web/src/stylesheets/layout/layout.styl
+++ b/clients/web/src/stylesheets/layout/layout.styl
@@ -19,7 +19,7 @@
   top 65px
   right 10px
   width 320px
-  z-index 11000
+  z-index $alertsZ
 
   .alert
     display none
@@ -48,7 +48,7 @@
   right 10px
   bottom 0
   width 400px
-  z-index 10000
+  z-index $progressZ
 
 .modal-body p
   overflow hidden
--- a/clients/web/src/stylesheets/widgets/markdownWidget.styl
+++ b/clients/web/src/stylesheets/widgets/markdownWidget.styl
@@ -56,7 +56,10 @@ $hoverBg = #fbfac0
   .g-markdown-preview
     min-height 80px
     padding 0 5px
-    overflow-x scroll
+    overflow-x auto
+
+    p
+      overflow-x auto  // necessary to override bootstrap styling
 
   .g-markdown-explanation
     display inline-block
--- a/clients/web/src/stylesheets/widgets/markdownWidget.styl
+++ b/clients/web/src/stylesheets/widgets/markdownWidget.styl
@@ -1,5 +1,8 @@
 @import 'nib'
 
+$hoverBorderColor = #aaa
+$hoverBg = #fbfac0
+
 .g-markdown-widget
   .g-tabs-container
     border-radius 5px
@@ -21,18 +24,34 @@
 
   .g-markdown-text
     height 100px
-    margin-bottom 4px
     padding 5px 9px
     background-color #fafafa
+    border 1px solid #ccc
     box-shadow inset 0 1px 2px rgba(0,0,0,0.075)
 
     &:focus
       background-color #fff
 
-    &[upload-enabled=true]
+    &[upload-enabled]
       border-bottom-left-radius 0
       border-bottom-right-radius 0
+      border-bottom 1px dashed #ccc
 
+    &.dragover
+      background-color $hoverBg
+      border-color $hoverBorderColor
+
+  .g-upload-footer
+    background-color #fafafa
+    color #777
+    border 1px solid #ccc
+    border-top none
+    padding 4px 10px
+    border-bottom-left-radius 3px
+    border-bottom-right-radius 3px
+
+    &.dragover
+      border-color $hoverBorderColor
 
   .g-markdown-preview
     min-height 80px
@@ -45,3 +64,27 @@
     right 14px
     bottom 15px
     font-size 12px
+
+  .g-upload-progress-container
+    margin-top 2px
+
+    .g-markdown-upload-progress
+      margin-bottom 9px
+
+      .progress-bar
+        transition none
+
+  .g-markdown-drop-zone
+    position relative
+
+  .g-upload-overlay
+    position absolute
+    top 0
+    right 0
+    bottom 0
+    left 0
+    padding 15px
+    background-color rgba(75,75,75,0.77)
+    border-radius 3px
+    z-index 2
+    color white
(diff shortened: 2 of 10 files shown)


zachmullen:
Ready for review.

manthey:
This uploads images to the current folder.  This is not surprising to me, but I could see it be surprising to a different user.  The folder contents should be updated if an image is uploaded so that the view is accurate when the description dialog is closed.

The images are uploaded upon selection, and persist even if the description change is cancelled. It is more complicated, but it would seem 'more correct' to have these images be deleted if the dialog was cancelled.

The interface says "Add images", but only allows adding one image at a time.  Perhaps it would be clearer to read "Add an image by clicking here, or drag and drop one into the editor".

Is there a reason why fake.jpg is a fake jpeg?  It superficially seems to work with a real jpeg.  I do enjoy the Magritte-like file, though.

zachmullen:
> Is there a reason why fake.jpg is a fake jpeg? It superficially seems to work with a real jpeg. I do enjoy the Magritte-like file, though.

I was under the impression we couldn't upload binary files via phantom, so I faked it since it still exercises all the important behavior.

> The images are uploaded upon selection, and persist even if the description change is cancelled. It is more complicated, but it would seem 'more correct' to have these images be deleted if the dialog was cancelled.

Yeah, I've thought about how to do this, but it's a little tricky depending on exactly what policies we want to enforce. For the moment I can justify this simple behavior by saying it's just a convenience for actions the user could already perform anyway: uploading + editing a description. But I agree that it may subvert their expectations in the case where they cancel. If we want more advanced/intuitive behavior, my vote would be to defer it to a future topic.

clients/web/src/stylesheets/layout/layout.styl
```diff
@@ -19,6 +19,7 @@
   top 65px
   right 10px
   width 320px
+  z-index 11000
```
manthey: I assume this is just to make sure it is above progress messages. 

Would it be better to have this set computationally?  That is, `$headerZ+2` instead of `11000` and `$headerZ+1` instead of `10000` elsewhere in the file (or define a different range for such values)?

clients/web/src/stylesheets/widgets/markdownWidget.styl
```diff
@@ -21,22 +24,39 @@
...
+    border-bottom-left-radius 3px
+    border-bottom-right-radius 3px
 
+    &.dragover
+      border-color $hoverBorderColor
 
   .g-markdown-preview
     min-height 80px
     padding 0 5px
+    overflow-x scroll
```
manthey: Can we make this `auto` so it only appears as needed?

manthey: When I upload a very wide image, the scroll bar isn't enabled.  Wide text wraps (breaks in the middle of a word), so I'm not sure what value this has.

clients/web/src/views/widgets/MarkdownWidget.js
```diff
@@ -4,7 +4,140 @@
...
+
+        'drop .g-markdown-drop-zone': function (e) {
+            e.stopPropagation();
+            e.preventDefault();
+            this.$('.g-markdown-text,.g-upload-footer').removeClass('dragover');
+            this.files = e.originalEvent.dataTransfer.files;
+            this.filesAdded();
+        }
+    },
+
```
manthey: None of the drag-and-drop functionality has coverage (neither here nor our existing drag-and-drop).  A short time ago I read that we should be able to test this in phantom; perhaps create a new issue to test drag-and-drop code.

clients/web/src/views/widgets/EditFolderWidget.js
```diff
@@ -28,8 +28,13 @@ girder.views.EditFolderWidget = girder.View.extend({
             text: this.folder ? this.folder.get('description') : '',
             prefix: 'folder-description',
             placeholder: 'Enter a description',
+            parent: this.folder,
+            allowedExtensions: ['png', 'jpg', 'jpeg'],
```
manthey: Is this the real format limitation?  Or could we include gif and others?

zachmullen: We could add anything that can be rendered in an `img` tag, I just picked this set for now. Do you think we should add other extensions?

clients/web/src/views/widgets/MarkdownWidget.js
zachmullen: #1041

clients/web/src/stylesheets/widgets/markdownWidget.styl
zachmullen: I saw this in the modal dialog but not in the covalic app where it's not inside a modal, then tracked down why. This is from bootstrap:

```
.modal-body p {
    overflow: hidden;
    text-overflow: ellipsis;
}
```

I'll add some overrides.

clients/web/src/views/widgets/MarkdownWidget.js
```diff
@@ -13,19 +146,35 @@ girder.views.MarkdownWidget = girder.View.extend({
...
+     *     certain file extensions, set this to a list of (lowercase) extensions,
+     *     not including periods. E.g.: ["png", "jpg", "jpeg"]
      */
     initialize: function (settings) {
         this.text = settings.text || '';
         this.placeholder = settings.placeholder || '';
         this.prefix = settings.prefix || 'markdown';
+        this.enableUploads = settings.enableUploads || false;
+        this.parent = settings.parent || null;
+        this.maxUploadSize = settings.maxUploadSize || null;
```
manthey: I'm not sure the default of this should be `null`.  Though I suppose only people with access to upload to the folder can edit this anyway, so it doesn't add any protection really.

Amusingly, I tried to upload a 4Gb file (renamed to .jpg).  This uploaded and now is downloading for the preview, at which point my Chrome tab died.  I'd recommend a default of 10 Mbytes (anything larger used to kill iOS 5, perhaps it is higher today).

manthey:
Phantom does strange things to binary files (for instance, it converts LF to CRLF), so if we compare the file, the tests fail.  In this instance, we could upload anything, and, since we don't use the results, it will pass.  For this purpose, a jpeg would work (even though it would get corrupted in the upload process).

I'm fine leaving images when the user cancels for now.  Since the images now show up in the folder contents when you upload them, it feels less strange then it did before.

clients/web/src/views/widgets/EditFolderWidget.js
manthey: I would add gif, because it still is used as a common web format.

zachmullen:
@manthey PTAL, I think I covered everything

clients/web/test/spec/folderSpec.js
```diff
@@ -57,9 +54,38 @@ function _editFolder(button, buttonText, testValidation)
...
         runs(function () {
             $('#g-name').val(oldval);
+            girderTest.sendFile('clients/web/test/testFile.txt',
+                                '.g-markdown-drop-zone .g-file-input');
+        });
+
+        waitsFor(function () {
+            return $('#g-alerts-container .alert-danger').text().indexOf(
+                'Only files with the following extensions are allowed: ' +
+                'png, jpg, jpeg.') !== -1;
```
manthey: You need to add `, gif` here.

manthey:
LGTM.

PR #1029, opened 2015-10-21
Move to latest pymongo major version

zachmullen:


zachmullen:
In doing this conversion I discovered what I consider a bug in our logic that I've fixed. Previously, we called ensure_index() on GridFS collections every time the adapter was instantiated. Now, we only call that when the assetstore document is saved.

This always *should* have broken the assetstore test -- it should be an exception to call ensure_index on a replica set that does not exist. However, it was apparently not in older versions of pymongo.

zachmullen:
@manthey PTAL, this touches a lot of code that you understand better than I do probably. Namely, that last commit makes me slightly nervous, and I want to make sure that we're using MongoProxy in a sensible way.

Good news -- this should hopefully fix some of the intermittent test failures we see on travis in web_client_item and server_upload.

zachmullen:
For instance, should we implement `__next__` in MongoProxy and have it also do auto-retry behavior (for cursors)?

manthey:
We are using an earlier version of https://github.com/arngarden/MongoDBProxy for our proxies.  The latest version has a "durable cursor" for auto-retry behavior in the cursor.  We should upgrade to the latest version of this, too.

We currently have placed a snapshot of this in the `girder/external` directory.  We can do the same, but perhaps there is cleaner way to pull in external github projects that aren't in pip.

zachmullen:
From looking at their DurableCursor, it's unclear from me how we would easily plug that into our code. The __init__ doesn't take an existing cursor, and we never manually instantiate a Cursor object ourselves. Maybe I'm missing something?

zachmullen:
It's also worth noting that their DurableCursor would only work on pymongo 2.x, they are using the old kwargs signature that is no longer compatible in 3.x. It may be worth having a custom cursor wrapper that takes a lot of the functionality from theirs, but I think we'd have to basically roll our own version of it to wrap an existing cursor.

manthey:
Looking at the difference in `mongodb_proxy.py` between the version we are using at the latest in the arngarden repo, they have extended the proxy to support the connection objects (which are not in pymongo 3) and explicitly removed cursors from the proxy (moving them to the durable cursor code).  Otherwise, changes include trying to reconnect an extra time.

We could just update `mongodb_proxy.py` to the latest version, but not pull in any of their other code (which is just a reorganization to support the durable cursor).

I think we should add our own cursor wrapper for reconnection.

girder/models/__init__.py
```diff
@@ -58,14 +69,17 @@ def getDbConnection(uri=None, replicaSet=None):
         uri = dbConf.get('uri')
         replicaSet = dbConf.get('replica_set')
     clientOptions = {
-        'connectTimeoutMS': 15000,
```
manthey: The default for this is 20000 ms, and has been officially since pymongo 2.7.  We were using that version when I added this parameter, and at the time the default was not being honored.  Do we know for certain that the default is now being used so that it is safe to remove this?

zachmullen: Hm, I'll reintroduce it in case that's still true. Thanks.

manthey:
A couple of missed items:

model_base.py, line 311, `save()` should either be `insert_one()` or `replace_one()`.

gridfs_assetstore_adapter.py, line 158, `insert()` should be `insert_one()`.
  line 177, `remove()` should be `remove_many()`
  line 282, `remove()` should be `remove_many()`
  line 292, `remove()` should be `remove_many()`

manthey:
I should have typed `delete_many` not `remove_many`.

zachmullen:
There we go, should be all sorted.

Do you want me to try and write the cursor proxy on this branch, or a future one?

manthey:
I'd say a future one.  I'll note that if we upgrade to the latest `mongodb_proxy.py`, we'd have to add or change code to avoid deprecated functions, but the version we have checked in should be okay as it is.

One more question: in model_base.py, should `findOne` filter kwargs?  Internally it calls `find()`, so we could have the same issue.  Though to be generic, you'd want to handle additional kwargs such as skip (or rename offset), limit (though it is sort-of ignored),  and timeout as well.

If you don't think we should filter kwargs in `findOne` (no code is using them), then LGTM.

zachmullen:
I added the filtering, but didn't expand to allow skip/timeout. My thought is that `findOne` should really only be used if you know *a priori* that only one document is going to match, and `find` should be used for everything else.

manthey:
LGTM.

PR #1030, opened 2015-10-22
Give the "limit" parameter of "filterResultsByPermission" a default value

brianhelba:
The other methods in the model API also have default values for "limit".

Note, the model API has defaults of "limit=0" for all of its functions. The HTTP API uses use a default of "limit=50".


brianhelba:
PTAL.

zachmullen:
LGTM.

PR #1031, opened 2015-10-22
Make "copyAccessPolicies" do a deep copy

brianhelba:
Previously, a call to "Folder.createFolder" had the possible side effect of setting additional ACLs on the "parent" object too.


brianhelba:
PTAL.

zachmullen:
LGTM

PR #1033, opened 2015-10-22
Fix a quota issue with large quotas.

manthey:
A comparison was being made to ensure that the quota was an integer.  In python 2, isinstance(quota, int) will fail if quota is a long, which happens when mongo stores the value as a NumberLong.  Python 3 was working properly.


zachmullen:
LGTM. @danlamanna I'll deploy this to data.kitware.com once it's merged. Thanks @manthey

PR #1036, opened 2015-10-24
Improvements to templated endpoints

brianhelba:
* Use the more appropriate PACKAGE_DIR constant in some places 
*  Move Webroot and Description (Swagger root) templates to separate files
 * This will make future maintenance of these files easier, as they are no longer string literals constrained by Python style restrictions.
 * Since the new implementation stores the HTML string as an instance variable, instead of a class variable, it is possible to instantiate multiple Webroots with different templates in the same Girder instance.
 * This also slightly changes whitespace in the HTML template files, to improve their style once they are actually sent to a client.
* Move template serving endpoint code to a common class
 * This reduces code duplication and makes it easier to add additional templated endpoints in the future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -116,7 +116,9 @@ setup(
             'girder-version.json',
             'conf/girder.dist.cfg',
             'mail_templates/*.mako',
-            'mail_templates/**/*.mako'
+            'mail_templates/**/*.mako',
+            'utility/webroot.mako',
+            'api/api_docs.mako'
         ]
     },
     install_requires=reqs,
--- a/girder/api/api_docs.mako
+++ b/girder/api/api_docs.mako
@@ -1,0 +1,56 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>${title}</title>
+    <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Droid+Sans:400,700">
+    <link rel="stylesheet" href="${staticRoot}/lib/fontello/css/fontello.css">
+    <link rel="stylesheet" href="${staticRoot}/built/swagger/css/reset.css">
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -116,7 +116,8 @@ def girder_install(self, component):
             'girder-version.json',
             'conf/girder.dist.cfg',
             'mail_templates/*.mako',
-            'mail_templates/**/*.mako'
```
zachmullen: I'm not sure we should get rid of this inclusion; in the future we may want to have subdirectories of mail templates and we'd want them to get sucked in automatically.

brianhelba: I suppose you could say this about a lot of places though. For right now, it's pointless cruft.

If anyone does want to add subdirectories in the future, they can just update the install script again accordingly; both these changes would have to be part of a commit to Girder proper, so this doesn't impact plugins or any development API.

zachmullen: That's if they remember to do so. Our test suite runs in a git clone environment, not the packaged environment, so it would be easy for someone to forget, and then we'd hear about it at deploy time when someone with a pip-installed server is seeing 500 exceptions when their server tries to send out emails.

Basically, the cost of having this line in there as protection against that future possibility is worth it to me. The cost is one extra line of code, the benefit is avoiding this possible headache in the future. I'm happy to document that we allow mail templates in subdirectories, so that it's an officially supported thing rather than just something we happen to support.

brianhelba:
I've removed the 2nd commit in response to @zachmullen 's feedback. Feel free to add some documentation if you'd like, but it would also be nice to have more comprehensive packaging tests that checked email template functionality, since that's a major thing that could go wrong during packaging.

zachmullen:
> it would also be nice to have more comprehensive packaging tests that checked email template functionality, since that's a major thing that could go wrong during packaging.

:+1: for that

brianhelba:
@zachmullen Once the tests come back green, are you satisfied with this PR?

zachmullen:
Yep, LGTM.

PR #1037, opened 2015-10-24
Add additional checks to packaging tests

brianhelba:
These will explicitly detect and report when a Girder server failed to start, or when a the server did not return a response.

Also, ensure that packaging tests will timeout when a server doesn't respond. This makes each HTTP request timeout after 5 seconds. The test still won't fail if the very first request takes significantly longer (as it's retried in a loop).

### DIFF
--- a/tests/packaging/test_rest_api.sh
+++ b/tests/packaging/test_rest_api.sh
@@ -24,7 +24,7 @@ echo "Detected api version ${version}"
 
 timeout=0
 until [ $timeout -eq 5 ]; do
-    json=$("${CURL}" --max-time 5 --silent http://localhost:${GIRDER_PORT}/api/v1/system/version)
+    json=$("${CURL}" --connect-timeout 5 --max-time 5 --silent http://localhost:${GIRDER_PORT}/api/v1/system/version)
     if [ -n "$json" ] && [[ $json == *shortSHA* ]]; then
         break
     fi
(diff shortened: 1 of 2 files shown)


brianhelba:
PTAL @jbeezley @manthey

jbeezley:
LGTM

tests/packaging/test_rest_api.sh
```diff
@@ -7,31 +7,42 @@ unset PYTHONPATH
...
+
 version="$(girder-install version)"
 echo "Detected api version ${version}"
 
-TIMEOUT=0
-until [ $TIMEOUT -eq 15 ]; do
-    json=$("${CURL}" -s http://localhost:${GIRDER_PORT}/api/v1/system/version)
+timeout=0
+until [ $timeout -eq 5 ]; do
+    json=$("${CURL}" --max-time 5 --silent http://localhost:${GIRDER_PORT}/api/v1/system/version)
```
manthey: With curl, I'd recommend also specifying `--connect-timeout 5` (besides the `--max-time` parameter).  In completely unrelated use of curl, I've had it stuck in the connection phase even with the max-time specified.  As long as we are adding the timeouts, we should have both of them.

brianhelba: Will do, thanks.

PR #1040, opened 2015-10-27
WIP: Fix 404 errors caused by Swagger throbber

brianhelba:
This completely resolves issue #446.

### DIFF
--- a/clients/web/test/spec/swaggerSpec.js
+++ b/clients/web/test/spec/swaggerSpec.js
@@ -25,7 +25,7 @@ function jasmineTests() {
                 return $('#system_getVersion:visible').length > 0;
             }, 'end points to be visible');
             runs(function () {
-                $('#system_getVersion h3 a').click();
+                $('#system_getVersion h3 a:first').click();
             });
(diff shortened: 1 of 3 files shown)


zachmullen:
What's the status of this PR?

brianhelba:
This seems to operate just fine during normal use, but the test is failing due to some changes (likely bugs) with the upgraded version of Swagger. I'll give it another look, but it's not worth trying too hard over. I think we'd be best served putting efforts on this into upgrading to Swagger UI 2.1, which has substantial internal and Swagger specification changes.

zachmullen:
Should we close this?

brianhelba:
It works absolutely fine in operation, but the tests fail. From reading upstream bug reports, it looks like the solution is to upgrade to Swagger 2.1.

Further, while the docs claim that Swagger 2.1 is backwards comparable with the old 1.x Swagger spec (which Girder currently uses), I could not get Swagger 2.1 to work correctly with it. So, we'd also need to update the server to send down the spec in the new 2.0 format.

I give up. Let's close.

Issue #1041, opened 2015-10-27
Get testing coverage for our drag-and-drop code


zachmullen:
* UploadWidget
* MarkdownWidget

manthey:
Added in PR #1043 and #1044.

PR #1043, opened 2015-10-28
Add tests for drag-and-drop uploading.

manthey:
Adjust test code to match our js style guides.

This addresses part of issue #1041.  This only tests the main uploads, not the markdown uploads.

### DIFF
--- a/clients/web/test/testUtils.js
+++ b/clients/web/test/testUtils.js
@@ -988,7 +988,7 @@ girderTest.testUpload = function (uploadItem, needResume, error) {
  * callbacks, however.
  *
  * @param itemSize: the size of the item to drop.
- * @param multiple: if a number, try to srop this many items.  Only one item is
+ * @param multiple: if a number, try to drop this many items.  Only one item is
  *                  ever actually uploaded, because phantomjs doesn't support
  *                  multiple items at one time, but this simulates dropping
  *                  multiples.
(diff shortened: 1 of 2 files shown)


clients/web/test/testUtils.js
```diff
@@ -970,13 +976,129 @@ girderTest.testUpload = function (uploadItem, needResume, error) {
...
+             suffix: girderTest._uploadSuffix});
+    });
+};
+
+/* Test upload drop events.  The drag and drop events are artificially
+ * constructed, so this isn't a perfect test.  It will exercise the event
+ * callbacks, however.
+ *
+ * @param itemSize: the size of the item to drop.
+ * @param multiple: if a number, try to srop this many items.  Only one item is
```
zachmullen: drop*

zachmullen:
While we're at it, can we extend or refactor this utility to make it accept the drop target element as a parameter? That'd make it easier to test things like the markdown widget.

manthey:
My initial hope was to reuse this for the markdown widget, but the two widgets are fairly different.  The Upload widget shows a different DOM element when you are hovering over it, versus adding a style to the drop zone, and there are a bunch of other differences.  I was going to extend it to the markdown widget in a separate PR, at which point I'll try to reuse as much of this code as possible.

zachmullen:
Ok, that sounds good. LGTM once travis is green

PR #1045, opened 2015-10-28
Batch install python deps to detect version collisions

jbeezley:
The `InstallPythonRequirements.py` script originally installed each requirement file one by one.  When two plugins share dependencies or conflict with girder's core dependencies, pip would overwrite the installation each time with the last requirements file processed.  This became apparent in other projects because [romanesco](https://github.com/Kitware/romanesco/blob/master/requirements.txt#L7) currently depends on `pymongo==2.7` which overwrites girder's `3.0` installation.

This PR saves up all requirements files and installs them all at once.  This has the advantage of throwing an error when there is a conflict like this.  Unfortunately, it will also throw an error if two requirements files have the same version of the same package as well.  My opinion is that it is better to raise the error here than try and figure out what went wrong later even if the checks are too strict.

We could parse the requirements files with the less than stable pip python API (`pip.req.parse_requirements`, ...) and resolve conflicts manually, but that seems like more effort than it is worth.

### DIFF
--- a/scripts/InstallPythonRequirements.py
+++ b/scripts/InstallPythonRequirements.py
@@ -24,23 +24,32 @@ import os
 import subprocess
 import sys
 
+files = []
 
-def installReqs(file, pip):
+
+def installReqs(file):
     if not os.path.exists(file):
         return
 
-    print('\033[32m*** Installing: %s\033[0m' % file)
-    if subprocess.call([pip, 'install', '-U', '-r', file]) != 0:
-        print('\033[1;91m*** Error installing %s, stopping.\033[0m' % file,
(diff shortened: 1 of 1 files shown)


zachmullen:
LGTM

jbeezley:
Before this goes in, we should plan how to fix minerva + any other projects using the script.  I'm not yet convinced this change is worth the trouble it will cause.

zachmullen:
Hm, that is problematic that you can't even have the exact same version dependency in two different files. Maybe we should close this...?

jbeezley:
I may have an alternate solution depending on how motivated I get during my next flight.

PR #1046, opened 2015-10-28
Fix static image path for json editor

danlamanna:
@zachmullen I remember us agreeing that the path I originally used was correct - but if you install girder fresh, the icons background image for the editor 404s. For some reason it seemed to be working for people using an already existing girder environment.

This should fix that.


zachmullen:
LGTM

zachmullen:
@danlamanna this change actually breaks the icons for me; it's now trying to fetch them from http://localhost:8080/static/built/static/built/jsoneditor/img/jsoneditor-icons.png.

I think I know why this is behaving so wonky. The following line in the vega plugin is importing this stylesheet, and the vega version is imported second and therefore overriding the core style rule and giving an incorrect relative path:

    @import '../../../../clients/web/src/stylesheets/widgets/metadataWidget.styl'

I'm not sure the best way to resolve that, but at least now we know why it's breaking conditionally.

zachmullen:
(All those icons could be replaced with our font icons if we are willing to override every class, that way we could avoid image refs entirely and use less bandwidth...)

danlamanna:
Thanks for tracking that down... I was just encountering the conditional behavior as well.

What do you think about using [this](https://github.com/gruntjs/grunt-contrib-stylus#define) in our gruntfile to have the static root defined in our stylus files?

zachmullen:
That seems like a great idea.

danlamanna:
Alright, I'll push a commit to this branch soon(ish).

danlamanna:
So this is broken in 2 respects:
- The path is incorrect due to plugin extending the stylesheet
- It hardcodes the static path as ```static``` despite the static path being configurable

After talking to @jbeezley we don't see how it will be possible to support defining the static root in stylus without re-introducing the cfg parsing into the grunt process. I know that task complicated things, however I don't see another way around this, given we agree that stylus should be able to refer to static assets.

Thoughts?

zachmullen:
Let's punt on this. We can arrange our stylesheets in such a way that the vega plugin doesn't need to import the image URL rule. I'm generally not wild about having to import sheets into other sheets, it's a lot of duplication, though I can see it is necessary in some cases.

jbeezley:
Crazy idea: Inject a [CSS variable](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables) into the main template in same way we do for the [html](https://github.com/girder/girder/blob/master/girder/utility/webroot.mako#L20).

danlamanna:
LGTM

PR #1048, opened 2015-10-29
Remove some redundant code in the folder creation HTTP API

brianhelba:
The "createFolder" function in the model API already grants admin access to the folder creator. There's no need for the HTTP API to do it too.

The duplicate functionality in the HTTP API also may override plugin customizations that expect only one call to "setUserAccess" within the model layer.


zachmullen:
LGTM once travis is happy. Glad you spotted this -- it's very old code.

PR #1053, opened 2015-10-30
Add an OAuth login button to the user registration window

brianhelba:
Previously, a first-time user could click "Register", and never know that the option to use OAuth was available.

### DIFF
--- a/plugins/oauth/web_client/templates/oauth_login.jade
+++ b/plugins/oauth/web_client/templates/oauth_login.jade
@@ -1,6 +1,6 @@
 .g-oauth-container
   .g-oauth-section-header
-    .g-oauth-section-header-text Or log in with
+    .g-oauth-section-header-text Or #{modeText} with
 
   each button in buttons
     .g-oauth-button(class="#{button.class}", g-provider="#{button.provider}")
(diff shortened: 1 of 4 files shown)


zachmullen:
This isn't showing up for me in the register dialog.

plugins/oauth/web_client/templates/oauth_login.jade
```diff
@@ -1,6 +1,6 @@
 .g-oauth-container
   .g-oauth-section-header
-    .g-oauth-section-header-text Or log in with
+    .g-oauth-section-header-text Or #{modeText} with
```
zachmullen: This variable is never passed into the template.

Also, it's fine for it to say "log in" even in the register case, most sites seem to phrase it that way.

zachmullen:
@brianhelba while QAing this I noticed the styling of the "Or log in with" text is done in a bad way, so if you don't mind I'm planning to commit a fix for that onto this topic.

zachmullen:
PTAL. I left the "register automatically" for now, but my vote would be to change it to just say "log in".

brianhelba:
I see the issue, I didn't originally notice that setup was instantiating a view, not a template.

I'd rather have the "register automatically". Indeed, most sites don't make a distinction, but I've encountered first hand a number of less-technical users that get confused by this process. OAuth login is still new to many people.

Please wait on any merge while I review @zachmullen 's latest change.

zachmullen:
Ok, in that case this change LGTM, just pending your approval on my commit.

brianhelba:
LGTM.

PR #1055, opened 2015-10-30
Accommodate missing access refs

zachmullen:
If ACL's contained references to users or groups that no longer existed, we were throwing KeyErrors. Now, we just quietly clean these references out of the document and continue on.


brianhelba:
Is this to benefit custom ``AccessControlledModel``s? The standard types with ACLs should already be cleaned up by https://github.com/girder/girder/blob/master/girder/models/user.py#L134 .

zachmullen:
Yep.

brianhelba:
I wonder if we could use something like ``AccessControlledModel.__subclasses__()`` to find all of the known plugin-defined ``AccessControlledModel`` types, whenever we call ``User.remove`` and ``Group.remove`` (right now it is hardcoded to ``Collection``, ``Folder``, ``User``, ``Group``).

This change might still be necessary as a final fallback, in case a user was deleted while a plugin was temporarily disabled, but I'd ideally like to clean up all invalid ACLs as soon as possible in the ``remove`` method.

brianhelba:
Per my discussion with @zachmullen, I've pushed an additional commit to generalize the removal of references to deleted users and groups for all ``AccessControlledModel`` subtypes.

There should be one final commit on this PR, to move the constant ``"core.cleanupDeletedEntity"`` into the ``constants.CoreEventHandler`` class, but that class is added in the currently-unmerged https://github.com/girder/girder/pull/1025 .

brianhelba:
@zachmullen The more that I think about it, we should be creating a new AccessControlledModel subclass, instantiating, sharing resources with, then deleting it. That was broken behavior before, so none of the tests could have been validating it.

Where do you think is the best place for this sort of test?

zachmullen:
Good idea, I'm not sure any of the existing test cases jump out at me as good locations to put such a test. Maybe we can add a new test case called `model_test.py`?

brianhelba:
@zachmullen ``CoreEventHandler.ACCESS_CONTROL_CLEANUP`` is the best name I could think of. If you have a better one, feel free to change it.

I'll add the ``model_test.py`` within the next day, then this PR should be ripe.

zachmullen:
@brianhelba What's the status on this PR?

zachmullen:
LGTM, thanks for adding the additional tests.

PR #1057, opened 2015-11-01
Allow MetadataMixin callbacks to be optional

danlamanna:
There are cases where I don't need to do anything if the call succeeded, and it would be cleaner not to have to pass a noop function.


zachmullen:
LGTM.

PR #1058, opened 2015-11-01
Monkey patch yet another bug in boto

### LINKED ISSUE #1056: Cannot create S3 Assetstore with anonymous
I just discovered that we can't create an S3 assetstore when connecting with anonymous (no credentials), due to the recent s3 sigv4 work, though really this looks like a bug in boto.

Current released version of boto is 2.38.

In the s3_assetstore_adapter we [`boto.config.set('s3', 'use-sigv4', 'True')`](https://github.com/girder/girder/blob/cc34c0727c62ffd0f3b2a71bac1300b6ff383927/girder/utility/s3_assetstore_adapter.py#L35).

Here's my stacktrace

>  File "girder/utility/s3_assetstore_adapter.py", line 659, in botoConnectS3
    return boto.connect_s3(**connectParams)
  File "/usr/local/lib/python2.7/dist-packages/boto/__init__.py", line 141, in connect_s3
    return S3Connection(aws_access_key_id, aws_secret_access_key, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/boto/s3/connection.py", line 190, in __init__
    validate_certs=validate_certs, profile_name=profile_name)
  File "/usr/local/lib/python2.7/dist-packages/boto/connection.py", line 569, in __init__
    host, config, self.provider, self._required_auth_capability())
  File "/usr/local/lib/python2.7/dist-packages/boto/auth.py", line 987, in get_auth_handler
    'Check your credentials' % (len(names), str(names)))
NoAuthHandlerFound: No handler was ready to authenticate. 1 handlers were checked. ['S3HmacAuthV4Handler'] Check your credentials

The problem seems to be when `S3Connection.__init__` calls the superclass `Connection.__init__`, which calls the `S3Connection.`[`_required_auth_capability`](https://github.com/boto/boto/blob/d9e5cfe900e1a58717e393c76a6e3580305f217a/boto/s3/connection.py#L200-L205) function , reproduced here

```
@detect_potential_s3sigv4
def _required_auth_capability(self):
    if self.anon:
        return ['anon']
    else:
        return ['s3']
```

We are correctly passing `anon=True` to `connect_s3` in these cases, which *should* be covered in `_required_auth_capability`, except for the [`detect_potential_s3sigv4`](https://github.com/boto/boto/blob/d9e5cfe900e1a58717e393c76a6e3580305f217a/boto/auth.py#L1002-L1020) wrapper which overrides this and returns `['hmac-v4']` instead.

This PR https://github.com/boto/boto/pull/3014 has a fix for this issue in 693b1de, but it hasn't moved in some time.

@manthey @zachmullen 

You guys are most familiar with the s3 assetstore adapter, is there a workaround?  

When I comment out `boto.config.set('s3', 'use-sigv4', 'True')` and my anonymous assetstore creation works.  

Could we maybe set this boto config value dynamically in [`botoConnectS3`](https://github.com/girder/girder/blob/cc34c0727c62ffd0f3b2a71bac1300b6ff383927/girder/utility/s3_assetstore_adapter.py#L639-L653) depending on the value `connectParams['anon']`?  Or something less :scream:?  I'd settle for :scream_cat: .

zachmullen:
Fixes #1056.

@mgrauer This is our nastiest monkey patch yet for boto. I'm getting kind of concerned about boto in general -- the project seems to be dead or dying, nobody's bugfix PRs are being merged anymore it seems. Sadly I think we're stuck with it, but patching internals like this makes me nervous for sure.

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -49,10 +49,23 @@ def authv4_determine_region_name(self, *args, **kwargs):
     return result
 
 
+def required_auth_capability_wrapper(fun):
+    def wrapper(self, *args, **kwargs):
+        if self.anon:
+            return ['anon']
+        else:
+            return fun(self, *args, **kwargs)
+    return wrapper
+
+
 authv4_orig_determine_region_name = \
     boto.auth.S3HmacAuthV4Handler.determine_region_name
 boto.auth.S3HmacAuthV4Handler.determine_region_name = \
     authv4_determine_region_name
+boto.s3.connection.S3Connection._required_auth_capability = \
+    required_auth_capability_wrapper(
+        boto.s3.connection.S3Connection._required_auth_capability)
+
 
 
 def _generate_url_sigv4(self, expires_in, method, bucket='', key='',
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -67,7 +67,6 @@ boto.s3.connection.S3Connection._required_auth_capability = \
         boto.s3.connection.S3Connection._required_auth_capability)
 
 
-
 def _generate_url_sigv4(self, expires_in, method, bucket='', key='',
                         headers=None, response_headers=None, version_id=None,
                         iso_date=None, params=None):


danlamanna:
Regarding boto - it seems like they are only accepting critical PRs while all of the development happens on boto3 (the 2 can be used alongside each other). [source](https://github.com/boto/boto#boto-3)

zachmullen:
Thanks for the info. I'll open an issue to explore porting to boto3.

mgrauer:
@zachmullen 

Thanks for the fix, LGTM.

Issue #1054, opened 2015-10-30
Swagger path is wrong on proxies


manthey:
PR #954 was a change to eliminate an extra `./` in the path of the swagger api calls.  Unfortunately, this was done by using the absolute path, which breaks easy use of the api when Girder is hosted through a proxy.

I'm not sure how to fix both problems with the current version of Swagger without monkey-patching.  Alternatively, we can use the current code when _not_ using a proxy and the old form of using an api base path of `.` when using a proxy.

manthey:
@zachmullen Since you made PR #954, I was wondering if you had a suggestion for this.

zachmullen:
Sorry, not sure how to get this working in all cases.

manthey:
I see four possible ways to address this:

* Revert to using `.` as the basePath.  This results in the problem original reported in #948, though this is essentially a cosmetic issue.

* Detect if we are in a proxy via tools.proxy.on, and, if so, use `.` and if not use `getApiUrl()`.  This means we'll still have the ugly extra `./` when behind a proxy, and if the user setting up girder fails to add the proxy setting, we'll still have the current issue.

* Have a config setting that can override this value.  Recommend setting it to `.` if you are in a proxy and need to access the swagger interface both inside and outside of the proxy, or to `(proxy url)/api/v1` if you only need to access it from outside of the proxy.

* Monkey patch swagger to strip out the offending `./`.  We would have to make sure this doesn't break any time we changed swagger-ui versions.

I think the third method is probably the best choice of these.

manthey:
Resolved in PR #1793.

PR #1061, opened 2015-11-02
[WIP] Determine API root behind a proxy

manthey:
When behind a proxy, swagger doesn't work because of how we get the api url.  This now has a setting to override it.  If set to ".", the url will look ugly (as it has an extra . in the path, but will work both inside and outside of the proxy.  If set to "https?://(proxy host and port)/(proxy path)/api/v1", it will look nice and work through the proxy, but not locally.

A better technique would be to modify swagger-ui, but that seems heavy-handed for now.

I need to add tests for this and update the documentation, but wanted some feedback first.

This addresses issue #1054.

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -59,18 +59,29 @@ def getUrlParts(url=None):
 def getApiUrl(url=None):
     """
     In a request thread, call this to get the path to the root of the REST API.
-    The returned path does *not* end in a forward slash.
+    The returned path does *not* end in a forward slash.  If no url is
+    specified, and there is a Referer header that includes the REST API root,
+    use that for determining the root.  This allows detection of the root from
+    within and outside of a proxy.
 
     :param url: URL from which to extract the base URL. If not specified, uses
-        `cherrypy.url()`
+        `cherrypy.url()` or the config value from ['server']['api_url'].
     """
+    apiPhrase = '/api/v1'
     url = url or cherrypy.url()
-    idx = url.find('/api/v1')
+    if url is None and 'Referer' in cherrypy.request.headers:
+        referer = cherrypy.request.headers['Referer']
+        if referer.endswith(apiPhrase) or (apiPhrase + '/') in referer:
+            url = referer
+    if url.endswith(apiPhrase):
+        idx = url.rfind(apiPhrase)
+    else:
+        idx = url.find(apiPhrase + '/')
 
     if idx < 0:
         raise GirderException('Could not determine API root in %s.' % url)
 
-    return url[:idx + 7]
+    return url[:idx + len(apiPhrase)]
 
 
 def _cacheAuthUser(fun):
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -65,14 +65,15 @@ def getApiUrl(url=None):
     within and outside of a proxy.
 
     :param url: URL from which to extract the base URL. If not specified, uses
-        `cherrypy.url()` or the config value from ['server']['api_url'].
+        `cherrypy.url()`.
     """
     apiPhrase = '/api/v1'
-    url = url or cherrypy.url()
     if url is None and 'Referer' in cherrypy.request.headers:
         referer = cherrypy.request.headers['Referer']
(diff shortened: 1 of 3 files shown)


manthey:
@zachmullen This addresses issue #1054.  If you would rather have a different setting name or think I should use a different solution, let me know.  If you like this, I'll update the documentation and add tests.

zachmullen:
It seems we should add this logic into `getApiUrl`, or is there some reason not to?

Also, I thought the `tools.proxy.base` config setting would take care of this for us. Does it not?

manthey:
`tools.proxy.base` doesn't fix the issue (I just checked).  Swagger doesn't reference things properly within the proxy (and it probably won't work both inside and outside of the proxy).

I can move the code to `getApiUrl`.

manthey:
@zachmullen, Using `tools.proxy.base` *and* `tools.proxy.local` does work around the problem, but another technique would be to use the `Referer` header to determine the api root. (see current changes).  The virtue of this change is that (a) it reduces the amount of configuration that is needed, and (b) it works both inside and outside of the proxy.

girder/api/rest.py
```diff
@@ -59,18 +59,29 @@ def getUrlParts(url=None):
...
     In a request thread, call this to get the path to the root of the REST API.
-    The returned path does *not* end in a forward slash.
+    The returned path does *not* end in a forward slash.  If no url is
+    specified, and there is a Referer header that includes the REST API root,
+    use that for determining the root.  This allows detection of the root from
+    within and outside of a proxy.
 
     :param url: URL from which to extract the base URL. If not specified, uses
-        `cherrypy.url()`
+        `cherrypy.url()` or the config value from ['server']['api_url'].
```
zachmullen: This doesn't seem to actually use the config value from ['server']['api_url'], am I missing something?

girder/api/rest.py
```diff
@@ -59,18 +59,29 @@ def getUrlParts(url=None):
...
     """
+    apiPhrase = '/api/v1'
     url = url or cherrypy.url()
-    idx = url.find('/api/v1')
+    if url is None and 'Referer' in cherrypy.request.headers:
```
zachmullen: Will this block ever be entered? Wouldn't `cherrypy.url()` have to return `None` for that?

girder/api/rest.py
manthey: My bad.  I removed the setting but didn't revert the text.

girder/api/rest.py
manthey: Apparently I'm just paying attention this morning...

zachmullen:
I'm not sure we should use the Referer header, for a couple of reasons:

1. In the normal case, the user agent sets the Referer header to the URL of the page that *referred* the user agent to the requested URL. So if someone makes a CORS request to our API, this would actually be set to the URL of the third-party site the user was looking at when the request was made.
2. In the malicious user case, the user can set the Referer header to whatever value they want, meaning someone could fool a romanesco task into uploading result data to any URL they wanted, for instance.

zachmullen:
To be fair, it seems to me that it would also be possible for a malicious user spoof the hostname in an HTTP request, meaning cherrypy.url() could also be wrong. I'm wondering if we should just have a config value or setting for this rather than trying to derive this info from request parameters, which can almost all be spoofed.

manthey:
If we care about header spoofing, then there is no safe way to determine the url both inside and outside of a proxy.  Inside of a proxy, the cherrypy.url() will be the internal url.  I think the only safe method that works on both sides of the proxy is using `.` as we used to do.

zachmullen:
True, though that still leaves us with the problem of 1). The referer header will almost never contain what we want it to, and there's a small chance it could contain something that will mess things up, i.e. if a third-party site makes a request to girder from a page that happens to contain "/api/v1" for whatever reason.

I'm thinking we may just want to improve our documentation for the cherrypy proxy config values, or add our own absolute URL config value to get around this issue completely.

zachmullen:
@manthey can we close this or do we have plans for it?

PR #1067, opened 2015-11-04
Make OAuth plugin able to support multiple providers

zachmullen:
This is a heavy refactoring of the oauth plugin that makes it
much easier to add new providers. As a demonstration of this, a
new provider (GitHub) has been added.


brianhelba:
@zachmullen Let me know when this is ready. I can quickly port over support for Facebook and Linkedin [from SlideAtlas](https://github.com/SlideAtlas/SlideAtlas-Server/tree/master/slideatlas/security/login_provider).

zachmullen:
This is now ready for review.

brianhelba:
I've reviewed this. This is overall a great improvement, but I'm hoping to have several hundred people using OAuth in the near term, so I want to make sure that we get it as good as possible.

I see a few opportunities for additional refinement:
* We should separately store each of the provider IDs, so a user can have multiple OAuth providers simultaneously associated with their Girder account. At the moment, each provider overwrites the ``user['oauth']`` object with their own ID.
* We should first lookup users by provider ID, since a user may change their primary email address with third-party providers. If that lookup fails, we should fall back to email.
* We should do more rigorous validation of the OAuth responses from providers, both during code-token exchange and API calls. I've had past experiences with the Google Plus API unexpectedly failing to return some fields (like an email address), probably due to either user's privacy settings or some sort of legacy/unverified Google accounts.

I've gone ahead and implemented most of my proposed changes, along with several smaller changes. Before I share these, I'm hoping @zachmullen can clarify a few minor questions. I'll post those inline, and once there's a response, I'll update my patch accordingly and publish it.

plugins/oauth/server/providers/base.py
```diff
@@ -0,0 +1,173 @@
...
+
+        :param email: The email address.
+        :type email: str
+        """
+        for login in self.generateLogins(email, firstName, lastName, userName):
+            login = login.lower()
+            if self.testLogin(login):
+                return login
+
+        return None
```
brianhelba: An exception should be raised at this point. If we absolutely can't create a login name, we shouldn't return the invalid login name ``None``, since that will just fail later in the more-difficult-to-trace ``User.validate`` method.

plugins/oauth/server/providers/google.py
```diff
@@ -0,0 +1,95 @@
...
+            'redirect_uri': self.redirectUri
+        }
+        resp = self.getJson(method='POST', url=constants.GOOGLE_TOKEN_URL,
+                            data=params)
+
+        headers = {
+            'Authorization': ' '.join((
+                resp['token_type'], resp['access_token']))
+        }
+        resp = self.getJson(method='GET', url=constants.GOOGLE_USER_URL,
```
brianhelba: Do you have any interest in passing ``fields`` into the query string of this Google Plus API request? We can specifically request only the fields of interest (id, name, email), rather than getting the entire "person" response. See https://developers.google.com/+/web/api/rest/#partial-response?

plugins/oauth/server/constants.py
```diff
@@ -20,9 +20,18 @@
 
 # Constants representing the setting keys for this plugin
 class PluginSettings:
+    PROVIDERS_ENABLED = 'oauth.providers_enabled'
+    GITHUB_CLIENT_ID = 'oauth.github_client_id'
+    GITHUB_CLIENT_SECRET = 'oauth.github_client_secret'
     GOOGLE_CLIENT_ID = 'oauth.google_client_id'
     GOOGLE_CLIENT_SECRET = 'oauth.google_client_secret'
 
+GITHUB_AUTH_URL = 'https://github.com/login/oauth/authorize'
```
brianhelba: Do you think we should move these provider-specific constants to the individual provider files? Once we have a half dozen different providers, the current layout could get a little messy, considering they're not properly namespaced.

plugins/oauth/server/rest.py
```diff
@@ -35,78 +33,9 @@ def __init__(self):
...
-    googleCallback.description = None
-
-    def _setCsrfToken(self, redirect):
-        """
-        Generate an anti-CSRF token and set it in a cookie for the user. It will
-        be verified upon completion of the OAuth login flow.
-        This can be used consistently across all of the OAuth providers.
-        """
+    def _sendCsrfToken(self, redirect):
         csrfToken = self.model('token').createToken(days=0.25)
```
brianhelba: The purpose of the [``state`` query string parameter](https://tools.ietf.org/html/rfc6749#section-4.2.1) is primarily to provide CSRF protection. By declining to validate ``state`` on the 2nd half of the login call and using the cookie instead, we are creating a very minor theoretical vulnerability, but more importantly, we're deviating from the OAuth2 standard.

At a minimum, this should be implemented where the cookie contains the redirect location and ``state`` contains the token.

However, I think it'd be even cleaner to get rid of the temporary cookie entirely, and changing ``state`` to be a concatenated combination of a token + a redirect location. Do you have any objections?

plugins/oauth/web_client/templates/oauth_config.jade
```diff
@@ -1,34 +1,36 @@
 .g-config-breadcrumb-container
 
 .g-oauth-provider-list-title Configure supported providers below
+p Only fill in the information for the OAuth providers you wish to enable.
```
brianhelba: We can call the plugin ``oauth``, but we should refer to the protocol as ``OAuth2`` in all user-facing configuration strings.

zachmullen: :+1:

plugins/oauth/server/rest.py
zachmullen: I'm fine with state containing both fields, perhaps the token, then a space, then the redirect location.

plugins/oauth/server/constants.py
zachmullen: I'm okay with either way.

plugins/oauth/server/providers/google.py
zachmullen: Sounds reasonable to me.

plugins/oauth/server/providers/base.py
zachmullen: I'm fine with that.

zachmullen:
@brianhelba Thanks for the QA... which parts do you want me to take care of and which did you take care of?

brianhelba:
@zachmullen I've made most of the changes already, I just need to do the ``state`` token thing and test. I'll do it.

brianhelba:
@zachmullen Should we bump the version in the ``plugin.json`` file? I don't know what the policy on plugin versioning is.

zachmullen:
Let's bump it to 1.1.0.

plugins/oauth/server/providers/base.py
```diff
@@ -0,0 +1,173 @@
...
+        Returns a user via this OAuth provider given an OAuth code. Subclasses
+        must implement this to retrieve an email address, first name, last name,
+        and optionally a login name from the OAuth provider, and then call
+        self.createOrReuseUser with that information to lookup or create a
+        corresponding Girder user.
+        """
+        raise NotImplementedError()
+
+    def createOrReuseUser(self, email, firstName, lastName, userName=None):
+        cursor = self.model('user').find({'email': email}, limit=1)
```
brianhelba: @zachmullen One last question, why are we doing this ``cursor.count`` stuff instead of ``findOne``?

zachmullen: Legacy. Feel free to fix.

plugins/oauth/server/providers/base.py
```diff
@@ -0,0 +1,237 @@
...
+            return json.loads(content)
+        except ValueError:
+            raise RestException('Non-JSON response: %s' % content, code=502)
+
+    def _createOrReuseUser(self, oauthId, email, firstName, lastName,
+                           userName=None):
+        providerName = self.getProviderName()
+
+        # Try finding by ID first, since a user can change their email address
+        user = self.model('user').findOne({
```
brianhelba: We should add an index to the 'oauth' field. To do this, we should store ``{'provider':..., 'id':....}``, so a compound index can be created.

Thanks @zachmullen .

plugins/oauth/server/providers/base.py
```diff
@@ -0,0 +1,237 @@
...
+            'oauthId.%s' % providerName: oauthId})
+        setId = not user
+
+        # Handle a legacy way of storing Google IDs
```
brianhelba: Expand on this comment.

brianhelba:
@zachmullen I'd consider this ready to merge. PTAL.

plugins/oauth/web_client/stylesheets/login.styl
```diff
@@ -29,15 +31,33 @@
     border-right 1px solid white
   .g-oauth-button-text
     display inline-block
-    padding 0 20px
+    text-align center
+    width 100px
```
brianhelba: I selected the ``width 160px`` for the whole button with the largest string of ``StackExchange`` in mind. Once that provider is added, we may need to expand this ``width 100px`` for the text as well. If it looks good, it's fine for now though.

zachmullen:
I added some very minor style changes onto the topic, everything else LGTM except for the one minor issue related to the `partition` call.

plugins/oauth/web_client/stylesheets/login.styl
zachmullen: The max amount we have to expand into is 109px (110 causes it to flow down below the icon). This is what it looks like as 100:

screen shot 2015-11-12 at 11 20 56 am

If you'd rather go to the full 109px right now, we can do that; if we want to go any further we'll need to bump up the button width over 160px.

brianhelba:
See my inline comment about the ``partition`` call.

zachmullen:
Fair enough, LGTM!

plugins/oauth/web_client/stylesheets/login.styl
brianhelba: I say leave it alone for now. We can always tweak things once the larger names actually exist.

brianhelba:
@zachmullen You can do the merge if you're ready. LGTM too.

plugins/oauth/web_client/stylesheets/login.styl
zachmullen: Actually just tested this out to make sure... it looks better as 109, so I'm going to change it to that. 

screen shot 2015-11-12 at 11 24 44 am

PR #1069, opened 2015-11-05
Make imported files trigger events with their extra attributes present

brianhelba:
This will make extra attributes ("path", "mtime", "imported") on imported files be present at the time of the "model.file.save.created" event, so event handlers can more intelligently handle imported files.

This will also remove an unnecessary extra call to the database.


zachmullen:
LGTM once travis is happy, seems like we should have been doing this anyway.

brianhelba:
I'm going to create another PR to handle a potential bug in how [``propagateSizeChange`` is called](https://github.com/girder/girder/blob/master/girder/models/file.py#L230) even when ``saveFile==False``. I'll move the ``propagateSizeChange`` to an event handler.

PR #1070, opened 2015-11-05
Docker update

Xarthisius:
This PR should fix build failure on docker hub, which was caused by `pip install -U six`. Additionally, node pulled from `ppa:chris-lea/node.js` was outdated, which resulted in multiple warnings during `npm install` of girder. Now, both node and npm can be regulated via changing ENV vars in Dockerbuild.

As a bonus I've added simple fig file allowing for a quick local deployment:
```
docker-compose build
docker-compose up -d
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,37 @@ EXPOSE 8080
 RUN mkdir /girder
 RUN mkdir /girder/logs
 
+RUN apt-get update && apt-get install -qy software-properties-common python-software-properties && \
+  apt-get update && apt-get install -qy \
+    build-essential \
+    curl \
+    git \
+    libffi-dev \
+    libpython-dev \
+    python-pip && \
+  apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Get node
+RUN set -ex \
+   && for key in \
+      7937DFD2AB06298B2293C3187D33FF9D0246406D \
+      114F43EE0176B71C7BC219DD50A3051F888C628D \
+   ; do \
+      gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key"; \
+   done
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -6,6 +6,37 @@ EXPOSE 8080
...
+ENV NPM_VERSION next
+
+RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-x64.tar.gz" \
+   && curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/SHASUMS256.txt.asc" \
+   && gpg --verify SHASUMS256.txt.asc \
+   && grep " node-v$NODE_VERSION-linux-x64.tar.gz\$" SHASUMS256.txt.asc | sha256sum -c - \
+   && tar -xzf "node-v$NODE_VERSION-linux-x64.tar.gz" -C /usr/local --strip-components=1 \
+   && rm "node-v$NODE_VERSION-linux-x64.tar.gz" SHASUMS256.txt.asc \
+   && npm install -g npm@"$NPM_VERSION" \
+   && npm cache clear
```
zachmullen: The complexity of the above command worries me a little, could we just do:

`curl -sL https://deb.nodesource.com/setup | bash -`

?

Xarthisius: That's straight from `node:0.10.40`, maybe it would be even easier to make `FROM node` rather than ubuntu and get rid of those bits altogether

zachmullen: I'd be fine with that :+1:

zachmullen:
I'm going to go ahead and merge this, our docker hub builds have been failing recently anyway, hopefully this'll fix it. Thanks @Xarthisius

PR #1071, opened 2015-11-05
Prevent item size from being updated if a file is created but not saved

brianhelba:
See comments in the code for rationale on when this may happen.

### DIFF
--- a/girder/models/file.py
+++ b/girder/models/file.py
@@ -253,7 +253,7 @@ class File(acl_mixin.AccessControlMixin, Model):
 
         fileDoc = event.info
         itemId = fileDoc.get('itemId')
-        if itemId:
+        if itemId and fileDoc.get('size'):
             item = self.model('item').load(itemId, force=True)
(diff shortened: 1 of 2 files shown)


girder/models/file.py
```diff
@@ -222,19 +228,35 @@ def createFile(self, creator, item, name, size, assetstore, mimeType=None,
...
+        """
+        # This task is not performed in "createFile", in case
+        # "saveFile==False". The item size should be updated only when it's
+        # certain that the file will actually be saved. It is also possible for
+        # "model.file.save" to set "defaultPrevented", which would prevent the
+        # item from being saved initially.
+
+        fileDoc = event.info
+        itemId = fileDoc.get('itemId')
+        if itemId:
```
zachmullen: `if itemId and fileDoc.get('size', 0):`

Link files don't have a size field, and we shouldn't call the method for 0-length files anyway.

zachmullen:
@brianhelba I pushed a small fix, the rest LGTM.

PR #1073, opened 2015-11-07
Make sure static/built/plugins dir always exists

zachmullen:
Some plugins with custom Gruntfiles (including COVALIC) expected
this directory to exist at the time their Gruntfile is first executed,
but prior to actual task execution. If users had installed such a
plugin without running grunt beforehand, they would encounter errors.
The easiest fix is just to ensure that the directory is present.


mgrauer:
LGTM.

PR #1074, opened 2015-11-09
Initialize the girder namespace in init.js

jbeezley:
Currently, the girder object is initialized by the jade template.  This makes the initialization explicit locally and protects against possible problems related to the order the javascript files are loaded.


zachmullen:
LGTM

PR #1075, opened 2015-11-09
Uglify was broken with the recent refactor, this fixes it

zachmullen:
@jbeezley could you take a look at this, I just noticed on a fresh girder install that none of the plugin javascripts were being loaded, which was because they had the wrong name and were being overwritten by the stylus step. It wasn't affecting environments where the js had already been built.

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -90,7 +90,7 @@ module.exports = function (grunt) {
                         '<%= pluginDir %>/<%= grunt.task.current.args[0] %>/web_client/js/**/*.js',
                         '<%= grunt.config.getRaw("jade.plugin.files")[0].dest %>'
                     ],
-                    dest: '<%= staticDir %>/built/plugins/<%= grunt.task.current.args[0] %>/plugin.min.css'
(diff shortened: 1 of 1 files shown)


jbeezley:
LGTM.  I'm somewhat concerned about why this didn't cause an error on travis because it is always built from scratch.  Do we not have any tests that use javascript from the plugins?

zachmullen:
We do, but we import them directly as source files rather than the minified versions so that we can perform coverage on the source files :)

jbeezley:
Okay, I see.

cpatrick:
Shouldn't we test on the minified versions as well...

zachmullen:
Yeah, we should. Maybe we can conditionally load the JS files one way or the other based on whether javascript coverage is turned on in our CMake cache. That way, travis could test the minified version (which is closer to a real deployment) and Aiur could test the source-dump mode and perform coverage.

zachmullen:
Actually I just realized Travis does in fact do the JS coverage since one of our tests is actually to measure that and make sure it's above a threshold. It just doesn't report it to CDash due to having a too-old version of ctest. I guess we could do it in our build matrix somehow...

jbeezley:
Travis reports it.  I fixed that a while back.

PR #1076, opened 2015-11-10
Always use npm installed jscs and jshint

jbeezley:
Without `NO_DEFAULT_PATH`, cmake will pick up globally installed* versions of jscs and jshint, which likely the wrong version.  This forces the local version.

<verb>*</verb> I use nvm, get off my lawn.


zachmullen:
LGTM

PR #1078, opened 2015-11-10
Allow simple specification of API URL for GirderClient

zachmullen:
Prior to this change, users had to pass the host, port, scheme,
etc. individually. This change enables users to simply pass
a single argument containing the full URL if they so desire,
which is much simpler in many cases.


mgrauer:
Can you add an example of using this to the python-client.rst ?

Nothing else other than these two comments.

zachmullen:
@mgrauer documents updated, PTAL

mgrauer:
LG, so fresh and so clean.

PR #1079, opened 2015-11-10
v1.1.1 python client bump

zachmullen:
I just added a new feature into the python client that romanesco needs, so I'm bumping the version and pushing to pypi. ping @mgrauer


mgrauer:
LGTM.

Issue #899, opened 2015-07-22
enable girder-install command line client to install 3rd party plugins


mgrauer:
allow girder-install to install 3rd party plugins, i.e. those not part of girder core.

This should allow for 

  * installing plugin's python server side code
  * pip installing plugin's requirements.txt specified dependencies

possibly doing these

  * installing plugin's npm dependencies
  * installing plugin's bower dependencies (?)
  * building and installing the plugin's web client code

but probably Grunt wouldn't be available in the case where a user is running girder-install, so instead girder-install should be able to install the web client code that the plugin should already have available in a built and packaged form.

mgrauer:
Add to our grunt package step, so that a 3rd party plugin will get its web client packaged up correctly so that it could be added to a github release for that plugin, and then this will be installable by girder-install.

jbeezley:
'girder-install` does (or should) install server side code and requirements for an arbitrary plugin packaged as a tarball.  One of my original intents with `girder-install` was to support installing from the plugin's git repository, but you are right about node possibly missing.

It would be awesome to have a searchable plugin repository somewhere containing packaged plugins.  The way homebrew does it's package installation from github is the best mechanism like this that I have seen.  Perhaps something hosted on data.kitware.com is another alternative.

zachmullen:
Yeah, the more I think about it the more I like the idea of requiring plugins to do all their own building/packaging so that we can simply download a tarball and plop it down. If we can provide a nice package registry (as a girder plugin!?!) that would be awesome.

mgrauer:
@jbeezley 

I think using girder-install you can install server side code for a 3rd party plugin, but there is a current limitation for installing the web client of the plugin.  I forget the exact details now.

jbeezley:
Yeah, I the limitation is that it won't install the client side content.  I'm not sure the conventions for that were nailed down at the time that I wrote the installer, or at least I wasn't aware of it.

jbeezley:
@zachmullen A girder plugin that hosts girder plugins... this sounds like a corollary to Russell's paradox.

zachmullen:
I'll try not to tear any holes in the universe when we implement this

brianhelba:
+1 for supporting Bower, either as a common tool that's installed available with Girder core, or with a clean procedure for a plugin requiring and using it.

+1 for allowing a plugin to specify its own NPM requirements, which are recursively installed (although if we wanted to only install them for enabled plugins, there are some issues with a circular workflow).

[...]

PR #1080, opened 2015-11-11
Filter out node_modules from packaged plugins

jbeezley:
The new `npm install` step that occurs in plugin directories can generate a large `node_modules` directory that is only needed at build time.  These are now filtered out in the packaging step.

### DIFF
--- a/grunt_tasks/package.js
+++ b/grunt_tasks/package.js
@@ -87,7 +87,8 @@ module.exports = function (grunt) {
                 filter: function (fname) {
                     return !fname.match(/\/plugin_tests/) &&
                            !fname.match(/cmake$/) &&
-                           !fname.match(/py[co]$/);
+                           !fname.match(/py[co]$/) &&
+                           !fname.match(/\/node_modules/);
                 }
             }
         }


cpatrick:
:clap:

cpatrick:
LGTM, assuming that no plugins depend on this. Should something be added to the docs?

zachmullen:
@jbeezley and others, just FYI I have a local branch working toward packaging girder's web client code and publishing it via npm, if you have any more planned work outstanding related to npm stuff, let's coordinate.

jbeezley:
@cpatrick The packaged tarball is mostly useless at the moment because it will only handle server side plugins (in which case the node_modules directory is truly pointless).  There is a related issue[#899] to add client side support the `girder-install`.  Until then, I don't expect anybody will be relying on the plugin release packages.

@zachmullen I have one more small change that is about done.  It fixes the `grunt watch` command and improves task that creates `testEnv.html`.  See [update-grunt-watch](https://github.com/girder/girder/compare/update-grunt-watch).

zachmullen:
Thanks @jbeezley , sounds good

PR #1082, opened 2015-11-12
Rename test assetstores to all be prefixed with "girder_test_"

brianhelba:


manthey:
Is there a reason for this change other than neatness?  If you are running tests locally and trying to make it easy to clean up, it would be straightforward to remove the test assetstores when the tests complete.

brianhelba:
Yes, its about neatness.

If we also want to start removing the test databases after tests complete, I'd support that change, but I wasn't sure if leaving them around was a "feature" to allow developers to inspect them manually.

manthey:
If we remove them when done testing. we can always add a flag that gets passed through cmake if we want to keep them (which we haven't).

brianhelba:
PTAL @girder/developers

zachmullen:
LGTM

PR #1085, opened 2015-11-13
Update to latest version of grunt watch

jbeezley:
This fixes an issue causing `grunt watch` to fail when the configured task matches no files.


zachmullen:
LGTM

PR #1087, opened 2015-11-16
Bump cffi version to 1.3.0

kotfic:
Minerva now depends on ```cryptography==1.1.0``` which requires ```cffi>=1.1.0```


jbeezley:
LGTM

PR #1088, opened 2015-11-16
Editable girder install plugins

kotfic:
Adds an ```--editable``` or ```-e``` flag to ```girder-install plugin``` that will symlink rather than copy the directory tree into girder/plugins/

@zachmullen PTAL  @jbeezley thanks!

### DIFF
--- a/girder/utility/install.py
+++ b/girder/utility/install.py
@@ -113,10 +113,10 @@ def install_plugin(opts):
 
         if (os.path.isdir(targetPath) and
                 os.path.samefile(pluginPath, targetPath) and not
-                opts.editable ^ os.path.islink(targetPath)):
+                opts.symlink ^ os.path.islink(targetPath)):
             # If source and dest are the same, we are done for this plugin.
             # Note: ^ is a logical xor - not xor means only continue if
-            # editable and islink() are either both false, or both true
+            # symlink and islink() are either both false, or both true
             continue
 
         if os.path.exists(targetPath):
@@ -129,7 +129,7 @@ def install_plugin(opts):
             else:
                 raise Exception('Plugin already exists at %s, use "-f" to '
                                 'overwrite the existing directory.')
-        if opts.editable:
+        if opts.symlink:
             os.symlink(pluginPath, targetPath)
         else:
             shutil.copytree(pluginPath, targetPath)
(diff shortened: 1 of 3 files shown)


girder/utility/install.py
```diff
@@ -112,20 +112,27 @@ def install_plugin(opts):
...
             continue
 
         if os.path.exists(targetPath):
             if opts.force:
                 print(constants.TerminalColor.warning(
                     'Removing existing plugin at %s.' % targetPath))
-                shutil.rmtree(targetPath)
+
+                shutil.rmtree(targetPath,
+                              onerror=lambda *args: os.unlink(targetPath))
```
zachmullen: I think it's considered bad form to use a lambda for side effects.

girder/utility/install.py
```diff
@@ -148,6 +155,9 @@ def main():
     plugin.set_defaults(func=install_plugin)
     plugin.add_argument('-f', '--force', action='store_true',
                         help='Overwrite plugins if they already exist.')
+    plugin.add_argument('-e', '--editable', action='store_true',
+                        help='Install by symlinking to the plugin directory.')
```
zachmullen: I appreciate that this is consistent with the pip install flag, but to me it seems much clearer to just use `-s, --symlink` as the name of this option. Do others feel strongly about this?

brianhelba: I'm in favor of sticking closer to ``pip install``'s conventions, but I see nothing wrong with also adding a ``--symlink`` flag as a documented alias.

girder/utility/install.py
kotfic: too much scala recently :stuck_out_tongue:

girder/utility/install.py
kotfic: symlink is good for me,  will push once i've added docs

kotfic:
Changes the ```-e``` flag to ```-s``` or ```--symlink```

zachmullen:
Travis is not obeying my instruction to restart that failing build... this LGTM, but I'd like to wait at least a little bit to see if travis gets it together.

brianhelba:
@kotfic , please be sure to delete branches once the corresponding PR is merged.

I'll leave this one for you to take care of when you see this.

PR #1089, opened 2015-11-17
Fixes #1068. Deprecate the plugin_directory setting

### LINKED ISSUE #1068: Grunt is not finding plugins from plugin_directory
Girder [server supports plugins](https://github.com/girder/girder/blob/d42b3817b16f69a3992c05c750c840956d3df31f/girder/utility/plugin_utilities.py#L195) from the location specified by the configuration option ``plugins.plugin_directory``.

However, Grunt only [looks for and builds plugins](https://github.com/girder/girder/blob/d42b3817b16f69a3992c05c750c840956d3df31f/Gruntfile.js#L451) in the ``plugins/*`` directory, and does not also search this configured path. The result is that while ``plugin_directory`` is quite useful for server-side plugins, is useless for client-side plugins.

The Gruntfile should be updated to remedy this.

zachmullen:


danlamanna:
LGTM.

PR #1092, opened 2015-11-17
Girder install dev options

kotfic:
@zachmullen PTAL thanks!

### DIFF
--- a/girder/utility/install.py
+++ b/girder/utility/install.py
@@ -58,12 +58,15 @@ def fix_path(path):
     return os.path.abspath(os.path.expanduser(path))
 
 
-def runNpmInstall(wd=None):
+def runNpmInstall(wd=None, dev=False):
     """
     Use this to run `npm install` inside the package.
     """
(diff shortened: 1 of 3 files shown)


girder/utility/install.py
```diff
@@ -77,7 +80,10 @@ def install_web(opts=None):
     Build and install Girder's web client. This runs `npm install` to execute
     the entire build and install process.
     """
-    runNpmInstall()
+    try:
+        runNpmInstall(dev=opts.development, npm=opts.npm)
+    except AttributeError:
+        runNpmInstall()
```
kotfic: ```install_web(opts=None)```  - if opts is None i want to catch and run runNpmInstall() with defaults

zachmullen: Ah, in that case let's just use a conditional

```
if opts is None:
    runNpmInstall()
else:
    runNpmInstall(....)

zachmullen:
LGTM

PR #1094, opened 2015-11-18
Include a human-readable name in the default email sender setting

brianhelba:
This provides a cue to administrators that they can use a human-readable display name in this field, per RFC2822-3.4.


zachmullen:
LGTM

PR #1097, opened 2015-11-18
Upgrade dependencies

brianhelba:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ girder-*.tar.gz
 __pycache__
 ### PyCharm ###
 .idea/
+.PyCharm50/
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
     # travis environment is bad
     - vmstat
     - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "3" ]; then export PY3="true"; else export PY2="true"; fi
-    - if [ -n "${PY3}" ]; then export MONGO_VERSION=3.0.2; export PY_COVG="OFF"; else export MONGO_VERSION=2.6.9; export PY_COVG="ON"; export DEPLOY=true; fi
+    - if [ -n "${PY3}" ]; then export MONGO_VERSION=3.0.7; export PY_COVG="OFF"; else export MONGO_VERSION=2.6.11; export PY_COVG="ON"; export DEPLOY=true; fi
     - if [ -n "${PY3}" ]; then export IGNORE_PLUGINS=hdfs_assetstore,metadata_extractor; fi
     - pushd "${HOME}"
     - curl "https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${MONGO_VERSION}.tgz" | gunzip -c | tar x
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -18,8 +18,8 @@ provide helpful development tools and to allow the test suite to run: ::
 
     pip install -r requirements-dev.txt
 
-.. note:: One of the development requirements, `httpretty`, can fail to install
-   if under certain system locales (see the `error report
+.. note:: One of the development indirect requirements, `httpretty`, can fail to
+   install if under certain system locales (see the `error report
    <https://github.com/gabrielfalcao/HTTPretty/issues/108>`_).  Changing to any
    UTF8 locale works around this problem.  For instance, on Ubuntu, you can
    change your system locale using the commands: ::
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -227,7 +227,6 @@ You will find many useful methods for client side testing in the ``girderTest`` 
 defined at ``/clients/web/test/testUtils.js``.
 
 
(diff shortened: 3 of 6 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.6)
+cmake_minimum_required(VERSION 2.8.12)
```
cpatrick: Do we need any CMake features that require this update? If not, I don't think we should update this.

brianhelba: After I get the rest of this WIP branch stabilized, I was planning on changing [Travis to use](https://github.com/girder/girder/blob/master/.travis.yml#L26) the same version as whatever we decide the minimum should be (it's using 3.1 now). That way, if a commit relies upon a newer CMake feature, the test will fail.

Note that [Ubuntu 12.04 uses CMake 2.8.7](http://packages.ubuntu.com/precise/cmake). I doubt we'll ever see anything older than that in the wild, and it's not unreasonable to ask for a newer just to run tests (it's not needed for deployment). The current requirement of CMake 2.8.6 seems kind of arbitrary (unless there's a good, undocumented reason).

zachmullen: We need to use CMake v3.1+ in testing for our coverage submission to work.

Since it's not feasible to make that our minimum requirement since it's not in any of the package managers, we should use the oldest possible CMake version that will successfully configure and build girder, which I believe is 2.8.6, but we can build 2.8.6 and see if that's true.

zachmullen:
On the pip requirements: before we merge this, we need to also put in parallel updgrade branches for both romanesco and minerva, and then simultaneously merge after all three are approved, a la Captain Phillips.

Going forward, I think we should try and do this on a regular schedule, perhaps in conjunction with a minor release, maybe no more often than 2-4 times a year unless there's a compelling reason to upgrade in between.

jbeezley:
There are a couple of options that I have seen around for this kind of thing.  For python packages there is [requires.io](https://requires.io/).  To see how it works with girder, take a look [here](https://requires.io/github/jbeezley/girder/requirements/?branch=master).  Note: they have a rest api and python client that you need an expensive enterprise account to use, but there are badges for free:
[Requirements Status](https://requires.io/github/jbeezley/girder/requirements/?branch=master)

A more basic option is [piprot](https://github.com/sesh/piprot).  Another tool validates the current environment is satisfied by a requirements file: [piplint](https://github.com/dcramer/piplint).

For node, I have used [outdated](https://docs.npmjs.com/cli/outdated) to get a list of outdated deps, but it looks like [npm-check](https://www.npmjs.com/package/npm-check) is the new hotness.  This site watches all github repos and does badges: [https://david-dm.org/](https://david-dm.org/).
[Dependency Status](https://david-dm.org/girder/girder)

brianhelba:
> Going forward, I think we should try and do this on a regular schedule, perhaps in conjunction with a minor release, maybe no more often than 2-4 times a year unless there's a compelling reason to upgrade in between.
@zachmullen You read my mind.

mgrauer:
Should we add a note about this to the [release process documentation](https://github.com/girder/girder/blob/a0300ce8c6043af9ab0f0def55420d469f2ff3a0/docs/development.rst#creating-a-new-release)?

brianhelba:
@mgrauer I'm less enthusiastic about making arbitrary dependency upgrades a requirement to do specific releases. Instead, once the upgrades happen, we should want to do a release anyway.

I've seen other projects that have a scheduled quarterly dependency review, and I'd suggest we target the same thing for Girder. We just need to pick some recurring date (how about the first day of each season, a.k.a the solstices and equinoxes?) and document that. If 4x ends up being too much churn, we can slow down to 2x.

I'll add something to the documentation to this once we agree.

brianhelba:
[requires.io](https://requires.io/) looks like a nice tool and while I have no objection to new badges, I'm concerned that it works against us having a dependable upgrade schedule. As soon as that badge turns red, we're going to want to upgrade something to fix it, which leads to excessive churn.

brianhelba:
On the client side, however, upgrading is less disruptive (since NPM does it automatically with every build), and security vulnerabilities are typically going to have a higher impact (with the exception of something in CherryPy).

So, I'd be more supportive of adding a [https://david-dm.org/](https://david-dm.org/) badge for NPM deps. What do @girder/developers think?

jbeezley:
I did try to update all npm deps recently, but hit a test error and promptly deleted the branch.

https://travis-ci.org/girder/girder/jobs/90115781

brianhelba:
PTAL @girder/developers 

This PR now upgrades all Python requirements that are unique to Girder, and not used in Romanesco or Minerva.

Please particularly note:
* I was not able to upgrade Moto to the latest release (0.4.18), since releases from 0.4.1 onwards
cause test failures in "web_client_data_s3".
* I added a basic policy outline for a quarterly dependency review to the documentation. Comments are encouraged.

zachmullen:
LGTM

PR #1100, opened 2015-11-18
Allow setuptools install to be called twice without errors

zachmullen:
This was breaking the RTD build also.


jbeezley:
LGTM :shipit:

PR #1102, opened 2015-11-18
Add a "name" field to the oauth provider list API

brianhelba:
Having a human-readable display name for the oauth provider is useful information for most clients to have.

This only adds an extra field to the objects returned by the HTTP API, so it requires only a minor version bump.


zachmullen:
LGTM

PR #1103, opened 2015-11-18
Fix API breakage of the OAuth provider list endpoint

zachmullen:
@brianhelba PTAL


brianhelba:
LGTM.

brianhelba:
@zachmullen Note, we theoretically could change the plugin version back to 1.x at this point.

zachmullen:
I thought of that, but I'm going to leave it at its current version since no harm can come of that, and since it feels a bit wrong to make a version not monotonically increase.

brianhelba:
Fair enough; I'm not necessarily advocating the idea. We did only bump it to 2.0 an hour ago.

PR #1107, opened 2015-11-19
Add a cookieAuth decorator to make endpoints accept cookie authentication

brianhelba:
This is cleaner and more Pythonic than the old ``endpoint.cookieAuth = True`` way. The old way will continue to work, but is now undocumented and should be considered deprecated.

### DIFF
--- a/girder/api/access.py
+++ b/girder/api/access.py
@@ -77,3 +77,38 @@ def token(fun):
         return fun(*args, **kwargs)
     accessDecorator.accessLevel = 'token'
     return accessDecorator
+
+
+def cookie(*args, **kwargs):
+    """
+    REST endpoints that allow the use of a cookie for authentication should be
+    wrapped in this decorator.
+
+    When used as a normal decorator, this is only effective on endpoints for
+    HEAD and GET routes (as these routes should be read-only in a RESTful API).
+
+    While allowing cookie authentication on other types of routes exposes an
+    application to Cross-Site Request Forgery (CSRF) attacks, an optional
+    ``force=True`` argument may be passed to the decorator to make it effective
+    on any type of route.
+    """
+    if len(args) == 1 and callable(args[0]):  # Used as a raw decorator
+        force = False
+        args[0].cookieAuth = (True, force)
+        return args[0]
+
+    else:  # Used with arguments
+        if len(args) == 1 and isinstance(args[0], bool):
+            force = args[0]
+        elif 'force' in kwargs:
+            force = kwargs['force']
+        elif (not args) and (not kwargs):
+            force = False
+        else:
+            raise TypeError('Unsupported extra decorator arguments.')
+
+        def decorator(fun):
+            fun.cookieAuth = (True, force)
+            return fun
+
+        return decorator
--- a/girder/api/access.py
+++ b/girder/api/access.py
@@ -57,11 +57,8 @@ def public(fun):
     Functions that allow any client access, including those that haven't logged
     in should be wrapped in this decorator.
     """
-    @functools.wraps(fun)
-    def accessDecorator(*args, **kwargs):
-        return fun(*args, **kwargs)
-    accessDecorator.accessLevel = 'public'
-    return accessDecorator
+    fun.accessLevel = 'public'
+    return fun
 
 
 def token(fun):
--- a/docs/developer-cookbook.rst
+++ b/docs/developer-cookbook.rst
@@ -210,6 +210,8 @@ body would be ``0123456789``.
 
 .. code-block:: python
 
+    from girder.api import access
+
     @access.public
     def rawExample(self, params):
         def gen():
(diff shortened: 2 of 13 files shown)


zachmullen:
Two suggestions:

1. I vote we move this out of `access` and into `rest`, lest people mistake it for another access level.
2. This is unrelated, but in looking at this, I noticed `access.public` is also a no-op that just needs to set a property on the function. Could you change that decorator to just return the input function rather than wrapping it in something useless?

brianhelba:
1. ``rest.py`` is actually the first place that I started to put it, but after seeing what a cluttered hodgepodge it was, and noticing that ``access.py`` was entirely other decorators with access-related functionality, I put it there. IMO, the fact that the other things in ``access.py`` happen to set the ``accessLevel`` attribute is an implementation detail; the more significant thing from an organizational perspective is their common functionality (controlling the who and how of first-line endpoint access control).

 If you still aren't convinced @zachmullen , I'll move it into the clutter.

2. Will do.

zachmullen:
Anyone else care to weigh in on how to namespace this symbol?

jbeezley:
To me, the `cookieAuth` decorator seems closely related to `access.token` so it makes sense to put it in the same place.  As long as it is in access, I would actually rename it to just `cookie` for consistency.

zachmullen:
Sounds good, let's keep in in `access`. I also like the idea of renaming the decorator to just "cookie".

brianhelba:
Updated. PTAL.

zachmullen:
LGTM once travis is green.

girder/api/access.py
```diff
@@ -80,3 +77,15 @@ def accessDecorator(*args, **kwargs):
...
     return accessDecorator
+
+
+def cookie(fun):
+    """
+    REST endpoints that allow the use of a cookie for authentication should be
+    wrapped in this decorator.
+
+    This should only be set on endpoints that are read-only (GET requests),
+    to avoid vulnerabilities to Cross-Site Request Forgery attacks.
```
cpatrick: The security-sensitive part of me wants to restrict this to not work on anything but a GET request unless explicitly stated through a parameter to the decorator...

brianhelba: @cpatrick good idea: ``@cookie`` or ``@cookie(force=True)``, allowing only GET or HEAD by default.

It significantly complicates the internal implementation to have a decorator optionally take a parameter, but that's not a problem.

zachmullen: On a long-lingering local branch of mine, I implemented such decorators for the access levels. They look like this:

```
def admin(*args, **kwargs):
    """
    Functions that require administrator access should be wrapped in this
    decorator.

    :param scope: To also expose this endpoint for certain token scopes,
        pass those scopes here. If multiple are passed, all will be required.
    :type scope: str or list of str
    """
    if len(args) == 1 and callable(args[0]):  # Raw decorator
        @six.wraps(args[0])
        def wrapped(*iargs, **ikwargs):
            rest.requireAdmin(rest.getCurrentUser())
            return args[0](*iargs, **ikwargs)
        wrapped.accessLevel = 'admin'
        return wrapped
    else:  # We should return a decorator
        def dec(fun):
            @six.wraps(fun)
            def wrapped(*iargs, **ikwargs):
                token = rest.getCurrentToken()
                if not _tokenModel.hasScope(token, kwargs.get('scope')):
                    rest.requireAdmin(rest.getCurrentUser())
                return fun(*iargs, **ikwargs)
            wrapped.accessLevel = 'admin'
            return wrapped
        return dec
```

Hopefully that helps :)

jbeezley: Implementing decorators always makes python look like javascript.

brianhelba: At least we don't have any code that messes with descriptors or metaclasses yet (though we will if I ever get time to work on v2.0 of our model layer).

zachmullen:
I'm handing QA of this off to @cpatrick

girder/api/access.py
```diff
@@ -80,3 +77,38 @@ def accessDecorator(*args, **kwargs):
...
+        if len(args) == 1 and isinstance(args[0], bool):
+            force = args[0]
+        elif 'force' in kwargs:
+            force = kwargs['force']
+        elif (not args) and (not kwargs):
+            force = False
+        else:
+            raise TypeError('Unsupported extra decorator arguments.')
+
+        def decorator(fun):
```
brianhelba: I just realized this should include ``@wraps``. I'll update later.

cpatrick:
It doesn't look like we test this cookie-based authentication for anything other than items, but we allow it for files, folders, collections, and notifications. I'd like tests for that, but I could live with a technical debt issue being open against this as it could be a nice on-ramp task for an aspiring developer. Other than that (and the `@wraps` issue), LGTM.

zachmullen:
The major gap in testing coverage here is for the new `force` option... can we add a test for that using a fake route?

girder/api/access.py
brianhelba: I was mistaken, this doesn't need ``@wraps``.

brianhelba:
PTAL again @zachmullen @cpatrick . I've added 3 additional commits to:
* Make cookie authorization warning print when route is first defined
* Allow cookie authorization to be used for "POST resource/download"
* Add testing of cookie authorization decorators

cpatrick:
@brianhelba I had one question regarding the POST do resource/download, but the tests are great.

PR #1108, opened 2015-11-20
Allow description property to be set via decorator

zachmullen:
The description property of route handlers can now be set
via a decorator rather than just being set as a property of the
function. This has a couple of minor benefits:

1. Users won't have to type the name of the function twice in
   order to describe it.
2. Descriptions can go above the function, rather than below it,
   which is a more familiar place to see documentation of a
   function.

@brianhelba this is obviously inspired by your recent similar change
regarding cookieAuth, PTAL.

### DIFF
--- a/docs/plugin-development.rst
+++ b/docs/plugin-development.rst
@@ -125,23 +125,25 @@ will default to being restricted to administrators.
 
 When you start the server, you may notice a warning message appears:
 ``WARNING: No description docs present for route GET item/:id/cat``. You
-can add self-describing API documentation to your route as in the following
+can add self-describing API documentation to your route using the
+``describeRoute`` decorator and ``Description`` class as in the following
 example: ::
 
-    from girder.api.describe import Description
+    from girder.api.describe import Description, describeRoute
     from girder.api import access
 
     @access.public
+    @describeRoute(
+        Description('Retrieve the cat for a given item.')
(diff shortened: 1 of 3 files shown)


brianhelba:
I like the spirit of this change, but since we're adding new syntax, I wonder whether there's a cleaner way than having to do ``@describeRoute(Description(...``, and require 2 different imports to use it.

Additionally, do you see this as the new preferred way to add route docs, such that we should deprecate and remove documentation for the old way? If so, we should create a low-priority isse to move all of the existing docs to use this format. Otherwise, I'm not sure why you just changed ``Folder.find``.

zachmullen:
> Otherwise, I'm not sure why you just changed Folder.find.

Didn't want to bother changing every single one if people didn't think this was a good idea :)

brianhelba:
@zachmullen What do you think of the syntax / need for 2 different symbol imports? I haven't thought of a more elegant way yet, but it's worth doing properly now, since we can't change it later.

zachmullen:
We actually could change it later by adding more kwargs to the decorator besides "description", so I'm OK with this syntax... it's at least better than the old way.

brianhelba:
LGTM.

Technically I'd say this is solid, and while I don't really care for the syntax / usage, I can't think of a better way myself. Most importantly though, it's an improvement to the current style of raw attribute assignment.

PR #1109, opened 2015-11-20
Fix pip install logic in an upgrade scenario

zachmullen:
During an upgrade pip install, pip simply merges the new files
into the existing directory. Our old logic only performed the
recursive copy of our package data (plugins, web client code)
if the old dirs didn't exist, which they did in an upgrade
scenario. This uses the distutils copy_tree, which performs
a directory merge to achieve the desired effect.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,15 +24,18 @@ import sys
 
 from setuptools import setup, find_packages
 from setuptools.command.install import install
+from distutils.dir_util import copy_tree
 from pkg_resources import parse_requirements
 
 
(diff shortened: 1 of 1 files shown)


zachmullen:
@girder/developers I discovered this issue after cutting our current release and installing it over a previous girder release that had used the old implementation of the `girder-install web` command. Once this goes in, I plan on cutting 1.4.1 so we can install cleanly over older versions.

cpatrick:
lgtm.

PR #1112, opened 2015-11-20
Add widgets to input and modify date/times

msmolens:
This commit adds two widgets: DateTimeWidget and DateTimeRangeWidget.

DateTimeWidget provides a text input field to specify a date/time. The user
chooses the date/time using a popup picker.

DateTimeRangeWidget provides text input fields to specify a range of
dates/times.  That is, the first field specifies "from" and the second field
specifies "to." The user chooses each date/time using a popup picker.

The widgets are implemented using the "eonasdan-bootstrap-datetimepicker"
module.

This commit also adds the Bootstrap 3.1.1 glyphicons font because
eonasdan-bootstrap-datetimepicker requires the font for its picker.


zachmullen:
@msmolens I've just added you to the collaborators for girder, you can now push and merge branches directly.

clients/web/static/lib/bootstrap/fonts/glyphicons-halflings-regular.svg
```diff
@@ -0,0 +1,229 @@
+<?xml version="1.0" standalone="no"?>
```
zachmullen: I'm going to experiment to see how hard it is to override the CSS of the datepicker widget to use our existing fontello icons so we can get rid of these.

clients/web/src/views/widgets/DateTimeWidget.js
```diff
@@ -0,0 +1,228 @@
...
+ */
+girder.views.DateTimeWidget = girder.View.extend({
+
+    /**
+     * @param [settings.prefix='default'] Prefix for element IDs in case
+     *     multiple DateTimeWidgets are rendered simultaneously.
+     */
+    initialize: function (settings) {
+        this.prefix = settings.prefix || 'default';
+        this.dateTimeId = '#' + this.prefix + '-datetime';
```
zachmullen: We generally try to avoid using id's unless we have to. In this case I think we could get away with a fixed class instead, like `.g-datetime`, and use `this.$('.g-datetime')` to select it without having to worry about the global uniqueness problem.

Also, for both of these widgets, let's make the initial date value something that can be passed in the settings object to `initialize`.

clients/web/src/views/widgets/DateTimeWidget.js
```diff
@@ -0,0 +1,228 @@
...
+     *     multiple DateTimeRangeWidgets are rendered simultaneously.
+     */
+    initialize: function (settings) {
+        this.prefix = settings.prefix || 'default';
+        this.dateTimeFromId = '#' + this.prefix + '-datetime-from';
+        this.dateTimeToId = '#' + this.prefix + '-datetime-to';
+    },
+
+    render: function () {
+        var view = this;
```
zachmullen: This is a minor nit, but I've started trying to avoid these sorts of `var foo = this` workaround bindings and just using `_.bind` instead to set the value of `this` explicitly. Not sure if it's worth being prescriptive about, just feels a bit cleaner to me.

clients/web/src/views/widgets/DateTimeWidget.js
```diff
@@ -0,0 +1,228 @@
...
+        // after "to" date
+
+        var options = {
+            showClear: true,
+            showTodayButton: true,
+            useCurrent: 'day'
+        };
+        view.$(view.dateTimeFromId).datetimepicker(options);
+
+        options['useCurrent'] = false;
```
zachmullen: This is what's causing the test failures, our style checkers prescribe using dot notation for object keys whenever syntactically possible.

clients/web/static/lib/bootstrap/fonts/glyphicons-halflings-regular.svg
zachmullen: Experiment complete, we just need to pass the following value into the widget's options when creating it:


    icons: {
                time: 'icon-clock',
                date: 'icon-calendar',
                up: 'icon-up-open',
                down: 'icon-down-open',
                previous: 'icon-left-open',
                next: 'icon-right-open',
                clear: 'icon-trash',
                close: 'icon-cancel',
                today: 'icon-target'
            }

I'm really impressed with this library, it's got great customizability, has a really nice look and feel and interaction, MIT license, it's in npm and is well maintained... all in all, you made a great choice amongst the hundreds of libraries that do the same thing. Nice work.

zachmullen:
Since it's so easy to make the widget switch between opening on focus on the text input vs. opening on clicking of a little icon on the right, we should make that a setting that can be passed into `initialize` so that downstreams can easily choose whichever behavior suits them.

msmolens:
@zachmullen Thanks for the feedback and suggestions! PTAL at the changes.

grunt_tasks/build.js
```diff
@@ -80,6 +80,14 @@ module.exports = function (grunt) {
...
                     dest: 'clients/web/static/built/jsoneditor'
                 }]
+            },
+            'eonasdan-bootstrap-datetimepicker': {
+                files: [{
+                    expand: true,
+                    cwd: 'node_modules/eonasdan-bootstrap-datetimepicker/build',
+                    src: 'css/bootstrap-datetimepicker.min.css',
+                    dest: 'clients/web/static/built/eonasdan-bootstrap-datetimepicker'
+                }]
```
zachmullen: Sorry for not noticing this earlier... rather than copying this, we can actually pass `node_modules/eonasdan-bootstrap-datetimepicker/build/css/bootstrap-datetimepicker.css` into our stylus inputs for `app.min.css`. The nice thing about doing it that way is that is it prevents our webroot and downstream webroots (such as COVALIC's) from needing to serve an extra file in their mako templates.

zachmullen:
LGTM

jbeezley:
:+1: on this.  I was just looking for a date widget yesterday.

PR #1113, opened 2015-11-21
Add a constructor to the Resource class

brianhelba:
All built-in Resource subclasses are updated to call the superclass constructor. However, non-migrated subclasses will continue to operate as before.


brianhelba:
PTAL @girder/developers

zachmullen:
LGTM

PR #1115, opened 2015-11-23
Experimental status ui

zachmullen:


zachmullen:
Point of discussion: after adding this icon, it becomes very clear that a *lot* of our plugins are pre-1.0. As part of this topic, we should consider whether we want to move any of these out of alpha. I think we should move **jobs** to v1.0 since it's got a solid API and is in use in the wild by various projects. Any other candidates?

zachmullen:
@girder/developers PTAL

zachmullen:
bump

clients/web/src/stylesheets/body/plugins.styl
```diff
@@ -1,5 +1,7 @@
 @import 'nib'
 
+$warnColor = #f0ad4e
+
```
danlamanna: Should we use this for the ```.g-plugin-warning``` below as well? I'm fine whichever one you think looks better, but it probably makes sense to only have one warn color (per file at least :P).

jbeezley:
Obligatory screenshot

screen shot 2015-12-08 at 1 30 48 pm

clients/web/src/stylesheets/body/plugins.styl
zachmullen: I agree that we should have only one warning color, I actually think `.g-plugin-warning` is a misnomer, though... it's actually more of an error message. Would you be OK if I added a `.g-plugin-error` class to the existing element and moved these styles to it?

danlamanna: Sounds good.

zachmullen:
@danlamanna should be fixed now.

clients/web/src/templates/body/plugins.jade
```diff
@@ -9,19 +9,25 @@ div.g-plugin-restart
...
+          span.g-experimental-notice(
+            title="WARNING: this plugin is an alpha version and is subject " +
+                  "to breaking changes; use at your own risk.") Experimental
+        if plugin.value.configRoute
+          a.g-plugin-config-link(g-route=plugin.value.configRoute, title="Configure Plugin")
             i.icon-cog
       .g-plugin-description #{plugin.value.description}
         if (!plugin.value.meetsDependencies)
-          .g-plugin-warning This plugin has unmet dependencies.
+          .g-plugin-warning.g-plugin-error-message This plugin has unmet dependencies.
```
danlamanna: Is there a reason you kept the plugin-warning class? I believe jasmine looks for it, but shouldn't we just remove it?

zachmullen: To some degree, we consider element class names to be part of our API surface, for the sake of selectors, not for style rules. Our web client plugins pattern of extension often involves wrapping a render method and then injecting elements into a page, and as such, these selectors become de facto hooks for such injection.

This isn't to say we'll *never* remove DOM elements or classes, just that we won't do it lightly.

danlamanna:
LGTM

Issue #1114, opened 2015-11-23
Add officially supported ways of pulling in extra js and css files for plugins


zachmullen:
Right now you have to hack it in some way or another, which is subject to breaking. The first thought that came to mind would be something in plugin.yml's grunt object:

```
grunt:
    prependJsFiles:
        - bower_components/foo/foo.js
        - node_modules/bar/bar.js
    appendCssFiles:
        - bower_components/foo/foo.css
```

There would be both prepend and append options for JS and CSS files to be built into plugin.min.js and plugin.min.css, and the paths would be relative to the plugin root directory.

A second option would be to expose some functions that could be called inside plugin Gruntfiles at config time.

ping @jbeezley

jbeezley:
I'll look into this along with #1104.

zachmullen:
thanks!

brianhelba:
@girder/developers I'm pretty sure this has gone past the point of being fixed to now be: "there are too many ways to do the same thing". I'm closing.

PR #1117, opened 2015-11-24
Simplify the plugin related build tasks

### LINKED ISSUE #1104: Fix grunt watch for plugins
There are a couple of problems with the grunt watch task particularly applied to plugin development.

1. The plugin specific tasks aren't initialized until the `plugin` task is run (see [here](https://github.com/girder/girder/blob/master/grunt_tasks/plugin.js#L212)).  For the moment, we can work around this by running `grunt plugin watch` instead, but that block of code should be moved to the previous loop where the plugins are detected.

2.  It appears as if some of the globs in the subtask definitions are picking up wrong and/or extra paths.  See https://github.com/Kitware/minerva/issues/208.

jbeezley:
This removes the grunt templates that were previously used to generate "generic" configurations for plugin tasks.  There are several benefits for this change:

  * you can now run task categories like `grunt uglify`
  * plugins can more easily change their own configuration
  * less black magic

Fixes #1104.  I tested it with minerva and there were no issues.  `grunt watch` works for files under `web_client`, and I verified that `touch plugins/minerva/web_client/templates/minerva_config.jade` run both the jade and uglify tasks.  I'm really trying to not break plugins this time, I promise.

For #1114.  With this change, there is no magic there so it should be possible to just append to the sources to the grunt config object if we want to make that part of the API surface.  Otherwise, this should make implementing the suggestion in #1114 straightforward.

For testing, the packaging tests do a descent job of covering this code except for the `watch` tasks.  I'm up for suggestions if and how that should be done.

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -24,98 +24,140 @@ module.exports = function (grunt) {
     var path = require('path');
 
     /**
-     * Register a task that will npm install inside plugin directories.
+     * Adds configuration for a plugin related multitasks:
+     *
+     *   * shell:plugin-<plugin name>
+     *      Runs npm install in the plugin directory
+     *   * jade:plugin-<plugin name>
+     *      Compiles jade templates into the plugin's template.js
+     *   * stylus:plugin-<plugin name>
+     *      Compiles stylus files into the plugin's main css file
+     *   * ugilfy:plugin-<plugin name>
+     *      Compiles all javascript sources into the plugin's main js file
+     *   * copy:plugin-<plugin name>
+     *      Copies an other files that are served statically
+     *
+     *  This will also add watch tasks for each of the above.
      */
-    grunt.config.merge({
-        shell: {
-            'plugin-install': {
-                /**
-                 * Run npm install in the plugin directory
-                 */
-                command: function (plugin) {
-                    var pth;
-                    if (!plugin) {
-                        return 'true';
-                    }
+    var configure_plugin = function (plugin) {
+        var pluginTarget = 'plugin-' + plugin,
+            pluginPath = path.resolve(grunt.config.get('pluginDir'), plugin),
+            staticPath = path.resolve(
+                grunt.config.get('staticDir'), 'built', 'plugins', plugin
+            ),
+            packageJson = path.resolve(pluginPath, 'package.json'),
+            cfg = {
+                shell: {},
+                jade: {},
+                stylus: {},
+                uglify: {},
+                copy: {},
+                watch: {},
+                default: {},
+                plugin: {},
+                'plugin-install': {}
+            };
 
-                    pth = path.resolve(grunt.config.get('pluginDir'), plugin);
-                    if (!fs.existsSync(path.resolve(pth, 'package.json'))) {
-                        // do nothing if no plugin was provided
-                        // or if it has no package.json file
-                        grunt.verbose.writeln('Skipping npm install');
-                        return 'true';
-                    }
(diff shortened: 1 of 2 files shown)


grunt_tasks/plugin.js
```diff
@@ -24,98 +24,140 @@ module.exports = function (grunt) {
...
-                    execOptions: {
-                        cwd: path.resolve(process.cwd(), 'plugins')
-                    }
-                },
-                src: ['<%= grunt.config.get("pluginDir") %>/<%= grunt.task.current.args[0] %>/package.json']
-            }
-        },
-        'plugin-install': {}
-    });
+    var configure_plugin = function (plugin) {
```
zachmullen: Let's use camel case rather than underscore

jbeezley: Done

zachmullen:
@jbeezley this fixed the issue with `grunt watch` for plugin template changes, but it's still not working for core template changes (maybe we want to put that off to another issue).

Do we now have some "official" (non hacky) way of allowing plugins to add their own JS and CSS files into their built files?

jbeezley:
I was targeting plugin issues here, but if you want me to address the core watch task as well, I don't think it should be too hard.

As for #1114, I was postponing a true fix for a later PR as we should discuss what we need in terms flexibility for the plugin build steps.  I don't necessarily consider monkey patching the `config.uglify.myPlugin.files` array as overly hacky.  This PR at the very least makes that possible again.

zachmullen:
>  I don't necessarily consider monkey patching the config.uglify.myPlugin.files array as overly hacky.

The syntax for doing it was pretty ugly since you have to explicitly read the value, change it, then write it back, and obviously I was not considering those things part of the API surface, just a quick and dirty hack that worked. I'm glad this at least restores that capability, but I'd like to go ahead and implement #1114 to make this easier on plugin devs with npm or bower deps. If you don't have time, let me know and I will go ahead and implement it.

This branch LGTM. @girder/developers do we think all of our grunt config names and behaviors should be part of the API surface? We should adopt a policy on this sooner rather than later.

jbeezley:
For the record, I wasn't advocating monkey patching, but it would expose the most power to the plugins.  I'm more in favor of a declarative approach.  For example, a list of files in `plugin.json`:

```javascript
{
  uglify: [...],
  jade: [...],
  stylus: [...]
}
```

I probably won't be able to implement any of this in the next week or so.  If you want it, feel free.

zachmullen:
I like that, maybe something like:

```
grunt:
    uglify:
        prepend: [file1, file2]
        append: [file1, file2]
        override: [<full list of files, will override default values>]
```

zachmullen:
And I might change "uglify" and "stylus" to "js" and "css" so that the declaration is technology-independent as well.

jbeezley:
This may be unrelated, but should there be a mechanism for plugins to bundle code for the `test-env-html` task? Is there a better way to do what [Minerva does](https://github.com/Kitware/minerva/blob/master/Gruntfile.js#L252)?

zachmullen:
I don't think a better way exists at this time, but it would be nice to add a utility function that could be used to aggregate the necessary sources.

jbeezley:
@zachmullen I threw in a fix for the `watch:js_core` task as well.

zachmullen:
Thanks, LGTM

mgrauer:
@jbeezley This was a nice improvement.

@zachmullen 

> do we think all of our grunt config names and behaviors should be part of the API surface?

on behalf of Minerva, yes.

PR #1122, opened 2015-12-01
Update file metadata in S3 if it changes

zachmullen:
This PR adds a new optional method to assetstore adapters to hook into when file information such as name and MIME type is changed. The only adapter that currently needs to take advantage of this is the S3 adapter.

### DIFF
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -325,7 +325,7 @@ class File(Resource):
     def updateFile(self, file, params):
         file['name'] = params.get('name', file['name']).strip()
         file['mimeType'] = params.get('mimeType',
-                                      file.get('mimeType', '')).strip()
+                                      (file.get('mimeType') or '').strip())
         fileModel = self.model('file')
         return fileModel.filter(
             fileModel.updateFile(file), self.getCurrentUser())
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -21,7 +21,7 @@ import cherrypy
 import errno
 import six
 
-from ..describe import Description
+from ..describe import Description, describeRoute
 from ..rest import Resource, RestException, loadmodel
 from ...constants import AccessType
 from girder.models.model_base import AccessException, GirderException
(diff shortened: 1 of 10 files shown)


zachmullen:
This is now ready for review.

mgrauer:
I'll take a look later in the week.

girder/utility/s3_assetstore_adapter.py
```diff
@@ -375,6 +375,7 @@ def finalizeUpload(self, upload, file):
                     'url': url,
                     'headers': headers
                 }
+                file['additionalFinalizeKeys'] = ('s3FinalizeRequest',)
```
mgrauer: what is the purpose of this?  was something broken before without it?

zachmullen: Our API prior to this change was not filtering file models that were getting returned, which was bad. I fixed that on this branch so that file models returned via REST only contain the keys they're supposed to. However, in the S3 case the client needs to know this additional information, so I added this behavior of including additional keys for the finalize step. It's rather specialized, so if you can think of a more elegant way I'm all ears.

mgrauer: makes sense.  I was more concerned about adding a test if this exposed a bug, but this fixed the problem (not filtering result) and preserved behavior (returning the S3FinalizeRequest).

girder/api/v1/file.py
```diff
@@ -261,20 +279,30 @@ def downloadWithName(self, id, name, params):
...
+    def deleteFile(self, file, params):
+        self.model('file').remove(file)
 
     @access.user
     @loadmodel(model='upload')
+    @describeRoute(
+        Description('Cancel a partially completed upload.')
+        .param('id', 'The ID of the upload.', paramType='path')
+        .errorResponse('ID was invalid.')
+        .errorResponse('You lack permission to cancel this upload.', 403)
```
mgrauer: sounds like a personal failing

:snowflake:

mgrauer:
LGTM.

mgrauer:
@zachmullen 

Before this PR, the item/getFiles endpoint included a `sha512` key with each item, because that endpoint  wasn't being filtered properly.  We counted on this key for a test in Minerva, which is why I'm back here now digging up the cause of the failing test.  We can work around this in Minerva, so I'm not worried about that aspect.

Just wanted to double check: do you deliberately want `sha512` blacklisted from the returned file documents?  Does this count as API breakage b/c that endpoint no longer includes the key?  I know S3 assetstore-d files wouldn't include the key anyway.

zachmullen:
> Does this count as API breakage b/c that endpoint no longer includes the key?

In the sense that every change is a breaking change :smile: 

Officially: the inclusion of assetstore impl-specific fields in returned file documents was simply a bug. As you've mentioned, downstreams that rely on the presence of this property should be considered broken anyway since they wouldn't support general files.

As a general policy: fields that have been explicitly whitelisted on a document are part of the API surface, and we will not remove them. However, accidental failure to filter returned documents is considered a bug and relying on this behavior is not supported.

PR #1123, opened 2015-12-02
Download content disposition

### LINKED ISSUE #1118: Be able to specify Content-Disposition=inline response header for downloads
I need to have pdfs displayed in the browser, and currently the only way I can get it to work is to have an HTML `object` element with the `data` attribute pointing to a file download path.  The problem is that I need to set

    cherrypy.response.headers['Content-Disposition'] = 'inline'

in order for this to display on Chrome and FF.  FF also requires me to change from the default PDF viewer to "Preview in Firefox", at least on Mac, but there's not much we can do about that here.  Safari works even without the change, which may mean it's doing the wrong thing, but it also works for "inline".

To support this, I was going to add an extra optional param (inlineContentDisposition=True) on the [File download endpoint](https://github.com/girder/girder/blob/7f9d0d14b051e8560e787f8d72f0e2f6c5e4beb5/girder/api/v1/file.py#L210-L247) and pass it through the File model, through the various assetstore_adapter implementations, and then back up to the abstract_assetstore_adapter [setContentHeaders](https://github.com/girder/girder/blob/7f9d0d14b051e8560e787f8d72f0e2f6c5e4beb5/girder/utility/abstract_assetstore_adapter.py#L199-L218) function.

Any concerns?

mgrauer:
This allows setting the content-disposition on a download of a single file.

This works for all files except for anonymously accessed files stored in S3 assetstores, due to a limitation from Amazon.  S3 allows you to specify the content-disposition parameter of a download URL, except in this case.

From the S3 [GET Object documentation](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html)

>You must sign the request, either using an Authorization header or a pre-signed URL, when using these parameters. They cannot be used with an unsigned (anonymous) request.

I've whitelisted two values for Content-Disposition at the REST layer, "inline" and "attachment", to address @brianhelba 's concerns.  

Still need to add tests on this PR.

Fixes #1118.

### DIFF
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -266,7 +266,15 @@ class File(Resource):
             if endByte is not None:
                 endByte = int(endByte)
 
-        return self.model('file').download(file, offset, endByte=endByte)
+        contentDisp = params.get('contentDisposition', None)
+        allowed = ['inline', 'attachment']
+        if contentDisp is not None and contentDisp not in allowed:
+            raise RestException('Unallowed contentDisposition type "%s".' %
+                                contentDisp)
+
+        return \
+            self.model('file').download(file, offset, endByte=endByte,
+                                        contentDisposition=contentDisp)
 
     @access.cookie
     @access.public
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -246,7 +246,7 @@ class File(Resource):
                required=False)
         .param('contentDisposition', 'Specify the Content-Disposition response '
                'header disposition-type value', required=False,
-               enum=['inline', 'attachment'])
+               enum=['inline', 'attachment'], default='attachment')
         .errorResponse('ID was invalid.')
         .errorResponse('Read access was denied on the parent folder.', 403)
     )
@@ -270,8 +270,8 @@ class File(Resource):
                 endByte = int(endByte)
 
         contentDisp = params.get('contentDisposition', None)
-        allowed = ['inline', 'attachment']
-        if contentDisp is not None and contentDisp not in allowed:
+        if (contentDisp is not None and
+           contentDisp not in {'inline', 'attachment'}):
             raise RestException('Unallowed contentDisposition type "%s".' %
                                 contentDisp)
 
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -244,6 +244,8 @@ class File(Resource):
                'so you should set it to the index of the byte one past the '
                'final byte you wish to receive.', dataType='integer',
                required=False)
+        .param('contentDisposition', 'Specify the Content-Disposition response '
+               'header disposition-type value', required=False)
         .errorResponse('ID was invalid.')
         .errorResponse('Read access was denied on the parent folder.', 403)
     )
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -245,7 +245,8 @@ class File(Resource):
                'final byte you wish to receive.', dataType='integer',
                required=False)
         .param('contentDisposition', 'Specify the Content-Disposition response '
-               'header disposition-type value', required=False)
+               'header disposition-type value', required=False,
+               enum=['inline', 'attachment'])
         .errorResponse('ID was invalid.')
         .errorResponse('Read access was denied on the parent folder.', 403)
     )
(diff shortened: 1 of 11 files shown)


mgrauer:
- [ ] add tests
- [ ] add filename to content-disposition response

mgrauer:
Ready for review.

If we decide to keep the whitelist of content-disposition for item and file download endpoints, we should unify it somewhere.  in girder.constants?

    allowed = ['inline', 'attachment']

zachmullen:
Whoops, I broke this... can you merge master

mgrauer:
merged

girder/api/v1/file.py
```diff
@@ -244,6 +244,8 @@ def readChunk(self, params):
                'so you should set it to the index of the byte one past the '
                'final byte you wish to receive.', dataType='integer',
                required=False)
+        .param('contentDisposition', 'Specify the Content-Disposition response '
+               'header disposition-type value', required=False)
```
zachmullen: We should use the `enum` kwarg to this function to enumerate the allowed strings.

girder/api/v1/file.py
```diff
@@ -266,7 +268,15 @@ def download(self, file, params):
...
 
-        return self.model('file').download(file, offset, endByte=endByte)
+        contentDisp = params.get('contentDisposition', None)
+        allowed = ['inline', 'attachment']
+        if contentDisp is not None and contentDisp not in allowed:
+            raise RestException('Unallowed contentDisposition type "%s".' %
+                                contentDisp)
+
+        return \
+            self.model('file').download(file, offset, endByte=endByte,
```
zachmullen: To avoid backslash here you can break after the opening paren `download(` and just indent the next line.

mgrauer:
@zachmullen 

I lost track of this, but have now addressed your comments.

brianhelba:
@mgrauer So I actually needed this branch last night, and I had to rebase it to my local deployment. I'd like to see it merged.

I have a few code comments though, I'll put them in.

girder/api/v1/file.py
```diff
@@ -244,6 +244,9 @@ def readChunk(self, params):
                'so you should set it to the index of the byte one past the '
                'final byte you wish to receive.', dataType='integer',
                required=False)
+        .param('contentDisposition', 'Specify the Content-Disposition response '
+               'header disposition-type value', required=False,
```
brianhelba: You can specify a ``default`` value here. Even though the API layer just sets it to ``None``, internally it will always default to ``attachment``.

girder/api/v1/file.py
```diff
@@ -266,7 +269,14 @@ def download(self, file, params):
             if endByte is not None:
                 endByte = int(endByte)
 
-        return self.model('file').download(file, offset, endByte=endByte)
+        contentDisp = params.get('contentDisposition', None)
+        allowed = ['inline', 'attachment']
```
brianhelba: ``allowed`` should be a set (just define the literal inside curly braces). You can just inline the variable too.

girder/utility/s3_assetstore_adapter.py
```diff
@@ -392,7 +392,13 @@ def downloadFile(self, file, offset=0, headers=True, endByte=None,
 
         if headers:
             if file['size'] > 0:
-                url = urlFn(key=file['s3Key'])
+                queryParams = {}
+                if contentDisposition == 'inline':
+                    # Tell S3 to set Content-Disposition response header,
+                    # though this only works for non-anonymous connections.
+                    queryParams['response-content-disposition'] = \
+                        'inline; filename="%s"' % file['name']
```
brianhelba: I assume S3 just defaults to ``attachment``? A code comment here noting that fact would be good.

mgrauer:
@brianhelba 

Girder is setting the Content-Disposition for S3 files on the initial [upload](https://github.com/girder/girder/blob/53a39f77892b13b4a7aae0640694d868e99c465d/girder/utility/s3_assetstore_adapter.py#L177-L178) to be "attachment; filename={$filename}" and then can be changed on any [updates](https://github.com/girder/girder/blob/53a39f77892b13b4a7aae0640694d868e99c465d/girder/utility/s3_assetstore_adapter.py#L501-L508).

I've addressed your comments.

brianhelba:
LGTM.

brianhelba:
LGTM.

PR #1124, opened 2015-12-03
Ensure job detail view updates automatically on status change

zachmullen:


mgrauer:
Works both in Girder and in the embedded widget (tested in Minerva).

Travis is unhappy, I saw some pip installation errors in a Python 3 build, restarted Travis.

LGTM.

PR #1125, opened 2015-12-03
[WIP] Add a setting in the web client to allow CORS via restRequest

zachmullen:
@jbeezley Can you let me know whether this fixes your CORS issue? Somewhere in your code before you start the girder app, just add the following line:

`girder.allowCors = true;`

### DIFF
--- a/clients/web/src/init.js
+++ b/clients/web/src/init.js
@@ -27,6 +27,7 @@ _.extend(girder, {
     staticRoot: $('#g-global-info-staticroot').text().replace(
         '%HOST%', window.location.origin),
     currentUser: null,
+    allowCors: false,
     events: _.clone(Backbone.Events),
     uploadHandlers: {},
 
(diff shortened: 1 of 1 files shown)


zachmullen:
Also it occurs to me we may want to change this setting name to `allowCorsAuth`, and also I'd be fine with having it default to `true`.

cpatrick:
This is going to require some documentation :smile_cat:

jbeezley:
I pushed this branch and added `<script>girder.allowCors=true;</script>` after `<script src="./static/built/app.min.js"></script>`, but that didn't solve the problem.

jbeezley:
This is another issue to be aware of when using this: 
```
A wildcard '*' cannot be used in the 'Access-Control-Allow-Origin' header when the credentials flag is true. Origin 'http://apps.bsvecosystem.net' is therefore not allowed access.
```

zachmullen:
I'm going to close this; the actual solution will not require this flag anyway.

PR #1128, opened 2015-12-03
CORS auth utilities

zachmullen:
ping @jbeezley you should now be able to set:

    girder.corsAuth = true;

Prior to girder app startup, and all should be well.


jbeezley:
Works like a charm.  LGTM

PR #1129, opened 2015-12-03
Add a superclass constructor call to the API's Resource class

brianhelba:
This fixes a warning during Girder startup.

### DIFF
--- a/girder/api/v1/resource.py
+++ b/girder/api/v1/resource.py
@@ -35,6 +35,7 @@ class Resource(BaseResource):
     API Endpoints that deal with operations across multiple resource types.
     """
     def __init__(self):
+        super(Resource, self).__init__()
         self.resourceName = 'resource'
         self.route('GET', ('search',), self.search)
         self.route('GET', ('lookup',), self.lookup)


zachmullen:
LGTMITIH

brianhelba:
>  LGTMITIH

Neither me nor Google know what that means, but it matches ``/^LGTM.*$/`` and the tests pass.

Merging.

jeffbaumes:
Oh, I think I know: "if travis is happy"

zachmullen:
:bell: :bell: :bell: That's a bingo!

cpatrick:
@mgrauer pulled this one out in another PR:

> TIRFMPBIWTTADITPRWI

I'm lost...

jbeezley:
My favorite part of searching google for that acronym is there are two results, each of which are in German text.

zachmullen:
fo**lgt mit Ih**rem

jeffbaumes:
Need to stop now or else I'll try Googlewhacking all afternoon https://en.wikipedia.org/wiki/Googlewhack

mgrauer:
@cpatrick It means `this is ready from my perspective but I want to trigger a discussion in the PR with it`.

It was actually a @jbeezley innovation.  I stood on the shoulders of giants.

mgrauer:
> LGTMITIH

I was going to criticize @zachmullen for being too obvious.

jbeezley:
ISOIA (I speak only in acronyms).

PR #1130, opened 2015-12-04
Create a filtermodel decorator

zachmullen:
@mgrauer PTALSTWYIO. If you are OK with how this looks, I will go ahead and convert all of our REST endpoints over. This will be a good exercise; in the folder resource alone I found two endpoints that were failing to filter their results.

### DIFF
--- a/girder/models/item.py
+++ b/girder/models/item.py
@@ -347,7 +347,7 @@ class Item(acl_mixin.AccessControlMixin, Model):
 
         :param item: The item whose root to find
         :type item: dict
-        :param user: The user running the test (not required if force=True).
+        :param user: The user making the request (not required if force=True).
         :type user: dict or None
         :param force: Set to True to skip permission checking. If False, the
             returned models will be filtered.
--- a/girder/models/item.py
+++ b/girder/models/item.py
@@ -347,6 +347,11 @@ class Item(acl_mixin.AccessControlMixin, Model):
 
         :param item: The item whose root to find
         :type item: dict
+        :param user: The user running the test (not required if force=True).
+        :type user: dict or None
+        :param force: Set to True to skip permission checking. If False, the
+            returned models will be filtered.
+        :type force: bool
         :returns: an ordered list of dictionaries from root to the current item
         """
         curFolder = self.model('folder').load(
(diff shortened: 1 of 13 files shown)


mgrauer:
> PTALSTWYIO

See if This is What You Intended Originally?

Such That Walrus Yawping Is Obstreperous?


Yes to both.  I like it.

zachmullen:
> it seems like I am adding fields to a filter.

You are adding fields to the filter. The filter says what fields are allowed in the document, and you're adding other fields to that filter... maybe the word "filter" itself is a source of confusion here?

mgrauer:
Indeed, I think of "filter" as what is removed. _.js uses "pluck" for the operation of passing through certain selected values.

I can live with it if the docstring clarified a bit more.

zachmullen:
Ah I see. I'll add more to the docstring to indicate that this is the "whitelist" sort of filter. I'm used to thinking of a filter as specifying what passes through rather than what doesn't, but I can definitely see the ambiguity.

zachmullen:
> PTALSTWYIO

Since this was your idea originally. I think I've addressed all your comments, so if you like what you see I will convert all the existing endpoints over.

mgrauer:
LGTM.  Go for it.

:checkered_flag:

zachmullen:
@girder/developers PTAL at this change, this will touch a lot of REST code. It's another step towards reducing a lot of our boilerplate into a set of decorator annotations.

jcfr:
:+1:

zachmullen:
Everything I care to convert for the moment has been converted.

zachmullen:
Fixed.

girder/models/item.py
```diff
@@ -339,6 +347,11 @@ def parentsToRoot(self, item, user=None, force=False):
 
         :param item: The item whose root to find
         :type item: dict
+        :param user: The user running the test (not required if force=True).
```
mgrauer: `running the test` ?  seems out of place here

zachmullen: This is one of those "did I really write that?" moments.

mgrauer:
I'm wondering if there is an inconsistency in when you add the `access` to the filter, but looking through the code, I couldn't find any.  There's a lot of code though.  Can you summarize the policy for when to add `access` to the filter?

zachmullen:
We more or less throw it into the document for any request where it could have changed. This minimizes the number of requests that client needs to make to the server. (At least, that's the goal... the execution might not be 100% correct.)

mgrauer:
> Everything I care to convert for the moment has been converted.

    f(sizeof(codebase), sizeof(your_heart))

LGTM.

PR #1132, opened 2015-12-08
Allow filtermodel to work for plugin models also

zachmullen:
@mgrauer this was a silly oversight on my part, PTAL.

### DIFF
--- a/plugins/jobs/server/job_rest.py
+++ b/plugins/jobs/server/job_rest.py
@@ -19,8 +19,8 @@
 
 from girder import events
 from girder.api import access
-from girder.api.describe import Description
-from girder.api.rest import Resource, loadmodel
+from girder.api.describe import Description, describeRoute
+from girder.api.rest import Resource, filtermodel, loadmodel
 from girder.constants import AccessType, SortDir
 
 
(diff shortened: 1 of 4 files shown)


plugins/jobs/server/job_rest.py
```diff
@@ -48,54 +57,26 @@ def listJobs(self, params):
             user = self.model('user').load(
                 params['userId'], user=currentUser, level=AccessType.READ)
 
-        jobs = self.model('job', 'jobs').list(
+        return list(self.model('job', 'jobs').list(
```
mgrauer: any reason you are preferring `list` over `[]`?

mgrauer:
I'm not impressed, I've seen much sillier oversights.

LGTM.

plugins/jobs/server/job_rest.py
brianhelba: @mgrauer I'm not sure whether you mean a list comprehension or a list literal. 
I don't see how either is really appropriate here, the constructor seems correct to me.

brianhelba:
LGTM.

Note that there is [a ``.. deprecated::`` block](http://sphinx-doc.org/markup/para.html#directive-deprecated), providing pretty formatting.

plugins/jobs/server/job_rest.py
mgrauer: I didn't read it carefully enough, thanks @brianhelba .

PR #1134, opened 2015-12-08
Fixes #1133. Allow raw responses to be specified for REST routes

### LINKED ISSUE #1133: Allow raw response to be sent without requiring generator
Right now a raw response can only be sent by returning a generator function that yields the output as a stream. In fact, "streaming" and "raw" can be orthogonal, so we should expose a simple function like `girder.api.rest.Resource.setRawResponse(self, val=True)` that a route handler could call to specify that its response should not be transformed.

ping @mgrauer this would solve the problem you described in your email just now, and would be simple to implement.

zachmullen:
@mgrauer @jcfr PTAL

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -305,12 +305,42 @@ class filtermodel(ModelImporter):  # noqa: class name
         return wrapped
 
 
+def setRawResponse(val=True):
+    """
+    Normally, non-streaming responses go through a serialization process in
+    accordance with the "Accept" request header. Endpoints that wish to return
+    a raw response without using a streaming response should call this, or use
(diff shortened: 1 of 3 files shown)


jcfr:
Thanks. 

Beside of few nitpicks. LGTM

zachmullen:
> getattr(cherrypy.request, 'girderRawResponse') is True:

> girderRawResponse is expected to exist.

Not sure what you mean by this... that attribute only gets set if `setRawResponse` is called, so we have to provide a default value.

jcfr:
> Not sure what you mean by this... that attribute only gets set if setRawResponse is called

Nevermind then.

jcfr:
LGTM :+1:

PR #1135, opened 2015-12-09
Ensure that custom access levels are respected

brianhelba:
In theory, an access level can be any numeric value.


zachmullen:
Looked at this twice. LGTM!

PR #1136, opened 2015-12-09
Add some branding to the docs

cpatrick:


aashish24:
Yay !!

zachmullen:
We are calling both girder and resonant a "platform"; is that how we officially want to brand them both? Is this equivocating on the meaning of the word "platform"?

jeffbaumes:
In this context, perhaps ecosystem is a better descriptive word for Resonant to avoid repeating the word, though both words seem nearly synonymous.

https://www.google.com/trends/explore#q=hadoop%20platform%2C%20hadoop%20ecosystem&cmpt=q&tz=Etc%2FGMT%2B5

ronichoudhury:
I think it's ok to call both things "platforms". Platforms are for building
applications on top of, which is possible both with Resonant and Girder.
The applications you build with each are of different kinds, hence "data
management platform" vs. "analytics and visualization platform".

zachmullen:
To clarify, I agree that it's OK to call both things platforms, just making sure it's OK to do so in the same sentence since we do mean *slightly* different things.

ronichoudhury:
Gotcha - I think repeating the word in this sentence is actually good. I
see no reason to downplay either component's "platformhood". However, we
could insert a word before the Resonant part, stressing the part-whole
relationship between Girder and Resonant:

"as part of the wider Resonant data and analytics platform"

Perhaps not "wider" but something to underline the difference between these
things.

jeffbaumes:
+1 to either "wider platform" or "ecosystem", I would LGTM either one. I agree it is more of a writing style question.

cpatrick:
I like ecosystem the best. How's that?

cpatrick:
@jeffbaumes @zachmullen @ronichoudhury @aashish24 lgtm?

zachmullen:
LGTM

jeffbaumes:
LGTM. Thanks for the bit o' branding.

PR #1138, opened 2015-12-10
Relax dependency versions in setup.py

jbeezley:
This PR is intended to ease the dependency hell that exists in downstream repos.  I removed all of the versioning information from `setup.py` except for `>=3.0` on `pymongo` because the is a known breakage.  If anyone knows of any similar hard requirements for the other packages, let me know.  For the most part, it doesn't matter much because [pip doesn't do real dependency resolution](https://github.com/pypa/pip/issues/988).

A couple of other small issues and improvements are addressed here as well:
  * Improved caching on travis (including updated pip that does caching automatically)
  * Fixed a race condition on the packaging tests
  * Added extra dependencies for plugins.  One for each plugin + a `plugins` extra that installs all
  * Downgraded jsoneditor to avoid requiring git on deployments

I still need to go over the documentation.  I also want to try add a new packaging test for the optionals if I can figure out a quick way to test that the plugin imports work.


zachmullen:
This seems to introduce some duplication that makes me a bit nervous.

* Duplication of full requirements lists between the requirements files and setup.py.
* Duplication of plugin requirements between their requirements files and the main requirements file.

Could we maybe avoid some of this if we expose another entry point that does the pip install for an individual plugin?

zachmullen:
Also, would it be possible to have setup.py still read the requirements files, but simply strip off all the versions except in special known cases like pymongo?

jbeezley:
> Duplication of full requirements lists between the requirements files and setup.py.

Because of the different intents of those two lists, I don't agree that it is a duplication.  They are highly correlated though.  One problem we would have trying to parse the requirements files is that doing so is [explicitly](https://github.com/pypa/pip/issues/2286#issuecomment-68285791) not supported.  Also, technically the requirements file should be a full [freeze](https://pip.readthedocs.org/en/stable/user_guide/#ensuring-repeatability) of every package, not just the direct dependencies.  I haven't done that yet, but I can if people think it is important.  I don't view the requirements.txt as something people should use outside of a CI environment.  I'm updating the RTD pages to reflect that point of view.  If pip did a better job of dependency resolution, the requirements files would be a lot like npm's [shrinkwrap files](https://docs.npmjs.com/cli/shrinkwrap#building-shrinkwrapped-packages) that contain similarly duplicated information.

> Duplication of plugin requirements between their requirements files and the main requirements file.

The plugin requirement files only exist right now to support the depreciated InstallPythonRequirements script.  Once that is removed, we should remove those files as well.

zachmullen:
Ok, thanks for the explanation. Sounds like this will be fine once you add your documentation changes.

jbeezley:
@zachmullen PTAL

I updated the devops to get rid of the requirements file usage as well.  I'm willing to remove that bit if you don't think it's a good idea.  My rational is that if there is a chance the user will pip install anything else into the python environment after girder, then the pinned version numbers should not be used.

I also didn't address the `girder-install` issue because that will be a breaking change with existing plugins.  I only removed the `-U` flag that it uses so that it keeps the current versions of dependencies intact rather than the prior behavior of  ¯\_(ツ)_/¯.  I think we can address that at a later point once we flesh out more the plugin pip packaging workflow.

jbeezley:
I added to this PR a full virtualenv freeze to protect against changes in deps of deps in our CI testing.  I also added a simple packaging test to assert that `pip install girder[plugins]` actually installs `celery` (an optional requirement).

zachmullen:
Can we also add something to the installation docs that recommends installing within a virtualenv?

jbeezley:
@zachmullen virtualenv instructions added.

zachmullen:
LGTM other than the one typo fix

jbeezley:
Amended.  I'll merge when the tests pass.

PR #1140, opened 2015-12-14
Send timestamps as ISO8601 format

zachmullen:
I plan to integrate this into my other outstanding topic branch, but wanted to put this forth as its own topic for discussion first. This change makes it much easier for clients to handle our date strings by putting them into the most widely-accepted standard format. Basically any date parser in existence should handle this.

@girder/developers PTAL


jbeezley:
This is a good change to make.  One additional thing we might consider is handling time zones.  Ideally, the returned string should be in UTC (thus ending in a `Z`).  I'm not sure what mongo provides when it returns datetime objects.

zachmullen:
It provides tz-agnostic stamps, but by convention we always use UTC stamps. So I suppose we should go ahead and add a Z or somehow pass in the tz info when serializing. I'll look into it.

zachmullen:
@jbeezley PTAL.

zachmullen:
@jbeezley be advised, I've added a new core requirement (pytz), which I imagine will affect your outstanding branch.

jbeezley:
:+1:

cpatrick:
If I'm reading this correctly, it subtly changes the API by changing the timestamp format on every endpoint that includes one.

zachmullen:
It changes the format of certain strings returned by the API. That said, the only way that such a format change would break any downstream code would be if someone had rolled their own date parser that only understood the (non-standard) date format that girder was sending before, which seems unlikely.

cpatrick:
I agree. I would make the case that this should increment the minor version.

zachmullen:
Yep, definitely. And now would be a good time to actually make the date format returned via REST as a guarantee by our API, now that we've moved to the most widely-accepted, standard date format possible.

zachmullen:
Another question is, should we remove the [old hack](https://github.com/girder/girder/blob/master/clients/web/src/utilities/MiscFunctions.js#L15) that was in place to workaround this.

jbeezley:
I'm leaning slightly towards keeping it, but I don't have a strong opinion.

jcfr:
> remove the old hack that was in place 

May be an email could be send on the user list


On a somewhat related topic ... 

If not already the case and if you think it would be helpful, would it make sense to include a parameter named `timezone` to every endpoint returning date/time. That way the server would take care of converting the date/time to the expected timezone ?

zachmullen:
@jcfr For now, I'd rather leave conversion of the stamps up to the clients (since that's a display issue) and have the server simply provide them with a tz-aware, unambiguous stamp that they can use how they wish.

jcfr:
>  conversion of the stamps up to the clients

Make sense

If relevant, should the girder js client code be updated to clearly indicate the date returned (for example in the metatdata viewer/editor) at in UTC ?

zachmullen:
This change actually alleviates the need to specify such a convention; now that the stamps contain timezone information, they are unambiguous. As an example, in JavaScript, passing the string to `new Date(str)` will always work on all browsers, and can be trivially displayed in the local timezone since the tz info is encoded in the stamp.

As for your example of the metadata editor, to my knowledge the metadata utilities don't actually do anything special related to date types; we just naively store whatever string is passed in. Please correct me if I'm wrong on that.

jcfr:
> will always work on all browsers, and can be trivially displayed in the local timezone

Great

> don't actually do anything special related to date types

Roger that.

zachmullen:
This is good to go now, @jbeezley please take one final look.

jbeezley:
LGTM

PR #1142, opened 2015-12-16
Gravatar plugin was causing an error on user info update

zachmullen:
Non-admin users were getting a cryptic "Administrator access
required" error when trying to update their user information. This
was because the gravatar plugin had registered a handler to a
REST hook without declaring an access level on itself.

@kotfic this fixes the bug I was running into when testing your branch.


zachmullen:
This is blocking a COVALIC ansible script, can someone PTAL?

cpatrick:
This LGTM.

PR #1144, opened 2015-12-17
Update Python requirement for Pillow to 3.0.0

brianhelba:
This is the latest version, which some other plugins may require.

The WIP [``large_image`` plugin](https://github.com/DigitalSlideArchive/large_image) requires v3.0.0.


jbeezley:
The Pillow [changelog](https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst#300-2015-10-01) seems reasonable.  LGTM.

PR #1145, opened 2015-12-23
Fix Backbone collection to allow altUrl without resourceName

danlamanna:
This fixes 2 issues:

1) `resourceName` isn't strictly required for a collection to properly fetch, an `altUrl` can instead be specified.
2) This assumes every object in this response will have an `_id` attribute which might not be the case for other applications using Girder. This is already handled by Backbone in the case of Girder models [here](https://github.com/girder/girder/blob/475bbf5fa4a5cf9f4beb41da03879fd619583cc7/clients/web/src/model.js#L7).


zachmullen:
LGTM.

PR #1146, opened 2015-12-26
Prefix search mode

zachmullen:
@jcfr you (and other users) have mentioned a desire for this feature. I still need to add testing for it and convert search fields in the web app where appropriate.


zachmullen:
Ready for review.

grunt_tasks/build.js
```diff
@@ -73,9 +73,22 @@ module.exports = function (grunt) {
...
+                    src: ['img/**'],
                     dest: 'clients/web/static/built/jsoneditor'
                 }]
+            },
+            bootstrap: {
+                files: [{
+                    expand: true,
+                    cwd: 'node_modules/bootstrap/dist/css',
+                    src: ['bootstrap.min.css'],
+                    dest: 'clients/web/static/built'
```
msmolens: The path to bootstrap.min.css should be updated in:
- https://github.com/girder/girder/blob/6fe7e9538227939ce62f70e8bd83f08fb7d1af0e/clients/web-external/index.html#L2
- https://github.com/girder/girder/blob/6fe7e9538227939ce62f70e8bd83f08fb7d1af0e/docs/external-web-clients.rst#using-girder-register-and-login-ui

msmolens:
On Chrome 47.0.2526.106 m (Windows) the plugin switches have a height problem:
switch

msmolens:
This looks like a useful enhancement and works as specified. I think it could be even more useful if the prefix search applied to each token instead of the entire string. For example, in Covalic, I'd like to search by typing the first few letters of any word in a challenge's name, not just the first word.

grunt_tasks/build.js
zachmullen: Good catch, thanks

zachmullen:
@msmolens I agree about the usefulness of prefix search of each word, however I think we probably want to look into other search technologies to achieve that rather than trying to implement it ourselves. It's not something we can trivially do using `$regex` or `$text` with mongo. (Admittedly, we could probably do it with `$regex`, but such a search would not be able to use indexes, and I consider that a deal-breaker.) We could look into something like Solr/Lucene or ElasticSearch, both of which I believe are designed to do that sort of thing in O(log n).

zachmullen:
Should be good to go now. I fiddled with the CSS for the bootstrap switches, but it actually appears to stem from some erroneous calculations of widths that takes place in the javascript, so I just removed the default "OFF" text that was breaking it.

clients/web/src/views/widgets/SearchFieldWidget.js
```diff
@@ -53,6 +49,23 @@ girder.views.SearchFieldWidget = girder.View.extend({
         }
     },
 
+    search: function () {
+        var q = this.$('.g-search-field').val();
+
+        if (!q) {
+            this.hideResults();
+            return;
```
cpatrick: This seems like odd return behavior compared to the normal case?

girder/models/model_base.py
```diff
@@ -281,6 +281,45 @@ def textSearch(self, query, offset=0, limit=0, sort=None, fields=None,
...
+        name, and the second element is a string representing the regex search
+        options.
+
+        :param query: The prefix string to look for.
+        :type query: str
+        :param filters: Any additional query operators to apply.
+        :type filters: dict
+        :param prefixSearchFields: To override the model's prefixSearchFields
+            attribute for this invocation, pass an alternate iterable.
+        :returns: A pymongo cursor. It is left to the caller to build the
```
cpatrick: A couple of things:

1. Should we return a pymongo cursor here? That seems to leak the model abstraction a bit. I think we already do this in other places, so that may be out-scope for this PR. I wanted to mention it here, though.
2. Is there equivalent user docs for this functionality?

cpatrick:
A couple of nits, but LGTM. I'd like @jcfr to comment before a merge, though.

girder/models/model_base.py
zachmullen: 1. We can document it as simply an "iterable of documents" if it makes more sense. The cursor has some additional functionality on it e.g. `count()` that could be useful, though.
2. Not sure what you mean by this question, could you elaborate?

zachmullen:
Any other comments on this branch or can it be merged?

jcfr:
Thanks for working on this.

> prefix search applied to each token instead of the entire string.

Assuming a username is "John Doe", and I type "doe", does it mean it won't auto complete ?

zachmullen:
It will work for doe. :)

jcfr:
Great. LGTM

Out of curiosity, it works because first and last name are in separate field ?

zachmullen:
Correct.

PR #1152, opened 2016-01-01
Make job status values extensible

zachmullen:
We add a simple event hook for allowing job status values
beyond the ones built into the jobs plugin.

We will need this for https://github.com/Kitware/romanesco/pull/200

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -39,22 +39,33 @@ plugins that are in charge of actually scheduling a job for execution should the
 call ``scheduleJob``, which triggers the ``jobs.schedule`` event with the job
 document as the event info.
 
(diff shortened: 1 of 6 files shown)


cpatrick:
LGTM. I'd like some more docs, but this *is* a simple feature.

zachmullen:
@cpatrick good thought, I added more docs.

zachmullen:
@cpatrick PTAL I've also made new changes to the web client code related to job statuses.

cpatrick:
I like the docs and changes. I'm slightly bothered by the duplication between JS and Python, but I'm not sure the best way around that. All things considered, LGTM.

PR #1155, opened 2016-01-09
Allow the "fields" parameter to "Model.load" to be used with "force"

brianhelba:
This change only has an effect in the case that the caller sets both "force" and "fields" (neither of which is the default).

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -218,7 +218,7 @@ class Model(ModelImporter):
         :param sort: The sort order.
         :type sort: List of (key, order) tuples.
         :param fields: A mask for filtering result documents by key.
-        :type fields: List of strings
+        :type fields: list[str]
         :param timeout: Cursor timeout in ms. Default is no timeout.
         :type timeout: int
         :returns: A pymongo database cursor.
@@ -933,7 +933,7 @@ class AccessControlledModel(Model):
         """
 
         # Ensure we load access and public, these are needed by requireAccess
-        loadFields = None
+        loadFields = fields
         if not force and fields:
             loadFields = list(set(fields) | {'access', 'public'})
 


brianhelba:
Ping @cjh1 . This code was originally written by you.

zachmullen:
I think the bug here is that instead of `loadFields = None` it should read `loadFields = fields`.

brianhelba:
@zachmullen I don't think so. ``force`` should have no effect on which fields are finally returned. If ``load(..., force=True, fields=['_id', 'foo'])`` is called, then the return value should not also contain ``access`` and ``public`` fields.

zachmullen:
Right, my point is that the addition of `access` and `public` is only required if `force=False`, which is why the conditional was written as such in the first place.

brianhelba:
Doh. Yup, you're right; I wasn't looking at this correctly.

Updated.

brianhelba:
I don't have time right now to do it myself, but the existence of this bug and the fact that my incorrect first fix didn't cause failures means this isn't tested well enough.

Reliably ``load``ing only those fields which are requested could have security implications for some usages. This needs adequate test coverage.

zachmullen:
It's not that your first fix wasn't correct, it just made it slightly less efficient than it needs to be in the `force` case. And before, it wasn't so much a bug as a lack of functionality (although that is arguable). I'm all for adding more testing, but I don't consider it a necessity before we merge this branch. LGTM.

cpatrick:
@brianhelba, @zachmullen don't forget to create the ticket prior to merging. :smile_cat:

brianhelba:
My first fix would have caused the ``access`` and ``public`` fields to be returned as extra fields any time ``force==True`` and ``fields`` was set, since the removal code would have been skipped. That's definitely a bug.

brianhelba:
@cpatrick Done: https://github.com/girder/girder/issues/1158

Issue #944, opened 2015-09-05
Add Javascript style checking / linting to Jade templates and test specs


brianhelba:
See https://github.com/girder/girder/pull/939 for beginning discussion of this issue.

brianhelba:
This has been fixed.

PR #1159, opened 2016-01-11
Converting core javascript style tests to eslint

### LINKED ISSUE #800: document JSHINT_EXTRA_CONFIGS and JSSTYLE_EXTRA_CONFIGS
Makes the most sense to document this in the plugin development section of the docs.  

Show how you would use this in a plugin.

### LINKED ISSUE #1150: Use eslint rather than jshint/jscs
I'm adding this issue for discussion/debate.  I have been doing some modernization of my javascript toolset and found that eslint is vastly superior jshint and jscs in a number of ways.
* [Babel](http://eslint.org/docs/2.0.0/user-guide/configuring.html#specifying-parser) can be used as a parser to support new language features
* [Extensible](http://eslint.org/docs/2.0.0/user-guide/configuring.html#extending-configuration-files) configuration files built-in so we no longer need to merge json files ourselves.
* [Integration](http://eslint.org/docs/1.10.3/user-guide/integrations.html#build-systems) with about every js build system I heard of
* [Plugins](https://www.npmjs.com/package/eslint-plugin-filenames) to enforce file name and path conventions
* About a bazillion other [plugins](https://www.npmjs.com/browse/keyword/eslintplugin)
* All the cool kids are using it<sup>[citation needed]</sup>

I have also found a fairly new "[standard](http://standardjs.com/)" javascript style guidelines that seems to be taking off.  It is trying to be something like what pep8 is for python.  The npm package installs a runnable script, but it is really just an ESLint configuration.  Some of style choices are a bit controversial, in particular the "no semicolons" rule.  (There is a variant called "[semistandard](https://github.com/Flet/semistandard)" that uses semicolons.)  One of the nice things about it is that there is a tool ([standard-format](https://github.com/maxogden/standard-format) or [semistandard-format](https://github.com/ricardofbarros/semistandard-format)) that will do most of the formatting for you, so it is a fairly painless process to do the conversion.

The "standard" conversion would be a pretty big change from the current code style, so I'm not really recommending using it at this point.  ESLint on the other hand can be configured to support the current code style.

jbeezley:
For this, I'm using the [semistandard](https://github.com/Flet/semistandard) as closely as possible without making any major code changes.  The overrides from the standard are in the [config file](https://github.com/girder/girder/blob/4568313f029d25358be5200a79c2dbd6a625e7c6/.eslintrc).  I left one rule (`no-throw-literal`) as a warning because fixing the few issues of it would technically be backwards incompatible.  I also didn't touch the plugins.  We'll have to leave jshint/jscs for downstream plugins anyway.  After this goes in, we should consider jshint/jscs as depreciated.

Fixes #1150.
Fixes #800.


clients/web/src/utilities/S3UploadHandler.js
```diff
@@ -90,7 +90,7 @@
             };
 
             xhr.upload.addEventListener('progress', function (event) {
-                handler._xhrProgress.call(handler, event);
+                handler._xhrProgress(event);
```
zachmullen: The fact that it catches this and our old tools didn't is pretty nice :)

scripts/json_config_expand_relpaths.js
```diff
@@ -97,7 +97,7 @@ var config = json.parse(fs.readFileSync(inputfile, {encoding: 'utf8'}).toString(
 // Expand relative paths
 
 if (typeof String.prototype.startsWith !== 'function') {
```
brianhelba: I think it should be fine to remove this whole block. ``String.prototype.startsWith`` is part of [ECMAScript6, and would only be unavailable on IE < 12](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith). Since this is an internal utility script, and not compiled into the web client, I'd think compatibility fixes should be unnecessary.

Ping @jcfr , since he added it.

jcfr: Makes sense.

jbeezley: Actually as it turns out, the polyfill is necessary for node (v0.10 at least).

brianhelba:
Heads up, this branch is going to have merge conflicts due to https://github.com/girder/girder/pull/1161 , specifically from to [changes in ``package.json``](https://github.com/girder/girder/pull/1161/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L14).

jbeezley:
PTAL

I added backbone and underscore plugins with most of the rules set as warnings.  I didn't add the jasmine plugin because we aren't actually linting our javascript tests at the moment anyway.  I also added a bit of documentation about extending the eslint rules for plugins in the spirit of #800.

tests/clients/web/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
 # All of our core javascript will be checked with a single test
-add_javascript_style_test(core src)
```
jcfr: Since we are keeping the function `add_javascript_style_test` around, would it make sense to style execute it  (that way it would be tested) ? and if executed outside of girder core ... it would display a CMake author warning. ?

jcfr:
Very nice :+1: 

Also really like the documentation you added.

.eslintrc
```diff
@@ -0,0 +1,63 @@
...
+        "backbone": {
+            "Collection": [
+                "girder.Collection"
+            ],
+            "Model": [
+                "girder.Model"
+            ],
+            "View": [
+                "girder.View",
+                "girder.views.MetadatumEditWidget"
```
danlamanna: What is the significance of this widget being a base view?

tests/clients/web/CMakeLists.txt
danlamanna: I might be missing something @jbeezley, but is this running on project_root/src or project_root/clients/web/src? I picked a random file in clients/web/src and started changing indentation and removing semicolons and everything still passed.

docs/plugin-development.rst
```diff
@@ -437,6 +437,49 @@ that can be used to import content:
   plugin that doesn't fall into one of the above categories can be placed here,
   such as static images, fonts, or third-party static libraries.
 
+Linting and Style Checking Client-Side Code
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Girder's uses `ESLint <http://eslint.org/>`_ to perform static analysis of its
```
danlamanna: Girder*

.eslintrc
jbeezley: That is how the plugin detects if the object is a view or not.  By default, it will only check against objects that follow this pattern:
```
Backbone.View.extend({...})
```
This makes it also check these cases
```
girder.View.extend({...})
girder.views.MetadatumEditWidget.extend({...})
```
I didn't add any other views because we never use them as base objects.

tests/clients/web/CMakeLists.txt
jbeezley: Yes, I think you're right.  That's one of the problems I've found with eslint; it doesn't error when you give it a non-existent path.  I'll fix it up later tonight.

As far as linting the test files, I think that would be most appropriate for a separate PR.

brianhelba: If you do start linting the test files, you can solve part of https://github.com/girder/girder/issues/944.

zachmullen: > it doesn't error when you give it a non-existent path

We should write a different JS style tool that doesn't have that bug! :wink:

brianhelba:
The documentation improvements should now close https://github.com/girder/girder/issues/800.

tests/clients/web/CMakeLists.txt
brianhelba: > We should write a different JS style tool that doesn't have that bug! :wink:

Only if we can concurrently invent several brand-new style standards: "semi-unified", "unified", and "grand unified", all of which will be objectively vastly superior to existing standards. Then we'll add a [Greenkeeper](http://greenkeeper.io/)-style feature to automatically send PRs to change the style in all repos that do not use our standard.

cpatrick: [image](http://www.reactiongifs.us/wp-content/uploads/2013/06/raining_david_tennant.gif)

jbeezley: Now we're thinking like javascript developers!

jbeezley:
@danlamanna I fixed the issues you raised.
@jcfr I deprecated the old style tests with an `AUTHOR_WARNING`.  I didn't put any tests in, but I can if people think it's worth it.

danlamanna:
I have a couple of notes:

1) So there are like ~170 warnings, it would be a pain to deal with them, but it also makes it a pain for people developing to realize which warnings they've added - which I think will lead to a larger number of warnings that are more or less ignored?

2) flake8, jshint, jscs configs are all in `/tests`, should `.eslintrc` move to the same place to be consistent?

jbeezley:
1) I don't know.  Fixing all of the warnings is pretty easy, but I can't spend the time doing so at the moment.  Maybe the answer is that this needs to wait until someone has time to fix the issues.

2) I put it in the main directory so that my editor detects the local style configuration automatically.  That is best practice in my opinion.

brianhelba:
> Fixing all of the warnings is pretty easy, but I can't spend the time doing so at the moment. Maybe the answer is that this needs to wait until someone has time to fix the issues.

@jbeezley Can we at least add an issue documenting where these warnings are displayed, and the general process for going about fixing them? This would be another good "onramp" issue.

jbeezley:
It would be a good "onramp" issue.  The question is whether we will have people looking to do onramp issues anytime soon.  I would like to avoid having the warnings there as much as possible.  I'll probably just fix them myself this weekend.  I doubt it will take more than an hour.

jbeezley:
Alright, I went through the warnings and fixed what I could fix quickly and disabled the rest.  The following disabled rules I'll open issues for:

* [`no-throw-literal`](http://eslint.org/docs/rules/no-throw-literal): Easy to fix and only three or four instances of it, but there are possible backward compatibility issues.
* [`backbone/model-defaults`](https://github.com/ilyavolodin/eslint-plugin-backbone/blob/master/docs/rules/model-defaults.md): I don't know enough about girder models to understand what the defaults should be.
* [`backbone/no-native-jquery`](https://github.com/ilyavolodin/eslint-plugin-backbone/blob/master/docs/rules/no-native-jquery.md): I think this one is a good idea in general, but it would take a bit of refactoring to fix all of the cases where this occurs.
* [`backbone/no-constructor`](https://github.com/ilyavolodin/eslint-plugin-backbone/blob/master/docs/rules/no-constructor.md): I don't understand the code well enough to know why the constructor is used [here](https://github.com/girder/girder/blob/master/clients/web/src/view.js#L2).
* [`backbone/no-view-onoff-binding`](https://github.com/ilyavolodin/eslint-plugin-backbone/blob/master/docs/rules/no-view-onoff-binding.md): This one is a persistent issue in this code that we should address at some point.  I suspect that a major reason why we need to register parent views and do a bunch of manual unbinding of events is because we are calling `on` rather than `listenTo`.

I recommend we keep the following rules disabled:

* [`underscore/prefer-underscore-method`](https://github.com/captbaritone/eslint-plugin-underscore/blob/master/docs/rules/prefer-underscore-method.md): This rule is very overzealous and results in a lot of false positives.
* [`backbone/event-scope`](https://github.com/ilyavolodin/eslint-plugin-backbone/blob/master/docs/rules/event-scope.md): Most of the time the context is not required for the event handlers and always requiring it goes a little overboard in my opinion.  Although, I did get a [PR](https://github.com/ilyavolodin/eslint-plugin-backbone/pull/54) accepted to fix an issue with this rule.

jbeezley:
I removed [`underscore/prefer-noop`](https://github.com/captbaritone/eslint-plugin-underscore/blob/master/docs/rules/prefer-noop.md) as well because `_.noop` doesn't exist in our version and updating it caused our js coverage handlers to break.

zachmullen:
> backbone/no-constructor: I don't understand the code well enough to know why the constructor is used here.

We should just disable that warning locally for that one view, it's our base view and we make use of the constructor to impart our memory management system on our views.

zachmullen:
> backbone/event-scope: Most of the time the context is not required for the event handlers and always requiring it goes a little overboard in my opinion. Although, I did get a PR accepted to fix an issue with this rule.

This is actually an important rule; while that variable does serve as the context for the bound function, it also has another purpose that is actually important to us: the [off function](http://backbonejs.org/#Events-off) will allow you to pass a context argument that will unbind all such events that have that context. This is very helpful for managing object lifecycles when events are in play between multiple objects and one of them needs to be destroyed.

jbeezley:
Okay, I'll make an issue for `backbone/event-scope` as well.

zachmullen:
This LGTM

Issue #805, opened 2015-04-23
Modular javascript libs


zachmullen:
Rather than build everything into a single JS file (app.min.js), we should divide it up into more files so people can import useful subsets of javascript functionality. I would propose the following:
- `girder.core.min.js`: all the utilities (init, misc functions, etc. All others would depend on this)
- `girder.models.min.js`: models (in case someone wants to build their own widgets
- `girder.widgets.min.js`: stuff inside views/widgets, meant for embedding inside other custom apps
- `girder.app.min.js`: everything else (the main app, layout stuff, and other views). Most custom apps would not need to import this file at all.

zachmullen:
And while we are breaking js import compatibility, I would rename `libs.min.js` to `girder.ext.min.js`.

jeffbaumes:
+1: rename `libs.min.js` to `girder.ext.min.js`.

ronichoudhury:
+7

cjh1:
@zachmullen Would the files that depend on girder.core.min.js load it automatically or would the user have to include it manually?

zachmullen:
@cjh1 It might make sense to investigate using require.js or something similar when we do this. We could also preserve backward compatibility by adding an additional grunt step that built stuff under the old names, though I would remove such a task from the default task graph.

manthey:
I like the idea of using require.js or similar so that when I want both girder and another package that both include the same external library, I only get it once.

cjh1:
I have looked a requires.js a couple of times and every time was scared off by the changes that would be required to use it in an existing application, but I may have missed something. I would be interesting in other peoples opinions ...

jeffbaumes:
I assume we would use browserify with require. I saw it being used with good success in a BioJS tutorial.

cpatrick:
+1 to something like require.js and +1 to changing the name. I'm a little nervous about how this affects our version... We need to make sure all users of JS things know of the change.

jbeezley:
It will probably require a significant change to do right.  This is what scared me away when I last looked at it.  I think you need to wrap all of your js modules into functions that look like this

```javascript
require(['dep1', 'dep2'], function (dep1, dep2) {
});
```

It might be worth it to look into ES6 shims that perform the same function if you are going to do this.

zachmullen:
Yeah, it's a sizeable undertaking; I'm not sure how it would play with our minification since we have a disconnect between modules and files.

zachmullen:
[...]

PR #1161, opened 2016-01-11
WIP: Move external dependencies into ext.min.js and ext.min.css

### LINKED ISSUE #1160: Create a libs.min.css
There are a number of CSS files that could be concatenated during build into a single ``libs.min.css`` file.

As part of this, the un-minified ``/lib/fontello/css/fontello.css`` and ``/lib/fontello/css/animation.css`` should be switched to use a minified version from upstream, or put through Girder's CSS minifier.

brianhelba:
* Make some docs use paragraph-level markup for warnings
* Move Bootstrap CSS into ``ext.min.css``
*  Rename ``libs.min.js`` to ``ext.min.js``
 * This makes it more clear that this file contains external Javascript dependencies, and is consistant with the file ``ext.min.css``.
* Make ext.min.js available under it's old name (libs.min.js)

Fixes #1160.


grunt_tasks/build.js
```diff
@@ -77,17 +77,11 @@ module.exports = function (grunt) {
...
-                    cwd: 'node_modules/bootstrap/dist/css',
-                    src: ['bootstrap.min.css'],
-                    dest: 'clients/web/static/built'
-                }, {
-                    expand: true,
-                    cwd: 'node_modules/bootstrap-switch/dist/css/bootstrap3',
-                    src: ['bootstrap-switch.min.css'],
-                    dest: 'clients/web/static/built'
+                    src: ['clients/web/static/built/ext.min.js'],
+                    dest: 'clients/web/static/built/libs.min.js'
```
zachmullen: I think we should just symlink this rather than copy it. I plan to do this for bootstrap and d3 as well when we move them.

brianhelba: I assume you mean Fontello + D3? Bootstrap has already been moved without leaving compatibility files behind.

Symlinks are a good idea, but that's just going to further break Windows; is this acceptable? I don't care either way.

zachmullen: We don't support windows for the server. I believe modern windows has symlinks with at least some level of support, plus windows users could just update to the new path.

manthey: Windows has supported symlinks (and hardlinks, too) since XP.  The command
mklink has existed since Vista.

On Mon, Jan 11, 2016 at 3:44 PM, Zach Mullen <notifications@github.com>
wrote:

> In grunt_tasks/build.js
> <https://github.com/girder/girder/pull/1161#discussion_r49375928>:
>
> >                  files: [{
> > -                    expand: true,
> > -                    cwd: 'node_modules/bootstrap/dist/css',
> > -                    src: ['bootstrap.min.css'],
> > -                    dest: 'clients/web/static/built'
> > -                }, {
> > -                    expand: true,
> > -                    cwd: 'node_modules/bootstrap-switch/dist/css/bootstrap3',
> > -                    src: ['bootstrap-switch.min.css'],
> > -                    dest: 'clients/web/static/built'
> > +                    src: ['clients/web/static/built/ext.min.js'],
> > +                    dest: 'clients/web/static/built/libs.min.js'
>
> We don't support windows for the server. I believe modern windows has
> symlinks with at least some level of support, plus windows users could just
> update to the new path.
>
> —
> <https://github.com/girder/girder/pull/1161/files#r49375928>.
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

brianhelba:
Two issues are outstanding:
* Stylus is failing with ``expected "indent", got ";"`` when it hits the CSS line
```css
filter: progid:DXImageTransform.Microsoft.gradient(enabled = false);
```
We could always use the pre-minified CSS from inside ``node_modules`` for this and all the other libraries, then just make the target concatenate them without running Stylus. If we try to run Stylus on some, and concatenate the others in afterwards, it's going to be too messy, IMO.
* Change the the file copy to a symlink operation.
  * Note, this will add the ``grunt-contrib-symlink`` dependency to ``package.json``, which is going to cause merge conflicts with changes in https://github.com/girder/girder/pull/1159

brianhelba:
I've fixed the Stylus error and changed file copy to a symlink, but now the symlinks are causing packaging test failures. I'll investigate in a bit.

brianhelba:
Ping @girder/developers . Can anyone help with these test failures?

The issue is that the packaging tests are failing with ``error: can't copy 'clients/web/static/lib/bootstrap/css/bootstrap-switch.min.css': doesn't exist or not a regular file`` when they try to operate on a file symlink.

For some reason, having a symlink to a directory apparently works fine (https://github.com/girder/girder/pull/1162 passed tests), but the file symlink fails. However, we need to have individual file symlinks, as the fix in https://github.com/girder/girder/pull/1162 (or any set of directory symlinks in general) isn't adequate to provide compatibility to all the files that have been removed as a result of this PR and https://github.com/girder/girder/pull/1146 .

zachmullen:
We just need to pass `preserve_symlinks=True` [here](https://github.com/girder/girder/blob/master/setup.py#L37).

clients/web/static/lib/bootstrap/css
```diff
@@ -1 +0,0 @@
-../../built/
```
zachmullen: Why are we removing this? @kotfic just put in a fix to preserve compatibility with the old path.

grunt_tasks/build.js
```diff
@@ -103,12 +90,24 @@ module.exports = function (grunt) {
...
+            ext_css: {
                 files: {
                     'clients/web/static/built/ext.min.css': [
-                        'node_modules/eonasdan-bootstrap-datetimepicker/build/css/bootstrap-datetimepicker.css',
-                        'node_modules/jsoneditor/dist/jsoneditor.css'
+                        'node_modules/bootstrap/dist/css/bootstrap.min.css',
+                        'node_modules/bootstrap-switch/dist/css/bootstrap3/bootstrap-switch.min.css',
+                        'node_modules/eonasdan-bootstrap-datetimepicker/build/css/bootstrap-datetimepicker.min.css',
+                        'node_modules/jsoneditor/dist/jsoneditor.min.css',
+                        'node_modules/as-jqplot/dist/jquery.jqplot.min.css'
```
zachmullen: I like this change. Every package under the sun ships their own minified css it seems, so this should be fine.

clients/web/static/lib/bootstrap/css
zachmullen: Nevermind, just realized why. So, in the end, this will in fact break any app that hasn't included `ext.min.css`, but that sort of needs to be included anyway at this point.

brianhelba: I changed things to create symlinks to the two moved files during the build process at https://github.com/girder/girder/pull/1161/files#diff-97825ec1d67dac5aa1eb3375f2dd68e4R174 . This way has a few benefits:
* All the compatibility symlinks are in the same place
* Since it's part of the build process, it doesn't clutter the repository with extra directories until build time
* Since individual files are linked, instead of a whole directory, it's more precise and prevents people from abusing the link
* Since it's part of the build process, in theory it might be easier to selectively turn off

zachmullen: Ah, great solution. I will test this out to make sure everything works.

grunt_tasks/build.js
brianhelba: We could change it to just concatenate the minified 3rd party JS too, but I doubt it will make a difference (except perhaps to make builds marginally faster), so I left it with the status quo.

The only other argument I can think of for leaving it the same is that perhaps some of the 3rd party libraries used a minifier that produced inferior output, and thus Uglify can do a better job, but this is a bit of a stretch.

I'll change it if you can think of a good reason (even just consistency).

zachmullen: Let's leave it for now since it isn't causing any issues, and it has the nice benefit of giving us consistent source maps.

clients/web/static/lib/bootstrap/css
brianhelba: @zachmullen What will it break? I've included symlinks (to the new file, or equivalent) for all the removed public-facing static files.

brianhelba:
@girder/developers PTAL. Ready for review.

zachmullen:
LGTM, though I'd like @kotfic to try it out and make sure it works with his app and give the final signoff.

brianhelba:
@kotfic @girder/developers PTAL again, added one last commit to address issues raised in https://github.com/girder/girder/issues/805 .

zachmullen:
LGTM.

jcfr:
:+1: Thanks again for working on this.

PR #1169, opened 2016-01-12
Timeline widget tweaks

msmolens:
This cleans up a couple issues in `TimelineWidget.js`:
- The `defaultClass` option was not used. This change adds separate default class options for segments and points and uses those options in the implementation.
- Remove nonexistent variable passed to template.


zachmullen:
I like this change, however let's retain the old property for compatibility. We can try to use the point or segment specific default first, and only drop back to the `defaultClass` if those aren't set.

jcfr:
Nice work :+1:

PR #1171, opened 2016-01-14
Fix CMake add_web_client_test flags

danlamanna:
@zachmullen 
The major fix here is `ENABLEDPLUGINS` now functions properly. 

@mgrauer This means that [this line](https://github.com/Kitware/minerva/blob/0e27615ed32c342bd048152a0a6899c2457c2d04/plugin.cmake#L68) in Minerva is mostly useless, only the gravatar plugin is actually loaded during client side testing (since it's first), this should fix that.

Also, I made `TIMEOUT` a regular argument instead of a multival argument (seemed like a bug).

Maybe a 2.0 note for consistency:
`ENABLEDPLUGINS` in CMake corresponds to `ENABLED_PLUGINS` in python
Same goes for `WEBSECURITY` -> `WEB_SECURITY`

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -223,11 +223,12 @@ and then run in your build directory ::
 
 before running your tests.
 
-A client side test can specify several options, including which plugins should be loaded at the time of the test, for example ::
+A client side test can specify several options, for example ::
 
-    add_web_client_test(some_client_test "someSpec.js" ENABLEDPLUGINS "gravatar" "jobs")
+    add_web_client_test(some_client_test "someSpec.js" PLUGIN "my_plugin" ENABLEDPLUGINS "my_plugin" "gravatar" "jobs")
 
-would add a test which ensured the gravatar and jobs plugins were loaded when run.
+The ``PLUGIN`` argument indicates that "my_plugin" is the owner of ``some_client_test``, while ``ENABLEDPLUGINS`` determines which plugins are going to be loaded when running
+``some_client_test``. In this case my_plugin, gravatar, jobs, and all of their dependencies will be loaded in the proper order at the time of this test.
 
 .. note:: Core functionality shouldn't depend on plugins being enabled, this test definition is more suitable for a plugin. Information for testing plugins can be found under :doc:`plugin-development`.
 
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -223,6 +223,14 @@ and then run in your build directory ::
 
 before running your tests.
 
+A client side test can specify several options, including which plugins should be loaded at the time of the test, for example ::
+
+    add_web_client_test(some_client_test "someSpec.js" ENABLEDPLUGINS "gravatar" "jobs")
+
+would add a test which ensured the gravatar and jobs plugins were loaded when run.
+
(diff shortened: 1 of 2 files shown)


danlamanna:
@manthey - @mgrauer and I were discussing this and it seems as though if a single argument is passed to `ENABLEDPLUGINS` then Girder will resolve dependencies appropriately, meaning we could really just go based off of the `PLUGIN` parameter and deprecate `ENABLEDPLUGINS` entirely. Unless we're missing something, care to chime in?

mgrauer:
@girder/developers

zachmullen:
There actually is a semantic difference between the PLUGIN and ENABLEDPLUGINS params. PLUGIN is used to declare which plugin is actually being tested, i.e. which plugin *owns* the test in question. In most cases web client tests will really only need to load that plugin, which will automatically load any of its explicit upstreams. ENABLEDPLUGINS is used to declare any additional plugins besides the one under test that should be enabled. So keeping them separate, and perhaps documenting this more explicitly, would be my preference.

manthey:
I don't see ENABLEDPLUGINS being used anywhere on the main branch, even historically.  I have a vague belief that I added it in 2014 to facilitate testing an external plugin before we added dependency resolution.  I'd say remove it.

zachmullen:
I think we should keep it in case someone wants to test their plugin in the presence of another plugin that is not a strict requirement.

jcfr:
> to test their plugin in the presence of another plugin that is not a strict requirement.

In that case, should the other plugin be a development dependency ?

Is there way to specify requirement on a per plugin basis ?

zachmullen:
> In that case, should the other plugin be a development dependency ?

We don't have the notion of development dependencies between plugins. The situation I'm imagining is more like a plugin that can optionally augment another plugin, but that other plugin is not a hard requirement, and in some (but not all) test scenarios, the other plugin should be loaded to test that interaction.

> Is there way to specify requirement on a per plugin basis ?

What do you mean? In girder, each plugin lists its own dependencies on other plugins, not sure if that is what you were asking though.

jcfr:
Thanks. Your explanation clarified everything.

docs/development.rst
```diff
@@ -223,6 +223,14 @@ and then run in your build directory ::
 
 before running your tests.
 
+A client side test can specify several options, including which plugins should be loaded at the time of the test, for example ::
+
+    add_web_client_test(some_client_test "someSpec.js" ENABLEDPLUGINS "gravatar" "jobs")
```
zachmullen: We should change these docs to mention both `PLUGIN` and `ENABLEDPLUGINS` and the behavior of each.

danlamanna: @zachmullen Let me know if the updated docs are good.

zachmullen:
LGTM

mgrauer:
@danlamanna Could you add a clarification to the comment in JavascriptTests.cmake to explicitly state that using `PLUGIN` will load all dependent plugins?  I couldn't find a mention of this in the cmake or the docs.  

    # PLUGIN (name of plugin) : this plugin is loaded (unless overridden with

I made a similar request in the docs.

danlamanna:
@mgrauer Take a look :)

mgrauer:
Great, thanks.  LGTM.

PR #1175, opened 2016-01-15
Return full user documents for group list endpoints

zachmullen:
There isn't a good reason to artificially limit them. This makes it much easier to, say, write a client script to scrape the emails of all members of a group.

### DIFF
--- a/girder/models/group.py
+++ b/girder/models/group.py
@@ -137,12 +137,11 @@ class Group(AccessControlledModel):
 
     def listMembers(self, group, offset=0, limit=0, sort=None):
         """
-        List members of the group, with names, ids, and logins.
+        List members of the group.
         """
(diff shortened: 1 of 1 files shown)


brianhelba:
This seems like a fine change on the model layer, but how does it impact the API, w.r.t. two things:
* an email address is a semi-private piece of data; it should not be available to untrusted other users
* plugins may store very private data in extra fields on the user document; is it going to be adequately filtered by default (I believe Girder's policy is always one of private-by default for fields)?

zachmullen:
The documents are still subject to filtering as usual, so email or any other non-whitelisted fields won't appear to anyone without admin access.

zachmullen:
@brianhelba are you OK with this PR?

brianhelba:
I've verified all of the calling locations have a ``@filtermodel``. LGTM.

PR #1180, opened 2016-01-17
Allow accordion collapse widgets to be clicked anywhere

brianhelba:
* Allow accordion collapse widgets to be clicked anywhere
 * The Bootstrap examples don't provide the optimal placement of the ``data-toggle`` property. It should be placed on the ``.panel-heading`` div, not the ``a`` element inside the ``.panel-title``.
* Convert some usage of Jade pipes into spans or block text continuations
 * In addition to being more maintainable, this actually slightly improves the spacing of the compiled HTML, making precise text selection in a browser easier.
 * Also, correct the spelling of "MongoDB".
* Remove inline nested elements from Jade templates
 * This syntax is less clear and more difficult to maintain.

### DIFF
--- a/clients/web/src/templates/widgets/groupInviteDialog.jade
+++ b/clients/web/src/templates/widgets/groupInviteDialog.jade
@@ -5,9 +5,9 @@
       h4.modal-title Invite user to group
       .g-dialog-subtitle
         i.icon-user
-        | #{user.text}
+        span #{user.text}
     .modal-body
-      label.control-label(for="g-role") Select a role for this user
+      label.control-label Select a role for this user
       .panel-group#g-invite-role-container
         .panel.panel-default
           .panel-heading(data-toggle="collapse",
@@ -23,11 +23,11 @@
                 group, but may not change properties of the group.
               .g-invite-as-member.btn.btn-primary
                 i.icon-mail
-                |  Invite as member
+                span Invite as member
               if mayAdd
                 .g-add-as-member.btn.btn-warning.g-add-user-button
                   i.icon-plus
-                  | Add as member
+                  span Add as member
 
         if (level >= accessType.ADMIN)
           .panel.panel-default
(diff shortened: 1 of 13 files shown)


brianhelba:
@girder/developers I doubt this is the sort of thing we'd want to write a test for (because maybe there are legitimate exceptions), but we should remember to put the ``data-toggle`` on the ``.panel-heading`` in all future accordion collapse widgets.

zachmullen:
What happens in this case for other elements embedded in panel headings that are clickable but not meant to toggle panel openness?

brianhelba:
@zachmullen I'm not sure, but it seems easy enough to try out if we're interested. If the parent ``.panel-heading`` really does steal (or duplicate) all click events from a child element that has a higher ``z-index`` and its own click handler, then perhaps that's one of the "legitimate exceptions".

At the moment though, we don't have any such embedded elements in Girder, do we?

zachmullen:
We do have some [in covalic](https://github.com/girder/covalic/blob/master/web_external/templates/body/scoringMetrics.jade#L33-L38).

brianhelba:
@zachmullen So.....DTPRLGTY?

clients/web/src/templates/widgets/groupInviteDialog.jade
```diff
@@ -39,17 +43,19 @@
...
+                    span Add as moderator
 
         if (level >= accessType.ADMIN)
           .panel.panel-default
-            .panel-heading: .panel-title
-              a(data-toggle="collapse", data-parent="#g-invite-role-container",
-                href="#g-invite-role-admin") Invite as administrator
+            .panel-heading(data-toggle="collapse",
+                           data-parent="#g-invite-role-container",
+                           data-target="#g-invite-role-moderator")
```
zachmullen: This should open the admin tab rather than the moderator tab.

zachmullen:
Other than that one bug, this LGTM.

clients/web/src/templates/widgets/groupInviteDialog.jade
brianhelba: Yikes, sorry.

brianhelba:
Fixed that typo and rebased.

PR #1181, opened 2016-01-17
Change all Python strings to use old-style formatting

brianhelba:
Without this, Unicode user names are causing 500 errors when ``/api/v1/group/<id>/access`` is called.

### DIFF
--- a/girder/utility/mail_utils.py
+++ b/girder/utility/mail_utils.py
@@ -41,7 +41,7 @@ def getEmailUrlPrefix():
         host = '://'.join((cherrypy.request.scheme,
                            cherrypy.request.local.name))
         if cherrypy.request.local.port != 80:
-            host += ':%s' % cherrypy.request.local.port
+            host += ':%d' % cherrypy.request.local.port
 
     return host
 
--- a/girder/utility/mail_utils.py
+++ b/girder/utility/mail_utils.py
@@ -41,7 +41,7 @@ def getEmailUrlPrefix():
         host = '://'.join((cherrypy.request.scheme,
                            cherrypy.request.local.name))
         if cherrypy.request.local.port != 80:
-            host += ':{}'.format(cherrypy.request.local.port)
+            host += ':%s' % cherrypy.request.local.port
 
     return host
 
--- a/plugins/geospatial/server/geospatial.py
+++ b/plugins/geospatial/server/geospatial.py
@@ -91,9 +91,9 @@ class GeospatialItem(Resource):
                     raise RestException('Property names must be at least one'
                                         ' character long.')
                 if '.' in key or key[0] == '$':
-                    raise RestException('The property name {} must not contain'
-                                        ' a period or begin with a dollar sign.'
-                                        .format(key))
+                    raise RestException('The property name %s must not contain'
+                                        ' a period or begin with a dollar'
+                                        ' sign.' % key)
 
             data.append({'name': name,
                          'description': description,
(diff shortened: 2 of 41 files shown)


zachmullen:
Thanks for fixing. I would actually consider this comment unnecessary since we really need to stop using `format` everywhere, since it's the only string building mechanism that doesn't do coercion to Unicode automatically.

zachmullen:
(This is already [documented](http://girder.readthedocs.org/en/latest/development.html#use-instead-of-format), BTW).

zachmullen:
@brianhelba Could you please rebase this so I can sign off on it?

girder/utility/mail_utils.py
```diff
@@ -41,7 +41,7 @@ def getEmailUrlPrefix():
         host = '://'.join((cherrypy.request.scheme,
                            cherrypy.request.local.name))
         if cherrypy.request.local.port != 80:
-            host += ':{}'.format(cherrypy.request.local.port)
+            host += ':%s' % cherrypy.request.local.port
```
zachmullen: `%d`?

plugins/geospatial/server/geospatial.py
```diff
@@ -263,15 +263,15 @@ def near(self, params):
                         raise ValueError
 
                 except ValueError:
-                    raise RestException("Parameter '{}' must be a number."
-                                        .format(param))
+                    raise RestException("Parameter '%s' must be a number." %
+                                        param)
 
-                condition['$'+param] = distance
+                condition['$%s' % param] = distance
```
zachmullen: `+` actually works fine in cases like this, it coerces unicode gracefully with the same behavior as `%`.

plugins/jobs/plugin_tests/jobs_test.py
```diff
@@ -43,7 +43,7 @@ def setUp(self):
         base.TestCase.setUp(self)
 
         self.users = [self.model('user').createUser(
-            'usr' + str(n), 'passwd', 'tst', 'usr', 'u{}@u.com'.format(n))
+            'usr' + str(n), 'passwd', 'tst', 'usr', 'u%s@u.com' % n)
```
zachmullen: `%d`

girder/utility/mail_utils.py
brianhelba: Fixed, though it shouldn't matter unless you want to do more advanced string formatting inside ``%d``, right?

plugins/jobs/plugin_tests/jobs_test.py
brianhelba: Fixed.

plugins/geospatial/server/geospatial.py
brianhelba: I'm not sure why I even changed this. My regex probably picked up something nearby, and I spotted it. I don't really care what this is.

Since you brought it up though, you're right it's faster with concatenation for 2 strings, but when you get to longer, non-trivial arrangements, formatting becomes much more maintainable and eventually faster. My thinking is to make everything use formatting, so future developers are encouraged to keep it that way if additional variables are necessary.

brianhelba: I've removed it from the rebased commit.

brianhelba:
@zachmullen Rebased and addressed your comments.

plugins/geospatial/server/geospatial.py
zachmullen: That's fine, we can leave this change in.

zachmullen:
LGTM once travis is green.

PR #1182, opened 2016-01-17
Fix validation in "load" and "setMetadata" model methods

brianhelba:
* Fix some incorrect documentation in ``setMetadata`` model methods 
* Use more specific and consistent exceptions in ``Model.load``

### DIFF
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -170,9 +170,9 @@ class Folder(AccessControlledModel):
 
     def setMetadata(self, folder, metadata):
         """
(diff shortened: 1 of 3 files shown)


zachmullen:
I think we should instead raise a ValidationException so we aren't leaking pymongo details.

brianhelba:
Added a new commit:
* Raise a ``ValidationException`` on ``Model.save`` failures
  * The most likely cause of this is an invalid field name.

@zachmullen PTAL again.

zachmullen:
LGTM

PR #1186, opened 2016-01-18
Fix plugin topological ordering

zachmullen:
Transitive plugin dependencies were broken. This unifies the code
path for DAG traversal and toposorting of plugins between the
storage code in the setting model and the loading code, and now
properly supports transitive dependency ordering. This also dies
gracefully if cycles exist.

### DIFF
--- a/girder/models/setting.py
+++ b/girder/models/setting.py
@@ -22,7 +22,7 @@ import cherrypy
 import six
 
 from ..constants import SettingDefault
-from .model_base import Model, ValidationException, GirderException
+from .model_base import Model, ValidationException
 from girder.utility import camelcase, plugin_utilities
 from bson.objectid import ObjectId
 
--- a/girder/models/setting.py
+++ b/girder/models/setting.py
@@ -49,7 +49,7 @@ class Setting(Model):
             getattr(self, funcName)(doc)
         else:
             raise ValidationException(
-                'Invalid setting key "{}".'.format(key), 'key')
+                'Invalid setting key "%s".' % key, 'key')
 
         return doc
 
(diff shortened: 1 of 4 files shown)


zachmullen:
@mgrauer PTAL to make sure this fixes your issue. In my manual testing it worked for transitive dependencies where the old one failed. If this fix works for you, I'll add some testing for it.

zachmullen:
@mgrauer that last commit should fix the issue for you.

mgrauer:
This solves the issue for me.  It sorts the plugins list twice, but it isn't a big deal and it only happens at server startup.

Looking at `loadPlugins`, it has [this comment](https://github.com/girder/girder/blob/e7b8af984447af76690d8342e6f4b10728206844/girder/utility/plugin_utilities.py#L47-L49):

```
Loads a set of plugins into the application. The list passed in should not
already contain dependency information; dependent plugins will be loaded
automatically.
```

Can you rephrase this comment to explain what `should not
already contain dependency information` means?  Does passing in a topo sorted list violate this?

zachmullen:
@mgrauer PTAL, I updated the comments and also made it so we don't build the DAG and sort it twice on server startup.

mgrauer:
Great, LGTM.

danlamanna:
We might not necessarily care, but this leaves Girder unable to start if a plugin that is expected no longer exists on disk. The only way I can see it being done is by manually altering the setting with mongo.

jbeezley:
Well, I care a little.  What is it that prevents it from starting?

zachmullen:
I'm putting in a fix shortly :)

PR #1187, opened 2016-01-18
Fix plugin copy task

### LINKED ISSUE #1179: Plugin "extras" directory is broken
On the latest Girder, the [``plugin-<name>:copy`` Grunt task](https://github.com/girder/girder/blob/master/grunt_tasks/plugin.js#L122) is not working. Although this task will run, the ``plugins/<name>/extra`` directory (and it's contents) is not actually being copied to ``clients/web/static/built/plugins/<name>/extra``.

This will need to be fixed before the next release.

jbeezley:
Fixes #1179

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -58,6 +58,9 @@ module.exports = function (grunt) {
                 'plugin-install': {}
             };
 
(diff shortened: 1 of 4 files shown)


zachmullen:
LGTM, nice test.

PR #1188, opened 2016-01-19
Allow server to startup with missing plugins

zachmullen:
@jbeezley @danlamanna this fixes the issue Dan mentioned on the other topic about the server failing to startup if a required plugin is missing.


cpatrick:
lgtm.

PR #1192, opened 2016-01-21
Trigger g:logout on logout instead of g:login

danlamanna:
On login:
- `g:login` is triggered
- `g:login.success` or `g:login.error` is triggered

On logout:
- `g:login` is triggered
- `g:logout.success` or `g:logout.error` is triggered

Girder uses `g:logout` in 2 places even though it's never defined. It happens to be in places where `g:login` is bound to the same function - so it accidentally works :).

This is a judgement call for Girder developers, as it technically breaks backwards compatibility.

### DIFF
--- a/clients/web/src/init.js
+++ b/clients/web/src/init.js
@@ -202,8 +202,8 @@ _.extend(girder, {
             girder.currentUser = null;
             girder.currentToken = null;
 
-            girder.events.trigger('g:login', null);
             girder.events.trigger('g:logout.success');
+            girder.events.trigger('g:logout');
         }, function (jqxhr) {
             girder.events.trigger('g:logout.error', jqxhr.status, jqxhr);
         });


zachmullen:
This is breaking our officially sanctioned and documented method of listening for user change events. It's not a great name, but `g:login` is sent anytime the user changes, even if it changes to null. That makes it so views that need to update themselves whenever the user changes can bind to just one event.

danlamanna:
Alright, I'll close this out.

mgrauer:
@danlamanna I tried the same thing about 6 months back.  

Sounds like g:logout is on the wishlist for G2.

danlamanna:
I suppose the 2 instances of `g:logout` should be removed since they're never being triggered?

zachmullen:
Perhaps. I'm also fine with adding a `g:logout` that only gets triggered on logout, if people are only interested in that part.

cpatrick:
:+1: to adding `g:logout`.

jeffbaumes:
The lack of `g:logout`, while not intuitive, at least makes people dig until they find that `g:login` is triggered on login and logout. If we supply `g:logout`, I think people are more likely to create two functions: one that does login logic that `g:login` triggers, and one that does logout logic that `g:logout` triggers. Then they stand to get even more confused when their login function is called at logout.

My suggestion is to fix this in v2.0 with three events: `g:user-changed`, `g:user-login`, and `g:user-logout` and deprecate `g:login` (which would have the same behavior as `g:user-changed`).

zachmullen:
:+1: for @jeffbaumes name suggestions.

jeffbaumes:
I suppose we could add the three new events now and just wait for v2 to deprecate/remove the original `g:login`.

danlamanna:
We would certainly need to document the differences then between `g:user-changed` and `g:login-changed`.

jeffbaumes:
I was unaware of `g:login-changed` :smile: - yes that would be confusing. It seems that before suggesting reconfiguring of event names we need what @danlamanna suggested - full documentation of all events currently in the system #1193. Since this is a closed PR, I'll make an issue for revisiting login/logout events and point here.

Issue #420, opened 2014-10-21
Javascript documentation tool


zachmullen:
We should add something that builds or javascript docs automatically from code annotations. I've used [jsdoc](https://github.com/jsdoc3/jsdoc) for this in the past, but I recall it being rather crappy. If sphinx supports javascript, we should just use that.

jbeezley:
I've done more exploring since this issue was last raised.  There are a few newer developments among javascript documentation generators.

* [dox](https://github.com/tj/dox) reads jsdoc style comments and outputs raw json.  The idea is that you can design your own templates to render it how you like.  It is a very nice project that does a much better job of generating consistent json than the [haruki](https://github.com/jsdoc3/jsdoc/tree/master/templates/haruki) jsdoc template.  The downside is that it doesn't use jsdoc and thus doesn't have full support for all jsdoc tags.  If we wanted to go the route of integrating the client-side documentation into our sphinx docs, this would probably be the way to go, but we would have to write the templates ourselves.
* A number of templates for rendering the output from dox have popped up as well,  but they all suffer from the limitations of dox itself.  To name a few in no particular order:
  * [dox-foundation](https://github.com/punkave/dox-foundation)
  * [mr-doc](https://github.com/mr-doc/mr-doc)
  * [markdox](https://github.com/cbou/markdox)
  * [doxumentor](https://github.com/dawnerd/doxumentor)
  * [dox-template](https://github.com/olivernn/dox-template)
  * [doxmate](https://github.com/JacksonTian/doxmate) (if you read Chinese)
  * [dox-basic](https://github.com/jepso/dox-basic)
  * [dox-docco](https://github.com/aearly/dox-docco)
* [documentation](https://github.com/documentationjs/documentation) is another (newer) generator that uses jsdoc style comments.  This one seems to be largely developed by the people behind mapbox and is under very active development.  It is what they use to generate their own top-notch [developer documentation](https://www.mapbox.com/mapbox-gl-js/api/).  I tried using it briefly on some of our projects.  It will definitely require some work to add the correct tags so that the documentation displays nicely, but I do have high hopes for it.

zachmullen:
Thanks for looking into this! Since you are now the most informed about this subject, what tool would you recommend as the best one for us to move forward with? We don't have many constraints, since we don't really have anything at all currently.

jbeezley:
[...]

PR #1195, opened 2016-01-23
[WIP] Event catalog

### LINKED ISSUE #1193: Create catalog of built-in events
Would there be interest in this? 

It seems like something similar to http://backbonejs.org/#Events-catalog could be useful.

mgrauer:
Fixes #1193.

This isn't done yet, but it's a lot of work, so I put an intermediate commit up now.  I'll continue on in this direction until I hear otherwise.


brianhelba:
I'm renaming the PR to add a ``[WIP]`` prefix.

mgrauer:
Thanks @brianhelba , I didn't put it on initially because I'd prefer feedback earlier.

brianhelba:
@mgrauer Ah, sorry. I'll let you control the naming.

mgrauer:
NP, it's kind of weird case.

danlamanna:
@mgrauer Thanks for doing this, I was actually just thinking of taking a stab at it so I'm glad I checked first :P.

Does now seem like a good time to add very rudimentary jsdoc tags? (re @jbeezley in #420)

Since it will be a lengthy process, simply adding [@fires](http://usejsdoc.org/tags-fires.html) annotations might be a decent start to building up our JS documentation. I can go through and add them in this PR if it seems appropriate.

mgrauer:
Sounds good to me @danlamanna .  I should be pushing another commit in the next hour.

mgrauer:
Ready for review, or handoff to @danlamanna

clients/web/src/views/widgets/MetadataWidget.js
```diff
@@ -53,7 +53,7 @@ girder.views.MetadataWidget = girder.View.extend({
     initialize: function (settings) {
         this.item = settings.item;
         this.accessLevel = settings.accessLevel;
-        this.item.on('g:changed', function () {
+        this.item.on('g:fetched', function () {
```
zachmullen: Why this change?

zachmullen:
I think we need to organize these by what prototype/object triggers the event. This is mostly a documentation of certain aspects of various prototypes, and in the special case of `girder.events`, a centralized object for global events.

clients/web/src/views/widgets/MetadataWidget.js
mgrauer: Those are item models, which as far as I can tell, never trigger `g:changed` (collections do though).  Models trigger `g:fetched`, and that seems to be what those listeners are intended to do.

jeffbaumes:
Much better than nothing - good start! Although I agree @zachmullen they need to be organized by the object that creates the event. It would make sense for these to be attached to the jsdoc for each type of object.

jbeezley:
This is where having jsdoc infrastructure in place would be helpful.  If you add the [`@fires`](http://usejsdoc.org/tags-fires.html) annotation to the classes, you will get that grouping automatically in the documentation for each class.

zachmullen:
Yep, that's what I would really love to have.

mgrauer:
@danlamanna I'm not going to have any more time on this--would you want to take it over?

danlamanna:
@mgrauer Yup I'll take a look

zachmullen:
@mgrauer @danlamanna mind if I close this PR for cleanliness (unless you're actively working on it)? I'll leave the branch around.

mgrauer:
no problem.  closed.

PR #1196, opened 2016-01-23
Only install what we need from clients dir

zachmullen:
The symlink in the test dir was breaking our `python setup.py install` command, which was also breaking our RTD build. This branch fixes that as a side effect; the real change here is to clean up our sdist to only include what we need from the clients dir, which is just the prod-required content within the web client dir.


cpatrick:
lgtm.

zachmullen:
Thanks, that [fixed RTD](https://readthedocs.org/projects/girder/builds/3664528/).

PR #1197, opened 2016-01-23
Scoped access decorators

zachmullen:
This branch has been lingering for a long while, I think it would be good to have @girder/developers look at it. This creates the initial infrastructure necessary for us to scope our tokens to allow third-party systems limited authenticated access to the server. This is the first step toward allowing girder to act as an OAuth provider.

The discussion of exactly what scopes should exist and what endpoints should correspond to what scopes is up for debate; I selected a minimal set that I think is sensible to start with. Anything that doesn't fall under these scopes is currently only accessible in the old way, i.e. normal authenticated user access. My hope is that we can slowly open up more and more endpoints to scoped access.

### DIFF
--- a/girder/api/access.py
+++ b/girder/api/access.py
@@ -76,11 +76,11 @@ def user(*args, **kwargs):
             @six.wraps(fun)
             def wrapped(*iargs, **ikwargs):
                 token = rest.getCurrentToken()
-                if (not _tokenModel.hasScope(token, kwargs.get('scope')) and
-                        not rest.getCurrentUser()):
+                if not rest.getCurrentUser():
                     raise AccessException('You must be logged in.')
                 return fun(*iargs, **ikwargs)
             wrapped.accessLevel = 'user'
+            wrapped.requiredScopes = kwargs.get('scope')
             return wrapped
         return dec
 
--- a/girder/api/access.py
+++ b/girder/api/access.py
@@ -141,26 +141,16 @@ def cookie(*args, **kwargs):
 
     While allowing cookie authentication on other types of routes exposes an
     application to Cross-Site Request Forgery (CSRF) attacks, an optional
-    ``force=True`` argument may be passed to the decorator to make it effective
+    ``force=True`` kwarg may be passed to the decorator to make it effective
     on any type of route.
     """
     if len(args) == 1 and callable(args[0]):  # Used as a raw decorator
         force = False
         args[0].cookieAuth = (True, force)
         return args[0]
-
     else:  # Used with arguments
-        if len(args) == 1 and isinstance(args[0], bool):
-            force = args[0]
-        elif 'force' in kwargs:
-            force = kwargs['force']
-        elif (not args) and (not kwargs):
-            force = False
-        else:
-            raise TypeError('Unsupported extra decorator arguments.')
-
         def decorator(fun):
-            fun.cookieAuth = (True, force)
+            fun.cookieAuth = (True, kwargs.get('force', False))
             return fun
 
         return decorator
--- a/girder/api/access.py
+++ b/girder/api/access.py
@@ -46,11 +46,10 @@ def admin(*args, **kwargs):
         def dec(fun):
             @six.wraps(fun)
             def wrapped(*iargs, **ikwargs):
-                token = rest.getCurrentToken()
-                if not _tokenModel.hasScope(token, kwargs.get('scope')):
-                    rest.requireAdmin(rest.getCurrentUser())
+                rest.requireAdmin(rest.getCurrentUser())
                 return fun(*iargs, **ikwargs)
             wrapped.accessLevel = 'admin'
+            wrapped.requiredScopes = kwargs.get('scope')
             return wrapped
         return dec
 
--- a/girder/api/access.py
+++ b/girder/api/access.py
@@ -19,37 +19,102 @@
 
 import six
 
-from girder.models.model_base import AccessException
 from girder.api import rest
+from girder.models.model_base import AccessException
+from girder.utility.model_importer import ModelImporter
+
+_tokenModel = ModelImporter.model('token')
 
 
-def admin(fun):
+def admin(*args, **kwargs):
     """
     Functions that require administrator access should be wrapped in this
     decorator.
+
+    :param scope: To also expose this endpoint for certain token scopes,
+        pass those scopes here. If multiple are passed, all will be required.
+    :type scope: str or list of str
     """
-    @six.wraps(fun)
-    def accessDecorator(*args, **kwargs):
-        rest.requireAdmin(rest.getCurrentUser())
-        return fun(*args, **kwargs)
-    accessDecorator.accessLevel = 'admin'
-    return accessDecorator
+    if len(args) == 1 and callable(args[0]):  # Raw decorator
+        @six.wraps(args[0])
+        def wrapped(*iargs, **ikwargs):
+            rest.requireAdmin(rest.getCurrentUser())
+            return args[0](*iargs, **ikwargs)
+        wrapped.accessLevel = 'admin'
+        return wrapped
+    else:  # We should return a decorator
+        def dec(fun):
+            @six.wraps(fun)
+            def wrapped(*iargs, **ikwargs):
+                token = rest.getCurrentToken()
+                if not _tokenModel.hasScope(token, kwargs.get('scope')):
+                    rest.requireAdmin(rest.getCurrentUser())
+                return fun(*iargs, **ikwargs)
+            wrapped.accessLevel = 'admin'
+            return wrapped
+        return dec
 
 
-def user(fun):
+def user(*args, **kwargs):
     """
-    Functions that allow any user (not just administrators) should be wrapped
-    in this decorator. That is, a token must be passed that has the
-    "core.user_auth" scope and a valid user ID.
+    Functions that require a logged-in user should be wrapped with this access
+    decorator.
+
+    :param scope: To also expose this endpoint for certain token scopes,
+        pass those scopes here. If multiple are passed, all will be required.
+    :type scope: str or list of str
     """
-    @six.wraps(fun)
-    def accessDecorator(*args, **kwargs):
-        user = rest.getCurrentUser()
-        if not user:
-            raise AccessException('You must be logged in.')
-        return fun(*args, **kwargs)
-    accessDecorator.accessLevel = 'user'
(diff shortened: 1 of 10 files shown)


girder/api/access.py
```diff
@@ -17,66 +17,118 @@
...
+        wrapped.accessLevel = 'admin'
+        return wrapped
+    else:  # We should return a decorator
+        def dec(fun):
+            @six.wraps(fun)
+            def wrapped(*iargs, **ikwargs):
+                rest.requireAdmin(rest.getCurrentUser())
+                return fun(*iargs, **ikwargs)
+            wrapped.accessLevel = 'admin'
+            wrapped.requiredScopes = kwargs.get('scope')
```
brianhelba: The ``cookie`` decorator allows either keyword or positional arguments for extra options. These new decorators expect extra options to only be keyword arguments. We should make this behavior consistent one way or another.

tests/cases/access_test.py
```diff
@@ -89,6 +96,34 @@ def cookieHandler(self, **kwargs):
     def cookieForceHandler(self, **kwargs):
         return
 
+    @access.admin
+    def rawAdmin(self, **kwargs):
```
brianhelba: This is redundant with ``adminHandler``. Same for ``userHandler``/``rawUser`` and ``publicHandler``/``rawPublic``.

girder/api/access.py
zachmullen: My first thought was to make these support positional or keyword, but thinking about it more I actually like the simplicity and explicitness of only supporting them as kwargs... are you OK with that?

zachmullen:
@brianhelba I think I addressed all the concerns, is this good to merge now?

jeffbaumes:
+1 for this (+1 !== LGTM). It will be great to fully support being an OAuth provider in the future.

girder/api/rest.py
```diff
@@ -170,7 +170,8 @@ def retVal(user, token):
         return retVal(None, token)
     else:
         try:
-            ensureTokenScopes(token, TokenScope.USER_AUTH)
+            ensureTokenScopes(token, getattr(
+                cherrypy.request, 'requiredScopes', TokenScope.USER_AUTH))
```
brianhelba: Do we need the fallback of `TokenScope.USER_AUTH` here? Every time `handleRoute` is called, `cherrypy.request.requiredScopes` should be getting set to something.

girder/api/access.py
```diff
@@ -89,26 +141,16 @@ def cookie(*args, **kwargs):
...
-        elif 'force' in kwargs:
-            force = kwargs['force']
-        elif (not args) and (not kwargs):
-            force = False
-        else:
-            raise TypeError('Unsupported extra decorator arguments.')
-
         def decorator(fun):
-            fun.cookieAuth = (True, force)
+            fun.cookieAuth = (True, kwargs.get('force', False))
```
brianhelba: Strictly speaking, this simplification of `cookie` is breaking some compatibility. I couldn't care less myself, but I have a duty to bring it up.

girder/api/rest.py
```diff
@@ -473,6 +478,9 @@ def ensureTokenScopes(token, scope):
     :type scope: str or list of str
     """
     tokenModel = ModelImporter.model('token')
+    if tokenModel.hasScope(token, TokenScope.USER_AUTH):
```
brianhelba: I take it this is for backwards compatibility? `TokenScope.USER_AUTH` is effectively a super-scope, that always provides permission?

girder/api/v1/system.py
```diff
@@ -108,7 +108,18 @@ def setSetting(self, params):
         .errorResponse('You are not a system administrator.', 403)
         .errorResponse('Failed to set system setting.', 500))
 
-    @access.admin
+    @access.admin(scope=TokenScope.SETTINGS_READ)
```
brianhelba: Just to clarify, the behavior here (and the other settings endpoints remains the same):
* A token must be for a user with >= the access level AND the correct scope to access the endpoint
* Ordinary user tokens still always have `TokenScope.USER_AUTH`
* `TokenScope.USER_AUTH` is a super-scope, so every ordinary user token satisfies this new `scope=TokenScope.SETTINGS_READ` requirement

Then at some point in the future, we'll introduce tokens which don't have `TokenScope.USER_AUTH`, but do have`TokenScope.SETTINGS_READ`, so this addition will become relevant then. However, those tokens still have to satisfy the admin user requirement, so I don't totally understand what their use case would be.

Let me know if I got this right.

girder/constants.py
```diff
@@ -195,6 +195,12 @@ class TokenScope:
     ANONYMOUS_SESSION = 'core.anonymous_session'
     USER_AUTH = 'core.user_auth'
     TEMPORARY_USER_AUTH = 'core.user_auth.temporary'
+    PLUGINS_ENABLED_READ = 'core.plugins.read'
```
brianhelba: Where'd this list come from? Administrative-ish tasks that we'd eventually want to delegate to a robot (partial upload clean)? that we'd eventually want non-admins to do (read settings)?

brianhelba:
@zachmullen Review complete. Please respond at your convenience.

girder/api/rest.py
zachmullen: USER_AUTH tokens are ones used by actual logged in users, as opposed to tokens being issued on behalf of the user to third-party clients. Users are allowed to do anything whilst logged in as themselves, but third party applications can be granted only specific scopes. So yes, it is a super-scope, but is used to distinguish between normal session tokens and OAuth tokens.

girder/api/v1/system.py
zachmullen: That's correct. 

> However, those tokens still have to satisfy the admin user requirement, so I don't totally understand what their use case would be.

The case here would be an admin user granting a third-party app the right to read system settings. That third-party app would then have permission that is normally reserved for admin user, but not *all* permissions granted to admins, just the ability to read system settings. This is why it's easiest (though not strictly required) to try and divide endpoints cleanly (read: unconditionally) into scopes.

girder/constants.py
zachmullen: Like I said in an earlier post, this list is currently put forth as a straw man. The actual decisions about which scopes should exist should be debated by the community at large.

girder/api/access.py
zachmullen: I think I made this change in response to your comment on Feb. 4, and I'm still OK with this. It does technically break compatibility, but I think it's a good idea to force people to use kwargs with these complicated functions that can be either decorators or not. @girder/developers please weigh in.

girder/api/rest.py
zachmullen: Just defense in depth, I don't think it's doing any harm.

girder/api/access.py
brianhelba: I'd much rather let this one slide, but again, I have a duty to bring it up. @.cpatrick should be so proud of me.

girder/api/v1/system.py
brianhelba: Got it.

girder/api/rest.py
brianhelba: Got it.

girder/constants.py
brianhelba: Well, you know how it is, as soon as you commit something, we're stuck supporting it forever, even if it was a bad idea. However, if you're comfortable with this list as a base (I assume you've put some thought into it), that's fine with me.

zachmullen: The burden of compatibility might not be quite so bad for these scopes. We shouldn't remove symbols under `TokenScope` once they're in master in case plugins are using them, but I think we should be OK with the idea of changing core and plugin endpoint scope requirements; the burden on users would simply be to log-out and reauthorize the app, which isn't so bad.

zachmullen: I say it's not so bad because this happens all the time with mobile apps, you'll notice they frequently change what scopes they require, and you simply have to reauthorize them whenever that happens.

girder/api/rest.py
brianhelba: I understand, but I'd argue that if `cherrypy.request.requiredScopes` is not set, something has gone horribly wrong and we *should* fail. Errors are not a bad thing, if you make a system that doesn't throw them, you end up with the mess that is HTML parsing. I worry that this sort of thing allows other bugs to fester, because they don't just throw an error quickly and loudly, they do some unintended thing silently. IMHO, defence in depth is more about gracefully handling errors from detached subsystems that we are willing to tolerate failure from.

I'd say that we decide "`cherrypy.request.requiredScopes` must now always be set by the time it reaches Girder's handlers", and then enforce that rule. Unless that's not a sensible rule, in which case nevermind everything. :wink:

brianhelba:
Once we resolve the question about `cherrypy.request.requiredScopes`, this LGTM. Obviously it requires a rebase.

girder/constants.py
zachmullen: Of course, that capability depends on a feature that doesn't yet exist: applications should be able to do a lookup to the server in which they send a list of all the API calls they wish to make and are sent back a set of required scope identifiers. If the apps do this lookup prior to sending the user through the login flow, they'll never lose compatibility.

brianhelba: Good points. I can tolerate some deprecated variable identifiers lingering in our internal code for the sake of plugins.

girder/api/access.py
cpatrick: 1. I'm also inclined to let this one slide, but it would be better to leave compatibility and deprecate--breaking the API later.
2. I'm very proud :smile_cat:.

brianhelba: Thanks dad.

girder/api/rest.py
zachmullen: Good point, I'm OK with this if the alternative is an error, I just wanted to be sure the alternative wasn't continuing on without validating scopes at all.

zachmullen:
> the question about cherrypy.request.requiredScopes

Having thought more about this, I think I'm going to leave it as is; the `getCurrentUser` function could, theoretically, be called in some context in which `cherrypy.request.requiredScopes` had not been previously set, if someone chooses to hook into the REST utilities at a lower level than we normally do. Since we haven't documented that `getCurrentUser` depends on the requiredScopes being set prior to invocation, I don't think we should make that a requirement now, and I think requiring a user auth token is a sensible default in such a case.

brianhelba:
@zachmullen Your reasoning is sound.

PR #1198, opened 2016-01-24
Proposal: change web client build process

zachmullen:
This is the first proposed v2.0 PR, so I don't expect it to be merged for quite some time, but I had talked about this last week as the first step toward a better, more abstracted web client build process. The main change here is simply that our web client build process changes from:

    npm install

to

    npm install && npm run build

While the decomposition into two separate steps isn't strictly necessary, it makes the process a bit cleaner and more standardized. With this change, `npm install` installs all required core npm packages, and then recursively installs any packages listed in the `package.json` of any plugin directories. It no longer performs the build. This change moves some of our plugin building logic out of grunt and into raw node scripts, which will make our build code more portable if/when we decide to use something other than grunt. `npm run build` is now the officially sanctioned command for building the web client assets.

I'd like us to also consider whether the `npm install` step should also install any bower requirements from plugins.

One other small API change here is that `grunt watch` will be officially supplanted by `npm run watch`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ before_install:
 install:
     - pip install -r requirements.txt -r requirements-dev.txt -e .
     - npm install
+    - npm run build
 script:
     - mkdir _build
     - cd _build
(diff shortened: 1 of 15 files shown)


zachmullen:
Closing this to reopen against the 2.0-integration branch.

PR #1207, opened 2016-01-25
add "autoconf" option to plugin.json

opadron:
Allows plugins to explicitly opt-out of the default grunt tasks created for each plugin.
Good idea? Bad idea? Got a better name for the option? Discuss.

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -185,13 +185,13 @@ module.exports = function (grunt) {
             }
 
             var doAutoConfig = (
-                typeof config.grunt !== "object" ||
-                config.grunt.autoconf === (void 0) ||
-                config.grunt.autoconf === (null) ||
+               !_.isObject(config.grunt) ||
+                _.isUndefined(config.grunt.autoconf) ||
+                _.isNull(config.grunt.autoconf) ||
               !!config.grunt.autoconf
             );
 
-            if(doAutoConfig) {
+            if (doAutoConfig) {
                 // merge in configuration for the main plugin build tasks
                 configurePlugin(plugin);
             }
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -177,14 +177,23 @@ module.exports = function (grunt) {
                 'Found plugin: ' + plugin
             ).bold);
 
-            // merge in configuration for the main plugin build tasks
-            configurePlugin(plugin);
-
             if (fs.existsSync(json)) {
                 config = grunt.file.readYAML(json);
             }
             if (fs.existsSync(yml)) {
                 config = grunt.file.readYAML(yml);
+            }
+
+            var doAutoConfig = (
+                typeof config.grunt !== "object" ||
+                config.grunt.autoconf === (void 0) ||
+                config.grunt.autoconf === (null) ||
+              !!config.grunt.autoconf
+            );
+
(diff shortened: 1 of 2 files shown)


grunt_tasks/plugin.js
```diff
@@ -187,6 +184,18 @@ module.exports = function (grunt) {
                 config = grunt.file.readYAML(yml);
             }
 
+            var doAutoConfig = (
+                typeof config.grunt !== "object" ||
+                config.grunt.autoconf === (void 0) ||
```
jbeezley: Just out of my own morbid curiosity, why `void 0`?

jbeezley:
I would be in favor of such a change.

zachmullen:
I like this feature, though it's failing style tests: http://my.cdash.org/testDetails.php?test=22954533&build=900004

jbeezley:
Yes with new stricter rules, this is going to be common.  I actually considered adding `npm run lint` as a pre-commit hook in my other branch.

grunt_tasks/plugin.js
opadron: Just a habit of mine from when you used to be able to redefine `undefined`.  I just checked on chrome and node.js and it looks like it's not necessary, anymore?

zachmullen:
BTW, thanks for adding that lint npm script. I want to add more scripts behind npm to start abstracting away our web tools that are not npm/node. (See https://github.com/girder/girder/pull/1198)

grunt_tasks/plugin.js
jbeezley: Your experience in javascript clear predates mine.  I've never heard of such a thing.

zachmullen: Back in the olden days it was always `typeof x === "undefined";`. Considering javascript has about 6,000 reserved words, it's pretty amazing that `undefined` isn't one of them, but an expression that returns it (`void`) is.

¯\\_(ツ)_/¯

opadron:
I'm not sure what to do about the remaining errors.

zachmullen:
Travis has been exceptionally flaky lately, though many of the intermittent test failures are actual race conditions in our code showing up.

opadron:
@girder/developers PTAL

zachmullen:
Let's add some documentation to plugin-development.rst, we have [some stuff in there already](http://girder.readthedocs.org/en/latest/plugin-development.html#executing-custom-grunt-build-steps-for-your-plugin) about the grunt config options.

zachmullen:
Somehow I didn't see your last commit, so ignore my comment. LGTM

mgrauer:
Sorry, didn't get a chance to review this earlier.

My concern here is that `autoconf` is an existing build system (and we are doing this work in a build system) and that could be grounds for confusion.  Could we change this key to `autoconfig` to provide slight separation?

Further, "autoconf" or even "autoconfig" is a bit terse to explain what it is doing.  Since this key (when true) means "allow Girder to build the plugin using the standard Grunt tools" something to make that clear, like "girder-grunt-plugin" or "girder-standard-plugin-grunt" or "standardConfig" or better ideas?

opadron:
@mgrauer I share your reservations.

Looks like I can't revert without creating a new PR, so I just went ahead and added #1211.  Let's move the discussion there.

mgrauer:
Thanks @opadron

PR #1208, opened 2016-01-27
girder_worker server side plugin

zachmullen:


zachmullen:
This is a much more minimal version of the server side code than currently exists in romanesco; this strips out all of the stuff that is flow-specific, i.e. the parts that let users run arbitrary code.

jeffbaumes:
Makes sense to have a minimal core. Do you have a plan on where the flow-specific stuff should go? Seems this should eventually be a Minerva-level capability but for now perhaps it goes into flow itself.

zachmullen:
I was thinking either flow (now that it's a proper girder plugin) or minvera as well. I want to talk with @mgrauer about it a bit.

zachmullen:
At long last, the plugin side of this is actually tested. This is now ready for review.

zachmullen:
@girder/developers can someone QA this?

jeffbaumes:
This is great that this is tested now! One thing - could we call it something other than Batch Worker in the docs - maybe Remote Worker? According to Google:

> A batch job is a computer program or set of programs processed in batch mode. This means that a sequence of commands to be executed by the operating system is listed in a file (often called a batch file, command file, or shell script) and submitted for execution as a single unit.

Since the worker can be scripting but can also include Docker and other execution, I'm not sure "batch" is the right word here.

jeffbaumes:
One other question: is there any code left that was in Romanesco but does not yet live in the `girder_worker` or `girder` repos? If Flow were to start using this, would it have all it needs?

zachmullen:
The endpoints in this plugin that allowed user-defined script execution have been removed, and to my knowledge they have not been implemented in flow yet. I know @danlamanna had been working a while ago on making flow a girder plugin, but I'm not sure the status on that.

zachmullen:
@jeffbaumes gave his LGTM via an alternate channel.

PR #1210, opened 2016-01-27
Serve our own license badge

zachmullen:
Two separate services have now failed us for serving up this badge.
It's only 749 bytes and it's ASCII... we'll do it ourselves.

/rant


jbeezley:
That doesn't appear to be working either: https://github.com/girder/girder/blob/serve-own-license-badge/README.rst

zachmullen:
I think it won't until we merge it, since the rst builder is changing the image path

jbeezley:
Well, it can't be any worse than the current badge.  LGTM.

brianhelba:
LGTM.

PR #1211, opened 2016-01-27
change autoconf option name to autobuild

opadron:
Supplants #1207.  I've had reservations with my initial choice of `autoconf` for the name of the option, and it looks like I'm not alone.

This PR renames the option to autobuild to avoid confusion with the `autoconf` build tool.

I'll let this PR hang for a few days in case other folks want to chime in with other name suggestions.

### DIFF
--- a/docs/plugin-development.rst
+++ b/docs/plugin-development.rst
@@ -534,7 +534,7 @@ key to your **plugin.json** file.
         {
         "file" : "Gruntfile.js",
         "defaultTargets": [ "MY_PLUGIN_TASK" ],
-        "autoconf": true
+        "autobuild": true
         }
     }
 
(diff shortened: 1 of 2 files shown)


danlamanna:
I think `grunt.autobuild` and `grunt.autoconf` are a bit vague/confusing. To me, `grunt.girderBaseTasks` as a plugin setting might be more explicit.

opadron:
More alternatives:

 - `inheritGirderBaseTasks`
 - `skipGirderBaseTasks` (default: `false`)
 - `inheritGirderTasks`
 - `skipGirderTasks` (default: `false`)
 - `inheritTasks`
 - `skipTasks` (default: `false`)

zachmullen:
I think we should go ahead and merge this before the old name stays in the wild any longer. autobuild is an improvement since it's not the name of an existing piece of configuration software.

mgrauer:
I can live with that.

PR #1212, opened 2016-01-27
Use a PNG instead of SVG for license badge

zachmullen:
We need a format that can be auto-detected via magic number.


zachmullen:
@jbeezley this works even in the preview, good to merge?

jbeezley:
LGTM.

PR #1213, opened 2016-01-27
Attempt to fix sporadic web client test failures

zachmullen:
Trying to fix an intermittent failure in the web_client_routing test on travis that I cannot reproduce locally. This one test accounts for the majority of our random travis failures.


zachmullen:
@girder/developers PTAL. I'm not certain that this actually fixes all the race conditions, but I just ran the head of this topic through travis for [16 builds](http://my.cdash.org/index.php?project=girder&date=2016-01-28); there was only 1 test failure during that time, which was an OOM killing.

cpatrick:
LGTM.

PR #1214, opened 2016-01-27
Provide an interface for testing with external data files

jbeezley:
Fixes https://github.com/DigitalSlideArchive/large_image/issues/11.

It should be fairly self explanatory with the tests and documentation that I added.  A few notes on the implementation:
* The wrapping of the `ExternalData` module happens in a separate module in case we want to provide similar functionality in other test environments (such as the client tests).
* The implementation in `TestData.cmake` is currently pretty basic, but as I mentioned in the comments it is designed to be extendable while providing backward compatibility.
* I choose to force plugins to namespace their own files (as in  `plugins/my-plugin/bigfile.txt`) so to make it possible for one plugin to access the data from another plugin.  It also makes the implementation quite a bit easier.


docs/plugin-development.rst
```diff
@@ -389,6 +389,18 @@ You can use all of the testing utilities provided by the ``base.TestCase`` class
...
 the main Girder coverage results if coverage is enabled.
 
+Plugins can also use the external data interface provided by Girder as described
+in :ref:`use_external_data`.  For plugins, the data key files should be placed
+inside a directory called ``plugin_tests/data/``.  When referencing the
+files, they must be prefixed by your plugin name as follows
+
+.. code-block:: cmake
+
+    add_python_test(my_test EXTERNAL_DATA plugins/cats/test_file.txt)
```
brianhelba: I question whether the fake ``plugins/<name>`` path is straining the limits of magical paths.

The ``add_python_test`` CMake function already takes a ``PLUGIN`` parameter. Why can't you just set the the data key file root path to ``plugins/<name>plugin_tests/data/`` whenever ``PLUGIN`` is set, then allow the ``EXTERNAL_DATA`` argument to just specify a straight file / directory name relative to that?

brianhelba: So to partially answer my own question, I suppose plugins may want to reference data key files from core. However, you could just have plugin data key file loader:
1. if this is a ``PLUGIN``, first search for the specified data key file in the ``plugins/<name>/plugin_tests/data/`` directory.
2. If not found, or not a ``PLUGIN``, search for the specified data key file in the ``tests/data/`` directory.

jbeezley: There can only be a single root path for locating the key files.  If we want to override that behavior, we will need to fork the `ExternalData` module locally.

tests/data/test_file.txt.md5
```diff
@@ -0,0 +1 @@
+169293f7c9138e4b50ebcab4358dc509
```
brianhelba: Does the current Midas3 support a more secure hash algorithm than MD5? If SHA, etc. is available, the tests should just always use that, since people will likely follow whatever the testing does.

To elaborate, while I agree with @cpatrick that there's a lot of FUD about MD5 floating around, using MD5 to verify external binary data (which may have been initially provided by untrusted third parties) is specifically what it's weak to.

tests/TestData.cmake
```diff
@@ -0,0 +1,109 @@
...
+
+# These are wrappers around ExternalData functions that have access to the
+# locally scoped variables such as the mirror list and the data source
+# root directory.
+#   See https://cmake.org/cmake/help/v3.3/module/ExternalData.html#module-functions
+function(girder_add_test)
+  ExternalData_add_test(${ARGN})
+endfunction()
+
+function(girder_add_target)
```
brianhelba: Are these the most appropriate names, since they're in the global CMake namespace? I could see some future functionality that adds a custom target in a more general way.

docs/plugin-development.rst
brianhelba: How do you feel about doing some more work in the [``file(COPY ${_base_data_files}``](https://github.com/girder/girder/pull/1214/files#diff-2b950cd7aa99f78420ee28f59a825688R139) step and/or the [``list(APPEND _data_files ``](https://github.com/girder/girder/pull/1214/files#diff-2b950cd7aa99f78420ee28f59a825688R139) to (only if the test is for a ``PLUGIN``) attempt to search / glob for files in the ``/tests/data/`` vs ``/tests/data/plugins/<name>/`` and add the appropriate prefix to each data key file?

jbeezley: That is possible, but it really begins to negate the advantage of using `ExternalData` in the first place.  If we are going to go through the file list and search for key files on our own, we are reimplementing a sizable portion of the original module.  It would also ruin much of the extensibility that I tried to bake in.  That is, you can still use functions like `girder_add_test` as you would use `ExternalData_add_test`.  The more magic we try to bake in the more we will break from the standard module and reimplement for ourselves.

jbeezley: Another reason, I'm building a new directory to store the key files is because *theoretically* the plugin directory doesn't have to be a subdirectory of girder.  In that case, inferring a relative path string (as required by `ExternalData`) to the plugin's data directory is difficult (or perhaps impossible).

tests/data/test_file.txt.md5
jbeezley: I tried to use sha1, but midas responded with "The bitstream does not exist or you do not have the permission".  I didn't see anywhere obvious what hashing functions midas supports.

cpatrick: Midas currently supports MD5 only.

tests/TestData.cmake
jbeezley: Any preferences on how to namespace them?  How about `girder_ExternalData_add_target`?

jbeezley:
@brianhelba I renamed the functions and rebased on master.

docs/developer-cookbook.rst
```diff
@@ -400,6 +400,44 @@ run.
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In some cases, it is necessary to perform a test on a file that is too big store
+inside a repository.  For tests such as these, Girder provides a way to link to
+test files served at `https://midas3.kitware.com`_ and have them automatically
+downloaded and cached during the build stage.  To add a new external file, first
+make an account at `https://midas3.kitware.com`_ and upload a publically accessible
+file.  When viewing the items containing those files on Midas, there will be a
+link to "Download key file" appearing as a key icon.  This file contains the
+the MD5 hash of the file contents and can be committed inside the
```
zachmullen: the the

docs/developer-cookbook.rst
```diff
@@ -400,6 +400,44 @@ run.
...
+``tests/data/`` directory of Girder's repository.  This file can then be
+listed as an optional ``EXTERNAL_DATA`` arguement to the ``add_python_test``
```
zachmullen: argument

tests/data/test_file.txt.md5
zachmullen: > using MD5 to verify external binary data (which may have been initially provided by untrusted third parties) is specifically what it's weak to.

Not sure what you mean here. What weakness are you referring to that would not be present with a different hashing algorithm?

zachmullen:
It feels wrong to me that we are pointing to midas3.kitware.com in our girder docs as a place to host data. This might put a spanner in the works of this effort, but could we instead create a very small plugin to deploy onto data.kitware.com that exposes stuff via its SHA512? That instance uses a filesystem assetstore, and those file documents are already indexed by the `sha512` field, so it would be a very simple matter of exposing a new endpoint that runs `self.model('file').find({'sha512': value})` and streams out the result if it exists and the user has access. Additionally, such a plugin could be made to support other assetstore adapters if it ever became necessary in the future by hooking into `data.process` and actually checksumming any file that was in an assetstore adapter that didn't contain that field already.

I could devote some of my own time toward this end if it's agreeable to people...

brianhelba:
@zachmullen Yes, that's what @vovythevov is working on now (with my support), in https://github.com/girder/girder/issues/961 and https://github.com/girder/girder/pull/1226 .

In the short term, we can use midas3.kitware.com , but as soon as @vovythevov finishes, I expect that we will move all references and data to data.kitware.com . Note, I've already created [a collection there to handle medically-oriented test data](https://data.kitware.com/#collection/568d85f28d777f429eac908a).

This is a chicken-and-egg problem, so I'd like to merge this PR first, since it's ready.

zachmullen:
Ok, if that is the plan then LGTM.

zachmullen:
(After fixing those typos, that is)

PR #1218, opened 2016-02-02
Test run restrictions

zachmullen:
This exposes two new CMake options, `RUN_CORE_TESTS`, a boolean option that can disable the core tests, and the related `TEST_PLUGINS` variable, a list that declares which plugins should be tested, or left empty to test them all. This makes it easier to setup dashboards and CI for external plugins that make use of girder's test infrastructure.


zachmullen:
@cpatrick PTAL, I'd like to convert some of our external plugins to use this for their CI.

cpatrick:
LGTMWTIH

PR #1219, opened 2016-02-02
Rework CMake plugin detection logic

zachmullen:
The "continue()" command was not added until CMake v3.2, and we
want to support older versions at this time.

ping @cpatrick


zachmullen:
@cpatrick gave this a LGTM in meatspace.

PR #1221, opened 2016-02-03
Make coverage thresholds configurable

zachmullen:
External plugins can now set their own values.

ping @cpatrick


cpatrick:
LGTM, but maybe put a comment saying you don't recommend lowering your standards :dancers:.

PR #1222, opened 2016-02-03
Docs about streaming/raw response were out of date

zachmullen:
@danlamanna your question about streaming responses led me to the docs and I realized they were out of date. PTAL.


danlamanna:
LGTM. Thanks again!

Issue #371, opened 2014-10-03
Swagger: update to 2.0 spec


jamiesnape:
Migration guide: https://github.com/wordnik/swagger-spec/wiki/Swagger-1.2-to-2.0-Migration-Guide

cjh1:
Looks like 2.1.1 now supports 2.0

brianhelba:
@girder/developers This didn't get auto-closed since it wasn't merged against master. I'm closing it now.

PR #1223, opened 2016-02-04
Update to Swagger Spec 2.0

msmolens:
The /describe endpoint now returns an API description that implements Swagger Spec 2.0:
https://github.com/OAI/OpenAPI-Specification/blob/0122c22e7fb93b571740dd3c6e141c65563a18be/versions/2.0.md

Previously, the /describe endpoint returned the Resource Listing and the
/describe/{resource} endpoint returned the API Declaration for a particular
resource. In Swagger Spec 2.0, the resource listing and API declaration are combined. Therefore, the /describe/{resource} endpoint is removed.

Also upgrades to swagger-ui 2.1.4 to support generating the front-end for a Swagger Spec 2.0 description.

Fixes https://github.com/girder/girder/issues/371

### DIFF
--- a/plugins/geospatial/server/geospatial.py
+++ b/plugins/geospatial/server/geospatial.py
@@ -125,13 +125,13 @@ class GeospatialItem(Resource):
         .param('geoJSON', 'A GeoJSON object containing the features or feature'
                           ' collection to add.', required=True,
                paramType='query')
-        .errorResponse()
-        .errorResponse('Invalid GeoJSON was passed in request body.')
-        .errorResponse('GeoJSON feature or feature collection was not passed in'
-                       ' request body.')
-        .errorResponse("GeoJSON feature did not contain a property named"
-                       " 'name'.")
-        .errorResponse('Property name was invalid.')
+        .errorResponse(("A parameter was invalid.",
+                        "Invalid GeoJSON was passed in request body.",
+                        "GeoJSON feature or feature collection was not passed "
+                        "in request body.",
+                        "GeoJSON feature did not contain a property named"
+                        " 'name'.",
+                        "Property name was invalid."))
         .errorResponse('Write access was denied on the parent folder.', 403)
         .notes("All GeoJSON features must contain a property named 'name' from"
                " which the name of each created item is taken."))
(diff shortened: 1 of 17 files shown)


plugins/geospatial/server/geospatial.py
```diff
@@ -492,10 +492,10 @@ def setGeospatial(self, item, params):
...
         .param('body', 'A JSON object containing the geospatial fields to add.',
                paramType='body')
-        .errorResponse('ID was invalid.')
-        .errorResponse('Invalid JSON was passed in request body.')
-        .errorResponse('Geospatial key name was invalid.')
-        .errorResponse('Geospatial field did not contain valid GeoJSON.')
+        .errorResponse('ID was invalid.\n\n'
+                       'Invalid JSON was passed in request body.\n\n'
+                       'Geospatial key name was invalid.\n\n'
+                       'Geospatial field did not contain valid GeoJSON.')
```
zachmullen: We could allow `errorResponse` to accept a list as the first parameter, and then inspect the type and call `'\n\n'.join(val)`.

zachmullen:
This is awesome, I had no idea you were working on this -- it's been on our wish-list for a really long time. :smile:

plugins/geospatial/server/geospatial.py
zachmullen: (I'm assuming the `\n\n` is a swagger detail, which would be best not to leak, but even if not, we are really trying to represent a list here, so it makes sense for our API to accept one and we can internally serialize it however we need to.)

zachmullen:
There is a small cosmetic issue with the front-end, a title is getting added in the page that wasn't there before:

screen shot 2016-02-04 at 1 21 37 pm

Hopefully there's a JS option to hide that, but if not we can hide it with CSS, it's got the `info_title` class.

zachmullen:
This is fantastic. My only question for @girder/developers is whether this is something we are considering a breaking change that should be labeled as targeting 2.0 integration branch rather than master. Technically, `/describe` and `/describe/:resource` are endpoints that clients other than our swagger page could use to programmatically inspect the API. This change rightly breaks compatibility of those endpoints.

cjh1:
I noticed that this change updated the description of lots of endpoints, are the same updates going to be need  for plugins?  Or is this just cleanup?

zachmullen:
@cjh1 it looks like everything will still work in plugins, however if you have multiple `errorResponse` on your endpoint for the same status code, you will now see warnings in the console. @msmolens correct me if that's wrong...

msmolens:
@zachmullen @cjh1 That's right. Later error responses with the same code override earlier ones and a warning is printed to the console.

msmolens:
It could be good to handle the legacy case and not discard earlier responses, but add them to a list instead. A warning message could still be appropriate.

cjh1:
Excellent, this should move us in the right direction for allow plugins to mount their endpoints under a prefix ...

zachmullen:
@msmolens given that v2.0 might be happening sooner than I initially anticipated, let's go ahead and close this and reopen it against the 2.0-integration branch.

PR #1225, opened 2016-02-04
Proposal: change web client build process

zachmullen:


zachmullen:
I've created the 2.0-integration branch and made this PR against it, so this is now ready for review and can be merged.

zachmullen:
@opadron PTAL

docs/installation.rst
```diff
@@ -115,7 +115,7 @@ to install the plugins as well.
 To build the client-side code project, cd into the root of the repository
 and run: ::
 
-    npm install
+    npm install && npm run build
 
 This will run multiple `Grunt <http://gruntjs.com>`_ tasks, to build all of
```
opadron: We're still using Grunt, right?  Either way, we should make sure to keep a close look at the docs as we make bigger changes.

grunt_tasks/package.js
```diff
@@ -38,8 +38,8 @@ module.exports = function (grunt) {
 
                         var fname = 'girder-' + grunt.config.get('pkg').version +
                             '.tar.gz';
-                        var stat = fs.statSync(fname);
-                        if (!err && stat.isFile() && stat.size > 0) {
+                        var stat = fs.existsSync(fname) && fs.statSync(fname);
+                        if (!err && stat && stat.isFile() && stat.size > 0) {
```
opadron: I'm not too familiar with this part.  Are these changes related to the build process?

zachmullen: This is kind of unrelated, it just fixes a bug in the case that packaging fails. Trying to `stat` a nonexistent file throws an exception.

grunt_tasks/plugin.js
```diff
@@ -246,25 +211,6 @@ module.exports = function (grunt) {
...
-            this.requiresConfig(tasks);
-
-            // queue the install tasks
-            grunt.config.get(tasks).forEach(function (task) {
-                grunt.task.run(task);
-            });
-        }
-    );
-
-    /**
```
opadron: Do these removals mean that plugins won't be able to use their own `packages.json` file, anymore?

zachmullen: No, this is now handled in the postinstall step by a new script.

docs/installation.rst
zachmullen: Yes, this branch is just changing the build commands, not the underlying technologies.

opadron:
LGTM

PR #1226, opened 2016-02-05
Add a utilty to serialize hash object states of several hash types

brianhelba:
This should help to implement https://github.com/girder/girder/issues/961 .


zachmullen:
While I am pretty certain nobody is using that utility outside this repo, we still should consider leaving the old symbols in place as aliases for the new ones, and then making a PR to remove the old file against the 2.0-integration branch.

brianhelba:
@zachmullen Done. Once this branch is merged, someone just needs to remember to add a 2.0-tagged PR which reverts https://github.com/girder/girder/commit/3c55e4c944618121d5e7aa32b1bfc77b80919f6e .

brianhelba:
This still needs testing. Additionally, given the amount of magic implementation details that this whole approach abuses (all of which will break as soon as someone isn't using cPython, complied against OpenSSL 1.0.x, on exactly an x86_64 platform), we need to directly unit test the functions, rather than just doing so implicitly as part of the FilesystemAssetStore code.

@vovythevov Care to add some direct unit testing? Note, you should to use data that is > the internal block size of the hashes (the ``block_size`` property of a hash instance, not the ``stateSize``) and that is a non-integer multiple of the block size.

brianhelba:
Also note, if we really need to, I could make this code support for 32-bit x86 platforms (currently it does not). If we really care about supporting more arbitrary platforms of either endianness, we *might* be able to start inspecting the actual OpenSSL ``structs`` by loading and reading the OpenSSL library header at runtime. 

Supporting other Python implantations, or cPython not linked against OpenSSL is also probably possible via a totally different route.

Of course, all of the aforementioned limitations existed before these improvements.

zachmullen:
Yeah, that's why I'm glad you've kept the `asserts` in there. We need to be able to know quickly and gracefully if the platform isn't going to support this, and as long as we have unit tests for each of these methods, it should be easy to test on whatever platforms we want.

zachmullen:
As far as supporting 32-bit and multiple endianness, I don't think we necessarily need to do that on this branch. If we eventually decide this is too dirty, we may need to think about rolling our own hashing implementation so we can enforce more guarantees than we can now.

brianhelba:
@zachmullen Yes, support for additional platforms / non-cPython implementations is way out the scope of this PR, but I just wanted to mention the idea. I don't think we need to re-implement the low-level hash algorithms, but we'd just do more poking around internal structures of other implementations (for instance, cPython has their own fallback implementation of hashes, in case OpenSSL is not linked against).

zachmullen:
The ideal thing would be a native object that has some public API methods for getting and setting internal hash state as a `bytes` object.

cpatrick:
It's pretty shocking how hard this is. People are going as far as forking hashlib to hand this sometimes: https://github.com/coreos/resumablehashlib.

cpatrick:
:octocat: buttons are hard.

zachmullen:
Yeah, it's quite surprising to find the lack of serializable hash objects, given that they are trivial to serialize.

brianhelba:
As sample data for hashing, how about using `from this import s`, where `s` is a ROT13 version of The Zen Of Python? It's 856 bytes long, which is about the perfect size. This way we can avoid committing more sample files with arbitrary data.

vovythevov:
@brianhelba: Would you mind taking a look at 915ef91 ?

tests/cases/hash_state_test.py
```diff
@@ -0,0 +1,68 @@
...
+    base.startServer()
+
+
+def tearDownModule():
+    base.stopServer()
+
+
+# Define a few variables used by all the tests:
+# Use the python's zen as 'long' string
+test_chunk = this.s.encode('utf8')
```
brianhelba: These should be moved to a ``def setUp(self):`` method inside the test class, which assigns any relevant data onto ``self``, where it can then be used by test methods.

The reason for doing this is to ensure the code runs at the time of the test, as opposed to whenever this file happens to be imported. It also ensures that if the data is ever modified by some test method, it gets freshly re-generated for the other test methods.

tests/cases/hash_state_test.py
```diff
@@ -0,0 +1,68 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+###############################################################################
+
+from .. import base
+from girder.utility import hash_state
+import hashlib
+import this
```
brianhelba: Since ``import this`` has side effects (printing to stdout), let's move it inside the new ``setUp`` method. Scope shouldn't be an issue, since it's only used to create ``test_chunk`` (which should be renamed to ``this.testChunk``).

tests/cases/hash_state_test.py
```diff
@@ -0,0 +1,68 @@
...
+class HashStateTestCase(base.TestCase):
+
+
+    def _simpleHashingTest(self, hash_type):
+        """
+        Test the hashing of hash_state against python's hashlib
+
+        :param hash_type: String representing the hash algorithm to use
+        """
+        official_hasher = hashlib.new(hash_type)
```
brianhelba: Girder's Python code uses camelCase for variable names. Let's change the underscore_names, to ensure consistent style.

tests/cases/hash_state_test.py
```diff
@@ -0,0 +1,68 @@
...
+        state = hash_state.serializeHex(hashlib.new(hash_type))
+
+        for chunk in chunks:
+            official_hasher.update(chunk)
+            checksum = hash_state.restoreHex(state, hash_type)
+            checksum.update(chunk)
+
+            self.assertEquals(official_hasher.hexdigest(), checksum.hexdigest())
```
brianhelba: There should be an additional check to ensure the hash states are equal for an empty hash, before any data gets added with ``update``.

tests/cases/hash_state_test.py
```diff
@@ -0,0 +1,68 @@
...
+            self.assertEquals(official_hasher.digest(), checksum.digest())
+
+            state = hash_state.serializeHex(checksum)
+
+
+    def testSimpleHashing(self):
+        """
+        Test all the base algorithms against the python's hashlib
```
brianhelba: Change ``the python's`` -> ``Python's`` here and in the docstring above. :wink:

brianhelba:
@vovythevov This looks really good so far, and seems to be fine in terms of functionality.

It needs a few style things fixed though. I've made most of my comments inline, but the style check test is also failing. See [the Travis output](https://travis-ci.org/girder/girder/jobs/108384756#L4561) for the issues that need to be fixed for the style test to pass.

If you want to set up Girder tests to run locally, [there's documentation on that](https://girder.readthedocs.org/en/latest/development.html#server-side-testing). Note you can also selectively run only some tests by using the ``-R`` option to ``ctest``, which allows you specify a regular expression for test names to run. You probably want to run ``ctest -R python_static_analysis_core_tests`` locally, since that's what's failing, and it's much faster to iteratively run on your machine than waiting 15 minutes for Travis to come back every time you make a change.

Personally, I will often run the style check and the new test that I'm working on locally, then just push the commits to GitHub. The full list of tests take a while to run, and I'd rather have Travis burn the CPU cycles than wait for my own machine to do it all.

brianhelba:
@vovythevov This iteration is improved, but see my inline comments and [the style tests](https://travis-ci.org/girder/girder/jobs/108624754#L4491) for the last of the issues that need to be fixed.

cpatrick:
CDash generally makes it easier to find the salient output http://my.cdash.org/testDetails.php?test=23296002&build=911302

brianhelba:
@cpatrick I knew that Girder's CI results are being sent to CDash, but I still don't know how to navigate directly to a given build's CDash summary. Is there a way beyond just guessing which of the listed "Continous" CDash builds corresponds to a given PR / Travis build?

cpatrick:
With Travis, mostly matching based on time.

brianhelba:
@girder/developers PTAL

zachmullen:
LGTM

PR #1227, opened 2016-02-05
Disable default error behavior on girder.login request

zachmullen:
Request failure on a login attempt is an error by the user, not
the system.

@cpatrick this meets your under 7 lines threshold, PTAL.


cpatrick:
How did we leave this here for so long...

LGTMITIH.

PR #1229, opened 2016-02-08
Add create user dialog to web client for admins

zachmullen:


zachmullen:
@cpatrick PTAL

cpatrick:
Nice test. LGTM.

PR #1230, opened 2016-02-09
Allow girder.EventStream to take arbitrary paths

danlamanna:
This allows Girder plugins to start their own event streams while still benefiting from the Backbone event triggering that Girder provides.

@aashish24 @kotfic This is what's being used for the Cumulus Minerva plugin (PR shortly) to stream log results.


kotfic:
+1

zachmullen:
Brilliant, LGTM.

PR #1231, opened 2016-02-10
Finally create a mkdir utility

zachmullen:
ping @cpatrick

### DIFF
--- a/girder/utility/__init__.py
+++ b/girder/utility/__init__.py
@@ -37,7 +37,7 @@ def camelcase(value):
                    re.split("[._]+", value))
 
 
-def mkdir(path, mode=0o777, recurse=True, safe=True):
+def mkdir(path, mode=0o777, recurse=True, existOk=True):
     """
     Create a new directory or ensure a directory already exists.
 
@@ -47,15 +47,15 @@ def mkdir(path, mode=0o777, recurse=True, safe=True):
     :type mode: int
     :param recurse: Whether intermediate missing dirs should be created.
     :type recurse: bool
-    :param safe: Set to True to suppress the error if the dir already exists.
-    :type safe: bool
+    :param existOk: Set to True to suppress the error if the dir exists.
+    :type existOk: bool
     """
     method = os.makedirs if recurse else os.mkdir
 
     try:
         method(path, mode)
     except OSError as exc:
-        if safe and exc.errno == errno.EEXIST and os.path.isdir(path):
+        if existOk and exc.errno == errno.EEXIST and os.path.isdir(path):
             pass
         else:
             raise
--- a/girder/utility/__init__.py
+++ b/girder/utility/__init__.py
@@ -18,7 +18,9 @@
 ###############################################################################
 
 import datetime
+import errno
 import json
+import os
 import pytz
 import re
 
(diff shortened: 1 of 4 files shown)


brianhelba:
Have you seen https://bugs.python.org/issue25583 ? This may cause issues on Windows (if we care).

zachmullen:
That problem doesn't bother me really, I don't think we ever try to create a directory that would be a drive name in Windows...

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -59,12 +59,13 @@ def validateInfo(doc):
...
-        if not os.path.isdir(doc['root']):
-            try:
-                os.makedirs(doc['root'])
-            except OSError:
-                raise ValidationException(
-                    'Could not make directory "%s".' % doc['root'])
+
+        try:
+            mkdir(doc['root'])
+        except Exception:
```
cpatrick: Why are we `except`ing two different ways around this?

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -85,14 +86,13 @@ def __init__(self, assetstore):
...
-            try:
-                os.makedirs(self.tempDir)
-            except OSError:
-                self.unavailable = True
-                logger.exception('Failed to create filesystem assetstore '
-                                 'directories %s' % self.tempDir)
-        elif not os.access(assetstore['root'], os.W_OK):
+        try:
+            mkdir(self.tempDir)
+        except OSError:
```
cpatrick: see above

girder/utility/plugin_utilities.py
```diff
@@ -250,15 +250,13 @@ def getPluginDirs(curConfig=None):
...
-                os.makedirs(pluginDir)
-            except OSError:
-                if not os.path.exists(pluginDir):
-                    print(TerminalColor.warning(
-                        'Could not create plugin directory %s.' % pluginDir))
-
-                    failedPluginDirs.add(pluginDir)
+        try:
+            mkdir(pluginDir)
+        except Exception:
```
cpatrick: see above

girder/utility/filesystem_assetstore_adapter.py
zachmullen: I think either one is fine, if we want them to be uniform which way would you prefer?

cpatrick: Ideally, we would `except` only the exceptions that can happen in those cases. I actually am not a big fan of `except Exception`.

zachmullen: That sounds good, I will make it `except OSError`.

zachmullen:
@cpatrick I changed the `except`s.

girder/utility/__init__.py
```diff
@@ -34,6 +37,30 @@ def camelcase(value):
...
+    """
+    Create a new directory or ensure a directory already exists.
+
+    :param path: The directory to create.
+    :type path: str
+    :param mode: The mode (permission mask) prior to applying umask.
+    :type mode: int
+    :param recurse: Whether intermediate missing dirs should be created.
+    :type recurse: bool
+    :param safe: Set to True to suppress the error if the dir already exists.
```
brianhelba: @zachmullen IMO, ``safe`` isn't a really clear name for a functionality which suppresses errors (I'd expect a safe mode to do more error checking).

[``os.makedirs`` in Python 3.2+ has the ``exist_ok`` parameter](https://docs.python.org/3/library/os.html#os.makedirs), which does the exact same thing as this parameter. How about just change the name to something like that?

brianhelba:
LGTM other than the parameter name issues I mentioned inline.

girder/utility/__init__.py
cpatrick: +1 to using exist_ok en lieu of safe.

Issue #1199, opened 2016-01-24
Make packaging tests run faster


zachmullen:
These are taking up a lot of our CI time. Some options:

* Make some of them mock out the npm calls so they don't all have to run through an entire install/build process
* Consider combining some of the plugin tests, such as packaging_plugin_install, packaging_extras_install, packaging_plugin_copy_extras so that they only need to actually do the npm install once for all three sets of assertions.

jbeezley:
I noticed this as well.  I might take a crack at mocking the npm calls.  Like I mentioned in #1189, there is a limit to what these packaging tests are even capable of detecting.  I think I might start running my nightly tests against an install from the generated packages.

zachmullen:
Mocking the npm calls should be simple, at least as I imagined, our wrapper test script would prepend something to that PATH that contained an executable shell script that just exits with 0.

jbeezley:
That's what I was thinking as well.

zachmullen:
It could also validate args or write something to disk that we could check and make sure it was actually called.

Issue #1141, opened 2015-12-15
Consider mocking npm during some of our tests


zachmullen:
Right now some of our packaging install tests are taking a great deal of time because they run npm. I think it's important for it to actually run npm at least one time, but we should consider mocking the executable for the rest of the invocations. This could cut ~2 minutes out of our CI response time probably.

PR #1234, opened 2016-02-12
Packaging test simplification

zachmullen:
Ran `time ctest -V -R packaging` on my macbook pro

Used to take 84s. With this change, takes 52s. On travis this will probably be an even bigger time saver.


zachmullen:
Fixes #1141 
Fixes #1199

jbeezley:
LGTM

PR #1236, opened 2016-02-12
Add model.file.finalizeUpload before and after event

vovythevov:
This event will be beneficial for future development as it's the only way
to get know the correspondence between a file and its upload.
Testing was added accordingly.

### DIFF
--- a/tests/cases/file_test.py
+++ b/tests/cases/file_test.py
@@ -27,6 +27,7 @@ import zipfile
 from hashlib import sha512
 from .. import base, mock_s3
 
+from girder import events
 from girder.constants import SettingKey
 from girder.models import getDbConnection
 from girder.models.model_base import AccessException
@@ -78,6 +79,14 @@ class FileTestCase(base.TestCase):
         }
         self.secondUser = self.model('user').createUser(**secondUser)
 
+        self.testForFinalizeUpload = False
+        self.finalizeUploadBeforeCalled = False
+        self.finalizeUploadAfterCalled = False
+        events.bind('model.file.finalizeUpload.before',
+                    '_testFinalizeUploadBefore', self._testFinalizeUploadBefore)
+        events.bind('model.file.finalizeUpload.after',
+                    '_testFinalizeUploadAfter', self._testFinalizeUploadAfter)
+
     def _testEmptyUpload(self, name):
         """
         Uploads an empty file to the server.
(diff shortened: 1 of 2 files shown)


brianhelba:
Let's hold off on this until https://github.com/girder/girder/pull/1239 is in.

tests/cases/file_test.py
```diff
@@ -100,10 +109,30 @@ def _testEmptyUpload(self, name):
...
+        self.finalizeUploadBeforeCalled = True
+        self._testFinalizeUpload(event)
+
+    def _testFinalizeUploadAfter(self, event):
+        self.finalizeUploadAfterCalled = True
+        self._testFinalizeUpload(event)
+
+    def _testFinalizeUpload(self, event):
+        self.assertTrue('file' in event.info)
+        self.assertTrue('upload' in event.info)
```
msmolens: Alternatively, these lines could use `self.assertIn()`, e.g. `self.assertIn('upload', event.info)`

zachmullen:
@vovythevov what is the status on this topic? Should we close the PR?

vovythevov:
@zachmullen: 
I haven't done any work on it since Feb 22nd. We were waiting on https://github.com/girder/girder/pull/1239 to be merged. It's mostly ready in my opinion, however I haven't addressed @msmolens 's comments.

How about I address the comments (within the next few days) and then we can merge it ?

zachmullen:
Sounds great, thanks :smile:

brianhelba:
Do we really need a before and after event? The original purpose of this feature was to enable non-SHA512 hashsums to be finalized for the hashsum_download plugin. For the sake of this plugin, either one will do.

I'd propose trimming this down to just the `after` event, since that gives us the fully saved `file` and `upload`. Any events that you'd normally have to do in the `before` event can be done just fine in `after`, just requiring perhaps 1 extra database save (which is de minimis for already-slow operations like file upload).

vovythevov:
@brianhelba: I only added the before event because it was easy and didn't seem to impact anything.
I agree with you that it might be superfluous. But it doesn't hurt to have it ?
@zachmullen: What's your opinion ?

zachmullen:
I don't have a strong opinion either way. I can't see a use for the `before` hook right now, but it's hard to try and think of all the possible ways someone might want to use the system. I'm fine with leaving it in.

vovythevov:
@zachmullen: Thanks for your input ! I added @msmolens 's comment so this should be ready to merge.

brianhelba:
Just to be clear, you should *in theory* be able to accomplish everything with just the `after` event, with the tradeoff that you might have to hit the database an extra time.

That said, while we could go overboard and proliferate new event hooks to many, many more places in our code, I think that uploading may be one area that plugin developers will want to customize the most, and it's not a speed-critical area, so checking for non-existent callbacks isn't a big deal.

TLDR: LGTM.

vovythevov:
@zachmullen: This is ready to be merged.

vovythevov:
Awesome ! Thanks @zachmullen and @brianhelba ! It's not always that old standing branches are actually picked and merged. Thanks for seeing this through !

zachmullen:
Thanks for implementing it! We try to keep the PR log as clean as possible :)

PR #1237, opened 2016-02-15
Download Fontello files at build time

### LINKED ISSUE #1164: Move Fontello to be built by Grunt
https://github.com/jubalm/grunt-fontello/pull/26 will have to be released before this will work, or we'll have to find another package to fetch and build Fontello via Grunt.

brianhelba:
* Download Fontello files at build time
* Add a compatibility symlink for Fontello files
* Remove the "clients/web/static/lib" directory entirely
 * This only holds compatibility symlinks for third-party libraries that used to live there.
* Update Fontello config to fix OAuth provider icons
 * This updates some unused icons for OAuth providers, which have the potential for misuse (as they are not the official branding) when additional providers are added in the future.

Fixes #1164


brianhelba:
@girder/developers PTAL.

zachmullen:
> * Add a compatibility symlink for Fontello files
* Remove the "clients/web/static/lib" directory entirely

@brianhelba where are the compatibility symlinks? Those two bullets seem contradictory...

brianhelba:
@zachmullen Sorry, to be more clear: the ``clients/web/static/lib`` directory no longer exists in the repo: the Fontello update removed the last of the real files in it, so the commit in question just [removes ``clients/web/static/lib/.gitignore``](https://github.com/girder/girder/commit/e1f6118d8443b82e62bcfbdfe58c49c88a4f6c61).

The build process will still create a "clients/web/static/lib" directory, which just contains an assortment of symlinks. These are all now ignored by [a new ``clients/web/static/.gitignore``](https://github.com/girder/girder/commit/e1f6118d8443b82e62bcfbdfe58c49c88a4f6c61).

zachmullen:
Ah, great, forgot about that build step.

brianhelba:
@zachmullen At some point, we could remove [the ``symlink::legacy_names`` step](https://github.com/girder/girder/blob/fontello-build/grunt_tasks/build.js#L178) from 2.0.

zachmullen:
We can do that now against the 2.0-integration branch.

zachmullen:
@brianhelba could you merge latest master onto this branch, and then add a snippet just below [here](https://github.com/girder/girder/blob/master/tests/packaging/pip_install_girder.sh#L27-L32) to make sure the built fontello files exist in the installed pip package?

zachmullen:
Sorry, I told you the wrong place to put that test, it should actually be after we run `girder-install web`.

brianhelba:
@zachmullen Yes, I figured that out. I've pushed an update.

zachmullen:
LGTMITIH

PR #1238, opened 2016-02-15
Add Python style tests to plugin tests

brianhelba:
* Add Python style tests to plugin tests
* Improve some test method documentation
* Fix style in hash_state_test
* Fix a bug in ``TestCase.tearDown``

### DIFF
--- a/plugins/hdfs_assetstore/plugin_tests/assetstore_test.py
+++ b/plugins/hdfs_assetstore/plugin_tests/assetstore_test.py
@@ -20,15 +20,20 @@
 import httmock
 import os
 import shutil
+import six
 import sys
 
 from girder.constants import AssetstoreType
 from tests import base
-from snakebite.client import Client
+from snakebite.client import Client  # noqa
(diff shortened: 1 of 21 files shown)


plugins/hdfs_assetstore/plugin_tests/assetstore_test.py
```diff
@@ -54,8 +55,8 @@ def serverdefaults(self):
 
     def df(self):
         return {
-            'used': 100L,
-            'capacity': 1000L
+            'used': 100,
```
brianhelba: @zachmullen This was causing the style checks to fail in Python 3. Since there is no ``long`` type or literals in Python 3, and since ``six`` does not have any support for ``long``, I changed these to ``int`` literals. 

If there is an important reason to make them ``long`` in Python2, we can add some extra logic.

brianhelba:
@girder/developers PTAL.

zachmullen:
LGTM, thanks for adding this.

PR #1239, opened 2016-02-15
Add a plugin to download files by their SHA512 hashsum

brianhelba:


brianhelba:
@vovythevov PTAL at the testing; it *should* be sufficient to finish implementing functionality for FilesystemAssetstore + SHA512 only.

plugins/hashsum_download/server/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        if algo not in self.SupportedAlgorithms:
+            msg = 'Invalid algorithm ("%s"). Supported algorithm are: %s.'\
+                  % (algo, self.SupportedAlgorithms)
+            raise RestException(msg, code=400)
+
+        query = {algo: hash.lower()}  # Always convert to lower case
+        fileModel = self.model('file')
+        file = fileModel.findOne(query)
+        if not file:
+            raise RestException('File not found.', code=404)
```
brianhelba: @girder/developers Note that when a file for an otherwise-valid hash value is not found, this endpoint returns a 404 code. This is in contrast to all other Girder endpoints, where the failure to find some ObjectId in the database returns a 400 error, even if the request is otherwise totally valid. I think that in general, the "right" thing is for database lookup failures to return a 404, but obviously changing this is way outside the scope of this PR. 

However, since this endpoint will likely be interacting with other third-party clients (e.g. CMake), I think it's very important to give more clear feedback on why a request failed. Also, since this is a plugin and it's doing a lookup by hash value and not ObjectId, I think we can be more tolerant of the inconsistency.

brianhelba:
@girder/developers PTAL.

plugins/hashsum_download/server/__init__.py
zachmullen: I think 404 is appropriate here. To me, the most important thing is that we return the exact same response in the following two cases:

1. No matching documents exists.
2. Matching documents exist, but current user does not have the required permission on any of them.

brianhelba: @zachmullen I don't want to sidetrack this too much, but I'm not totally clear on what you're asking for. Here's what we do now for every other route:
1. Unmatched route -> 404
2. Matched route, ObjectId syntax invalid -> 400
3. Matched route, ObjectId syntax valid, document missing from DB -> 400
4. Document found, anonymous user access denied -> 401
5. Document found, logged-in user access denied -> 403
6. Document found, logged-in user access denied -> 200

If your concern is that we are leaking information about the presence of absence of documents to unauthorized users, that's legitimate, but it's a deeper problem with the fact that [``AccessControlledModel.load``](https://github.com/girder/girder/blob/master/girder/models/model_base.py#L915) raises either a ``ValidationException`` on 2 or 3 or an ``AccessException`` (from [``AccessControlledModel.requireAccess``](https://github.com/girder/girder/blob/master/girder/models/model_base.py#L894); also this should be noted in docstring of ``AccessControlledModel.load``) on 4 or 5, then status codes are set in [``_handleRestException``](https://github.com/girder/girder/blob/master/girder/api/rest.py#L366) or [``_handleAccessException``](https://github.com/girder/girder/blob/master/girder/api/rest.py#L375).

So, all I'm proposing is that for this route, an unfound hash value (case 3, since we don't check hash syntax) returns 404. Everything else will be the same as the rest of the API (for better or worse).

zachmullen: I meant for this route in particular, given that it is a lookup by hash of contents. I'm fine with this behavior.

brianhelba: We're are indeed exposing even more information (effectively "does this file exist in this Girder instance") to unauthorized users. However the current functionality is such that cases 4 and 5 are the same as other API routes.

plugins/hashsum_download/plugin_tests/hashsum_download_test.py
```diff
@@ -0,0 +1,206 @@
...
+            user=self.user,
+            mimeType='image/png'
+        )
+
+        self.otherUser = self.model('user').createUser(
+            login='zorg',
+            password='mortis',
+            firstName='Jean-Baptiste',
+            lastName='Zorg',
+            email='nullus@universe.org'
```
zachmullen: Getting very creative with this dummy data...

plugins/hashsum_download/server/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+    SupportedAlgorithms = [
+        'sha512',
+    ]
+
+    def __init__(self, apiRoot):
+        super(File, self).__init__()
+
+        self.resourceName = 'file'
+        apiRoot.file.route('GET', ('hashsum', ':algo', ':hash', 'download'),
```
zachmullen: Should we also add `GET /file/hashum/:algo/:hash` to retrieve the file document? It's not necessary for our immediate needs, but should be easy to add.

plugins/hashsum_download/server/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+                           self.downloadWithHash)
+
+    @access.public
+    @describeRoute(
+        Description('Download a file by its hashsum.')
+        .param('algo', 'The type of the given hashsum.',
+               paramType='path', enum=SupportedAlgorithms)
+        .param('hash', 'The hashsum of the file to do wnload.',
+               paramType='path')
```
zachmullen: There's a space in the word download here. Also, would be nice to specify "in hexadecimal, case insensitive".

plugins/hashsum_download/server/__init__.py
brianhelba: I'd like to have that too, but I'd rather add it on a future iteration.

plugins/hashsum_download/server/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+        self.resourceName = 'file'
+        apiRoot.file.route('GET', ('hashsum', ':algo', ':hash', 'download'),
+                           self.downloadWithHash)
+
+    @access.public
+    @describeRoute(
+        Description('Download a file by its hashsum.')
+        .param('algo', 'The type of the given hashsum.',
+               paramType='path', enum=SupportedAlgorithms)
+        .param('hash', 'The hashsum of the file to download.',
```
zachmullen: My comment here got clobbered by a rebase... could we document here that this should be in hexadecimal and is case insensitive?

vovythevov:
@zachmullen: Thanks for your comment ! I updated the code to add the documentation.

plugins/hashsum_download/plugin_tests/hashsum_download_test.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def tearDownModule():
+    base.stopServer()
+
+
+class HashsumDownloadTest(base.TestCase):
+    def setUp(self):
+        base.TestCase.setUp(self, assetstoreType='filesystem')
+
```
jcfr: Could you add a comment summarizing the organization of the data in this test.

Something like this:

```
# Create two users: leeloo, zorg
# Create hiearchy:
#   folder1
#     |- file1  [readable by ....]
```

plugins/hashsum_download/plugin_tests/hashsum_download_test.py
```diff
@@ -0,0 +1,246 @@
...
+            path='/file/hashsum/%s/%s/download' % ('crc32', '1a2b3c4d'),
+            method='GET', user=self.user, isJson=False,
+        )
+        self.assertStatus(resp, 400)
+
+        for hashAlgorithm in ['sha512']:
+            publicDataHash = self._hashSum(self.userData, hashAlgorithm)
+            privateDataHash = self._hashSum(self.privateOnlyData, hashAlgorithm)
+
+            # Test normal use
```
jcfr: Add one private function for each sub tests.

your test would then be:

```
self._testNormalUse()
self._testNonExistentFile()
...
```

plugins/hashsum_download/server/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+        super(File, self).__init__()
+
+        self.resourceName = 'file'
+        apiRoot.file.route('GET', ('hashsum', ':algo', ':hash', 'download'),
+                           self.downloadWithHash)
+
+    @access.public
+    @describeRoute(
+        Description('Download a file by its hashsum.')
+        .param('algo', 'The type of the given hashsum. '
```
jcfr: Can you list supported hashsum. In that case would be `sha512`

plugins/hashsum_download/server/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+                       'This parameter is case insensitive.',
+               paramType='path', enum=SupportedAlgorithms)
+        .param('hash', 'The hexadcimal hashsum of the file to download. '
```
jcfr: Typo `hexadcimal`

plugins/hashsum_download/server/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+                       'This parameter is case insensitive.',
+               paramType='path')
+        .errorResponse()
+        .errorResponse('Read access was denied on the file.', 403)
```
jcfr: @brianhelba Is there a global list of all error string ? Or does it make sense to copy the text ?

jcfr: Never-mind. These errors are specific to the external data domain.

plugins/hashsum_download/server/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+        file = self._getFirstFileByHash(algo, hash)
+        if not file:
+            raise RestException('File not found.', code=404)
+
+        return self.download(id=file['_id'], params=params)
+
+    def _getFirstFileByHash(self, algo, hash, user=None):
+        """
+        Return the first file that the user has access to given a hash and an
+        algorithm.
```
jcfr: `given a hash and an algorithm.` -> `given its hash and associated hash sum algorithm name.`

plugins/hashsum_download/server/__init__.py
jcfr: `hashsum` -> `hash sum`

See https://en.wikipedia.org/wiki/Checksum

plugins/hashsum_download/server/__init__.py
vovythevov: Good catch thanks.

plugins/hashsum_download/server/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+
+        :param algo: Algorithm the given hash is encoded.
```
jcfr: `Algorithm the given hash is encoded.` -> `Algorithm the given hash is encoded with.`

plugins/hashsum_download/plugin_tests/hashsum_download_test.py
```diff
@@ -0,0 +1,246 @@
...
+            resp = self.request(
+                path='/file/hashsum/%s/%s/download' % (
+                    hashAlgorithm, publicDataHash),
+                method='GET', isJson=False, params={
+                    'offset': 15})
+            self.assertStatus(resp, 206)
+            self.assertEqual(self.userData[15:],
+                             self.getBody(resp, text=False))
+
+            # Test downloading with a range header and query range params
```
jcfr: Can you test that the mime type of the download file is correct.

plugins/hashsum_download/server/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+from girder.api import access
+from girder.api.describe import describeRoute, Description
+from girder.api.rest import RestException
+from girder.api.v1.file import File
+from girder.constants import AccessType
+
+
+class HashedFile(File):
+
+    SupportedAlgorithms = [
```
danlamanna: If I can nitpick, I think lowercase camelCase is the convention here.

jcfr: Our of curiosity, is pep8/flake8 catching this currently ?

vovythevov: This isn't caught by the style testing. (See the travis checks).

plugins/hashsum_download/server/__init__.py
vovythevov: The supported hashes will be displayed in the API page because of ```enum=supportedAlgorithms```

plugins/hashsum_download/server/__init__.py
brianhelba: @jcfr @girder/developers Indeed, I've noticed that the Flake8 style tests do not enforce camelCase for variable names. We should file this as a bug.

jbeezley: I think we ignore that error ([`N806`](https://github.com/girder/girder/blob/master/tests/flake8.cfg#L57)).

vovythevov:
@girder/developers : What is the status for this branch ? Is there something I should do to get this merged in ?

zachmullen:
I am OK with it, @jcfr added the last comments so I leave it up to him to give final approval.

jcfr:
@vovythevov LGTM after addressing the last minor comments. Since I am out travelling, I will then defer to @zachmullen @brianhelba .

brianhelba:
This LGTM.

@voythevov you can go ahead and merge it (and delete the branch afterwards).

mgrauer:
:confetti_ball:

Issue #1184, opened 2016-01-17
Add feature to explicitly link OAuth logins with a Girder account


brianhelba:
Currently, if a user logs in with an OAuth provider for the first time, and the email address returned by the provider matches an email address of an existing Girder account, that OAuth login is automatically "linked" to the matching Girder account. This allows a user to log in to the same account using multiple methods.

There are several limitations to this method:
* The primary email addresses of all OAuth logins must match the email address of the Girder account
 * If an OAuth email address does not match, a branch-new Girder user is created, which can be worse tn a simple failure, since it leads to multiple Girder accounts, which is confusing to both users and administrators
* The process is implicit, and users may know it's available

To fix this, several sub-tasks must be undertaken:
* Support for users having multiple email address must be added to Girder, with one designated as the "primary" address.
 * While not strictly necessary, it's important that GUI elements be added for users to add and remove email addresses, and designate a new primary.
* GUI elements need to be added to the user account screen, showing the currently-associated OAuth logins, and the ability to explicitly link a new OAuth login for any other enabled providers
 * When the user selects the option to link an account, a modified OAuth flow should be entered
  * Either the ``state`` variable or the temporary Girder token can be used to mark this modified flow (depending on security implications); duplicate endpoints should not be necessary
 * When a user is redirected back from this modified flow, Girder should always associate the OAuth login email with the triggering Girder account, and not implicitly create a new account
* Support should also be added to unlink any existing associated OAuth logins

Finally, the ability to merge user accounts will be helpful to clean up duplicated users caused by the current OAuth policy:
* One Girder user should be designated as the "original", and other "duplicate" user accounts should be deleted, with all ACLs in the database updated to replace the duplicate user IDs with the original user ID.

brianhelba:
@cryos Feel free to add any other requirements you may have.

cryos:
This looks pretty complete @brianhelba, if this gets implemented and merged I think it would satisfy my use case very well. I would be happy to help test this as pieces get added, it reflects well what other services do that I would say act as a good model of the right way of doing this.

PR #1240, opened 2016-02-15
Add LinkedIn provider to OAuth plugin

brianhelba:


brianhelba:
This depends on https://github.com/girder/girder/pull/1237 , for the icons to work correctly.

brianhelba:
@girder/developers Ready for review. PTAL.

zachmullen:
LGTM

brianhelba:
FYI, I also plan to add StackExchange (which seems worthwhile) and Facebook (which SlideAtlas needs for historical reasons).

@zachmullen Are there any other providers you'd like to see? 

Otherwise, we can move on to worry about #1184 and LDAP, once the last additional providers are added

zachmullen:
Google, Facebook, and Twitter seem to me to be the "big 3" right now. I'm fine with waiting on any other providers until we actually need or want them.

PR #1241, opened 2016-02-15
Add optional reference to uploads.

manthey:
This is sent to the data.process event.

This is tested in the python client.

The worker plugin defaults to placing the job id in the reference.

### DIFF
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -36,7 +36,7 @@ class Upload(Model):
         self.name = 'upload'
 
     def uploadFromFile(self, obj, size, name, parentType=None, parent=None,
-                       user=None, mimeType=None):
+                       user=None, mimeType=None, reference=None):
         """
         This method wraps the entire upload process into a single function to
         facilitate "internal" uploads from a file-like object. Example:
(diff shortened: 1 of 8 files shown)


girder/models/upload.py
```diff
@@ -172,10 +175,13 @@ def finalizeUpload(self, upload, assetstore=None):
...
         # Add an async event for handlers that wish to process this file.
-        events.daemon.trigger('data.process', {
+        eventParams = {
             'file': file,
             'assetstore': assetstore
-        })
+        }
+        if 'reference' in upload:
+            eventParams['reference'] = upload['reference']
+        events.daemon.trigger('data.process', eventParams)
```
vovythevov: Hey @manthey,
FIY, we also have a pull request to reference the upload object here, see https://github.com/girder/girder/pull/1236

manthey: Great.  I think the two PRs do fundamentally different things (both useful) -- this allows the initiator of a job run via girder_worker or romanesco to monitor the data.process events and determine if a file was the result of a particular job.

vovythevov: Arf, you're right. I didn't look to close and thought we were doing the same thing. Sorry for the noise.

zachmullen:
LGTM

zachmullen:
Do you have permission to publish v1.1.2 of girder-client? If so, feel free to cut it.

PR #1242, opened 2016-02-15
Exclude test directories from python packaging

jbeezley:
Having the test directories included in the python packaging will cause race conditions such as what happened here: http://my.cdash.org/testDetails.php?test=23360663&build=913481.  This is an attempt to fix it by limiting the files that are copied to the staging area during the packaging step.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,6 +5,7 @@ include girder/girder-version.json
 include package.json
 include README.rst
 include girder/conf/girder.dist.cfg
+recursive-exclude tests *
 recursive-include clients/web *
 recursive-include girder/mail_templates *
 recursive-include grunt_tasks *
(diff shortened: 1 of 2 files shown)


zachmullen:
I support this change, but out of curiosity, how is this causing race conditions?

jbeezley:
The cause of the failure I linked was because the asset store tests add and delete files inside `tests/assetstore`.  The packaging tests copy every file included by the manifest to a staging area.  `shutil.copytree` apparently globs a list of files and then loops through to copy them.  One of the assetstore tests deleted a file between the globbing and copying phase.

The relevant line in this is `recursive-exclude tests *`.  The rest is there because I noticed that the `plugin_tests` directories were being packaged with the rest of the source files.

jbeezley:
On a related note, I wonder if we should move the test assetstores to `tempfile.gettempdir()`.  Is there a reason it needs to be under `tests/`?

zachmullen:
That sounds reasonable to me.

zachmullen:
LGTM

PR #1248, opened 2016-02-17
Expose coverage config as a cmake variable

jbeezley:
This is needed in Minerva to configure the coverage report for submission to CDash.


jeffbaumes:
```python
int(''.join([str(pow(2,i)) for i in range(4)]))
```

zachmullen:
> `int(''.join([str(pow(2,i)) for i in range(4)]))`

Is this the new "LGTM"?

jbeezley:
`base64.b64decode('TEdUTQ==')`

zachmullen:
@jbeezley your change LGTM

@jeffbaumes most modern processors have barrel shifters in their ALUs, so you could save clock cycles by writing as:

    int(''.join([str(1 << i) for i in range(4)]))

:wink:

jeffbaumes:
> Is this the new "LGTM"?

No but good idea, I should randomly do something like @jbeezley's obfuscation sometime. Let the record show I know nothing about this branch. I just liked the issue number.

PR #1249, opened 2016-02-18
Fixes #1247. Don't pass invalid enabled plugins

### LINKED ISSUE #1247: missing enabled plugins break plugin page
@cdeepakroy just ran into this. He had an external plugin that he had enabled, but then removed its directory. It then made the plugin management UI throw errors anytime a plugin was turned on.

zachmullen:
@girder/developers PTAL

### DIFF
--- a/clients/web/src/views/body/PluginsView.js
+++ b/clients/web/src/views/body/PluginsView.js
@@ -120,6 +120,10 @@ girder.views.PluginsView = girder.View.extend({
     },
 
     _updatePlugins: function () {
+        // Remove any missing plugins from the enabled list. Can happen
+        // if the directory of an enabled plugin disappears.
+        this.enabled = _.intersection(this.enabled, _.keys(this.allPlugins));
+
         girder.restRequest({
             path: 'system/plugins',
             type: 'PUT',
--- a/clients/web/src/views/body/PluginsView.js
+++ b/clients/web/src/views/body/PluginsView.js
@@ -39,12 +39,16 @@ girder.views.PluginsView = girder.View.extend({
 
     render: function () {
         _.each(this.allPlugins, function (info, name) {
-            if (this.enabled.indexOf(name) >= 0) {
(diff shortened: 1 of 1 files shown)


manthey:
Does this properly disable plugins that required the missing plugins?  (I haven't looked if it does or not).

zachmullen:
Those will work as before, showing the unmet dependency error message, and can be disabled normally without error.

zachmullen:
Does someone have time to review this today? I'd like to get this bug fixed ASAP.

manthey:
I started girder with a missing plugin, and another plugin requiring the missing one is shown as "On", but cannot be toggled (the control is disabled).

Unrelated, since a style change some weeks ago, when plugins are turned off, the control is blank (it doesn't have the word "off" or any other indicator of its state), which seems confusing.

zachmullen:
> I started girder with a missing plugin, and another plugin requiring the missing one is shown as "On", but cannot be toggled (the control is disabled).

Hm, I'll take a look at this.

>  the control is blank (it doesn't have the word "off" or any other indicator of its state), which seems confusing.

This was actually done because bootstrap-switch appeared to break their own styling in a recent update, which was causing the word "OFF" to break and the second F to flow onto a new line. I couldn't find an easy fix so I removed the text for now.

danlamanna:
> I started girder with a missing plugin, and another plugin requiring the missing one is shown as "On", but cannot be toggled (the control is disabled).

This symptom (and that of #1247) are just side effects of the weird state Girder is in where a plugin is enabled that doesn't exist anymore. I wouldn't be surprised if more of these types of bugs popped up given this is a scenario that isn't immediately obvious to program for.

That being said, I think one solution (which would make this PR unnecessary) is for plugins (and their dependent plugins) to be immediately disabled when they fail to load due to not existing on disk. Definitely interested to hear what others think about that though.

zachmullen:
@manthey I fixed the bug you discovered, PTAL.

manthey:
LGTM.

PR #1252, opened 2016-02-18
Wait for data.process events in python client test.

manthey:
Events can be asynchronous, so wait a short time for them.

Delete the test directory before the client test to make sure it is a clean test.  If the test is interrupts, an old test directory can cause the next run to fail.


mgrauer:
LGTM.

PR #1253, opened 2016-02-19
Make it possible to execute jasmine test specs in the browser

jbeezley:
This won't work for tests that rely on `callPhantom`, but you can at least load up some of the specs in the browser to debug them.


zachmullen:
Nice! @danlamanna and I were just lamenting yesterday that we couldn't do this.

zachmullen:
LGTMITIH

jbeezley:
Yeah, I found out from @danlamanna that while you were on the hangout, I was struggling with exactly the same problem.

PR #1254, opened 2016-02-22
Custom web client test module

jbeezley:
For minerva, we want to be able to customize the setup and teardown behavior for the client tests.  Custom client test runners can derive from `web_client_test.WebClientTestCase` to extend the default behavior.

### DIFF
--- a/tests/JavascriptTests.cmake
+++ b/tests/JavascriptTests.cmake
@@ -220,6 +220,7 @@ function(add_web_client_test case specFile)
   string(REPLACE ";" " " plugins "${plugins}")
 
   set_property(TEST ${testname} PROPERTY ENVIRONMENT
+    "PYTHONPATH=$ENV{PYTHONPATH}"
     "SPEC_FILE=${specFile}"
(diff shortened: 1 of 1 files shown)


tests/JavascriptTests.cmake
```diff
@@ -212,6 +220,7 @@ function(add_web_client_test case specFile)
   string(REPLACE ";" " " plugins "${plugins}")
 
   set_property(TEST ${testname} PROPERTY ENVIRONMENT
+    "PYTHONPATH=$ENV{PYTHONPATH}"
```
zachmullen: We are freezing the pythonpath to whatever it is during config time? What's the reasoning for doing that?

jbeezley: 1. It's the same thing we do in the [`server tests`](https://github.com/girder/girder/blob/master/tests/PythonTests.cmake#L117).
2. I need to make the `minerva` directory resolvable as a python package.  The only other way is to register it with setuptools.

zachmullen:
LGTM

PR #1255, opened 2016-02-22
Register plugins via a setuptools entry_point

jbeezley:
I came up with a way to do this without breaking backward compatibility so it doesn't necessarily have to wait until version 2.0.  You can test drive it with a standalone version of the geospatial plugin [here](https://github.com/jbeezley/girder-geospatial).  Just `python setup.py develop` or `python setup.py install` this branch of girder (unnecessary once this branch is on PyPI) then
```
pip install git+https://github.com/jbeezley/girder-geospatial.git
```
The plugin `Geospatial (standalone)` should now be listed on the plugins page.

Note: this won't work for client side plugins, so I'll understand if we decide to hold off on this PR until that is resolved.

### DIFF
--- a/girder/utility/plugin_utilities.py
+++ b/girder/utility/plugin_utilities.py
@@ -42,7 +42,7 @@ from pkg_resources import iter_entry_points
 from girder.constants import PACKAGE_DIR, ROOT_DIR, ROOT_PLUGINS_PACKAGE, \
     TerminalColor
 from girder.models.model_base import ValidationException
-from girder.utility import config, mail_utils, mkdir
+from girder.utility import config as _config, mail_utils, mkdir
 
 
 def loadPlugins(plugins, root, appconf, apiRoot=None, curConfig=None,
@@ -72,7 +72,7 @@ def loadPlugins(plugins, root, appconf, apiRoot=None, curConfig=None,
     # Register a pseudo-package for the root of all plugins. This must be
     # present in the system module list in order to avoid import warnings.
     if curConfig is None:
-        curConfig = config.getConfig()
+        curConfig = _config.getConfig()
 
     if 'plugins' in curConfig and 'plugin_directory' in curConfig['plugins']:
         print(TerminalColor.warning(
@@ -109,7 +109,7 @@ def getToposortedPlugins(plugins, curConfig=None, ignoreMissing=False):
     Given a set of plugins to load, construct the full DAG of required plugins
     to load and yields them in toposorted order.
     """
-    curConfig = curConfig or config.getConfig()
+    curConfig = curConfig or _config.getConfig()
     plugins = set(plugins)
 
     allPlugins = findAllPlugins(curConfig)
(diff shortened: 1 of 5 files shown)


zachmullen:
This is great, especially how dirt simple the setup.py file is in the plugin. I'm happy to merge this now given that it's tested and doesn't break anything.

We should consider exposing the config as a decorator from girder to decorate the load method instead of having to set an attribute on it, just to keep it consistent with how we typically do things in core now.

```
@girder.plugin_utilities.config(
    name='Geospatial (standalone)',
    description='Enables the storage and querying of GeoJSON formatted geospatial data',
    version='0.1.0'
)
def load(info):
```

Feel free to pick a better name...

jbeezley:
@zachmullen PTAL

girder/utility/plugin_utilities.py
```diff
@@ -35,11 +35,14 @@
...
 import yaml
+import importlib
+
+from pkg_resources import iter_entry_points
 
 from girder.constants import PACKAGE_DIR, ROOT_DIR, ROOT_PLUGINS_PACKAGE, \
     TerminalColor
 from girder.models.model_base import ValidationException
-from girder.utility import config, mail_utils, mkdir
+from girder.utility import config as _config, mail_utils, mkdir
```
zachmullen: Hm, this makes me want to rename our new symbol to `pluginConfig` to be less vague...

zachmullen: Can somebody come up with an acceptable name?

zachmullen: I guess not. @jbeezley do you like `pluginConfig` better as a name? If so we can change it to be more descriptive, or if you have a better idea altogether, go for it.

jbeezley: My least favorite part of programming is naming things.  :/

My only thought was that many decorators are named so that they describe the function they are wrapping.  `@endpoint` means the the wrapped function is an endpoint.  I was thinking something like `@loader`.  I'm not sure that I agree that `config` -> `pluginConfig` is very beneficial.  The namespace should provide the specificity not the symbol name.  I.e. if I could totally rearrange everything, I might do something like this:
```python
from girder import plugin

@plugin.loader(config)
def loader(info):
    pass
```

That works equally well with `girder.plugin.config`,

zachmullen: In girder/\__init\__.py, we can just add `import girder.utility.plugin_utilities as plugin`, then we would have the namespace you want as an alias. I certainly like that name better.

While I agree that a decorator should normally describe the wrapped function, the function in question is actually not a decorator, it's a function that takes some arguments and returns a decorator. That might be splitting hairs, but to me the passing of parameters is an important distinction that makes me believe it's more useful to describe the decorator-returning function than the wrapped function. So, I lean toward `girder.plugin.config` over `girder.plugin.loader`, but we should decide democratically. @girder/developers

jbeezley: Between `config` and `loader`, I definitely don't care enough to advocate for it.  I was only throwing out and idea because you seemed disappointed nobody else had :).

Do you want me to put in the alias?

zachmullen: Yep!

zachmullen:
LGTMITIH

jbeezley:
Looking at that `__init__.py` file, we should really be setting `__all__`.  In any case, PTAL.

jbeezley:
Fixed

zachmullen:
LGTM

PR #1257, opened 2016-02-23
begin to add item previews, starting with image previews. Cc: @girder…

hwrod:
Adding item previews (like JSON file previews, image previews). 

Movie demo: [http://www.planetharold.com/demos/girder-item-previews.mov](http://www.planetharold.com/demos/girder-item-previews.mov)

Screenshot: 
girder-item-previews

### DIFF
--- a/clients/web/src/stylesheets/widgets/itemPreviewWidget.styl
+++ b/clients/web/src/stylesheets/widgets/itemPreviewWidget.styl
@@ -1,0 +1,20 @@
+maxImageHeight = 400px
+
+.g-widget-itempreview-header
+  margin-top: 10px
+  background-color #f0f0f0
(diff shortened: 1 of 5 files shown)


zachmullen:
@hwrod this is a really cool feature, thanks for adding this! I am hoping to have some time to review it this week.

zachmullen:
@hwrod I just talked with @cpatrick and we both think it makes sense to move this new functionality into a plugin. I'm happy to do the conversion effort myself if you'd like, or if you want to take a crack at it that's fine too. Let me know which is best for you.

hwrod:
@zachmullen Sounds good, I'll convert this feature into a plugin and update the PR.

PR #1258, opened 2016-02-24
Metadata widget flexibility

### LINKED ISSUE #1224: Allow MetadataWidget to use item fields other than "meta"
The [``MetadataWidget``](https://github.com/girder/girder/blob/master/clients/web/src/views/widgets/MetadataWidget.js) should be able to take a parameter (something like ``fieldName``) to determine which field on an item will be used for the getting / setting of metadata. If the parameter is not specified, the field will still default to ``meta``.

zachmullen:
ping @brianhelba 

fixes #1224

### DIFF
--- a/clients/web/src/views/widgets/MetadataWidget.js
+++ b/clients/web/src/views/widgets/MetadataWidget.js
@@ -19,6 +19,7 @@ girder.views.MetadataWidget = girder.View.extend({
      * @param [settings.fieldName='meta'] {string} The name of the model attribute
      *    to display/edit. The model attribute with this name should be an object
      *    whose top level keys represent metadata keys.
+     * @param [settings.title='Metadata'] {string} Title for the widget.
      * @param [settings.apiPath] {string} The relative API path to use when editing
      *    metadata keys for this model. Defaults to using the MetadataMixin default path.
      * @param [settings.accessLevel=girder.AccessType.READ] {girder.AccessType} The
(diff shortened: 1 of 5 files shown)


clients/web/src/views/widgets/MetadataWidget.js
```diff
@@ -13,7 +13,10 @@ girder.views.MetadataWidget = girder.View.extend({
 
     initialize: function (settings) {
         this.item = settings.item;
+        this.fieldName = settings.fieldName || 'meta';
         this.accessLevel = settings.accessLevel;
+        this.onMetadataEdited = settings.onMetadataEdited;
```
brianhelba: Couldn't this just be ``this.onMetadataEdited = settings.onMetadataEdited || this.item.addMetadata;``? This removes the need to do a conditional call below (where both functions are passed the exact same arguments).

brianhelba: Nevermind, ``girder.views.MetadatumEditWidget`` is part of our public API surface (although most people would not instantiate it directly), so we have to support it being created without the ``onMetadataEdited`` setting.

brianhelba:
@zachmullen The JS code looks fine, and while I haven't yet tested the CSS, it looks sane.

However, I don't think this goes far enough to address #1224 in making it easy to drop in a ``MetadataWidget`` to work against a different ``fieldName``. At the moment, the [``addMetadata``](https://github.com/girder/girder/blob/master/clients/web/src/model.js#L273) and [``editMetadata``](https://github.com/girder/girder/blob/master/clients/web/src/model.js#L295) functions will need to be entirely copy-paste reimplemented just to change the 1 line in each where ``'meta'`` is hardcoded. Additionally, reimplementing ``addMetadata`` and ``editMetadata`` is a little more burdensome than just creating an anonymous function, since both rely on ``this``.

Of course, if this were an easy fix, I would have done it myself already.

zachmullen:
@brianhelba PTAL

zachmullen:
ping @brianhelba is this good to merge?

brianhelba:
LGTM. This should satisfy the use case (and more, since you can configure the path too).

PR #1259, opened 2016-02-24
Prevent division by zero in js coverage reporting

jbeezley:
https://travis-ci.org/Kitware/minerva/builds/111517943#L4229-L4238

### DIFF
--- a/tests/js_coverage_tool.py
+++ b/tests/js_coverage_tool.py
@@ -93,12 +93,24 @@ def combine_report(args):
     report(args, combined, stats)
 
 
+def safe_divide(numerator, denominator):
+    """
+    Return numerator / denominator or 0 if denominator <= 0.
+    """
+    numerator = float(numerator)
+    denominator = float(denominator)
+    if denominator > 0:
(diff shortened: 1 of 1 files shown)


tests/js_coverage_tool.py
```diff
@@ -93,12 +93,24 @@ def combine_report(args):
     report(args, combined, stats)
 
 
+def safe_divide(numerator, denominator):
+    """
+    Return numerator / denominator or 0 if denominator <= 0.
+    """
+    numerator = float(numerator)
+    denominator = float(denominator)
+    if denominator > 0:
```
zachmullen: Sort of meaningless here, but this should be != instead of >.

jbeezley: That's a habit from my hpc days where the general rule of thumb is to never compare equality between floats.  Especially with zero because it is slow.  Here you can make the philosophical argument that if there are -1 lines of code, then none of them can be covered by tests regardless of what the numerator is.

zachmullen:
LGTM

PR #1260, opened 2016-02-25
Add test labels and use a Travis build matrix

jbeezley:
This splits the tests into 4 groups (server, client, packaging, and other) that will run in independent containers on Travis.  This has a few benefits:
* tests are less likely to interfere with each other
* rerunning a single test group is faster (5 minutes vs 10 minutes)
* you can use up to 5 containers in parallel on Travis
* javascript coverage reports now show up on cdash

There are a couple of other tangentially related changes here as well:
* The external data tests are now turned off with the `RUN_CORE_TESTS` option
* Updated cmake to 3.4.3 because I thought I was encountering bugs related to test labels
* The `node_modules` directory is now cached between runs (the `npm prune` line ensures that it remains pristine)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_install:
     - pip install -U pip virtualenv
 install:
     - if [ "$GIRDER_TEST_GROUP" == "server" -o "$GIRDER_TEST_GROUP" == "client" ] ; then pip install -r requirements.txt -r requirements-dev.txt -e . ; fi
-    - if [ "$GIRDER_TEST_GROUP" == "packaging" -o "$GIRDER_TEST_GROUP" == "client" -o "$GIRDER_TEST_GROUP" == "other" ] ; then npm install ; fi
+    - npm install
 script:
     - mkdir _build
     - cd _build
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,12 @@ before_install:
     - npm --version
     - pip install -U pip virtualenv
 install:
-    - pip install -r requirements.txt -r requirements-dev.txt -e .
-    - npm install
+    - if [ "$GIRDER_TEST_GROUP" -eq "server" -o "$GIRDER_TEST_GROUP" -eq "client" ] ; then
+    -   pip install -r requirements.txt -r requirements-dev.txt -e .
+    - fi
+    - if [ "$GIRDER_TEST_GROUP" -eq "packaging" -o "$GIRDER_TEST_GROUP" -eq "client" -o "$GIRDER_TEST_GROUP" -eq "other" ] ; then
+    -   npm install
+    - fi
 script:
     - mkdir _build
     - cd _build
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,12 @@ python:
     - "2.7"
     - "3.4"
 
+env:
+    - GIRDER_TEST_GROUP=server
+    - GIRDER_TEST_GROUP=client
+    - GIRDER_TEST_GROUP=packaging
+    - GIRDER_TEST_GROUP=other
+
 cache:
     directories:
         - $HOME/.cache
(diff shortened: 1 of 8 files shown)


scripts/tests/test_json_config_expand_relpaths_script/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ function(add_json_config_expand_relpaths_test name)
     COMMAND ${NODEJS_EXECUTABLE} ${JSON_CONFIG_EXPAND_RELPATHS_SCRIPT} ${ARGN}
     )
   set(testname ${testname} PARENT_SCOPE)
+  set_property(TEST "${testname}" PROPERTY LABELS girder_devel)
```
zachmullen: Just to make sure I'm reading this properly, any label other than `(girder_server|girder_client|girder_packaging)` falls into the "other" category, right?

jbeezley: Exactly.  In case people forget to add a label to a test, it will be run in this group.

.travis.yml
```diff
@@ -18,17 +24,20 @@ before_install:
     - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "3" ]; then export PY3="true"; else export PY2="true"; fi
     - if [ -n "${PY3}" ]; then export MONGO_VERSION=3.0.7; export PY_COVG="OFF"; else export MONGO_VERSION=2.6.11; export PY_COVG="ON"; export DEPLOY=true; fi
     - if [ -n "${PY3}" ]; then export IGNORE_PLUGINS=hdfs_assetstore,metadata_extractor; fi
-    - CACHE=$HOME/.cache source ./scripts/install_mongo.sh
+    - if [ "$GIRDER_TEST_GROUP" != "other" ] ; then CACHE=$HOME/.cache source ./scripts/install_mongo.sh ; fi
```
zachmullen: And is this line saying that we don't install mongo at all in the "other" case?

jbeezley: I was originally trying to limit the number of things being installed into each container to speed up the setup time.  Unfortunately, we need basically a full setup in each one.  This is the only optimization that survived.  In reality, this step probably doesn't save any time.

zachmullen:
I don't think I've seen the tests fail like [this](http://my.cdash.org/viewTest.php?onlyfailed&buildid=918842) before, do you think it's a result of your change?

jbeezley:
I know I've seen it before.  It's a random failure that this change may have made more likely.

I'm fairly certain there is a race condition in the way we are loading test specs.  That plus the blanket instrumentation makes for some complicated asynchronous interactions.  I hit it when trying to do client side tests in minerva because there are more script tags being loaded.  I also suspect it is the cause of the failures when updating to phantomjs 2.x.

zachmullen:
Just restarted the failed server build and it turned around in 4 minutes. LGTM!

brianhelba:
@jbeezley How confident are you that CMake 3.4.3 is required? Test labels are a feature that should be 100% supported for developers to run locally, so if a later version of CMake is required, the [``cmake_minimum_required``](https://github.com/girder/girder/blob/master/CMakeLists.txt#L1) should be updated too. Of course, requiring the *very* latest CMake 3.4.3 is going to burdensome to most developers, so if we can require an older version that still provides full support for labels, that would be best.

jbeezley:
Not very confident.  The actual issue was with the `ctest_test(INCLUDE_LABEL ...)` function not with the `ctest -L` command line interface, so it is less likely to affect users.

zachmullen:
@brianhelba I don't believe it's required, in fact our `cmake_minimum_required(VERSION` remains unchanged at 2.8.6. This is just a change to the version travis uses, and a good way to dogfood the latest CMake release.

brianhelba:
@zachmullen I think there are two versions of CMake that we *should* test:
1. The oldest allowed version (from ``cmake_minimum_required``), guaranteeing that new Girder infrastructure features (for example, test labels) still function with old versions of CMake that developers may have installed.
2. The latest released version, checking that upstream CMake doesn't add any regressions that break Girder.

However, I don't necessarily think this is important enough to add another dimension to our build matrix. If I had to pick one of the cases to test on Travis though, I'm intuitively more concerned about case 1, especially since many developers will just use the distro-packaged version of CMake, which tends to be quite old. For this reason, I'd advocate making Travis use the same version of CMake as ``cmake_minimum_required``.

jbeezley:
I understand that sentiment, but I don't think the CI process is where we want build out that kind of complexity.  We have a hard enough time getting tests to reliably pass as is.  Maybe we can set up some nightly tests for this purpose.

zachmullen:
I agree with both @brianhelba and @jbeezley .

PR #1261, opened 2016-02-26
Allow altUrl to be specified on a girder.Model

danlamanna:
This is the analog to #1145.

I'm using models in another Girder application that inherit from girder.Model but have entirely different endpoints, so I need an altUrl. We may want to conform to the `Backbone.Model.url` convention in the future.


zachmullen:
LGTM, thanks

PR #1262, opened 2016-02-27
update the get-pip.py location as requested by get-pip.py itself

hwrod:
Looks like the get-pip.py shell script needs updating; the latest URL is provided here.


zachmullen:
Thanks for the fix, LGTM!

PR #1263, opened 2016-02-29
Vagrant provision improvements

brianhelba:
* Improve Vagrantfile
 * Default VM will have 768MB of memory, which should fix npm's OOM failures.
 * Add future support for 'ansible_local' provisioner (once bugs are fixed in the next release), so Ansible will not be required on the host machine.
 * Add a message explaining where the new Girder web server is running
 * Make VM naming more clear and consistent
 * Mount the shared directory in a more obvious and accessible location.
* Update documentation with additional prerequisites
 * Pillow 3+ requires libjpeg and zlib.
 * Also note that npm 4.x should be installed on RHEL.
* Refactor Ansible playbook for Vagrant
 * Fix system dependencies.
 * Use Node 4.x.
 * Use MongoDB 3.2.

Fixes #861.
Fixes #916.

PR #1265, opened 2016-02-29
Provision improvements persistent girder server

hwrod:
Builds on @brianhelba's great job on updating the Vagrantfile.

Adds a way to persistently run a Girder server, with automatic `grunt watch` and Girder CherryPy server available on localhost:8080. Fixes https://github.com/girder/girder/pull/1263 WIP TODO #1

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,8 +1,33 @@
 Vagrant.configure("2") do |config|
   config.vm.box = "ubuntu/trusty64"
+
+  config.vm.hostname = "girder"
+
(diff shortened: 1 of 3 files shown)


zachmullen:
Thanks for these improvements. Since there is still work-in-progress on @brianhelba 's topic branch (#1263), could you close this and instead make a PR against the `provision-improvements` branch? Then we can have @brianhelba review that.

hwrod:
@zachmullen will do

Issue #1268, opened 2016-03-07
Fontello build task failing


jbeezley:
The grunt task `fontello:ext_font` is currently failing with the following stack trace:
```
Running "fontello:ext_font" (fontello) task
[D] Task source: /Users/jbeezley/git/girder/node_modules/grunt-fontello/tasks/fontello.js
Verifying property fontello.ext_font exists in config...OK
File: [no files]
Options: host="http://fontello.com", config="clients/web/fontello.config.json", fonts="clients/web/static/built/fontello/font", styles="clients/web/static/built/fontello/css", exclude=[], zip=false, scss=false, force, updateConfig=false
Verify paths...OK
[D] clients/web/static/built/fontello/font verified!
[D] clients/web/static/built/fontello/css verified!
Fetching archive...Creating session...OK
[D] sid: 95d43a8575e3b6fccfdc27fe9305aab9
Fatal error: Cannot read property 'config' of null
TypeError: Cannot read property 'config' of null
    at setSession (/Users/jbeezley/git/girder/node_modules/grunt-fontello/tasks/lib/fontello.js:33:43)
    at fetchStream (/Users/jbeezley/git/girder/node_modules/grunt-fontello/tasks/lib/fontello.js:129:3)
    at /Users/jbeezley/git/girder/node_modules/grunt-fontello/tasks/lib/fontello.js:106:12
    at done (/Users/jbeezley/git/girder/node_modules/needle/lib/needle.js:357:7)
    at PassThrough.<anonymous> (/Users/jbeezley/git/girder/node_modules/needle/lib/needle.js:546:9)
    at emitNone (events.js:73:20)
    at PassThrough.emit (events.js:167:7)
    at endReadableNT (_stream_readable.js:906:12)
    at nextTickCallbackWith2Args (node.js:455:9)
    at process._tickCallback (node.js:369:17)
```

After this happens, the fontello config file is mutated:
```
$ git diff
diff --git a/clients/web/fontello.config.json b/clients/web/fontello.config.json
index 4a52502..a374f6c 100644
--- a/clients/web/fontello.config.json
+++ b/clients/web/fontello.config.json
@@ -1,5 +1,5 @@
 {
-   "name": "3643f1665274345f0d7b88fbdb539555",
+   "name": "",
    "css_prefix_text": "icon-",
    "css_use_suffix":
```
after which the task will succeed.

manthey:
[...]

PR #1269, opened 2016-03-09
Workaround for the grunt-fontello task.

manthey:
The grunt-fontello task mutates the fontello config file.  If this file has a name entry that is an invalid or expired fontello session ID, the task fails.  To work around this, copy the fontello config file to an other directory and have the grunt-fontello task use that.  This means that we always will get a new fontello session, and the task will avoid the condition of having a bad session.  This also avoids mutating the source file.

### DIFF
--- a/clients/web/fontello.config.json
+++ b/clients/web/fontello.config.json
@@ -1,5 +1,5 @@
 {
-	"name": "3643f1665274345f0d7b88fbdb539555",
+	"name": "",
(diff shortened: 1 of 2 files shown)


zachmullen:
LGTM. I wish it did some intelligent caching, but since this is in the init step and only takes about 1-2 seconds, seems fine.

manthey:
Resolves issue #1268.

PR #1272, opened 2016-03-14
Provision improvements persistent girder server

hwrod:
Port of master PR: https://github.com/girder/girder/pull/1265

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -3,8 +3,8 @@ Vagrant.configure("2") do |config|
 
   config.vm.hostname = "girder"
 
-  config.vm.network "forwarded_port", guest: 80, host: 9080
-  config.vm.post_up_message = "Girder is running at http://localhost:9080"
+  config.vm.network "forwarded_port", guest: 8080, host: 8080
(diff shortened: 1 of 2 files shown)


brianhelba:
Thanks for the PR @hwrod (and sorry for the delay in doing a review).

Unfortunately, I don't think this is the right approach. The steps you added will by run once by Ansible upon the initial provisioning process. This results in the following problems:
* The commands block and never return, causing Ansible to hang and never complete the provisioning process.
* Even if you were able to run the commands in the background, when Ansible completes, it will disconnect from the server, causing a `SIGHUP` to be sent, killing the background job.
* Even if you ran the commands backgrounded and inside the `nohup` command, the server would only persist until the machine was shut down or restarted. Then you'd have to re-run the entire Ansible playbook again (typically via `vagrant provision`) just to get a server running.
* Even if you got it all to work, this just isn't how Ansible was ever intended to be used.

However, I'll try to finish off my PR #1265 so that we can get this functionality out.

brianhelba:
I'm closing this PR. Feel free to re-open this or another new PR if you do have another approach you'd like to share.

PR #1273, opened 2016-03-14
Fix race conditions in client side testing

jbeezley:
I *think* this should resolve the race conditions that I have seen recently while writing plugin tests.  I found a few problems with the test runner:

* Blanket instruments and loads code asynchronously, so occasionally it would load a source files out of order.
* Scripts add by `addCoveredScripts` were also loaded asynchronously so their order was **not** respected as advertised.
* The jasmine test suite was being executed before plugin sources were loaded.
* `window.onload` isn't sufficient to guarantee that the instrumented scripts have been loaded.

To fix this, I created a promise chain as `girderTest.promise` that loads each script in sequence and is fully resolved when all of the scripts are loaded.  I also added a standard `girderTest.startApp` method that will load the application after all of the sources have been loaded.

I also removed the manual invocation of jasmine's test runner.  It was necessary before because we weren't using the `blanket.requiringFile` method correctly.

### DIFF
--- a/clients/web/test/spec/adminSpec.js
+++ b/clients/web/test/spec/adminSpec.js
@@ -1,14 +1,7 @@
 /**
  * Start the girder backbone app.
  */
-$(function () {
-    girder.events.trigger('g:appload.before');
-    var app = new girder.App({
-        el: 'body',
-        parentView: null
-    });
-    girder.events.trigger('g:appload.after');
-});
+girderTest.startApp();
 
 describe('Create an admin and non-admin user', function () {
 
--- a/clients/web/test/spec/collectionSpec.js
+++ b/clients/web/test/spec/collectionSpec.js
@@ -1,14 +1,7 @@
 /**
  * Start the girder backbone app.
  */
-$(function () {
-    girder.events.trigger('g:appload.before');
-    var app = new girder.App({
-        el: 'body',
-        parentView: null
-    });
-    girder.events.trigger('g:appload.after');
-});
+girderTest.startApp();
 
 describe('Test collection actions', function () {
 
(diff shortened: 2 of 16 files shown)


zachmullen:
This change LGTM, but could you link an example of the sort of error you were seeing before?

jbeezley:
It is tough to link to any concrete examples.  For plugins loaded via the `addCoveredScripts` method, you can pretty easily observe the jasmine suite starting before the scripts are executed (add `blanket.options('debug', true)` in the test environment).  The same happened when adding scripts via normal script tags when I was writing tests for Minerva.  It is the reason I had to resort to a `setTimeout` [here](https://github.com/Kitware/minerva/blob/13f6b1f8092fa43d05e9525783db01487d861dc6/plugin_tests/client/minervaSpec.js#L4).

The issues with `addCoveredScripts` are the easiest to reproduce.  The rest of the issues are more theoretical, but I suspect it might be the cause of the fairly consistent 2-3 client tests failing on my dashboard every night.  It may be less likely to occur on travis because it is run in parallel on a VM with 1.5 cores available so very little parallel processing is actually occurring.  My dashboard machine has 8 cores and the tests are run in serial.

What this really does is add some explicit blocking to blanket, which if you look at the source is an absolute [mess of callbacks](https://github.com/alex-seville/blanket/blob/master/src/blanket_browser.js#L236-L274) literally everywhere.  Even the author of blanket is now [recommending](https://github.com/alex-seville/blanket/commit/baea610953cf3481f54a1113fade06975ce45c39) transitioning to istanbul.

zachmullen:
Thanks for the explanation. Guess we should switch to Istanbul?

cpatrick:
It does look like Istambul is more popular.

jbeezley:
I think we should switch to Istanbul, but not now.  It will be a much easier task once our webpack build is in place.

PR #1274, opened 2016-03-15
Add method to upload content to existing file

cjh1:


cjh1:
@mgrauer PTAL

clients/python/girder_client/__init__.py
```diff
@@ -486,6 +486,60 @@ def uploadFileToItem(self, itemId, filepath, reference=None):
...
+        :param progressCallback: If passed, will be called after each chunk
+            with progress information. It passes a single positional argument
+            to the callable which is a dict of information about progress.
+        :type progressCallback: callable
+    """
+    def _uploadContents(self, uploadObj, stream, size, progressCallback=None):
+        offset = 0
+        uploadId = uploadObj['_id']
+        while True:
+            data = stream.read(self.MAX_CHUNK_SIZE)
```
mgrauer: @cjh1 

This concerns me, a small amount anyway.  I realize it was here before and just got moved.

If a stream is passed with more bytes than `size`, we'll potentially read more than `size`.  

Should this be something like

    data = stream.read(min(self.MAX_CHUNK_SIZE, (size - offset)))

?

zachmullen: `stream.read(n)` should read n bytes or until the end of stream, I believe.

cjh1: Yes, @zachmullen is right it reads at most n

mgrauer: right, what I'm worried about (maybe uselessly) is that we pass in a stream, which could be unlimited (reading from a socket) in size, so we could read more bytes from the stream than the `size` param.  `size` param is described as the length of the file, which helps to cover the case.

It's getting stuck in my craw for some reason.

:bird:

clients/python/girder_client/__init__.py
```diff
@@ -520,48 +574,43 @@ def uploadFile(self, parentId, stream, name, size, parentType='item',
...
+        return self._uploadContents(
+            obj, stream, size, progressCallback=progressCallback)
 
-            if '_id' not in obj:
-                raise Exception('After uploading a file chunk, did'
-                                ' not receive object with _id. Got instead: ' +
-                                json.dumps(obj))
+    def uploadFileContents(self, file_id, stream, size, reference=None):
+        """
+        Uploads the contents of an existing file..
```
mgrauer: I can't tell if this is a period or ellipsis, though I'm enjoying the ambiguity!

clients/python/girder_client/__init__.py
```diff
@@ -520,48 +574,43 @@ def uploadFile(self, parentId, stream, name, size, parentType='item',
...
 
-            if callable(progressCallback):
-                progressCallback({
-                    'current': offset,
-                    'total': size
-                })
+        :param file_idd: ID of file to update
```
mgrauer: file_id

clients/python/girder_client/__init__.py
```diff
@@ -486,6 +486,60 @@ def uploadFileToItem(self, itemId, filepath, reference=None):
...
+        :param size: The length of the file. This must be exactly equal to the
+            total number of bytes that will be read from ``stream``, otherwise
+            the upload will fail.
+        :type size: str
+        :param progressCallback: If passed, will be called after each chunk
+            with progress information. It passes a single positional argument
+            to the callable which is a dict of information about progress.
+        :type progressCallback: callable
+    """
+    def _uploadContents(self, uploadObj, stream, size, progressCallback=None):
```
mgrauer: This function is wearing its docstring on the outside!

:nail_care:

clients/python/girder_client/__init__.py
```diff
@@ -486,6 +486,60 @@ def uploadFileToItem(self, itemId, filepath, reference=None):
                                 ' not receive object with _id. Got instead: ' +
                                 json.dumps(obj))
 
+    """
+        Uploads contents of a file
```
mgrauer: This could take one of those spare periods from below.

clients/python/girder_client/__init__.py
cjh1: I am happy to make the change you suggest, however, in that case the upload would fail as you are try to upload more data than was specified I think.

clients/python/girder_client/__init__.py
cjh1: How embarrassing for it!

clients/python/girder_client/__init__.py
zachmullen: Also let's remain consistent and camelCase our vars.

clients/python/girder_client/__init__.py
```diff
@@ -520,48 +574,43 @@ def uploadFile(self, parentId, stream, name, size, parentType='item',
...
+        :param stream: Readable stream object.
+        :type stream: file-like
+        :param size: The length of the file. This must be exactly equal to the
+            total number of bytes that will be read from ``stream``, otherwise
+            the upload will fail.
+        :type size: str
+        :param reference: optional reference to send along with the upload.
+        :type reference: str
+        """
+        path = 'file/' + file_id + '/contents'
```
zachmullen: Minor nit, but let's use `%s` interpolation for this var since it's in the middle of the string

mgrauer:
works, LGTM.  

:fish: :fries:

PR #1280, opened 2016-03-22
Dockerfile: Add missing grunt_tasks

Xarthisius:
Without `grunt_tasks` static files are not generated.


cpatrick:
LGTM! Merging.

Thanks @Xarthisius!

PR #1281, opened 2016-03-30
Enforce more of PEP257 -- Docstring Conventions

msmolens:
Enforce the following warnings from flake8-docstrings for [PEP257](https://www.python.org/dev/peps/pep-0257/):

    D208 - Docstring over-indented.
    D209 - Put multi-line docstring closing quotes on separate line.
    D300 - Use """triple double quotes""".

### DIFF
--- a/plugins/provenance/server/resource.py
+++ b/plugins/provenance/server/resource.py
@@ -91,8 +91,10 @@ class ResourceExt(Resource):
                 self.boundResources[resource] = True
 
     def unbindModels(self, resources={}):
-        """Unbind any models that were bound and aren't listed as needed.
-        :param resources: resources that shouldn't be unbound."""
+        """
+        Unbind any models that were bound and aren't listed as needed.
+        :param resources: resources that shouldn't be unbound.
+        """
         # iterate over a list so that we can change the dictionary as we use it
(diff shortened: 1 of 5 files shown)


plugins/provenance/server/resource.py
```diff
@@ -323,8 +325,10 @@ def snapshotResource(self, obj):
         return snap
 
     def fileSaveHandler(self, event):
-        """When a file is saved, update the provenance of the parent item.
-        :param event: the event with the file information."""
+        """
+        When a file is saved, update the provenance of the parent item.
+        :param event: the event with the file information.
```
zachmullen: Sphinx actually wants a blank line above the start of `:param` list

msmolens: Agreed, but that seems like a separate, larger issue. That blank line is missing in every method in this file, as well as in other places in this project.

Could documentation generation be covered by a test? Running sphinx shows some errors and warnings, and I don't think it even processes the plugin code right now.

zachmullen: I think it would be great to run sphinx as a test and look for errors. It's fine to do that on a different branch.

zachmullen:
LGTM

PR #1282, opened 2016-03-30
Use absolute path for database volume in docker-compose.yml

jbeezley:
Using the latest version of docker-compose you get this error:
```
ERROR: Invalid volume spec "data/db": volumeabs: Invalid volume destination path: 'data/db' mount path must be absolute.
```


zachmullen:
LGTM

PR #1283, opened 2016-03-30
Trigger g:changed event *after* rerendering

jbeezley:
I'm not sure if there is a reason the `g:changed` event was being triggered before rendering, but this change makes the `FileListWidget` and `ItemListWidget` consistent with the behavior [`FolderListWidget`](https://github.com/girder/girder/blob/master/clients/web/src/views/widgets/FolderListWidget.js#L29-L30).


zachmullen:
LGTM

PR #1289, opened 2016-04-03
Show folder counts for users and collections in hierarchy widget

zachmullen:


zachmullen:
@brianhelba this is now ready for review.

clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -352,30 +352,30 @@ girder.views.HierarchyWidget = girder.View.extend({
...
+        if (this.parentModel.has('nFolders')) {
+            showCounts();
+        } else {
+            this.parentModel.set('nFolders', 0); // prevents fetching details twice
+            this.parentModel.once('g:fetched.details', function () {
+                showCounts();
+            }, this).fetch({extraPath: 'details'});
         }
 
+        this.parentModel.off('change:nItems', showCounts, this)
```
brianhelba: Why are the callbacks being re-added?

clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -352,30 +352,30 @@ girder.views.HierarchyWidget = girder.View.extend({
...
-                            .on('change:nItems', showCounts, this)
-                            .off('change:nFolders', showCounts, this)
-                            .on('change:nFolders', showCounts, this);
+        if (this.parentModel.has('nFolders')) {
+            showCounts();
+        } else {
+            this.parentModel.set('nFolders', 0); // prevents fetching details twice
+            this.parentModel.once('g:fetched.details', function () {
+                showCounts();
+            }, this).fetch({extraPath: 'details'});
```
brianhelba: Just to be clear, since we no longer check that the the resource is a `folder`, every single resource type (`user`, `collection`, `folder`) that the `HierarchyWidget` displays is expected to have a `GET .../details` REST endpoint, correct? Is it ever intended that new custom resource types be displayed with the current hierarchy widget?

brianhelba:
I have a few questions out of curiosity, but this LGTM otherwise, and I just tested it locally.

clients/web/src/views/widgets/HierarchyWidget.js
zachmullen: Probably to avoid duplicate bindings. I wish there was a way in Backbone to say "bind unless it's not already bound"...

clients/web/src/views/widgets/HierarchyWidget.js
zachmullen: Good point. I hadn't intended as much, this is really just meant to traverse the core data hierarchy. If someone wants to repurpose this widget to display other types in the future, we can tweak it to support that, though I would guess there are other reasons besides this that it might break on other types.

PR #1296, opened 2016-04-08
Hierarchy widget improvements

zachmullen:
* The "info" and "access control" buttons should also appear in the hierarchy widget for collections.
* Descriptions (just the text, not any markup) of currently viewed collections and folders now appear to the right of the breadcrumbs; overflow of this text does not wrap, but is elided, and clicking on it opens the full info dialog that will render the markdown and provide other details.
* Collection actions now appear in the action dropdown in the hierarchy widget when viewing a collection.
* Action dropdown in the widget when viewing a collection now correctly shows the collection icon rather than the open folder icon.
* Once a user descends in the hierarchy widget below the collection level, the "collection actions" button at the top of the collection view now disappears to avoid confusion.


zachmullen:
Depends on #1289

kotfic:
just an FYI:  minerva's S3 browser depends on the API in HierarchyWidget.js [here](https://github.com/Kitware/minerva/blob/master/web_external/js/views/widgets/ReadOnlyHierarchyWidget.js)

zachmullen:
Hm, I don't foresee this breaking that in any obvious way, however I think it would be great if we could get something upstream that supports your use case so you wouldn't have to hack it with your own subclass. A read-only flag at widget instantiation time would be generally useful.

zachmullen:
@brianhelba can you take a look at the UI changes in this branch and let me know if they are to your liking? If so I will add some asserts for them in the tests.

zachmullen:
@girder/developers PTAL

jbeezley:
It looks like "Total size" in the collection info widget doesn't update when you add or remove files.  Other than that, this LGTM.

PR #1300, opened 2016-04-08
Miscellaneous code quality improvements

brianhelba:
This a PR full of assorted small changes that I've accumulated locally over the past few months. None were significant to make a dedicated PR for, but I'd like to get them in.

This will probably be easiest to review on a commit-by-commit basis. None of the changes should have any immediate impact on behavior, but are still desirable for code quality.
 
If desired, I can split out some commits into their own PR(s).

* Use six.wraps instead of functools.wraps
 * Python 3 didn't have a "functools.wraps" until 3.4.
* Call superclass constructor in AssetstoreAdapter classes
 * This doesn't change behavior now, but is more semantically correct and could be useful in the future.
* Clean some whitespace
* Use isinstance to determine Python types
 * Python is a language for consenting adults. Using direct type comparison via "type(...) ==" or "type(...) is" prevents downstream developers from ever subclassing. Unless subclassing needs to be explicitly forbidden, type checking should always use "isinstance".
* Fix install docs to recommend using Node 4.x on RPM distros
 * This is equivalent to what is already recommended for DEB distros.
* Refactor a portion of test code to eliminate duplication
* Generalize the AccessControlledModel.list method
 * Previously, Group and Collection had identical implementations of the "list" method. This code can be generalized to all AccessControlledModel without affecting any behavior.
 * Note that Job.list has different semantics than the AccessControlledModel.list which it now directly overrides.

### DIFF
--- a/docs/prerequisites.rst
+++ b/docs/prerequisites.rst
@@ -100,7 +100,7 @@ Install MongoDB server using YUM: ::
 
 Enable the Node.js YUM repository: ::
 
-    curl -sL https://rpm.nodesource.com/setup_4.x | sudo bash
+    curl -sL https://rpm.nodesource.com/setup_4.x | sudo bash -
 
 Install Node.js and NPM using YUM: ::
 
--- a/docs/prerequisites.rst
+++ b/docs/prerequisites.rst
@@ -100,7 +100,7 @@ Install MongoDB server using YUM: ::
 
 Enable the Node.js YUM repository: ::
 
-    curl -sL https://rpm.nodesource.com/setup | sudo bash -
+    curl -sL https://rpm.nodesource.com/setup_4.x | sudo bash
 
 Install Node.js and NPM using YUM: ::
 
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -165,12 +165,12 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
         """
         :param assetstore: The assetstore to act on.
         """
+        super(S3AssetstoreAdapter, self).__init__(assetstore)
         if ('accessKeyId' in assetstore and 'secret' in assetstore and
                 'service' in assetstore):
             assetstore['botoConnect'] = makeBotoConnectParams(
                 assetstore['accessKeyId'], assetstore['secret'],
                 assetstore['service'])
-        self.assetstore = assetstore
 
     def _getRequestHeaders(self, upload):
         return {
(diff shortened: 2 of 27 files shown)


brianhelba:
@girder/developers PTAL.

docs/prerequisites.rst
```diff
@@ -100,7 +100,7 @@ Install MongoDB server using YUM: ::
 
 Enable the Node.js YUM repository: ::
 
-    curl -sL https://rpm.nodesource.com/setup | sudo bash -
+    curl -sL https://rpm.nodesource.com/setup_4.x | sudo bash
```
msmolens: Is there a good reason to remove the `-`? Both ways should work, but these instructions are now inconsistent with the APT repository instructions, as well as the [node.js instructions](https://nodejs.org/en/download/package-manager/#enterprise-linux-and-fedora).

brianhelba: You are correct. I'll fix.

girder/utility/s3_assetstore_adapter.py
```diff
@@ -165,12 +165,12 @@ def __init__(self, assetstore):
         """
         :param assetstore: The assetstore to act on.
         """
+        super(S3AssetstoreAdapter, self).__init__(assetstore)
         if ('accessKeyId' in assetstore and 'secret' in assetstore and
                 'service' in assetstore):
             assetstore['botoConnect'] = makeBotoConnectParams(
                 assetstore['accessKeyId'], assetstore['secret'],
                 assetstore['service'])
-        self.assetstore = assetstore
```
zachmullen: It doesn't matter currently, but would it be more appropriate to call `super` after modifying the document for forward compatibility?

zachmullen:
LGTM other than my one comment (and Max's comment)

girder/utility/s3_assetstore_adapter.py
brianhelba: Good point, but I'd say just change all the `assetstore[...] ` to use `self.assetstore[...]`. It's idiomatic that the superclass constructor be called first, unless you need to do some special setup.

zachmullen: Yep, that works also, let's do one or the other.

brianhelba:
Added:
* Fix a typo in install documentation
* Construct AssetstoreAdapter in a forwards-compatible way
 * In the future, the "self.assetstore" that is set by "AbstractAssetstoreAdapter.__init__" may not be reference-equal with the "assetstore" parameter.

As soon as the tests pass and I get a final :+1: from @zachmullen on this last commit, I'll merge.

zachmullen:
LGTM

PR #1301, opened 2016-04-08
Fix a potential issue with uploading files.

manthey:
If you have the setting MINIMUM_UPLOAD_CHUNK_SIZE set to an absurdly high value, uploading files could fail.  This fixes that by using the larger of the system setting or 32MB, rather than 32MB regardless.


zachmullen:
LGTM

PR #1305, opened 2016-04-08
Introduce new command: localsync

Xarthisius:
'localsync' can be used to update existing directory that was previously
downloaded using 'girder-cli download'. It compares items' metadata and
skips download if nothing has changed.

Useful for huge folders

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -871,7 +871,7 @@ class GirderClient(object):
         :param dest: The local download destination.
         """
 
-        with open(os.path.join(dest, ".girder_metadata"), "w") as fh:
+        with open(os.path.join(dest, '.girder_metadata'), 'w') as fh:
             fh.write(json.dumps(self.incomingMetadata))
 
     def loadLocalMetadata(self, dest):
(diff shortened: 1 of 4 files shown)


clients/python/girder_client/cli.py
```diff
@@ -94,6 +95,15 @@ def main():
             print('download command only accepts parent-type of folder')
         else:
             g.downloadFolderRecursive(args.parent_id, args.local_folder)
+            g.saveLocalMetadata(args.local_folder)
```
zachmullen: I think we should make the saving of local metadata optional and off by default when using the existing download command so that we don't accidentally break anybody's existing workflows.

clients/python/girder_client/__init__.py
```diff
@@ -743,12 +757,40 @@ def downloadFolderRecursive(self, folderId, dest):
...
+        Reads item metadata from a local folder.
+
+        :param dest: The local download destination.
+        """
+
+        try:
+            with open(os.path.join(dest, ".girder_metadata"), "r") as fh:
+                self.local_metadata = json.loads(fh.read())
+        except OSError:
+            print("Local metadata does not exists. Falling back to download.")
```
zachmullen: I know we have done a bad job of this in the python client, but we are trying to stick to camelCase symbol names moving forward.

clients/python/girder_client/__init__.py
```diff
@@ -839,12 +853,40 @@ def downloadFolderRecursive(self, folderId, dest):
...
+        Reads item metadata from a local folder.
+
+        :param dest: The local download destination.
+        """
+
+        try:
+            with open(os.path.join(dest, ".girder_metadata"), "r") as fh:
+                self.localMetadata = json.loads(fh.read())
+        except (IOError, OSError):
+            print("Local metadata does not exists. Falling back to download.")
```
zachmullen: For consistency, let's stick to just using single quotes.

zachmullen:
LGTM, thanks @Xarthisius

PR #1307, opened 2016-04-11
Fixes #1217. Support MIME type for uploads via girder_client

### LINKED ISSUE #1217: Python client does not specify MIME type in uploads
The [``GirderClient.uploadFile``](https://github.com/girder/girder/blob/master/clients/python/girder_client/__init__.py#L483) (and similar) methods should allow a MIME type to be specified as an optional parameter.

Additionally, Girder anticipates that most web browser clients will autodetect and set the MIME type of uploaded files. The Python client can also perform this function, using [mimetypes.guess_type](https://docs.python.org/2/library/mimetypes.html#mimetypes.guess_type).

zachmullen:


mgrauer:
Can you test the usage of this in `uploadFileToItem` and also mention the feature in the docs?

zachmullen:
@mgrauer I added the testing of `uploadFileToItem` (though I think that method should be deprecated). As for the documentation, the new arguments should be automatically documented via sphinx in [this section](http://girder.readthedocs.org/en/latest/python-client.html#module-girder_client).

mgrauer:
LGTM.

>I think that method should be deprecated

I'm good with that, if you want to tag it somehow, or add a comment.

PR #1313, opened 2016-04-18
Expand icon set

zachmullen:


zachmullen:
This change isn't as big as it looks, fontello just changed their white space in the config file dump. Best to look at the changes with whitespace comparison turned off.

msmolens:
LGTM

PR #1315, opened 2016-04-19
When updating jobs, do not include kwargs in notifications

zachmullen:
@girder/developers this fixes a bug I just ran into, can someone PTAL real quick?


manthey:
LGTM.

PR #1322, opened 2016-04-24
Fix several bugs associated with data copying

### LINKED ISSUE #1318: Reported item size of copied items is doubled.
Rough steps to reproduce:

1. import large file in a folder
2. create another folder using web UI
3. cherry pick first folder
4. copy first folder to the 2nd folder

zachmullen:
* Fixes #1318. When copying any file, the propagated size was
  doubled in the destination tree.
* The endpoints for copying always returned the new documents
  showing size "0" because we didn't reload the document after
  the copy operation.


zachmullen:
@manthey could you take a look at these bugfixes to make sure I haven't done anything silly?

manthey:
LGTM.

Issue #1319, opened 2016-04-22
Plugins loaded via entry points fail if they depend on other plugins


manthey:
When we load a plugin via an entry point, the plugin is imported as part of `findAllPlugins`.  Unfortunately, if the plugin itself tries to import another girder plugin, this will fail (as we haven't loaded the required plugins yet).  A plugin can be rewritten to defer these imports, but it would be better to fix this so that plugin dependencies are honored.

zachmullen:
Normally we handle this via toposorting an explicit dependency list; do we not have that ability for entrypoint plugins?

manthey:
In the toposort function, we call `findAllPlugins` which is doing the import (in an effort to get the config information needed for the toposort).

manthey:
There are other issues with using a plugin via an entry point.  Sub-imports within the plugin behave very differently when imported via an entry point and via a directory.

For instance, models aren't imported, even if they exist in the models module of the plugin.  They throw an ImportError, since girder.plugins.(plugin).models is NOT in the sys.modules path.

I think we *should* set the plugin's `PLUGIN_ROOT_DIR`.

We need to add the plugin to the `girder.plugins` module via `girder.plugins.__dict__[name] = module`, so that other plugins need not care if the plugin came via an entry point or a directory. 

We need the `sys.modules` dictionary to be away that the plugin is reachable this way as well (via `sys.modules[moduleName] = module`).  I don't think this has any repercussions, but I don't really like fiddling sys.modules directly.

danlamanna:
@jbeezley these issues should be fixed from pip-ip right?

jbeezley:
Yes, no longer relevant.

PR #1324, opened 2016-04-25
Fix some namespace issues when importing a plugin via entry_point.

manthey:
When importing a module by entry point, make sure it is available via girder.plugins[modulename] so that its submodules are always reachable in a consistent fashion.

Read a plugin.json or plugin.yml file from the packaged plugin rather than requiring this information to be duplicated at the entry point with the @plugin decorator.  This expects the plugins file within the module to be a girder/plugin.json or girder/plugin.yml.

Note that plugins must not import other plugins at the time that the module with the entry point is imported, as checking the @plugin decorator OR reading module files (such as plugin.json) import the module BEFORE module dependencies have been sorted and loaded.

The most useful feature of this branch is making sure the plugin is in the appropriate spot of the sys.modules namespace so that it can import its own modules.  This isn't currently tested, as it will require installing and enabling a plugin that isn't in the plugins directory.

### DIFF
--- a/girder/utility/plugin_utilities.py
+++ b/girder/utility/plugin_utilities.py
@@ -37,6 +37,7 @@ import traceback
 import yaml
 import importlib
 
+import pkg_resources
 from pkg_resources import iter_entry_points
 
 from girder.constants import PACKAGE_DIR, ROOT_DIR, ROOT_PLUGINS_PACKAGE, \
@@ -188,9 +189,10 @@ def loadPlugin(name, root, appconf, apiRoot=None, curConfig=None):
         # Try to load the plugin as an entry_point
         for entry_point in iter_entry_points(group='girder.plugin', name=name):
             pluginLoadMethod = entry_point.load()
-            pluginDir = os.path.dirname(
-                importlib.import_module(entry_point.module_name).__file__
-            )
+            module = importlib.import_module(entry_point.module_name)
+            pluginDir = os.path.dirname(module.__file__)
+            module.PLUGIN_ROOT_DIR = pluginDir
+            girder.plugins.__dict__[name] = module
             isPluginDir = True
 
     if not os.path.exists(pluginDir):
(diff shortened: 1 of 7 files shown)


manthey:
This is related to issue #1319.  I'm not sure if that issue can actually be fixed, as it turns out that asking for module resources imports the module.  I think that that the only way to discover the dependencies of a plugin without importing it would be to flog the `extras` of the entry_point.

To add tests for importing modules within an entry_point accessed plugin would require installing such a plugin for the main tests.  Ideally, we'd always run girder in a virtualenv so that we wouldn't have to install a test plugin into the system girder installation (which we do on travis for all but the packaging tests).

zachmullen:
This LGTM, I'd like @jbeezley to give a sign-off on it though

girder/utility/plugin_utilities.py
```diff
@@ -327,10 +325,49 @@ def findAllPlugins(curConfig=None):
...
+                        entry_point.name, configYml) as conf:
+                    try:
+                        data = yaml.safe_load(conf)
+                    except yaml.YAMLError as e:
+                        print(
+                            TerminalColor.error(
+                                'ERROR: Plugin "%s": plugin.yml is not valid '
+                                'YAML.' % entry_point.name))
+                        print(e)
+        except ImportError:
```
jbeezley: Why is `ImportError` caught here?  I don't see anything in the try block that should raise it.

manthey: `pkg_resources.resource_exists` can raise it.  Specifically, it imports the package via __import__.

If the package tries to import something that is unavailable the import error is thrown.  It will probably also be thrown by entry_point.load(), but the error message could be different, and, if I remember, it seemed more lucid.

jbeezley: I see, you noted that in your second comment.  Sounds okay then.

jbeezley:
LGTM

PR #1326, opened 2016-04-26
Add constants for mounted dirs in docker tasks

zachmullen:
ping @cdeepakroy


cdeepakroy:
@zachmullen This works! Thanks!

PR #1328, opened 2016-04-29
API keys

### LINKED ISSUE #995: Consider user-facing explicit token management
@jcfr just brought up this issue -- it may be nice to allow users of programmatic clients to be able to create (auth) tokens explicitly, even controlling the duration, via a REST endpoint. This would ideally be able to be done *without* needing to provide your password, via either a new REST endpoint or via modification of the existing `GET user/authentication` route.

This could be achieved by allowing users to create an API key that could be a proxy for their password, much like we had in Midas. This could be exposed with a new kwarg in the python client, e.g.

```
client = GirderClient(...)
client.authenticate(apiKey='foo')
```

zachmullen:
fixes #995


zachmullen:
@girder/developers this is a large topic branch, and the testing effort is also going to be large, so before I write tests to assert this behavior, I'd like folks to take a look at this functionality and give the OK on the design and implementation. A good starting point would be to look at what I added to the user guide docs, since that's a pretty accurate description of the functionality.

ping @cryos @jcfr

clients/python/girder_client/cli.py
```diff
@@ -42,7 +43,11 @@ def __init__(self, username, password, dryrun, blacklist,
                               apiRoot=apiRoot, scheme=scheme, dryrun=dryrun,
                               blacklist=blacklist, apiUrl=apiUrl)
         interactive = password is None
-        self.authenticate(username, password, interactive=interactive)
+
```
jcfr: raise exception if `apiKey != None and (username != None or password!= None)` ?

zachmullen: Is it really an error though if someone passes multiple ways of authenticating?

clients/web/src/views/widgets/ApiKeyListWidget.js
```diff
@@ -0,0 +1,169 @@
+girder.TokenScopeFull = {
```
jcfr: Should token listed here be retrieved using an end point ?

zachmullen: Thinking about it more, I think that would be a better way to do it, so that plugins that want to add scopes only have to add stuff on the server side.

clients/python/girder_client/cli.py
jcfr: Or may be a warning to indicate which parameter are ignored ?

jcfr:
Awesome work @zachmullen  

This is inlined with what I was thinking. :smile: 

Are the js widget to manage token etc ... reusable outside of main girder ui ?

jcfr:
@zachmullen Thanks for the slick demo :8ball:

zachmullen:
> Are the js widget to manage token etc ... reusable outside of main girder ui ?

Yes, and this is generally true of any of Girder's widgets.

zachmullen:
The tests are written, this is now ready for final review.

zachmullen:
Any comments on this topic?

girder/constants.py
```diff
@@ -200,7 +200,76 @@ class TokenScope:
...
+    'view the value of any system setting.', admin=True)
+TokenScope.describeScope(
+    TokenScope.ASSETSTORES_READ, 'View assetstores', 'Allows clients to see '
+    'all assetstore information.', admin=True)
+TokenScope.describeScope(
+    TokenScope.PARTIAL_UPLOAD_READ, 'View unfinished uploads.',
+    'Allows clients to see all partial uploads.', admin=True)
+TokenScope.describeScope(
+    TokenScope.PARTIAL_UPLOAD_CLEAN, 'Remove unfinished uploads.',
+    'Allows clients to remove unfinished uploads.', admin=True)
```
jeffbaumes: It seems that `admin=True` here can be confused with the naming `TokenScope.DATA_ADMIN`, and in general the admin term should be reserved for administrator-account things. I wonder if you should call the token something like `TokenScope.DATA_FULL_ACCESS` so that developers and users aren't confused seeing the word "administration" on something that does not require an administrative account.

jeffbaumes:
Is there a way for the checkbox text to look disabled (gray) when "Allow all" is selected? Currently there is only a slight shift and visually it looks like you can click the checkboxes anytime.

screen shot 2016-05-12 at 8 10 47 am

jeffbaumes:
Other than my naming and UI suggestions, everything seems to be working very well. I created, edited, disabled, enabled, deleted a key and girder_cli worked correctly in all cases. Great work.

zachmullen:
I made the disabled checkboxes more conspicuous and changed the name to DATA_OWN. These scopes are actually flat rather than hierarchical like the access levels, so you can theoretically give DATA_WRITE scope without granting DATA_READ scope, which is actually a feature, for instance if you want a third party CI service to be able to upload CI artifacts to girder, but not necessarily be able to read all the data you can access.

jeffbaumes:
LGTM! I agree that flat scopes are best, and that `DATA_FULL_ACCESS` was not the right term.

PR #1331, opened 2016-05-03
Resolves bug by explicitly defining argspect type

kotfic:
https://github.com/ansible/ansible/commit/627658 introduces explicity
defaulting module paramaters to strings.  This means we need to be
explicity about the type each function expects.  In this case default to
'dict' and only switch to list if vararg is present (e.g. like in
plugins()).

This is backward compatible but should future proof us against when 627658 is released.


zachmullen:
LGTM

PR #1333, opened 2016-05-06
Fix showing 'OFF' on disabled plugins.

manthey:
Rather than hide the word 'OFF', fix the CSS.


clients/web/src/views/body/PluginsView.js
```diff
@@ -57,7 +57,7 @@ girder.views.PluginsView = girder.View.extend({
 
         var view = this;
         this.$('.g-plugin-switch').bootstrapSwitch({
-            offText: '&nbsp;'
+            offText: 'OFF'
```
zachmullen: I think this is the default, so you can probably remove the line altogether.

zachmullen:
LGTM

PR #1334, opened 2016-05-06
Add apiUrl, patch and apiKey support to girder ansible module

kotfic:


zachmullen:
We may want to wait and batch in one more change on this topic, #1328 adds the API key authentication option to the client that we will definitely want to support in the ansible module.

kotfic:
@zachmullen should be a minor change,  i'm happy to wait on this until https://github.com/girder/girder/pull/1328 is merged

zachmullen:
The other branch is now merged, so we can add API key authentication on this topic.

kotfic:
@zachmullen PTAL,  

Note:  the big line +/- line count is because this also removes gobig roles in favor of the girder-ansible-role for provisioning the test VM via ansible-galaxy

zachmullen:
LGTM

PR #1340, opened 2016-05-12
Add license field to Item

### LINKED ISSUE #1320: Add dedicated metadata field for license
It could be nice to support a dedicated metadata field for an item's license. The field could be set at the time of upload/creation, potentially selected from a list of standard open source licenses.

msmolens:
Items gain a 'license' field.

In the web client, users can select a license from a dropdown when uploading an item, creating an item, or editing an item.

The list of licenses is a site-wide setting. The default licenses include the some of the "popular" open source licenses listed on https://opensource.org/licenses/category, the Creative Commons licenses listed on http://creativecommons.org/licenses/, as well as CC0, Public Domain, and All Rights Reserved.

Fixes #1320

### DIFF
--- a/clients/web/src/models/ItemModel.js
+++ b/clients/web/src/models/ItemModel.js
@@ -39,6 +39,19 @@ girder.models.ItemModel = girder.Model.extend({
         }, this)).error(_.bind(function (err) {
             this.trigger('g:error', err);
         }, this));
+    },
+
+    /**
+     * Get the list of licenses.
+     */
+    getLicenses: function (callback) {
+        girder.restRequest({
+            path: this.resourceName + '/licenses'
+        }).done(_.bind(function (resp) {
+            callback(resp);
+        }, this)).error(_.bind(function (err) {
+            this.trigger('g:error', err);
+        }, this));
     }
 });
 
--- a/clients/web/src/stylesheets/body/systemConfig.styl
+++ b/clients/web/src/stylesheets/body/systemConfig.styl
@@ -6,6 +6,11 @@
   margin-bottom 3px
   height 120px
 
+#g-core-licenses
+  font-family monospace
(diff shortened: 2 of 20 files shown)


zachmullen:
This is great. I know at some point somebody is going to ask for the feature of recursively setting a license, but that's something we can add later on.

Could you merge latest master into this so it will merge cleanly?

zachmullen:
@girder/developers PTAL as this is a widely useful feature. I've reviewed the code and functionality and I approve of it, but I'd like others to weigh in also.

cpatrick:
I like this and it's a great feature. The code looks good.

That said, I vote this be a built-in plugin rather than a core feature. A lot of use cases don't care about data licensing.

zachmullen:
> A lot of use cases don't care about data licensing.

That's a good point. If we don't want to go to the trouble of refactoring all this into a plugin, I'd be fine with the compromise of a system setting checkbox to enable or disable their visibility.

jeffbaumes:
> I'd be fine with the compromise of a system setting checkbox to enable or disable their visibility.

Seems an ok strategy to me. Should be defaulted to off to have the UI uncluttered by default.

brianhelba:
@msmolens How much additional work would it be to refactor this into a plugin, as opposed to all the updates you'll have to make anyway so that all the logic checks the setting value?

I feel bad, because I was the one who asked for this feature to be in core, but after looking at the code, I agree with @cpatrick that this would objectively be better as a plugin.

msmolens:
@brianhelba I agree this would be better as a plugin. If the system fully supports integrating these types of changes into the core web client, I don't expect the conversion would take long.

msmolens:
Superseded by https://github.com/girder/girder/pull/1347

PR #1344, opened 2016-05-17
Fix some CTest issues

### LINKED ISSUE #1336: smtpd.SMTPServer DeprecationWarning with Python 3
Running the `server_mail` test shows a `DeprecationWarning` with Python 3:

    $ python --version
    Python 3.5.1
    $ ctest -R server_mail -V
    ...
    /path/to/girder/tests/mock_smtp.py:57: DeprecationWarning: The decode_data default of True will change to False in 3.6; specify an explicit value for this keyword
      self.smtp = MockSmtpServer(self.address, None)

From https://docs.python.org/3/library/smtpd.html#smtpserver-objects:

> decode_data specifies whether the data portion of the SMTP transaction should be decoded using UTF-8. The default is True for backward compatibility reasons, but will change to False in Python 3.6; specify the keyword value explicitly to avoid the DeprecationWarning.

brianhelba:
* Make CTest fail gracefully if eslint is not available
* Provision Vagrant with Python 2.7.11
 * This fixes an issue where the Girder package fails to build on Virtualbox.
* Silence a test output warning in Python 3.5
 * The "decode_data" parameter to SMTPServer should be explicitly provided
in Python 3.5+.

Fixes #1336

### DIFF
--- a/tests/mock_smtp.py
+++ b/tests/mock_smtp.py
@@ -21,6 +21,7 @@ import asyncore
 import email
 import os
 import smtpd
+import sys
 import threading
 import time
 
@@ -32,6 +33,16 @@ _maxTries = 100
 
 class MockSmtpServer(smtpd.SMTPServer):
     mailQueue = queue.Queue()
+
+    def __init__(self, localaddr, remoteaddr, decode_data=False):
+        kwargs = {}
+        if sys.version_info >= (3, 5):
+            # Python 3.5+ prints a warning if 'decode_data' isn't explicitly
+            # specified, but earlier versions don't accept the argument at all
+            kwargs['decode_data'] = decode_data
+        # smtpd.SMTPServer is an old-style class in Python2,
+        # so super() can't be used
+        smtpd.SMTPServer.__init__(self, localaddr, remoteaddr, **kwargs)
 
     def process_message(self, peer, mailfrom, rcpttos, data):
         self.mailQueue.put(data)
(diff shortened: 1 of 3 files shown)


tests/mock_smtp.py
```diff
@@ -33,6 +34,12 @@
 class MockSmtpServer(smtpd.SMTPServer):
     mailQueue = queue.Queue()
 
+    def __init__(self, localaddr, remoteaddr, decode_data=False):
+        kwargs = {}
+        if sys.version_info[0:2] >= (3, 5):
+            kwargs['decode_data'] = decode_data
```
zachmullen: Let's add an inline comment to explain this.

tests/mock_smtp.py
```diff
@@ -33,6 +34,14 @@
 class MockSmtpServer(smtpd.SMTPServer):
     mailQueue = queue.Queue()
 
+    def __init__(self, localaddr, remoteaddr, decode_data=False):
+        kwargs = {}
+        if sys.version_info[0:2] >= (3, 5):
```
zachmullen: Python trivia time, you can actually just do `if sys.version_info >= (3, 5):`

brianhelba: Mind :boom:.

Actually, that makes perfect sense. I just copied the code that `six` uses; someone should tell them too.

brianhelba:
@girder/developers PTAL.

zachmullen:
LGTM

PR #1345, opened 2016-05-17
Force newer node version in travis

zachmullen:


zachmullen:
Latest stable version of node (6.0.0) causes a failure since it apparently requires a C++11 compiler for building contextify, so I pegged to the version I have on my local machine for now.

zachmullen:
@jbeezley PTAL

jbeezley:
:shipit:

brianhelba:
It's good that we got this fixed ASAP, but I was just about to suggest that we switch to using Node 4.4.4, since that's the latest LTS release, and what our Vagrant environment provisions.

What do you think?

zachmullen:
Sure, feel free to open a topic to move to that, as long as travis is green that's fine. I just wasn't sure when the C++11 dependency was introduced for contextify.

PR #1347, opened 2016-05-17
Add item licenses plugin

### LINKED ISSUE #1320: Add dedicated metadata field for license

msmolens:
When the item licenses plugin is enabled, items gain a 'license' field.

In the web client, users can select a license from a dropdown when uploading an
item, creating an item, or editing an item.

The list of licenses is configurable in the plugin settings. The default
licenses include the some of the "popular" open source licenses listed on
https://opensource.org/licenses/category, the Creative Commons licenses listed
on http://creativecommons.org/licenses/, as well as CC0, Public Domain, and All
Rights Reserved.

Fixes #1320


plugins/item_licenses/plugin.cmake
```diff
@@ -0,0 +1,30 @@
...
+add_python_style_test(python_static_analysis_item_licenses
+                      "${PROJECT_SOURCE_DIR}/plugins/item_licenses/server")
+add_python_style_test(python_static_analysis_item_licenses_tests
+                      "${PROJECT_SOURCE_DIR}/plugins/item_licenses/plugin_tests")
+
+# XXX: add web client tests
+# add_web_client_test(
+#    item_licenses
+#    "${PROJECT_SOURCE_DIR}/plugins/item_licenses/plugin_tests/itemLicensesSpec.js"
+#    PLUGIN item_licenses)
```
cpatrick: ಠ_ಠ

cpatrick:
Needs tests on the web client side (as noted by ಠ_ಠ).

plugins/item_licenses/plugin.cmake
msmolens: Ah, "ಠ_ಠ", the universal symbol for missing web client tests!

msmolens:
@cpatrick I added web client tests

plugins/item_licenses/plugin.cmake
```diff
@@ -0,0 +1,29 @@
...
+add_python_style_test(python_static_analysis_item_licenses
+                      "${PROJECT_SOURCE_DIR}/plugins/item_licenses/server")
+add_python_style_test(python_static_analysis_item_licenses_tests
+                      "${PROJECT_SOURCE_DIR}/plugins/item_licenses/plugin_tests")
+
+add_web_client_test(
+   item_licenses
+   "${PROJECT_SOURCE_DIR}/plugins/item_licenses/plugin_tests/itemLicensesSpec.js"
+   PLUGIN item_licenses)
+add_eslint_test(
```
brianhelba: You should also `eslint` the `itemLicensesSpec.js`.

plugins/item_licenses/web_client/js/setup.js
```diff
@@ -0,0 +1,162 @@
...
+    render.call(this);
+
+    return this;
+});
+
+/**
+ * Allow selecting license when uploading an item.
+ */
+girder.wrap(girder.views.HierarchyWidget, 'uploadDialog', function (uploadDialog) {
+    girder.restRequest({
```
brianhelba: This means that the pop-up and rendering of the upload dialog is completely blocked until this REST request resolves.

Ideally, I'd hope to see the dialog render immediately, and the dropdown with license options get populated asynchronously.

brianhelba: This can probably wait until a v0.2.0 though.

plugins/item_licenses/web_client/js/ConfigView.js
```diff
@@ -0,0 +1,84 @@
...
+            event.preventDefault();
+
+            this.$('#g-item-licenses-error-message').empty();
+
+            this._saveSettings([{
+                key: 'item_licenses.licenses',
+                value: this.$('#g-item-licenses').val().trim()
+            }]);
+        },
+        'click #g-item-licenses-defaults': function (event) {
```
brianhelba: This is a nice feature.

plugins/item_licenses/web_client/templates/item_licenses_config.jade
```diff
@@ -0,0 +1,11 @@
+.g-config-breadcrumb-container
+p.g-item-licenses-description
+  | Configure the available licenses for items.
+form#g-item-licenses-settings-form(role="form")
+  .form-group
+    label.control-label(for="g-item-licenses") Licenses
+    textarea#g-item-licenses.form-control= licenses
```
brianhelba: If this plugin becomes popular, we'll want to make this a more user-friendly series of widgets, but this is fine for now.

plugins/item_licenses/server/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+def updateItemLicense(event):
+    """
+    REST event handler to update item with license parameter, if provided.
+    """
+    params = event.info['params']
+    if 'license' in params:
+        itemModel = ModelImporter.model('item')
+        item = itemModel.load(event.info['returnVal']['_id'], force=True,
+                              exc=True)
+        item['license'] = params['license']
```
brianhelba: It's possible to add a quick extra optimization here, and only save if the license on the item is different from what we're updating it to.

brianhelba: Nevermind, I see what you're doing with `preventDefault`.

brianhelba: Oops, `preventDefault` is just to make `handleRoute` return the updated item. My original optimization suggestion still stands.

plugins/item_licenses/server/constants.py
```diff
@@ -0,0 +1,98 @@
...
+                        'name': 'Mozilla Public License 2.0'
+                        }
+                    ]
+                },
+            {
+                # See http://creativecommons.org/licenses/
+                'category': 'Content Licenses',
+                'licenses': [
+                    {
+                        'name': 'CC0'
```
brianhelba: I think this should be `Creative Commons Public Domain Dedication (CC0)`, considering we're being verbose about the GPL licenses too.

plugins/item_licenses/server/constants.py
```diff
@@ -0,0 +1,98 @@
...
+                        },
+                    {
+                        'name': 'CC Attribution (BY)'
```
brianhelba: This should be `Creative Commons Attribution (CC BY)`.

plugins/item_licenses/server/constants.py
```diff
@@ -0,0 +1,98 @@
...
+                    {
+                        'name': 'CC Attribution-NonCommercial-ShareAlike '
+                        '(BY-NC-SA)'
+                        },
+                    {
+                        'name': 'CC Attribution-NonCommercial-NoDerivs '
+                        '(BY-NC-ND)'
+                        },
+                    {
+                        'name': 'Public Domain'
```
brianhelba: I wonder if we should try to name this differently to prevent people from misusing it. In the US, you cannot affirmatively just choose to place something into the public domain. I'm worried people will say "I want to share this with everyone" and chose this license, when in fact, it provides no legal basis for sharing.

plugins/item_licenses/server/constants.py
```diff
@@ -0,0 +1,98 @@
...
+                'category': 'Content Licenses',
+                'licenses': [
+                    {
+                        'name': 'CC0'
+                        },
+                    {
+                        'name': 'CC Attribution (BY)'
+                        },
+                    {
+                        'name': 'CC Attribution-ShareAlike (BY-SA)'
```
brianhelba: The same for the rest of the CC licenses.

plugins/item_licenses/server/__init__.py
brianhelba: Additionally, we're not enforcing that the `params['license']` (which could be arbitrary input from any client) is actually a member of the allowed `PluginSettings.LICENSES`. I think we should add a check for that.

brianhelba: The check that the license is one of the enumerated values would normally appear in `validateItem`, but that's going to slow down every future item save with an extra database lookup for the settings. I'd vote to put it here on the REST handler for efficiency.

plugins/item_licenses/server/constants.py
msmolens: Good catch. Perhaps we should just remove it for now to avoid confusion?

plugins/item_licenses/web_client/js/setup.js
msmolens: Yes, that would be better.

plugins/item_licenses/web_client/templates/item_licenses_config.jade
msmolens: Agreed, editing these options in a naive textarea is unwieldy.

plugins/item_licenses/server/constants.py
brianhelba: Yeah. Datasets produced by the US Government are public domain, and so is anything for which the copyright term has lapsed. Since it's much harder to rename or remove licences without compatibility breaking, maybe we should drop it until we can reach a consensus.

Any thoughts from others about how to name "Public Domain" while letting people know that what they *really* want is "CC0"?

zachmullen: What about http://opendatacommons.org/licenses/pddl/ ?

brianhelba: I'd say we should add [all 3 Open Data Commons licenses](http://opendatacommons.org/licenses/) to the list.

The PDDL looks to be attempting the same thing as the CC0. We just need to find a way to encourage users to mark their content as either of those, and to avoid inappropriately using "Public Domain" unless they really understand what it actually indicates.

Of course, from the perspective some data store software, we shouldn't care about any of these semantics. As advocates of open source, though, I think we should do our part to ensure content is actually made free for as many people as possible.

manthey: Perhaps next to CC0 or PDDL we could have additional text reading something
like "Use this to waive copyrights and allow full public use", and,
similarly, additional text next to Public Domain reading something like
"Only for works already in the public domain".

On Wed, May 18, 2016 at 3:47 PM, Brian Helba <notifications@github.com>
wrote:

> In plugins/item_licenses/server/constants.py
> <https://github.com/girder/girder/pull/1347#discussion_r63769235>:
>
> > +                    {
> > +                        'name': 'CC Attribution-NoDerivs (BY-ND)'
> > +                        },
> > +                    {
> > +                        'name': 'CC Attribution-NonCommercial (BY-NC)'
> > +                        },
> > +                    {
> > +                        'name': 'CC Attribution-NonCommercial-ShareAlike '
> > +                        '(BY-NC-SA)'
> > +                        },
> > +                    {
> > +                        'name': 'CC Attribution-NonCommercial-NoDerivs '
> > +                        '(BY-NC-ND)'
> > +                        },
> > +                    {
> > +                        'name': 'Public Domain'
>
> I'd say we should add all 3 Open Data Commons licenses
> <http://opendatacommons.org/licenses/> to the list.
>
> The PDDL looks to be attempting the same thing as the CC0. We just need to
> find a way to encourage users to mark their content as either of those, and
> to avoid inappropriately using "Public Domain" unless they really
> understand what it actually indicates.
>
> Of course, from the perspective some data store software, we shouldn't
> care about any of these semantics. As advocates of open source, though, I
> think we should do our part to ensure content is actually made free for as
> many people as possible.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/girder/girder/pull/1347/files/687eaa599b413848381c3b3fbc14b43db4b12292#r63769235>
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

plugins/item_licenses/server/__init__.py
```diff
@@ -39,22 +39,46 @@ def validateString(value):
...
+
+    # Ensure that new license name is in configured list of licenses.
+    #
+    # Enforcing this here, instead of when validating the item, avoids an extra
+    # database lookup (for the settings) on every future item save.
+    if newLicense:
+        validLicense = False
+        licenseSetting = ModelImporter.model('setting').get(
+            PluginSettings.LICENSES)
+        for group in licenseSetting:
```
brianhelba: Try:
```
validLicense = any(
    license['name'] == newLicense
    for group in licenseSetting
    for license in group['licenses']
)
```
instead.

plugins/item_licenses/server/constants.py
cpatrick: The CC licenses are vastly more popular that the ODC ones--and better understood by most folks. We should encourage CC0 for people who think they want "Public Domain".

brianhelba:
All code LGTM.

@msmolens and I discussed using the license names from https://github.com/ufal/public-license-selector , since we may eventually want to add that widget. After those names are changed, I think it's good to merge.

msmolens:
@brianhelba I resolved a merge conflict then squashed the commits. This should be ready to go.

plugins/item_licenses/plugin_tests/itemLicensesSpec.js
```diff
@@ -0,0 +1,425 @@
...
+            $('a.g-create-item:visible').click();
+        });
+        girderTest.waitForLoad();
+
+        waitsFor(function () {
+            return Backbone.history.fragment.slice(-18) === '?dialog=itemcreate';
+        }, 'the URL state to change');
+        waitsFor(function () {
+            return $('button.g-save-item:visible').text() === 'Create';
+        }, 'the create item dialog to appear');
```
zachmullen: Could we just do a `girderTest.waitForDialog` here?

plugins/item_licenses/plugin_tests/itemLicensesSpec.js
```diff
@@ -0,0 +1,425 @@
...
+        waitsFor(function () {
+            return Backbone.history.fragment.slice(-16) === '?dialog=itemedit';
+        }, 'the url state to change');
+        girderTest.waitForDialog();
+        waitsFor(function () {
+            return $('#g-name').val() !== '';
+        }, 'the dialog to be populated');
+        waitsFor(function () {
+            return $('button.g-save-item:visible').text() === 'Save';
+        }, 'the edit item dialog to appear');
```
zachmullen: Multiple consecutive `waitsFor` calls could be combined into a single one.

plugins/item_licenses/server/__init__.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    Make sure a value is a string.
+
+    :param value: the value to coerce into a string if it isn't already.
+    :returns: the string version of the value.
+    """
+    if value is None:
+        value = ''
+    if not isinstance(value, six.string_types):
+        value = str(value)
```
zachmullen: `str` in python2 is roughly equivalent to `bytes` in python3, and `str` in python3 is roughly equivalent to `unicode` in python2. This code will probably work in either version, but we may want to clarify the docstring and the return value to be either always an encoded string or always a text string (see `six.text_type`).

brianhelba: Just make this check for and coerce to `six.text_type`. It should always be `str` in Python3, and making it always `unicode` in Python2 should be harmless.

plugins/item_licenses/plugin_tests/itemLicensesSpec.js
msmolens: The `girderTest.waitForLoad` on line 65 is incorrect, and yes there should be a `girderTest.waitForDialog` after waiting for the URL state change.

That said, are you recommending replacing both of these waitFor() blocks with `girderTest.waitForDialog`? Currently, after the changes mentioned above, this is very similar to what's done here: https://github.com/girder/girder/blob/f6f259e3e6c999542cc5e134da1200eb4b71ff21/clients/web/test/spec/itemSpec.js#L28-L39

plugins/item_licenses/server/__init__.py
msmolens: Makes sense; I'll make the change. Also note that this function is similar to one used in the initial PR that implemented item licenses in Girder core: https://github.com/girder/girder/blob/f6f259e3e6c999542cc5e134da1200eb4b71ff21/girder/models/item.py#L68-L79

zachmullen:
All good by me, I'll let @brianhelba give the final pass.

zachmullen:
@msmolens could you merge latest master into this?

zachmullen:
@msmolens I just tested this and it works well, and the data format will lend itself to improving the admin config UI without breaking the data spec, so LGTM! If you approve my small change, go ahead and merge.

msmolens:
Thanks @zachmullen, your changes look good. I'd still like to defer to @brianhelba before merging this.

Issue #803, opened 2015-04-21
girder_client: token kwarg for `authenticate()`


zachmullen:
You can sort of do this already by calling `client.token = ...`, but it would be better if we either expose a `token` kwarg on `authenticate()` that could be used instead of username and password, or simply create an alternative method called `authenticateWithToken()`.

jcfr:
:+1:  Where can I get the token from the current girder UI ?

zachmullen:
Closing this for now, `client.token = ...` is relatively pythonic and not so bad really.

danlamanna:
This behavior took me off guard (only for a moment)... but talking with others I found out it confused them as well.

Not that `client.token` is difficult, but being totally undocumented and not part of the authentication function (which we have a level of compatibility for) is counter intuitive.

zachmullen:
Setting the token directly is a bit of an edge case for usage of girder_client. Doing so means you've somehow used a different client to obtain a token. The prescribed ways of authentication for the client are to use a username/password or an API key. What is the use case for wanting to create the token yourself?

danlamanna:
When I schedule jobs I create a token with the specific scope needed for the job, and pass that along. So the job uses the token when it uses Girder client.

zachmullen:
Fair enough. I will reopen this, but it's quite an old issue -- does someone have time to actually add this in soon?

PR #1349, opened 2016-05-19
Add modal showing current user's token

Xarthisius:
Provide convenient shortcut in user menu for obtaining authentication token. This addresses the 2nd [comment](https://github.com/girder/girder/issues/803#issuecomment-139577988) on #803


brianhelba:
Is this something we even want users to use?

We might be better to encourage them to use API keys, and improve that interface if necessary.

Xarthisius:
To be honest I completely missed pull request 1328 while writing this.

PR #1350, opened 2016-05-20
Add optional json parameter to client

cjh1:


zachmullen:
Do we have a test that makes sure the json argument works?

cjh1:
Yes, I added [these]( https://github.com/girder/girder/blob/2a192805711bff3661be010fc71217d06204290b/tests/cases/py_client/lib_test.py#L361-L378), I changed the implementation of these to use the json parameter.

zachmullen:
Ah cool. LGTM!

PR #1352, opened 2016-05-21
Add a utility method to support stream-reading request bodies

zachmullen:
This method is particularly useful for cases when the client
is sending the request using Transfer-Encoding: chunked, though
it is also valuable for normal Content-Length requests insofar
as it abstracts cherrypy functions in addition to the transfer
method.

This is the "final" integration step related to the functionality added in https://github.com/girder/girder_worker/pull/36. This fulfills our desired use case of a docker container writing a chunk to a file within the container that happens to be a named pipe, and receiving that chunk in Girder while the container is still running, so that we can display real-time streaming output from containers that are potentially generating a large amount of data.


tests/cases/stream_test.py
```diff
@@ -0,0 +1,104 @@
...
+
+class StreamTestResource(Resource):
+    def __init__(self):
+        super(StreamTestResource, self).__init__()
+        self.resourceName = 'stream_test'
+        self.route('POST', ('input_stream',), self.inputStream)
+
+    @access.public
+    def inputStream(self, params):
+        for chunk in iterBody(8):  # Read body 8 bytes at a time
```
cpatrick: why 8?

tests/cases/stream_test.py
```diff
@@ -0,0 +1,104 @@
...
+            a separate chunk.
+            """
+            for i in range(1, 4):
+                buf = 'chunk%d' % i
+                yield buf
+                start = time.time()
+                while True:  # Wait for server thread to get the chunk
+                    if len(_chunks) == i:
+                        break
+                    if time.time() - start > 5:
```
cpatrick: why 5?

girder/api/rest.py
```diff
@@ -73,6 +75,36 @@ def getApiUrl(url=None):
...
+        cherrypy.request.rfile.bufsize = length
+        while True:
+            cherrypy.request.rfile._fetch()
+            if cherrypy.request.rfile.closed:
+                break
+            buf = cherrypy.request.rfile.buffer
+            cherrypy.request.rfile.buffer = ''
+            yield buf
+    elif 'Content-Length' in cherrypy.request.headers:
+        while True:
```
cpatrick: Why not while buf?

girder/api/rest.py
```diff
@@ -34,6 +34,8 @@
 from girder.utility import config, JsonEncoder
 from six.moves import range, urllib
 
+READ_BUFFER_LEN = 65536
```
cpatrick: why 65536?

tests/cases/stream_test.py
zachmullen: Why not 5? Seems like a reasonable timeout in which another thread should really have done something.

tests/cases/stream_test.py
zachmullen: Easy number to demonstrate that we are indeed receiving the correct chunk lengths. This number is essentially test fixture data, it correlates to our other arbitrary fixture data which is the dummy chunks we are sending in each case.

girder/api/rest.py
zachmullen: Why Patrick instead of Pat? :wink:

cpatrick: :D

cpatrick: You know me, I just like comments that say a little bit about why. Even if the answer is: "This is arbitrary."

girder/api/rest.py
zachmullen: Because python does not have a `do-while` looping construct :(

cpatrick: This is preferred over

```
buf = cherrypy.request.body.read(length)
while buf:
  yield buf
  buf = cherrypy.request.body.read(length)
```
? I'm not sure which I really prefer, but I think I tend toward the latter.

zachmullen: I don't have strong feelings either way, though I'm also not fond of seeing `cherrypy.request.body.read(length)` duplicated.

cpatrick: Yeah. Anyway, ignore this critique.

zachmullen:
This is now working in both python 2 and 3, and we have tests that will fail if the functionality breaks in the future, so I think it's ready for review.

tests/cases/stream_test.py
```diff
@@ -49,7 +49,7 @@ def __init__(self):
     @access.public
     def inputStream(self, params):
         # Read body 8 bytes at a time so we can test chunking a small body
-        for chunk in iterBody(8):
```
cpatrick: now the comment doesn't agree with the code...

zachmullen: This is the main substantive point I want to discuss about this feature. The underlying cherrypy API gives us two options for reading these chunked TE requests, and I'm not sure which one makes for a better API, or if we really need to expose both. Hoping folks can weigh in.

1. Block until the last byte of the current chunk is read. This allows clients to (perhaps) intellegently decide what should be encoded in each chunk, which could be useful if we expect significant delays between discrete documents being sent over the socket. The drawback here is that the server side logic no longer has any control over the read buffer size, and also we have to make use of a private method in cherrypy, although I'm not that worried about using the private method since our testing is sufficient.
2. Block on read until a certain number of bytes (decided by the server logic) is available. The drawback here is that we will block until that number of bytes is available, which may be a larger buffer than the chunk size being sent by the client, meaning we might be sitting there waiting while we actually have discrete documents we could be processing.

If we choose to support both options, we just have to have a very long docstring describing all of these behaviors. I want someone to be able to call the `iterBody` method regardless of whether the client sent the body using TE: chunked or a normal known-length content body.

I guess a third option is to try and dig further down to see if we could add a timeout to the blocking read, but from my browsing cherrypy internals for a while, it didn't look simple.

girder/api/rest.py
```diff
@@ -76,7 +76,7 @@ def getApiUrl(url=None):
     return url[:idx + 7]
 
 
-def iterBody(length=READ_BUFFER_LEN):
+def iterBody(length=READ_BUFFER_LEN, strictLength=False):
```
mgrauer: docstring for `strictLength`

girder/api/rest.py
```diff
@@ -73,6 +76,51 @@ def getApiUrl(url=None):
...
+    function works regardless of whether the body was sent with a Content-Length
+    or using Transfer-Encoding: chunked, but the behavior is slightly different
+    in each case.
+
+    If `Content-Length` is provided, the `length` parameter is used to read the
+    body in chunks up to size `length`. This will block until end of stream or
+    the specified number of bytes is ready.
+
+    If `Transfer-Encoding: chunked` is used, the `length` parameter is ignored
+    by default the generator yields each chunk that is sent in the request,
```
mgrauer: this phrasing is a bit choppy, maybe `parameter is ignored by default; the generator` or `parameter is ignored by default.  The generator` ?

tests/cases/stream_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+class StreamTestResource(Resource):
+    def __init__(self):
+        super(StreamTestResource, self).__init__()
+        self.resourceName = 'stream_test'
+        self.route('POST', ('input_stream',), self.inputStream)
+
+    @access.public
+    def inputStream(self, params):
+        # Read body 8 bytes at a time so we can test chunking a small body
```
mgrauer: 8 or 5?

tests/cases/stream_test.py
```diff
@@ -0,0 +1,121 @@
...
+        This exercises the behavior of iterBody in the case of requests with
+        Content-Length passed.
+        """
+        resp = requests.post(self.apiUrl + '/stream_test/input_stream',
+                             data='hello world')
+        resp.raise_for_status()
+        self.assertEqual(resp.json(), ['hello', ' worl', 'd'])
+
+    def testStrictLength(self):
+        resp = requests.post(
```
mgrauer: add a docstring, and move this above ` testKnownLengthBodyReady` so that both chunked tests are together

mgrauer:
LGTM?TH

jeffbaumes:
I think you mean

TH?LGTM:👎

PR #1355, opened 2016-05-23
Fix default parameter value for json

cjh1:


cjh1:
@zachmullen Just noticed this!

zachmullen:
Oops, wonder how that didn't break our tests?

cjh1:
The test uses PUT but not PATCH, I can add a test?

zachmullen:
Ah, I see. We never really use PATCH in girder so that makes sense. I'll cut a point release of the client once this is merged.

cjh1:
Do you want me to add a test?

zachmullen:
I guess if we are going to expose a `patch` method then we should test it in some way or another.

cjh1:
I will take a look, can probably use httpmock

cjh1:
@zachmullen PTAL I added a test for PATH

zachmullen:
LGTMITIH

PR #1357, opened 2016-05-25
Add configuration options for application startup

jbeezley:
This moves most of the `initialize` method into a new `start` method, which is called by default during initialization.  The new method support a few configuration flags that girder derived applications can use to customize the startup behavior.

### DIFF
--- a/clients/web/src/app.js
+++ b/clients/web/src/app.js
@@ -1,43 +1,82 @@
 girder.App = girder.View.extend({
-    initialize: function () {
-        girder.fetchCurrentUser()
-            .done(_.bind(function (user) {
-                girder.eventStream = new girder.EventStream({
-                    timeout: girder.sseTimeout || null
-                });
+    /**
+     * @param {object} [settings]
+     * @param {bool} [settings.start=true] Run start after initialization
+     */
+    initialize: function (settings) {
+        this._started = false;
+        settings = settings || {};
+        if (settings.start === undefined || settings.start) {
+            this.start();
+        }
+    },
 
-                this.headerView = new girder.views.LayoutHeaderView({
-                    parentView: this
-                });
+    /**
+     * Start the application with optional components.
+     * @param {object} [settings]
+     * @param {bool} [settings.fetch=true] Fetch the current user modal
+     * @param {bool} [settings.render=true] Render the layout after starting
+     * @param {bool} [settings.history=true] Start backbone's history api
+     * @returns {$.Deferred} A promise-like object that resolves when the app is ready
+     */
+    start: function (settings) {
+        // start is a noop if the app is already running
+        var promise = new $.Deferred().resolve(null).promise();
+        if (this._started) {
+            return promise;
+        }
+        this._started = true;
 
-                this.globalNavView = new girder.views.LayoutGlobalNavView({
-                    parentView: this
-                });
+        // set defaults
+        settings = _.defaults(settings || {}, {
+            fetch: true,
+            render: true,
+            history: true
+        });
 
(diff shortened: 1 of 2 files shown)


clients/web/src/app.js
```diff
@@ -1,43 +1,82 @@
...
-                });
+    /**
+     * @param {object} [settings]
+     * @param {bool} [settings.start=true] Run start after initialization
+     */
+    initialize: function (settings) {
+        this._started = false;
+        settings = settings || {};
+        if (settings.start === undefined || settings.start) {
+            this.start();
```
zachmullen: Don't we need to pass settings into start?

zachmullen:
Having run into difficulties myself trying to extend this prototype for custom apps, this is a nice improvement.

clients/web/src/app.js
jbeezley: Nope.  Line 31 has `settings || {}`.

clients/web/src/app.js
```diff
@@ -1,43 +1,82 @@
...
-                });
+            this._createLayout();
 
-                if (user) {
-                    girder.currentUser = new girder.models.UserModel(user);
-                    girder.eventStream.open();
-                }
+            if (user) {
+                girder.currentUser = new girder.models.UserModel(user);
+                girder.eventStream.open();
```
zachmullen: We should probably make creation of the eventStream optional while we're at it.

clients/web/src/app.js
```diff
@@ -1,43 +1,82 @@
...
+        // If fetching the user from the server then we return the jqxhr object
+        // from the request, otherwise just call the callback.
+        if (settings.fetch) {
+            promise = girder.fetchCurrentUser()
+                .done(afterFetch);
+        } else {
+            afterFetch(null);
+        }
+
+        this.bindGirderEvents();
```
zachmullen: Might want to be able to control some or all of these event bindings too.

zachmullen:
This may be an appropriate time to finally write some developer docs for "how to write a custom backbone app using girder's core javascript library"

clients/web/src/app.js
jbeezley: I thought about adding more options, but it gets dicey.  For example, if the event stream object is assumed to exist all over the place, including in the `_createLayout` function.  I think if you really want to do that, you will have to override the `start` method anyway.

clients/web/src/app.js
jbeezley: You can always just override the `bindGirderEvents` method.  I can see adding possibly a `bindEvents` setting, but I wouldn't want to be more granular than that.

clients/web/src/app.js
zachmullen: Ah, so in order to customize the start options, you have to set start: false and then just call start() yourself.

jbeezley: Yes, that's the idea.  I wanted to decouple the construction of the object from starting the application.  Start is only called here to retain backward compatibility.  From what I've seen, it's a fairly common practice to have an `app.start` method ([`Marionette.Application`](http://marionettejs.com/docs/v2.4.5/marionette.application.html) for example).

zachmullen:
Code looks good, could you just add a bit of developer docs describing the use of this API?

jbeezley:
PTAL.  I think what is really needed is a more complete example.  The content of `clients/web-external` is pretty minimal as far as examples go.  I think a stripped down version of Covalic would be more appropriate.  To really motivate why I added this would require a bigger departure from Girder's client.  I intend to extract pieces of what I'm working on for Emory as a boilerplate application integrating Girder with Marionette.

zachmullen:
LGTM

PR #1360, opened 2016-05-27
Python client added features

sutartmelson:
Modified downloadFile function to allow downloading to a file-like object rather than just downloading to a local directory. 

Added listFile function that is similar to listFolder and listItem. It simply lists all the files in a given item.

Added optional limit parameter to the aforementioned list functions.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -227,16 +227,12 @@ class GirderClient(object):
             Note that the path string should not begin or end with the path
             separator, '/'.
         :param parameters: A dictionary mapping strings to strings, to be used
-<<<<<<< HEAD
             as the key/value pairs in the request parameters.
-=======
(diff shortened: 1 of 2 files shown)


clients/python/girder_client/__init__.py
```diff
@@ -288,6 +288,21 @@ def listResource(self, path, params):
         """
         return self.get(path, params)
 
+    def listFiles(self, itemId, limit=None):
```
zachmullen: Let's call this `listFile` so it agrees with the naming of the other methods.

zachmullen:
One last thing, could you add a test for your new method? [This file](https://github.com/girder/girder/blob/master/tests/cases/py_client/lib_test.py) contains our testing of the python client library.

sutartmelson:
I've written tests for all of the code I wrote/altered.

zachmullen:
LGTM

PR #1361, opened 2016-05-30
Zip download mimetype filter

zachmullen:
@brianhelba PTAL

### DIFF
--- a/girder/models/item.py
+++ b/girder/models/item.py
@@ -397,7 +397,7 @@ class Item(acl_mixin.AccessControlMixin, Model):
         return newItem
 
     def fileList(self, doc, user=None, path='', includeMetadata=False,
-                 subpath=True):
+                 subpath=True, mimeFilter=None):
         """
         Generate a list of files within this item.
 
@@ -417,6 +417,9 @@ class Item(acl_mixin.AccessControlMixin, Model):
                         metadata, or the sole file is not named the same as the
                         item, then the returned paths include the item name.
         :type subpath: bool
+        :param mimeFilter: Optional list of MIME types to filter by. Set to
+            None to include all files.
+        :type mimeFilter: list or tuple
         :returns: Iterable over files in this item, where each element is a
                   tuple of (path name of the file, stream function with file
                   data).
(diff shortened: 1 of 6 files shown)


girder/models/item.py
```diff
@@ -438,6 +443,15 @@ def stream():
                 yield json.dumps(doc['meta'], default=str)
             yield (os.path.join(path, metadataFile), stream)
 
+    def _mimeFilter(self, file, mimeFilter):
+        """
+        Returns whether or not the given file should be passed through the given
+        MIME filter. If no MIME filter is specified, all files are allowed.
+        """
+        if not mimeFilter:
```
brianhelba: By using truth checking, instead of `mimeFilter is None`, an empty `mimeFilter` list will actually allow everything. Of course, the semantics of passing an empty `mimeFilter` list are a little unclear to begin with.

brianhelba:
While this excludes any non-matched files from being in the download, all of the now-empty items (with possible metadata) which contained those non-matched files are still included in the download. Is this the behavior most users would want or expect?

girder/models/item.py
zachmullen: This was by design, since it is assumed that nobody would request an empty set to download.

zachmullen:
> While this excludes any non-matched files from being in the download, all of the now-empty items (with possible metadata) which contained those non-matched files are still included in the download. Is this the behavior most users would want or expect?

Isn't that only true when the user explicitly requests to include metadata? Otherwise, files are the only thing that actually cause data to be written into the archive.

brianhelba:
@zachmullen Yup, you're right, I forgot that metadata is not included by default.

girder/models/item.py
brianhelba: Request filters could be built programmatically, and having the edge case be "an empty whitelist really means include everything" is a little weird. You could also want to use an empty filter with the metadata option turned on to download only the metadata as a ZIP. IMO, it should be that an empty list filters all files, but it's your call.

brianhelba: This brings up another question. Is `mimeFilter` the best name for the option, since it's actually a list of what to include (not what to filter / exclude)?

zachmullen: I think `filter` is acceptable nomenclature for whitelist semantics; in fact, it's similar to the semantics of the python builtin of the same name. If I were making a blacklist instead, I'd probably call it a mask rather than a filter.

PR #1363, opened 2016-05-31
Update to PhantomJS 1.9.20

jbeezley:
This release uses the same binary as 1.9.18, but it is served from github by default to avoid the rate limiting on bitbucket.


zachmullen:
Nice, LGTMITIH

kotfic:
yay! :clap:

brianhelba:
Was this the cause of some of the `ELIFECYCLE` errors during `npm install`?

jbeezley:
Probably, yes.

PR #1367, opened 2016-06-04
Fixes #1366. Document need for nginx max body size config

### LINKED ISSUE #1366: upload issues with girder
is there a test suite we can use to test a girder install remotely. things we are interested in:
- user creation
- user authentication
- file upload of different sizes (we seem to not be able to upload files of 1MB or larger)

since we are running behind an nginx proxy, we would also like to run the test with or without the proxy to isolate effects.

zachmullen:
Thanks @satra for reporting this issue.

### DIFF
--- a/docs/deploy.rst
+++ b/docs/deploy.rst
@@ -73,6 +73,15 @@ Nginx can be used by adding a block such as:
         proxy_send_timeout 600s;
     }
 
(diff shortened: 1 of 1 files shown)


zachmullen:
Anybody object to merging this documentation change?

mgrauer:
LGTM

PR #1368, opened 2016-06-07
Upgrade CherryPy, fixing a Python3 incompatibility

### LINKED ISSUE #1362: CherryPy 5.4 breaks tests
A [change was made for CherryPy 5.4](https://github.com/cherrypy/cherrypy/pull/1432) that causes `cherrypy.serving.response.headers['Location']` to be returned as `str` in Python2 (the same as before) and `bytes` in Python3 (different).

In addition to failing tests, this change may cause unexpected results in how some URL manipulation in the `hdfs_assetstore`, `oauth`, etc. is done. As the tests are fixed to work with CherryPy 5.4 on Python 3, a more thorough examination should be made.

In the meantime, [CherryPy 5.4 was blacklisted from `setup.py`](https://github.com/girder/girder/commit/ae940338e6d931dc67d76e310b2f2f11f9474095), so new installations will not use it. Once this bug is resolved, the blacklisting should be removed.

brianhelba:
* Upgrade CherryPy to 6.0.1
 * CherryPy changes at: https://github.com/cherrypy/cherrypy/blob/master/CHANGES.txt
 * This resolves an bug with Python3 and CherryPy 5.4.0: cherrypy/cherrypy@0d48639
 * Closes #1362
* Upgrade Girder development package requirements
 * Sphinx changes at: http://www.sphinx-doc.org/en/stable/changes.htm

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,7 +7,7 @@ httmock==1.2.5
 mock==2.0.0
 # Versions >0.4.0 of moto cause test failures
 moto==0.4.0
-Sphinx==1.4.2
+Sphinx==1.4.3
 sphinx_rtd_theme==0.1.9
 virtualenv==15.0.2
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,8 +2,7 @@
 bcrypt==2.0.0
 boto==2.40.0
 cffi==1.6.0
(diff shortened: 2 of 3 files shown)


brianhelba:
@girder/developers PTAL.

zachmullen:
LGTM

PR #1369, opened 2016-06-08
Consistency refactor

### LINKED ISSUE #1321: PUT /system/check does not recompute sizes for top level entities

fogleman:
@zachmullen @mgrauer 

This PR fixes #1321. The code for the systemConsistencyCheck was refactored. The check runs three subtasks: fixBaseParents, pruneOrphans, recalculateSizes. Tests were added as well.

The code is a bit hairy, but I blame it on MongoDB and the way the models work. An ORM like SQLAlchemy would be nicer. In particular, it's difficult dealing with stale data when model objects are just dictionaries...

   - fetch a model record
   - execute an update for that model
   - now you have to re-fetch and not forget that your original dict is stale now

Anyway, that's another discussion...

This is my first PR at Kitware, so definitely give me all the feedback you can think up!


zachmullen:
Note that this removes the `checkConsistency` method from the item model, as well as changing the JSON structure returned by the endpoint. The new design is better IMO, and I don't see any harm in changing the return value structure since it didn't make much sense before, but we should decide whether we are OK removing the old method from the model, or whether we should deprecate it and ignore it from test coverage. I can't imagine anyone outside of this core endpoint would have any use for it, but we should consider it nonetheless. ping @girder/developers

girder/api/v1/system.py
```diff
@@ -422,3 +404,55 @@ def stream():
                         break
                     yield data
         return stream
+
+    def fixBaseParents(self, progress):
```
mgrauer: `def _fixBaseParents(self, progress):`

zachmullen:
Great test, LGTM, but let's wait on merging this until people can comment about the breakage issue.

girder/api/v1/system.py
```diff
@@ -422,3 +404,55 @@ def stream():
...
+                        doc['baseParentId'] != baseParentId):
+                    self.model(model).update({'_id': doc['_id']}, update={
+                        '$set': {
+                            'baseParentType': baseParentType,
+                            'baseParentId': baseParentId
+                        }})
+                    fixes += 1
+        return fixes
+
+    def pruneOrphans(self, progress):
```
mgrauer: `def _pruneOrphans(self, progress):`

girder/api/v1/system.py
```diff
@@ -422,3 +404,55 @@ def stream():
...
+            total=steps, current=0, title='Checking for orphaned records')
+        for model in models:
+            for doc in self.model(model).find():
+                progress.update(increment=1)
+                if self.model(model).isOrphan(doc, user=user):
+                    self.model(model).remove(doc)
+                    count += 1
+        return count
+
+    def recalculateSizes(self, progress):
```
mgrauer: `def recalculateSizes(self, progress):`

mgrauer: errrr

`def _recalculateSizes(self, progress):`

mgrauer:
@fogleman Nice work in your first Girder PR!

Can you add docstrings to all of the API and model layer functions?  And prefix the private methods in the API with `_`.

manthey:
One petty thing: the progress will go from 0 to the full value three times.  It would be nicer for progress to increase monotonically.  This would require calculating the steps for each part first, then performing the steps.

fogleman:
@manthey Yes. My hope was to keep the code simpler - so I guess that was a trade off I made.

manthey:
@fogleman Rather than clutter the code, we could have in the progress message something like "Checking for incorrect base parents (step 2 of 3)".  Perhaps the step text would be passed, so only the caller needs to know how many steps: e.g., `results['orphansRemoved'] = self._pruneOrphans(pc, ' (step 1 of 3)')` in `systemConsistencyCheck`, so that `_proneOrphans` would just add that text to the progress message.

girder/api/v1/system.py
```diff
@@ -422,3 +404,55 @@ def stream():
...
+
+    def _pruneOrphans(self, progress):
+        count = 0
+        user = self.getCurrentUser()
+        models = ['folder', 'item', 'file']
+        steps = sum(self.model(model).find().count() for model in models)
+        progress.update(
+            total=steps, current=0, title='Checking for orphaned records')
+        for model in models:
+            for doc in self.model(model).find():
```
cpatrick: nit: Should we save the result of find at line 435?

fogleman:
Good idea. Added that.

girder/api/v1/system.py
```diff
@@ -347,46 +347,28 @@ def systemStatus(self, params):
...
-            # * check that all resources validate
-            # * for filesystem assetstores, find files that are not tracked.
-            # * for gridfs assetstores, find chunks that are not tracked.
-            # * for s3 assetstores, find elements that are not tracked.
-        return results
+        user = self.getCurrentUser()
+        title = 'Running system consistency check'
+        with ProgressContext(progress, user=user, title=title) as pc:
+            results = {}
+            results['orphansRemoved'] = self._pruneOrphans(pc)
```
mgrauer: @fogleman I think he meant change this (and the next two lines similarly) to

`results['orphansRemoved'] = self._pruneOrphans(pc, ' (step 1 of 3)')`

girder/api/v1/system.py
```diff
@@ -422,3 +404,58 @@ def stream():
...
         return stream
+
+    def _fixBaseParents(self, progress):
+        fixes = 0
+        user = self.getCurrentUser()
+        models = ['folder', 'item']
+        steps = sum(self.model(model).find().count() for model in models)
+        progress.update(
+            total=steps, current=0,
+            title='Checking for incorrect base parents (Step 2 of 3)')
```
mgrauer: and then take in the `(Step 2 of 3)` text in this function as a param which is concatenated here.

If some arbitrary caller calls this function, it may not be step 2 and it may not be out of 3 steps total.

fogleman:
I did something a little different that I think is cleaner. Just set the messages all in the same area of the code before calling the private helper functions.

mgrauer:
That change works for me.

The client side tests are failing due to presumed flakiness.  I restarted.

zachmullen:
LGTM

( @fogleman that means you get to click the merge button )

PR #1370, opened 2016-06-09
Add Homepage Markdown Plugin

fogleman:
@zachmullen @mgrauer 

This plugin allows an admin to replace the girder homepage with markdown-formatted content. Currently, image uploads are not handled in the markdown editor.

### DIFF
--- a/plugins/homepage/plugin_tests/homepage_test.py
+++ b/plugins/homepage/plugin_tests/homepage_test.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#  Copyright 2014 Kitware Inc.
+#
+#  Licensed under the Apache License, Version 2.0 ( the "License" );
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


plugins/homepage/plugin_tests/homepage_test.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#  Copyright 2014 Kitware Inc.
```
mgrauer: `Copyright 2016 Kitware Inc.`

Although, it is kinda goofy that we have copyright on Python files and no others.  Not sure why that is.

plugins/homepage/server/__init__.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#  Copyright Kitware Inc.
```
mgrauer: `Copyright 2016 Kitware Inc.`

mgrauer:
I have a mild preference to call this landing page instead of home page, but currently not enough to ask you to rename anything.  If no one chimes in before the PR is done, it can stay.

zachmullen:
The overridden view is called "Front Page" in the code, which is another option for the name.

mgrauer:
- [ ] Add a brief README explaining what this does with a screenshot, the screenshot should show a before and after of the Girder front page with markdown

cpatrick:
Sphinx docs would be preferable to a readme.

mgrauer:
LGTM.

🔫 

@fogleman  is killing it.

PR #1371, opened 2016-06-09
Show collection public / private icon in list

fogleman:
@zachmullen @mgrauer 

The hierarchyWidget shows public / private status on folders and such, but the collectionList didn't show it. So I added it:

screen shot 2016-06-09 at 12 56 03 pm

### DIFF
--- a/clients/web/src/templates/body/collectionList.jade
+++ b/clients/web/src/templates/body/collectionList.jade
@@ -12,6 +12,13 @@
 each collection in collections
   .g-collection-list-entry
     .g-collection-right-column
+      .g-folder-privacy
+        if collection.get('public')
+          i.icon-globe
+          | Public
+        else
+          i.icon-lock
+          | Private
       .g-collection-created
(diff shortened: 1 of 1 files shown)


brianhelba:
This seems like a strict improvement, thanks.

It does look like the collection list entries now take up slightly more vertical space, but I can't totally tell because the test collection in your screenshot doesn't include collection descriptions. Is this the case? If so, can anything be done to reduce the impact of now having 3x lines of height?

zachmullen:
I was suggesting we might copy github's UI for this, where the background color changes for public vs. private, and the icon for public or private goes to the left of the title and is slightly larger.

fogleman:
I quickly tried your suggestion but it doesn't look great (without more effort), so I'd rather not spend more cycles on it for now. Also that's something that might warrant changes on other pages too for consistency.

fogleman:
@zachmullen did it better with #1376, closing.

PR #1372, opened 2016-06-09
Allow for plugins to modify the assetstore adapters

ronichoudhury:
attn: @zachmullen @manthey @alex-r-bigelow

### DIFF
--- a/girder/utility/assetstore_utilities.py
+++ b/girder/utility/assetstore_utilities.py
@@ -62,6 +62,16 @@ def getAssetstoreAdapter(assetstore, instance=True):
 
 
 def setAssetstoreAdapter(storeType, cls):
+
+    """
+    This updates the internal assetstore adapter table with either a new entry,
+    or a modification to an existing entry. Subsequent calls to
+    getAssetstoreAdapter() will return the modified class (or instance thereof),
+    allowing for dynamic updating of assetstore behavior at runtime.
+
+    :param storeType: The assetstore type to create/modify.
+    :type storeType: enum | any
+    """
     _assetstoreTable[storeType] = cls
 
 
--- a/girder/utility/assetstore_utilities.py
+++ b/girder/utility/assetstore_utilities.py
@@ -24,7 +24,7 @@ from girder.constants import AssetstoreType
 from girder import events
 
 
-assetstoreTable = {
+_assetstoreTable = {
     AssetstoreType.FILESYSTEM: FilesystemAssetstoreAdapter,
     AssetstoreType.GRIDFS: GridFsAssetstoreAdapter,
     AssetstoreType.S3: S3AssetstoreAdapter
@@ -47,7 +47,7 @@ def getAssetstoreAdapter(assetstore, instance=True):
     """
     storeType = assetstore['type']
 
-    cls = assetstoreTable.get(storeType)
+    cls = _assetstoreTable.get(storeType)
     if cls is None:
         e = events.trigger('assetstore.adapter.get', assetstore)
         if len(e.responses) > 0:
@@ -62,10 +62,10 @@ def getAssetstoreAdapter(assetstore, instance=True):
 
 
 def setAssetstoreAdapter(storeType, cls):
-    if storeType not in assetstoreTable:
+    if storeType not in _assetstoreTable:
         raise Exception('Illegal assetstore type code: "%s"' % (storeType))
 
-    assetstoreTable[storeType] = cls
+    _assetstoreTable[storeType] = cls
 
 
 def fileIndexFields():
--- a/girder/utility/assetstore_utilities.py
+++ b/girder/utility/assetstore_utilities.py
@@ -71,6 +71,9 @@ def setAssetstoreAdapter(storeType, cls):
 
     :param storeType: The assetstore type to create/modify.
     :type storeType: enum | any
+    :param cls: The new assetstore adapter class to install in the table. This
+    should be an adapter descending from AbstractAssetstoreAdapter.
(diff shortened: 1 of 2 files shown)


girder/utility/assetstore_utilities.py
```diff
@@ -60,6 +59,15 @@ def getAssetstoreAdapter(assetstore, instance=True):
         return cls
 
 
+def setAssetstoreAdapter(storeType, cls):
+    global assetstoreTable
```
fogleman: This isn't needed unless you reassign the variable (which you aren't)

girder/utility/assetstore_utilities.py
```diff
@@ -24,6 +24,11 @@
 from girder import events
 
 
+assetstoreTable = {AssetstoreType.FILESYSTEM: FilesystemAssetstoreAdapter,
```
fogleman: Is this the preferred formatting for dicts in girder? I'd do this:

```python
assetstoreTable = {
    AssetstoreType.FILESYSTEM: FilesystemAssetstoreAdapter,
    AssetstoreType.GRIDFS: GridFsAssetstoreAdapter,
    AssetstoreType.S3: S3AssetstoreAdapter,
}
```

zachmullen: Let's also make this variable private for now (prefix with _).

girder/utility/assetstore_utilities.py
```diff
@@ -60,6 +59,15 @@ def getAssetstoreAdapter(assetstore, instance=True):
         return cls
 
 
+def setAssetstoreAdapter(storeType, cls):
+    global assetstoreTable
+
+    if storeType not in assetstoreTable:
+        raise Exception('Illegal assetstore type code: "%s"' % (storeType))
```
zachmullen: I think this is actually fine, I'd like for plugins to be able to call this as a way to register new assetstore adapters as opposed to using the old `assetstore.adapter.get` hook.

ronichoudhury: Does that mean we'd dump the logic for that particular hook as well?

girder/utility/assetstore_utilities.py
ronichoudhury: Good catch :-D

girder/utility/assetstore_utilities.py
ronichoudhury: I prefer that style too but I've seen the other one in Python projects
before. I'll change it unless @zachmullen objects.

girder/utility/assetstore_utilities.py
zachmullen: Maybe in 2.0, we'd leave it around for now for backward compatibility, but this would be the "recommended" way moving forward.

girder/utility/assetstore_utilities.py
ronichoudhury: Just FYI, I'll be out until Monday but I'll go through and make all the
requested changes then. Have a good weekend everyone!

girder/utility/assetstore_utilities.py
brianhelba: It would be good if registering a listener for the `assetstore.adapter.get` event printed a deprecation warning log message (as we have in some other places).

girder/utility/assetstore_utilities.py
brianhelba: If we're going to have an 80-character width limit, then we really have to globally adopt the dict style rule expressed by @fogleman . This should indeed be changed to that.

brianhelba:
:+1: Great improvement.

ronichoudhury:
I made an attempt at warning about deprecated events, but the approach seems a bit heavy-handed. Let me know if there's a better way to handle it.

girder/utility/assetstore_utilities.py
```diff
@@ -60,6 +61,10 @@ def getAssetstoreAdapter(assetstore, instance=True):
         return cls
 
 
+def setAssetstoreAdapter(storeType, cls):
```
zachmullen: Cool, just needs a good docstring now.

zachmullen:
I like how you handled the deprecation, that is sufficient for now.

ronichoudhury:
Obiter dicta: @fogleman's admonishment about `global` seems to apply in other parts of the codebase too. In particular, I notice it's being used the way I tried to use it (to mark a global variable for mutation rather than assignment) in also every function in events.py.

zachmullen:
Feel free to remove those uses also.

ronichoudhury:
I'll do that in a separate PR to avert the possibility of introducing unrelated bugs into this one.

girder/utility/assetstore_utilities.py
```diff
@@ -62,6 +62,16 @@ def getAssetstoreAdapter(assetstore, instance=True):
...
 def setAssetstoreAdapter(storeType, cls):
+
+    """
+    This updates the internal assetstore adapter table with either a new entry,
+    or a modification to an existing entry. Subsequent calls to
+    getAssetstoreAdapter() will return the modified class (or instance thereof),
+    allowing for dynamic updating of assetstore behavior at runtime.
+
+    :param storeType: The assetstore type to create/modify.
+    :type storeType: enum | any
```
zachmullen: Could we also document the `cls` arg, in particular mention that it should be a subclass of `AbstractAssetstoreAdapter`.

ronichoudhury: Oops, don't know how I missed that :monkey_face:

zachmullen:
LGTM

PR #1375, opened 2016-06-09
Add 'extraParameters' argument to file download endpoint

ronichoudhury:
@zachmullen `extraParameters` is probably a terrible name. Please give me a better suggestion and I will change it :)

Also, I am not sure if any other sites need to be changed in response to this PR.

### DIFF
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -251,8 +251,8 @@ class File(Resource):
         .param('contentDisposition', 'Specify the Content-Disposition response '
                'header disposition-type value', required=False,
                enum=['inline', 'attachment'], default='attachment')
-        .param('extraParameters', 'Arbitrary data to send along with the download '
-               'request.', required=False)
+        .param('extraParameters', 'Arbitrary data to send along with the '
+               'download request.', required=False)
         .errorResponse('ID was invalid.')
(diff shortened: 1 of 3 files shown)


zachmullen:
I'm fine with `extraParameters`. Also really glad we had the foresight to add `**kwargs` to all the assetstore adapter download methods, but it seems we should probably also add it to the abstract one to make it official.

girder/api/v1/file.py
```diff
@@ -251,6 +251,8 @@ def readChunk(self, params):
         .param('contentDisposition', 'Specify the Content-Disposition response '
                'header disposition-type value', required=False,
                enum=['inline', 'attachment'], default='attachment')
+        .param('extraParameters', 'Arbitrary data to send along with the download '
```
zachmullen: This line is too long

ronichoudhury:
Is that abstract method not the one here?
https://github.com/girder/girder/pull/1375/files#diff-7a4bb7daa3b91512f4624e3962ed7906R129

zachmullen:
Yep, just suggesting we should add a `**kwargs` to that so that when people come along and make their own assetstore adapters, they will know to include it in their own implementations, which will increase forward compatibility.

ronichoudhury:
@zachmullen ping

zachmullen:
LGTM

PR #1376, opened 2016-06-09
Conspicuous public status

zachmullen:
ping @fogleman this supersedes PR #1371

### DIFF
--- a/clients/web/src/stylesheets/body/collectionList.styl
+++ b/clients/web/src/stylesheets/body/collectionList.styl
@@ -1,17 +1,11 @@
 .g-collection-list-entry
-  padding 11px 14px
-  margin 2px 0
-  border-radius 3px
-  background-color #fafafc
-  border 1px solid #eaeaea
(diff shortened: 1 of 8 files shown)


fogleman:
Very nice, I'll close my PR.

brianhelba:
@zachmullen Got any screenshots?

zachmullen:
screen shot 2016-06-09 at 4 16 55 pm

brianhelba:
Are you going to change the folder listing too, to be consistent? Users seem regularly confused at the distinction between the collection list and a folder list.

zachmullen:
I hadn't planned on it, that's part of the HierarchyWidget which has a different layout altogether.

fogleman:
I agree with @brianhelba. But for the HierarchyWidget I think a yellow background on Private folders would be sufficient. (No need to move the icon.)

zachmullen:
Oh, you just mean the color? I can add that to the HierarchyWidget.

zachmullen:
Done

fogleman:
LGTM

PR #1377, opened 2016-06-10
Fixed boundHandler so it works without arguments

### LINKED ISSUE #1293: The boundHandler decorator does not work without arguments
The [`boundHandler` decorator](https://github.com/girder/girder/blob/master/girder/api/rest.py#L983) can only be be called as:
```
@boundHandler()
def myRoute(...
```

It cannot currently be called as:
```
@boundHandler
def myRoute(...
```

This second usage without parentheses should be supported. The [`access.cookie` decorator](https://github.com/girder/girder/blob/master/girder/api/access.py#L82) provides a good example of how this is implemented.

salamb:
boundHandler decorator can be used without arguments by
saving the function passed into init and evaluating it when bound handler is
called. May want to rename ctx since it is used as an instance of
Resource or the function depending on whether boundhandler
is passed arguments.
PTAL
Fixes #1293

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -1057,10 +1057,26 @@ class boundHandler(object):  # noqa: class name
     ``self.boolParam``, ``self.requireParams``, etc.
     """
     def __init__(self, ctx=None):
-        self.ctx = ctx or _sharedContext
 
-    def __call__(self, fn):
-        @six.wraps(fn)
-        def wrapped(*args, **kwargs):
-            return fn(self.ctx, *args, **kwargs)
-        return wrapped
+        if ctx is None or isinstance(ctx, Resource):  # Used with arguments
+            self.ctx = ctx or _sharedContext
+            self.func = None
+
+        elif callable(ctx):  # Used as a raw decorator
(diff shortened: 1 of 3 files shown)


girder/api/rest.py
```diff
@@ -1057,10 +1057,22 @@ class boundHandler(object):  # noqa: class name
     ``self.boolParam``, ``self.requireParams``, etc.
     """
     def __init__(self, ctx=None):
-        self.ctx = ctx or _sharedContext
+        if callable(ctx):  # Used as a raw decorator
```
zachmullen: The only concern here would be that someone might pass a `ctx` variable that was actually callable, i.e. they had some Resource class that overrode `__call__` for whatever reason, which would make this case ambiguous.

brianhelba: @salamb How about flipping the logic, and check if the passed is an instance of Resource or a subclass?

salamb:
Added check to look for instances of Resource and also fixed styling issues.
PTAL

zachmullen:
LGTM

PR #1383, opened 2016-06-13
Update requests for girder client

jbeezley:
Version 2.4.2 of requests is required for the json parameter added in #1350.


zachmullen:
LGTM

PR #1384, opened 2016-06-14
Upgrade base node image in our Dockerfile

zachmullen:
ping @Xarthisius


Xarthisius:
tested and deployed in production :+1:

PR #1385, opened 2016-06-14
Girder ansible module ctest

kotfic:
Setting up cmake/ctest to work with the ansible module tests. This is a precursor to setting up nightly tests.  

This is also adding better assertion testing to the yaml playbooks


kotfic:
To test you should be able to:
```sh
git checkout girder_ansible_module-ctest
cd clients/ansible/tests
mkdir build/
cd build/
cmake ../
ctest -V
```

This will: 
1. Launch vagrant box in ```clients/ansible/test``` with girder on port ```20202``` 
2. Provision the [ansible-role-girder](https://github.com/girder/ansible-role-girder) role 
3. Run each of the ```test_*.yml``` files against the VM
4. Destroy the VM.

A significant number of assertions have been added to the test yaml files.  @zachmullen @cpatrick PTAL I'm hoping to get this in before the hackathon.  Thanks!

clients/ansible/test/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+cmake_minimum_required(VERSION 3.5)
```
zachmullen: This is a really new version of CMake, do we actually need this? If not, we should set the value to something easy to obtain, like 2.8.

kotfic:
@zachmullen updated,  thanks!

zachmullen:
The tests are mostly failing for me, with similar error messages:

```
9: TASK [Create 'admin' User] *****************************************************
9: fatal: [girder]: FAILED! => {"changed": false, "failed": true, "module_stderr": "", "module_stdout": "Traceback (most recent call last):\r\n  File \"/tmp/ansible_V_Zv5t/ansible_module_girder.py\", line 1025, in <module>\r\n    class GirderClientModule(GirderClient):\r\nNameError: name 'GirderClient' is not defined\r\n", "msg": "MODULE FAILURE", "parsed": false}
```

I have girder-client installed in my system.

kotfic:
@zachmullen girder client needs to be installed on the VM,  this is usually an indication that the 'vagrant up' didn't work.  Can you run ctest -V  and paste the ansible output from the first test?

Thanks!

zachmullen:
I found the part from the `vagrant_up` test where the error occurs:

```
1: TASK [girder : Download Girder] ************************************************
1: fatal: [girder]: FAILED! => {"failed": true, "msg": "Failed to set permissions on the temporary files Ansible needs to create when becoming an unprivileged user. For information on working around this, see https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user"}
```

kotfic:
sorry,  what version of ansible are you using?

zachmullen:
2.1.0.0

kotfic:
Hmm.. no issues on my side. This looks like some kind of local-config/ansible-role-girder interaction. I will see if i can get someone over here to reproduce.  This is usually the error you get when you SSH in as vagrant and then try to su to another non-privileged user (e.g.  girder).  But i'm not sure why you're seeing it and I'm not.

zachmullen:
I'd love to have a new travis build to run these tests, but it'd basically be a totally different script. Is there a way in travis to use a different set of steps for a given build matrix cell, or would we have to hack it with conditionals?

jbeezley:
You can write shell scripts to manage the builds from the env variables provided by the build matrix.

kotfic:
My original plan was to set up a nightly on cdash seeing as this takes ~10 minutes to run and should really be checked against different versions of ansible and, at least for the short term, the current release, master and 2.0-integration branches

zachmullen:
That's reasonable. We can go ahead and merge this, it's not going to break anything and it seems this failure is only on my local machine. Any further issues with it will be sorted out once we setup the nightly dashboard.

kotfic:
I agree - though this code is actually going to migrate into a PR that includes the ansible-role-girder code that i'll be working on as apart of the hackathon.  Merging!

PR #1386, opened 2016-06-15
Remove erroneous EXTERNAL_DATA reference

msmolens:
The external_data_plugin test definition references an external data file used
only by the external_data_core test. This commit removes the erroneous
reference.


zachmullen:
LGTM... though I'd love to get rid of the dependence of our tests on downloading stuff from Midas

PR #1387, opened 2016-06-15
Use VIRTUAL_ENV Python executable for tests

fogleman:
Just getting started with the web client testing and it was breaking with Python imports. Turns out it was not using my virtual environment, so I made this change. It works, but not sure if it's the best approach (maybe PYTHON_EXECUTABLE shouldn't be overwritten or maybe this check should be done at some higher level, which I tried but failed at), so please review. @zachmullen @cpatrick

### DIFF
--- a/tests/JavascriptTests.cmake
+++ b/tests/JavascriptTests.cmake
@@ -214,6 +214,11 @@ function(add_web_client_test case specFile)
     set(test_module "tests.web_client_test")
   endif()
 
+  set(VIRTUAL_ENV $ENV{VIRTUAL_ENV})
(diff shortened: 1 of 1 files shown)


zachmullen:
I don't think this is a good idea, since it would overwrite a user-specified PYTHON_EXECUTABLE value on each new configure. If you run cmake inside the desired virtualenv in a clean configure, it should pick up the correct python anyway.

fogleman:
Turns out CMakeCache.txt had cached my non-virtualenv python executable. So this patch isn't actually needed. Closing.

PR #1388, opened 2016-06-15
Add curation plugin to girder

fogleman:
@zachmullen @mgrauer @cpatrick - please review

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -180,17 +180,27 @@ and select the option to log in with Google.
 Curation
 --------
 
-This plugin adds curation functionality to Girder, similar to that in MIDAS.
+This plugin adds curation functionality to Girder, allowing content to be
+assembled and approved prior to publication. Admin users can activate curation
+for any folder, and users who are then granted permission can compose content
+under that folder. The users can request publication of the content when it is
+ready, which admins may approve or reject. The plugin provides a UI along with
+workflow management, notification, and permission support for these actions.
+
 The standard curation workflow works as follows.
 
 - Site admins can enable curation for any folder.
 - Users with write access can populate the folder with data.
-- When ready, a user can request approval from the admin. The folder becomes read-only at this point to avoid further changes being made while the admin is reviewing.
+- When ready, a user can request approval from the admin. The folder becomes
+  read-only at this point to avoid further changes being made while the admin
+  is reviewing.
 - The admin can approve or reject the folder contents.
 - If approved, the folder becomes public.
-- If rejected, the folder becomes writeable again by the users who can make changes and resubmit for approval.
+- If rejected, the folder becomes writeable again by the users who can make
+  changes and resubmit for approval.
 
-The curation dialog is accessible from the Folder actions menu and shows the following information.
+The curation dialog is accessible from the Folder actions menu and shows the
+following information.
 
 - Whether curation is enabled or disabled for the folder.
 - The current curation status: construction, requested, or approved.
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -89,6 +89,7 @@ Python package, which may be installed using **pip**: ::
 
 Once the package is installed, the plugin may be enabled via the admin console.
 
+
 Google Analytics
 ----------------
 
@@ -100,6 +101,15 @@ it does not technically trigger routing events for hierarchy navigation).
 
 To use this plugin, simply copy your tracking ID from Google Analytics into the
 plugin configuration page.
+
+
+Homepage
+--------
+
+The Homepage plugin allows the default Girder front page to be replaced by
+content written in [Markdown](https://daringfireball.net/projects/markdown/)
+format. After enabling this plugin, visit the plugin configuration page
+to edit and preview the Markdown.
 
 
 Metadata Extractor
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -187,17 +187,20 @@ under that folder. The users can request publication of the content when it is
 ready, which admins may approve or reject. The plugin provides a UI along with
 workflow management, notification, and permission support for these actions.
 
-The standard curation workflow works as follows.
+The standard curation workflow works as follows, with any operations affecting
+privacy or permissions being applied to the folder and all of its descendent
+folders.
 
-- Site admins can enable curation for any folder.
+- Site admins can enable curation for any folder, which changes the folder to
+  Private.
 - Users with write access can populate the folder with data.
 - When ready, a user can request approval from the admin. The folder becomes
-  read-only at this point to avoid further changes being made while the admin
-  is reviewing.
+  read-only at this point for any user or group with write access, to avoid
+  further changes being made while the admin is reviewing.
 - The admin can approve or reject the folder contents.
-- If approved, the folder becomes public.
-- If rejected, the folder becomes writeable again by the users who can make
-  changes and resubmit for approval.
+- If approved, the folder becomes Public.
+- If rejected, the folder becomes writeable again by any user or group with read
+  access, enabling users to make changes and resubmit for approval.
 
 The curation dialog is accessible from the Folder actions menu and shows the
 following information.
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -177,6 +177,27 @@ then be able to log in with their Google account when they click the log in page
 and select the option to log in with Google.
 
 
+Curation
+--------
+
+This plugin adds curation functionality to Girder, similar to that in MIDAS.
+The standard curation workflow works as follows.
+
+- Site admins can enable curation for any folder.
+- Users with write access can populate the folder with data.
+- When ready, a user can request approval from the admin. The folder becomes read-only at this point to avoid further changes being made while the admin is reviewing.
+- The admin can approve or reject the folder contents.
+- If approved, the folder becomes public.
+- If rejected, the folder becomes writeable again by the users who can make changes and resubmit for approval.
+
+The curation dialog is accessible from the Folder actions menu and shows the following information.
+
+- Whether curation is enabled or disabled for the folder.
+- The current curation status: construction, requested, or approved.
+- A timeline of status changes, who performed them and when.
+- Context-dependent action buttons to perform state transitions.
+
+
 Provenance Tracker
 ------------------
 
--- a/plugins/curation/server/__init__.py
+++ b/plugins/curation/server/__init__.py
@@ -59,6 +59,7 @@ class CuratedFolder(Resource):
         result = dict(DEFAULTS)
         result[TIMELINE] = []
         result.update(folder.get(CURATION, {}))
+        result['public'] = folder.get('public')
         return result
 
     @access.user(scope=TokenScope.DATA_READ)
@@ -154,6 +155,7 @@ class CuratedFolder(Resource):
                 'curation.rejected.mako')
 
         self.model('folder').save(folder)
+        curation['public'] = folder.get('public')
         return curation
 
     def _makeReadOnly(self, folder):
--- a/plugins/curation/server/__init__.py
+++ b/plugins/curation/server/__init__.py
@@ -38,6 +38,11 @@ CONSTRUCTION = 'construction'
 REQUESTED = 'requested'
 APPROVED = 'approved'
 
+DEFAULTS = {
+    ENABLED: False,
+    STATUS: CONSTRUCTION,
+}
+
 
 class CuratedFolder(Resource):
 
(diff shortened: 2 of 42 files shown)


docs/plugins.rst
```diff
@@ -177,6 +177,27 @@ then be able to log in with their Google account when they click the log in page
 and select the option to log in with Google.
 
 
+Curation
+--------
+
+This plugin adds curation functionality to Girder, similar to that in MIDAS.
```
jeffbaumes: Since people (like me) don't immediately know what "curation" means in this case and may not know about MIDAS, I'd suggest replacing "similar to that in MIDAS" with a sentence or two about what it means to curate data. Something like "This enables administrators to give certain users write access to folders while maintaining control of how its contents change through an extra approval process." I'm sure this could be explained better.

mgrauer: suggested

This plugin adds curation functionality to Girder, allowing content to be assembled and approved prior to publication.  Admin users can activate curation for any folder, and users who are then granted permission can compose content under that folder.  The users can request publication of the content when it is ready, which admins may approve or reject.  The plugin provides a UI along with workflow management, notification, and permission support for these actions.

plugins/curation/server/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+        .param('enabled', 'Enable or disable folder curation.',
+               required=False, dataType='boolean')
+        .param('status', 'Set the folder curation status.',
+               required=False, dataType='string')
+        .errorResponse('ID was invalid.')
+        .errorResponse('Write permission denied on the folder.', 403)
+    )
+    def setCuration(self, folder, params):
+        user = self.getCurrentUser()
+        if CURATION not in folder:
```
mgrauer: not that you need to change it, but I recently found `setdefault` and thought it was useful

it would allow you to replace

    if CURATION not in folder:
        folder[CURATION] = dict(DEFAULTS)
    curation = folder[CURATION]

with

    curation = folder.setdefault(CURATION, dict(DEFAULTS))

plugins/curation/server/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+DEFAULTS = {
+    ENABLED: False,
+    STATUS: CONSTRUCTION,
+}
+
+
+class CuratedFolder(Resource):
+
+    @access.public
+    @loadmodel(model='folder', level=AccessType.READ)
```
mgrauer: I think we should require user access to see the curation status, and also allow clients with tokens to access this endpoint.  So change

    @access.public

to

    @access.user(scope=TokenScope.DATA_READ)

plugins/curation/server/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+        Description('Set curation details for the folder.')
+        .param('id', 'The folder ID', paramType='path')
+        .param('enabled', 'Enable or disable folder curation.',
+               required=False, dataType='boolean')
+        .param('status', 'Set the folder curation status.',
+               required=False, dataType='string')
+        .errorResponse('ID was invalid.')
+        .errorResponse('Write permission denied on the folder.', 403)
+    )
+    def setCuration(self, folder, params):
```
mgrauer: There is a lot of boolean logic in this endpoint that was hard for me to reason about, as I needed to think about enabled/disabled + state transitions + user perms.

I'd prefer to see this broken into  separate endpoints, and we should also be able to rely on the decorators to enforce permissions in most (all?) of the cases as well as getting more parameter validation.

The following is just a suggestion, feel free to go with a different design.  I think it simplifies reasoning, especially about permissions --which I'd rather not do, so as to minimize access bugs, and because I'm not good at it. 🐒 

     @access.user(scope=TokenScope.DATA_OWN)
     @loadmodel(model='folder', level=AccessType.OWN)
     @describeRoute(
         Description('Enable or disable curation for the folder.')
         .param('id', 'The folder ID', paramType='path')
         .param('enabled', 'Enable or disable folder curation.',
                required=True, dataType='boolean')
         .errorResponse('ID was invalid.')
         .errorResponse('Admin permission denied on the folder.', 403)
     )
     def setCuration(self, folder, params):
     ...

     @loadmodel(model='folder', level=AccessType.WRITE)
     @describeRoute(
         Description('Change the curation state from construction to requested for the folder.')
         .param('id', 'The folder ID', paramType='path')
         .errorResponse('ID was invalid.')
         .errorResponse('Write permission denied on the folder.', 403)
     )
     def setCurationRequested(self, folder, params):
         raise something if not curationEnabled(folder) 
         ...

     @access.user(scope=TokenScope.DATA_OWN)
     @loadmodel(model='folder', level=AccessType.OWN)
     @describeRoute(
         Description('Change the curation state for the folder.')
         .param('id', 'The folder ID', paramType='path')
         .param('status', 'Set the folder curation status.',
                required=True, dataType='string')
         .errorResponse('ID was invalid.')
         .errorResponse('Admin permission denied on the folder.', 403)
     )
     def setCurationState(self, folder, params):
         raise something if not curationEnabled(folder) 
         ...

plugins/curation/server/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+            self._sendMail(
+                folder, curation.get(REQUEST_USER_ID),
+                'REJECTED: ' + folder['name'],
+                'curation.rejected.mako')
+
+        self.model('folder').save(folder)
+        return curation
+
+    def _makeReadOnly(self, folder):
+        for doc in folder['access']['users'] + folder['access']['groups']:
```
mgrauer: both of these access control change methods should be recursive, currently they are this folder only.  I've got to run so don't have time to look up how to do it correctly.

plugins/curation/server/__init__.py
mgrauer: A few other things to note that might be helpful:

* The `describeRoute` decorator generates the swagger, but will not actually validate the params
* The `id` param will be validated through the `loadModel` decorator, against the `level` param to that decorator, to ensure that the folder with that `id` exists and that the current user has the `level` of access
* For required params, you can use [requireParams](https://github.com/girder/girder/blob/e46ff18eb0d0b3c34ca6432c188ba5cdbe477f40/plugins/thumbnails/server/rest.py#L55)
* You can use an [enum](https://github.com/girder/girder/blob/e46ff18eb0d0b3c34ca6432c188ba5cdbe477f40/plugins/thumbnails/server/rest.py#L47-L48) as a param type in the `describeRoute` decorator, but you would need to [enforce](https://github.com/girder/girder/blob/e46ff18eb0d0b3c34ca6432c188ba5cdbe477f40/plugins/thumbnails/server/rest.py#L61-L64) it yourself

plugins/curation/server/__init__.py
zachmullen: Use the `setAccessList` on the model with `recurse=True`.

mgrauer: Here's the API [endpoint](https://github.com/girder/girder/blob/b319a1f9aed8fdd866527e4b2dc3fc5a48220e10/girder/api/v1/folder.py#L219-L240) for updating folder access, you can follow what it does for recursive access updates.  It's a good idea to use a ProgressContext for notifications because these trees could get large.

mgrauer:
Currently if the folder is not curation enabled, for a user that doesn't have admin on the folder (and hence can't affect its curation status), they still get a line in the folder actions dialog for `Folder curation`.

screen shot 2016-06-16 at 10 56 59 am


When they click on this, a dialog pops up that tells them curation is not available.

screen shot 2016-06-16 at 10 57 06 am


I have a mild preference towards not displaying the `Folder curation` line in the folder actions in this case, so as to not confuse end users as they navigate around.  That might require an additional API call though, or else an additional exposed field on the Folder model coming from the curation plugin.

plugins/curation/server/__init__.py
fogleman: Not sure if this will work without changing `setAccessList`, since the user won't necessarily have admin rights on the folder?

plugins/curation/server/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+            if doc['level'] == AccessType.WRITE:
+                doc['level'] = AccessType.READ
+
+    def _makeWriteable(self, folder):
+        for doc in folder['access']['users'] + folder['access']['groups']:
+            if doc['level'] == AccessType.READ:
+                doc['level'] = AccessType.WRITE
+
+    def _addTimeline(self, oldCuration, curation, text):
```
mgrauer: add docstrings to all of these internal functions

plugins/curation/server/__init__.py
mgrauer: I may not be following your question exactly.

One approach could be to change these helper functions so that they only update a passed in access document.  Then the caller (who would have admin access to the folder in the case of the 3 separated endpoints, or who could at least check on their perms) would be able to call setAccessList passing their current user down.

e.g.

    @access.user(scope=TokenScope.DATA_OWN)
    @loadmodel(model='folder', level=AccessType.OWN)
    def setCurationState():
        ...
        if stateTransitionIncludesWriteable:
            access = self._makeWriteable(folder['access'])
            self.model('folder').setAccessList(folder, access, save=True, recurse=True, user=user, ...)
        self.model('folder').save(folder)
        return curation
        
    def _makeReadOnly(self, access):
        for doc in access['users'] + access['groups']:
            if doc['level'] == AccessType.WRITE:
                doc['level'] = AccessType.READ

fogleman: When `setAccessList` recursively checks folders, it only looks at folders where the user has `ADMIN` privilege.

In the case of the `construction -> approved` transition, where we downgrade write permissions to read permissions, the user is just a regular user who has `WRITE` access but not `ADMIN`. So this user is not able to alter the folder permissions recursively using the existing `setAccessList` logic.

mgrauer: Makes sense, thanks for clarifying.

It would be easy enough to copy functionality from `setAccessList` here, but probably better to amend that function.  Maybe with an additional `force=False` param to skip the permission filtering of subfolders?

ping @zachmullen

zachmullen: I'm fine with that.

plugins/curation/server/__init__.py
fogleman: Yeah, I know `setdefault` but it's not the most readable pattern, I usually reserve it for this use case:

    d.setdefault(k, []).append(foo)

plugins/curation/server/__init__.py
fogleman: I agree the `setCuration` method is a little long, but I'd rather not refactor it at this point as a lot of the code is simpler due to it being a single REST endpoint.

fogleman:
I've addressed all of the comments - please have another look.

fogleman:
@zachmullen How can I test the sending of emails from my girder plugin?

zachmullen:
Look through [this test](https://github.com/girder/girder/blob/master/tests/cases/mail_test.py) for examples, in particular see references to `mockSmtp`.

fogleman:
What if I want to actually receive an email to check the formatting and whatnot?

zachmullen:
Oh, for manual testing you mean? In the past I've just used my own SMTP server, but if you send more than one or two emails like that you can get our IP range flagged, which is bad... to use services like mailgun, you need to do domain validation. One thing you could do would be to write the testing code using mockSmtp, then write the contents out to an HTML file on your local disk and see how it looks in your browser.

msmolens:
I believe I once set up a local Girder instance to use Gmail's SMTP server for testing. (But it might have been a dream.) See: https://support.google.com/a/answer/176600?hl=en.

fogleman:
@msmolens Cool, it worked using `smtp-relay.gmail.com` on port 465 with SSL/TLS.

mgrauer:
- [ ] change the access levels for `setCuration` to

     @loadmodel(model='folder', level=AccessType.WRITE)

- [ ] add progress notifications to the set public/read/write functions

Let's look for an opportunity to make `setCuration` more readable when we iterate on the functionality further.

Note that the set public and set write/read functions will loop over the full folder tree 2X with save and load in some cases.

mgrauer:
One last change, to the docs, adding a bit more detail about permissions.

```
The standard curation workflow works as follows, with any operations affecting privacy or permissions being applied to the folder and all of its descendent folders.

- Site admins can enable curation for any folder, which changes the folder to Private.
- Users with write access can populate the folder with data.
- When ready, a user can request approval from the admin. The folder becomes read-only at this point for any user or group with write access, to avoid further changes being made while the admin is reviewing.
- The admin can approve or reject the folder contents.
- If approved, the folder becomes Public.
- If rejected, the folder becomes writeable again by any user or group with read access, enabling users to make changes and resubmit for approval.
```

mgrauer:
LGTMITIH (If Travis Is Happy).

Awesome work.

PR #1389, opened 2016-06-15
Fixes #1306. Support efficient job log updates

### LINKED ISSUE #1306: Make the Job log more efficient at handling larger logs
Currently, the entire Job log gets stored on one continually-concatenated string, which gets entirely reallocated and re-send to the database every time it is appended to.

The Job log should be improved so that it's not reallocating new concatenated strings every time a Job updates it, and so it doesn't have to push the entire history to the server every time it's updated.

zachmullen:


zachmullen:
@brianhelba PTAL

zachmullen:
@girder/developers PTAL, this fixes potential race conditions when updating job information

brianhelba:
I'll QA this soon.

zachmullen:
Could someone QA this in the near future? I'd like to integrate this in front of our 2.0 release push.

mgrauer:
Claiming for QA

mgrauer:
Seems as though `AccessControlledModel. prefixSearch` is missing the crucial `prefixSearchFields` param.

plugins/worker/server/__init__.py
```diff
@@ -86,8 +86,9 @@ def schedule(event):
 
         # Set the job status to queued and record the task ID from celery.
         job['celeryTaskId'] = asyncResult.task_id
-        ModelImporter.model('job', 'jobs').updateJob(
-            job, status=JobStatus.QUEUED)
+        jobModel = ModelImporter.model('job', 'jobs')
+        job = jobModel.save(job)
+        jobModel.updateJob(job, status=JobStatus.QUEUED)
```
mgrauer: This strikes me as odd, we need to save the job before calling `updateJob`.  Is there no way to include the update to `job['celeryTaskId']` in `updateJob`?

plugins/jobs/web_client/js/views/JobDetailsWidget.js
```diff
@@ -1,12 +1,26 @@
...
+            }
+        }, this);
+
+        girder.eventStream.on('g:event.job_log', function (event) {
+            var info = event.data;
+            if (info._id === this.job.id) {
+                var container = this.$('.g-job-log-container');
+                if (info.overwrite) {
+                    this.job.set({log: [info.text]});
+                    container.text(info.text);
```
mgrauer: why do you escape the `info.text` for `container.append` below but not this `container.text`?

mgrauer:
@zachmullen My QA pass is done, and I tried to clarify and add consistency to some docstrings.

The two main issues I see here

 - [ ] escaping text (my inline comment) in the JobDetailsWidget.js
 - [ ] questions about `updateJob` in the model `job.py`
    - [ ] I think a method called `updateJob` should update the job fully, it seems odd to have to save it first to get the celeryTaskId updated and then call `updateJob`, IIUC
    - [ ] There is a comment in the docstring of `updateJob` that says `Any of the updateable fields that are set to None in the kwargs will not be modified.`, but I don't see where this actually comes into play in that method.  Is this related to the celeryTaskId question above?

plugins/jobs/web_client/js/views/JobDetailsWidget.js
zachmullen: `$.fn.text` escapes its input, `$.fn.append` does not.

plugins/worker/server/__init__.py
zachmullen: Good point, I can add something to the job model to allow us to do all this in a single step.

zachmullen:
> There is a comment in the docstring of updateJob that says Any of the updateable fields that are set to None in the kwargs will not be modified., but I don't see where this actually comes into play in that method.

It's the lines like `if log is not None:` and `if status is not None:`.

mgrauer:
Ah, I was confused reading the docstring whether you were talking about the job['kwargs'] or the method kwargs.  I added a commit to clarify.

LGTM.  LGTY?

PR #1390, opened 2016-06-16
Trigger events upon importing files and folders into filesystem assetstore

Xarthisius:
This PR adds few additional events that happen while files are imported


zachmullen:
What's the use case here that you only want to do this in the case of a filesystem assetstore?

Xarthisius:
I'm storing physical `importPath` in items/folder metadata. It only makes sense for import not general upload/creation

zachmullen:
That makes sense. I think we should remove the `_` from the beginning of the event name if we plan to make it officially supported. Also, having the type in both the event name and as a field in the info is redundant, so I think we should either have a single event and pass the `type` field, or have two different events and remove the `type` field.

zachmullen:
LGTM

PR #1391, opened 2016-06-16
Add missing superclass constructor to Cats example

Xarthisius:
If someone bases their plugin on Cats example he/she is greeted by:
` WARNING: Resource subclass "Cats" did not call "Resource__init__()" from its constructor.`
when starting girder.


zachmullen:
Thanks @Xarthisius LGTM

PR #1392, opened 2016-06-16
[RFC] Allow to call a custom function while iterating over files in a folder

Xarthisius:
I'd like to iterate over files in a given Girder's folder and perform some action. Current implementation of `.fileList()` method does exactly that, yet doesn't allow to do anything except get raw data stream. With this change `fileList` becomes more flexible, e.g.:

```
@access.public(scope=TokenScope.DATA_READ)
@loadmodel(model='folder', level=AccessType.READ)
@describeRoute(
    Description('Get physical paths for files in folder.')
    .param('id', 'The ID of the folder.', paramType='path')
    .errorResponse('ID was invalid.')
    .errorResponse('Read access was denied for the folder.', 403)
)
@boundHandler()
def myHandler(self, folder, params):
    user = self.getCurrentUser()
    result = {}
    for (path, item) in self.model('folder').fileList(
            folder, user=user, subpath=True, stream=False):
        assetstore = self.model('assetstore').load(item['assetstoreId'])
        adapter = assetstore_utilities.getAssetstoreAdapter(assetstore)
        result[path] = adapter.fullPath(item)
    return result

def load(info):
    info['apiRoot'].folder.route('GET', (':id', 'contents'), myHandler)
```

allows to get mapping between Girder's filenames and physical paths in filesystem assetstore.

### DIFF
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -658,17 +658,17 @@ class Folder(AccessControlledModel):
                                      user=user):
             if sub['name'] == metadataFile:
                 metadataFile = None
-            for (filepath, func) in self.fileList(
+            for (filepath, file) in self.fileList(
                     sub, user, path, includeMetadata, subpath=True,
                     mimeFilter=mimeFilter, stream=stream):
-                yield (filepath, func)
+                yield (filepath, file)
         for item in self.childItems(folder=doc):
             if item['name'] == metadataFile:
                 metadataFile = None
-            for (filepath, func) in self.model('item').fileList(
+            for (filepath, file) in self.model('item').fileList(
                     item, user, path, includeMetadata, mimeFilter=mimeFilter,
                     stream=stream):
-                yield (filepath, func)
+                yield (filepath, file)
         if includeMetadata and metadataFile and doc.get('meta', {}):
             def stream():
                 yield json.dumps(doc['meta'], default=str)
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -624,9 +624,14 @@ class Folder(AccessControlledModel):
         return count
 
     def fileList(self, doc, user=None, path='', includeMetadata=False,
-                 subpath=True, mimeFilter=None, stream=True):
+                 subpath=True, mimeFilter=None, data=True):
         """
-        Generate a list of files within this folder.
+        This function generates a list of 2-tuples whose first element is the
+        relative path to the file from the folder's root and whose second
+        element depends on the value of the `data` flag. If `data=True`, the
+        second element will be a generator that will generate the bytes of the
+        file data as stored in the assetstore. If `data=False`, the second
+        element is the file document itself.
 
         :param doc: The folder to list.
         :param user: The user used for access.
@@ -643,12 +648,12 @@ class Folder(AccessControlledModel):
         :param mimeFilter: Optional list of MIME types to filter by. Set to
             None to include all files.
         :type mimeFilter: list or tuple
-        :param stream: Return stream function with file data or file model
-            object if False
-        :type stream: bool
+        :param data: If True return raw content of each file as stored in the
+            assetstore, otherwise return file document.
+        :type data: bool
         :returns: Iterable over files in this folder, where each element is a
                   tuple of (path name of the file, stream function with file
-                  data).
+                  data or file object).
         :rtype: generator(str, func)
         """
         if subpath:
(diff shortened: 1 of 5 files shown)


zachmullen:
Cool, I like this in theory, will look a little further into it shortly.

girder/models/item.py
```diff
@@ -396,7 +396,7 @@ def copyItem(self, srcItem, creator, name=None, folder=None,
         return newItem
 
     def fileList(self, doc, user=None, path='', includeMetadata=False,
-                 subpath=True, mimeFilter=None):
+                 subpath=True, mimeFilter=None, streamCallback=None):
```
zachmullen: I think it would be better to break this up from a tuple into three separate parameters, then the types can be expressed more cleanly in the sphinx docs:

```
:param callback: A callback that will be called for each file found. If not set, defaults to the `download` method of the file model. This is always called with the file as the first argument.
:type callback: callable
:param callbackPargs: Optional additional positional arguments to be passed into the callback after the file argument.
:type callbackArgs: iterable
:param callbackKwargs: Optional additional keyword arguments to be passed into the callback.
:type callbackKwargs: dict
```

zachmullen:
After thinking about this some more, I think rather than pass in a callback, since this is a generator it would be better to just have a switch that would change the behavior to just yield the file rather than yielding a generator function as the second part of the tuple. That would allow the caller to do whatever they want with the file. Does that make sense?

Xarthisius:
That'd be also good. Should I modify 'user.fileList' and 'collection.fileList' methods too?

zachmullen:
Yep

girder/models/folder.py
```diff
@@ -655,16 +658,17 @@ def fileList(self, doc, user=None, path='', includeMetadata=False,
                                      user=user):
             if sub['name'] == metadataFile:
                 metadataFile = None
-            for (filepath, file) in self.fileList(
+            for (filepath, func) in self.fileList(
```
zachmullen: I vote we keep this as "file" rather than changing it to func.

girder/models/item.py
```diff
@@ -430,13 +434,20 @@ def fileList(self, doc, user=None, path='', includeMetadata=False,
                     (includeMetadata and doc.get('meta', {}))):
                 path = os.path.join(path, doc['name'])
         metadataFile = 'girder-item-metadata.json'
+
+        if stream:
+            returnFunc = functools.partial(self.model('file').download,
+                                           headers=False)
+        else:
+            def returnFunc(item):
+                return item
```
zachmullen: We can write this more simply as:

```
if stream:
    val = self.model('file').download(file, headers=False)
else:
    val = file

...

yield (os.path.join(path, file['name']), val)
```

Xarthisius: I did it on purpose to avoid `if` clause in tight loop. Wouldn't it cause performance issues for large number of files?

fogleman: Negligible difference. Opt for readability / simplicity over premature optimization.

zachmullen:
Logic looks good! Can you just add a test for the `stream=False` mode?

zachmullen:
Thanks for the test. I would like to make one more proposal to change this to make it easier for developers to understand. I would suggest changing the parameter name from `stream=True` to `data=True`. To motivate this, I'd suggest adding the following text to the docstrings:

> This function returns a 2-tuple whose first element is the relative path to the file from the root node and whose second element depends on the value of the `data` flag. If `data=True`, the second element will be a generator that will generate the bytes of the file data as stored in the assetstore. If `data=False`, the second element is the file document itself.

That elucidates the purpose of this flag, which is to distinguish between file data and file metadata.

tests/cases/file_test.py
```diff
@@ -341,6 +341,14 @@ def _testDownloadFolder(self):
             path='/file/chunk', user=self.user, fields=fields, files=files)
         self.assertStatusOk(resp)
 
+        # List files in the folder
+        testFolder = self.model('folder').load(test['_id'], force=True)
+        fileList = [(path, file['name'])
+                    for (path, file) in self.model('folder').fileList(
+                        testFolder, user=self.user,
+                        subpath=True, stream=False)]
```
zachmullen: Forgot to change this one 😄

Xarthisius: ups... Fixed

zachmullen:
LGTM, many thanks @Xarthisius

PR #1394, opened 2016-06-20
Email verification

fogleman:
A new core setting allows email verification to be disabled, optional or required.

If disabled, no verification email is sent.
If optional, verification email is sent but verification is not required to login.
If required, verification email is sent and verification is required to login.

This doesn't yet address the need for an admin to create an account and let the user set the password after clicking the link in the email.

We might need to add a way for a user to re-send the verification email if they can't find the original for whatever reason. Or, the forgot-password email could also set emailVerified to true.

### DIFF
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -53,6 +53,7 @@ class User(Resource):
         self.route('PUT', ('password', 'temporary'),
                    self.generateTemporaryPassword)
         self.route('DELETE', ('password',), self.resetPassword)
+        self.route('GET', ('verify', ':id'), self.verifyEmail)
 
     @access.public
     @filtermodel(model='user')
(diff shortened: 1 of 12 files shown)


girder/models/setting.py
```diff
@@ -163,6 +163,13 @@ def validateCoreRegistrationPolicy(self, doc):
                 'Registration policy must be either "open" or "closed".',
                 'value')
 
+    def validateCoreEmailVerification(self, doc):
+        doc['value'] = doc['value'].lower()
+        if doc['value'] not in ('required', 'optional', 'disabled'):
+            raise ValidationException(
+                'Email verification must be either "required" or "disabled".',
```
danlamanna: Does this message need to be updated to include `optional`?

fogleman: Yep, good catch.

fogleman:
Currently, verifying an email is a GET: /user/verify/{id}?token={token} ... should this be a PUT? Only issue is I used PUT /user/verify to send a new verification email. Suggestions for these endpoints?

zachmullen:
Typically we would organize it so that the user ID comes after the user literal, so `user/{id}/verify`. I think it should be a PUT rather than a GET since the purpose is to modify a field in the system.

We should also consider whether "verify" is the right word, since it's more properly RESTful to use nouns rather than verbs. "Verification" would work, but there may be other alternatives.

fogleman:
Cool, now it's like this:

- `PUT user/{id}/verification` verifies an email address using a token
- `PUT user/verification` sends a verification email using a login name or email

zachmullen:
Cool. I actually think the `user/verification` action makes more sense as either a POST (create me an email) or GET (requesting an email from the server, in a way) than as a PUT (which implies modification of some existing data).

fogleman:
`POST` makes sense, changed.

fogleman:
No longer WIP, please review.

fogleman:
Bump. @zachmullen @mgrauer @cpatrick

girder/api/v1/user.py
```diff
@@ -411,3 +424,54 @@ def getUserDetails(self, user, params):
...
+    )
+    def verifyEmail(self, user, params):
+        self.requireParams('token', params)
+        token = self.model('token').load(
+            params['token'], force=True, objectId=False, exc=True)
+        delta = (token['expires'] - datetime.datetime.utcnow()).total_seconds()
+        hasScope = self.model('token').hasScope(
+            token, TokenScope.EMAIL_VERIFICATION)
+
+        if token.get('userId') != user['_id'] or delta <= 0 or not hasScope:
```
cpatrick: nit: Three conditions on the `or` is where I start thinking about `any`.

fogleman: Not sure I've ever seen `any` used with a set of arbitrary boolean expressions, looks weird. Usually it's a list or a list comprehension.

    if any([token.get('userId') != user['_id'], delta <= 0, not hasScope]):

Also, I directly pulled that code from `checkTemporaryPassword` :)

cpatrick: I was thinking more like:

```
if any([token.get('userId') != user['_id'],
        delta <= 0,
        not hasScope]):
```

but it's just a nit :smile_cat:.

cpatrick:
LGTM. :shipit:

Issue #1393, opened 2016-06-20
resource/lookup should have a mode for testing existence


zachmullen:
Current behavior raises 400 status if the requested path does not exist; we should allow another parameter to make the function simply return `None` in this case, perhaps something like `test=true`.

PR #1395, opened 2016-06-20
Added resource lookup existence testing.

xerxes455:
Responding to issue #1393.

### DIFF
--- a/girder/api/v1/resource.py
+++ b/girder/api/v1/resource.py
@@ -110,6 +110,7 @@ class Resource(BaseResource):
         Validate a JSON string listing resources.  The resources parameter is a
         JSON encoded dictionary with each key a model name and each value a
         list of ids that must be present in that model.
+
         :param params: a dictionary of parameters that must include 'resources'
         :param allowedModels: if present, an iterable of models that may be
                               included in the resources.
@@ -134,6 +135,7 @@ class Resource(BaseResource):
     def _getResourceModel(self, kind, funcName=None):
         """
         Load and return a model with a specific function or throw an exception.
+
         :param kind: the name of the model to load
         :param funcName: a function name to ensure that each model contains.
         :returns: the loaded model.
@@ -149,6 +151,7 @@ class Resource(BaseResource):
     def _lookUpToken(self, token, parentType, parent):
         """
         Find a particular child resource by name or throw an exception.
+
         :param token: the name of the child resource to find
         :param parentType: the type of the parent to search
         :param parent: the parent resource
--- a/girder/api/v1/resource.py
+++ b/girder/api/v1/resource.py
@@ -183,11 +183,12 @@ class Resource(BaseResource):
     def _lookUpPath(self, path, user, test=False):
         """
         Look up a resource in the data hierarchy by path.
+
         :param path: path of the resource
-        :param user: user with correct privledges to access path
+        :param user: user with correct privileges to access path
         :param test: defaults to false, when set to true
-        will return None instead of throwing exception when
-        path doesn't exist
+            will return None instead of throwing exception when
+            path doesn't exist
         """
         pathArray = [token for token in path.split('/') if token]
         model = pathArray[0]
@@ -251,7 +252,7 @@ class Resource(BaseResource):
     def lookup(self, params):
         self.requireParams('path', params)
         test = self.boolParam('test', params, default=False)
-        return self._lookUpPath(params['path'], test, self.getCurrentUser())
+        return self._lookUpPath(params['path'], self.getCurrentUser(), test)
 
     @access.cookie(force=True)
     @access.public(scope=TokenScope.DATA_READ)
--- a/girder/api/v1/resource.py
+++ b/girder/api/v1/resource.py
@@ -180,7 +180,7 @@ class Resource(BaseResource):
         raise RestException('Child resource not found: %s(%s)->%s' % (
             parentType, parent.get('name', parent.get('_id')), token))
 
-    def _lookUpPath(self, path, user):
+    def _lookUpPath(self, path, test, user):
         pathArray = [token for token in path.split('/') if token]
         model = pathArray[0]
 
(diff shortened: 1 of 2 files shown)


girder/api/v1/resource.py
```diff
@@ -180,7 +180,7 @@ def _lookUpToken(self, token, parentType, parent):
         raise RestException('Child resource not found: %s(%s)->%s' % (
             parentType, parent.get('name', parent.get('_id')), token))
 
-    def _lookUpPath(self, path, user):
+    def _lookUpPath(self, path, test, user):
```
mgrauer: Thanks for the contribution @xerxes455 .

Change this function def to better support backwards compatibility and provide default value for `test`

    def _lookUpPath(self, path, user, test=False):

Can you also add a docstring for this function?

girder/api/v1/resource.py
```diff
@@ -190,15 +190,21 @@ def _lookUpPath(self, path, user):
             parent = self.model('user').findOne({'login': username})
 
             if parent is None:
-                raise RestException('User not found: %s' % username)
+                if test is False:
```
mgrauer: I think these blocks should be rewritten to be more Pythonic, like

    if test:
        return None
    else:
        raise RestException('User not found: %s' % username)

girder/api/v1/resource.py
```diff
@@ -223,13 +232,17 @@ def _lookUpPath(self, path, user):
                'path starting with either "/user/[user name]", for a user\'s '
                'resources or "/collection/[collection name]", for resources '
                'under a collection.')
+        .param('test',
+               'Specify whether to return null if path doesn\'t exist.',
```
mgrauer: `Specify whether to return None instead of throwing an exception when path doesn\'t exist.`

Sorry for being so nit-picky.

tests/cases/resource_test.py
```diff
@@ -455,11 +455,21 @@ def testGetResourceByPath(self):
                 str(resp.json['_id']), str(item['_id']))
 
         # test bogus path
+        # test is set to false, response code should be 400
```
mgrauer: Can you add another test, replicating the old behavior?  i.e., Don't pass a value for `test` and ensure you get the 400 exception.

girder/api/v1/resource.py
```diff
@@ -180,7 +180,15 @@ def _lookUpToken(self, token, parentType, parent):
         raise RestException('Child resource not found: %s(%s)->%s' % (
             parentType, parent.get('name', parent.get('_id')), token))
 
-    def _lookUpPath(self, path, test, user):
+    def _lookUpPath(self, path, user, test=False):
+        """
+        Look up a resource in the data hierarchy by path.
+        :param path: path of the resource
+        :param user: user with correct privledges to access path
```
mgrauer: privileges

girder/api/v1/resource.py
```diff
@@ -180,7 +180,15 @@ def _lookUpToken(self, token, parentType, parent):
         raise RestException('Child resource not found: %s(%s)->%s' % (
             parentType, parent.get('name', parent.get('_id')), token))
 
-    def _lookUpPath(self, path, test, user):
+    def _lookUpPath(self, path, user, test=False):
```
mgrauer: You'll need to change the order below when you call this, as the param order changed

    return self._lookUpPath(params['path'], self.getCurrentUser(), test)

mgrauer:
I'm satisfied once these two additional changes come in.

@zachmullen, does this close the issue for you?

girder/api/v1/resource.py
zachmullen: Also add a blank line between the description and the param list so that sphinx-build doesn't complain.

girder/api/v1/resource.py
```diff
@@ -180,7 +180,15 @@ def _lookUpToken(self, token, parentType, parent):
...
 
-    def _lookUpPath(self, path, user):
+    def _lookUpPath(self, path, user, test=False):
+        """
+        Look up a resource in the data hierarchy by path.
+        :param path: path of the resource
+        :param user: user with correct privledges to access path
+        :param test: defaults to false, when set to true
+        will return None instead of throwing exception when
+        path doesn't exist
```
zachmullen: Sphinx will also complain if you don't indent continuations of these descriptions, e.g.

    :param blah: blah blah ...
        blah.

tests/cases/resource_test.py
```diff
@@ -455,10 +455,25 @@ def testGetResourceByPath(self):
...
+                                    'test': False})
+        self.assertStatus(resp, 400)
+
+        # test is set to true, response code should be 200 and response body
+        # should be null (None)
+        resp = self.request(path='/resource/lookup',
+                            method='GET', user=self.user,
+                            params={'path': '/collection/bogus/path',
+                                    'test': True})
+        self.assertStatusOk(resp)
```
zachmullen: Let's add `self.assertEqual(resp.json, None)`

xerxes455:
@zachmullen  when you want blank line between the description and the param list for the _lookUpPath method, do you want me to go through the rest of the doc strings of the other methods and do the same? Because none of them are currently like that.

zachmullen:
That would be great 😄

tests/cases/resource_test.py
```diff
@@ -455,10 +455,26 @@ def testGetResourceByPath(self):
                 str(resp.json['_id']), str(item['_id']))
 
         # test bogus path
+        # test is not set
         resp = self.request(path='/resource/lookup',
                             method='GET', user=self.user,
-                            params={'path': '/bogus/path'})
+                            params={'path': '/collection/bogus/path'})
```
zachmullen: Can you change this back to just `/bogus/path`? It was covering a branch that is no longer covered after this change.

sutartmelson: the '/bogus/path' isn't really checking for a path that doesnt exists though, it just instantly fails as it is not in the proper path format. So maybe have both? One that checks for a bogus path, and one that checks for improper path formatting?

zachmullen: The two requests below test the case of invalid collection paths, this one was testing [this branch](https://github.com/girder/girder/blob/master/girder/api/v1/resource.py#L204). I think just changing it back to `/bogus/path` covers all of our bases.

tests/cases/resource_test.py
```diff
@@ -458,20 +458,20 @@ def testGetResourceByPath(self):
...
         resp = self.request(path='/resource/lookup',
                             method='GET', user=self.user,
-                            params={'path': '/collection/bogus/path'})
+                            params={'path': '/bogus/path'})
         self.assertStatus(resp, 400)
         # test is set to false, response code should be 400
         resp = self.request(path='/resource/lookup',
                             method='GET', user=self.user,
-                            params={'path': '/collection/bogus/path',
+                            params={'path': '/bogus/path',
```
zachmullen: Sorry for the confusion, what I meant in my earlier suggestion was to keep lines 466 and 474 as `/collection/bogus/path`, but only revert the one on line 461 to `/bogus/path` 😄 

Alternatively, you could just add another block altogether to test `/bogus/path` if you wanted to still test `/collection/bogus/path` using the default value for the `test` parameter.

zachmullen:
LGTM, thanks!

mgrauer:
Nice, thanks @xerxes455 .

PR #1396, opened 2016-06-21
Display the login name alongside first and last names

ronichoudhury:
This is helpful for distinguishing multiple accounts created under a single name.

### DIFF
--- a/clients/web/src/stylesheets/layout/header.styl
+++ b/clients/web/src/stylesheets/layout/header.styl
@@ -20,11 +20,12 @@
   margin-right 12px
   vertical-align top
   margin-top (($headerHeight - 30px) / 2)
+  width $quickSearchWidth
 
 .g-current-user-wrapper
   display inline
   float right
 
-@media (max-width: 500px)
+@media (max-width: 720px)
   .g-app-title
     display none
--- a/clients/web/src/stylesheets/layout/headerUser.styl
+++ b/clients/web/src/stylesheets/layout/headerUser.styl
@@ -3,8 +3,13 @@
 .g-user-text
   line-height $headerHeight
 
-  .g-user-dropdown-link
+  a
     color white
+
+    &:hover
+      color #e0e0e0
+
+  .g-user-dropdown-link
     display inline-block
     max-width calc(100vw - 300px)
     overflow hidden
@@ -13,8 +18,6 @@
 
     i
       margin-left 2px
-    &:hover
-      color #e0e0e0
 
   .g-or-text
     vertical-align top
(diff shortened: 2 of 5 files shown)


fogleman:
Could also drop in a `i.icon-user` before the name and it'd look like this. Just a thought. :)

screen shot 2016-06-21 at 1 19 44 pm

zachmullen:
If you make the screen really narrow, does overflow stuff happen? I think that's mostly why I've avoided adding more text there.

ronichoudhury:
At some point, the name text simply disappears. I'd expect it to become one
of those ugly hamburger icons instead (to maintain the ability to open the
menu).

roni

zachmullen:
It looks like it actually flows outside the black area and just wonkily floats underneath. I think we should figure out a way to handle that text overflow. A couple options would be:

1. Just show the user icon and login name, and put the first and last name inside the dropdown as a header
2. If gravatar plugin is enabled, hide everything but the gravatar and the little dropdown icon.

ronichoudhury:
I see the wonky floating now.

I'm not sure (1) actually fixes the problem; at some point you will still trigger the overflow behavior.

zachmullen:
> I'm not sure (1) actually fixes the problem; at some point you will still trigger the overflow behavior.

I left this off of my original comment, but I think we need to add text-overflow styling such as ellipsis in the case of (1).

ronichoudhury:
That being said, I can certainly implement (1) now and we can work on a more thorough solution in a separate PR. I would need some help with (2) because I've never used the gravatar stuff.

zachmullen:
That would be great

ronichoudhury:
Ok I'll give ellipsis-truncation a go :racehorse:

ronichoudhury:
@zachmullen, I added a commit that implements (1) but I poked around with text-overflow and ellipsis et al. and couldn't figure out how to make it work for Girder's navbar. Could you or someone else take a look at that?

zachmullen:
Sure, I'll give it a shot

zachmullen:
I got the ellipsis working with some CSS3 hackage, however I removed the user fullname header from the dropdown list for now; it looks nice, but unfortunately broke assumptions for a common way that our plugins extend that menu, which is by prepending:

screen shot 2016-06-22 at 10 53 55 am

I've got a good idea for how to fix that in v2.0 to make extending the menu easier.

fogleman:
Haven't looked at the DOM, but can the full name go outside of whatever container the plugins are prepending to?

zachmullen:
Let's just say that's nontrivial :)

ronichoudhury:
LGTM!

PR #1397, opened 2016-06-21
Do not return full user document with API key token creation

zachmullen:
ping @ronichoudhury 

I considered an alternate solution where we conditionally include the full user document only if the token has the USER_INFO_READ permission. If you'd prefer that behavior, I can change it to do that instead.


ronichoudhury:
LGTM

PR #1400, opened 2016-06-21
Reduce number of CI builds

zachmullen:
This removes a couple of CI matrix indexes that seem redundant;
I don't see much benefit doing the client tests in both python
2 and 3, since all the python code assertions are supposed to be
made by the server tests. Also, the "other" category of tests do
not depend on python at all.


zachmullen:
@jbeezley I wanted to get your thoughts on this. If we are OK with this, it would have two benefits:

1. Cutting the number of spurious test failures by almost half.
2. Reducing CI response time; we have so many parallel builds happening now and we are rate limited in terms of both overall builds per unit time and simultaneous builds, so cutting it down from 8 to 6 will make our queue wait times significantly shorter.

Also, I'd like to propose changing the category names a little:

* "server" --> "python" (to indicate that the main dependency here is the python environment, since this set also includes the python client tests)
* "client" -> "browser"

Having these categories correspond more closely with the associated technology stack will make it easier for future developers to decide where to put novel test types.

jbeezley:
LGTM

brianhelba:
+1

Is there a rationale for the decision to use Py2 as the environment for "server  and "client", as opposed to Py3?

cpatrick:
Is anyone using Py3 and girder in production?

.travis.yml
```diff
@@ -1,14 +1,21 @@
 language: python
 python:
-    - "2.7"
-    - "3.4"
+    - 2.7
+    - 3.4
```
Xarthisius: Is there any reason it's `3.4` rather than `3.5`?

zachmullen: We want to test on our minimum py3 version, which is 3.4. Since nothing should be broken from 3.4 to 3.5, we assume that it works on 3.5 also.

zachmullen:
> Is there a rationale for the decision to use Py2 as the environment for "server and "client", as opposed to Py3?

Nope, just arbitrary.

PR #1403, opened 2016-06-22
Force compliant JSON output from REST layer

zachmullen:
Infinity, -Infinity and NaN are valid literal tokens in
javascript, but there is no way to represent them in JSON,
so we disallow their usage and leave it up to the route
handlers themselves to determine how to encode such values.

@cpatrick PTAL


cpatrick:
LGTM. Nice test.

cpatrick:
I will say that this does _technically_ break the API by producing valid JSON, but only for a broken edge case.

PR #1405, opened 2016-06-22
Fix /resource/search TypeError

### LINKED ISSUE #1404: Error using /resource/search for models that aren't access controlled
Using the /resource/search endpoint to search for models that aren't access controlled gives an error like:

    TypeError: TypeError("prefixSearch() got an unexpected keyword argument 'user'")

This happens because the endpoint calls the model search method with keyword arguments that only exist on, for example, AccessControlledModel.prefixSearch(), and not Model.prefixSearch(); see https://github.com/girder/girder/blob/ccd2b162ca315931190fec5b958c4409b0d0c2cb/girder/api/v1/resource.py#L101-L103.

msmolens:
This commit fixes an error when using the /resource/search endpoint to search
for models that aren't access controlled. The error message is similar to:

    TypeError: TypeError("prefixSearch() got an unexpected keyword argument 'user'")

Fixes #1404

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -255,7 +255,7 @@ class Model(ModelImporter):
         return self.collection.find_one(query, projection=fields, **kwargs)
 
     def textSearch(self, query, offset=0, limit=0, sort=None, fields=None,
-                   filters=None):
+                   filters=None, **kwargs):
         """
         Perform a full-text search against the text index for this collection.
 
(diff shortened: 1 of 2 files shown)


brianhelba:
LGTMITIH.

Note, the order and available parameters between the `prefixSearch` and `textSearch` methods of `Model` and `AccessControlledModel` still breaks polymorphism, but I guess fixing this would be a sometimes-backwards-incompatible change to the model API.

PR #1407, opened 2016-06-22
Add download and inline-view links to item list

ronichoudhury:
This adds download and inline-view links to each item in an item list, enabling to, e.g., open a sequence of images without having to click into each item and then back out again.

### DIFF
--- a/clients/web/src/templates/widgets/itemList.jade
+++ b/clients/web/src/templates/widgets/itemList.jade
@@ -3,9 +3,13 @@ ul.g-item-list
     li.g-item-list-entry
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-item-cid="#{item.cid}")
-      a.g-item-list-link(g-item-cid="#{item.cid}")
+      a.g-item-list-link.g-right-border(g-item-cid="#{item.cid}")
         i.icon-doc-text-inv
         | #{item.get('name')}
+      a.g-right-border(href=item.downloadUrl())
+        i.icon-download
+      a.g-view-inline(target="_blank" href=item.downloadUrl({contentDisposition: 'inline'}))
+        i.icon-eye
       .g-item-size #{girder.formatSize(item.get('size'))}
   if (hasMore)
     li.g-show-more
--- a/clients/web/src/templates/widgets/itemList.jade
+++ b/clients/web/src/templates/widgets/itemList.jade
@@ -6,7 +6,7 @@ ul.g-item-list
       a.g-item-list-link.g-right-border(g-item-cid="#{item.cid}")
         i.icon-doc-text-inv
         | #{item.get('name')}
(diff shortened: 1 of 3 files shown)


ronichoudhury:
ping @zachmullen

clients/web/src/templates/widgets/itemList.jade
```diff
@@ -3,9 +3,13 @@ ul.g-item-list
     li.g-item-list-entry
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-item-cid="#{item.cid}")
-      a.g-item-list-link(g-item-cid="#{item.cid}")
+      a.g-item-list-link.g-right-border(g-item-cid="#{item.cid}")
         i.icon-doc-text-inv
         | #{item.get('name')}
+      a.g-right-border(href=item.downloadUrl())
```
zachmullen: I think we should remove the `g-right-border` class here so there's just one separator instead of two.

zachmullen:
Could you add tooltips similar to the ones we have for files?

ronichoudhury:
@zachmullen PTAL

fogleman:
LGTM.

If an item has multiple files, the item will be a zip download, so we might want to hide the "View in browser" icon in that case, but that can be worried about later IMO.

ronichoudhury:
It had occurred to me to do that but I guess it got away from me. Standby, commit coming...

clients/web/src/templates/widgets/itemList.jade
```diff
@@ -3,9 +3,15 @@ ul.g-item-list
     li.g-item-list-entry
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-item-cid="#{item.cid}")
-      a.g-item-list-link(g-item-cid="#{item.cid}")
+      a.g-item-list-link.g-right-border(g-item-cid="#{item.cid}")
         i.icon-doc-text-inv
         | #{item.get('name')}
+      a(title="Download item" href=item.downloadUrl())
+        i.icon-download
+      if item.collection.length === 1
```
fogleman: This doesn't work. `item.collection` is the collection of items that this item belongs to, not its children files.

ronichoudhury:
@fogleman @zachmullen let me know if this is ok.

clients/web/src/templates/widgets/itemList.jade
ronichoudhury: Ah, right. How do I get a list of files contained in the item?

fogleman: I don't think you can easily - I say forget about it for now.

ronichoudhury: Ok, I'll zap that commit out hang on.

roni

fogleman:
LGTMITIH

PR #1408, opened 2016-06-23
Registration approval

fogleman:
@zachmullen @mgrauer @cpatrick 

This PR adds a new "Registration policy" option called "Admin approval required"

In this mode, users can register themselves but cannot login until an admin approves their account.

As an added bonus, admins can also "disable" user accounts to prevent them from logging in. This may be useful for the case where you don't want to delete the user account but don't want them logging in anymore.

The user list is updated so admins can see users with disabled or pending approval status:

screen shot 2016-06-23 at 9 58 29 am

To approve an account, the admin clicks on a user and then clicks Actions > Approve account as shown here:

screen shot 2016-06-23 at 9 59 58 am

Admins are emailed when a new user registers, with a link to go directly to this page.

Once approved, the user is notified by email that they can now login.

### DIFF
--- a/clients/web/src/stylesheets/body/userList.styl
+++ b/clients/web/src/stylesheets/body/userList.styl
@@ -17,6 +17,10 @@
   .g-space-used
     margin-top 2px
 
+  .g-user-status
+    margin-top 2px
+    color #d9534f
+
   .g-user-right-column
     border-left 1px solid #ddd
     padding-left 8px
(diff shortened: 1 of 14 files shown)


cpatrick:
Code LGTM, but I think it would be better to have a single-click approval from the user list for admins.

fogleman:
That would be nice, I guess I was trying not to change the UI too much. How important do you think it is?

cpatrick:
Not terribly. I'd be content with a follow-up ticket for someone to grab later.

fogleman:
See #1423

PR #1409, opened 2016-06-23
Add extraParameters parameter to FilesystemAssetstoreAdapter.downloadFile()

ronichoudhury:
This should have been done in #1375.


zachmullen:
LGTM

I've already forgotten, what was your use case for these extra params?

ronichoudhury:
We're looking to implement specialized semantic behavior for certain kinds of files (such as, e.g., CSV); something along the lines of database-style offset/limit/filtering access to the data.

PR #1410, opened 2016-06-23
Don't show exact byte count twice

### LINKED ISSUE #1401: Don't show exact byte count twice
I.e., this case:

screen shot 2016-06-22 at 9 40 04 am

fogleman:
Fixes #1401 @zachmullen

### DIFF
--- a/clients/web/src/templates/body/itemPage.jade
+++ b/clients/web/src/templates/body/itemPage.jade
@@ -37,7 +37,7 @@
     | Info
   .g-item-size.g-info-list-entry
     i.icon-hdd
-      if item.get('size') < 20000
+      if item.get('size') < 1024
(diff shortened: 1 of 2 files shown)


zachmullen:
I've never been too fond of our convention of using 20000 as the threshold... Is this a common thing? ping @manthey

fogleman:
I just used what was already in place. :shrug: Seems like anything over 1kB could be formatted.

zachmullen:
I would prefer just using 1024 as the max number of bytes before we reformat, just wanted to see if there is some logic behind the choice of 20000.

fogleman:
I doubt there's a specific reason, so I went ahead and made the change.

zachmullen:
LGTM

PR #1413, opened 2016-06-24
Add originalItemId flag to copied items

### LINKED ISSUE #1412: Add provenance information to copied items
Specifically, when [`copyItem`](https://github.com/girder/girder/blob/master/girder/api/v1/item.py#L295) is called, a reference to the original item's id is stored.

alex-r-bigelow:
Fix #1412

I *think* this is probably best stored in the root level, as opposed to inside `'meta'`, but feel free to correct me.


zachmullen:
This is indeed the way to do it. I think we should maybe change the field name, though, to indicate that it was a copy. Maybe `copyOfItem`?

zachmullen:
Great, could you just add an assertion into [this test](https://github.com/girder/girder/blob/master/tests/cases/item_test.py#L556)?

zachmullen:
LGTM

PR #1414, opened 2016-06-24
Attempt to fix race condition in web_client_folder test

zachmullen:
This is the one that's failing the most often


zachmullen:
I've rerun the browser test suite several times now on travis and these tests haven't failed yet, any objections to merging this?

fogleman:
LGTM

PR #1415, opened 2016-06-24
Auto join plugin

fogleman:
This PR adds a new plugin called "Auto Join".

The Auto Join plugin allows you to define rules to automatically assign new users to groups based on their email address domain. Typically, this is used in conjunction with email verification.

Currently, it's a case-insensitive substring match (not a regex).

screen shot 2016-06-24 at 3 08 16 pm

Tests not implemented yet (WIP), but it works.


fogleman:
@zachmullen @cpatrick Ready for review.

cpatrick:
I believe the string matching is fine and likely more straightforward for many users. LGTM.

PR #1419, opened 2016-06-27
Bump py client version to 1.2.1

zachmullen:


zachmullen:
Thanks @law12019 for reporting the related bug, we hadn't released the python client since @jbeezley fixed our minimum required requests version.

PR #1420, opened 2016-06-28
Fix race condition in curation plugin test

fogleman:
@zachmullen 

Since the dialog refreshes after a REST request, doing two calls to waitForDialog seems to fix the issue. I tested by making the curation endpoints sleep for one second before responding.


zachmullen:
LGTMITIH

Issue #1421, opened 2016-06-28
"Copy to clipboard" widget


zachmullen:
This should be an inline icon link that we wrap into a reusable widget which, when clicked, copies the specified text to the user's clipboard and maybe shows an alert indicating a successful copy. Example use case:

```
new girder.views.ClipboardCopyWidget({
    el: '.g-copy-link-container',
    parentView: this,
    text: someResource.id
}).render();
```

zachmullen:
This should probably have three modes:

1. `raw` (default). Whatever is passed in as the target is assumed to be a string and is copied on click.
2. `el`: Whatever is passed in is an HTML element and when this widget is clicked, we copy its `$(target).text()` value.
3. `input`: Whatever is passed in is an HTML input tag and its `$(target).val()` value is copied.

So, e.g.

```
new girder.views.ClipboardCopyWidget({
    el: this.$('.g-copy-link-container'),
    parentView: this,
    target: this.$('.g-some-input-field'),
    mode: 'input'
})
```

PR #1428, opened 2016-06-28
Expose checksum keyfiles and file info in the UI

zachmullen:
ping @cdeepakroy 

This still needs testing before it gets merged.

### DIFF
--- a/clients/web/src/views/widgets/FileListWidget.js
+++ b/clients/web/src/views/widgets/FileListWidget.js
@@ -7,6 +7,22 @@ girder.views.FileListWidget = girder.View.extend({
             this.collection.fetchNextPage();
         },
 
+        'click a.g-show-info': function (e) {
+            var cid = $(e.currentTarget).attr('file-cid');
+            new girder.views.FileInfoWidget({
+                el: $('#g-dialog-container'),
+                model: this.collection.get(cid),
+                parentView: this
+            }).on('g:uploadFinished', function () {
+                girder.events.trigger('g:alert', {
+                    icon: 'ok',
+                    text: 'File contents updated.',
+                    type: 'success',
+                    timeout: 4000
+                });
+            }, this).render();
+        },
+
         'click a.g-update-contents': function (e) {
             var cid = $(e.currentTarget).parent().attr('file-cid');
             this.uploadDialog(cid);
--- a/clients/web/src/views/widgets/FileListWidget.js
+++ b/clients/web/src/views/widgets/FileListWidget.js
@@ -13,14 +13,7 @@ girder.views.FileListWidget = girder.View.extend({
                 el: $('#g-dialog-container'),
                 model: this.collection.get(cid),
                 parentView: this
-            }).on('g:uploadFinished', function () {
-                girder.events.trigger('g:alert', {
-                    icon: 'ok',
-                    text: 'File contents updated.',
-                    type: 'success',
-                    timeout: 4000
-                });
-            }, this).render();
+            }).render();
         },
 
         'click a.g-update-contents': function (e) {
--- a/plugins/hashsum_download/server/__init__.py
+++ b/plugins/hashsum_download/server/__init__.py
@@ -95,8 +95,8 @@ class HashedFile(File):
         Print an exception if a user requests an invalid checksum algorithm.
         """
         if algo not in self.supportedAlgorithms:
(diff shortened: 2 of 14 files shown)


cdeepakroy:
@zachmullen Can you also add the URL template for data.kitware.com [here](https://github.com/girder/girder/blob/master/tests/TestData.cmake#L15-L17) and test if it works? 

Rest LGTM!

clients/web/src/views/widgets/FileListWidget.js
```diff
@@ -7,6 +7,22 @@ girder.views.FileListWidget = girder.View.extend({
...
         },
 
+        'click a.g-show-info': function (e) {
+            var cid = $(e.currentTarget).attr('file-cid');
+            new girder.views.FileInfoWidget({
+                el: $('#g-dialog-container'),
+                model: this.collection.get(cid),
+                parentView: this
+            }).on('g:uploadFinished', function () {
+                girder.events.trigger('g:alert', {
```
brianhelba: Who is listening for this `'g:alert'` event?

zachmullen: Thanks for finding this, this is cruft left over from copypasta.

To answer your question, that's a global event that can be triggered anywhere in the app, the handler shows a popup with a message in the upper right corner for some period of time.

plugins/hashsum_download/server/__init__.py
```diff
@@ -88,4 +128,6 @@ def _getFirstFileByHash(self, algo, hash, user=None):
 
 
 def load(info):
-    HashedFile(info['apiRoot'])
+    HashedFile(info['apiRoot'].file)
+    ModelImporter.model('file').exposeFields(
+        level=AccessType.READ, fields=HashedFile.supportedAlgorithms)
```
brianhelba: Note, this means that as future hash types are added, they'll be tacked on to the root of file objects, as opposed to nested in some sub-field. This isn't really a problem though.

plugins/hashsum_download/web_client/templates/hashsum_download_fileInfo.jade
```diff
@@ -0,0 +1,6 @@
+if file.has('sha512')
```
brianhelba: As additional hashsum types are added, they'll have to be known ahead of time by the client, rather than looping through every member of an embedded field, I guess.

plugins/hashsum_download/web_client/templates/hashsum_download_fileInfo.jade
```diff
@@ -0,0 +1,6 @@
...
+  .g-file-info-line(property="sha512")
+    i.icon-barcode
+    span SHA-512:
+    span.g-bold-part
+      a(href=keyfileUrl(file.id, 'sha512'))  #{abbreviate(file.get('sha512'))}
```
brianhelba: Could we have an easy way for a user to select this hash value (for example, to copy to the clipboard) without having to download the file?

zachmullen: I do plan to add that once we resolve #1421

brianhelba: How about just a read-only text box right now, so the text can easily be selected, but doesn't take up too much horizontal space?

plugins/hashsum_download/web_client/templates/hashsum_download_fileInfo.jade
zachmullen: That would be more elegant. Unfortunately this plugin is taking advantage of a data model that predates it. That data model was not created with consideration for this use case 😄

brianhelba: I agree. The one possible workaround is for the API document to be look different from the internal model document. I think that violates too many other Girder expectations though.

brianhelba:
LGTM, just needs testing.

@zachmullen I can quickly add the server-side testing, if you aren't already working on it right now. I'll need help for the client-side.

zachmullen:
Server-side testing is done, I'm just writing some client tests now

zachmullen:
Testing finished, this is ready for final QA

brianhelba:
LGTMWTIH.

brianhelba:
Please deploy to [data.kitware.com](data.kitware.com) when this is merged.

zachmullen:
It is now deployed there.

PR #1429, opened 2016-06-29
Allow plugins to override JSON encoding

cjh1:


zachmullen:
Hm, I'm trying to think of some other way to resolve the import cycle that doesn't break compatibility. Any good ideas?

zachmullen:
(One obvious solution is deferring an import into a function body somewhere, but if we could avoid doing that it would be cleaner.)

cjh1:
Yes, I thought of that, not very nice. I guess I didn't realize that the JsonEncoder is part of the public interface?

zachmullen:
Essentially any symbol within the girder package that isn't prefixed with an underscore is fair game for downstreams to use. 😄

cjh1:
Well we could leave the class in utility ( deprecate it if we want ) and create a copy in the new module.

zachmullen:
I'm fine with that solution, so long as we mark the old version as deprecated.

cjh1:
This is another alternative. In girder/event.py instead of doing ```from girder import logger``` to just ``import girder``` and change referenced to ```logger``` to ```girder.logger```, then the move it no longer required. Thoughts?

zachmullen:
I don't think that will work since the logger is created inside girder's `__init__.py`

cjh1:
It seems to, its the fact the logger is referenced in the import ( before it created ) that causes the problem.

zachmullen:
Ah, you're right. That works then, let's do it.

zachmullen:
LGTM

PR #1430, opened 2016-06-29
Support file uploads in homepage markdown widget

### LINKED ISSUE #1374: Add the ability to upload a file in the homepage plugin
The folder markdown widget supports this, but provides that folder as a target for the upload.

An upload target would need to be specified or provided by the plugin (perhaps a new private collection named "homepage", a new folder under that user's space, etc).

follow up to #1370.

fogleman:
Fixes #1374

@brianhelba

### DIFF
--- a/girder/models/collection.py
+++ b/girder/models/collection.py
@@ -102,7 +102,8 @@ class Collection(AccessControlledModel):
             progress.update(increment=1, message='Deleted collection ' +
                             collection['name'])
 
-    def createCollection(self, name, creator, description='', public=True):
+    def createCollection(self, name, creator=None, description='', public=True,
+                         reuseExisting=False):
         """
         Create a new collection.
 
@@ -114,22 +115,33 @@ class Collection(AccessControlledModel):
         :type public: bool
         :param creator: The user who is creating this collection.
         :type creator: dict
+        :param reuseExisting: If a collection with the given name already exists
+            return that collection rather than creating a new one.
+        :type reuseExisting: bool
         :returns: The collection document that was created.
         """
+        if reuseExisting:
+            existing = self.findOne({
+                'name': name
+            })
+            if existing:
+                return existing
+
         now = datetime.datetime.utcnow()
 
         collection = {
             'name': name,
             'description': description,
-            'creatorId': creator['_id'],
+            'creatorId': creator['_id'] if creator else None,
             'created': now,
             'updated': now,
             'size': 0
         }
 
         self.setPublic(collection, public, save=False)
-        self.setUserAccess(collection, user=creator, level=AccessType.ADMIN,
-                           save=False)
+        if creator:
+            self.setUserAccess(
+                collection, user=creator, level=AccessType.ADMIN, save=False)
 
         return self.save(collection)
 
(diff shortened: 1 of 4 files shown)


plugins/homepage/server/__init__.py
```diff
@@ -36,14 +38,27 @@ def __init__(self):
...
     if event.info['key'] == KEY:
         event.preventDefault().stopPropagation()
 
 
+def getOrCreateAssetsFolder():
+    collection = ModelImporter.model('collection').createCollection(
+        NAME, public=False, reuseExisting=True)
+    folder = ModelImporter.model('folder').createFolder(
+        collection, NAME, parentType='collection', public=False,
+        reuseExisting=True)
```
zachmullen: Shouldn't the folder at least be public so that everyone can see the images?

plugins/homepage/server/__init__.py
```diff
@@ -21,8 +21,10 @@
 from girder.api import access
 from girder.api.describe import Description, describeRoute
 from girder.api.rest import Resource
+from girder.utility.model_importer import ModelImporter
 
 KEY = 'homepage.markdown'
+NAME = 'Homepage Assets'
```
brianhelba: These are typically in a `constants.py` file, as members of a class. It's probably not as big of a deal for a plugin this size.

plugins/homepage/server/__init__.py
brianhelba: Indeed. Make the collection private and the folder public. Then everyone can download from it without seeing it in their listing.

girder/models/collection.py
```diff
@@ -114,22 +115,33 @@ def createCollection(self, name, creator, description='', public=True):
...
+            if existing:
+                return existing
+
         now = datetime.datetime.utcnow()
 
         collection = {
             'name': name,
             'description': description,
-            'creatorId': creator['_id'],
+            'creatorId': creator['_id'] if creator else None,
```
zachmullen: This introduces the notion of collections with no creator, so we'll definitely want to add some testing since that's a new feature.

plugins/homepage/server/__init__.py
fogleman: Oops, fixing.

brianhelba:
Feature-wise, this LGTM. The collection having no creator needs to be resolved one way or another (as a supported feature with testing, or by making it owned by some arbitrary site admin).

fogleman:
Doesn't look like collection creator is really used for much of anything. I can add a test making sure the creator can now be `None`, but not sure what else to test?

brianhelba:
I share @zachmullen hesitation about the implications of allowing collections to have no creator, but I'll let him make the call on this.

zachmullen:
The alternative is to have the asset folder be a configurable system setting and only allow uploading in the widget if the value is set

fogleman:
Folders can also have no creator, that was already the case. I don't see any real implications.

zachmullen:
I'm honestly not sure what assumptions this might break. It clearly doesn't break anything just by calling `createCollection`, if a problem exists it would come up somewhere else where the collection was being passed in and someone was assuming it has a creatorId. It's pretty hard to figure out how to test that, though, so I guess it's OK as is.

fogleman:
LGTY?

zachmullen:
LGTM

PR #1433, opened 2016-06-29
Fixes #1432. Allow test harnessing of individual classes or methods

### LINKED ISSUE #1432: Allow add_python_test to harness individual classes or test functions
Unittest supports this via the command line.

zachmullen:
@cdeepakroy @brianhelba This addresses what we just discussed on the call. Here's an example of using this feature, shown on one of our core tests. Before:

    add_python_test(api_key)

After (to test a single method):

    add_python_test(api_key NAME api_key.listKeys SUBMODULE ApiKeyTestCase.testListKeys)

After (to test a single class);

    add_python_test(api_key NAME api_key.ApiKeyTestCase SUBMODULE ApiKeyTestCase)

### DIFF
--- a/tests/PythonTests.cmake
+++ b/tests/PythonTests.cmake
@@ -83,7 +83,7 @@ function(add_python_test case)
   set(name "server_${case}")
 
   set(_options BIND_SERVER PY2_ONLY)
-  set(_args NAME PLUGIN SUBMODULE)
+  set(_args PLUGIN SUBMODULE)
   set(_multival_args RESOURCE_LOCKS TIMEOUT EXTERNAL_DATA)
   cmake_parse_arguments(fn "${_options}" "${_args}" "${_multival_args}" ${ARGN})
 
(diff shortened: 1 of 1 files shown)


zachmullen:
I decided to keep the automatic naming convention, so my example would actually just be:

    add_python_test(api_key SUBMODULE ApiKeyTestCase.testListKeys)

Which would create a test called `server_api_key.ApiKeyTestCase.testListKeys`.

cdeepakroy:
Let me test this tomorrow and let you know.

Maybe we should also add a test in girder proper that uses this cmake function with the `SUBMODULE` option.

zachmullen:
@cdeepakroy I added the option we discussed, PTAL.

cdeepakroy:
@zachmullen I confirm that it works for both `class-level` and `a-method-in-class-level`.

The only issue is when i run `ctest -N` the names specified with the `DBNAME` parameter do not show up. Instead it shows what you provide for the `SUBMODULE` parameter. Not sure if this is what you intended. But I like the fact that i shows the longer version. Maybe you want to display both when the test runs to ease debugging.

zachmullen:
That's intentional, the DBNAME can just be used to control the database name in case it would otherwise be too long.

I actually pushed a patch to CTest that causes all environment variables to be printed when the verbose flag is on during test execution, but it's only available in very new version of CTest at the moment.

zachmullen:
@cdeepakroy is this OK to merge?

cdeepakroy:
Yes. Please merge.

PR #1434, opened 2016-06-29
[WIP] Give item and folder links correct href property

ronichoudhury:
This allows for item and folder links to be, e.g., middle-clicked to open them in a new window, etc.

### DIFF
--- a/clients/web/src/templates/widgets/folderList.jade
+++ b/clients/web/src/templates/widgets/folderList.jade
@@ -3,7 +3,7 @@ ul.g-folder-list
     li.g-folder-list-entry(public=(folder.get('public') ? 'true' : 'false'))
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-folder-cid=folder.cid)
-      a.g-folder-list-link(g-folder-cid=folder.cid, href="#{pathname}#{baseHash}/folder/#{folder.id}")
+      a.g-folder-list-link(g-folder-cid=folder.cid, href="#folder/#{folder.id}")
         i.icon-folder
         = folder.get('name')
         i.icon-right-dir
--- a/clients/web/src/templates/widgets/folderList.jade
+++ b/clients/web/src/templates/widgets/folderList.jade
@@ -3,7 +3,7 @@ ul.g-folder-list
     li.g-folder-list-entry(public=(folder.get('public') ? 'true' : 'false'))
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-folder-cid=folder.cid)
-      a.g-folder-list-link(g-folder-cid=folder.cid)
+      a.g-folder-list-link(g-folder-cid=folder.cid href="#{pathname}#{baseHash}/folder/#{folder.id}")
         i.icon-folder
         = folder.get('name')
         i.icon-right-dir
--- a/clients/web/src/templates/widgets/folderList.jade
+++ b/clients/web/src/templates/widgets/folderList.jade
@@ -3,7 +3,7 @@ ul.g-folder-list
     li.g-folder-list-entry(public=(folder.get('public') ? 'true' : 'false'))
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-folder-cid=folder.cid)
-      a.g-folder-list-link(g-folder-cid=folder.cid, href="#folder/#{folder.id}")
+      a.g-folder-list-link(g-folder-cid=folder.cid, href=(navigate ? "#folder/#{folder.id}" : false))
         i.icon-folder
         = folder.get('name')
         i.icon-right-dir
--- a/clients/web/src/templates/widgets/folderList.jade
+++ b/clients/web/src/templates/widgets/folderList.jade
@@ -3,7 +3,7 @@ ul.g-folder-list
     li.g-folder-list-entry(public=(folder.get('public') ? 'true' : 'false'))
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-folder-cid=folder.cid)
-      a.g-folder-list-link(g-folder-cid=folder.cid href="#{pathname}#{baseHash}/folder/#{folder.id}")
+      a.g-folder-list-link(g-folder-cid=folder.cid, href="#{pathname}#{baseHash}/folder/#{folder.id}")
         i.icon-folder
         = folder.get('name')
         i.icon-right-dir
--- a/clients/web/src/views/widgets/HierarchyWidget.js
+++ b/clients/web/src/views/widgets/HierarchyWidget.js
@@ -61,6 +61,7 @@ girder.views.HierarchyWidget = girder.View.extend({
         this.folderEdit = settings.folderEdit;
         this.itemCreate = settings.itemCreate;
         this.breadcrumbs = [this.parentModel];
+        this.navigate = settings.navigate || false;
 
         // Initialize the breadcrumb bar state
         this.breadcrumbView = new girder.views.HierarchyBreadcrumbView({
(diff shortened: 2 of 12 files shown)


ronichoudhury:
Handling item links was straightforward and clean. Doing the same for folder links, however, was a bit messier. I had to use global state to propagate URL components to the href property in a way that is clearly brittle if the URL scheme ever changes (or, if the folder list widget were to be used outside the context of the hierarchy widget, etc.). Let me know if it's unacceptable, or if there's a better way.

clients/web/src/templates/widgets/folderList.jade
```diff
@@ -3,7 +3,7 @@ ul.g-folder-list
     li.g-folder-list-entry(public=(folder.get('public') ? 'true' : 'false'))
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-folder-cid=folder.cid)
-      a.g-folder-list-link(g-folder-cid=folder.cid)
+      a.g-folder-list-link(g-folder-cid=folder.cid href="#{pathname}#{baseHash}/folder/#{folder.id}")
```
zachmullen: For strict jade we should actually put a comma between these two attributes.

zachmullen:
> Doing the same for folder links, however, was a bit messier. I had to use global state to propagate URL components to the href property in a way that is clearly brittle if the URL scheme ever changes

Actually, a lot of people have complained that the user or collection stays visible even when you start dropping down into subfolders, so I think it would actually be fine to just go straight to `#folder/{id}`.

clients/web/src/templates/widgets/folderList.jade
ronichoudhury: Huh I thought strict jade omitted the commas. Will fix.

ronichoudhury:
I was wondering why the collection/user needed to be there. If we can simplify that out that would be great. Would you like me to look into changing the global router so left-clicks also result in the correct URL?

zachmullen:
That would be great. We can leave the old route handling in place so that we don't break existing URLs, but I'm fine with doing away with it during navigation.

ronichoudhury:
I believe I've addressed both concerns. I didn't have to touch any routing logic, as URLs just mentioning a folder id seem to work properly.

zachmullen:
@ronichoudhury could you resolve conflicts on this branch so we can merge?

ronichoudhury:
Done; just waiting for Travis.

The code/results look ok to you?

zachmullen:
Not sure about this actually. I think the reason it was left as an event handler was so that downstreams using the FolderListWidget and ItemListWidget could use it for more than just navigation, e.g. selection of an item or folder, which is why they trigger an event when one is clicked. So, I think we need to make this functionality conditional based on an initialization option for the widget, and ideally, make it able to be passed down through the HierarchyWidget. Maybe `this.navigate = settings.navigate || false;`?

ronichoudhury:
@zachmullen was that last comment (about the widgets) meant to be a line comment?

zachmullen:
It applies to both line changes that you made, and my suggestion in the last sentence was meant to refer to the `initialize` methods of the HierarchyWidget, FolderListWidget, and ItemListWidget.

ronichoudhury:
Ah ok I think I get it.

Just FYI, I'll be out starting in a few minutes, until next Monday. I'll be sure to take one more look at this when I get back. In the meantime I'll mark this PR as WIP.

ronichoudhury:
@zachmullen PTAL. I think I did this right, but I had to change code in a bunch of places (which might be a code smell independent of this PR).

Essentially, this makes the client change to use href'ed navigation links, but leaves it up to the classes mentioned above to enable/disable this behavior.

zachmullen:
That looks about right. We should also change near line 55 of HierarchyWidget.js so that the onItemClick default binding is a no-op rather than doing the navigation, so we don't do the navigation twice.

ronichoudhury:
So I use a noop as the default action if navigation is on; otherwise it defaults to the function that was there before.

clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -46,15 +46,17 @@ girder.views.HierarchyWidget = girder.View.extend({
...
 
         this._showActions = _.has(settings, 'showActions') ? settings.showActions : true;
         this._showItems = _.has(settings, 'showItems') ? settings.showItems : true;
         this._checkboxes = _.has(settings, 'checkboxes') ? settings.checkboxes : true;
         this._routing = _.has(settings, 'routing') ? settings.routing : true;
         this._appendPages = _.has(settings, 'appendPages') ? settings.appendPages : false;
-        this._onItemClick = settings.onItemClick || function (item) {
+        this._onItemClick = settings.onItemClick || (
+          this.navigate ? this.$.noop : function (item) {
             girder.router.navigate('item/' + item.get('_id'), {trigger: true});
```
zachmullen: Weird, this should be causing our eslint_core test to fail, and it's doing that for me locally, but not in travis. I'm going to investigate that...

zachmullen: Nevermind, it is catching them.

zachmullen:
This is causing a bunch of failing tests after I merged master into it. Looking at the screenshots of the failures, it looks like the navigation down to the item view isn't working in the tests.

zachmullen:
It turns out that what's breaking our test is that in the version of phantomjs we are using (1.9.20, the last 1.x release), calling `click()` on an anchor tag with an href attribute that points to a hash does not actually work as expected (the hash state does not update).

There are a couple options to sort this out:

1. In testUtils.js, override `$.fn.click` to call the normal implementation, and then if the element is an `<a>` with a `href` that starts with `#`, we call `window.location.hash = ...`.
2. Attempt to upgrade to phantom 2.x and see if this gets fixed. I'm not sure how much of our testing infrastructure code will break when we do this.

clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -46,15 +46,17 @@ girder.views.HierarchyWidget = girder.View.extend({
     initialize: function (settings) {
         this.parentModel = settings.parentModel;
         this.upload = settings.upload;
+        this.navigate = settings.navigate || false;
```
zachmullen: Could you add a docstring for this new parameter?

zachmullen:
After further reading, this appears to actually be jQuery behavior. It fires the high-level click event, but does not do the browser default fallback behavior of following the href attribute. In light of that, I think we should implement option 1.

zachmullen:
That should fix the issue of links not getting followed during testing. I discovered a bug, which is that clicking on a breadcrumb when in `navigate=true` mode breaks navigation.

ronichoudhury:
I'd like to get this finalized if possible. @zachmullen, were you planning to take another look at the remaining issues?

Maybe we can try to fix this up during tomorrow hackfest.

zachmullen:
I fixed the testing problem, but as I mentioned:

> I discovered a bug, which is that clicking on a breadcrumb when in navigate=true mode breaks navigation.

We can definitely try and sort that out tomorrow.

ronichoudhury:
Sounds good. Looking forward to it!

roni

ronichoudhury:
@zachmullen I've fixed the breadcrumb navigation but I'm having trouble running the client tests locally.

ronichoudhury:
I fixed an outstanding issue that was making a test fail (where the "empty folder message" was not properly appearing - though the fix is dubious, mainly because the code surrounding it is a bit dubious), but I'm not sure what to do about the test failures. @manthey helped me debug a bit but I probably need some help from above.

zachmullen:
I'm going to keep the branch, but close this PR since it's very out of date. If this becomes a priority again, we should reopen against 2.x.

PR #1435, opened 2016-06-29
Allow importing of individual files

zachmullen:


zachmullen:
@cpatrick PTAL, this will better facilitate synchronization of a local filesystem to a girder hierarchy.

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -332,19 +332,39 @@ def _importDataAsItem(self, name, user, folder, path, files,
             self.importFile(item, os.path.join(path, fname),
                             user, name=fname)
 
+    def _hasOnlyFiles(self, path, files):
+        return all(map(os.path.isfile, (os.path.join(path, fname)
```
cpatrick: nit: Since this is a comprehension in a map in an all, can we have a small `#` comment. :smile:

cpatrick:
Other than the one nit, LGTM.

girder/utility/filesystem_assetstore_adapter.py
fogleman: IMO, just get rid of the map:

    return all(os.path.isfile(os.path.join(path, fname)) for fname in files)

PR #1437, opened 2016-06-30
Include dicom in metadata extraction plugin

sutartmelson:


sutartmelson:
It appears that this only extracts metadata from dicom files, and not from any other files.

sutartmelson:
Perhaps there could be a plugin configuration page where you specify which headers to include in the metadata rather than dumping everything.

PR #1440, opened 2016-07-01
list/get/createUser methods in py client

fogleman:


fogleman:
WIP will write tests soon.

fogleman:
@zachmullen

clients/python/girder_client/__init__.py
```diff
@@ -391,6 +391,42 @@ def listItem(self, folderId, text=None, name=None, limit=None):
...
+    def listUser(self, limit=None):
+        """
+        Retrieves a list of users.
+
+        :param limit: the result set size limit.
+        """
+        params = {}
+        if limit:
+            params['limit'] = limit
+        return self.listResource('user', params)
```
zachmullen: I think we should improve all these listing endpoints; they take a `limit` but not offset, which isn't terribly useful. I think it would be better to make them act as generators that will internally handle the paging and yield one record at a time until the caller is satisfied, and they should also expose `params` in their API so that users can pass whatever else they want, including their own limit and offset if they want to do only a single specific slice.

zachmullen:
```
def listUser(self, params=None):
    params = params or {}
    params['offset'] = params.get('offset', 0)
    params['limit'] = params.get('limit', 50)

    while True:
        # fetch a page
        # increment offset
        # if empty set, break
        for record in page:
             yield record

fogleman:
@zachmullen I just added an offset argument to all the list methods for now.

zachmullen:
LGTM. We should be able to preserve this API even once we change it to the unlimited generator.

PR #1441, opened 2016-07-01
Add settings' option to modify default file permission in assetstore

Xarthisius:
This option is really useful if you want to touch assetstore directly from outside of Girder

### DIFF
--- a/clients/web/src/templates/widgets/editAssetstoreWidget.jade
+++ b/clients/web/src/templates/widgets/editAssetstoreWidget.jade
@@ -14,6 +14,9 @@
             .form-group
               label.control-label(for="g-edit-fs-root") Root directory
               input.input-sm#g-edit-fs-root.form-control(type="text", placeholder="Root directory")
+            .form-group
+              label.control-label(for="g-edit-fs-perms") File creation permissions (octal)
+              input.input-sm#g-edit-fs-perms.form-control(type="text", placeholder="600")
           if assetstore.get('type') === types.GRIDFS
             .form-group
               label.control-label(for="g-edit-gridfs-db") Database name
--- a/clients/web/src/views/widgets/EditAssetstoreWidget.js
+++ b/clients/web/src/views/widgets/EditAssetstoreWidget.js
@@ -79,11 +79,13 @@ girder.views.EditAssetstoreWidget = girder.View.extend({
     fieldsMap[girder.AssetstoreType.FILESYSTEM] = {
         get: function () {
             return {
-                root: this.$('#g-edit-fs-root').val()
+                root: this.$('#g-edit-fs-root').val(),
+                perms: this.$('#g-edit-fs-perms').val()
             };
         },
         set: function () {
             this.$('#g-edit-fs-root').val(this.model.get('root'));
+            this.$('#g-edit-fs-perms').val(this.model.get('perms'));
         }
     };
 
--- a/girder/utility/filesystem_assetstore_adapter.py
+++ b/girder/utility/filesystem_assetstore_adapter.py
@@ -70,6 +70,20 @@ class FilesystemAssetstoreAdapter(AbstractAssetstoreAdapter):
             raise ValidationException(
                 'Unable to write into directory "%s".' % doc['root'])
 
+        if doc.get('perms') is None:
+            doc['perms'] = oct(stat.S_IRUSR | stat.S_IWUSR)
+        else:
+            try:
+                perms = int(doc['perms'], 8)
+                # Make sure that mode is still rw for user
+                if not perms & stat.S_IRUSR or \
+                        not perms & stat.S_IWUSR:
+                    raise ValidationException(
+                        'Default permissions must allow "rw" for user')
+            except ValueError:
+                raise ValidationException(
+                    'Default permissions must be an octal integer')
+
     @staticmethod
     def fileIndexFields():
         """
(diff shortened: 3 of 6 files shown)


zachmullen:
This would probably be better as simply a field on the assetstore. That would also save us the extra database lookup.

girder/models/setting.py
```diff
@@ -209,6 +210,18 @@ def validateCoreUploadMinimumChunkSize(self, doc):
             'Upload minimum chunk size must be an integer >= 0.',
             'value')
 
+    def validateCoreFilesystemDefaultFileMode(self, doc):
+        try:
+            doc['value'] = int(doc['value'])
+            # Make sure that mode is still rw for user
+            if (stat.S_IRUSR | stat.S_IWUSR) & doc['value']:
```
zachmullen: This only tests that one or the other flag is set, not both. We want to do

```
rw = stat.S_IRUSR | stat.S_IWUSR
if doc['value'] & rw != rw:
    raise ValidationException(
        'File mode must grant the user "rw" access.', 'value')
```

zachmullen: Or it might be easier to read as

    if not doc['value'] & stat.S_IRUSR or not doc['value'] & stat.S_IWUSR:
        raise ValidationException(...)

girder/models/setting.py
```diff
@@ -209,6 +210,18 @@ def validateCoreUploadMinimumChunkSize(self, doc):
             'Upload minimum chunk size must be an integer >= 0.',
             'value')
 
+    def validateCoreFilesystemDefaultFileMode(self, doc):
+        try:
+            doc['value'] = int(doc['value'])
```
zachmullen: Shouldn't this be `int(doc['value'], 8)`, since most users would probably expect this to be in octal notation?

clients/web/src/templates/widgets/editAssetstoreWidget.jade
```diff
@@ -14,6 +14,9 @@
             .form-group
               label.control-label(for="g-edit-fs-root") Root directory
               input.input-sm#g-edit-fs-root.form-control(type="text", placeholder="Root directory")
+            .form-group
+              label.control-label(for="g-edit-fs-perms") Default file permissions
+              input.input-sm#g-edit-fs-perms.form-control(type="text", placeholder="Default permissions")
```
zachmullen: Let's make the placeholder be the actual default value (600), and make the label say "File creation permissions (octal)"

clients/web/src/views/widgets/EditAssetstoreWidget.js
```diff
@@ -79,11 +79,13 @@ girder.views.EditAssetstoreWidget = girder.View.extend({
...
         get: function () {
             return {
-                root: this.$('#g-edit-fs-root').val()
+                root: this.$('#g-edit-fs-root').val(),
+                perms: this.$('#g-edit-fs-perms').val()
             };
         },
         set: function () {
             this.$('#g-edit-fs-root').val(this.model.get('root'));
+            this.$('#g-edit-fs-perms').val("0o" + this.model.get('perms').toString(8));
```
zachmullen: Let's remove the "0o", that's a somewhat python-specific notation.

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -70,6 +70,20 @@ def validateInfo(doc):
             raise ValidationException(
                 'Unable to write into directory "%s".' % doc['root'])
 
+        if doc['perms'] is None:
+            doc['perms'] = (stat.S_IRUSR | stat.S_IWUSR)
```
zachmullen: No need for the parens

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -70,6 +70,20 @@ def validateInfo(doc):
...
+            try:
+                doc['perms'] = int(doc['perms'], 8)
+                # Make sure that mode is still rw for user
+                if not doc['perms'] & stat.S_IRUSR or \
+                        not doc['perms'] & stat.S_IWUSR:
+                    raise ValidationException(
+                        'Default permissions must allow "rw" for user')
+            except ValueError:
+                raise ValidationException(
+                    'Default permissions must be an integer')
```
zachmullen: 'Default permissions must be an octal integer'

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -70,6 +70,20 @@ def validateInfo(doc):
             raise ValidationException(
                 'Unable to write into directory "%s".' % doc['root'])
 
+        if doc['perms'] is None:
```
zachmullen: For supporting legacy assetstores, we can't assume the existence of this key, so we should change this to:

    if doc.get('perms') is None:

Xarthisius:
I had to keep `perms` as an octal string. Going back and forth with GET and PUT sometimes led to loosing context.

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -70,6 +70,20 @@ def validateInfo(doc):
             raise ValidationException(
                 'Unable to write into directory "%s".' % doc['root'])
 
+        if doc.get('perms') is None:
```
zachmullen: This should just be `if not doc.get('perms'):` so that if the user leaves that field blank they won't get errors.

zachmullen:
@Xarthisius I'm going to push this upstream and make some small tweaks on top of it, so I'll close this out and will ping you on the new PR.

PR #1442, opened 2016-07-05
Change python client list methods into generators

zachmullen:
ping @fogleman this implements the change we discussed, whereby users of the listing methods will no longer have to manage paging of results unless they want to, and changes the methods to be generators that will yield results until they're exhausted or the caller decides to stop iterating.

This technically breaks compatibility of those methods by changing their return type from list to generator; callers attempting to use `len` or `__getitem__` on the return value will be unable to do so after this change, so we need to decide how to version girder-client to reflect the breakage.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -343,11 +343,7 @@ class GirderClient(object):
         a single page. Passing an ``offset`` will work in both single-page and
         exhaustive modes.
         """
-        if params is None:
-            params = {}
-        else:
-            params = params.copy()
-
+        params = dict(params or {})
         params['offset'] = offset or 0
         params['limit'] = limit or DEFAULT_PAGE_LIMIT
 
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -353,7 +353,7 @@ class GirderClient(object):
                 yield record
 
             n = len(records)
-            if limit or not n or n < params['limit']:
+            if limit or n < params['limit']:
                 # Either a single slice was requested, or this is the last page
                 break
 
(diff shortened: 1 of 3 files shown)


clients/python/girder_client/__init__.py
```diff
@@ -329,28 +334,45 @@ def resourceLookup(self, path, test=False):
...
         """
-        search for a list of resources based on params.
+        This is a generator that will yield records using the given path and
+        params until exhausted. Paging of the records is done internally, but
+        can be overriden by manually passing a ``limit`` value to select only
+        a single page. Passing an ``offset`` will work in both single-page and
+        exhaustive modes.
         """
-        return self.get(path, params)
+        if params is None:
```
fogleman: Could just say `params = dict(params or {})`

clients/python/girder_client/__init__.py
```diff
@@ -329,28 +335,41 @@ def resourceLookup(self, path, test=False):
...
+        params['offset'] = offset or 0
+        params['limit'] = limit or DEFAULT_PAGE_LIMIT
+
+        while True:
+            records = self.get(path, params)
+            for record in records:
+                yield record
+
+            n = len(records)
+            if limit or not n or n < params['limit']:
```
fogleman: I think `not n` is superfluous here?

fogleman:
LGTMITIH

cjh1:
Ok, so this just broke some of my code, not a big deal, however, I am not sure I would have expected a breaking change with a minor version bump?

zachmullen:
I was following the policy we outlined here:  http://girder.readthedocs.io/en/latest/development.html#releasing-the-python-client-package

That decision was made a long time ago, so maybe it's time to rethink it, but at the time it seemed there was agreement that maintaining matching major version numbers between client and server was more important than semantic versioning of the client by itself.

PR #1445, opened 2016-07-07
Fix support for multiple allowed origins

zachmullen:
The access-control-allow-origin header will only cause a CORS
request to succeed if it contains a single domain value, which
must either be '*' or the value of the 'Origin' request header.


zachmullen:
@Xarthisius this should fix the problem you encountered when using a comma-separated allowed origins setting.

Xarthisius:
Works as advertised, thanks!

zachmullen:
@girder/developers PTAL

PR #1446, opened 2016-07-08
Add PUT /resource/:id/timestamp to set timestamps

fogleman:
For Midas to Girder migration, it is desirable to maintain the 'created' timestamp for the entities.

This PR adds a single endpoint, /resource/:id/timestamp, accessible only to admins for this purpose, instead of updating all of the endpoints for collections, users, folders, items, etc.

It also updates the girder python client with a method for this endpoint.

Tests WIP. Feedback welcome.


zachmullen:
LGTM

Issue #912, opened 2015-08-07
girder_client - Fail to upload data when assestore is a S3 bucket


jcfr:
@mgrauer and I could reproduce the issue on challenge.kitware.com and also on an instance using girder master.

The client has been installed using `pip install girder_client`

```
$ girder-cli --host challenge.kitware.com --scheme https --port 443  --parent-type folder 5519921acad3a55533264fd5 TCGA-17-Z011
[...]
Creating Folder from TCGA-17-Z011
Uploading Item from 000005 Unknown-label.nrrd
Uploading Item from 000005 Unknown.nrrd
Traceback (most recent call last):
  File "/home/jcfr/.virtualenvs/girder/bin/girder-cli", line 9, in <module>
    load_entry_point('girder-client==1.0.0', 'console_scripts', 'girder-cli')()
  File "/home/jcfr/.virtualenvs/girder/local/lib/python2.7/site-packages/girder_client/cli.py", line 96, in main
    reuse_existing=args.reuse)
  File "/home/jcfr/.virtualenvs/girder/local/lib/python2.7/site-packages/girder_client/__init__.py", line 895, in upload
    leaf_folders_as_items, reuse_existing)
  File "/home/jcfr/.virtualenvs/girder/local/lib/python2.7/site-packages/girder_client/__init__.py", line 852, in _upload_folder_recursive
    entry, folder['_id'], full_entry, reuse_existing)
  File "/home/jcfr/.virtualenvs/girder/local/lib/python2.7/site-packages/girder_client/__init__.py", line 758, in _upload_as_item
    local_file, current_item['_id'], file_path)
  File "/home/jcfr/.virtualenvs/girder/local/lib/python2.7/site-packages/girder_client/__init__.py", line 741, in _upload_file_to_item
    self.uploadFileToItem(parent_item_id, file_path)
  File "/home/jcfr/.virtualenvs/girder/local/lib/python2.7/site-packages/girder_client/__init__.py", line 470, in uploadFileToItem
    obj = self.post(path, parameters=parameters, files=filedata)
  File "/home/jcfr/.virtualenvs/girder/local/lib/python2.7/site-packages/girder_client/__init__.py", line 220, in post
    return self.sendRestRequest('POST', path, parameters, files=files)
  File "/home/jcfr/.virtualenvs/girder/local/lib/python2.7/site-packages/girder_client/__init__.py", line 214, in sendRestRequest
    text=result.text)
[...]

PR #1448, opened 2016-07-11
Fix multipart uploads for s3

fogleman:
@zachmullen 

I added the `response_headers` clause when I noticed it differed from the boto source: https://github.com/boto/boto/blob/dd5f87c7d5c55b69aa23bb4c21010897eb0deb42/boto/s3/connection.py#L354-L378

But it's possible you removed it intentionally? I didn't check the history. It probably has nothing to do with the error I was getting.

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -89,6 +89,9 @@ def _generate_url_sigv4(self, expires_in, method, bucket='', key='',
     if version_id is not None:
         params['VersionId'] = version_id
 
+    if response_headers is not None:
+        params.update(response_headers)
+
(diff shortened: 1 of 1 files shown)


fogleman:
I was able to reproduced the failure when including all headers, and it works with only content-type header.

zachmullen:
Excellent. It would be nice to figure out exactly which header was causing the problem, but I think it might be better to just plan to migrate to boto3 in the not-too-distant future.

LGTM

zachmullen:
For posterity, this was the issue @jcfr ran into almost a year ago when uploading via the client to an S3 assetstore.

jcfr:
xref #912

Nice. Good to see this being fixed :tada:

PR #1449, opened 2016-07-11
Add a browser widget modal dialog

jbeezley:
This adds a modal dialog for selecting a path graphically.  The dialog returns a model id that can be a collection, user, folder, or item.  For a usage example, I added a button to the `FilesystemImportView` so that you can browse to the import path rather than enter an id manually.

It looks something like this:
screen shot 2016-07-11 at 12 22 44 pm

### DIFF
--- a/clients/web/src/templates/widgets/browserWidget.jade
+++ b/clients/web/src/templates/widgets/browserWidget.jade
@@ -20,4 +20,4 @@
       a.btn.btn-small.btn-default(data-dismiss='modal') Cancel
       button.g-submit-button.btn.btn-small.btn-primary(type='submit')
         i.icon-edit
-          | Save
+        | Save
--- a/clients/web/src/templates/widgets/browserWidget.jade
+++ b/clients/web/src/templates/widgets/browserWidget.jade
@@ -10,9 +10,10 @@
       .g-hierarchy-widget-container.g-wait-for-root.form-group.hidden
       .help-block.g-wait-for-root.hidden
         | #{help}
-      .g-selected-model.g-wait-for-root.form-group.hidden
-        label(for='g-selected-model').control-label Selected id
-        input#g-selected-model.form-control(type='text' readonly)
+      if preview
+        .g-selected-model.g-wait-for-root.form-group.hidden
+          label(for='g-selected-model').control-label Selected id
+          input#g-selected-model.form-control(type='text' readonly)
       .g-validation-failed-message.hidden
 
     .modal-footer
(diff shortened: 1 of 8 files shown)


zachmullen:
Woohoo, this is an awesome feature. Will QA this soon.

manthey:
Excellent.  It would be nice if the Browser widget had a convenient way to override the text for the Save button (for instance, in the filesystem import dialog it might be more appropriate to be 'Select Destination').

clients/web/src/templates/widgets/browserWidget.jade
```diff
@@ -0,0 +1,23 @@
+.modal-dialog
+  .modal-content
+    .modal-header
+      button.close(data-dismiss='modal', aria-hidden='true', type='button') &times;
+      h4.modal-title
+        | #{title}
```
zachmullen: This is equivalent to

    = title

zachmullen:
This is really handy. How well does it handle pagination of collections and users? It might be good to add a SearchFieldWidget on collections and users to make it easier to pick the top level entity to use, though we could do that on another topic.

clients/web/src/views/widgets/BrowserWidget.js
```diff
@@ -0,0 +1,112 @@
...
+    _renderRootSelection: function () {
+        this._rootSelectionView.setElement(this.$('.g-hierarchy-root-container')).render();
+        this._renderHierarchyView();
+    },
+
+    _renderHierarchyView: function () {
+        if (this._hierarchyView) {
+            this.stopListening(this._hierarchyView);
+            this._hierarchyView.off();
+            this.$('.h-hierarchy-widget').empty();
```
jbeezley: Yes, fixed now.

jbeezley:
It doesn't directly handle pagination.  For the moment, the caller can handle it manually by calling methods on the collections exposed on the view's object (`.collection` and `.user`).  I could add special "Fetch more..." options in the select box to handle it automatically if desired.

Another idea I had after doing this is that I could make an option in the hierarchy widget to substitute the clickable link in the breadcrumb bar at the root level with the root selection view.  I had some concerns about the extra width that would add though.

zachmullen:
> Another idea I had after doing this is that I could make an option in the hierarchy widget to substitute the clickable link in the breadcrumb bar at the root level with the root selection view. I had some concerns about the extra width that would add though.

I like that idea a lot. If we make the dropdown link something other than a native `select` element, that could work.

zachmullen:
In any case, LGTM. I'll open issues for our feature requests.

jbeezley:
Sounds good.

PR #1450, opened 2016-07-12
Collapse multiple spaces in s3 filenames

fogleman:
@zachmullen

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -195,7 +195,7 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
         # this is due to a bug in S3 that does not properly handle filenames
         # with multiple spaces in a row, resulting in a SignatureDoesNotMatch
         # error
-        upload['name'] = ' '.join(filter(None, upload['name'].split(' ')))
+        upload['name'] = re.sub('\s+', ' ', upload['name'])
 
         uid = uuid.uuid4().hex
         key = '/'.join(filter(None, (self.assetstore.get('prefix', ''),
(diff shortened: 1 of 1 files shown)


girder/utility/s3_assetstore_adapter.py
```diff
@@ -191,6 +191,12 @@ def initUpload(self, upload):
         if upload['size'] <= 0:
             return upload
 
+        # collapse consecutive spaces in the filename into a single space
+        # this is due to a bug in S3 that does not properly handle filenames
+        # with multiple spaces in a row, resulting in a SignatureDoesNotMatch
+        # error
+        upload['name'] = ' '.join(filter(None, upload['name'].split(' ')))
```
zachmullen: I just tested, and S3 actually supports names with both leading and trailing spaces, so I would propose a slightly less restrictive name modification:

    upload['name'] = re.sub('\s+', ' ', upload['name'])

In case some crazy person really wants files with spaces at the beginning or end.

fogleman: Should we use `\s` or a more specific `\ `?

fogleman: (There's a space there, which somewhat ironically seems to not be showing.)

zachmullen: If they manage to get tabs or newlines in the filename, I definitely think we should replace those with spaces also, so I'm fine with \s.

fogleman: No backslash needed actually.

```
>>> import re
>>> def f(x):
...  return re.sub(' +', ' ', x)
... 
>>> f(' a b  c  ')
' a b c '
>>> 
```

zachmullen:
LGTM, thanks for tracking this one down

Issue #14, opened 2013-08-06
Migration script to move Midas 3.X data to Girder


cpatrick:
Need to make sure we can move data from a Midas 3.X instance into girder.

cpatrick:
This is fixed!

PR #1451, opened 2016-07-12
Add scripts for midas migration

fogleman:
@zachmullen @cpatrick @mgrauer

### DIFF
--- a/scripts/midas/requirements.txt
+++ b/scripts/midas/requirements.txt
@@ -1,0 +1,3 @@
+girder-client
+joblib
+pydas
--- a/scripts/midas/README.md
+++ b/scripts/midas/README.md
@@ -1,0 +1,41 @@
+# Midas to Girder Migration
+
+This folder contains scripts for migrating data from Midas to Girder.
+
+It can migrate users, collections (communities in Midas), folders, items, item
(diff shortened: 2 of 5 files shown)


zachmullen:
Fixes #14

mgrauer:
LGTM

zachmullen:
Hang on, can we add a README mentioning the prerequisites to install, and also mentioning that several values near the top of the script need to be replaced for it to work?

cjh1:
Is it worth adding a requirements.txt to make it easy for a user to install the deps?

scripts/midas/requirements.txt
```diff
@@ -0,0 +1,3 @@
+girder-client
+joblib
+pydas
```
zachmullen: What about sqlite?

cjh1: Its in the standard lib I think

fogleman: It is.

zachmullen: :mind_blown:

fogleman: sqlite is awesome.

zachmullen:
Cool, LGTM

PR #1452, opened 2016-07-12
Collapse build matrix to improve CI response time

zachmullen:
Given the limits on concurrent build runs, plus the overhead associated with each one, it makes sense to combine the `packaging` and `python` labels. This means we now have only one label that needs to be run in both python 2 and 3, making our total build count 4 instead of 6, which will increase our CI response time significantly in high traffic situations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ python:
 env:
     - GIRDER_TEST_GROUP=python
     - GIRDER_TEST_GROUP=browser
-    - GIRDER_TEST_GROUP=packaging
     - GIRDER_TEST_GROUP=other
 
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ before_install:
     - npm prune
     - pip install -U pip virtualenv
 install:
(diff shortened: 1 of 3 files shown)


jbeezley:
This looks okay to me.  You might also get rid of the now point condition [here](https://github.com/girder/girder/pull/1452/files#diff-354f30a63fb0907d4ad57269548329e3R48).  I was originally hoping limit the setup per test group, but it turned out less useful that I had hoped.

zachmullen:
It was a good experiment, even though in the end the time spent waiting in the queue and the time spent on setup/teardown ended up being too much.

zachmullen:
OK, I've also done away with the "other" row, those tests run really really quickly and are independent of python, so we now have a really simple matrix with just two rows, one for python tests (which should run against both py2 and py3) and one for everything else, which can just run once instead of twice. The "everything else" label is still called "browser" for lack of a better idea.

zachmullen:
@girder/developers any objections?

fogleman:
🐳  I guess not.

PR #1453, opened 2016-07-12
Allow to upload/download files using resource path

Xarthisius:
Utilizing resource path for target during upload/download is a bit more human-friendly:

```
girder-cli -c upload --api-url ... --api-key ... /user/xarthisius/Public ~/ala.txt
```

versus

```
girder-cli -c upload --api-url ... --api-key ... 570bd8fc2f2b14000176822d ~/ala.txt
```

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -922,11 +922,11 @@ class GirderClient(object):
         """
         Download a folder recursively from Girder into a local directory.
 
-        :param folderId: Id of the Girder folder to download.
+        :param folderId: Id of the Girder folder or resource path to download.
         :param dest: The local download destination.
         """
         offset = 0
-
+        folderId = self._checkResourcePath(folderId)
         while True:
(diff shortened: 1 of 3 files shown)


clients/python/girder_client/__init__.py
```diff
@@ -1243,3 +1244,10 @@ def upload(self, file_pattern, parent_id, parent_type='folder',
                     leaf_folders_as_items, reuse_existing)
         if empty:
             print('No matching files: ' + file_pattern)
+
+    def _checkResourcePath(self, objId):
+        if objId.startswith('/'):
+            obj = self.resourceLookup(objId, test=True)
+            if obj is not None:
+                return obj["_id"]
```
zachmullen: Let's use single quotes to be consistent.

Xarthisius: I'm sorry, it looks like I'll never learn

zachmullen:
This is a great feature, thanks! LGTM

PR #1454, opened 2016-07-12
Change python line length limit to 100

zachmullen:
@girder/developers rejoice!

### DIFF
--- a/girder/__init__.py
+++ b/girder/__init__.py
@@ -41,26 +41,23 @@ class LogLevelFilter(object):
 
 class LogFormatter(logging.Formatter):
     """
-    Custom formatter that adds useful information about the request to the logs
-    when an exception happens.
+    Custom formatter that adds useful information about the request to the logs when an exception
+    happens.
     """
     def formatException(self, exc):
         info = '\n'.join((
-            '  Request URL: %s %s' % (cherrypy.request.method.upper(),
-                                      cherrypy.url()),
+            '  Request URL: %s %s' % (cherrypy.request.method.upper(), cherrypy.url()),
             '  Query string: ' + cherrypy.request.query_string,
             '  Remote IP: ' + cherrypy.request.remote.ip
         ))
-        return ('%s\n'
-                'Additional info:\n'
-                '%s' % (logging.Formatter.formatException(self, exc), info))
+        return ('%s\nAdditional info:\n%s' % (logging.Formatter.formatException(self, exc), info))
 
 
 def getLogPaths():
     """
-    Return the paths to the error and info log files. These are returned as
-    a dict with "error" and "info" keys that point to the respective file,
-    as well as a "root" key pointing to the log root directory.
+    Return the paths to the error and info log files. These are returned as a dict with "error" and
+    "info" keys that point to the respective file, as well as a "root" key pointing to the log root
+    directory.
     """
     cfg = config.getConfig()
     logCfg = cfg.get('logging', {})
--- a/girder/__main__.py
+++ b/girder/__main__.py
@@ -27,22 +27,18 @@ except ImportError:
     # Update python path to ensure server respawning works. See #732
     source_root_dir = os.path.dirname(os.path.dirname(__file__))
     import sys
-    cherrypy.engine.log("[Girder] Appending source root dir to 'sys.path': %s"
-                        % source_root_dir)
+    cherrypy.engine.log("[Girder] Appending source root dir to 'sys.path': %s" % source_root_dir)
     sys.path.append(source_root_dir)
     from girder.utility import server
 
 
 def main():
-    parser = argparse.ArgumentParser(
-        description='Girder: data management platform for the web.')
-    parser.add_argument("-t", "--testing", help="run in testing mode",
-                        action="store_true")
-    parser.add_argument("-d", "--database",
-                        help="to what database url should Girder connect")
-    parser.add_argument("-p", "--port",
-                        help="on what port should Girder serve")
+    parser = argparse.ArgumentParser(description='Girder: data management platform for the web.')
+    parser.add_argument("-t", "--testing", help="run in testing mode", action="store_true")
+    parser.add_argument("-d", "--database", help="to what database url should Girder connect")
+    parser.add_argument("-p", "--port", help="on what port should Girder serve")
     args = parser.parse_args()
+
     if args.database:
         cherrypy.config['database']['uri'] = args.database
     if args.port:
--- a/girder/api/access.py
+++ b/girder/api/access.py
@@ -93,8 +93,7 @@ def token(*args, **kwargs):
         @six.wraps(args[0])
         def wrapped(*iargs, **ikwargs):
             if not rest.getCurrentToken():
-                raise AccessException(
-                    'You must be logged in or have a valid auth token.')
+                raise AccessException('You must be logged in or have a valid auth token.')
             return args[0](*iargs, **ikwargs)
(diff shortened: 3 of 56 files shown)


manthey:
I'm sort of against this.  Why 100 and not a larger number?  I know, PEP 8 sort of sanctifies 72, 79, 80, 99, or 100 columns, but it feels like a whimsical change.  I think we should either leave the length limit alone or abolish it entirely.

zachmullen:
I think having a limit is a good idea, but I also think 80 characters is too constraining once you add in several tabs; things can get jumbled up and start to look ugly due to all the newlines and indentation. I'd like people to not have to scroll their editors sideways, but I also want the code to look pretty, and this 100 number was agreeable to @cpatrick and several others who happened to be present 😄

manthey:
I still say getting rid of the limit and objecting to code with absurd lengths in code review would be better.  Will we change to 120 in a few years once we've softened up those who just now thought 100 was reasonable?

zachmullen:
> Will we change to 120 in a few years once we've softened up those who just now thought 100 was reasonable?

Maybe. But moving it up doesn't break anything, and moving it down does, so I'd rather go incrementally than make it too long now.

fogleman:
My only gripe is with this style of indentation:

```python
foo = long_function_name(var_one, var_two,
                         var_three, var_four)
```

Everything gets bunched up on the right side of the screen, at some arbitrary location. And you have to insert some also arbitrary number of spaces to make it line up.

It's "OK" according to PEP-8, but so is an alternative method that's better IMO:

```python
foo = long_function_name(
    var_one, var_two, var_three, var_four)
```

fogleman:
As far as 80 vs. 100, I find that Python makes it really easy to stay under the 80 limit. Until you're dealing with verbose APIs, or APIs that go crazy with nested packages, etc. So a lot of it has to do with coding style.

fogleman:
Perhaps 80 should be considered a soft limit and 100 a hard limit?

Aim for 80 but if it reads better to go a little over, that's OK?

zachmullen:
I've only run into a handful of situations where it's been difficult to stay under 80, although I'm guilty of shortening variable names on several occasions to avoid newlines, which probably detracts from readability. But in general I'd prefer to be less constrained unless there's a compelling reason for constraint.

opadron:
-1 for me, too.  I've ran into a few cases where staying under 80 got tricky, and I've found that just lifting subexpressions into another variable on the line above was sufficient for just about all these cases.  I also think it helps with readability since most of the time, the expression is wieldy at those line lengths.

And for those crazy APIs/edge cases where you just can't make 80 characters work, or to do so would make the code so much worse to read?  Just break the rule.  I'm pretty sure PEP8 allows for this for such extenuating circumstances.  These edge cases shouldn't be coming up often enough to be a problem.

fogleman:
I do prefer 80 as well, with the occasional line going over a little if needed. In particular, docstrings and comments and such should remain wrapped at 80 IMO.

zachmullen:
> I do prefer 80 as well, with the occasional line going over a little if needed. In particular, docstrings and comments and such should remain wrapped at 80 IMO.

Why?

cpatrick:
I share @fogleman's preference. Ideally we have an 80 char soft limit and and a 100 char hard limit. I take blame for this as I was a hard-line 80-char guy for a very long time and caved in the aforementioned meeting.

zachmullen:
What exactly is a "soft limit" when it comes to setting a policy?

sebastienbarre:
And I recorded the meeting. I will make a GIF if I need to :)

cpatrick:
Please don't gif me :smile_cat:.

As far as a soft limit, it's mentioned in our code review docs but not required in our style tests.

sebastienbarre:
A "soft" limit would generate a linting warning, whereas a "hard" limit 
would generate a linting error?
Personally warnings or errors, I would rather not see or being flagged 
by my editor (but it's probably toggable).

I'm +1 on the 100 characters line length (I switched KWiK to 100 earlier 
this year, I find it a lot more clear, but then again, PHP).

zachmullen:
To avoid bikeshedding on this subject, let's just put this to a vote. Please mark your preference with a thumbs up on one of the following comments. I'd like to get all @girder/developers @girder/core and anybody who reads or writes girder python code or plugins to vote. Thank you for your input!

zachmullen:
Keep the 80 character limit

zachmullen:
Change to 100 character limit

zachmullen:
(I have intentionally left out more complex schemes of "hard" and "soft" limits because I think they're too confusing as a policy.)

fogleman:
I'm reminded of this talk that I saw at PyCon: https://www.youtube.com/watch?v=wf-BqAjZb8M&feature=youtu.be&t=40

ronichoudhury:
I think refactoring your code or shortening variable names, etc., just to meet an antiquated line length limit rule is a sort of code smell. You wanted the longer line for a reason, so if we trust that you are a good coder, then changing the line to meet what is essentially a bureaucratic regulation, by definition, makes the code less optimal (readable, grokkable, etc.) than it was.

To add a bit of evidence for this position from the other end: a line that is too long can easily become unreadable as well, so at a programmer's discretion she might decide to refactor, etc., to *reduce* the length. But I've never liked having to do it by fiat.

fogleman:
I don't think it'd be very confusing to have more flexible limits:

- Comments and docstrings should be wrapped to 80 characters.
- Code should also aim to wrap around 80 characters, but if it hurts readability, going a little over is OK - subject to the developer's and reviewers' discretion. A hard limit of 100 is enforced by the linter.

Why 80? Because everyone uses different editors / IDEs / configurations and it's the closest thing we have to a global standard. Why 80 for comments? Because it's not hard to wrap plain text to 80 and longer lines of text can be harder to read.

zachmullen:
I have selfish reasons for wanting a simple policy here. I spend a lot of time doing QA on pull requests to girder and external plugins. I want our linter to be the thing that handles our style rules to the greatest extent possible; I don't want to have to post that bullet point block every time an external contributor makes a PR, I'd rather focus my QA time on substance and let the linter take care of style. So picking between one of these two simple policies is a compromise I hope we can all make.

sebastienbarre:
> Why 80? Because everyone uses different editors / IDEs / configurations and it's the
> closest thing we have to a global standard. 

But times have changed since [the IBM punch card](http://programmers.stackexchange.com/a/148678), screen resolutions on laptops are a lot higher, I think that's why we can now reasonably push changes such has 100 chars line limit.

opadron:
I suppose I should also point out that, although against the change overall, it's not that big of a deal to me in the grand scheme of things.  If the change ultimately does more good than harm, then it seems logical to introduce it.

manthey:
@zachmullen is there a reason you're editing lots of files as part of this change?  It seems like it should just be the change to the flake8.cfg and that is all.  Every other change should, at best, do nothing useful, but each change risks introducing bugs.  It also mucks up the blame history, so that all of the changes most recently come from this PR instead of whatever caused someone to make the actual code change.

zachmullen:
Just showing how things look by changing some places in core to fit the new length.

This change is part of our 2.0 changes, but I didn't open it against 2.0-integration since it's not actually breaking anything. My plan is to integrate all outstanding python changes prior to merging this. Compared to some of our other planned 2.0 changes, this is actually pretty low-impact; we'll need to do a full manual test cycle prior to release anyway.

zachmullen:
Also, I'm happy to be the blame target for all these lines ;)

PR #1455, opened 2016-07-13
Support file caching in girder python client

fogleman:
This PR lets the girder python client cache files. The `downloadFile` method will use a disk-backed cache if configured. Downloaded files will be saved in the cache and later calls to `downloadFile` can simply fetch the file from the cache instead of downloading it again from Girder.

The [diskcache](https://pypi.python.org/pypi/diskcache/) library is used as the cache. It should work well even for very large files as it can store file-like objects by reading them in chunks.

Here is an example configuration:

```python
CACHE_SETTINGS = {
    'directory': '/tmp/cache',
    'eviction_policy': 'least-recently-used',
    'size_limit': 10 * 2 ** 30, # 10 GB
}

gc = girder_client.GirderClient(apiUrl=GIRDER_URL, cacheSettings=CACHE_SETTINGS)
# ...
gc.downloadFile(fileId, path)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ before_install:
     - npm prune
     - pip install -U pip virtualenv
 install:
-    - pip install -r requirements-dev.txt -e . -e .[plugins] -e clients/python
+    - pip install -r requirements-dev.txt -e .[plugins] -e clients/python
     - npm install
 script:
     - mkdir _build
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ before_install:
     - npm prune
     - pip install -U pip virtualenv
 install:
-    - pip install -r requirements.txt -r requirements-dev.txt -e .
+    - pip install -r clients/python/requirements.txt -r requirements-dev.txt -e .
     - npm install
 script:
     - mkdir _build
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,8 @@ before_install:
     - npm prune
     - pip install -U pip virtualenv
 install:
-    - pip install -r clients/python/requirements.txt -r requirements-dev.txt -e .
+    - pip install -r requirements.txt -r requirements-dev.txt -e .
+    - pip install -r clients/python/requirements.txt -e clients/python
     - npm install
 script:
     - mkdir _build
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,8 +42,7 @@ before_install:
     - npm prune
     - pip install -U pip virtualenv
 install:
-    - pip install -r requirements.txt -r requirements-dev.txt -e .
-    - pip install -r clients/python/requirements.txt -e clients/python
+    - pip install -r requirements-dev.txt -e . -e .[plugins] -e clients/python
     - npm install
 script:
     - mkdir _build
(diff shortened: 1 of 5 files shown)


zachmullen:
Cool! TODO:

- Add diskcache to the python client's requirements.txt and setup.py
- Add a test that uses httmock to verify that we don't download a file that is already cached, and that we do re-download a file if its mtime changes.

manthey:
This is more of a heads-up than a request for a change.  I'm working on a database assetstore, where each file is a view of a table.  The file model download function returns that view as JSON or CSV data.  Using the extraParameters field that was added recently, you can change the view and get different data.  Eventually, the girder python client should support extraParameters as well, and the caching should make sure that those are part of the cache key.  Addiitonally, there should be a mechanism to ask to have a non-cached version of the file.

I think the later of the created and updated timestamps should be used for the cachekey, not the created timestamp by itself.

zachmullen:
> Addiitonally, there should be a mechanism to ask to have a non-cached version of the file.

Definitely. This is still the default behavior (you have to opt-in to caching). Once we plug this into girder_worker, girder_io input specs will be able to override caching parameters at runtime if they desire.

> I think the later of the created and updated timestamps should be used for the cachekey, not the created timestamp by itself.

These blobs were originally immutable, then we added the notion of replacing the blob. Doing so actually updates its `created` field rather than using an `updated` field, for whatever reason.

Once we have to think about exposing `extraParameters` in the client, we can decide what to do there. I would suggest that, at least at first, the use of `extraParameters` should preclude using the cache altogether since at that point we are breaking the traditional opaque blob model and all notion of immutability may have gone out the window.

.travis.yml
```diff
@@ -42,7 +42,7 @@ before_install:
     - npm prune
     - pip install -U pip virtualenv
 install:
-    - pip install -r requirements.txt -r requirements-dev.txt -e .
+    - pip install -r clients/python/requirements.txt -r requirements-dev.txt -e .
```
zachmullen: Maybe this should be:

    - pip install -r requirements-dev.txt -e . -e clients/python

So we can test out our setup.py

fogleman: I dunno, it doesn't seem to happy that I removed `-r requirements.txt`

zachmullen: `pip install -e .[plugins]` will also include all the plugin requirements

.travis.yml
```diff
@@ -42,8 +42,7 @@ before_install:
     - npm prune
     - pip install -U pip virtualenv
 install:
-    - pip install -r requirements.txt -r requirements-dev.txt -e .
-    - pip install -r clients/python/requirements.txt -e clients/python
+    - pip install -r requirements-dev.txt -e . -e .[plugins] -e clients/python
```
zachmullen: You can actually remove the other `-e .`

zachmullen:
LGTM

cjh1:
Just out of interest what was the use case driving this change. Was it for MIDAS migration?

zachmullen:
The main driving use case is the `girder_io` plugin of girder_worker. COVALIC is a good example of where this will be useful: for grand challenges, the ground truth data is downloaded each and every time a new submission is made. We also use the girder_io module for HistomicsTK when running image analysis on very large microscopy images, so this could save time there too.

There will be a subsequent PR in girder_worker soon that integrates this change.

cjh1:
Ok, thanks, It currently doesn't support cache invalidation if the file changes?

fogleman:
File changes update the `created` timestamp on the file, which is part of the cache key.

cjh1:
Ok, thanks missed that.

zachmullen:
LGTM

PR #1460, opened 2016-07-14
Bump girder python client to 1.3.1

fogleman:
@zachmullen I will package & upload to pypi once merged.

v1.3.1 adds support for diskcache.

PR #1461, opened 2016-07-14
PEP 396 compliance

zachmullen:
@girder/developers PTAL. [PEP 396](https://www.python.org/dev/peps/pep-0396/) provides a standard way of including a version identifier inside packages.

This shouldn't be merged until after #1460 is merged.

### DIFF
--- a/clients/python/setup.py
+++ b/clients/python/setup.py
@@ -17,7 +17,6 @@
 #  limitations under the License.
 ###############################################################################
 
-import girder_client
 from setuptools import setup, find_packages
 
 install_reqs = [
@@ -28,10 +27,16 @@ install_reqs = [
 with open('README.rst') as f:
     readme = f.read()
 
+init = os.path.join(os.path.dirname(__file__), 'girder_client', '__init__.py')
+with open(init) as fd:
+    version = re.search(
+        r'^__version__\s*=\s*[\'"]([^\'"]*)[\'"]',
+        fd.read(), re.MULTILINE).group(1)
+
 # perform the install
 setup(
     name='girder-client',
-    version=girder_client.__version__,
+    version=version,
     description='Python client for interacting with Girder servers',
     long_description=readme,
     author='Kitware, Inc.',
--- a/clients/python/setup.py
+++ b/clients/python/setup.py
@@ -17,11 +17,8 @@
 #  limitations under the License.
 ###############################################################################
 
-
+import girder_client
 from setuptools import setup, find_packages
-
-
-CLIENT_VERSION = '1.3.0'
 
 install_reqs = [
     'diskcache',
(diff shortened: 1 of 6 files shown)


jbeezley:
👍

zachmullen:
Now I see why this is problematic, it precludes installing directly from setup.py in a clean environment. Looking around, I notice [requests is just hacking around this issue](https://github.com/kennethreitz/requests/blob/master/setup.py#L50-L52). Are we comfortable doing the same thing in our setup.py files?

clients/python/setup.py
```diff
@@ -31,10 +27,16 @@
 with open('README.rst') as f:
     readme = f.read()
 
+init = os.path.join(os.path.dirname(__file__), 'girder_client', '__init__.py')
+with open(init) as fd:
+    version = re.search(
+        r'^__version__\s*=\s*[\'"]([^\'"]*)[\'"]',
+        fd.read(), re.MULTILINE).group(1)
+
```
ronichoudhury: This feels weird. Can we not simply import `girder_client` and extract `__version__` from it directly?

jbeezley: No, because `girder_client` is not installed yet when this is run.

zachmullen: The problem is that when you `import girder_client`, it [transitively imports a bunch of stuff](https://github.com/girder/girder/blob/master/clients/python/girder_client/__init__.py#L20-L31). This works fine if you are installing via an egg or have already installed the requirements, but if you checked out the girder source and tried to do a `pip install -e .` (which normally works), it would no longer work because none of the required third party libraries would have been installed yet.

zachmullen:
This is ready for review.

zachmullen:
Does anyone object to this being merged?

jbeezley:
I'll give it an LGTM.

zachmullen:
Thanks!

cjh1:
To late, but wouldn't it be nice to have the version string in a separate .py file. That can be imported in girder/__init__.py and loaded using exec(...) in setup.py, so we aren't scanning a file using a regex?

zachmullen:
That's a good idea, feel free to make that change if you want.

jbeezley:
I understand being uncomfortable with the regex, but trying to use the import mechanism inside a package being installed is pretty dangerous.  I would avoid it and trust that the requests people know what they're doing.

zachmullen:
The regex is obviously not elegant, but it's also low risk; the code isn't part of the girder package, and it's well tested by our CI, so we'll know if it breaks.

cjh1:
@jbeezley the exec(...) approach could be used in both places, it would at least get rid of the regex. I will leave well alone, just an observation.

PR #1463, opened 2016-07-14
Fixes #1462. Do not accept invalid JSON in metadata

### LINKED ISSUE #1462: Invalid item metadata renders folder inaccessible
To replicate:

1. Create a folder A
2. Create an item B inside folder A
3. Use `/item/{id}/metadata` to attach `{"value":Infinity}` to the item.

    This will respond with a cherrypy-style, HTML internal server error (not the regular kind):

   `ValueError: Out of range float values are not JSON compliant: inf`

4. From this point, further attempts to access Folder A in the girder interface will also yield this error.

zachmullen:
ping @alex-r-bigelow

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -242,14 +242,26 @@ def requireAdmin(user, message=None):
         raise AccessException(message or 'Administrator access required.')
 
 
-def getBodyJson():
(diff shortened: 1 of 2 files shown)


alex-r-bigelow:
I can confirm that #1462 is fixed. Thanks!

PR #1464, opened 2016-07-14
Add support for config file

Xarthisius:
For all those lazy people like me tired of typing so many options each time.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -29,6 +29,24 @@ import requests
 import shutil
 import six
 import tempfile
+from six.moves.configparser import SafeConfigParser
+
+_config_defaults = {
+    'host': 'localhost',
+    'scheme':  'http',
+    'apiRoot': '/api/v1',
+    'apiUrl': None,
+    'port': None,
+    'apiKey': None,
+    'username': None,
+    'password': None,
+}
+CONFIG_DIR = os.environ.get(
+    'XDG_CONFIG_HOME', os.path.join(os.path.expanduser('~'), '.config'))
+config = SafeConfigParser(_config_defaults, allow_no_value=True)
+config.read([os.path.join(CONFIG_DIR, "girder-cli.conf")])
+if not config.has_section("girder_client"):
+    config.add_section("girder_client")
 
 __version__ = '2.0.0'
 __license__ = 'Apache 2.0'
@@ -147,16 +165,23 @@ class GirderClient(object):
         :param cacheSettings: Settings to use with the diskcache library, or
             None to disable caching.
         """
-        if apiUrl is None:
+        configApiUrl = config.get('girder_client', 'apiUrl')
+        if apiUrl is None and configApiUrl is None:
             if apiRoot is None:
-                apiRoot = '/api/v1'
-
-            self.scheme = scheme or 'http'
-            self.host = host or 'localhost'
-            self.port = port or (443 if scheme == 'https' else 80)
+                apiRoot = config.get('girder_client', 'apiRoot')
+            self.scheme = scheme or config.get('girder_client', 'scheme')
+            self.host = host or config.get('girder_client', 'host')
+            try:
+                configPort = config.getint('girder_client', 'port')
+            except TypeError:
+                configPort = None
+            self.port = port or configPort or \
+                (443 if self.scheme == 'https' else 80)
 
             self.urlBase = '%s://%s:%s%s' % (
                 self.scheme, self.host, str(self.port), apiRoot)
+        elif configApiUrl is not None:
+            self.urlBase = configApiUrl
         else:
             self.urlBase = apiUrl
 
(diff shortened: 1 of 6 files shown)


clients/python/girder_client/cli.py
```diff
@@ -42,11 +42,12 @@ def __init__(self, username, password, dryrun, blacklist,
         GirderClient.__init__(self, host=host, port=port,
                               apiRoot=apiRoot, scheme=scheme, dryrun=dryrun,
                               blacklist=blacklist, apiUrl=apiUrl)
-        interactive = password is None
+        interactive = password is None and \
+            config.get("gitder_config", "password", fallback=None)
```
danlamanna: Typo?

clients/python/girder_client/__init__.py
```diff
@@ -29,6 +29,17 @@
...
 import tempfile
+from six.moves.configparser import SafeConfigParser
+
+_config_defaults = {
+    'host': 'localhost',
+    'scheme':  'http',
+    'apiRoot': '/api/v1'
+}
+CONFIG_DIR = os.environ.get(
+    'XDG_CONFIG_HOME', os.path.join(os.path.expanduser('~'), '.config'))
```
danlamanna: +1

We could also put `girder-cli.conf` in a top level `girder` directory under `CONFIG_DIR`, though I can't think of a use case for this at the moment.

clients/python/girder_client/__init__.py
```diff
@@ -140,16 +151,20 @@ def __init__(self, host=None, port=None, apiRoot=None, scheme=None,
         :param cacheSettings: Settings to use with the diskcache library, or
             None to disable caching.
         """
-        if apiUrl is None:
+        configApiUrl = config.get('girder_client', 'apiUrl', fallback=None)
```
zachmullen: We should use the config file as a fallback if the options aren't passed to the constructor rather than having the config file take precedence.

clients/python/girder_client/cli.py
```diff
@@ -39,16 +39,25 @@ def __init__(self, username, password, dryrun, blacklist,
         :param password: password to authenticate to Girder instance, leave
             this blank to be prompted.
         """
+        if not all((apiUrl, apiRoot, host, port, scheme)):
```
zachmullen: Shouldn't this be `if not any`?

Xarthisius: oops

zachmullen:
Most command-line tools that can use config files also have an option for pointing to a config file at a specific location. Could you add such an option to the `girder-cli`?

zachmullen:
Sorry it's been forever since I looked at this.

One thing that makes me a little hesitant here is that this will fail on environments where there is no home directory. That may sound crazy, but I just encountered such an environment while trying to work with Google App Engine recently, and girder client is something we may want to deploy there at some point. In fact, on GAE any attempt to do filesystem IO or lookups will throw exceptions.

So the solution for me would be to pull out all the references to the config file into as little code as possible in `__init__.py`, and wrap the call(s) up in a try/except block that just keeps to the existing logic if an exception occurs.

Xarthisius:
@zachmullen what would cause the issue in that case? `os.path.expanduser`? `config.read` is already fail-proofed for non-existing, non-accessible files:

```
>>> from six.moves.configparser import ConfigParser
>>> config = ConfigParser()
>>> config.read(['/asdasd/asdas/d'])
[]
```

zachmullen:
`os.path.expanduser` will fail.

Xarthisius:
I'll wrap it in `try: except:` but it looks like it's a bug in GAE, see https://github.com/google/oauth2client/issues/578 They claim it should work on GAE already, but doesnt' work in SDK yet

clients/python/girder_client/__init__.py
```diff
@@ -29,6 +29,24 @@
...
+    'apiKey': None,
+    'username': None,
+    'password': None,
+}
+CONFIG_DIR = os.environ.get(
+    'XDG_CONFIG_HOME', os.path.join(os.path.expanduser('~'), '.config'))
+config = ConfigParser(_config_defaults, allow_no_value=True)
+config.read([os.path.join(CONFIG_DIR, "girder-cli.conf")])
+if not config.has_section("girder_client"):
+    config.add_section("girder_client")
```
zachmullen: All of this code should move inside `config.py` and should get wrapped in a function. We should not call that code at the time `girder_client` is imported; rather, it should be up to users to conditionally read the config file, and then pass that into the `GirderClient` constructor.

zachmullen: We can have the config read and passed in by default in `cli.py`, but then we should add a CLI switch like `--no-config-file` to turn off the behavior of reading the config file.

Xarthisius:
@zachmullen I think I've addressed all your comments, PTAL.

clients/python/girder_client/configure.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def write_config(self, fd=None):
+        _safeMakedirs(self.config_dir)
+        if fd is None:
+            with open(self.config_file, 'w') as fd:
+                self.config.write(fd)
+        else:
+            self.config.write(fd)
+
+    def rm_config(self, section, option):
```
zachmullen: Let's change the symbols in this module to be camelCase in keeping with the convention of the rest of the package.

zachmullen:
Looks good other than the minor nit. Could you merge latest master into this?

zachmullen:
I've reopened this as #1850, thanks @Xarthisius !

PR #1469, opened 2016-07-19
Use HTML doctype for building Jade templates

brianhelba:
This primarily affects self closing tags and boolean attributes.

### DIFF
--- a/grunt_tasks/build.js
+++ b/grunt_tasks/build.js
@@ -46,6 +46,7 @@ module.exports = function (grunt) {
             options: {
                 client: true,
                 compileDebug: false,
+                doctype: 'html',
                 namespace: 'girder.templates',
                 processName: function (filename) {
                     return path.basename(filename, '.jade');


zachmullen:
Ping @sebastienbarre this change might not be valid after the build process changes, so let's make sure to integrate it on your PR as well.

sebastienbarre:
Noted, thanks. Yes, this code is gone in build.js. Will advise.

zachmullen:
LGTM

brianhelba:
@sebastienbarre Just so I understand, how does the webpack-based build compile Jade templates? This option should still be placed in there somewhere.

sebastienbarre:
@brianhelba using webpack [loaders](https://webpack.github.io/docs/using-loaders.html), namely the [jade-loader](https://github.com/pugjs/pug-loader). I'll figure something out.

sebastienbarre:
`doctype=html` is the default for the jade loader, [as I found out](https://github.com/pugjs/pug-loader/commit/d9caf449dad6458ebc8d970cad330394cfdac424). To be on the safe side, I will still pass it on the loader's query string.

PR #1470, opened 2016-07-19
[WIP] Add an option for new users to have private visibility

brianhelba:


brianhelba:
This supports a use case for @satra . PTAL.

brianhelba:
@girder/developers To fully support this, since `User` is an `AccessControlledModel` type, I think we should also expose a `/user/:id/access` endpoint, and add a `girder.views.AccessWidget` to the `girder.views.UserAccountView`. Does this sound reasonable?

zachmullen:
There are certain related scenarios I'd want to test, like what happens if I'm a group admin and a user belongs to that group but I don't have permission to view the user?

clients/web/src/templates/body/systemConfiguration.jade
```diff
@@ -77,6 +77,21 @@ form.g-settings-form(role="form")
...
+          selected=(
+            settings['core.user_default_public'] === null ?
+            defaults['core.user_default_public'] === true :
+            settings['core.user_default_public'] === true)
+          ) Public
+        option(value="false",
+          selected=(
+            settings['core.user_default_public'] === null ?
+            defaults['core.user_default_public'] === false :
+            settings['core.user_default_public'] === false)
```
zachmullen: Maybe we should capture this structure in a function so we don't have to repeat it

brianhelba:
@zachmullen I can tell you the answer from my own experience using private users: the private user actually shows up in the group list to everyone (even anonymous users). But you're right, this should be tested further.

satra:
thanks @brianhelba 

@zachmullen - i think this speaks to the granularity of access and visibility more generally. brian and i had a bit of a discussion over file level access permissions as well. i think there is a place for a broader discussion of acl's amongst groups, users, collections, items, and files. just noting it here for now, but happy to move to a new issue for discussion.

zachmullen:
@brianhelba what's the status of this PR, are you comfortable that everything regarding private users works properly?

jcfr:
@brianhelba This was closed during today's hackfest because it didn't see activity for a while, feel free to re-open or comment. Thanks

PR #1474, opened 2016-07-20
Refactor python client's commands into subparsers. Fixes #1471

### LINKED ISSUE #1471: girder_client should use subparsers for commands
I'd like to add a new command to `girder_client`. However it takes a different number of arguments than `upload/download/localsync`. Providing command via `-c` switch is a bit cumbersome this way. 

I'd like to propose switching commands to separate subparsers:
```
girder-cli upload ...
girder-cli download ...
```

Since it breaks backward compatibility, I'd like to request comments from @girder/developers before implementing.

Xarthisius:
This splits girder-cli's subcommands into subparsers

### DIFF
--- a/clients/python/girder_client/cli.py
+++ b/clients/python/girder_client/cli.py
@@ -18,6 +18,8 @@
 ###############################################################################
 
 import argparse
+import sys
+from six import add_metaclass, string_types
 from girder_client import GirderClient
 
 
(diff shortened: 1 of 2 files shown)


clients/python/girder_client/cli.py
```diff
@@ -50,67 +52,141 @@ def __init__(self, username, password, dryrun, blacklist,
...
         if args.parent_type != 'folder':
             print('localsync command only accepts parent-type of folder')
-        else:
-            g.loadLocalMetadata(args.local_folder)
-            g.downloadFolderRecursive(args.parent_id, args.local_folder,
-                                      sync=True)
-            g.saveLocalMetadata(args.local_folder)
-    else:
-        print('No implementation for command %s' % args.c)
+            sys.exit(0)
```
zachmullen: Why do we need `sys.exit` here?

clients/python/girder_client/cli.py
```diff
@@ -50,67 +52,141 @@ def __init__(self, username, password, dryrun, blacklist,
...
+parser.add_argument('--scheme', required=False, default=None)
+parser.add_argument('--host', required=False, default=None)
+parser.add_argument('--port', required=False, default=None)
+parser.add_argument('--api-root', required=False, default=None,
+                    help='relative path to the Girder REST API')
+parser.add_argument('--blacklist', default='', required=False,
+                    help='comma separated list of filenames to ignore'),
+parser.add_argument('--dryrun', action='store_true',
+                    help='will not write anything to Girder, only report'
+                    ' on what would happen'),
```
zachmullen: I'd rather `--blacklist` and `--dryrun` be applied on the subparsers since there's no guarantee that they'll be sensible options on all future subcommands.

zachmullen: To be honest I'd be OK with keeping `--dryrun` here since we could come up with a reasonable general implementation for that, but `--blacklist` is only used for uploading. I realize these are both options to the GirderClient class constructor, but I think we should remove `blacklist` as a constructor argument.

zachmullen:
@Xarthisius thanks for this, I'm going to go ahead and merge it and apply the dryrun and blacklist changes on top of it.

PR #1476, opened 2016-07-21
Webpack

sebastienbarre:
Here we go.

## Scope

This PR holds all changes related to the [ES6/ES2015 modularization](http://www.2ality.com/2014/09/es6-modules-final.html) of the Girder web client and 20+ of its plugins, as well as the subsequent bundling using [Webpack](https://webpack.github.io/).

Almost all source files have been updated to no longer rely on the global `girder` object/singleton. Specific functionalities, Jade templates, as well as Stylus/CSS styles must now be imported *explicitly*. For example:

```javascript
import _ from 'underscore';

import FileModel from 'girder/models/FileModel';
import View from 'girder/views/View';
import events from 'girder/events';
import { renderMarkdown, formatSize } from 'girder/misc';

import MarkdownWidgetTemplate from 'girder/templates/widgets/markdownWidget.jade';
import 'girder/stylesheets/widgets/markdownWidget.styl';

var MarkdownWidget = View.extend({
    events: {
        'show.bs.tab .g-preview-link': function () {
            renderMarkdown(this.val().trim() || 'Nothing to show', this.$('.g-markdown-preview'));
        },
...
});

export default MarkdownWidget;
```

Importing on a "per-needed" basis will provide the added benefit of facilitating [tree shaking](http://www.2ality.com/2015/12/webpack-tree-shaking.html) and build-time checks.

## Directory Structure

The directory structure in `clients\web\src` was not altered, but some functions/classes were moved to their own files. For example:

- authentication-related code is now in `auth.js`,
- constants are now in `constants.js`,
- dialog (UI) related code is in `dialog.js`,
- the event emitter is now in `events.js`,
- miscellaneous functions are now in `misc.js`,
- REST-related code is now in `rest.js`,
- the router is now in `router.js`,
- routes definitions are *all* in `routes.js`,
- server related code is in `server.js`,
- `collection.js` was moved to `collections\Collection.js`,
- `model.js` was split into `Model.js`, `AccessControlledModel.js` and `MetadataMixin.js` inside the `models` subdir,
- `plugin_utils.js` is now in `utilities\PluginUtils.js`,
- jQuery plugins are now in the `utilities\jquery` subdir,

## Plugins

All plugins were cleaned-up to adopt the same directory structure and naming conventions as the core files, and to better convey which core files are being extended or wrapped. Specifically:

- all plugins feature a `main.js` entry point,
- routes, if any, are defined in `routes.js`,
- views are to be found in a `views` subdir,
- all templates, stylesheets, and views were renamed, when needed, to be consistent with each-others (e.g. `views/ItemPreviewWidget.js`, `templates/itemPreviewWidget.jade`, `stylesheets/itemPreviewWidget.styl`),
- any core model/collection/view being extended prototypically or `wrap()`'ed is now to be found in a single file matching the same model/collection/view name and location (say, `plugins/curation/web_client/views/HierarchyWidget.js`).

## Migration

I would recommend checking any collection, model, view, or plugin to help transitioning your own code. Search for instances of the `girder.` string to find the functions previously attached to the `girder` global namespace, and bring them into your code using `import... from...`. Export your own code using `export default ...` to make it available to the outside world as a module. Start a new plugin by copy/pasting an existing one.
   
Note that it is still possible, although not recommended, to use a higher-level namespace. 
For example, the old:
```javascript
girder.fetchCurrentUser(blah blah);
```
is now coded this way:
```javascript
import { fetchCurrentUser } from 'girder/auth';
[...]
fetchCurrentUser(blah blah);
```
yet, it could also be written this way:
```javascript
import * as auth from 'girder/auth';
[...]
auth.fetchCurrentUser(blah blah);
```
...but your bundler of choice (say, Webpack) would no longer be able to check at build-time if `fetchCurrentUser` is to be found in `girder/auth`. It could also be rewritten as:
```javascript
import * as girder from 'girder';
[...]
girder.auth.fetchCurrentUser(blah blah);
```
... but the risk here would be even higher, should the `auth` namespace or its contents be moved or renamed.

Be aware that if you passed the `girder` global to a Jade template, you will now need to pass individual functions separately, and update the Jade template accordingly. For example:

in `clients/web/src/views/body/CollectionsView.js`:
```javascript
       this.$el.html(girder.templates.collectionList({
             collections: this.collection.toArray(),             
             girder: girder
         }));
```
becomes:
```javascript
import { formatSize } from 'girder/misc';
import { getCurrentUser } from 'girder/auth';
import CollectionListTemplate from 'girder/templates/body/collectionList.jade'; 
[...]
       this.$el.html(CollectionListTemplate({
             collections: this.collection.toArray(),
             getCurrentUser: getCurrentUser,
             formatSize: formatSize
         }));
``` 
and in the corresponding `clients/web/src/templates/body/collectionList.jade`
```jade
if girder.currentUser
[...]
        span.g-user-space-used  #{girder.formatSize(collection.get('size'))}
```
becomes:
```
if getCurrentUser()
[...]
        span.g-user-space-used  #{formatSize(collection.get('size'))}
```

Also note that a `girder` global is still available to facilitate testing, but only *after* the core Javascript bundles and all plugins have been loaded. It is nearly identical to the previously available global, but some functions were shuffled around, as described in the "Directory Structure" section.

Tip: use `console.dir(girder)` to browse the new tree.
```
$ console.dir(girder)
> views
> auth
> collections
> utilities
> server
> dialog
  > confirm
     [...]
[...]
```
then follow the same "path" to migrate to the new `import...from` statement:
```javascript
import { confirm } from 'girder/dialog';
```
Importing from one plugin to another can be done using the `plugins/` prefix. For example,
consider `JobStatus` being exported by the `jobs` plugin in `/plugins/jobs/web_client/JobStatus.js`:
```javascript
var JobStatus = {
    registerStatus: function (status) {
        [...]
    }
}
export default JobStatus;
```
then later on used by the `worker` plugin in `/plugins/worker/web_client/JobStatus.js`, like this (note that there is no `web_client` subdir here):
```javascript
import JobStatus from 'plugins/jobs/JobStatus';
JobStatus.registerStatus({
    WORKER_FETCHING_INPUT: {
        [...]
    }
});
```

## TODO

- [x] This work was done on Mac OSX, which is case-insensitive -- there may be some lingering, trivial file naming issues that Unix users will notice right away.

### Grunt Build-process

The old, Grunt build process is still in place. I do not advocate removing it actually, as Webpack is a bundler, not a task runner. It was nonetheless slimmed down, as `grunt_tasks/build.js` is almost 100% Webpack code at this point. A few Grunt plugins remain, and could potentially be removed in favor of Webpack (see such plugins as [kevlened/copy-webpack-plugin](https://github.com/kevlened/copy-webpack-plugin) as a replacement to `grunt-contrib-copy`). 

- [ ] **Testing**: `grunt_tasks/dev.js`still uses `grunt-contrib-uglify`; it needs to be exercised, converted to the new modules, and likely fixed before it is Webpack'ed; UPDATE: soon to be webpack'ed,
- [ ] **Packaging**: `grunt_tasks/package.js` uses `grunt-shell`; **it needs to be exercised and likely fixed**,
- [x] **Fontello**: `grunt_tasks/fontello.js` uses `grunt-contrib-copy` and `grunt-fontello`; it is functional and would likely be fine as is,
- [x] **Plugins**: `grunt_tasks/plugin.js` still uses `grunt-contrib-copy` and `grunt-npm-install`; the code that handles Javascript, Jade, and Stylus for each plugin was moved and integrated into the single Webpack target in `grunt_tasks/build.js`; while plugins appear to be generated correctly, ***a large section (starting at `grunt.file.expand()`) needs to be exercised and likely be fixed or removed**,
- [x] **Sphinx**: `grunt_tasks/sphinx.js` uses `grunt-shell`; it appears functional and would likely be fine left as is,
- [x] **Swagger**: `grunt_tasks/swagger.js` uses `grunt-contrib-copy` and `grunt-contrib-stylus`; it is functional but could be Webpack'ed, or left as is,
- [x] **Versioning**: `grunt_tasks/version.js` uses `grunt-file-creator` and `grunt-gitinfo`; it is functional and would likely be fine as is,

### Webpack

The Webpack target generates, in `clients\web\static\built`:
- `girder.ext.min.js` and `girder.ext.min.css` (and corresponding source maps): all third-party/vendor Javascript and CSS (e.g. imported from `node_modules`), 
- `girder.app.min.js` and `girder.app.min.css` (and corresponding source maps): all Girder core Javascript/Jade/CSS (e.g. found in `clients\web\src`) used by the client *and* the plugins,
- `plugin.min.js` and `plugin.min.css` (and corresponding source maps) for each plugin, in `plugins\name_of_plugin`: plugin code only, typically small, making calls into `girder.app`  (and `girder.ext`),
- glyphicons in the `assets` subdir

However:
- [ ] There still appears to be too much duplicated code going in `girder.main.min.js` and each `plugin.min.js`; this needs to be investigated and fixed by myself (Seb; consider [robertknight/webpack-bundle-size-analyzer](https://github.com/robertknight/webpack-bundle-size-analyzer)), also consider the `DllPlugin`, which would increase performance as well ([Webpack Plugins we been keepin on the DLL](https://medium.com/@soederpop/webpack-plugins-been-we-been-keepin-on-the-dll-cdfdd6cb8cd7), [Optimizing Webpack for Faster React Builds](http://engineering.invisionapp.com/post/optimizing-webpack/), [Optimizing Webpack build times and improving caching with DLL bundles](https://robertknight.github.io/posts/webpack-dll-plugins/), [webpack/examples/dll](https://github.com/webpack/webpack/tree/master/examples/dll)). Also check [trivago/parallel-webpack: Builds multi-config webpack projects in parallel](https://github.com/trivago/parallel-webpack)
- [x] The Grunt `watch` target in `build.js` needs to be converted to a similar capability compatible with Webpack; (use `grunt --watch` or `grunt webpack:watch`, for `grunt-contrib-watch` is too slow),
- [x] Support for ES2015 code through Babel should be enabled (by myself, Seb),
- [x] Importing `$.girderModal()` looks too much like a side effect: `import 'girder/utilities/jQuery';`; let's find a way to make it more explicit (check how jQuery does it nowadays), or restructure the files so that we have one prototype method per file, so that it would be more obvious why it's getting imported, e.g. `import 'girder/utilities/jQuery/girderModal';`
- [x] It is still unclear at this point how minimization and debugging will ultimately be enabled in Webpack 2.0 (see [What's new in webpack 2](https://gist.github.com/sokra/27b24881210b56bbaff7#loader-options--minimize), [Why does the UglifyJsPlugin affect other loaders?](https://github.com/webpack/webpack/issues/283), [this comment](https://github.com/webpack/webpack/issues/2061#issuecomment-228932941)), so the 2 (or 3) different ways to do so are used at the same time (see how `environment` and `debugJs` are used in `grunt_tasks/build.js`), but this will need some clean up later on,

### Testing

- [x] Programmatic testing of the web client was disabled, and **needs to be converted and re-enabled ASAP**, or Continuous Integration will fail (see `grunt_tasks/dev.js`, remove early `return` in (`grunt.registerTask`); I (Seb) could use some help for this one, from the original testing author **Update: most tests are passing, but the way the test driver exists make it look like they are failing on some laptops**,

### Plugins

All plugins were converted. Manual testing was performed to check that each plugin's UI is in place, using eye-ball technology.

- [x] Programmatic testing, if any, **needs to be converted and re-enabled ASAP** (say, `plugins\autojoin\plugin_tests\autoJoin.spec`)

### Misc, Minor

- [ ] Consider moving constants from `constants.js` back to more relevant files, closer to the functions that most often use them?
- [x] Consider files like `auth.js`, `events.js`, `rest.js`, `router.js`, which essentially bundle utility functions, vs. the files in `utilities` such as `DialogHelper.js`, `MiscFunctions.js`, or `PluginUtils.js`, which *also* bundle utility functions. Are the former considered more important "utilities" then the latter? Or should they all be in the `utilities` dir? Should some be merged somehow in the process? Or named more consistently? 
- [x] `girder/router` exports `{ router, splitRoute }`. Therefore, accessing the router from the `girder` singleton will look like `girder.router.router`; not ideal; consider renaming `girder/router` to `girder/routing` to use `girder.routing.router` instead?
- [x] `girder/events` exports `{ events, eventSream }`. Therefore, accessing the events object from the `girder` singleton will look like `girder.events.events`; not ideal; consider renaming `girder/events` to `girder/eventing` to use `girder.eventing.events` instead?


sebastienbarre:
And we are back.

cjh1:
Looks like swagger doc needs looking at as well ( the Web API link at the bottom of the page )

```
swagger-ui.min.js:3 Uncaught TypeError: Cannot read property '$ref' of undefined
```

sebastienbarre:
@cjh1 I can't reproduce your Swagger issue -- I'm able to look at it, browse it, and test it. Maybe you have some straggler files, try wiping out your `clients/web/static/built/swagger`?

cjh1:
So after a little digging with @sebastienbarre the swagger issue turned out to be a environment issue, the Girder server serving the swagger doc didn't match the webpack branch ... works fine after this was resolved.

sebastienbarre:
@zachmullen after doing some research ([here](http://stackoverflow.com/questions/35136874/extend-es6-plugin-to-jquery-prototype) and [here](http://stackoverflow.com/questions/35136874/extend-es6-plugin-to-jquery-prototype)), it appears your suggestion to move `$.girderModal` to `utilities/jquery/girderModal` is the sensible one. The change is in.

grunt_tasks/build.js
```diff
@@ -14,301 +14,98 @@
...
-                    test: /\.svg(\?v=\d+\.\d+\.\d+)?$/,
-                    loaders: [urlLoader({ mimetype: 'image/svg+xml' }), fileLoader()]
+        // The grunt-contrib-watch task below is one option for dev, as described here:
+        // https://github.com/webpack/webpack-with-common-libs/blob/master/Gruntfile.js
+        // but it could also be done using the webpack dev server.
+        // It is however A LOT SLOWER than using the built-in watch options in grunt-webpack
+        watch: {
+            app: {
+                files: getWatchFiles(paths.web_src),
+                tasks: ['warnWatch', isDev ? 'webpack:dev' : 'webpack:prod'],
```
zachmullen: We have a `npm run watch` command that will be our documented way to watch, so we can point that to the "proper" command to use and get rid of everything else.

sebastienbarre: It's good to hear. I did put that warning because there still is *one* remaining `grunt-contrib-watch` task at play here, the one used to watch the Sphinx doc (see `sphinx.hs`). Now the issue is that you can't use both at the same time, because both `grunt-contrib-watch` or `grunt-webpack` (in watch mode) would try take control (or `webpack-dev-server` for that matter). It was working before because `grunt-contrib-watch` would watch everything (doc files, javascript, jade, etc) but this is neither viable nor desirable now that Webpack is in control of the build process. My gut feeling is that "watching" the build process is a more common case than watching the documentation process, so I would probably favor `npm run watch` triggering the `webpack:watch` task. We can chew on that.

zachmullen:
@sebastienbarre FYI I'm about to take a pass at the testing related stuff. Stop me if you are already down that road.

sebastienbarre:
@zachmullen Yes I am doing that this afternoon. What about I make a pass at it first and come back to you with my quesstions?

zachmullen:
Great idea :) it's all yours

sebastienbarre:
@zachmullen this is progressing, I'll likely commit something tomorrow. For the sake of simplicity, I'm not going to webpackify the tests -- they will access the API through the `girder` singleton I'm exposing after all webpacked bundles have loaded. This will exercise said singleton at the same time -- I think it should still be around in 2.0 for people to test or play around from the console (my 2 cents). I'm not exactly sure how code coverage will work; I see `dev.js` was extracting grunt dependencies (aka source files) from the `girder.app.min.js` task, and loading all of them as separate scripts -- this can no longer be done, dependencies are gone and computed by Webpack itself, and sources can't be loaded as in because of the `from/import`. II'll cross that bridge later, I'm sure this has been solved by someone, somewhere. Maybe the source maps can help.

manthey:
@sebastienbarre Jon Beezley switched [geojs](https://github.com/OpenGeoscience/geojs) to use webpack and run client-side tests via Istanbul and phantomjs.  This has been working just fine, though the error messages are in a different format then before the change.

zachmullen:
Switching to Istanbul is certainly fine. We can adopt any new tools that make our lives easier on the testing side of things.

sebastienbarre:
Sounds good, I've used Istanbul with node.js/ES2015 in the past, but not w/ Phantomjs -- I will poke at geojs and go pester @jbeezley eventually :) Thank you.

sebastienbarre:
Here is new commit -- lots of fixes on the tests.
I've about 3 left to fix out of 30.

However, while the specs are passing, most of them still *look* like they are failing, and I suspect this is on the Python side of things -- the way the server is shut down, or the way the mock SMTP server is closed, etc. I'm looking at you, `/tests/web_client_test.py`. I'm not a Python guy (understatement of the year), help appreciated here.

For example:
```
$ ctest -V -I 62,62
[...]
62: ----------------
62: Testing Finished
62: ----------------
62: 29 specs, 0 failures in 4.634s.
62:
62: ConsoleReporter finished
62: FAIL
62: Exception in thread Thread-104:
62: Traceback (most recent call last):
62:   File "/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 801, in __bootstrap_inner
62:     self.run()
62:   File "/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 754, in run
62:     self.__target(*self.__args, **self.__kwargs)
62:   File "tests/mock_smtp.py", line 86, in loop
62:     asyncore.loop(timeout=0.5, use_poll=True)
62:   File "/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/asyncore.py", line 216, in loop
62:     poll_fun(timeout, map)
62:   File "/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/asyncore.py", line 145, in poll
62:     r, w, e = select.select(r, w, e, timeout)
62: error: (9, 'Bad file descriptor')
62: ======================================================================
62: FAIL: testWebClientSpec (tests.web_client_test.WebClientTestCase)
62: ----------------------------------------------------------------------
62: Traceback (most recent call last):
62:   File "tests/web_client_test.py", line 201, in testWebClientSpec
62:     self.assertEqual(returncode, 0)
62: AssertionError: 1 != 0
62:
62: ----------------------------------------------------------------------
62: Ran 1 test in 10.281s
62:
62: FAILED (failures=1)
1/1 Test #62: web_client_data_filesystem .......***Failed   11.26 sec
```

As you can see, all the specs *did* pass, then things went wrong at exit time. Something in `mock_smtp.py` is failing in *all* the tests. It is unclear if the latter is triggering the big "FAIL", since test #127 will pass entirely even with the `mock_smtp.py` error. But it wouldn't hurt to eliminate that variable. 

For the sake of clarity, it would probably be good for the below warnings to disappear too; they seem to be on the server side (swagger?):
```
62: WARNING: Error response for code '400' is already defined (old: 'Path is invalid.', new: 'Path refers to a resource that does not exist.')
62: WARNING: Error response for code '400' is already defined (old: 'Unsupport or unknown resource type.', new: 'Invalid resources format.')
62: WARNING: Error response for code '400' is already defined (old: 'Invalid resources format.', new: 'No resources specified.')
62: WARNING: Error response for code '400' is already defined (old: 'No resources specified.', new: 'Resource not found.')
62: WARNING: Error response for code '400' is already defined (old: 'Unsupport or unknown resource type.', new: 'Invalid resources format.')
62: WARNING: Error response for code '400' is already defined (old: 'Invalid resources format.', new: 'No resources specified.')
62: WARNING: Error response for code '400' is already defined (old: 'No resources specified.', new: 'Resource not found.')
```

Finally, one of the 3 tests, `web_client_swagger`, is eluding me, if somebody wants to look at it... It doesn't exercise anything related to the web client code itself, it goes to the Swagger API page, and spits some errors.

Thanks

jbeezley:
`web_client_data_filesystem` passed for me.  Here is the result from the full test suite: <http://my.cdash.org/viewTest.php?onlyfailed&buildid=1027313>

jbeezley:
I reran `web_client_data_gridfs`, and it passed.  I'm guessing we have randomly occurring error in the `web_client_data_*` tests.

sebastienbarre:
thanks @jbeezley ; that's OSX?

jbeezley:
Yes OSX.  Running the `server_upload` test again, it passed as well.

zachmullen:
Hi @sebastienbarre , could you merge the latest 2.0-integration into this topic? Thanks!

sebastienbarre:
@zachmullen done

girder/api/api_docs.mako
```diff
@@ -51,7 +51,7 @@
     <script src="${staticRoot}/built/swagger/lib/underscore-min.js"></script>
     <script src="${staticRoot}/built/swagger/lib/backbone-min.js"></script>
     % if mode == 'testing':
-    <script src="${staticRoot}/built/polyfill.min.js"></script>
+    <script src="${staticRoot}/built/testing/polyfill.min.js"></script>
```
sebastienbarre: @jourdain sorry for the waste of time :)

sebastienbarre:
@jeffbaumes I'm getting this from your PR:
```
Failed to load vega/Gruntfile.js:
>>> Error: Cannot find module 'girder/plugins/vega/Gruntfile.js'
```

jeffbaumes:
@sebastienbarre see https://github.com/girder/girder/pull/1536 for a fix for the error message (which is actually a warning message re-printing a caught error).

sebastienbarre:
Here we go, the web_client tests are all passing on my machine.
```
(env) barre@killy $ PHANTOMJS_RETRY=1 ctest -R web_client
Test project /Users/Shared/Src/girder/girder-build
      Start  62: web_client_data_filesystem
 1/27 Test  #62: web_client_data_filesystem .....................   Passed   11.32 sec
      Start  63: web_client_data_gridfs
 2/27 Test  #63: web_client_data_gridfs .........................   Passed   11.75 sec
      Start  64: web_client_data_gridfsrs
 3/27 Test  #64: web_client_data_gridfsrs .......................   Passed   34.93 sec
      Start  65: web_client_data_s3
 4/27 Test  #65: web_client_data_s3 .............................   Passed   12.47 sec
      Start  66: web_client_admin
 5/27 Test  #66: web_client_admin ...............................   Passed   27.59 sec
      Start  67: web_client_collection
 6/27 Test  #67: web_client_collection ..........................   Passed    8.77 sec
      Start  68: web_client_group
 7/27 Test  #68: web_client_group ...............................   Passed   17.32 sec
      Start  69: web_client_user
 8/27 Test  #69: web_client_user ................................   Passed   12.36 sec
      Start  70: web_client_item
 9/27 Test  #70: web_client_item ................................   Passed    8.76 sec
      Start  71: web_client_folder
10/27 Test  #71: web_client_folder ..............................   Passed    8.84 sec
      Start  72: web_client_routing
11/27 Test  #72: web_client_routing .............................   Passed   11.84 sec
      Start  73: web_client_version
12/27 Test  #73: web_client_version .............................   Passed    5.33 sec
      Start  74: web_client_custom_widgets
13/27 Test  #74: web_client_custom_widgets ......................   Passed    6.00 sec
      Start  75: web_client_widgets
14/27 Test  #75: web_client_widgets .............................   Passed   14.91 sec
      Start  76: web_client_datetime_widget
15/27 Test  #76: web_client_datetime_widget .....................   Passed    5.86 sec
      Start  77: web_client_empty_layout
16/27 Test  #77: web_client_empty_layout ........................   Passed    6.77 sec
      Start  78: web_client_swagger
17/27 Test  #78: web_client_swagger .............................   Passed    8.90 sec
      Start  79: web_client_browser
18/27 Test  #79: web_client_browser .............................   Passed    5.81 sec
      Start  85: web_client_autojoin.autojoin
19/27 Test  #85: web_client_autojoin.autojoin ...................   Passed    8.38 sec
      Start  91: web_client_curation.curation
20/27 Test  #91: web_client_curation.curation ...................   Passed    6.22 sec
      Start  95: web_client_geospatial.geospatial
21/27 Test  #95: web_client_geospatial.geospatial ...............   Passed    6.26 sec
      Start 104: web_client_hashsum_download.hashsum_download
22/27 Test #104: web_client_hashsum_download.hashsum_download ...   Passed    5.08 sec
      Start 120: web_client_item_licenses.item_licenses
23/27 Test #120: web_client_item_licenses.item_licenses .........   Passed    8.14 sec
      Start 127: web_client_jobs.jobs
24/27 Test #127: web_client_jobs.jobs ...........................   Passed    5.24 sec
      Start 142: web_client_provenance.provenance
25/27 Test #142: web_client_provenance.provenance ...............   Passed    6.66 sec
      Start 147: web_client_thumbnails.thumbnails
26/27 Test #147: web_client_thumbnails.thumbnails ...............   Passed    6.71 sec
      Start 152: web_client_user_quota.user_quota
27/27 Test #152: web_client_user_quota.user_quota ...............   Passed   13.36 sec

100% tests passed, 0 tests failed out of 27
```

Thanks to the girdster who helped me out.

@zachmullen I'd love for TravisCI to finally go green, can we push for that?

@jbeezley will you be taking ownership of the ESLint test? Per discussion, this might require upgrading to ESLint 2 or 3. It shouldn't be too bad, I've cleaned most of the client code when SublimeText would report a linting issue. Thanks

jbeezley:
Sure.  I'll give it a shot.

sebastienbarre:
Alright, I looked at the TravisCI logs.

Once I ignored the bazillions yellow "WARNING: Error response for code '400' is already defined (old: 'Path is invalid.', new: 'Path refers to a resource that does not exist.')", I noticed:

```
 2/99 Test  #29: python_static_analysis_core_tests .....................***Failed    6.64 sec
```
Could a Pythoner take a quick look at this?

Then the following:
```
164: Running "npm-install:vega@^2.6.0" (npm-install) task
...
164: > canvas@1.4.0 install /home/travis/build/girder/girder/_build/env/lib/python2.7/site-packages/girder/node_modules/canvas
...
164: make: *** [Release/obj.target/canvas/src/CanvasPattern.o] Error 1
164: make: Leaving directory `/home/travis/build/girder/girder/_build/env/lib/python2.7/site-packages/girder/node_modules/canvas/build'
164: gyp ERR! command "/home/travis/.nvm/versions/io.js/v2.3.3/bin/iojs" "/home/travis/build/girder/girder/_build/env/lib/python2.7/site-packages/gir
```
@jeffbaumes Seems vega is bringing Canvas, which fails to compile. However, from the above, it looks like we are requiring `iojs`. In `.travis.yml` I can confirm that `nvm install v2.3.3` is bringing `iojs`. Now this is fairly old, with respect to `node`'s history -- both `iojs` and `node` have re-merged. @zachmullen If you don't mind, I'll require `node` v5 to see if this fixes the build (I use v6).

Also:
```
	 25 - js_coverage_combine_report (Failed)
	 27 - eslint_grunt_modules (Failed)
	 68 - web_client_admin (Failed)
	161 - eslint_core (Failed)
```
- The ESLint tests, that was expected, and once @jbeezley makes the upgrade, I'm sure we can tackle all the linting issues quickly,
- the coverage I'll have a look, since it's... well... gone for now,
- `web_client_admin` passes on my machine, but let's see if it remains after requiring node 5.

Thanks

zachmullen:
@sebastienbarre I find it easier to look at the test failure outputs on CDash:

http://my.cdash.org/viewTest.php?onlyfailed&buildid=1028400
http://my.cdash.org/viewTest.php?onlyfailed&buildid=1028413

sebastienbarre:
OK I think we are getting closer. What still fails:
- ESLint
- js_coverage_combine_report (I'll look)
- the packaging tests (I think I may have broken that?)
- one web_client test

cjh1:
I am taking a look at python_static_analysis_core_tests for @sebastienbarre

sebastienbarre:
Alright guys, it seems we have a green dashboard, thanks. 
I'll work on the coverage this week.
One of the TravisCI build still [fails on vega](https://travis-ci.org/girder/girder/jobs/153821142), not sure why (@jeffbaumes?)

jeffbaumes:
@sebastienbarre yes Vega brings in canvas I believe for offline rendering, which we are not using. Looks like there is not yet a way to decouple it from Vega dependencies, see https://github.com/vega/vega/issues/274

jbeezley:
To fix the vega install on travis, try adding the apt packages that vega's own repository uses.

<https://github.com/vega/vega/blob/master/.travis.yml>

sebastienbarre:
OK let me try

sebastienbarre:
Afraid not.
```
../../nan/nan.h:43:3: error: #error This version of node/NAN/v8 requires a C++11 compiler
```

jbeezley:
Possible fix here: https://github.com/girder/girder/commit/dcbc91f979cd1d8d02be0ecab6b302817eeacfd0

This is a little bit unfortunate because it requires a bunch of system level dependencies for what is only an optional plugin.  The only way around it that I can think of is to get vega through bower rather than npm.

zachmullen:
There's one more major change to our build system that I plan to implement once the webpack stuff is in, which will be to disable building of plugins via the default task, and create a task for building each plugin. So, that should alleviate the problem of plugins that are expensive to build.

sebastienbarre:
Thanks @jbeezley ; feel free to push that commit, I would like to make Travis happy somehow.
@zachmullen I actually had already added `--skip-plugins` to help debugging this branch (see `build.js`)

jbeezley:
That should fix it.  Everything looks to be passing now.  There does seem to be a frequent spurious failure of `web_client_group` though.

zachmullen:
Just wanted to check in, I think our checklist may be stale at this point. What's still left to be done at this point? Anything I can help with?

sebastienbarre:
Thanks @zachmullen ; I just updated some of the code snippet in the PR ~~novel~~ description to reflect changes made during the hackathon, and I updated the checklist. 

For convenience, here are the ones remaining:

- [ ] `grunt_tasks/dev.js`still uses `grunt-contrib-uglify`; it could be Webpack'ed (by Seb),
- [ ] `grunt_tasks/package.js` seems to be Python (server?) related, not client; could a Python person clarifies what it is for, and if it needs fixing? I'm not entirely confident anything related to `isSourceBuild` has really been tested, 
- [ ] There still appears to be too much duplicated code put by Webpack in `girder.main.min.js` and each `plugin.min.js`; this needs to be investigated and fixed by myself (Seb) [...]
- [ ] Consider moving constants from `constants.js` back to more relevant files, closer to the functions that most often use them?

First and third I'll do. Second I could use some help. Last one is minor.

I do not think any of these would prevent us from merging this branch inside `2.0-integration` rather sooner than later, so that we can move on with other items on the 2.0 roadmap (in the Google doc). All of the above could be fine-tuned or fixed if they were in `2.0-integration`. Your call of course.

Thanks

jbeezley:
For the package task, I think it would be appropriate to remove it.  It is only there for publishing on pypi, which most people will never do.  We could possibly put the equivalent command as a one liner npm script.

jbeezley:
Are we punting on coverage for now, or is that still in the works?

sebastienbarre:
@jbeezley oh sorry, no punting, I had forgotten to put it there, good catch. I need to do a quick experiment with [SitePen/remap-istanbul: A tool for remapping Istanbul coverage via Source Maps](https://github.com/SitePen/remap-istanbul) (and [gotwarlost/babel-coverage](https://github.com/gotwarlost/babel-coverage) to some extent).

jeffbaumes:
@sebastienbarre https://github.com/Kitware/candela also performs source-mapped coverage through Babel. If what we do can help inform your process (or vice-versa) it would be good to have a uniform solution.

jbeezley:
Candela uses karma to handle those complexities.  Porting existing tests to use karma would be a significant undertaking which is why I suspect @sebastienbarre has chosen a more direct approach.

zachmullen:
Thanks for the write-up. My vote would be that we should get coverage working on this branch, then merge it and leave additional stuff to subsequent PR's.

jeffbaumes:
Ah ok, thanks for the explanation @jbeezley.

jbeezley:
There is also [isparta](https://github.com/douglasduteil/isparta) for babel coverage if that helps.

sebastienbarre:
Alright I'll write a quick remapper for the Blanket coverage files.
For example, if you consider `js_coverage/user.cfg` in the build tree (ellipsis mine):
```
Fclients/web/static/built/girder.app.min.js
...
L1 1
...
L23 167
```
Line 23 was hit 167 times in `girder.app.min.js`. Now open `girder.app.min.js` (ellipsis mine):
```javascript
...
var View = __WEBPACK_IMPORTED_MODULE_1_backbone___default.a.View.extend({
   constructor: function constructor(opts) {
       // eslint-disable-line backbone/no-constructor
       if (opts && __WEBPACK_IMPORTED_MODULE_0_underscore___default.a.has(opts, 'parentView')) {
...
```
Line 23 is:
```javascript
       if (opts && __WEBPACK_IMPORTED_MODULE_0_underscore___default.a.has(opts, 'parentView')) {
```

This piece of Javascript I just tested:
```javascript
const sourceMap = require('source-map');
const fs = require('fs');

const SourceMapConsumer = sourceMap.SourceMapConsumer;

const checkSourceMap = filename => {
  const rawSourceMap = JSON.parse(fs.readFileSync(filename));
  const smc = new SourceMapConsumer(rawSourceMap);
  console.log(smc.originalPositionFor({
    line: 23,
    column: 1
  }));
};
checkSourceMap('girder/clients/web/static/built/girder.app.min.js.map');
```
will output:
```
{ 
  source: 'webpack:///clients/web/src/views/View.js',
  line: 9,
  column: 8,
  name: null 
}
```
now open `View.js`:
```javascript
import _ from 'underscore';
import Backbone from 'backbone';

import events from 'girder/events';
import eventStream from 'girder/utilities/EventStream';

var View = Backbone.View.extend({
    constructor: function (opts) { // eslint-disable-line backbone/no-constructor
        if (opts && _.has(opts, 'parentView')) {
```
and line 9 is indeed:
```javascript
        if (opts && _.has(opts, 'parentView')) {
```
Thanks [mozilla/source-map](https://github.com/mozilla/source-map)

zachmullen:
Great! Sounds like we are very close 😄

sebastienbarre:
Pythoners? @zachmullen? Unfortunately I need to get going on my journey to NC. Could you guys take over please? :)

Two issues:

- Travis CI seems to think it's a good idea to fail the test `js_coverage_combine_report` that creates the Javascript coverage XML file for CDash because the code is not looking "good enough" (is it linting it at the same time? why oh why?),
```
 6: /home/travis/build/girder/girder/tests/js_coverage_tool.py:48: [C901] 'combine_report' is too complex (16)
```

- The remapping from `blanket`'s `.cvg` coverage to the original source files using Webpack's source maps works "fine" on my machine:
```
env) girder-build$ /Users/Shared/Src/girder/env/bin/python2.7 "/Users/Shared/Src/girder/girder/tests/js_coverage_tool.py" "--threshold=75" "--source=/Users/Shared/Src/girder/girder" "--include-core" "combine_report" "/Users/Shared/Src/girder/girder-build/js_coverage"
Overall total: 20350 / 27032 (75.28%)
```
but the [mattrobenolt/python-sourcemap](https://github.com/mattrobenolt/python-sourcemap) package I'm using in `tests/js_coverage_tool.py`'s `combine_report` has issues. Its [`lookup` API](https://github.com/mattrobenolt/python-sourcemap/blob/master/sourcemap/objects.py#L61) expects a line *and* column in a destination file (say in `girder.app.min.js`), which are then remapped to a new line and column in a different (original) file (say in `clients/web/src/views/layout/ProgressListView.js`) using the source map (say `girder.app.min.js.map`). However, `blanket` only generates line numbers in its `.cvg` files (see `girder_build/js_coverage/user.cvg` for example). Even when calling the API with `                                token = currentRemapper.lookup(line=lineNum, column=0)
`, `sourcemap` still throws `IndexError` exceptions where it shouldn't -- it already has the line number we need, but fails trying to compute a column number. We are not interested in the column number and that exception makes us miss a good amount of coverage.

What's the protocol?
- send a patch and create a PR to add an API to `sourcemap` that only accepts a line number?
- fork it, patch it, and use the fork?
- download it and include the source straight in the girder?

Thanks

zachmullen:
I'm on it. See you soon!

sebastienbarre:
Thanks Zach. To help you a bit:
- you can generate a `.cvg` file inside `girder_build/js_coverage/` by running any single test using `ctest -R`
- then run `js_coverage_combine_report` using `ctest -R` as well, this will create the `js_coverage.xml` file in the build tree.
- I've left two debug statements in the code; the second one, after the exception, is the problematic one, it is triggered because of `sourcemap`'s issue with column numbers. The one immediately above it is of the same nature, but is not triggered if `column=0` in the call to `.lookup(` (it *would be* if `column=1`, for example). Either statements would not be hit if there was an API remapping a line number to another line number.

Thanks!

jbeezley:
Travis is only showing [two problems](http://my.cdash.org/viewTest.php?onlyfailed&buildid=1040637).  One is the static analysis (McCabe complexity) of the combining script.  The other is the fontello issue I already fixed on master.

zachmullen:
@jbeezley there are several browser tests failing also, it's just that CTest is segfaulting during the coverage aggregation step, so there is no submission to CDash for the browser build.

zachmullen:
We are green, but we still have web client test failures on travis that are masked by CTest segfaulting. Anyone object if I go ahead and merge this into 2.0-integration and then fix our coverage and web client tests on a subsequent branch?

jeffbaumes:
👏 👏

jcfr:
:fireworks: :tada: :balloon:

grunt_tasks/fontello.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Define tasks that download and copy fontello font files.
+ */
+module.exports = function (grunt) {
+    var path = require('path');
+
+    grunt.config.merge({
+        curl: {
+            fontello: {
+                src: 'https://data.kitware.com/api/v1/file/57c5d1fc8d777f10f269dece/download',
```
sebastienbarre: So I'm curious. Why are we not sticking the Fontello files directly in the repo?

jbeezley: I considered it, but it's around a 0.5MB diff every time we want to add or remove an icon.  I don't care much either way though.

zachmullen: I consider this a temporary solution until the network problems are resolved, or we find a different service entirely.

PR #1477, opened 2016-07-22
Dicom file metadata extractor plugin

sutartmelson:
Similar to the metadata_extractor plugin, this plugin works exclusively on dicom files and only server side. Could easily be expanded to work client side if desired. 

Only public attributes are extracted from the metadata.


zachmullen:
@sutartmelson I solved the mystery of the failing test. In the [adminSpec](https://github.com/girder/girder/blob/master/clients/web/test/spec/adminSpec.js#L572) we now have two plugins that coincidentally happen to match that selector, and somehow activating both of them at once causes things to blow up. I'd suggest just changing the adminSpec.js to use a different test plugin, the following works when I change it on lines 572 and 612:

    $('.g-plugin-list-item:contains(OAuth2 login) .g-plugin-switch').click();

plugins/dicom_metadata_extractor/server/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+###############################################################################
+
+from girder import events
+from girder.constants import AssetstoreType
+
+from . dicom_metadata_extractor import ServerDicomMetadataExtractor
+
+
+def handler(event):
+    if event.info['assetstore']['type'] == AssetstoreType.FILESYSTEM:
```
zachmullen: Is there any reason why this can't work in other assetstore types? Can pydicom read from a stream?

plugins/dicom_metadata_extractor/server/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+        metadataExtractor = ServerDicomMetadataExtractor(event.info['assetstore'],
+                                                         event.info['file'])
```
zachmullen: Maybe we should only call this if a given set of file extensions is present, and only if DICOM metadata doesn't already exist on the parent item?

plugins/dicom_metadata_extractor/server/__init__.py
sutartmelson: I mostly copied this from the metadata_extractor plugin, but I think you make a good point. For some reason it uses paths to the remote files. Instead, I'll write it to use fileId. That way the asset store type shouldn't matter.

cpatrick:
@fogleman ptal

plugins/dicom_metadata_extractor/server/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+#  limitations under the License.
+###############################################################################
+
+from girder import events
+from . dicom_metadata_extractor import ServerDicomMetadataExtractor
+from girder.utility.model_importer import ModelImporter
+
+
+def handler(event):
+    if event.info['file']['exts'] == ['dcm']:
```
fogleman: DICOM files often have no extension.

plugins/dicom_metadata_extractor/server/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+        itemId = event.info['file']['itemId']
+        itemModel = ModelImporter.model('item')
+        item = itemModel.load(itemId, force=True)
+        if 'meta' in item:
```
fogleman: These three `if`s could just be:

    if item.get('meta', {}).get('DICOM Elements Extracted'):

plugins/dicom_metadata_extractor/server/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+            if 'DICOM Elements Extracted' in item['meta']:
```
fogleman: `DICOM Elements Extracted` is a bit verbose IMO.

plugins/dicom_metadata_extractor/server/dicom_metadata_extractor.py
```diff
@@ -0,0 +1,129 @@
...
+            self._setMetadata()
+
+    def _extractMetadata(self):
+        """
+        Extract metadata from file on client or server using hachoir-metadata
+        or pydicom for DICOM images.
+        """
+
+        try:
+            dcMetadata = dicom.read_file(self.path, stop_before_pixels=True)
```
fogleman: Does this plugin only work with filesystem assetstore? It'd be nice if it worked with any assetstore.

plugins/dicom_metadata_extractor/server/dicom_metadata_extractor.py
```diff
@@ -0,0 +1,129 @@
...
+            if dcMetadata:
+                self.metadata = dict()
+                for item in dcMetadata:
+                    # pass forces rawDataElement to DataElement conversion
```
fogleman: What? How?

plugins/dicom_metadata_extractor/server/dicom_metadata_extractor.py
```diff
@@ -0,0 +1,129 @@
...
+                self.metadata = None
+
+        except dicom.filereader.InvalidDicomError:
+            self.metadata = None
+
+    def _setMetadata(self):
+        """
+        Attach metadata to item on server.
+        """
+        pass
```
fogleman: Is this supposed to be blank?

fogleman: Oh, I see you've got some inherited classes below.

plugins/dicom_metadata_extractor/server/__init__.py
zachmullen: I'd suggest checking the MIME type (`application/dicom`) instead.

manthey: Or both.  The `file['exts']` array could have multiple extensions, so the comparison would be better as  `if event.info['file']['exts'][-1:] == ['dcm']:`.  That way, a file name `dicom.with.dots.in.its.name.dcm` would still be handled.

adsorensen:
Sorry things have been a little hectic. I can get to these change requests later today or tomorrow.

zachmullen:
No rush, I was just trying to go through the PR backlog and mark status on stuff using the new github features :smile:

plugins/dicom_metadata_extractor/server/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+###############################################################################
+
+from girder import events
+from . dicom_metadata_extractor import ServerDicomMetadataExtractor
+from girder.utility.model_importer import ModelImporter
+
+
+def handler(event):
+    mime = mimetypes.guess_type(event.info['file'])
+    if event.info['file']['exts'][-1] == ['dcm'] || 'application/dicom':
```
zachmullen: This `||` is not valid python syntax.

PR #1478, opened 2016-07-25
Py client breaking changes

zachmullen:


zachmullen:
@girder/developers PTAL

clients/python/girder_client/cli.py
```diff
@@ -29,10 +28,10 @@ class GirderCli(GirderClient):
...
-    def __init__(self, username, password, dryrun, blacklist,
-                 host=None, port=None, apiRoot=None, scheme=None, apiUrl=None,
-                 apiKey=None):
-        """initialization function to create a GirderCli instance, will attempt
+    def __init__(self, username, password, host=None, port=None, apiRoot=None,
+                 scheme=None, apiUrl=None, apiKey=None):
+        """
+        Initialization function to create a GirderCli instance, will attempt
         to authenticate with the designated Girder instance. Aside from username
         and password, all other kwargs are passed directly through to the
```
mgrauer: Aside from username, password and apiKey, all other ...

mgrauer:
Since you changed the constructor of the GirderCli to use `super`, let's change all the rest to be consistent in style.

    super(GirderCli, self).__init__(


    GirderCommandSubtype
    HttpError
    IncorrectUploadLengthError

mgrauer:
I'm dogpiling on some housekeeping.

Can you move `chunk_size=65536` to a class constant, from `downloadFile`? 

:dog:  :pill:

mgrauer:
Please update python-client.rst with the migrated method names.  From what I saw, it was only the block at the very bottom with callback examples.

mgrauer:
The [cli arg parsing](https://github.com/girder/girder/blob/1f40ded69bbe8590eef96c694e503924a5ae52e9/clients/python/girder_client/cli.py#L70-L168) is too clever for me to understand with a quick read through.  I think breaking it out into subparsers was a great change though.

I'm OK with not rewriting this for now, but let's try to add some comments in the future explaining the magic.  The biggest smell for me is that many of those `_COMMON_OPTIONS` apply only to uploading.  A comment near `_COMMON_OPTIONS` to that effect would be useful as a stopgap.

clients/python/girder_client/cli.py
```diff
@@ -64,44 +62,41 @@ def __init__(self, username, password, dryrun, blacklist,
...
     parent_type=dict(
         longname='--parent-type', required=False, default='folder',
-        help='type of Girder parent target, one of '
-             '(collection, folder, user)'),
+        help='type of Girder parent target, one of (collection, folder, user)'),
+    blacklist=dict(
+        longname='--blacklist', required=False, default='',
+        help='comma-separated list of filenames to ignore'),
+    dryrun=dict(
+        longname='--dryrun', required=False, action='store_true',
```
msmolens: Nit: I'd like this to be named `--dry-run` and the associated variables updated appropriately, e.g. `dryRun`. This would make this option consistent with the others that are longer than one word (e.g. `--parent-type`, `--leaf-folders-as-items`). Moreover, it would make this option consistent with that in other tools (e.g. git).

mgrauer:
Thanks for simplifying, LGTM.

PR #1481, opened 2016-07-26
Fix typo in --parent-type option name

zachmullen:
@mgrauer I made a typo during refactoring that was not caught by our tests, PTAL.


mgrauer:
LGTM, thanks for adding the test.

Issue #1479, opened 2016-07-26
Settings collection keys should be unique


manthey:
At least two people have accidentally duplicated settings, so that the mongo setting collection contains multiple entries with the same `key` value.  The `key` index in setting should be specified as `self.ensureIndices([('key', {'unique': True})])`.

This is easy to modify, but, we need to choose how Girder would migrate to this setting (at the least, it will conflict with an extant index autonamed `key_1`.   Do we automatically drop and readd indices that collide with identical auto-generated names, or do we just make up a new name for this index?  If there are non-unique settings, does upgrading girder break until the admin manually cleans up these issues, or should we attempt to fix extant problems?

zachmullen:
What does mongo do if you try to add a unique index to a collection that would violate the unique constraint?

manthey:
It throws a `pymongo.errors.DuplicateKeyError` exception.

manthey:
Just giving the unique index a different name is insufficient.  Mongo objects
```
pymongo.errors.OperationFailure: Index with pattern: { key: 1 } already exists with different options
```

I think this means we would need to drop the extant index.

zachmullen:
But the data itself would still violate the new index by virtue of having two settings with the same key field, so wouldn't adding a new index also be problematic?

manthey:
Yes.  If we had had a unique key requirement all along then we would have prevented such problems.  If we want to switch to having a unique key requirement and there are duplicate keys, then we either have to "fix" the existing data (whatever that means), fail and require an admin to fix it, or display a warning on start and not switch to a unique key constraint.

zachmullen:
How about we try to make the index, and if we catch the DuplicateKeyError, we issue a warning, arbitrarily choose one and remove the rest, and then try again?

manthey:
Fixed in PR #1482.

PR #1482, opened 2016-07-27
Add a unique-key constraint to the setting collection.

manthey:
This addresses issue #1479.

If there are duplicate settings with the same key, the one with the earliest Mongo _id is kept, and later ones discarded.  Generally, this seems to be the value returned by findOne, but it isn't guaranteed.


zachmullen:
I think as an arbitration policy it probably makes more sense to keep the latest value and throw out older ones.

manthey:
I'm torn on this.  When we use a settings end point, we get the result of findOne(), which returns the " first document according to the natural order which reflects the order of documents on the disk", which is more likely to be the earliest value than the latest value.

zachmullen:
Good point, LGTM.

PR #1483, opened 2016-07-28
Change how we validate settings

zachmullen:
We now provide a convenient decorator method for registering setting validation functions. This alleviates the needs for plugins to bind to the setting validate event and check the keys themselves, although that will still work as before.

Thanks to @mgrauer for floating this idea in the 2.0 wish list. Since it turns out it's not a breaking change, I've PRed it against master.

### DIFF
--- a/girder/models/setting.py
+++ b/girder/models/setting.py
@@ -22,9 +22,10 @@ import cherrypy
 import pymongo
 import six
 
-from ..constants import SettingDefault, TerminalColor
+from ..constants import SettingDefault, SettingKey, TerminalColor
 from .model_base import Model, ValidationException
-from girder.utility import camelcase, plugin_utilities
+from girder.utility import plugin_utilities, setting_utilities
+from girder.utility.model_importer import ModelImporter
 from bson.objectid import ObjectId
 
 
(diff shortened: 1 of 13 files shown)


zachmullen:
It's also worth noting that this is a lot more efficient; rather than linearly iterating through all registered callbacks, we can just look up the required key in a hash table.

manthey:
I like this.  I note that there is at most one registered validator per setting key.  I don't have any specific case where the validation could be split across two functions, but we might want to add a note that this can be accomplished with event handling, and throw a warning if `registerValidator` replaces an existing validator (with a flag to skip such a warning).

zachmullen:
Alternatively, we could add a `replace` bool argument to `registerValidator` that, if set to False, would make it so the new function was just chained together after the existing one via function wrapping.

cjh1:
Am I right is thinking that setting keys are global, so a poor choice of name can result in two plugins trying to using the same setting?

zachmullen:
Setting keys are global, but all the core ones start with `core.`, and we instruct plugin authors to prefix the name of their setting keys with their plugin name followed by a dot. So, yes, violating this convention could cause problematic collisions, though that was true before this change.

zachmullen:
@manthey I wrote it out, what do you think?

```
def registerValidator(key, fn, replace=False):
    """
    Register a validator for a given setting key.

    :param key: The setting key.
    :type key: str
    :param fn: The function that will validate this key.
    :type fn: callable
    :param replace: If a validator already exists for this key, set this to True to replace the
        existing validator. The default is to add the new validator in addition to running the
        old validation function.
    :type replace: bool
    """
    if not replace and key in _validators:
        old = _validators[key]
        def wrapper(doc):
            old(doc)
            fn(doc)
        _validators[key] = wrapper
    else:
        _validators[key] = fn
```

manthey:
Because validation functions can alter values, would it make more sense to call the new function *before* the old function?  For instance, suppose there is a validator that checks for a valid email address.  A plugin could add a default domain to an email address.  The plugin adds a validator that adds the domain if necessary, then the original validator of the email address would pass instead of failing.

zachmullen:
Fixed, PTAL

manthey:
@zachmullen Can you add a test for the `default` decorator?

manthey:
LGTM.

PR #1484, opened 2016-07-28
Add route table and global webroot registry

danlamanna:
This PR proposes that plugins which want to create their own webroot do so without modifying the global configuration object directly.

Instead, the plugin would do something like this:
```
from girder.utility.plugin_utilities import registerPluginWebroot

def load(info):
    registerPluginWebroot(SomeWebroot(), info)
```

This creates a mapping in a global `_plugin_webroots` object from plugin name to webroot object. Girder will pick up on this (on server start) and create an entry in the `core.route_table` setting which defaults to `pluginName: /pluginName`.

The route table can be modified under `Settings -> Server Configuration`. Some clarity for this:
1) Changing routes requires a server restart
2) Setting a route to the empty string will effectively disable its webroot
3) Girder has to be mounted somewhere

I still need to add something to developer docs to clarify the workflow here, and possibly make the naming clearer - but I'll ask for some feedback before going that far.


clients/web/src/templates/body/systemConfiguration.jade
```diff
@@ -129,6 +129,18 @@ form.g-settings-form(role="form")
...
+    h4 Routing
+    p.
+      Altering routes requires a restart of the Girder server to take effect.
+    .form-group
+      each route, name in settings['core.route_table'] || defaults['core.route_table']
+        label #{name}
+        input.g-core-route-table.form-control.input-sm(
+          type="text",
+          value="#{route || ''}",
+          data-webroot-name="#{name}",
```
zachmullen: `data-webroot-name=name,`

clients/web/src/templates/body/systemConfiguration.jade
```diff
@@ -129,6 +129,18 @@ form.g-settings-form(role="form")
         type="text", value="#{settings['core.smtp.password'] || ''}",
         placeholder="Default: #{defaults['core.smtp.password'] || 'none'}",
         title="The password, if any, used to authenticate to the SMTP server used to send emails.")
+  .g-settings-form-container
+    h4 Routing
+    p.
+      Altering routes requires a restart of the Girder server to take effect.
+    .form-group
+      each route, name in settings['core.route_table'] || defaults['core.route_table']
+        label #{name}
```
zachmullen: `label=name`

girder/models/setting.py
```diff
@@ -215,6 +215,22 @@ def validateCoreUserDefaultFolders(self, doc):
                 'User default folders must be either "public_private" or '
                 '"none".', 'value')
 
+    def validateCoreRouteTable(self, doc):
```
zachmullen: Note that this will break after #1483 ... we should rebase this on top of that change once it's merged.

girder/models/setting.py
```diff
@@ -215,6 +215,22 @@ def validateCoreUserDefaultFolders(self, doc):
...
+        nonEmptyRoutes = [route for route in doc['value'].values() if route]
+        if 'girder' not in doc['value'] or not doc['value']['girder']:
+            raise ValidationException('Girder must be routeable.')
+
+        if not all([route.startswith('/') for route in nonEmptyRoutes]):
```
zachmullen: `all(route.startswith('/') for route in nonEmptyRoutes)`

girder/models/setting.py
```diff
@@ -215,6 +215,22 @@ def validateCoreUserDefaultFolders(self, doc):
                 'User default folders must be either "public_private" or '
                 '"none".', 'value')
 
+    def validateCoreRouteTable(self, doc):
+        nonEmptyRoutes = [route for route in doc['value'].values() if route]
+        if 'girder' not in doc['value'] or not doc['value']['girder']:
```
zachmullen: `if not doc['value'].get('girder'):`

girder/utility/plugin_utilities.py
```diff
@@ -46,6 +46,8 @@
 from girder.models.model_base import ValidationException
 from girder.utility import config as _config, mail_utils, mkdir
 
+_plugin_webroots = {}
```
zachmullen: camelCase for consistency

girder/utility/server.py
```diff
@@ -20,10 +20,12 @@
 import cherrypy
 import mimetypes
 import os
+import six
 
 import girder.events
 from girder import constants
 from girder.utility import plugin_utilities, model_importer
+from girder.utility.plugin_utilities import _plugin_webroots  # noqa
```
zachmullen: If we need access to this from outside the module, let's make an accessor method `getPluginWebroots()` in the public namespace.

girder/models/setting.py
```diff
@@ -215,6 +215,22 @@ def validateCoreUserDefaultFolders(self, doc):
...
+            raise ValidationException('Girder must be routeable.')
```
mgrauer: Something more expansive here, perhaps

    Girder must have a valid webroot and route

mgrauer:
👍 for the explicitness of requiring a server restart

danlamanna:
@zachmullen @jeffbaumes This should be ready for review.

jeffbaumes:
Code logic, testing, and docs LGTM. Good sign that this simplifies the "Serving a custom app from the server root" documentation. +1 for the restart link and generalizing that code.

I will leave to @zachmullen for a final sign off.

docs/developer-cookbook.rst
```diff
@@ -447,34 +447,17 @@ the server root instead, and they may also want to move the Girder web client to
...
-
-This will now serve the api out of *both* ``/api`` and ``/girder/api``, which
-may be desirable. If you only want it to be served out of ``/api`` and not
-``/girder/api``, just add a final line below that:
-
-.. code-block:: python
-
-    del info['serverRoot'].girder.api
+This will register your ``CustomAppRoot`` with Girder so that it can then be mounted
+whereever an Administrator specifies using the Server Configuration Panel. See
```
zachmullen: wherever*

docs/developer-cookbook.rst
```diff
@@ -447,34 +447,17 @@ the server root instead, and they may also want to move the Girder web client to
 be served out of an alternative route so they can still use it in addition to
 their custom front-end application.
```
jeffbaumes: Since by default apps are not served at the root, we should probably change the section heading to something like "Mounting a custom application" and modify the first paragraph to something like

```
Mounting a custom application
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Normally, the root node (``/``) of the server will serve up the Girder web client.
A plugin may contain an entire application separate from the default Girder
web client. This plugin may be written in a way which enables administrators
to mount the application at a configured endpoint, including the option of
replacing the root node with the plugin application.
```

zachmullen:
Do we enforce that something always has to be mounted on `/`? I was able to mount girder somewhere else with nothing mounted at the root.

docs/configuration.rst
```diff
@@ -76,3 +76,16 @@ to take effect.
...
 ``/etc/security/limits.conf`` for details.
+
+.. _managing-routes:
+
+Managing Routes
+---------------
+
+When plugins which have their own custom webroot are enabled, they are mounted at /pluginName.
+In certain cases it may be desirable for the site administrator to mount such plugins at their own
+specificied paths.
```
mgrauer: specified

docs/developer-cookbook.rst
```diff
@@ -438,43 +438,26 @@ by the environment variable ``GIRDER_TEST_DATA_PREFIX`` as follows
...
+Mounting a custom application
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Normally, the root node (``/``) of the server will serve up the Girder web client.
-Some plugins will wish to change this so that their own custom app gets served out of
-the server root instead, and they may also want to move the Girder web client to
-be served out of an alternative route so they can still use it in addition to
-their custom front-end application.
+A plugin may contain an entire application separate from the default Girder
+web client. This plugin may be written in a way which enables administrators
```
mgrauer: What does it mean that the "plugin may be written in a way which enables administrators to mount the application" ?  Is the previous sentence sufficient?  This sentence I'm asking about makes me think my plugin would have to do something specific beyond providing an entire web application.

mgrauer: Never mind, I just needed to read a bit further down.

danlamanna:
@zachmullen We don't, my main concern was keeping Girder associated to *some* route, aside from that I would prefer less restrictions (as long as it makes sense). @jeffbaumes Do you have an opinion one way or the other?

jeffbaumes:
I feel it is more flexible to allow for the case that nothing is at the root. It also seems hard to enforce. For example, if a plugin is assigned to the root and then the plugin is disabled, what do we do to enforce that constraint?

zachmullen:
Sounds good. My one other suggestion would be to make the UI show the core app always at the top of the list, with some special label or styling to indicate that it is different from the others.

danlamanna:
@zachmullen Right - I found it a bit difficult to do elegantly, open to critiques.

zachmullen:
Maybe just order the list before passing it into the template, then do some different logic for the first list element?

zachmullen:
One thing I'm noticing is that if I move the core app, the `static` subdirectory also moves with it, so that static content is no longer served out of `/static`. Is that expected? Should we change that behavior?

zachmullen:
Could you merge master into this to hopefully get CI green again?

danlamanna:
>
One thing I'm noticing is that if I move the core app, the static subdirectory also moves with it, so that static content is no longer served out of /static. Is that expected? Should we change that behavior?

Yeah, this is tricky. Ideally everything for Girder would always be attached to it and we wouldn't pollute anything else. However, we made an exception for `api/v1` since determining the location of it from a purely HTML application is a bit tricky.. maybe we should revisit that.

zachmullen:
> However, we made an exception for api/v1 since determining the location of it from a purely HTML application is a bit tricky.. maybe we should revisit that.

Whatever rules we apply to the API root should also apply to the static root I think.

zachmullen:
The API root and static root are one of the only things that are configured via the config file instead of the database; maybe it is time to move those things into database settings and then expose them right alongside this new routing table.

jeffbaumes:
@zachmullen I like the db config idea. One question: Does the static dir serve anything other than files to support the standard Girder client? Why is it configurable at all - can't it just always sit with the Girder app?

jeffbaumes:
For that matter, why is the API root configurable? Why can't it always be at api/v1?

jbeezley:
If girder is mounted on the webserver somewhere other than at the root level via apache or nginx, then it needs to be changed so that the client side code hits the correct path.

zachmullen:
We could force it so that the API had to be served out of `/api/v1`, but then if someone wanted to host a girder instance at a path on their server besides the root path, it would make that impossible. That's the reason it's in config files and not the database right now: people typically only interact with the database *via* the API, which could lead to some cases of being unable to use the web client to interact with the settings.

jeffbaumes:
Thanks @jbeezley @zachmullen. So if I understand correctly, the API config parameter is what is used client-side to access the API, not to serve the API. So regardless of the setting, the API will be served at `api/v1` relative to the Girder CherryPy root, but the client app may need to access it from another place because of web server configuration?

zachmullen:
That's right, although plugins could also move the API if they had some reason to do so, but I doubt they would.

danlamanna:
Static dir is now configurable via the web app. I *think* this is ready to be looked at again.

zachmullen:
@danlamanna sorry to do this, but would you mind converting this to target 2.0-integration since it touches some of the javascript? I'm happy to help with that if you want.

danlamanna:
@zachmullen This is rebased against 2.0. There are some changes in the diff that will go away the next time master is merged into 2.0 I think.

Unfortunately the branch isn't actually working on my machine, so I may have messed something up. I'll have to look into it next week.

zachmullen:
Just merged master into 2.0-integration.

danlamanna:
Perfect, thanks.

clients/web/src/views/body/SystemConfigurationView.js
```diff
@@ -51,7 +60,8 @@ var SystemConfigurationView = View.extend({
         },
         'click .g-edit-collection-create-policy': function () {
             this.collectionCreateAccessWidget.render();
-        }
+        },
+        'click #g-restart-server': girder.restartServerPrompt
```
zachmullen: Shouldn't this be imported rather than referenced under ``girder``?

girder/utility/server.py
```diff
@@ -131,24 +129,97 @@ def configureServer(test=False, plugins=None, curConfig=None):
...
+            cherrypy.tree.mount(pluginWebroots[name], route, appconf)
 
     if test:
         application.merge({'server': {'mode': 'testing'}})
 
     return application
 
 
+class BaseWebroot(object):
+    expose = True
```
zachmullen: Did you mean `exposed = True`? Or is this something else?

girder/utility/server.py
```diff
@@ -131,24 +129,97 @@ def configureServer(test=False, plugins=None, curConfig=None):
...
+
+    def __init__(self):
+        pass
```
zachmullen: Seems like we don't need to include this `__init__` definition at all.

zachmullen:
When I try to start my server after checking out this branch, I get the following error:

```
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/zach/dev/girder/girder/__main__.py", line 57, in <module>
    main()
  File "/Users/zach/dev/girder/girder/__main__.py", line 50, in main
    server.setup(args.testing)
  File "girder/utility/server.py", line 187, in setup
    girderWebroot, appconf = configureServer(test, plugins, curConfig)
  File "girder/utility/server.py", line 116, in configureServer
    'staticRoot': os.path.relpath(routeTable[constants.GIRDER_STATIC_ROUTE_ID],
KeyError: 'core_static_root'
```

danlamanna:
@zachmullen My guess it that you're not loading the new key that was added to the default setting. Are you using this with same db you used before? If so I think you'll have to manually delete the setting key via mongo so it picks up the default first.

zachmullen:
@danlamanna I fixed the two issues I cited on that last commit.

Issue #1486, opened 2016-07-29
girder ansible module: making public collection doesn't seem to work


zachmullen:
The following doesn't work:

```
girder:
    username: "{{ girder_admin_user }}"
    password: "{{ girder_admin_password }}"
    port: "{{ girder_socket_port }}"
    collection:
      name: TCGA
      public: true
    state: present
```

The collection is created as private. I also tried `public: yes`, but no luck that way either.

@kotfic is this an easy fix?

PR #1487, opened 2016-08-01
Fix issue with set/update of public on collections

kotfic:
This should resolve an issue with setting and updating the 'public' attribute on collections.  It bumps the girder ansible module client to 0.2.4 (ref #1486) @zachmullen PTAL

### DIFF
--- a/clients/ansible/test/test_access.yml
+++ b/clients/ansible/test/test_access.yml
@@ -253,8 +253,8 @@
           name: "Test Collection"
           description: "A test collection"
           public: yes
-          accesss: []
       register: test_collection
+
 
     # Check that test_collection ispublic again
 
--- a/clients/ansible/test/test_access.yml
+++ b/clients/ansible/test/test_access.yml
@@ -212,8 +212,20 @@
 
 ####
 #  Test access control
+    - name: Create collection with public access
+      girder:
+        port: 8080
+        username: "admin"
+        password: "letmein"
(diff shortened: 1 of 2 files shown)


kotfic:
Yes? I assumed this was the intended functionality when making a collection public?

kotfic:
Nevermind,  as soon as i posted that I realized why it was stupid.  I will make sure to pull the access and then resend it with the updated public flag.  This is a little awkward though in terms of the API.

zachmullen:
Roger. I'm OK with adding a public flag to the collection creation endpoint, so that it will at least be easier in the future.

kotfic:
@zachmullen had to tweak the _access function to allow for processing access lists from girder (e.g. lists with levels and Id's rather than names and human-readable levels). but otherwise this is ready for review

clients/ansible/test/test_access.yml
```diff
@@ -230,6 +242,38 @@
...
+    - name: Make the collection public again
+      girder:
+        port: 8080
+        username: "admin"
+        password: "letmein"
+        collection:
+          name: "Test Collection"
+          description: "A test collection"
+          public: yes
+          accesss: []
```
zachmullen: This typo is still present

zachmullen:
I've verified that this works in both the create and update case now, thanks for the fix! LGTM once you fix that typo in the test.

PR #1488, opened 2016-08-01
use shutil.move instead of os.rename

opadron:
Using `os.rename` causes problems when renaming across filesystems.

See [here](http://stackoverflow.com/a/11593707),
[here](https://mail.python.org/pipermail/tutor/2011-April/083076.html),
and the documentation [here](https://docs.python.org/3.4/library/shutil.html#shutil.move).

We ran into an issue on SUMO, and we think fixing this issue will address it.  Details [here](https://github.com/osumo/osumo-project/pull/3).


zachmullen:
LGTM

PR #1490, opened 2016-08-02
Thumbnails dicom

adsorensen:
Modified thumbnails plugin to handle dicom images.


plugins/thumbnails/server/worker.py
```diff
@@ -157,3 +159,20 @@ def attachThumbnail(file, thumbnail, attachToType, attachToId, width, height):
...
+    Check extension of image and opens it.
+
+    :param extension: The extension of the image that needs to be opened.
+    :param data: The image file stream.
+    """
+    if 'dcm' in extension:
+        # Open the dicom image
+        dicomData = dicom.read_file(six.BytesIO(data))
+        imageData = dicomData.pixel_array
+        return Image.fromarray(imageData).convert("I")
```
fogleman: Looks like this assumes a particular pixel format for the image data. Probably won't work for all DICOM images.

sutartmelson: Hm. As far as I understand, dicom.pixel_array should be consistent. [link](http://pydicom.readthedocs.io/en/stable/working_with_pixel_data.html)

Also, the convert doesn't assume a specific pixel format that its converting from, so it should work regardless.. right?

plugins/thumbnails/server/worker.py
```diff
@@ -157,3 +159,20 @@ def attachThumbnail(file, thumbnail, attachToType, attachToId, width, height):
...
+
+def getImage(extension, data):
+    """
+    Check extension of image and opens it.
+
+    :param extension: The extension of the image that needs to be opened.
+    :param data: The image file stream.
+    """
+    if 'dcm' in extension:
+        # Open the dicom image
```
sutartmelson: Need to check if this dicom file's pixel data is 2d or 3d and generate thumbnail accordingly.

plugins/thumbnails/server/worker.py
fogleman: Ah, maybe you're right.

fogleman: Try this file. The output is basically black. Levels need to be scaled up I think.

[000000.dcm.zip](https://github.com/girder/girder/files/399609/000000.dcm.zip)

plugins/thumbnails/server/worker.py
```diff
@@ -157,3 +160,53 @@ def attachThumbnail(file, thumbnail, attachToType, attachToId, width, height):
     }
 
     return ModelImporter.model('file').save(thumbnail)
+
+
+def getImage(extension, data):
```
zachmullen: Let's make this method private for the moment (`_getImage`).

plugins/thumbnails/server/worker.py
```diff
@@ -157,3 +160,53 @@ def attachThumbnail(file, thumbnail, attachToType, attachToId, width, height):
...
+    """
+    Check extension of image and opens it.
+
+    :param extension: The extension of the image that needs to be opened.
+    :param data: The image file stream.
+    """
+    if 'dcm' in extension:
```
zachmullen: Let's also pass the `file['mimeType']` so it can be used to detect type. Then we can do:

    if  (extension and extension[-1] == 'dcm') or mimeType == 'application/dicom':

plugins/thumbnails/server/worker.py
```diff
@@ -157,3 +160,53 @@ def attachThumbnail(file, thumbnail, attachToType, attachToId, width, height):
...
+        finalImage = __win_lev(imageData[0], maximum-minimum, (maximum+minimum)/2)
+        return Image.fromarray(finalImage).convert("I")
+    else:
+        minimum = imageData.min() + offset
+        maximum = imageData.max() + offset
+        finalImage = __win_lev(imageData, maximum-minimum, (maximum+minimum)/2)
+        return Image.fromarray(finalImage).convert("I")
+
+
+def __win_lev(img, window, level, maxc=255):
```
zachmullen: Let's name this `_scaleIntensity`.

zachmullen:
This is a great feature, but I'm not wild about the fact that it introduces numpy as a hard dependency. I think an easy way around this is to try/except the numpy import, and if numpy isn't found, if someone tries to thumbnail a dicom file, we raise an exception telling them it's because numpy isn't installed.

plugins/thumbnails/server/worker.py
```diff
@@ -22,7 +22,11 @@
 import sys
 import traceback
 import dicom
-import numpy as np
+try:
+    import numpy as np
+    numpy = True
+except Exception:
```
zachmullen: `except ImportError:`

plugins/thumbnails/server/worker.py
```diff
@@ -162,14 +165,16 @@ def attachThumbnail(file, thumbnail, attachToType, attachToId, width, height):
...
     """
     Check extension of image and opens it.
 
     :param extension: The extension of the image that needs to be opened.
     :param data: The image file stream.
     """
-    if 'dcm' in extension:
+    if (extension and extension[-1] == 'dcm') or mimeType == 'application/dicom':
+        if not numpy:
+            raise Exception('Missing dependancy: numpy.')
```
zachmullen: Minor nit, but let's change this error message to:

    raise Exception('Could not create DICOM thumbnail: numpy not installed.')

zachmullen:
I just tested this and it works, now it just needs testing! [This article](http://danielnouri.org/notes/2012/11/23/use-apt-get-to-install-python-dependencies-for-travis-ci/) shows how to install the pre-built numpy binary package via apt and get it working for travis. Since it's prebuilt, it shouldn't take too long.

sutartmelson:
@zachmullen  The solution in the article you linked appears to only work in python 2.7, [link](https://docs.travis-ci.com/user/languages/python#Travis-CI-Uses-Isolated-virtualenvs)

zachmullen:
LGTM, thanks for implementing this awesome feature!

PR #1493, opened 2016-08-08
Fixes #1492. Hide SMTP password setting in UI

### LINKED ISSUE #1492: SMTP and any other passwords in browser views should not be plain text

zachmullen:


cpatrick:
LGTMWTIH.

PR #1496, opened 2016-08-09
Make sure web_client_admin test does not run parallel to any other tests

zachmullen:
The `web_client_admin` test sporadically fails by getting SIGKILLed by the system due to OOM. This attempts to workaround the memory limitations on travis by ensuring that test does not run in parallel with any others.


cpatrick:
LGTM, but I'd rather use a resource than go serial.

PR #1499, opened 2016-08-10
DICOM Viewer Plugin

fogleman:
@cpatrick

This plugin adds DICOM viewing / playback to the item page. Place your DICOM files in a single item to view all of the slices.

screen shot 2016-08-10 at 2 30 58 pm

VTK.js is used for rendering. Some changes were made in the VTK.js repo to support this plugin. Later, we may move some of the DICOM logic into VTK.js itself.

DICOM tags are parsed out on file upload and stored in mongo. This way multiple images can be quickly sorted by SeriesNumber and InstanceNumber. A new endpoint `/item/{id}/dicom` provides DICOM metadata for all DICOM files in the item. You can specify filters here if you only care about certain DICOM tags.

Currently only uncompressed monochrome DICOM files are supported.


package.json
```diff
@@ -16,32 +16,45 @@
...
+    "dicom-parser": "^1.7.2",
+    "grunt-webpack": "^1.0.14",
+    "jade-loader": "^0.8.0",
+    "kw-web-suite": "^2.0.1",
+    "load-grunt-tasks": "^3.5.0",
+    "path": "^0.12.7",
+    "string-replace-loader": "^1.0.3",
+    "stylus-loader": "^2.1.2",
+    "webpack": "^1.13.1",
+    "webpack-dev-server": "^1.14.1"
```
zachmullen: I think we should add these new requirements inside the plugin rather than to the core package.json.

This is also a good time to start thinking about a way to only build the web client code for plugins that are actually enabled, or at least some specified subset of the plugins so we can skip the ones we don't need.

fogleman: I gave that a shot but I'm having trouble getting it to work. Might need some help.

jbeezley: Some of these will probably need to be provided as "grunt.dependencies" in the `plugin.json` (some explanation [here](http://girder.readthedocs.io/en/latest/plugin-development.html#installing-custom-dependencies-from-npm)).  I also wouldn't be surprised if the post install step of vtk.js causes problems here as well.  I can provide some advice if that is the case.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+        .param('id', 'The item ID', paramType='path')
+        .param('filters', 'Filter returned DICOM tags (comma-separated).', required=False)
+        .errorResponse('ID was invalid.')
+        .errorResponse('Read permission denied on the item.', 403)
+    )
+    def getDicom(self, item, params):
+        filters = set(filter(None, params.get('filters', '').split(',')))
+        files = list(ModelImporter.model('item').childFiles(item))
+        # process files if they haven't been processed yet
+        for i, file in enumerate(files):
```
cpatrick: I'm not a bit reserved word-as-variable fan.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+            if 'dicom' not in file:
+                files[i] = process_file(file)
+        # filter out non-dicom files
+        files = [x for x in files if x.get('dicom')]
+        # sort files
+        files = sorted(files, key=sort_key)
+        # execute filters
+        if filters:
+            for file in files:
```
cpatrick: reserved word again

plugins/dicom_viewer/server/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+                dicom = file['dicom']
+                dicom = dict((k, dicom[k]) for k in filters if k in dicom)
+                file['dicom'] = dicom
+        return files
+
+
+def sort_key(file):
```
cpatrick: reserved word

cpatrick:
LGTM :).

This is really cool.

cpatrick:
I do think it would be useful to store the DICOM headers in [the standard way](http://dicom.nema.org/dicom/2013/output/chtml/part18/sect_F.2.html).

msmolens:
We have some WIP conversion routines between pydicom and the DICOM JSON Model: https://github.com/girder/girder/blob/7ec860feab6c46db9d1e50dbe1de912613a86200/plugins/dicom/plugin_tests/conversion_test.py

fedorov:
As pointed out by @pieper, you should fix handling of patient orientation - it looks like the patient and the table were hanging from the ceiling in this image!

jcfr:
@fedorov We indeed noticed that last week and handling of orientation will be addressed.

fogleman:
Superseded by https://github.com/girder/girder/pull/1717

PR #1500, opened 2016-08-10
Add a utility for filtering cherrypy log messages.

manthey:
This lets you reduce cherrypy logging for specific endpoints.  For example, suppose you are tired of seeing every /api/v1/notification/stream request, you can do
```
from girder.api import filter_logging
filter_logging.addLoggingFilter('GET (/[^/ ?#]+)*/notification/stream[/ ?#]', None)
```
and never see them again.  Or you could ask to log only one request in _n_.


girder/api/filter_logging.py
```diff
@@ -0,0 +1,89 @@
...
+
+import cherrypy
+import logging
+import re
+
+
+LoggingFilters = []
+
+
+class regexLoggingFilter(logging.Filter):
```
zachmullen: Pep8 normally complains if you have a class name starting with a lowercase, are we sure flake8 is checking this file?

manthey:
I realized if you were having cherrypy logging to both stdout and a file, this was counting a message to each log as a separate instance of the message.  I refactored things a bit to work around that.

brianhelba:
I'm concerned that this may not be the best way to accomplish this objective:
* For something as fundamental as logging, should we be configuring this at run-time? Since log files are only readable by the machine administrator, shouldn't they be fully setup via the `girder.local.config` file too?
* Do we need to add a brand-new API category for this, or would it belong under `/system`? I'm concerned about Swagger becoming cluttered with a functionality most users will never need or have permissions to use.
* It's it a standard devops practice to your logs via `grep` when you need something?
* If we want to support more sophisticated logging redirection and filtering, we should be using Python's logging mechanism properly, right now, we're outputting some things to the logging mechanism which gets written to `~/.girder/logs/` and other things via print statements to stdout and stderr. IMO, a daemon application (web servers included) should not output anything directly to stdout or stderr, but should send everything to the log mechanism. As a convenience, we can then output some log content to stdout / stderr when in development mode.

manthey:
@brianhelba This is intended to reduce log volume in specific instances, regardless of where the logs are sent.  Most of your points relate to girder logging in general, rather than this PR.

- The log files are expected to be configured via the girder config file.  However, being a python program, there is no reason that a plugin has to only use this mechanism.  This code is agnostic to how and where the logs are configured (though the test attaches a handler to explicitly test things).

- This doesn't add any endpoints.  It is expected that a few specific routes would add filters to reduce logging volume.  I have a requested use case for a high-volume endpoint to reduce log file size.  In some instances, that endpoint could generate 32kb of logs per second from a single user, and the concern was that if there were multiple users, this could start becoming significant.

  Conceptually, this could be a route option (but since two routes could theoretically hit the same endpoint function, and could be filtered differently, this wouldn't be an endpoint decorator).

- grep works wonderfully until you have a lot of churn and the files rotate away before you can use them.  We probably should (in an unrelated PR) make it so that MAX_LOG_SIZE and LOG_BACKUP_COUNT are configurable.

- I agree that we should send everything though the logging mechanism, but that it unrelated to this PR and issue.

zachmullen:
LGTM

PR #1503, opened 2016-08-12
Fix a string encoding error on test failure

jbeezley:
I've seen the following error from time to time (e.g. see [here](http://my.cdash.org/testDetails.php?test=26779805&build=1022716)):
```
  File "tests/web_client_test.py", line 185, in testWebClientSpec
    sys.stdout.write(line)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 29-30: ordinal not in range(128)
```

While this isn't causing any errors, it is preventing the actual error message from being printed out.  I'm not entirely sure why stdout is being coerced into ascii, it could be the locale on travis or something having to do with piping the output through ctest, but this should ensure the conversion happens without raising any errors in the future.

### DIFF
--- a/tests/web_client_test.py
+++ b/tests/web_client_test.py
@@ -182,7 +182,7 @@ class WebClientTestCase(base.TestCase):
                     hasJasmine = True
                 if 'Testing Finished' in line:
                     jasmineFinished = True
-                sys.stdout.write(line)
+                sys.stdout.write(line.encode('utf8', 'replace'))
                 sys.stdout.flush()
             returncode = task.wait()
             if not retry and hasJasmine and jasmineFinished:


msmolens:
I've hit this locally when printing to the console. IIRC I set `PYTHONIOENCODING` to resolve the issue. See http://chase-seibert.github.io/blog/2014/01/12/python-unicode-console-output.html as an example.

tests/web_client_test.py
```diff
@@ -182,7 +182,7 @@ def testWebClientSpec(self):
                     hasJasmine = True
                 if 'Testing Finished' in line:
                     jasmineFinished = True
-                sys.stdout.write(line)
+                sys.stdout.write(line.encode('ascii', 'replace'))
```
zachmullen: Why not `line.encode('utf8')`?

zachmullen: It's trying to internally coerce it and using the default 'ascii' codec because it is being passed a string rather than bytes, but I'm curious if we pass it utf8-encoded bytes if it will be able to display them rather than having to hide all the non-ascii.

jbeezley:
@zachmullen It looks like encoding as utf8 does the trick.  I went ahead and made it ignore encoding errors as well to avoid raising exceptions in case invalid bytes happen to be returned for whatever reason.

zachmullen:
Great, LGTM.

Issue #1072, opened 2015-11-05
Create utility for gating multiple async events


zachmullen:
It's a fairly common pattern to want to wait for several asynchronous events at the same time, and only proceed with a callback once all of them have finished. It's trivial in the case of a single asynchronous event, but when there is more than one, you either have to chain them consecutively, or maintain your own data structure and logic for testing whether they're all finished or not, which is cumbersome. It would be nice to add a utility to Girder core with the following semantics:

```
var gate = new girder.AsyncGate(function () {
    // behavior to be executed when all registered events have finished
    this.render();
}, this)
    .waitFor(someModel, 'g:fetched')
    .waitFor(someCollection, 'g:fetched');
```

Then it's easy enough to add things like timeout, error handlers, etc. onto this prototype.

zachmullen:
Sorry, not sure why "gate" was in my head, what I was thinking of was "fence". If anyone can come up with better nomenclature I'm all ears.

manthey:
jquery uses `deferred`, `when`, and `promise`.

mgrauer:
Yes, this would definitely be useful, and I hope would have a slimming effect.

```
  session.set({
        _id: id
  }).once('m:fetched', function () {
        var datasetsCollection = new minerva.collections.DatasetCollection();
        datasetsCollection.once('g:changed', function () {
            var analysisCollection = new minerva.collections.AnalysisCollection();
            analysisCollection.once('g:changed', function () {
                var sourceCollection = new minerva.collections.SourceCollection();
                sourceCollection.once('g:changed', function () {
                    girder.events.trigger('g:navigateTo', minerva.views.SessionView, {
                        analysisCollection: analysisCollection,
                        datasetsCollection: datasetsCollection,
                        sourceCollection: sourceCollection,
                        session: session
                    });
                }).fetch();
            }).fetch();
        }).fetch();
  }, this).on('g:error', function () {
        minerva.router.navigate('sessions', {trigger: true});
  }, this).fetch();
```

cjh1:
angular has:

```
$q.all(...)
```

This takes an array promises and return a promise that is resolved when all the others have been resolved ....

cjh1:
(Javascript 6 uses the same sort of syntax ...) [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all]

jbeezley:
[...]

PR #1504, opened 2016-08-12
Return promises for async methods on models

jbeezley:
Note: This is a PR targeted at the 2.0 integration branch.

This makes the backward incompatible change to all async methods on models and collections to return `jqXHR` objects rather than `this`.  This change is made for two reasons:

1. This is the signature of the [superclass methods](http://backbonejs.org/#Model-fetch).
2. It allows you to avoid the callback hell described in #1072 by using `Promise.all` or `$.when` on the returned objects.


jbeezley:
PTAL, rebased on 2.0-integration

Issue #602, opened 2015-01-07
Girder output doesn't use standard logging facilities


ronichoudhury:
While figuring out just how to do logging for Tangelo, I discovered that, while CherryPy's "error" log (all log statements generated by calls to ``cherrypy.log()``) is by default written to stderr, but that some of Girder's output goes to stdout instead.  I came upon this when I saw that reports about Girder connecting to Mongo, etc., appear on stdout even as the rest of Tangelo's logs go to stderr.

The reason for this is that Girder, at least in that instance, uses ``print`` statements to make this report.  Should this be part of the logging output (either CherryPy's, or Python's) instead?

zachmullen:
Probably. As long as our messages still appear in the console I am fine with changing this; having the colorful errors and warnings are kind of handy for development.

ronichoudhury:
Colors are great.  In fact I poached the code to make rainbow logs into Tangelo :metal:

manthey:
Added in PR #1505.

PR #1505, opened 2016-08-16
Instead of just printing to stdout, also send information to logs.

manthey:
Except for the install utilities, all print statements now go through the `logprint` function.  This still prints the value to stdout with the appropriate terminal color.  It also send the information to the configure logs (if any).  This simplifies printing in color, as instead of `print(TerminalColor.info(msg))`, one can use `logprint.info(msg)`.  It also adds exception printing support and a few additional log levels.

This resolves issue #602.

### DIFF
--- a/girder/__init__.py
+++ b/girder/__init__.py
@@ -123,11 +123,13 @@ logger = _setupLogger()
 def logprint(*args, **kwargs):
     """
     Send a message to both stdout and the appropriate logs.  This behaves like
-    Python3's print statement, plus takes additional named parameters of
-    level (log level), color (one of the constants.TerminalColor values), and
-    exc_info (either True for the last exception or exception information).
+    Python3's print statement, plus takes additional named parameters:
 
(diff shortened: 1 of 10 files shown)


girder/__init__.py
```diff
@@ -113,5 +119,46 @@ def _setupLogger():
...
+    level (log level), color (one of the constants.TerminalColor values), and
+    exc_info (either True for the last exception or exception information).
+
+    Ideally, we should expose a config option to make this silent.
+    """
+    data = six.StringIO()
+    kwargs = (kwargs or {}).copy()
+    level = kwargs.pop('level', logging.DEBUG)
+    color = kwargs.pop('color', None)
+    exc_info = kwargs.pop('exc_info', None)
```
zachmullen: Let's document these kwargs in the docstring.

manthey:
@zachmullen This is ready for review.

zachmullen:
LGTM

PR #1506, opened 2016-08-16
Add an endpoint to lookup the path of a resource

jbeezley:
This is roughly the inverse of `GET /resource/lookup` and is helpful for client codes that treat girder objects as file-like paths.  It is valid for collections, users, folders, items, and files.

A couple of sticky points here:
* A name containing a slash (`/`) will not be escaped and will break lookups in the other direction.  If desired, I can add some logic to handle that case for both endpoints.
* It is possible (I believe) for a folder to be readable that has a parent folder that is not readable.  In this case, the code here will respond with a `403` if the user lacks read permissions on any document in the parent hierarchy because that would leak private information (e.g. the document's name).

### DIFF
--- a/docs/api-docs.rst
+++ b/docs/api-docs.rst
@@ -194,6 +194,9 @@ Utility
 .. automodule:: girder.utility.progress
    :members:
 
+.. automodule:: girder.utility.path
+   :members:
+
 Constants
 ---------
 .. automodule:: girder.constants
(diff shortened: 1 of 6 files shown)


fogleman:
> A name containing a slash (/) will not be escaped and will break lookups in the other direction. If desired, I can add some logic to handle that case for both endpoints.

Yes please. That caused me headaches with Midas -> Girder migration.

jbeezley:
For escaping, how about this:  `\/` -> `/` and `\\` -> `\` so we have:

`/collection/bad\/collection\\name/folder` -> `['bad/collection\name', 'folder']`

Any better ideas?

fogleman:
Works for me.

jbeezley:
Okay, done.  PTAL

zachmullen:
LGTM

PR #1507, opened 2016-08-16
Add include/exclude regexes for assetstore import process

zachmullen:


girder/api/v1/assetstore.py
```diff
@@ -150,6 +150,10 @@ def createAssetstore(self, params):
                enum=('folder', 'collection', 'user'))
         .param('progress', 'Whether to record progress on the import.',
                dataType='boolean', default=False, required=False)
+        .param('fileIncludeRegex', 'If set, only filenames matching this regular '
+               'expression will be imported.', required=False)
+        .param('fileExcludeRegex', 'If set, only filenames that do not match this regular '
+               'expression will be imported.', required=False)
```
jeffbaumes: May be good to clarify with "If a filename matches both inclusion and exclusion patterns, exclusion takes precedence over inclusion." in the docs here.

zachmullen:
@manthey PTAL, I noticed we needed this feature in our histomicsTK provisioning scripts in order to exclude the .gitignore file from the import.

tests/cases/assetstore_test.py
```diff
@@ -193,12 +188,26 @@ def testFilesystemAssetstoreImport(self):
...
+        # Test importing directory with include & exclude regexes; file should be excluded
+        params['importPath'] = os.path.join(ROOT_DIR, 'tests', 'cases', 'py_client')
+        self.request(
+            path, method='POST', params=dict(
+                params, fileIncludeRegex='world.*', fileExcludeRegex='world.*'))
+        resp = self.request('/resource/lookup', user=self.admin, params={
+            'path': '/user/admin/Public/world.txt/world.txt',
+            'test': True
+        })
+        self.assertStatusOk(resp)
```
manthey: The coverage report implies that this call isn't doing what we expect (the line where the exclusion filter occurs is reported as uncovered).

tests/cases/assetstore_test.py
```diff
@@ -193,12 +188,26 @@ def testFilesystemAssetstoreImport(self):
...
-        params['importPath'] = os.path.join(
-            ROOT_DIR, 'tests', 'cases', 'py_client')
-        resp = self.request(path, method='POST', params=params, user=self.admin)
+        self.model('file').remove(file)
+
+        # Test importing directory with include & exclude regexes; file should be excluded
+        params['importPath'] = os.path.join(ROOT_DIR, 'tests', 'cases', 'py_client')
+        self.request(
+            path, method='POST', params=dict(
+                params, fileIncludeRegex='world.*', fileExcludeRegex='world.*'))
```
manthey: This is missing `user=self.admin` as a parameter to the POST call.

zachmullen: Good catch, I forgot to assert that this request returned OK.

manthey:
LGTM

PR #1513, opened 2016-08-18
WIP: Create an Ansible role for Girder

brianhelba:


kotfic:
@brianhelba Looking at this the next day i am having some concerns about bloat in the role,  particularly supporting mongo and proxies.  These feel like things that belong in separate roles (that either we maintain,  or that are brought in from third parties).  

My sense from yesterday was that your primary concern with this approach was issues with packaging multiple role together (e.g. problems with galaxy)? I'm wondering if there is a separate technical approach we can take that would alleviate those concerns (or did I misunderstand?).

I'd like to see something like the following:
- Fail if mongo-db connection information is not provided (this role should not be responsible for setting up mongo) 
- Check if the min version of each of system prereqs are available:
  - take OS specific action to install required package if possible
  - Fail if not installed
- Clone girder
- Install girder & dependencies (e.g. with pip, in virtualenv)
  - Install default plugin dependencies (pip)
- Provide OS specific services tied to a variable (e.g. ```girder_daemon: supervisord```)
  - if enabled these setup and run default services
- Provide handlers to restart & rebuild girder

The rest of the components necessary to coordinate other services (e.g. mongo,  proxy servers, etc should be handled in project specific playbooks). We should have a number of example playbooks that people can start from but these should ultimately be copied/modified by individual projects.   

I know there are several other people who have a pressing stake in this role coming together so i'm curious what @cjh1 @danlamanna and @mgrauer think?

cjh1:
IMHO The the role should limit itself to Girder specific operations. I agree with @kotfic setting up mongodb is outside the scope of this role. Everyone is going to have their own requirements for a mongo install, its probably going to be on another machine, maybe a hosted service, sharding etc. etc. We don't want to try to support this ... I say keep it simple even if it requires the user to have more in their playbook. As @kotfic says there is nothing to stop use providing examples.

danlamanna:
I agree with most of what's been said by @kotfic and @cjh1. I would rather Mongo/Nginx installation be separate from the Girder role. Let the marrying of Nginx and Girder take place in a site-specific playbook.

mgrauer:
Agree with these comments to keep it just to Girder.  Anything else assumes too much.

I've had deployments where we were given a Mongo instance already, and needed to start up a Girder only instance.  Similarly, we need to be able to use nginx or apache or whatever, but that is outside of Girder.

Restricting the scope will also make it easier to appease all the disparate needs of devs who want to use this in various ways

jeffbaumes:
FYI I'm having @danlamanna look at cleaning up the Girder role on this PR since he has some time and there seems to be general consensus on where to take the role. It's going to be a stripped down version of the current role - we should also look at how to also have some examples on how to deploy Girder with mongo, nginx, apache, etc.

kotfic:
@danlamanna Keep in mind that the ```girder.py``` module will need to be updated before this can be merged. There are PR's that are going in to girder that update the library (e.g. https://github.com/girder/girder/pull/1513)

zachmullen:
@kotfic @brianhelba @danlamanna is this PR now obsolete?

kotfic:
@zachmullen yes

PR #1515, opened 2016-08-18
Skip replicaset tests when mongod is not present

### LINKED ISSUE #1216: Certain tests fail if mongod isn't present
Tests dependent on mongo replica sets (`server_upload` and `web_client_data_gridfsrs`) require the `mongod` binary exist on the local machine, while the rest of Girder functions with either a local or remote mongo uri.

I think we should use [REQUIRED_FILES](https://cmake.org/cmake/help/v3.4/prop_test/REQUIRED_FILES.html) so the tests will register as "not run" rather than failing.

jbeezley:
Fixes #1216


zachmullen:
LGTM

Issue #927, opened 2015-08-20
track down assets from an assetstore


mgrauer:
I want to delete an S3 assetstore, but there are files stored in it and I can't delete it. I'd like some easier way of tracking down these files so that I can remove the assetstore.

zachmullen:
:+1:

manthey:
I think the simplest way is to add an endpoint for GET /assetstore/{id}/files .

manthey:
Added in PR #1516.

PR #1516, opened 2016-08-18
Add an endpoint to list files owned by an assetstore.

manthey:
This resolves issue #927.

### DIFF
--- a/girder/api/v1/assetstore.py
+++ b/girder/api/v1/assetstore.py
@@ -272,7 +272,7 @@ class Assetstore(Resource):
         .errorResponse('You are not an administrator.', 403)
     )
     def getAssetstoreFiles(self, assetstore, params):
-        limit, offset, sort = self.getPagingParameters(params, 'lowerName')
+        limit, offset, sort = self.getPagingParameters(params, '_id')
         return list(self.model('file').find(
             query={'assetstoreId': assetstore['_id']},
             offset=offset, limit=limit, sort=sort))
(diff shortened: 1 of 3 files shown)


girder/api/v1/assetstore.py
```diff
@@ -260,3 +261,18 @@ def updateAssetstore(self, assetstore, params):
...
+    @loadmodel(model='assetstore')
+    @describeRoute(
+        Description('Get a list of files controlled by an assetstore.')
+        .param('id', 'The assetstore ID.', paramType='path')
+        .pagingParams(defaultSort='_id')
+        .errorResponse()
+        .errorResponse('You are not an administrator.', 403)
+    )
+    def getAssetstoreFiles(self, assetstore, params):
+        limit, offset, sort = self.getPagingParameters(params, 'lowerName')
```
mgrauer: why is there a different sort order within the method `'lowerName'` compared to the `@describeRoute`, `'_id'`?

manthey: Copy and paste stupidity...  I meant for it to be by _id.

mgrauer:
@manthey 

Could I prevail upon you to sneak in a similar fix in the [folder find endpoint](https://github.com/girder/girder/blob/master/girder/api/v1/folder.py#L62)?

It should be `'lowerName'`.

mgrauer:
Thanks for the fix @manthey.  LGTM.

You squashed this bug 2 days shy of its first birthday.

🍰

PR #1518, opened 2016-08-18
Fixes #1509. Support github oauth users with no name listed

### LINKED ISSUE #1509: Log in to hub.yt through Github fails
When trying to log into hub.yt using my github account I get the following message:

{
    "message": "Invalid CSRF token (state=\"z7XYtLC6b63gUAThc0FcaEniJzqxagxsxA56BSPi9dLdqRIQmnCLKooft0fZc1tj.https://girder.hub.yt/#\").",
    "type": "rest"
}

{
    "message": "AttributeError: AttributeError(\"'NoneType' object has no attribute 'split'\",)",
    "type": "internal"
}

How can I fix this or is this some bug?

Thanks,

zachmullen:
ping @CamiloHPenaloza @Xarthisius

### DIFF
--- a/plugins/oauth/server/providers/github.py
+++ b/plugins/oauth/server/providers/github.py
@@ -84,8 +84,7 @@ class GitHub(ProviderBase):
         # Get user's email address
         # In the unlikely case that a user has more than 30 email addresses,
         # this HTTP request might have to be made multiple times with pagination
-        resp = self._getJson(method='GET', url=self._API_EMAILS_URL,
-                             headers=headers)
+        resp = self._getJson(method='GET', url=self._API_EMAILS_URL, headers=headers)
         emails = [
             email.get('email')
             for email in resp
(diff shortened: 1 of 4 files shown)


plugins/oauth/server/providers/github.py
```diff
@@ -98,18 +97,13 @@ def getUser(self, token):
...
-                             headers=headers)
+        resp = self._getJson(method='GET', url=self._API_USER_URL, headers=headers)
         oauthId = resp.get('id')
         if not oauthId:
             raise RestException('GitHub did not return a user ID.', code=502)
 
-        login = resp.get('login', None)
+        login = resp.get('login', '')
+        names = (resp.get('name') or login).split()
+        firstName, lastName = names[0], names[-1]
```
Xarthisius: That's gonna create 'login login', wasn't it better with the previous default for `lastName` to `''`?

zachmullen: Actually girder internally doesn't allow empty first name or last name (maybe this is a bad idea, but that's how it is currently written). So, for lack of any better idea, we set this as the initial value.

Xarthisius: I guess for a consistency sake, the same change should be applied to the other providers. At least Bitbucket, I don't think LinkedIn allows for creating users without full name

zachmullen: Ok, will do for bitbucket also.

cpatrick:
LGTM, though it would be better to break out the line changes into separate commits next time.

PR #1522, opened 2016-08-18
[WIP] Add support to validate using JSON schema

cjh1:


cjh1:
@mgrauer PTAL, this is WIP, this need to work out how to support $ref resolution ...

girder/models/model_base.py
```diff
@@ -192,8 +206,13 @@ def validate(self, doc):
         :param doc: The document to validate before saving to the collection.
         :type doc: dict
         """
-        raise Exception('Must override validate() in %s model.'
-                        % self.__class__.__name__)  # pragma: no cover
+        if self.schema:
+            try:
+                jsonschema.validate(doc,  self.schema)
```
mgrauer: two spaces here

requirements.txt
```diff
@@ -11,6 +11,7 @@ python-dateutil==2.5.3
 pytz==2016.4
 requests==2.10.0
 six==1.10.0
+jsonschema==2.5.1
```
mgrauer: I think you would need to add this (modified in some way to make it less specific) to the setup.py also

cjh1: Good catch.

girder/models/model_base.py
```diff
@@ -61,8 +63,20 @@ def __init__(self):
...
+    def _validateSchema(self):
+        """
+        Validate schema if one has been provided.
+        """
+        if self.schema:
+            try:
+                validator = jsonschema.validators.validator_for(self.schema)
+                validator.check_schema(self.schema)
+            except jsonschema.exceptions.SchemaError as se:
+                raise Exception('Invalid JSON schema: %s' % se.message)
```
mgrauer: Maybe you could add a bit of debugging here, like the name of the model at least.

I've had a hard time hunting down these json schema validation errors in the past.

cjh1: Sure, good idea

mgrauer:
@cjh1 I like your approach here for adding schemas to models and validating them.

I'm having slight concerns about the whole idea and wanted to at least describe them.  We are storing models in Mongo, which is schemaless, and this will start introducing schemas to models, which could mean issues like data migration (e.g., update saved models when a schema changes).  We would also want to ensure that developers add schemas that don't restrict the ability to add new fields, which is a feature Girder relies upon to allow plugins to extend core models.

Of course, this is written as an opt-in, so you aren't forcing anyone to adopt schemas for their models.

girder/models/model_base.py
```diff
@@ -192,8 +206,13 @@ def validate(self, doc):
...
         """
-        raise Exception('Must override validate() in %s model.'
-                        % self.__class__.__name__)  # pragma: no cover
+        if self.schema:
+            try:
+                jsonschema.validate(doc, self.schema)
+            except jsonschema.ValidationError as ve:
+                raise ValidationException(ve.message)
+
+        return doc
```
cjh1: One issue here is this removes the requirement to override the validate method ...

zachmullen: That's no problem.

cjh1:
@mgrauer Not sure how useful the latest addition is, but was experimenting with validation of incoming requests.

cjh1:
Another possibility is to add another decorator to do validation of parameters on an endpoint. Something like:

```
@validateJson(schema=..., param='name')
```

Any value? Thoughts?

zachmullen:
That would be great for handling requests that accept JSON in the body of the request.

cjh1:
For the request body it would be nice to have a decorator, that injects the result of calling getBodyJson(...), something like:

```
@jsonBody(schema=...)
def myEndpoint(body, ...)
```

zachmullen:
That is cool, though I fear we may be starting to have too many decorators on our REST endpoints. Let's hold off on that for now and evaluate later on whether we want to make it into a decorator.

zachmullen:
@cjh1 PTAL at #1701 

I think this feature would best be implemented within the newly proposed `autoDescribeRoute`; there is a new method on the description called `jsonParam`, and I envision (but have not yet implemented) a `jsonSchema` kwarg to that which could validate both body JSON and JSON in form or query params. What do you think?

cjh1:
@zachmullen Yes, I saw #1701 and though of this branch. I think when I started the schema work we discussed using a decorator, but there where concerns about there proliferation, as we are adding this one I think its a natural fit for this functionality. I would be happy to add this on a follow up PR to #1701.

zachmullen:
We now have json schema support in autoDescribeRoute, so I'm closing this. If further action is needed, go ahead and open a PR against latest master.

PR #1527, opened 2016-08-18
WIP: Updating vagrant VM for development

chetnieter:
Updated ansible playbook so vagrant VM can be used for development.
Now installing cmake and ccmake, installing python modules from
requirements-dev.txt and installing the npm modules needed for
development.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -28,6 +28,6 @@ Vagrant.configure("2") do |config|
 
   config.vm.provider "virtualbox" do |virtualbox|
     virtualbox.name = "girder"
-    virtualbox.memory = 768
+    virtualbox.memory = 1536
   end
 end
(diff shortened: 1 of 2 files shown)


chetnieter:
This fixes the various missing python package errors I was getting when running the tests and insures that the needed tools and packages are installed. I am still getting quite a few test failures and the mongo database seems pretty fragile and occasionally crashes while the tests are running.

brianhelba:
LGTM, just bump the Vagrant memory size for VirtualBox to 1.5GB first.

chetnieter:
Increasing the memory made a big difference  but it looks like most/all of the web_client tests still fail with:

`/home/vagrant/girder/node_modules/phantomjs/lib/phantom/bin/phantomjs: 4: /home/vagrant/girder/node_modules/phantomjs/lib/phantom/bin/phantomjs: Syntax error: Unterminated quoted string`

mgrauer:
These should be fixed by work already merged into master.  If there is anything not addressed, please open a new issue.

PR #1528, opened 2016-08-18
Update grunt-contrib-symlink and add the force option

### LINKED ISSUE #1264: Grunt fails with NodeJS 5.x, due to symlink task
To fix this, the ``force: true`` argument needs to be added to [the ``symlink.options`` object](https://github.com/girder/girder/blob/18796e2353b6f41da73d235c391ac8f36fdc2cd6/grunt_tasks/build.js#L175) in the Grunt configuration.

However, this option depends on the a new version of ``grunt-contrib-symlink`` being released by upstream. Once https://github.com/gruntjs/grunt-contrib-symlink/issues/12 and https://github.com/gruntjs/grunt-contrib-symlink/issues/21 are resolved, we can upgrade the version used by Girder.

As a last resort, we can always fork ``grunt-contrib-symlink`` and re-publish it ourselves, but I don't think this "nuclear option" should be necessary; the upstream maintainers appear to still be active.

jbeezley:
Fixes #1264

### DIFF
--- a/grunt_tasks/build.js
+++ b/grunt_tasks/build.js
@@ -180,7 +180,8 @@ module.exports = function (grunt) {
 
         symlink: {
             options: {
-                overwrite: true
+                overwrite: true,
+                force: true
             },
             legacy_names: {
                 // Provide static files under old names, for compatibility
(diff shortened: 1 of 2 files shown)


cpatrick:
LGTM.

cpatrick:
...WTIH.

PR #1529, opened 2016-08-18
WIP: Dicom plugin

cpatrick:
This is a pretty stale topic that is worth reviving, I think.


plugins/dicom/utils/record.py
```diff
@@ -0,0 +1,251 @@
...
+import sys
+import traceback
+import json
+import dicom
+import girder_client
+import argparse
+
+from girder_client import HttpError
+
+sys.path.append('../server')
```
zachmullen: Naw

zachmullen:
Not sure what the status is on this, closing it due to staleness. If this is an ongoing effort we can re-open.

Issue #330, opened 2014-09-17
Ability to move files from one assetstore to another


zachmullen:
This should be implemented such that it calls the existing download function and pipes the data (in chunks) into the upload function. When done, it would update the file's assetstoreId and delete it from the old assetstore.

manthey:
Resolved by PR #1531.

PR #1531, opened 2016-08-18
Move files between assetstores.

manthey:
Move files between assetstores.

Allow uploads to target a specific assetstore.

Fixed a bug in GridFS assetstore where, if enough data is in the assetstore, downloads could fail because there was no index just on n.

This still needs tests.

This resolves issue #330.


girder/utility/gridfs_assetstore_adapter.py
```diff
@@ -65,6 +65,7 @@ def validateInfo(doc):
                 ('uuid', pymongo.ASCENDING),
                 ('n', pymongo.ASCENDING)
             ], unique=True)
+            chunkColl.create_index([('n', pymongo.ASCENDING)])
```
zachmullen: Why is this index required? Don't we always want to always filter by uuid?

manthey: In using Mongo version 3.2.9, I was getting an error when trying to download a large file that the data could not fit in memory and that an index was needed.  The lines
```
 cursor = self.chunkColl.find({
            'uuid': file['chunkUuid'],
            'n': {'$gte': n}
        }, projection=['data']).sort('n', pymongo.ASCENDING)
```
were not using the compound index.  Adding this index solves that.

zachmullen: Hm, that's odd. Any idea why that query wouldn't use the compound index? I would expect it to do so...

manthey: I don't know.  I will look into it further.

manthey: It turns out that somehow I had dropped the indices from the GridFS database.  I then uploaded a large file to it, and it wouldn't download due to the lack of any index on `n`.  The addition of the `n`-only index was a red herring.  We really just need to ensure the index exists when we attach to the database (in the `__init__` method), though we should still do so in the `validate` method as well.

manthey:
This is ready for review.

girder/utility/gridfs_assetstore_adapter.py
```diff
@@ -85,6 +85,10 @@ def __init__(self, assetstore):
                 self.assetstore.get('mongohost', None),
                 self.assetstore.get('replicaset', None)
             )[self.assetstore['db']].chunk
+            self.chunkColl.create_index([
+                ('uuid', pymongo.ASCENDING),
+                ('n', pymongo.ASCENDING)
+            ], unique=True)
```
zachmullen: If we want to call this logic in two places, let's make a small helper method.

zachmullen:
Could you merge master into this for CI

zachmullen:
LGTM

Issue #1498, opened 2016-08-09
Disabled buttons are not actually disabled


zachmullen:
We are adding the disabled class to make them look disabled, but not the disabled property to actually disable them.

zachmullen:
I plan to fix this in 2.0-integration since it will touch many files. I'm going to add another jquery plugin that will toggle both the class and the attribute on a selector.

PR #1532, opened 2016-08-18
Add girderEnable method to factorize enabling/disabling inputs

vovythevov:
See Issue #1498

### DIFF
--- a/clients/web/src/utilities/jquery/girderEnable.js
+++ b/clients/web/src/utilities/jquery/girderEnable.js
@@ -1,0 +1,25 @@
+import jQuery from 'jquery';
+
+/**
+ * Define jQuery plugins within this scope.
+ */
+(function ($) {
+    /**
+     * Helper selector to enable/disable inputs elements
+     *
+     * @param enable Whether to enable the element or not
+     */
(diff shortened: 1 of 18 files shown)


clients/web/src/utilities/jquery/girderEnable.js
```diff
@@ -0,0 +1,26 @@
...
+ * Define jQuery plugins within this scope.
+ */
+(function ($) {
+    /**
+     * Helper selector to enable/disable inputs elements
+     *
+     * @param enable Whether to enable the element or not
+     */
+    $.fn.girderEnable = function (enable) {
+        if($(this).is(':input')){
```
msmolens: Would it make sense to cache $(this) in a variable?

vovythevov: Good catch. I'll change that.

zachmullen:
@vovythevov could you re-open this PR against `2.0-integration`? Thanks!

zachmullen:
Actually don't worry about this since you are on vacation; I'll cherry-pick it onto my current branch and close this out. Thanks!

PR #1536, opened 2016-08-19
Add another mode for adding npm deps to fix Vega Gruntfile error

jeffbaumes:
Separates out an addDependencies() function for adding npm dependencies to Girder root from plugins. Now plugins can specify npm dependencies from either `grunt.dependencies` (meant for deps needed for the plugin build process) or `npm.dependencies` (meant for deps needed by "in-place build" plugins such as the core plugins). The latter is needed by the Vega plugin (which does not have a grunt build) and hadn't been supported until now - a warning about not having a Gruntfile occurs when trying to use `grunt.dependencies`.

`grunt.dependencies` could be deprecated in favor of `npm.dependencies` in the future, but this fix is backwards-compatible.

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -107,14 +107,10 @@ module.exports = function (grunt) {
                 configurePlugin(plugin);
             }
 
-            if (config.grunt) {
-                grunt.log.writeln((
-                    'Found plugin: ' + plugin + ' (custom Gruntfile)'
-                ).bold);
-
+            function addDependencies (deps) {
                 // install any additional npm packages during init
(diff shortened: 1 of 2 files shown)


jeffbaumes:
@sebastienbarre @jbeezley PTAL.

jbeezley:
We should probably update the documentation here: https://github.com/girder/girder/blob/master/docs/plugin-development.rst#installing-custom-dependencies-from-npm

jeffbaumes:
Good idea. I think the docs don't delineate enough between plugins that are built in-place and plugins that have their own build system. I can work to clarify.

Issue #1537, opened 2016-08-19
girder-client: GirderClient._upload_file_to_item does not use local_file argument


thewtex:
The name of the local_file should be used in uploadFileToItem

thewtex:
A new argument added to uploadFileToItem, `localFile`, and change

```
filename = os.path.basename(filepath)
```
to
```
filename = os.path.basename(localFile)
```

PR #1538, opened 2016-08-19
Python client: Support different filename on local filesystem

thewtex:
girder-client: GirderClient._upload_file_to_item does not use local_file
argument #1537 

This only came about when developing a script to archive CMake ExternalData; it has only been tested there. I did not find any testing for the Python girder client. *This should be tested locally before merging.*

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -615,7 +615,7 @@ class GirderClient(object):
         # to upload anyway in this case also.
         return (None, False)
 
-    def uploadFileToItem(self, itemId, filepath, reference=None, mimeType=None):
+    def uploadFileToItem(self, itemId, filepath, reference=None, mimeType=None, filename=None):
         """
         Uploads a file to an item, in chunks.
         If ((the file already exists in the item with the same name and size)
@@ -627,9 +627,12 @@ class GirderClient(object):
         :type reference: str
         :param mimeType: MIME type for the file. Will be guessed if not passed.
         :type mimeType: str or None
+        :param filename: path with filename used in Girder. Defaults to basename of filepath.
         :returns: the file that was created.
         """
-        filename = os.path.basename(filepath)
+        if filename is None:
+            filename = filepath
+        filename = os.path.basename(filename)
         filepath = os.path.abspath(filepath)
         filesize = os.path.getsize(filepath)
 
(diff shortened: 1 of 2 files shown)


fogleman:
Python client tests are here: https://github.com/girder/girder/blob/master/tests/cases/py_client/lib_test.py

clients/python/girder_client/__init__.py
```diff
@@ -650,9 +651,13 @@ def uploadFileToItem(self, itemId, filepath, reference=None, mimeType=None):
         :type reference: str
         :param mimeType: MIME type for the file. Will be guessed if not passed.
         :type mimeType: str or None
+        :param localFile: path with filename used in Girder. Defaults to
+        filepath.
```
zachmullen: Line 641 (which is breaking the style test) and 655 (which sphinx will complain about) can both be moved to the preceding line since we now allow 100 characters per line. 👍

thewtex:
Thanks for the reviews and pointers to the tests.  I will follow-up on this hopefully later this week.

zachmullen:
Hi @thewtex , could you go change this PR to be against the 2.0-integration branch?

thewtex:
@zachmullen yes, rebased against 2.0-integration.

thewtex:
Green :-)

tests/cases/py_client/lib_test.py
```diff
@@ -616,3 +616,17 @@ def testUploadWithPath(self):
...
+        itemName = 'MyStash'
+        item = self.client.createItem(self.publicFolder['_id'],
+                                      itemName)
+
+        path = os.path.join(self.libTestDir, 'sub1', 'f1')
+        uploadedFile = self.client.uploadFileToItem(item['_id'],
+                                                    path,
+                                                    reference=None,
+                                                    mimeType=None,
+                                                    localFile='g1')
```
zachmullen: We tend to prefer next-line indented instead of this visual indentation style. Also, we can avoid passing `reference` and `mimeType` arguments here since they are the defaults.

thewtex: A style test was complaining about next-line indentation.

I'll try with all on the same line.

zachmullen: It just needs to be 4-spaces to the right of the previous line's indentation, our pep8 tool allows for that.

zachmullen:
Looks fine, but I think we should come up with a different name besides localFile, because that makes me think it's referring to how it appears locally rather than on the server. Maybe call it just `filename` and explain the docstring it's how it will appear in Girder?

thewtex:
OK, will change `localFile` to `filename`.

thewtex:
Thanks!

zachmullen:
No problem. Does this need to be in a release for your ITK stuff, or is it OK just being in master?

thewtex:
It will be helpful if it gets in a release by December.

PR #1542, opened 2016-08-22
Add docs for new npm.dependencies config for plugins

jeffbaumes:
@jbeezley PTAL. I also noticed a main heading that I believe should have been a subheading.


jbeezley:
LGTM

jeffbaumes:
Thanks. Merging docs into webpack since this is where this parameter was introduced.

PR #1543, opened 2016-08-22
Add Globus OAuth provider

Xarthisius:
This PR adds support for [Globus](https://www.globus.org/) as an OAuth provider, which is quite popular in Academia.

### DIFF
--- a/plugins/oauth/server/providers/globus.py
+++ b/plugins/oauth/server/providers/globus.py
@@ -30,32 +30,6 @@ from .base import ProviderBase
 from .. import constants
 
 
-def decode_value(val):
-    # Borrowed from 'cryptography' for py2 backward compat
-    if hasattr(int, 'from_bytes'):
-        int_from_bytes = int.from_bytes
-    else:
-        def int_from_bytes(data, byteorder, signed=False):
-            assert byteorder == 'big'
-            assert not signed
-
-            if len(data) % 4 != 0:
-                data = (b'\x00' * (4 - (len(data) % 4))) + data
-
-            result = 0
-
-            while len(data) > 0:
-                digit, = struct.unpack('>I', data[:4])
-                result = (result << 32) + digit
-                data = data[4:]
-            return result
-
-    if isinstance(val, string_types):
-        val = val.encode('utf-8')
-    decoded = base64url_decode(val)
-    return int_from_bytes(decoded, 'big')
-
-
 class Globus(ProviderBase):
     _AUTH_URL = 'https://auth.globus.org/v2/oauth2/authorize'
     _AUTH_SCOPES = ('urn:globus:auth:scope:auth.globus.org:view_identities',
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -77,8 +77,10 @@ def run(self, *arg, **kw):
     'geospatial': ['geojson'],
     'thumbnails': ['Pillow', 'pydicom', 'numpy'],
     'worker': ['celery'],
-    'plugins': ['celery', 'geojson', 'Pillow', 'pydicom', 'numpy']
+    'oauth': ['pyjwt', 'cryptography']
 }
+all_extra_reqs = [y for x in extras_reqs.values() for y in x]
```
zachmullen: I find this construct difficult to read, would you mind changing to:

    all_extra_reqs = itertools.chain.from_iterable(extras_reqs.values())

zachmullen:
We should also add a `requirements.txt` to the oauth plugin with specific versions of the dependencies listed.

plugins/oauth/server/providers/globus.py
```diff
@@ -0,0 +1,149 @@
...
+from cryptography.hazmat.backends import default_backend
+from six.moves import urllib
+from six import string_types
+
+from girder.api.rest import getApiUrl, RestException
+from .base import ProviderBase
+from .. import constants
+
+
+def decode_value(val):
```
zachmullen: Since this is a public symbol, let's make it camelCase and also add a docstring. (Or just make it private if you prefer.)

plugins/oauth/server/providers/globus.py
```diff
@@ -0,0 +1,149 @@
...
+        if not oauthId:
+            raise RestException(
+                'Globus identity did not return a valid ID.', code=502)
+
+        email = identity.get('email')
+        if not email:
+            raise RestException(
+                'Globus identity did not return a valid email.', code=502)
+
+        name = identity.get('name').split()
```
zachmullen: This will only work if `identity` actually contains a `'name'` key. Otherwise, we are attempting to split `None`, which will throw a confusing error message. We should either

* Change to use `identity['name'].split()` if we are sure we will be given a name.
* Change to `identity.get('name', some_default_string).split()`
* If there is no name field or an empty name field, we should raise a more useful exception message.

plugins/oauth/web_client/js/ConfigView.js
```diff
@@ -29,6 +29,14 @@ girder.views.oauth_ConfigView = girder.View.extend({
                           'Developer Console. When creating your client ID there, ' +
                           'use the following values:'
         }, {
+            id: 'globus',
+            name: 'Globus',
+            icon: 'gplus',
```
zachmullen: `icon: 'globe',`

zachmullen:
LGTM other than the icon

zachmullen:
Thanks @Xarthisius,  you rock!

PR #1544, opened 2016-08-24
Add a periodic event trigger. Closes #1533

### LINKED ISSUE #1533: [RFC] Periodic events
I'd like to suggest creation of periodic events that would be triggered e.g. every hour, 24h etc. I have at least two applications that would benefit from this:

1. short lived objects with an expiration attribute (in my case these are notebook containers)
2. embargoed data, that cannot be made public until specific date

Proof of concept:

```
diff --git a/girder/__main__.py b/girder/__main__.py
index 415cbae..8b8743c 100644
--- a/girder/__main__.py
+++ b/girder/__main__.py
@@ -20,9 +20,11 @@
 import cherrypy  # pragma: no cover
 import argparse  # pragma: no cover
 import os  # pragma: no cover
+from cherrypy.process.plugins import BackgroundTask  # pragma: no cover
 
 try:  # pragma: no cover
     from girder.utility import server
+    from girder.events import trigger
 except ImportError:
     # Update python path to ensure server respawning works. See #732
     source_root_dir = os.path.dirname(os.path.dirname(__file__))
@@ -31,6 +33,7 @@ except ImportError:
                         % source_root_dir)
     sys.path.append(source_root_dir)
     from girder.utility import server
+    from girder.events import trigger
 
 
 def main():
@@ -49,7 +52,12 @@ def main():
         cherrypy.config['server.socket_port'] = int(args.port)
     server.setup(args.testing)
 
+    def hourly_heartbeat():
+        trigger('heartbeat.hourly')
+    cron = BackgroundTask(3600, hourly_heartbeat)
+
     cherrypy.engine.start()
+    cron.start()
     cherrypy.engine.block()
```

Xarthisius:
This PR adds a simple `Monitor` thread to the main cherrypy bus that triggers _heartbeat_ event every `server.heartbeat` seconds.


zachmullen:
This looks good, we should document it, including its caveats. Like, does this work in WSGI wrappers besides cherrypy? We should also document that it will execute on every instance of the app server, so in a multi-node server deployment, people must assume this could be running in parallel on many instances at once, and a different solution should be employed if that's problematic.

mgrauer:
This needs to be thought of in a broader context, running horizontally scaled or in other deployment setups.  We'll close this and keep the issue open for discussion.

PR #1545, opened 2016-08-25
Ansible client testing fixes

danlamanna:
@kotfic 
- `vagrant_destroy` test now properly depends on every other test
- pulls ansible galaxy role from the correct branch
- force ansible client tests to run serially (since they share the same database)


kotfic:
@danlamanna Thanks for catching these,  LGTM

PR #1547, opened 2016-08-26
Move some Plugin doc sections to a seemingly more appropriate location

Purg:
Not sure if this is supposed to be the case, but it looked like some client-side javascript extension documentation was mis-placed under the section about linting and style checking JS code. This branch moves the sub-sections out from under the linting sub-section and under the Client-side extension sub-section.

### DIFF
--- a/docs/plugin-development.rst
+++ b/docs/plugin-development.rst
@@ -450,80 +450,8 @@ that can be used to import content:
   plugin that doesn't fall into one of the above categories can be placed here,
   such as static images, fonts, or third-party static libraries.
 
-JavaScript extension capabilities
-*********************************
-
-Plugins may bind to any of the normal events triggered by core via the
-``girder.events`` object. This will accommodate certain events, such as before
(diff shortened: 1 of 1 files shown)


zachmullen:
I think the actual bug here is that the linting and style checking section is using the wrong underlining character, I think it should be `*` and not `^`.

Purg:
Are what look like that section's children sections ("Installing custom dependencies from npm", "Executing custom Grunt build steps for your plugin", etc.) actually child sections under the "Linting" section? Or should those stay as "*" headings?

zachmullen:
Those should stay as is, they are not subsections.

Purg:
Ok. Commit incoming.

zachmullen:
@Purg please merge latest master into this topic so that our CI will pass. Thanks!

Purg:
Rebased.

zachmullen:
LGTM

PR #1548, opened 2016-08-29
Update grunt-fontello.

manthey:
I was getting failure messages from npm when it ran the fontello task.  This appears to correct it.


zachmullen:
I was seeing this also, it was happening intermittently. I think this is an HTTP failure of some sort and it's trying to check the status code on a response, but there was no response (couldn't connect probably).

zachmullen:
In any case, this fix can't hurt, so LGTM.

PR #1549, opened 2016-08-30
The change to log errors despite unicode issues fails in Python3.

manthey:
This should always log something, though the log will be uglier if it would have a Unicode error.

PR #1503 attempted to fix this.  This is another attempt.


zachmullen:
LGTM

PR #1550, opened 2016-08-30
Retrieve fontello icons from girder

jbeezley:
I gave up trying to get fontello.com working.  Should fix all the deployment issues we've been having.


zachmullen:
LGTM

PR #1551, opened 2016-08-31
Fix failing documentation generation test

zachmullen:
When we merged master into 2.0-integration, this test was
introduced onto this branch for the first time.


zachmullen:
Any objections?

jcfr:
LGTM :+1:

zachmullen:
I had to change one more thing on this topic; I noticed another test was failing occasionally, it turns out we were depending on an undefined dict ordering.

PR #1552, opened 2016-08-31
Move job update events into model layer, add post-update event

zachmullen:


zachmullen:
This moves the job update event out of the REST layer and into the model layer, which is a more appropriate place.

@manthey PTAL.

manthey:
There is still a minor style issue.  Once that is fixed, LGTM.

manthey:
@zachmullen I just realized we actually removed some functionality as part of moving the `jobs.job.update` event from the rest endpoint to the updateJob.  Specifically, an event handler could change the parameters of what was being updated in the job (since it was in the params dictionary), whereas now, it cannot.  Was this desired functionality?

zachmullen:
I don't think anybody was doing that, but if so, they will hopefully speak up. In general, events that are meant to be able to modify inputs like that should use `getResponse` rather than modifying their input parameters for side effects, so anyone who was using it for that purpose would have been relying on undefined behavior.

PR #1553, opened 2016-08-31
Allow alternate tasks to be sent to girder worker

kotfic:
Maintains default ```girder_worker.run task```,  but allows alternate tasks to
be sent if 'celeryTaskName' is set on the job.

Also adds otherFields to createJob to mimik updateJob.

@cjh1 @mgrauer FYI
@zachmullen PTAL


zachmullen:
LGTM

PR #1556, opened 2016-09-01
Use rel=noopener on user-defined links that open in new tab

zachmullen:
@girder/developers PTAL at this security fix. https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

### DIFF
--- a/clients/web/src/templates/widgets/itemList.jade
+++ b/clients/web/src/templates/widgets/itemList.jade
@@ -8,7 +8,7 @@ ul.g-item-list
         | #{item.get('name')}
       a(title="Download item" href=item.downloadUrl())
         i.icon-download
-      a.g-view-inline(title="View in browser" target="_blank"
+      a.g-view-inline(title="View in browser" target="_blank", rel="noopener noreferrer",
                       href=item.downloadUrl({contentDisposition: 'inline'}))
         i.icon-eye
       .g-item-size #{girder.formatSize(item.get('size'))}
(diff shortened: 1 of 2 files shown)


danlamanna:
Do we need to add `noreferrer` (for FF support)?

cpatrick:
yes

zachmullen:
Done

cpatrick:
LGTMWTIH.

clients/web/src/templates/widgets/itemList.jade
```diff
@@ -2,16 +2,16 @@ ul.g-item-list
   each item in items
     li.g-item-list-entry
       if checkboxes
-        input.g-list-checkbox(type="checkbox", g-item-cid="#{item.cid}")
-      a.g-item-list-link.g-right-border(g-item-cid="#{item.cid}")
+        input.g-list-checkbox(type="checkbox", g-item-cid=item.cid)
+      a.g-item-list-link.g-right-border(g-item-cid=item.cid)
         i.icon-doc-text-inv
-        | #{item.get('name')}
```
msmolens: This, and the other unrelated changes, should be part of a separate commit, or even a separate PR.

msmolens:
Add a description of the security problem to the commit message.

danlamanna:
I found a number of other results when grepping for `_blank` - is there a reason those are ignored?

zachmullen:
I was only targeting user-defined data, rather than static links. Are there any of those that I missed?

danlamanna:
@zachmullen Nope :)

jcfr:
Is there a way to provide a jade mixin or similar to encapsulate this ?

zachmullen:
Backed out the style improvements.

PR #1557, opened 2016-09-02
Change girder.dist.cfg to use equals delimiter instead of colon

danlamanna:
If we want programmatic manipulation of our `.cfg` (via ansible or otherwise) then the easiest solution is to use `=` as the delimiter in our `.cfg` files.

More detail:
The built in `ConfigParser` (Python2) reads with either delimiter in mind, but is hardcoded to write out using `=` as the delimiter. This becomes an issue when using the `ini_file` module from Ansible as it duplicates entries. I can put in a PR to fix this in Ansible, but I imagine it will take a long time for it to be incorporated.

### DIFF
--- a/girder/conf/girder.dist.cfg
+++ b/girder/conf/girder.dist.cfg
@@ -1,27 +1,27 @@
 [global]
-server.socket_host: "0.0.0.0"
-server.socket_port: 8080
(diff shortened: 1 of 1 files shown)


zachmullen:
LGTM

PR #1558, opened 2016-09-02
Bespoke job states

cjh1:
Allow the job model to accept job states that are strings.

### DIFF
--- a/plugins/jobs/plugin_tests/jobs_test.py
+++ b/plugins/jobs/plugin_tests/jobs_test.py
@@ -365,6 +365,27 @@ class JobsTestCase(base.TestCase):
             with self.assertRaises(ValidationException):
                 jobModel.updateJob(job, status=4321)  # Should fail
 
+    def testCustomStrStates(self):
+        jobModel = self.model('job', 'jobs')
+        job = jobModel.createJob(title='test', type='x', user=self.users[0])
+
+        def validateStatus(event):
+            states = ['a', 'b', 'c']
+
+            if event.info in states:
+                event.preventDefault().addResponse(True)
+
+        with self.assertRaises(ValidationException):
+            jobModel.updateJob(job, status='a')
+
+        with events.bound('jobs.status.validate', 'test', validateStatus):
+            jobModel.updateJob(job, status='a')
+            self.assertEqual(job['status'], 'a')
+
+        with self.assertRaises(ValidationException), \
+                events.bound('jobs.status.validate', 'test', validateStatus):
+            jobModel.updateJob(job, status='foo')
+    
     def testUpdateOtherFields(self):
         jobModel = self.model('job', 'jobs')
         job = jobModel.createJob(title='test', type='x', user=self.users[0])
--- a/plugins/jobs/plugin_tests/jobs_test.py
+++ b/plugins/jobs/plugin_tests/jobs_test.py
@@ -365,7 +365,7 @@ class JobsTestCase(base.TestCase):
             with self.assertRaises(ValidationException):
                 jobModel.updateJob(job, status=4321)  # Should fail
 
-    def testCustomStrStates(self):
+    def testValidateCustomStrStatus(self):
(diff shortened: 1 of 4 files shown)


cjh1:
@mgrauer PTAL

kotfic:
integer job statuses imply a certain order/relationship,  without wanting to blow up the scope of your PR,  is there a way to implicitly (or better yet, explicitly) define custom job status relationships?

zachmullen:
Actually ordinality of job states was not part of the design; if we want to codify the relationship between states, I think that's a much larger undertaking than this PR, and also not something we could accomplish with integers alone anyway.

kotfic:
Cumulus currently uses something like [this](https://github.com/Kitware/cumulus/blob/master/cumulus/constants.py#L70-L167), which could potentially be generalized. It provides a nice (maybe too magical?) API by which [assigning status to a cluster adapter](https://github.com/Kitware/cumulus/blob/master/girder/cumulus/server/utility/cluster_adapters.py#L128) automatically validates the transition via [properties on the cluster_adapater](https://github.com/Kitware/cumulus/blob/master/girder/cumulus/server/utility/cluster_adapters.py#L44-L55).  This allows status to remain a property of the cluster_adapter,  while ceding all transitions/validation to a configurable state machine.

cjh1:
@kotfic I can see that validating of order/relationships it useful as we have seen in cumulus, but as @zachmullen says its probably out of the scope of this PR. Here we are just trying to removing the requirement to use ints for states, baby steps :smile:

kotfic:
Sure, especially if @zachmullen is saying it wasn't part of the original design then it makes sense to me to look at that in a different PR.

cjh1:
OK, so I fixed up this branch so the tests pass ( I hope, I will add some tests for the new functionality). We now try to coerce the job state into int but will not fail if its not possible. 

There is one inconsistency with the current interface related to this type conversion. If we pass a string such as "3" for the status to the update endpoint then when we get the status back its been converted to an int for us ( whether we like it or not ). I guess that's a reason to use a JSON encoded request body rather than query parameters as in this case the type could be preserved, but I'm not proposing to change this and query parameters seem to be the norm for girder updates.

mgrauer:
@cjh1 

I also assumed that the integer status codes implied an order.

Would you add a comment to the integer job status codes to clarify that no order is implied? The order of the job states comes from the timestamps, as I understand things, this could be useful to mention.

Since there is already a test method called `testValidateCustomStatus`, could you add the body of `testCustomStrStates` to that, with an appropriate string-focused comment?  I do feel conflicted on this, and if you feel strongly, then keep them separate.  My concern is that the test of custom state validation behavior is now split between the two methods.

I'm OK with the inconsistency you mention ("3" => 3), maybe a brief comment is in order on that endpoint explaining the behavior, but it seems unlikely that someone would want to have an int-convertible status string that they want to remain a string.

cjh1:
@mgrauer No, there is no order, they are just an int enum. You want a comment [here](https://github.com/girder/girder/blob/master/plugins/jobs/server/constants.py#L26)

The order of that states is currently just provided by the application code that updates the jobs ( not necessary related to timestamps either ).

Personally, I think each test case should have a narrow focus, that is why I separated the two, it makes it much easier to track down problems in my experience. Plus the state validation behavior is different if you are using ints as the state will be converted to an int. Is it the name that bothers you? I can change them.

mgrauer:
Right, there is no order implied by the ints, and there where you pointed is where I'd like a comment.

I agree with you about the application code that updates jobs sets the order.  I had meant (but didn't state) that the timestamp is used as an ordering for display, in the timeline JS widget and possibly the job detail status JS widget.

You make a good case to keep the test methods separate.  Perhaps a name change would help show that they are quite similar, one just has a particular focus.  Maybe `testValidateCustomStrStatus`?

cjh1:
@mgrauer Done

mgrauer:
Thanks, LGTM.

PR #1559, opened 2016-09-06
Enable database_assetstore in Ansible library module

ronichoudhury:
attn: @danlamanna and @kotfic

### DIFF
--- a/clients/ansible/girder.py
+++ b/clients/ansible/girder.py
@@ -189,7 +189,7 @@ options:
                           - Name of the assetstore
                   type:
                       required: true
-                      choices: ['filesystem', 'gridfs', 's3', 'hdfs']
+                      choices: ['filesystem', 'gridfs', 's3', 'hdfs', 'database']
                       description:
                           - Currently only 'filesystem' has been tested
                   readOnly:
@@ -1638,7 +1638,8 @@ class GirderClientModule(GirderClient):
         "filesystem": 0,
         "girdfs": 1,
         "s3": 2,
-        "hdfs": "hdfs"
+        "hdfs": "hdfs",
+        "database": "database"
     }
 
     def __validate_hdfs_assetstore(self, *args, **kwargs):
@@ -1649,6 +1650,7 @@ class GirderClientModule(GirderClient):
                    replicaset='', bucket=None, prefix='', accessKeyId=None,
                    secret=None, service='s3.amazonaws.com', host=None,
                    port=None, path=None, user=None, webHdfsPort=None,
+                   dbtype=None, dburi=None,
                    readOnly=False, current=False):
 
             # Fail if somehow we have an asset type not in assetstore_types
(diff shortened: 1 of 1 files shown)


zachmullen:
LGTM

zachmullen:
(Though I'd like @kotfic to give approval before this gets merged)

kotfic:
@ronichoudhury If you want to be able to update ```dbtype``` and ```dburi``` attributes of an existing database assetstore  you will need to add them to the [updateable](https://github.com/girder/girder/blob/cff242f3c870e180f404f544762d8cb4e26d858d/clients/ansible/girder.py#L1723-L1726) list.  See the source code comment for why this is unfortunately necessary.  Note also that this does not implement a ```__validate_database_assetstore``` [function](https://github.com/girder/girder/blob/cff242f3c870e180f404f544762d8cb4e26d858d/clients/ansible/girder.py#L1754-L1755). That is fine,  but it means users will be able to pass invalid database types/URIs.  Finally,  please bump the [minor version](https://github.com/girder/girder/blob/cff242f3c870e180f404f544762d8cb4e26d858d/clients/ansible/girder.py#L37).  Thanks!

zachmullen:
Could you also add a comment in there saying that we should remove this option from the assetstore once this makes it into a girder built-in plugin, since the plan is to change this from an assetstore into its own domain object.

manthey:
@zachmullen I'm not at all convinced that databases shouldn't be assetstores.  The ability for a file to be a view in the database is quite useful, and assetstores that aren't fully controlled by Girder will have mutable files, too (such as the filesystem assetstore with imported files).  We need to have a design discussion (but not in this PR).

zachmullen:
> The ability for a file to be a view in the database is quite useful

That certainly wouldn't be precluded by using something other than an assetstore, since files can already be used as redirects, thereby becoming a view of another data source.

> and assetstores that aren't fully controlled by Girder will have mutable files, too

One could mutate files even if they are fully controlled by girder, it's just not a good idea. Mutability in the underlying storage media is only part of the issue; it's really an issue of structured vs. unstructured data.

ronichoudhury:
On Tue, Sep 6, 2016 at 11:31 AM Christopher Kotfila <
notifications@github.com> wrote:

> @ronichoudhury <https://github.com/ronichoudhury> If you want to be able
> to update dbtype and dburi attributes of an existing database assetstore
> you will need to add them to the updateable
> <https://github.com/girder/girder/blob/cff242f3c870e180f404f544762d8cb4e26d858d/clients/ansible/girder.py#L1723-L1726>
> list. See the source code comment for why this is unfortunately necessary.
>
Does this mean I should make these updateable if they could change in the
course of using the assetstore? What's an example use case of an updateable
field? I don't think I want this for dbtype and dburi, as I can't imagine
these two values shifting around for a database assetstore, even if the
assetstore weren't read-only (as currently it is).

> Note also that this does not implement a __validate_database_assetstore
> function
> <https://github.com/girder/girder/blob/cff242f3c870e180f404f544762d8cb4e26d858d/clients/ansible/girder.py#L1754-L1755>.
> That is fine, but it means users will be able to pass invalid database
> types/URIs.
>
I noticed that the only other `__validate_*_assetstore()` function in this
file is a no-op. Where can I see an example of such a function that
actually does something?

> Finally, please bump the minor version
> <https://github.com/girder/girder/blob/cff242f3c870e180f404f544762d8cb4e26d858d/clients/ansible/girder.py#L37>.
> Thanks!
>
Will do.

Thanks @kotfic!

ronichoudhury:
@kotfic and I discussed my questions in meatspace. I've made the requested updates. PTAL!

kotfic:
@ronichoudhury Once travis is happy this LGTM

PR #1561, opened 2016-09-06
Ansible Role

danlamanna:
This PR glues together a lot of the work that has been done with Ansible and Girder over the past year. The majority of the work on the actual role comes from @kotfic and @brianhelba. (It's based off of #1513 but diverges quite a bit, hence the separate PR).

There's been a lot of discussion internally about how a Girder role will function and what it will support. The consensus seems to be that the Girder role will install Girder and only Girder, and anything beyond that must be done in a playbook designed specifically for the site the user is deploying to (we break this a little bit by installing node in our role).

### Examples (seen [here](https://github.com/girder/girder/tree/ansible-role-refactor/devops/ansible/examples))
A large part of this work is having enough examples that people can look through to match their real world needs for deploying Girder. After combing through the repos mentioned recently on girder-devel I assembled a number of examples illustrating common use cases for Girder deployment.

Primarily:
- Deploying Girder under NGINX
- Installing Girder under a virtual environment
- Using the Girder Ansible Client library to configure a running instance of Girder
- Altering the Girder configuration file (for altering server settings, alternative mongo URIs, etc)

### Docs
Some docs were added, see [provisioning.rst](https://github.com/girder/girder/blob/ansible-role-refactor/docs/provisioning.rst).

### To test
Note: The tests have to be run serially (it's possible to change this, but it's not low hanging fruit) and take a while
```
cmake -D ANSIBLE_TESTS=ON -D ANSIBLE_CLIENT_TESTS=ON
ctest -L '^girder_ansible'
```

### To try out
```
GIRDER_EXAMPLE=girder-nginx vagrant up

# this can be done with any of the examples, e.g
GIRDER_EXAMPLE=girder-configure-lib vagrant up
```

### Other notes
- CMake logic has been added/unified to allow our testing of Ansible to be in our main test suite (disabled by default) and labeled to allow for easy testing
- I'm in the middle of setting up nightly builds for this and forcing @jbeezley to do the same so we're covering Linux and Mac hosts
- Testing the examples just aims for a clean vagrant up, though it should probably be ironed out in the future

### Future Work
- Supporting Ubuntu 16.04    
  This role does support 16.04 (@cjh1) (despite Vagrant having many issues with 16.04), however the 3rd party Mongo role we're
      using in our base playbooks doesn't. I've submitted a [PR](https://github.com/Stouts/Stouts.mongodb/pull/28) and hope that will be accepted so we don't have to create our own Mongo role.
- Support different python binaries    
  Right now it just uses the system python, ideally installing/specifying python3 for example would be simple to do.
- Convert Travis to use Ansible    
  This isn't urgent at all, but there's no reason that I can see that would block us from having our travis environment provisioned via ansible.    

~~This is still pretty raw and I've uncovered a few more bugs just today, so it's marked WIP for now but feedback is still appreciated.~~


kotfic:
@danlamanna  https://github.com/girder/girder/pull/1559 has been merged, please update the girder.py module from the one in master. Thanks!

kotfic:
Just an FYI

```
Convert Travis to use Ansible
This isn't urgent at all, but there's no reason that I can see that would block us from having our travis environment provisioned via ansible.
```
This actually is much more challenging that it initially seems. Playbooks would have to be written so that they can work without sudo as travis prevents this. That would include third party roles such as mongo/node which makes it basically impossible

devops/ansible/examples/girder-configure-lib/site.yml
```diff
@@ -0,0 +1,59 @@
...
+        update_cache: yes
+        cache_valid_time: 3600
+      become: yes
+      become_user: root
+
+  roles:
+    - role: Stouts.mongodb
+      become: yes
+      become_user: root
+    - girder
```
kotfic: can we do ```- role: girder``` here to be consistent

devops/ansible/examples/girder-dev-environment/site.yml
```diff
@@ -0,0 +1,44 @@
...
+        update_cache: yes
+        cache_valid_time: 3600
+      become: yes
+      become_user: root
+
+  roles:
+    - role: Stouts.mongodb
+      become: yes
+      become_user: root
+    - girder
```
kotfic: can we do - role: girder here to be consistent

devops/ansible/examples/girder-external-mongo/site.yml
```diff
@@ -0,0 +1,25 @@
...
+  pre_tasks:
+    - name: Update package cache
+      apt:
+        update_cache: yes
+        cache_valid_time: 3600
+      become: yes
+      become_user: root
+
+  roles:
+    - girder
```
kotfic: can we do ```- role: girder``` here to be consistent

devops/ansible/examples/girder-nginx/site.yml
```diff
@@ -0,0 +1,65 @@
...
+        update_cache: yes
+        cache_valid_time: 3600
+      become: yes
+      become_user: root
+
+  roles:
+    - role: Stouts.mongodb
+      become: yes
+      become_user: root
+    - girder
```
kotfic: can we do ```- role: girder``` here to be consistent

devops/ansible/roles/girder/README.md
```diff
@@ -0,0 +1,41 @@
...
+--------------
+
+The following variables may be overridden:
+
+* `girder_path`: Path to download and build Girder in.
+* `girder_version`: Git commit-ish for fetching Girder.
+* `girder_virtualenv`: Path to a Python virtual environment to install Girder in.
+* `girder_update`: Whether provisioning should fetch new versions via git.
+* `girder_force`: Whether provisioning should discard modified files in the working directory.
+* `girder_web`: Whether to build the Girder web client.
```
kotfic: Can we include required/optional (with defaults)  in the documentation here?

kotfic:
It would be nice to fail early if trying to provision a system that isn't suported. What do you think about putting in a ```tasks/required.yml``` that checks OS info and including it at the top of ```tasks/main.yml```?  Could be generally responsible for verifying requirements that the role shouldn't or won't be taking responsibility for.

kotfic:
We should consider the implications of making [Build Girder](https://github.com/girder/girder/pull/1561/files#diff-d72e393515ab307f3cf939768eed1bcaR27) a handler,  that way it is only run once at the end of the play (See [Intro to Playbooks](http://docs.ansible.com/ansible/playbooks_intro.html#handlers-running-operations-on-change)).  I am thinking specifically about the case where someone may want to install a number of girder plugins (potentially from other roles) and not build the frontend until  all plugins are installed.  This may need a little testing,  i'm not sure if handlers are run before/after post_tasks,  which may effect the decision.

devops/ansible/roles/girder/tasks/girder.yml
```diff
@@ -0,0 +1,31 @@
...
+  git:
+    repo: "https://github.com/girder/girder.git"
+    dest: "{{ girder_path }}"
+    version: "{{ girder_version }}"
+    update: "{{ girder_update|default(omit) }}"
+    force: "{{ girder_force|default(omit) }}"
+  register: vc
+
+- name: Build Girder
+  command: "npm install --production"
```
zachmullen: We should use `girder-install web` for this if possible.

Purg:
Should there be some handlers added for restarting girder server?

Vagrantfile
```diff
@@ -1,33 +1,35 @@
...
   config.vm.hostname = "girder"
 
   config.vm.network "forwarded_port", guest: 8080, host: 9080
   config.vm.post_up_message = "Girder is running at http://localhost:9080"
 
   config.vm.synced_folder ".", "/vagrant", disabled: true
-  config.vm.synced_folder ".", "/home/vagrant/girder"
+  #config.vm.synced_folder ".", "/home/vagrant/girder"
 
-  provisioner_type = if
```
brianhelba: Why are you removing support for `ansible-local` provisioning with Vagrant?

kotfic:
@danlamanna Being able to run nightly tests against the 2.0-integration branch would be extremely helpful.  At least for the configuration module it would be nice to catch/fix any failures before the release. Are there technical reasons that prevent us from being able to do this?

danlamanna:
@kotfic Not at all. My plan was to run this against master and the latest release each night (once a release with this was made), I'll just add the 2.0 integration branch to the list.

danlamanna:
@kotfic This is ready for a final pass.

kotfic:
another comment

docs/provisioning.rst
```diff
@@ -0,0 +1,52 @@
...
+
+.. note:: Our playbooks are only currently supporting Ubuntu 14.04 and require Ansible >= 2.1.1.
+
+Example: Deploying Girder with NGINX to a VirtualBox
+####################################################
+Assuming a working copy of Girder is in your current directory: ::
+
+   GIRDER_EXAMPLE=girder-nginx vagrant up
+
+.. note:: A full list of examples that can be used exist in devops/ansible/examples.
```
kotfic: 'A full list of usable examples is available in ...'

mgrauer:
🎊 🐺

zachmullen:
:clap:

cjh1:
:beers:

PR #1562, opened 2016-09-07
Add -U flag when installing plugin requirements files

zachmullen:
@manthey @cdeepakroy the issue we just ran into when installing HistomicsTK pip requirements made me realize we should probably be passing the `-U` flag, which applies recursively. I haven't yet verified that this would fix the issue we saw, but it seems likely, and seems like a good idea regardless.

### DIFF
--- a/girder/utility/install.py
+++ b/girder/utility/install.py
@@ -129,7 +129,7 @@ def install_plugin(opts):
                     print(constants.TerminalColor.info(
                         'Installing pip requirements for %s from %s.' % (name, reqs)))
 
-                    if pip.main(['install', '-r', reqs]) != 0:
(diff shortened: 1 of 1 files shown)


manthey:
This touches on the whole conflicting requirements issue, where we really should be using >= requirements where possible.  Otherwise, if we up the version required by one plugin, other plugins will compete with it.    At present, the first installed wins, which isn't desirable either.

jbeezley:
The change is okay with me, but I did remove the `-U` a while back on purpose.  The flag will force all deps to be updated to the recent version even those that were explicitly pinned during a prior pip install.  Without `-U`, my understanding is it will still update deps that are strictly required to be updated.  I haven't checked this though.

zachmullen:
If it was removed intentionally, I'll back this out.

PR #1563, opened 2016-09-08
Allow use of local fontello archive instead of downloading

zachmullen:
This is controlled via the `GIRDER_LOCAL_FONTELLO_ARCHIVE` variable.

@cpatrick PTAL


cpatrick:
LGTMWTIH. It would great if we could test this, but that might be too difficult to bother.

zachmullen:
Actually we can test it, I'll go ahead and add something to the `packaging_install` test.

PR #1566, opened 2016-09-12
Fix last failing web client tests

zachmullen:


sebastienbarre:
Of the remaining tests failing [here](http://my.cdash.org/viewTest.php?onlyfailed&buildid=1046096):
- web_client_folder
- web_client_data_gridfsrs
- web_client_data_filesystem

or [here](http://my.cdash.org/viewTest.php?onlyfailed&buildid=1046291):
- web_client_thumbnails.thumbnails
- web_client_geospatial.geospatial
- web_client_user

or [here](http://my.cdash.org/viewTest.php?onlyfailed&buildid=1046099)
- web_client_autojoin.autojoin
- web_client_routing
- web_client_group

None fail on my OSX laptop.

@zachmullen  Should we rope in a Unix guy? They seem to be stuck on "Sign up" dialog. Is it the expected workflow? Or should they be signing in?

zachmullen:
I suspect these are simply race conditions in our tests; travis has always been quite good at finding these due to having much more limited resources than the average developer machine.

zachmullen:
Well, the tests are passing on travis. I'm certain the race conditions still exist, but we can go ahead and merge this since it does have some legitimate fixes in it.

PR #1567, opened 2016-09-13
Consistancy check fixes

brianhelba:
* Make Model "updateSize" methods not require a user
  * Access controls never have an impact on the "size" of Collections, Users, Folders, or Items. Sizes for Collections and Users represent the total size of all content recursively within them, regardless of whether a given user has access to any of it.
  * As a practical matter, the system consistancy check which typically calls "updateSize" may only be run by a site admin user, who will have access to everything.
* Remove user access checks from _fixBaseParents in consistancy check
  * A user may have access to a sub-folder that is within a parent folder or collection that the user does not have access to (this is how "unlisted" folders can be implemented). Regardless, these sub-folders should always have an accurate "baseParentId" and "baseParentType", even if a user does not have rights to actually traverse to this base.
  * As a practical matter, the system consistancy check may only be run by a site admin user, who will have access to everything.
* Remove user access checks from _pruneOrphans in consistancy check
  * A user may have access to a sub-folder that is within a parent folder or collection that the user does not have access to (this is how "unlisted" folders can be implemented). Thus, these sort of sub-folders are not technically 'orphans' and definitely should not be removed.
  * As a practical matter, the system consistancy check may only be run by a site admin user, who will have access to everything.
* Make "File.isOrphan" work with references to non-access-controlled models
  * Plugins may attach Files to non-access-controlled models. Previously, this was causing "File.isOrphan" to raise unhandled exceptions.
* Make "File.isOrphan" work with plugin "attachedToType" Models
  * Plugins may want to attach Files to plugin-based Model types. These should be handled gracefully and not cause Files to appear orphaned.
  * These plugins are expected to set "attachedToType" to be a list of the form ["model_type", "plugin_name"].

### DIFF
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -449,13 +449,12 @@ class System(Resource):
 
     def _recalculateSizes(self, progress):
         fixes = 0
-        user = self.getCurrentUser()
         models = ['collection', 'user']
         steps = sum(self.model(model).find().count() for model in models)
         progress.update(total=steps, current=0)
         for model in models:
             for doc in self.model(model).find():
                 progress.update(increment=1)
-                _, f = self.model(model).updateSize(doc, user)
+                _, f = self.model(model).updateSize(doc)
                 fixes += f
         return fixes
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -413,14 +413,13 @@ class System(Resource):
 
     def _fixBaseParents(self, progress):
         fixes = 0
-        user = self.getCurrentUser()
         models = ['folder', 'item']
         steps = sum(self.model(model).find().count() for model in models)
         progress.update(total=steps, current=0)
         for model in models:
             for doc in self.model(model).find():
                 progress.update(increment=1)
-                baseParent = self.model(model).parentsToRoot(doc, user=user)[0]
+                baseParent = self.model(model).parentsToRoot(doc, force=True)[0]
                 baseParentType = baseParent['type']
                 baseParentId = baseParent['object']['_id']
                 if (doc['baseParentType'] != baseParentType or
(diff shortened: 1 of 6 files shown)


brianhelba:
@fogleman PTAL.

brianhelba:
This fixes some issues with https://github.com/girder/girder/pull/1369

jbeezley:
When I added the [`/resource/:id/path`](https://github.com/girder/girder/blob/master/girder/api/v1/resource.py#L287-L312) endpoint I was unsure how to handle "unlisted" folders.  The current behavior is to return `403 Read access was denied for the resource`, which could be confusing.  Since you are bringing this sort of folder up, do you have any better ideas for how to handle it?

brianhelba:
@jbeezley Short answer: That endpoint should 403 only if the original resource is inaccessible. It should still print the names of any inaccessible resources that lie along the path up to the root.

Let me lay out my thinking:

I think the concept of 'unlisted' folders is quite useful, and analogous to a POSIX directory with permissions `--x`: you can `cd` to it, but not list the files or get any metadata. However, if you happened to know the exact name of a subdirectory that's `r-x`, you can further `cd` to it and behave normally again.

To follow this analogy, the full 'traversal path' of inaccessible parent folders should not be protected, but the metadata of each component component along that path is. For most of Girder, this 'traversal path' is just a list of ObjectIds, which I'd feel comfortable revealing. However a lot of things within the `/resource` endpoint are starting to attach special semantics to the `'name'` attribute, using it as a navigation mechanism.

So now that the cat is out of the bag with this `/resource` endpoint, I think we have to accept that the `'name'` of an otherwise-inaccessible resource can be revealed if a user is able to access a sub-folder.

jbeezley:
Another behavior to note with unlisted folders is the current behavior of the girder client.  It looks like this:

screen shot 2016-09-13 at 12 35 55 pm

The page loads, but the breadcrumb bar emits an access error and doesn't load correctly.

cpatrick:
@jbeezley I've seen that to. It's irritating...

fogleman:
LGTM

PR #1568, opened 2016-09-13
Prevent all disabled users from logging in

brianhelba:
* Perform filtering of all returned user documents in the user API
  * Models returned through the API should almost always be filtered. Users may not necessarily have access to every field in their own user document.
* Check for site admin in a more consistent way
  * User documents have had the `'admin'` property since the beginning of Girder, and it may only be `True` or `False`.
* Check for "user['emailVerified']" in a more consistent way
  * User documents have also had the 'emailVerified' property since the beginning of Girder, and it has always defaulted to False.
* Fix access to unset `user['status']` field
  * User documents from legacy databases do not have the `status` field set. While `status` gets set on `User.save` (see 3f450cc9a2540942f46db90aaaee939bbe85488b), it may not be set in time for some other operations.
* Prevent disabled users from logging in
  * Disabled users should not be able to log in, regardless of the registration policy. This includes disabled admin users.
  * `User.emailVerificationRequired` now uses the same logic for both admin and non-admin users. This means that newly created admin users will still have to verify their email addresses if the instance requires it generally.
  * `User.adminApprovalRequired` now uses the same logic for both admin and non-admin users. This means that newly created admin users will still have to have their accounts approved by another admin user if the instance requires it.
* Don't send approval emails to re-enabled users
  * An email message will only be sent on the transition from `'pending'` -> `'enabled'`. This should help to prevent misleading "approval" emails from being sent when users are changed `'disabled'` -> `'enabled'`.

### DIFF
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -268,7 +268,8 @@ class User(Resource):
                     raise AccessException('Only admins may change admin state.')
 
         # Only admins can change status
-        if 'status' in params and params['status'] != user['status']:
+        if 'status' in params and \
+                params['status'] != user.get('status', 'enabled'):
             if self.getCurrentUser()['admin']:
                 user['status'] = params['status']
                 if user['status'] == 'enabled':
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -212,15 +212,15 @@ class User(Resource):
             email=params['email'], firstName=params['firstName'],
             lastName=params['lastName'], admin=admin)
 
+        outputUser = self.model('user').filter(user, user)
         if not currentUser and self.model('user').canLogin(user):
             setattr(cherrypy.request, 'girderUser', user)
             token = self.sendAuthTokenCookie(user)
-            user['authToken'] = {
+            outputUser['authToken'] = {
                 'token': token['_id'],
                 'expires': token['expires']
             }
-
-        return user
+        return outputUser
 
     @access.user
     @loadmodel(map={'id': 'userToDelete'}, model='user', level=AccessType.ADMIN)
@@ -276,7 +276,8 @@ class User(Resource):
             else:
                 raise AccessException('Only admins may change status.')
 
-        return self.model('user').save(user)
+        user = self.model('user').save(user)
+        return self.model('user').filter(user, user)
 
     @access.admin
     @loadmodel(model='user', level=AccessType.ADMIN)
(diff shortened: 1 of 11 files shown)


brianhelba:
@fogleman PTAL. This fixes some issues with #1394 and #1408.

@zachmullen PTAL.

brianhelba:
@girder/developers 

The client tests are failing because they try to turn off email verification, then log in an un-verified admin user. Strictly speaking, this should now fail, but it does bring up the scenario where a site admin accidentally disables their own account and has no alternative site admin accounts to correct the problem.

For the case where an un-verified site admin enables email verification, the user can correct the problem by simply verifying their account's email address. If that email address is inaccessible, then I'd be concerned that there could be some serious security shenanigans going on, which Girder should not try to work around. The admin user should stay inaccessible.

For the case where a site admin disables their own account, we can handle it one of 3 ways:
1. Prohibit site admins from disabling their own accounts
2. Prohibit site admins from disabling their own accounts if it's the last enabled one remaining
3. Always allow site admins to disable their own accounts, at their own risk

I'm leaning towards 1., but I'd like to get some feedback on this whole situation before I implement the restriction.

brianhelba:
@girder/developers 

I've just considered the case where an inexperienced site admin on a fresh instance turns on email verification before they actually set up SMTP.

We could deal with this by either:
1. Prohibiting the enabling of email verification if there are no verified site admins.
2. When the first Girder user is created, mark their account as verified, in addition to being a site admin.

I'd say 1. is the less magical option, but again, I'd like to hear feedback.

fogleman:
For option 1, how would the admin verify their email address before enabling email verification?

zachmullen:
I vote for #2.

brianhelba:
Any feedback on my 3 options for the disabled site admin problem?

zachmullen:
I vote # 3 for that one.

brianhelba:
Additional changes:
* Ensure that the default site admin can always log in
* Fix a failing test for disabled users
  * This actually makes the test more specific, as previouslay it was treating a disabled user as the same as an unapproved user.

brianhelba:
@girder/developers PTAL.

girder/api/v1/user.py
```diff
@@ -268,15 +275,17 @@ def updateUser(self, user, params):
...
 
         # Only admins can change status
-        if 'status' in params and params['status'] != user['status']:
-            if self.getCurrentUser()['admin']:
-                user['status'] = params['status']
-                if user['status'] == 'enabled':
-                    self.model('user')._sendApprovedEmail(user)
-            else:
+        if 'status' in params and \
+                params['status'] != user.get('status', 'enabled'):
```
zachmullen: We have 100 character line limit now, so this backslash can go away.

brianhelba: Fixed.

brianhelba: Ping @zachmullen

PR #1569, opened 2016-09-13
Capture the return value of "Model.save" where it is used subsequently

brianhelba:
The contract of `Model.save` does not guarantee that the argument will be correctly mutated. Rather, the caller should overwrite the passed document with the return value.


zachmullen:
LGTM

PR #1570, opened 2016-09-13
Use "setResponseHeader" to set all response headers

brianhelba:
These abstracted methods should be preferred over direct use of CherryPy whenever possible.

### DIFF
--- a/girder/utility/abstract_assetstore_adapter.py
+++ b/girder/utility/abstract_assetstore_adapter.py
@@ -14,15 +14,15 @@
 #  limitations under the License.
 ###############################################################################
 
-import cherrypy
 import os
 import re
 import six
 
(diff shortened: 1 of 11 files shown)


girder/utility/abstract_assetstore_adapter.py
```diff
@@ -156,8 +156,8 @@ def downloadFile(self, file, offset=0, headers=True, endByte=None,
         """
         This method is in charge of returning a value to the RESTful endpoint
         that can be used to download the file. This can return a generator
-        function that streams the file directly, or can modify the cherrypy
-        request headers and perform a redirect and return None, for example.
+        function that streams the file directly, or can modify the response
```
mgrauer: good catch

mgrauer:
LGTM

PR #1571, opened 2016-09-13
Add script for unzipping zipped items in girder

fogleman:
This is intended to be run after migrating from Midas to Girder.

During migration, Midas items that had multiple "bitstreams" are downloaded as a zip file and stored in Girder as such. This script looks for zip files that are flat (have no folders) and have multiple file entries and unzips them into the same item, removes the zip file, and removes the .zip extension from the item.

I've run the script on QIDW. (Still running actually.)

### DIFF
--- a/scripts/midas/unzip_items.py
+++ b/scripts/midas/unzip_items.py
@@ -1,0 +1,104 @@
+import girder_client
+import os
+import tempfile
+import zipfile
+
+DRY_RUN = False
+
+MAX_ZIP_FILE_SIZE = -1
+# MAX_ZIP_FILE_SIZE = 1024 * 1024 * 1024 # 1GB
+
+GIRDER_URL = 'http://127.0.0.1/api/v1'
+GIRDER_API_KEY = 'API_KEY'
+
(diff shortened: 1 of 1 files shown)


cpatrick:
I really like your description here, could you add it as a block comment at the top of the script?

A few other notes:
* Could we add style tests to this scripts directory?
* Could you add a small argparse interface to this?

After that, LGTMWTIH.

zachmullen:
Is there not a way to download the bitstreams from Midas separately?

fogleman:
@zachmullen Not in pydas. Not sure about otherwise.

zachmullen:
That's something we should add to pydas. It would be possible by calling the lower-level `request` method of the driver.

fogleman:
@zachmullen Agreed. I can take a look at some point in the future. For now, this script has done everything I needed.

zachmullen:
I'm going to close this for now since I don't think we'd run it against most girder instances since it acts globally on all .zip files. I've recorded the underlying bug in pydas here: https://github.com/midasplatform/pydas/issues/36

PR #1573, opened 2016-09-13
Increase file descriptor resource limit in travis

zachmullen:
I had to increase the ulimit for file handles on my local machine in order to get mongo not to terminate, and the errors I was seeing locally were similar to the errors we have been seeing in the web client tests. This is an attempt to mitigate the travis failures in the web client testing, I have no idea if it will work though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,8 @@ before_install:
     # Show the memory state; this lets us more quickly determine when the
     # travis environment is bad
     - vmstat
+    # Mongo can use lots of file descriptors
+    - ulimit -S -n 2048
     - export CXX=g++-4.8
     - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "3" ]; then export PY3="true"; else export PY2="true"; fi
     - if [ -n "${PY3}" ]; then export MONGO_VERSION=3.0.7; export PY_COVG="OFF"; else export MONGO_VERSION=2.6.11; export PY_COVG="ON"; export DEPLOY=true; fi
--- a/clients/web/src/views/body/ItemView.js
+++ b/clients/web/src/views/body/ItemView.js
@@ -113,6 +113,7 @@ var ItemView = View.extend({
         // it. TODO: load the page and adjust only the action menu once
         // the access level is fetched.
         this.model.getAccessLevel(_.bind(function (accessLevel) {
+            this.accessLevel = accessLevel;
             this.$el.html(ItemPageTemplate({
                 item: this.model,
                 accessLevel: accessLevel,
--- a/clients/web/test/lib/jasmine-1.3.1/ConsoleReporter.js
+++ b/clients/web/test/lib/jasmine-1.3.1/ConsoleReporter.js
@@ -99,6 +99,10 @@
                     new Date().toISOString().replace(/:/g, '.') + '.png');
                 this.log('\n Error: ' + item.message, 'red');
                 this._printStackTrace(item.trace.stackArray);
+                window.callPhantom({
+                    action: 'exit',
+                    code: 1
+                });
             }
         }, this);
     };
--- a/clients/web/test/specRunner.js
+++ b/clients/web/test/specRunner.js
@@ -133,6 +133,10 @@ page.onCallback = function (data) {
                 fs.remove(uploadTemp);
             }
             break;
+        case 'exit':
+            // The "Testing Finished" string is magical and causes web_client_test.py not to retry
+            console.log('Testing Finished with status=' + data.code);
+            phantom.exit(data.code);
     }
 };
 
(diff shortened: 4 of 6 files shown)


zachmullen:
If this works I'll probably commit it to master too, it's just easier to get these problems to appear on 2.0-integration.

jbeezley:
My current theory is that the blanket callback doesn't fire at the correct time.  The additional complexity of the webpacked source means that the instrumented code takes longer to load and is more likely to cause a failure.

zachmullen:
Well, my idea didn't work, so I'm entertaining any and all theories. 😄 

Any idea as to how to fix this blanket callback issue, or at least test it?

zachmullen:
The geospatial one is actually failing for me locally on a consistent basis with the same message as travis. The others all appear to be failing in the same way, which is waiting for the login dialog to be rendered, even though the screenshots show that it has.

zachmullen:
I found the race condition in the geospatial test. Looking into the other ones now.

jbeezley:
Well when I looked at it before, I tried the best I could to ensure that we are always waiting for blanket correctly by loading them in order and making a promise available to the test code.  It didn't really help except for a couple of edge cases I had in HistomicsTK.  Now, the only solution I can think of is to not use blanket.

Of course, I could also be wrong and switching to istanbul would not help.  I was secretly hoping @sebastienbarre would be forced to make the change in his branch, and it would all go away by itself.  But, alas...

zachmullen:
I'm still all for switching to Istanbul, feel free to take a look in that direction if it interests you.

sebastienbarre:
The author of blanket himself [recommends Istanbul](https://github.com/alex-seville/blanket/commit/baea610953cf3481f54a1113fade06975ce45c39), so ultimately I agree it should be on the roadmap too. I kinda liked the idea of writing tests without involving Webpack, e.g. use the `girder` singleton for convenience and remap the coverage locations back to the original files via source map, but I don't blame you if it looks like a hack :)

Now the timing problems might be due to a dubious change [I made in the girderModal code](https://github.com/girder/girder/blob/2.0-integration/clients/web/src/utilities/jquery/girderModal.js#L34-L43).

zachmullen:
@sebastienbarre I found another issue, not sure if it's causing this last failure or not though. We used to have a separate file called main.js that actually created the main app and bound it to the selector `'body'`, but we are now doing that inside the top level main.js file, which is probably not what we want; we want to be able to import girder without having it try to bind to the DOM. We were also instantiating the app and binding to the `'body'` selector inside most of our test cases via the call to `girderTest.startApp()`, so everything was being bound twice.

I think the solution is to just have a small main file like we did before that is separate and contains just this logic:


    events.trigger('g:appload.before');
    var mainApp = new App({
        el: 'body',
        parentView: null
    });
    events.trigger('g:appload.after', mainApp);

sebastienbarre:
@zachmullen I'm confused. [Isn't it done already in main.js](https://github.com/girder/girder/blob/2.0-integration/clients/web/src/main.js#L22-L27)?

zachmullen:
Right, we should remove it from that entry point and make a separate file for that logic; right now that code is getting built into `girder.app.min.js`, meaning if you want to use any of the girder code, it's instantiating an App view and binding it to the DOM.

sebastienbarre:
Oh OK. Yes I actually had set it up the way you are describing, e.g. there were 3 bundles generated by Webpack: `girder.ext` for all external libs, `girder.app` for Girder code, and `girder.main` for, well, simply `main.js`. But Webpack was still pulling too much code in `girder.main`. I can look at it again. I'd rather leave the code `var mainApp = new App` in `main.js` and move `window.girder = girder;` to `global.js`, which make sense anyway.

zachmullen:
For main.js, we don't even need to build it with webpack, it can really just be a static single file that contains that logic and only works if girder.app.min.js is already there.

sebastienbarre:
I see, and rewrite it to use the singleton.  I had not noticed that double-binding, you might be onto something there. So basically you want the code that is in [clients/web/tests/testUtils.js](https://github.com/girder/girder/blob/2.0-integration/clients/web/test/testUtils.js#L1279-L1284) into a separate file, or directly in `girder/utility/webroot.mako`?
```
        girder.events.trigger('g:appload.before');
        var app = new girder.views.App({
            el: 'body',
            parentView: null
        });
        girder.events.trigger('g:appload.after');
```

zachmullen:
Yep, whichever one of those things you find easier is fine by me.

sebastienbarre:
OK let me try.

sebastienbarre:
Alright, let's see what Travis has to say about bf78ba3

sebastienbarre:
I received an email from Travis with a strange color, something I had never seen before. It looked like... green? With a strange word that said "passed". Doesn't ring a bell but I'll merge. Be still my heart.

zachmullen:
@sebastienbarre your change is now merged into this topic and we are green! If this LGTY, we can merge this PR.

PR #1574, opened 2016-09-13
Job list filters

cjh1:
This PR expose the job "type" and adds the ability to file the job list based on type and status, see screenshot below:

filter


cjh1:
@mgrauer PTAL 

One open issue is what todo out status changes, do they disappear from the view if the user has a filter set?

plugins/jobs/web_client/js/views/JobListWidget.js
```diff
@@ -1,16 +1,19 @@
...
     events: {
-        'click .g-job-trigger-link': function (e) {
+        'click .g-job-checkbox-menu input': function (e) {
             var cid = $(e.target).attr('cid');
             this.trigger('g:jobClicked', this.collection.get(cid));
         }
     },
 
     initialize: function (settings) {
+        var self = this;
```
mgrauer: Can you use `_.bind(function (e) {}, this);` instead of `self = this`?  That is more consistent with Girder js generally.

plugins/jobs/web_client/js/views/JobListWidget.js
```diff
@@ -85,6 +129,30 @@ girder.views.jobs_JobListWidget = girder.View.extend({
         window.setTimeout(function () {
             tr.removeClass('g-highlight');
         }, 1000);
+    },
+    _filterJobs: function (jobs) {
+        var filterJobs = [], self = this;
```
mgrauer: `_.bind` instead of `self`

plugins/jobs/web_client/templates/jobs_checkBoxMenu.jade
```diff
@@ -0,0 +1,11 @@
+a.g-job-checkbox-menu-link(data-toggle="dropdown", data-target="#g-job-checkbox-menu-#{title}")
+      | #{title}
+      i.icon-down-dir
```
mgrauer: I'd prefer `filter` to `down-dir`

plugins/jobs/web_client/js/views/JobListWidget.js
```diff
@@ -85,6 +129,30 @@ girder.views.jobs_JobListWidget = girder.View.extend({
...
+        filterJobs = this.collection.filter(function (job) {
+            return ((_.isEmpty(self.typeFilter) || _.isUndefined(job.attributes.type) ||
+                        self.typeFilter[job.attributes.type]) &&
+                    (_.isEmpty(self.statusFilter) || _.isUndefined(job.attributes.status) ||
+                        self.statusFilter[girder.jobs_JobStatus.text(job.attributes.status)]));
+        });
+
+        return filterJobs;
+    },
+    _updateFilter: function (filter, newValues) {
```
mgrauer: Can you provide a comment here explaining why you go to this trouble of updating the existing filter rather than just generating a new one based on the current set of unique values?

I think it relates to maintaining the current state of the filter for any filter states that will remain (were not removed), but I'm not sure.

plugins/jobs/web_client/js/views/JobListWidget.js
```diff
@@ -85,6 +129,30 @@ girder.views.jobs_JobListWidget = girder.View.extend({
         window.setTimeout(function () {
             tr.removeClass('g-highlight');
         }, 1000);
+    },
+    _filterJobs: function (jobs) {
+        var filterJobs = [], self = this;
+
+        filterJobs = this.collection.filter(function (job) {
+            return ((_.isEmpty(self.typeFilter) || _.isUndefined(job.attributes.type) ||
```
mgrauer: If I'm reading this right, the filter will include all jobs that are of the filtered set of types/status, and also those jobs that don't have a type or status?

mgrauer: Actually, please comment this to explicitly describe the intended behavior.

plugins/jobs/web_client/js/views/JobListWidget.js
danlamanna: Actually, for listeners using `.on` you can use the [context argument](http://backbonejs.org/#Events-on) to bind it.

mgrauer:
Also, client side tests, if possible.

plugins/jobs/web_client/js/views/JobListWidget.js
cjh1: Done

cjh1: Done

plugins/jobs/web_client/js/views/JobListWidget.js
cjh1: Done

cjh1: Done

plugins/jobs/web_client/js/views/JobListWidget.js
cjh1: Done

plugins/jobs/web_client/templates/jobs_checkBoxMenu.jade
cjh1: I have changed this, not sure its as good as dir-down, but not big preference.

cjh1:
@mgrauer Can you point me to a good example of a plugin client test?

cjh1:
@mgrauer Sorry I didn't see it already had one, I will take a look at that one.

plugins/jobs/web_client/templates/jobs_checkBoxMenu.jade
mgrauer: I had a slight preference also--my real concern was that a filter could be on and the user forget about that, then not find something they were looking for.  I don't think changing the icon gets around that problem, but my hope was it helped a bit.

mgrauer:
Because this is web UI, we will need to also make changes to target the 2.0 integration branch.

@zachmullen said he could help guide those changes.

cjh1:
@mgrauer I think I have addressed all your feedback. PTAL

mgrauer:
This LGTM, but it should be created as a new PR targeting the 2.0-integration branch, but should probably wait until #1598 gets merged into 2.0-integration and then this PR will need to be updated a bit.

mgrauer:
Also, nice tests!

🌛

mgrauer:
@cjh1 I saw @zachmullen in 🍖 👾  and told him that as part of targeting the 2.0-integration branch we would push the client side part of the 1-to-2 migration guide.  So let's sync up before you start that work.

PR #1577, opened 2016-09-14
Use npm-install-retry to help with intermittent errors

cjh1:


zachmullen:
The place I more often than not see the error is during the `packaging_install` test, where we are still running regular old `npm install`. Maybe we could augment that shell script to check for the presence of `npm-install-retry` and, if found, pass that as the `--npm` argument to `girder-install web`.

zachmullen:
On second thought, that would still try to pass the `install` command line arg, so it may need more work than that unless `npm-install-retry install` is a valid command.

cjh1:
Seems that ```npm-install-retry install``` is a valid command :smile:

zachmullen:
Great! Here's the place: https://github.com/girder/girder/blob/master/tests/packaging/pip_install_girder.sh#L91

cjh1:
Case in point!  :smile:

cjh1:
@zachmullen Any pointers on getting the packaging_install test running locally, sees to be looking for a girder tarball in my source tree?

zachmullen:
You can run `grunt package` to build the package, or just run all of the packaging tests since one of the prior ones builds the package.

cjh1:
How are we expecting the process package to be installed,  referenced [here](https://github.com/girder/girder/blob/master/grunt_tasks/fontello.js#L2), I am getting process not found ...

cjh1:
```
>> Error: Cannot find module 'process'
```

zachmullen:
I thought that was part of the node standard library...

cjh1:
Not with the old version I was using, upgrading to stable, resolve it, thanks.

cjh1:
The "@zachmullen approved this change" sounds very grand :smile:

zachmullen:
That feature apparently just got released live in the last few minutes, I like it.

cjh1:
Yep, very nice.

PR #1578, opened 2016-09-14
Make use of new github PR code review features

zachmullen:
@girder/developers take note, now that we have a more official channel for marking final approval of code review, we can use that rather than the old "LGTM".

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,9 +45,8 @@ Getting your contributions integrated is relatively straightforward, here
 is the checklist:
 
 * All tests pass
-* Consensus is reached. This usually means that at least one reviewer added a `LGTM` comment
-and a reasonable amount of time passed without anyone objecting. `LGTM` is an
-acronym for _Looks Good to Me_.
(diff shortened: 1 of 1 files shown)


zachmullen:
That language predates this PR, and is also intentionally vague. There is legal precedent for intentionally vague use of the word "reasonable", so as to leave room for human judgement on an ad hoc basis.

zachmullen:
I was hoping to lose "LGTM" altogether as a practice, or do people feel strongly about keeping it?

fogleman:
You can approve without leaving a comment, so I think writing "LGTM" is superfluous now.

jbeezley:
As long as we can still post LTGM gifs. ;)

fogleman:
iz9rzzo

cpatrick:
image

cpatrick:
That's weird....

fogleman:
I guess I shouldn't use the first Google result.

manthey:
What is the equivalent of "LGTMWTIH"?

zachmullen:
Just approval is equivalent; we are programmatically disallowed from merging until tests pass anyway. 😄

PR #1581, opened 2016-09-16
Use Istanbul for javascript coverage

### LINKED ISSUE #1576: Replace blanket with istanbul for javascript coverage
I decided to take this on.  It doesn't look like it will be incredibly difficult.  It will work like this:

1. Instrument the code with [babel-plugin-istanbul](https://github.com/istanbuljs/babel-plugin-istanbul).  I checked that it works perfectly with our webpack setup.
2. After a test passes pass the global `__coverage__` object the instrumentation produces back to phantomjs.
3.  Phantomjs writes the coverage objects from each test to a unique file.
4. Run `istanbul report` to combine the coverage objects into standard coverage formats (and stdout reporting).
5.  Submit coverage files to CDash, codecov, coveralls, etc.
6. Profit

I think the most difficult part will be surgically removing all of the infrastructure in place for blanket coverage.

jbeezley:
This almost works; however there is something going on with the spec runner that I haven't been able to diagnose.  For one, if I remove [this line](https://github.com/girder/girder/compare/2.0-integration...istanbul?expand=1#diff-de2e7347d54425ec072a4f7d5d88647cR60), then all of the client tests retry five times and then claim to pass, but the coverage information makes clear none of the tests were ever run.  With that line, the tests usually work correctly, but sometimes timeout.

I assume it has to do with the script loading race condition that this branch was supposed to fix... I tried updating to phantomjs 2.x, but there was no change.  Updating jasmine would also be useful to try, but that would require many changes to the test code to switch to the new async syntax.

One other idea is to generate static html files for each test and avoid dynamically loading scripts.  I started trying that, but it turned into more work than I wanted to undertake without checking in here first.  If anyone has any advice, I'm all ears.

To get this to work, you have to build with coverage on by setting the environment variable `BABEL_ENV=cover`.  This is done [on travis](https://github.com/girder/girder/compare/2.0-integration...istanbul?expand=1#diff-354f30a63fb0907d4ad57269548329e3R65), but we will have to do so manually locally.  I couldn't come up with a better approach than this without making more invasive changes.

With the branch, we get better test coverage output, which gets submitted to [coveralls](https://coveralls.io/github/girder/girder?branch=istanbul) and [codecov](https://codecov.io/gh/girder/girder/branch/istanbul).  It also outputs the results html locally as shown in the screenshot below.

A couple of more things we probably want to add here:
* Documentation updates
* Combine the js coverage output with the python coverage (apparently you can post twice to codecov, and it will combine them for you), but I assume for cdash and coveralls we will need to do it manually.  Luckily there are a number of scripts that do that for standard formats.
* Add a failure threshold, though we could rely on the codecov service for this.

Closes #1576 

screen shot 2016-09-16 at 1 18 10 pm


jbeezley:
Yeah, we probably don't want both services post to the PR's.  We can configure that.

zachmullen:
>  if I remove this line, then all of the client tests retry five times and then claim to pass

We do have to pipe the output from the underlying window to the console, because our [wrapper script](https://github.com/girder/girder/blob/2.0-integration/tests/web_client_test.py#L182-L185) inspects the output and looks for certain messages to mark success. If we no longer need that complicated loop, it would be great to remove it.

jbeezley:
Oh I see, it's in a [chunk of code](https://github.com/girder/girder/pull/1581/files#diff-de2e7347d54425ec072a4f7d5d88647cL90) I deleted.  The only issue is with the flakiness then, I guess.

tests/js_coverage_tool.py
```diff
@@ -1,228 +0,0 @@
...
-running in the phantom environment. It is used to reset the coverage, as well
-as to combine and report it once all of the tests have been run.
-"""
-
-import argparse
-import collections
-import glob
-import os
-import six
-import sourcemap
```
zachmullen: This was the only place we were using sourcemap, so we should remove it from requirements-dev.txt.

clients/web/test/specRunner.js
```diff
@@ -78,20 +75,14 @@ page.onConsoleMessage = function (msg) {
...
-            console.log('Exception writing coverage results: ', e);
+    } else if (msg === 'ConsoleReporter finished') {
+        var success = this.page.evaluate(function () {
+            return window.jasmine_phantom_reporter.status === 'success';
+        });
+        if (success) {
+            reportCoverage();
         }
-    } else {
-        console.log(msg);
```
zachmullen: Did this have to be moved up to the top and out of the `else` block?

jbeezley: I didn't notice it before, I can move it after the conditional so the `__SCREENSHOT__` lines are not shown.

jbeezley:
I apparently lack the necessary permissions to change notification settings on coveralls.  Getting one on each push will probably become irritating. For codecov, we just need to add a [codecov.yml](http://docs.codecov.io/docs/codecov-yaml) file to the repo, but it just updates the original comment by default.

sebastienbarre:
@jbeezley nice!

a) What was the line you referenced in "For one, if I remove this line"? 

b) "BABEL_ENV=cover. This is done on travis, but we will have to do so manually locally.". That could be taken care of by a new entry in the `scripts` section of `package.json`, the same way you would do it to transpile your ES6 code to ES5 for distribution, in production mode:
```
    "dist": "NODE_ENV=production babel src -d dist --no-comments",
```
e.g.
```
    "coverage": "BABEL_ENV=cover ... rest of call, I assume instanbul something something
```

jbeezley:
It was [this one](https://github.com/girder/girder/pull/1581/commits/74881dff4bc27c44f726e46f173eb556338c8fbc).  @zachmullen explained why it was necessary.

Yeah, we could add a script for it.  I think it would just be
```
"coverage": "BABEL_ENV=cover npm run build"
```

PR #1582, opened 2016-09-17
Allow parallel testing of browser tests on travis

zachmullen:
I had removed this in an earlier branch to avoid out of memory issues, this is an experiment to see if we've resolved the underlying issue.


zachmullen:
I re-ran a couple times and everything is still green, so this seems safe. Seems to finish in around 8 minutes now instead of around 10.

PR #1585, opened 2016-09-19
Do not append a / in email links

zachmullen:
ping @ronichoudhury


zachmullen:
Nope, no need for LGTM

PR #1589, opened 2016-09-20
Buttons that looked disabled were actually not

### LINKED ISSUE #1498: Disabled buttons are not actually disabled

zachmullen:
Closes #1498 

I am reopening this on behalf of @vovythevov , originally was in PR #1532

### DIFF
--- a/clients/web/src/utilities/jquery/girderEnable.js
+++ b/clients/web/src/utilities/jquery/girderEnable.js
@@ -1,8 +1,5 @@
 import jQuery from 'jquery';
 
-/**
- * Define jQuery plugins within this scope.
- */
 (function ($) {
     /**
      * Helper selector to enable/disable inputs elements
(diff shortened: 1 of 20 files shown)


zachmullen:
```
Running "webpack:dev" (webpack) task
Warning: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration has an unknown property 'debug'. These properties are valid:
   object { amd?, bail?, cache?, context?, devServer?, devtool?, entry, externals?, loader?, module?, name?, dependencies?, node?, output?, plugins?, profile?, recordsInputPath?, recordsOutputPath?, recordsPath?, resolve?, resolveLoader?, stats?, target?, watch?, watchOptions? }
 - configuration.resolve.extensions[4] should not be empty. Use --force to continue.
```

That problem wasn't happening to me locally until I blew away node_modules and re-ran `npm install`, meaning this is something that changed in an upstream npm package (presumably webpack itself). @sebastienbarre any idea? Google wasn't very helpful

sebastienbarre:
I'll check today. This is an issue I had mentioned in the PR -- the debug
options were in flux in the 2.0 beta so I added all of them and a cleanup
might be needed.

zachmullen:
That makes sense, thanks for looking into it.

sebastienbarre:
Still working on this, I'm afraid the `grunt-webpack` plugin itself [is having issue passing an option not allowed by Webpack anymore](https://github.com/webpack/grunt-webpack/blob/master/tasks/webpack.js#L29), but it's unclear if sokra is still updating it ([I reported it](https://github.com/webpack/grunt-webpack/issues/65#issuecomment-248523613)). Might just execute webpack from grunt.

zachmullen:
@sebastienbarre PTAL, I just pinned to the last known working version of the webpack 2.0 beta package for now.

PR #1591, opened 2016-09-20
Fix a few issues when files don't have an itemId.

manthey:
When a file doesn't have an itemId OR an attached type and ID (which could happen while uploading a thumbnail, for instance), the acl mixin would fail.

When a file doesn't have an itemId, the provenance plugin would fail.

### DIFF
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -81,7 +81,7 @@ class Upload(Model):
             unspecified, the current assetstore is used.
         :type reference: str
         :param attachParent: if True, instead of creating an item within the
-            parent or giving the file an itemID, set itemID to None and set
+            parent or giving the file an itemId, set itemId to None and set
             attachedToType and attachedToId instead (using the values passed in
             parentType and parent).  This is intended for files that shouldn't
             appear as direct children of the parent, but are still associated
@@ -311,7 +311,7 @@ class Upload(Model):
         :param assetstore: An optional assetstore to use to store the file.  If
             unspecified, the current assetstore is used.
         :param attachParent: if True, instead of creating an item within the
-            parent or giving the file an itemID, set itemID to None and set
+            parent or giving the file an itemId, set itemId to None and set
             attachedToType and attachedToId instead (using the values passed in
             parentType and parent).  This is intended for files that shouldn't
             appear as direct children of the parent, but are still associated
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -49,7 +49,7 @@ class Upload(Model):
 
     def uploadFromFile(self, obj, size, name, parentType=None, parent=None,
                        user=None, mimeType=None, reference=None,
-                       assetstore=None):
+                       assetstore=None, attachParent=False):
         """
         This method wraps the entire upload process into a single function to
         facilitate "internal" uploads from a file-like object. Example:
@@ -67,10 +67,11 @@ class Upload(Model):
         :param size: The total size of
         :param name: The name of the file to create.
         :type name: str
-        :param parent: The parent (item or folder) to upload into.
-        :type parent: dict
         :param parentType: The type of the parent: "folder" or "item".
         :type parentType: str
+        :param parent: The parent (item or folder) to upload into, or the id of
+            the parent.
+        :type parent: dict or str
         :param user: The user who is creating the file.
         :type user: dict
         :param mimeType: MIME type of the file.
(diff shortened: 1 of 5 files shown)


girder/utility/acl_mixin.py
```diff
@@ -52,16 +52,15 @@ def load(self, id, level=AccessType.ADMIN, user=None, objectId=True,
...
                 loadId = doc[self.resourceParent]
             else:
-                loadType = doc['attachedToType']
-                loadId = doc['attachedToId']
-
-        if not force and doc is not None:
-            self.model(loadType).load(loadId, level=level, user=user, exc=exc)
+                loadType = doc.get('attachedToType')
+                loadId = doc.get('attachedToId')
+            if loadType is not None and loadId is not None:
```
brianhelba: I don't think the failure of this condition should silently pass. `Model.load` is a security-critical method. If something inheriting from `AccessControlMixin` fails to either specify `self.resourceColl` / `self.resourceParent` or have an `'attachedToType'` `'attachedToId'`, I think it should be an error, and the model should fail to load.

plugins/provenance/server/resource.py
```diff
@@ -394,7 +394,13 @@ def fileRemoveHandler(self, event):
         if 'itemId' not in file:
```
brianhelba: This check now seems superfluous.

brianhelba:
@manthey While reviewing this, I also noticed that the logic of [`getProvenanceUser`](https://github.com/girder/girder/blob/f2122672653f4f3d5948c6ca7a5caf7f3b80e0c2/plugins/provenance/server/resource.py#L211) does not seem to match the docstring; namely, the request's current user (if present) always overrides the document's user / creator.

girder/utility/acl_mixin.py
manthey: One problem is that we call `models.upload.uploadFromFile` with a `parentType=None` and `parent=None` because we are *going* to attach the upload via the `attachedToId` values, but it isn't attached *yet*.  Events that respond to `data.process` need access to the file, but, if we throw an exception here, they fail.

For a brief duration, we have a resource that has no parent nor attached id.  Either we have to let this pass (a mixed-in controlled resource with no owner allows permission, as this code allows), or we have additional logic to the `uploadFromFile` code to attach the file as it is uploaded.

There may be other cases where it is appropriate to have an ownerless resource (but I don't know what those would be).

plugins/provenance/server/resource.py
manthey: Why?  Files can either have `itemId`s or be `attachedToId`.  If they have an `itemId`, we add provenance to the item.  If they don't, we don't.  When a file with `attachedToId` is removed, we shouldn't even be trying to load an item (even if that would fail, which it might do otherwise, for instance, if the item were deleted before the file).

manthey:
Good point about `getProvenanceUser`.  The function does the right thing (since it is intended to be the user that caused the provenance change), but the docstring was wrong.

manthey:
@brianhelba I made the changes we discussed in person.

girder/models/upload.py
```diff
@@ -171,6 +178,8 @@ def finalizeUpload(self, upload, assetstore=None):
             file['assetstoreId'] = assetstore['_id']
             file['size'] = upload['size']
         else:  # Creating a new file record
+            if upload.get('attachParent'):
+                pass
```
brianhelba: I think you meant to also make the line below an `elif`.

girder/models/upload.py
```diff
@@ -80,11 +81,17 @@ def uploadFromFile(self, obj, size, name, parentType=None, parent=None,
         :param assetstore: An optional assetstore to use to store the file.  If
             unspecified, the current assetstore is used.
         :type reference: str
+        :param attachParent: if True, instead of giving the file a parentType
```
brianhelba: Strictly speaking, files don't have a `parent` or `parentType` property. Rather, the `itemId` property of the file is getting set to `None`.

girder/models/upload.py
```diff
@@ -297,6 +311,12 @@ def createUpload(self, user, name, parentType, parent, size, mimeType=None,
         :type reference: str
         :param assetstore: An optional assetstore to use to store the file.  If
             unspecified, the current assetstore is used.
+        :param attachParent: if True, instead of giving the file a parentType
```
brianhelba: Same issue with this docstring as above.

girder/models/upload.py
```diff
@@ -286,8 +299,9 @@ def createUpload(self, user, name, parentType, parent, size, mimeType=None,
         :type name: str
         :param parentType: The type of the parent being uploaded into.
         :type parentType: str ('folder' or 'item')
-        :param parent: The document representing the parent.
-        :type parentId: dict
+        :param parent: The document representing the parent or the id of the
```
brianhelba: Personally, I don't think we should go down this road of model methods accepting an `ObjectId` (or even just a `str`-type ID) in place of a model document. If we do so, we should be consistent about allowing this all across the model API, and such a change would add a huge amount of additional logic to handle all the coercions. As I understand the current design, the model API does not do heavy validation of its parameters and generally takes full documents as parameters.

Furthermore, the fact that the old code was doing `ObjectId(parent['_id'])` seems nonsensical, as a model document should never contain a `str`-typed `_id` field. Indeed, the only time an ID should be `str` typed is right after it's extracted as an API parameter; at that point, it should be passed to `load`, to validate it and do access checking.

So, I think this method should require that `parent` be a `dict`-typed model document. Of course, as an optimization, callers can always pass `parent={'_id': parentId}` when calling the function, if they don't want to do a full lookup of the parent document.

girder/models/upload.py
```diff
@@ -319,10 +339,13 @@ def createUpload(self, user, name, parentType, parent, size, mimeType=None,
 
         if parentType and parent:
             upload['parentType'] = parentType.lower()
-            upload['parentId'] = ObjectId(parent['_id'])
+            upload['parentId'] = ObjectId(
```
brianhelba: Per my comment above, I think this should just be
```
upload['parentId'] = parent['_id']
```

PR #1593, opened 2016-09-21
Py client download root resources

### LINKED ISSUE #1592: Add ability to download collections and users in python client

zachmullen:
Closes #1592

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -985,7 +985,7 @@ class GirderClient(object):
                 local = os.path.join(dest, self.transformFilename(folder['name']))
                 _safeMakedirs(local)
 
-                self.downloadFolderRecursive(folder['_id'], local)
+                self.downloadFolderRecursive(folder['_id'], local, sync=sync)
 
             offset += len(folders)
             if len(folders) < DEFAULT_PAGE_LIMIT:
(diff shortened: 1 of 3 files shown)


clients/python/girder_client/__init__.py
```diff
@@ -1012,6 +1012,42 @@ def downloadFolderRecursive(self, folderId, dest, sync=False):
             if len(items) < DEFAULT_PAGE_LIMIT:
                 break
 
+    def downloadResource(self, resourceId, dest, resourceType='folder', sync=False):
+        """
+        Download a folder recursively from Girder into a local directory.
+
+        :param resourceId: ID or path of the resource to download.
```
mgrauer: I'd like this to be more clear.  I looked through the overall file, and there is a mix of `resourceId` type params where some are an Id only and some are an Id or a path.  It would be nice to unify these, which is probably out of scope here.

Even better might be to rename these type of params to be `resourceIdentifier` (avoiding `resourceId` as this can be an Id or a path) and have the docstring indicate that they can be an ID or a path--this would be a breaking change, but it is small enough that we could get it into the 2.0 branch reasonably quickly.

There's a good docstring in `resourceLookup` that can be the basis here, though that one doesn't include `folder`, probably because you can't look up a folder by name?
 
        :param path: The path of the resource. The path must be an absolute
            Unix path starting with either "/user/[user name]" or
            "/collection/[collection name]".

My ideal (breaking change for other methods, but OK for this one) would be

    : param resourceIdentifier: resource Id or Girder resource path.  The path must be an absolute
            Unix path starting with either "/user/[user name]" or "/collection/[collection name]".

This is fine as a simple fix and a new issue.  I'd be happy to do the cleanup.

clients/python/girder_client/__init__.py
```diff
@@ -1012,6 +1012,42 @@ def downloadFolderRecursive(self, folderId, dest, sync=False):
...
+        :param dest: The local download destination.
```
mgrauer: specify if this is an absolute path or relative to cwd

clients/python/girder_client/__init__.py
zachmullen: Well, ID is short for identifier, so I'm not sure about that nomenclature. `idOrPath` seems more descriptive, what do you think?

mgrauer:
@zachmullen docstring changes only

zachmullen:
Your changes LGTM (I'd add a review but it won't let me since it's my own PR).

PR #1598, opened 2016-09-27
Jade to pug conversion 

zachmullen:
It makes sense to go ahead and push this through prior to 2.0 since it's necessarily a breaking change. There are several changes on this branch:

* All files with the .jade extension have been changed to have the .pug extension. The webpack config and all import refs within the JS code have been updated to reflect this change.
* Introduced the ``pug-lint`` tool as a dev dependency, and added a set of tests to perform static analysis of our pug code to make sure it is sensibly uniform and readable. This will help catch lingering broken things like old-style interpolation in attributes that won't be caught during compilation.
* In adding the pug-lint tests, I discovered that we had silently broken all of our eslint testing of plugins during the move to webpack. This restores that testing, and fixes several eslint findings in the plugins.
* Added babel-loader to the end of our template build pipeline to support the use of ES2015+ features within our templates.
* The migration guide was updated with information about the breaking changes.
* Plugin development docs were updated to document the new way to extend the web client via plugins.


zachmullen:
The good news is, we now actually have sourcemaps and coverage of our template files!

The bad news is, CTest dies when trying to read certain pug files for coverage...

```
Reading file: ../clients/web/src/templates/body/assetstores.pug
*** glibc detected *** ctest: malloc(): memory corruption: 0x00000000026077f0 ***
```

It may be easiest to just exclude them from coverage for now.

zachmullen:
Actually, I'm going to disable ctest_coverage in the browser build. It'll save time and won't be a big deal because we get a better vis UI from codecov anyway.

zachmullen:
Update: after looking at the [coverage report](https://codecov.io/gh/girder/girder/tree/80c101f26df12a0036c8af0d384bc7431937e068/clients/web/src/templates/body) of some of our template files, something went wrong in the instrumentation pipeline, and we aren't mapping correctly back to the source template files. For the moment I've simply excluded the templates from coverage as we were doing before. It would be great to get this working in the future, though.

grunt_tasks/webpack.config.js
```diff
@@ -189,15 +189,15 @@ module.exports = {
...
+            // /node_modules\/pug/,
             // /node_modules\/remarkable/
         ]
     },
     resolve: {
         alias: {
             'girder': paths.web_src
         },
-        extensions: ['.styl', '.css', '.jade', '.js', ''],
+        extensions: ['.styl', '.css', '.pug', '.jade', '.js', ''],
```
zachmullen: I had initially removed it, but then I added it back in case plugins still wanted to use the .jade extension. I don't have a strong preference one way or the other, though.

jbeezley: It's fine.  I don't have a strong preference either.

tests/.pug-lintrc
```diff
@@ -0,0 +1,19 @@
...
+"requireClassLiteralsBeforeAttributes": true,
+"requireIdLiteralsBeforeAttributes": true,
+"requireLowerCaseAttributes": true,
+"requireLowerCaseTags": true,
+"requireSpaceAfterCodeOperator": true,
+"requireStrictEqualityOperators": true,
+"validateAttributeSeparator": { "separator": ", ", "multiLineSeparator": ",\n    " },
+"validateDivTags": true,
+"validateTemplateString": true
+}
```
jbeezley: Could we move this to the project root or add the `pugLintConfig` key to the `package.json` as described in the [documentation](https://github.com/pugjs/pug-lint#configuration-file) so that editors will pick up the config?

zachmullen: I'll add the key. We have so many files at the root level now that I was trying to avoid adding another 🍂

danlamanna: Moving it to the root would bring us closer to balancing the number of linting configs we have in `/` and `/tests`.

jbeezley: Actually, I think the 2.0 branch only has the flake config in `/tests`.  Personally, hidden files don't bother me.  YMMV.

PR #1606, opened 2016-09-28
Specify minimum version of Node.js

fedorov:


docs/prerequisites.rst
```diff
@@ -6,7 +6,7 @@ The following software packages are required to be installed on your system:
 * `Python 2.7 or 3.4 <https://www.python.org>`_
 * `pip <https://pypi.python.org/pypi/pi>`_
 * `MongoDB 2.6+ <http://www.mongodb.org/>`_
-* `Node.js <http://nodejs.org/>`_
+* `Node.js 4.x+ <http://nodejs.org/>`_
```
manthey: In the .travis.yml file, we specify node v2.3.3.  Shouldn't these be the same, whatever version we pick?

jbeezley:
Is 4.x still needed after https://github.com/girder/girder/pull/1596 went in?

fedorov:
I just wanted to formalize the instructions that @zachmullen provided me over gitter ...

zachmullen:
I'm not actually sure what the precise minimum version we require is, I've just been recommending 4.x to people because I know it works. It seems that it must be lower than that since that's what travis uses.

fedorov:
@zachmullen makes sense. So we know (from my experience) it doesn't work with v0.10.46 (the stock version supplied by CentOS 6.7) and it works with whatever is used on travis, so should be something in the middle ...

zachmullen:
Someone can do a binary search through that version space if they want to suss it out. I'm OK with just recommending a new-ish version, which has the added benefit of being more compatible with future npm-related stuff we might add.

fedorov:
How about using the "community search" approach - list the lowest known working version, and add a comment asking users to report/submit a PR if they encounter a lower version that works! :)

zachmullen:
I'm reopening this PR via a local branch, thanks @Xarthisius

mgrauer:
@fedorov 

We understand your frustration here. We are working towards more explicitly stating our dependencies.  https://github.com/girder/girder/issues/1851 https://github.com/girder/girder/issues/1848.

I'm going to close this PR and we can follow up with those GH issues.

fedorov:
@mgrauer no frustration really, just a friendly suggestion.

PR #1609, opened 2016-10-03
Update base image in the Dockerfile

Xarthisius:
With `node:4` npm is unable to handle `grunt-webpack`, which
requires `webpack-1.x`. With `node:5` it only issues warning.
Dropping --production flag, since some of the development
dependencies are explicitly required now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:4.4.5
+FROM node:5
 MAINTAINER Patrick Reynolds <patrick.reynolds@kitware.com>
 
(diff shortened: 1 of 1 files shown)


zachmullen:
Thanks @Xarthisius ! I just opened #1610 to fix a bug, but once that is merged, we should be able to replace all these npm lines with just `girder-install web`, so I'd suggest just using this branch to do that.

zachmullen:
We need to remove the `--production` flag from npm install, and we can remove the grunt-cli install and cache clear step altogether.

PR #1617, opened 2016-10-04
Add docs for systemd mongod init script with Ubuntu 16.04

thewtex:
As suggested in the MongoDB docs:

  https://docs.mongodb.com/manual/tutorial/install-mongodb-on-ubuntu/

Thanks @zachmullen -- this works :-)


danlamanna:
Just an FYI - the mongo role used in some of the Ansible examples for Girder also [provides this](https://github.com/Stouts/Stouts.mongodb/blob/develop/files/mongod.service).

PR #1623, opened 2016-10-06
Add 'size' attribute to the 'linkUrl' file objects.

Xarthisius:
It doesn't hurt to know the size of a file before downloading. \<shameless plug\> I also need this property for the GirderFS fuse filesystem I'm developing. \</shameless plug\>

### DIFF
--- a/girder/api/sftp.py
+++ b/girder/api/sftp.py
@@ -55,10 +55,7 @@ def _handleErrors(fun):
 
 
 def _getFileSize(file):
-    if file.get('assetstoreId'):
-        return file['size']
-    elif file.get('linkUrl'):
-        return len(file['linkUrl'])
+    return file.get('size', len(file.get('linkUrl', '')))
 
 
 def _stat(doc, model):
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -104,10 +104,11 @@ class File(Resource):
                                              level=AccessType.WRITE, exc=True)
 
         if 'linkUrl' in params:
+            size = int(params.get('size', 0))
             return self.model('file').filter(
                 self.model('file').createLinkFile(
                     url=params['linkUrl'], parent=parent, name=params['name'],
-                    parentType=parentType, creator=user), user)
+                    parentType=parentType, creator=user, size=size), user)
         else:
             self.requireParams('size', params)
             assetstore = None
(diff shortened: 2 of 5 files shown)


danlamanna:
Side note: This is another area where something like http://docs.pyfilesystem.org/en/latest/ could come in handy as it already supports [exposing their filesystems via FUSE](http://docs.pyfilesystem.org/en/latest/expose.html).

zachmullen:
+1, although pyfilesystem currently has limited support for python 3.

zachmullen:
@Xarthisius I'm going to hold off on this for a bit, I'm planning to cut 2.0 release tomorrow, so we can rebase this on top of 2.0 afterward.

zachmullen:
Could we just HEAD it at the time it's requested by the client? That feels better to me than trying to cache the value at all.

zachmullen:
Just curious, does all this assume that the upstream HTTP server accepts byte range requests?

manthey:
There is no requirement that a linkUrl use the http scheme (you can store any value as the link).  If the scheme isn't http or https, is there any chance that performing a head request results in something undesirable?

zachmullen:
> There is no requirement that a linkUrl use the http scheme (you can store any value as the link). If the scheme isn't http or https, is there any chance that performing a head request results in something undesirable?

This is a good point, and outside just the scope of this PR. We may want to enforce only http and https schemes as valid for link URLs.

Xarthisius:
Anything scheme besides http{,s} would raise `RequestException`

Xarthisius:
@zachmullen in my use case urls' destinations are immutable. I haven't really thought about how to handle url that changes.

Xarthisius:
With addition of `size` as an optional argument to `createLinkFile` I'm ok with dropping HEAD checking altogether if that's controversial

zachmullen:
Let's go ahead and drop that part then 😄

girder/api/sftp.py
```diff
@@ -55,10 +55,7 @@ def wrapped(*args, **kwargs):
 
 
 def _getFileSize(file):
-    if file.get('assetstoreId'):
-        return file['size']
-    elif file.get('linkUrl'):
-        return len(file['linkUrl'])
+    return file.get('size', len(file.get('linkUrl', '')))
```
zachmullen: It looks like with the new model API, link files will *always* have a ``size`` field that defaults to 0, so this will always return 0 if none is passed. Is that the desired behavior? It seems like it might be better to only put a ``size`` field in the document if it was explicitly passed, and change the default value in the signature to ``None``.

PR #1625, opened 2016-10-07
Enforce max buffer length in SFTP download

zachmullen:
@cjh1 I just realized I forgot to add this change you requested yesterday. PTAL.

### DIFF
--- a/girder/api/sftp.py
+++ b/girder/api/sftp.py
@@ -33,6 +33,7 @@ from girder.utility.model_importer import ModelImporter
 from six.moves import socketserver
 
 DEFAULT_PORT = 8022
+MAX_BUF_LEN = 10 * 1024 * 1024
 
 
 def _handleErrors(fun):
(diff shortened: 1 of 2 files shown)


zachmullen:
In my testing, both the CLI sftp client that comes with OSX and Filezilla use a chunk size of 32768, so 10MB seemed like a reasonable threshold.

mgrauer:
Can you add a test for this case?

mgrauer:
Will this message be sent out to an SFTP client?  

    'Requested chunk length (%d) is larger than the maximum allowed.' % length)

If so, maybe it would be useful to also include our max value in the message so they can change their client setting (is that something you can even change on a client?).

zachmullen:
Sadly no. Paramiko converts all exceptions to SFTP standard status codes and messages, so clients never get to see these custom ones.

zachmullen:
@cjh1 I'm not able to reproduce that error

cjh1:
ok, got to the bottom of that error, not related to this change ( environmental, its a long story )

zachmullen:
Thanks!

PR #1629, opened 2016-10-07
Remove support for nonstandard filter() kwargs

zachmullen:
Fingers crossed, this is the last breaking change before we can cut 2.0.


cpatrick:
LGTM.

PR #1630, opened 2016-10-07
2.0 integration

zachmullen:
Here it is, try not to crash your browser!


zachmullen:
Thanks Seb! @girder/developers it would be great if some other devs can mark their review status on this just to make sure we have consensus that it's OK to go ahead with the release.

Thanks all.

mgrauer:
Way to bring this home team.

Soon it will be time to hoist a celebratory glass of 🐗 💣

PR #1636, opened 2016-10-12
Regular files should not show up in available plugin list

zachmullen:
".gitignore" and ".DS_Store" were showing up as plugins in my list prior to this change, which must have come in when we removed the deprecated feature of multiple plugin dirs.


zachmullen:
ping @girder/developers could anyone take a look at this small bugfix?

PR #1640, opened 2016-10-14
Fix vega build

zachmullen:
A recent release of one of vega's dependencies requires json
files to be loaded as part of the build, so we use webpack's
json-loader.

@girder/developers could someone PTAL, this is breaking all of our CI.

### DIFF
--- a/grunt_tasks/webpack.config.js
+++ b/grunt_tasks/webpack.config.js
@@ -105,6 +105,11 @@ module.exports = {
                 loader: 'babel-loader',
                 exclude: [paths.node_modules]
             },
+            // JSON files
+            {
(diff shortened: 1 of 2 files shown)


mgrauer:
Tested and works.  This was my first time turning on a new plugin with 2.0 after I had built the web client, I was a bit confused by the UI asking me about rebuilding the web code and restarting the server.  I did them both in that order, but I think there could be better guidance provided from this UI about when to do what.

screen shot 2016-10-14 at 5 15 08 pm

zachmullen:
Thanks Mike! Your intuition about the order was correct, but a message popping up when you switch a plugin on would be helpful.

PR #1642, opened 2016-10-14
Add config option for disabling the events daemon

zachmullen:
@mgrauer PTAL. This allows a config file switch to be set such that `girder.events.daemon` simply becomes an alias for the normal, synchronous `girder.events`, which makes us no longer depend on having a background thread on the server.

### DIFF
--- a/girder/events.py
+++ b/girder/events.py
@@ -42,9 +42,11 @@ receive the Event object as its only argument.
 """
 
 import contextlib
+import girder
 import threading
 
-import girder
+
+from girder.utility import config
 from six.moves import queue
 
 
@@ -107,6 +109,26 @@ class Event(object):
         self.responses.append(response)
 
 
+class ForegroundEventsDaemon(object):
+    """
+    This is the implementation used for ``girder.events.daemon`` if the
+    config file chooses to disable using the background thread for the daemon.
+    It executes all bound handlers in the current thread, and provides
+    no-op start() and stop() implementations to remain compatible with the
+    API of AsyncEventsThread.
+    """
+    def start(self):
+        pass
+
+    def stop(self):
+        pass
+
+    def trigger(self, eventName, info=None, callback=None):
+        event = trigger(eventName, info, async=False, daemon=True)
+        if callable(callback):
+            callback(event)
+
+
 class AsyncEventsThread(threading.Thread):
     """
     This class is used to execute the pipeline for events asynchronously.
@@ -115,6 +137,7 @@ class AsyncEventsThread(threading.Thread):
     """
     def __init__(self):
         threading.Thread.__init__(self)
+
         self.daemon = True
         self.terminate = False
         self.eventQueue = queue.Queue()
@@ -130,13 +153,12 @@ class AsyncEventsThread(threading.Thread):
         while not self.terminate:
             eventName, info, callback = self.eventQueue.get(block=True)
             try:
-                event = trigger(eventName, info, async=True)
+                event = trigger(eventName, info, async=True, daemon=True)
                 if callable(callback):
                     callback(event)
             except Exception:
-                girder.logger.exception(
-                    'In handler for event "%s":' % eventName)
-                pass  # Must continue the event loop even if handler failed
+                # Must continue the event loop even if handler failed
+                girder.logger.exception('In handler for event "%s":' % eventName)
 
         girder.logprint.info('Stopped asynchronous event manager thread.')
 
@@ -146,6 +168,9 @@ class AsyncEventsThread(threading.Thread):
 
         :param eventName: The event name to pass to the girder.events.trigger
         :param info: The info object to pass to girder.events.trigger
+        :param callback: Optional callable to be called upon completion of
+            all bound event handlers. It takes one argument, which is the
+            event object itself.
         """
         self.eventQueue.put((eventName, info, callback))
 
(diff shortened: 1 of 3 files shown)


girder/utility/config.py
```diff
@@ -40,15 +40,15 @@ def _loadConfigsByPrecedent():
...
-        os.path.join(PACKAGE_DIR, 'conf', 'girder.dist.cfg'))
-    configPaths.append(
-        os.path.join(PACKAGE_DIR, 'conf', 'girder.local.cfg'))
-    configPaths.append(
-        os.path.join('/etc', 'girder.cfg'))
-    configPaths.append(
-        os.path.join(os.path.expanduser('~'), '.girder', 'girder.cfg'))
+    configPaths = [os.path.join(PACKAGE_DIR, 'conf', 'girder.dist.cfg')]
+
+    if 'GIRDER_TEST_DB' not in os.environ:
```
mgrauer: why this change?  is it necessary for the new test?

girder/utility/server.py
```diff
@@ -103,8 +103,9 @@ def configureServer(test=False, plugins=None, curConfig=None):
     root = webroot.Webroot()
     api_main.addApiToNode(root)
 
-    cherrypy.engine.subscribe('start', girder.events.daemon.start)
-    cherrypy.engine.subscribe('stop', girder.events.daemon.stop)
+    if not curConfig['server'].get('disable_event_daemon', False):
+        cherrypy.engine.subscribe('start', girder.events.daemon.start)
```
mgrauer: add a comment explaining why you don't do this, in this case

tests/cases/events_test.py
```diff
@@ -136,3 +139,14 @@ def callback(event):
...
+    def testDisableDaemon(self):
+        self.assertTrue(hasattr(events.daemon, 'start'))
+
+        # Disable the event daemon and reload the module
+        config.getConfig()['server']['disable_event_daemon'] = True
+        del sys.modules['girder.events']
+        eventsModule = importlib.import_module('girder.events')
+
+        self.assertFalse(hasattr(eventsModule.daemon, 'start'))
+        self.assertTrue(callable(eventsModule.daemon.trigger))
```
mgrauer: is it possible to run `testAsyncEvents` with `disable_event_daemon` set?  it would be nice to test the behavior in some way in the case of no async thread, as clients can still call daemon methods.

girder/events.py
```diff
@@ -266,4 +268,9 @@ def trigger(eventName, info=None, pre=None, async=False):
 
 _deprecated = {}
 _mapping = {}
-daemon = AsyncEventsThread()
+
+if config.getConfig()['server'].get('disable_event_daemon', False):
+    # If daemon is disabled, just use the current module to run those event synchronously
```
mgrauer: events

zachmullen:
> it might be better to raise an exception in the case where an async event is attempted when running without the async thread

In principal I agree, but there are core actions that call `girder.events.daemon.trigger`, and the `bind` part of things does not distinguish between binding synchronously or asynchronously, so I don't think that's possible. For instance, uploading a file would always raise an exception in that case. A compromise solution might be to print a warning message if someone has bound to one of these events that is normally triggered on the daemon, but is being triggered synchronously.

zachmullen:
I worded that badly: it's possible, and we could throw an exception in that case, but I think a warning would be a little better than just shutting down the request.

girder/utility/config.py
zachmullen: I noticed that the test execution was being affected by how I had my girder.local.cfg set up, which is not desirable, so I went ahead and fixed that issue here.

zachmullen:
@mgrauer I changed this around a bit. Added warning in the case you mentioned, and improved the testing.

tests/cases/events_test.py
```diff
@@ -136,3 +137,28 @@ def callback(event):
...
+                events.daemon.trigger('_test.event', None, callback)
+
+        with events.bound('_test.event',  '_test.handler', self._incrementWithResponse):
+            events.daemon.trigger('_test.event', {'amount': 2}, callback)
+
+        self.assertEqual(self.ctr, 3)
+        self.assertEqual(self.responses, ['foo'])
+
+        events.daemon.stop()
+        events.daemon = oldDaemon
```
mgrauer: Resetting state at the end makes me nervous.  Can you do the above testing work in a context manager and then reset the state outside the CM?

tests/cases/events_test.py
```diff
@@ -136,3 +138,25 @@ def callback(event):
             self.assertEqual(events.daemon.eventQueue.qsize(), 0)
             self.assertEqual(self.ctr, 3)
             self.assertEqual(self.responses, ['foo'])
+            events.daemon.stop()
+
+    @mock.patch.object(events, 'daemon', new=events.ForegroundEventsDaemon())
+    def testForegroundDaemon(self):
```
mgrauer: This works for the context manager, thanks.

One last worry is that this doesn't actually test that we are using the ForegroundEventsDaemon for events.daemon. Am I right to worry about that?  Couldn't this test code also work if the standard async daemon is used?

I guess it comes down to trusting that the mock.patch works.  Or, trust, but verify.  Perhaps just a type check of events.daemon, that it is a ForegroundEventsDaemon?

zachmullen:
@mgrauer any other issues or is this good to merge?

tests/cases/events_test.py
zachmullen: > Am I right to worry about that? Couldn't this test code also work if the standard async daemon is used?

It would create a race condition, and this test would probably fail intermittently if the mock was failing. I added the assertion that the mock was working, there's certainly no harm in that.

mgrauer:
LGTM.  I'm also getting a couple messages in the logs when I run the tests that tell me the right thing is occurring.  Hmmm, is that something we can test for?  

    WARNING: Handler "_test.handler" for event "_test.event" was triggered on the daemon, but is actually running synchronously.

PR #1643, opened 2016-10-18
Authorized third party upload plugin

zachmullen:
Still need to write tests, but the functionality is all here.


zachmullen:
This is now ready for review, but depends on #1647

zachmullen:
@bcd123 has seen the demo of this feature and gives his stamp of approval. Just needs code review now.

zachmullen:
@mgrauer in case you were interested in QAing.

girder/api/rest.py
```diff
@@ -227,6 +227,17 @@ def retVal(user, token):
         return retVal(user, token)
 
 
+def setCurrentUser(user):
```
mgrauer: @zachmullen 

Since you introduced this method, could you update [api/v1/user.py](https://github.com/girder/girder/blob/master/girder/api/v1/user.py) and [api/rest.py](https://github.com/girder/girder/blob/master/girder/api/rest.py) to use this method instead of the calls to 

    setattr(cherrypy.request, 'girderUser', user)

for consistency and for the benefit of hiding the use of `cherrypy` a bit?

Perhaps also hide cherrypy in [these lines](https://github.com/girder/girder/blob/master/girder/api/rest.py#L131-L132), while you're at it.

clients/web/src/templates/widgets/uploadWidgetMixins.pug
```diff
@@ -17,7 +17,7 @@ mixin g-upload-widget-browse-or-drop
     else
       |  Browse or drop files here
   .form-group.hide
-    if (parentType === 'file')
+    if (parentType === 'file' || !multiFile)
```
mgrauer: The upload widget says "files" in two places, but in this case, it should say "file" as only one is allowed.

screen shot 2016-11-13 at 1 49 06 pm

plugins/authorized_upload/web_client/templates/authorizeUpload.pug
```diff
@@ -0,0 +1,17 @@
+.g-body-title Authorize upload into folder #[i= folder.name()]
```
mgrauer: This UI flow strands the user in this authorize upload view with no way to get back to the folder they were in.

Could this be done as a modal, or else add a "return to folder" or "done" button?

screen shot 2016-11-13 at 1 49 44 pm

zachmullen:
@mgrauer I went ahead and did some general cleanup and improvements of the upload widget.

* Made the text more uniform, and only store it in one place rather than two. Also made it consistent in the single vs. multi-file case.
* Disallowed selection of multiple files via drag & drop when the widget is in single-file mode.
* Allowed drag and drop selection by dragging files from the chrome download bar. This enables a workflow I've wanted for a while, which is to download a file from somewhere on the web, and then upload it into girder without having to mess with any filesystem browsers.

zachmullen:
I can't reproduce the .docx issue. I can take a look on your machine tomorrow.

PR #1647, opened 2016-10-19
Upgrade to phantom 2.x

zachmullen:
1.9 is no longer supported, and segfaults on startup on the
latest version of OSX (Sierra).

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "eslint-plugin-underscore": "0.0.10",
     "istanbul": "1.1.0-alpha.1",
     "phantomjs": "^2.1.7",
-    "phantomjs-polyfill": "0.0.x",
     "webpack-dev-server": "^2.1.0-beta.0"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "eslint-plugin-standard": "^2.0.0",
     "eslint-plugin-underscore": "0.0.10",
     "istanbul": "1.1.0-alpha.1",
-    "phantomjs": "^2.1.7",
+    "phantomjs-prebuilt": "^2.1.13",
     "webpack-dev-server": "^2.1.0-beta.0"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "eslint-plugin-standard": "^2.0.0",
     "eslint-plugin-underscore": "0.0.10",
     "istanbul": "1.1.0-alpha.1",
-    "phantomjs": "1.9.x",
+    "phantomjs": "^2.1.7",
     "phantomjs-polyfill": "0.0.x",
     "webpack-dev-server": "^2.1.0-beta.0"
   },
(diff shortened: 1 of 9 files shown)


zachmullen:
@girder/developers PTAL, this is now ready.

jbeezley:
Most everything works, but I'm getting a timeout running the swagger tests locally.  I noticed a couple of related issues, but haven't been able to fix it yet:
* The bind polyfill should be removed from the [swagger mako template].
* On [travis](https://travis-ci.org/girder/girder/jobs/168969256#L2553), the swagger test says `0 of 0 passed`, which would indicate that the test ran before the spec was loaded, but this was also the case with the last run on master.

package.json
```diff
@@ -70,8 +70,7 @@
     "eslint-plugin-standard": "^2.0.0",
     "eslint-plugin-underscore": "0.0.10",
     "istanbul": "1.1.0-alpha.1",
-    "phantomjs": "1.9.x",
-    "phantomjs-polyfill": "0.0.x",
+    "phantomjs": "^2.1.7",
```
jbeezley: This package is deprecated, we should move to `phantomjs-prebuilt`, which has a newer version.  This will also require changing the executable path in the test runner.

zachmullen:
> On travis, the swagger test says 0 of 0 passed, which would indicate that the test ran before the spec was loaded, but this was also the case with the last run on master.

That's actually the number of assertions made using `expect`. The test itself is running, but it contains no assertions, instead using implicit assertions via `waitsFor`, so I think that's working as expected.

zachmullen:
@jbeezley that last commit fixed the swagger failure on my local machine. I also added an `expect` assertion so it wouldn't say 0 of 0 anymore 😄

jbeezley:
Actually, now that I took another look at this, I think there is a problem.  The generated testEnv looks like this:
```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>Girder test environment</title>
    <link href="../../../test/lib/jasmine-1.3.1/jasmine.css" rel="stylesheet">
    <script src="testing.min.js"></script>
    <link href="/clients/web/static/built/fontello/css/fontello.css" rel="stylesheet" type="text/css">
    <link href="/clients/web/static/built/girder.ext.min.css" rel="stylesheet" type="text/css">
    <link href="/clients/web/static/built/girder.app.min.css" rel="stylesheet" type="text/css">
  </head>
  <body>
    <div class="hide" id="g-global-info-apiroot">%HOST%/api/v1</div>
    <div class="hide" id="g-global-info-staticroot">%HOST%/static</div>
    <script src="/clients/web/static/built/girder.ext.min.js"></script>
    <script src="/clients/web/static/built/girder.app.min.js"></script>
    <script type="text/javascript">
      (function () {
        var consoleReporter = new jasmine.ConsoleReporter();

        window.jasmine_phantom_reporter = consoleReporter;
        jasmine.getEnv().addReporter(consoleReporter);

        girder.utilities.eventStream.settings.timeout = 10;
      })();
    </script>
  </body>
</html>
```

I think you need to move the `testing.min.js` below `girder.ext.min.js` for the jquery dependency.

zachmullen:
Yep, I fixed that a little while ago, just waiting on travis to get through the queue.

PR #1653, opened 2016-10-25
Explicitly enable Babel ES2015 transpilation

### LINKED ISSUE #1648: Rebuild Webcode error via /api/v1/system/web_build
Context: Extending girder/girder docker image to make one with custom plugins added.

Found that when I enabled plugins (even just one that's included with the base girder and then even when no plugins are enabled), the web_build endpoint would fail even when manually calling "npm run build" succeeds after ``docker exec -it <name> bash`` into the running container.

The following gist shows output from the same container using two methods that I'm assuming should be resulting in the same output:

[1] Calling ``http://localhost:8080/api/v1/system/web_build?progress=false&dev=false``
[2] Calling ``npm install --unsafe-perm && npm run build`` (what the girder Dockerfile does)

Gist: https://gist.github.com/Purg/6f49b7cf4587242ba920130e81bb4d80

Things I've noticed:
- The ``npm run build`` version looks like its running the ``webpack:dev`` task vs. the web_build call running the ``webpack:prod`` task.

zachmullen:
Fixes #1648. This preset was enabled by default on some platforms,
but not others, including in our docker container. This fixes the
production build on those platforms.

This also changes the docker container to build the web client
code in production mode.

@Purg PTAL and let me know if this fixes your issue with the docker container.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -9,7 +9,7 @@
       "cover": {
         "plugins": [
           ["istanbul", {
-            "exclude": ["**/*.pug", "**/*.jade"]
+            "exclude": ["**/*.pug", "**/*.jade", "node_modules/**/*"]
           }]
         ]
       }
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ COPY Gruntfile.js /girder/Gruntfile.js
 COPY setup.py /girder/setup.py
 COPY package.json /girder/package.json
 COPY README.rst /girder/README.rst
+COPY .babelrc /girder/.babelrc
 
 RUN pip install -e .[plugins]
 RUN npm install --production --unsafe-perm
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ COPY package.json /girder/package.json
 COPY README.rst /girder/README.rst
 
 RUN pip install -e .[plugins]
-RUN npm install --unsafe-perm
-RUN npm run build
+RUN npm install --production --unsafe-perm
+RUN npm run build -- --env=prod
 
 ENTRYPOINT ["python", "-m", "girder"]
--- a/grunt_tasks/webpack.config.js
+++ b/grunt_tasks/webpack.config.js
@@ -103,7 +103,10 @@ module.exports = {
             {
                 test: /\.js$/,
                 loader: 'babel-loader',
-                exclude: [paths.node_modules]
+                exclude: [paths.node_modules],
+                query: {
+                    presets: ['es2015']
+                }
             },
             // JSON files
             {
@@ -136,7 +139,12 @@ module.exports = {
             {
                 test: /\.(pug|jade)$/,
                 loaders: [
-                    'babel-loader',
+                    {
+                        loader: 'babel-loader',
+                        query: {
+                            presets: ['es2015']
+                        }
+                    },
                     'pug-loader'
                 ]
             },
--- a/grunt_tasks/webpack.config.js
+++ b/grunt_tasks/webpack.config.js
@@ -104,7 +104,7 @@ module.exports = {
             {
                 test: /\.js$/,
                 loader: 'babel-loader',
-                exclude: [paths.node_modules],
+                exclude: /node_modules/,
                 query: {
                     presets: [es2015Preset]
                 }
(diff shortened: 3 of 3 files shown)


zachmullen:
ping @nholthaus since you ran into the same issue.

jbeezley:
There is a chance this is also affecting me in https://github.com/DigitalSlideArchive/large_image/pull/111.  At one point, I was seeing a syntax error while loading the javascript bundle in the test runner.

jbeezley:
I get a build error in the large_image plugin.  It's looking for the preset relative to the plugin directory rather than girder's.

https://travis-ci.org/DigitalSlideArchive/large_image/builds/170502928#L3905

jbeezley:
The commit I just pushed fixes both the build and the test in my large_image branch.

zachmullen:
Great! Now we just have to figure out why this is happening on travis:

```
ERROR: TypeError: undefined is not an object (evaluating 'root._')
53: TRACE:
53:  -> http://localhost:30006/clients/web/static/built/girder.ext.min.js: 164
53:  -> http://localhost:30006/clients/web/static/built/girder.ext.min.js: 1717
53:  -> http://localhost:30006/clients/web/static/built/girder.ext.min.js: 53 (in function "__webpack_require__")
53:  -> http://localhost:30006/clients/web/static/built/girder.ext.min.js: 40610
53:  -> http://localhost:30006/clients/web/static/built/girder.ext.min.js: 53 (in function "__webpack_require__")
53:  -> http://localhost:30006/clients/web/static/built/girder.ext.min.js: 63742
53:  -> http://localhost:30006/clients/web/static/built/girder.ext.min.js: 53 (in function "__webpack_require__")
53:  -> http://localhost:30006/clients/web/static/built/girder.ext.min.js: 138
53: 
53:   phantomjs://code/specRunner.js:143 in onError
```

I can't reproduce it locally...

sebastienbarre:
The `var es2015Preset = require.resolve('babel-preset-es2015');` is puzzling me? Where did you find that trick, @jbeezley? I see you are now referencing `es2015` explicitly in `webpack.config.js` now, is`.babelrc` being ignored?

jbeezley:
Google-fu

http://stackoverflow.com/questions/34574403/how-to-set-resolve-for-babel-loader-presets

zachmullen:
`.babelrc` was not being respected on some platforms (not sure why). It happened in our docker build and also inside someone's ubuntu VM.

jbeezley:
The fun part is that this fixes the tests in my plugin, but breaks the client tests here.

jbeezley:
I wonder it it matters when the plugin is symlinked in rather than copied?

zachmullen:
Ok, everything seems in order now. @jbeezley does the head of this branch still work for large_image?

sebastienbarre:
I was about to suggest "are we copying .babelrc?" over, but you took care of it, @zachmullen. This makes more sense now. My gut feeling would be to revert the explicit mention of `es2015` in `webpack.config.js` so that it does not become out of sync with the presets in `.babelrc`. Or I could see some headaches down the road...

zachmullen:
Well, I thought that too, but there's still the issue that @jbeezley was experiencing on a downstream plugin, which I don't think was related to the docker environment.

jbeezley:
Yeah
```
ERROR in /home/travis/build/DigitalSlideArchive/large_image/web_client/main.js
Module build failed: Error: Couldn't find preset "es2015" relative to directory "/home/travis/build/DigitalSlideArchive/large_image/web_client"
```
so `.babelrc` isn't in the directory hierarchy.  That's the problem, I think.  A lot of people symlink plugins in place like this currently.

jbeezley:
This branch still works with `large_image`.

zachmullen:
Ok, just needs a code review.

sebastienbarre:
I read it differently:
```
Module build failed: Error: Couldn't find preset "es2015" relative to directory "/home/travis/build/DigitalSlideArchive/large_image/web_client"
```
doesn't that mean that it knows `es2015` is required (maybe by finding it in `.eslintrc`) but can't resolve it? A problem that you solved using `require.resolve('babel-preset-es2015');`, right? Or are you confident it only knows about `es2015` because it was found in the now modified `webpack.config.js`?

jbeezley:
> Or are you confident it only knows about es2015 because it was found in the now modified webpack.config.js?

That is my best guess.  It could be unrelated, but it seems as if babel is using the "real" path of the file's directory rather than the symlinked path.

zachmullen:
Thanks for the help and review on this issue @jbeezley @sebastienbarre

PR #1655, opened 2016-10-27
Fixes #1652. Put girder core web client code into common chunk

### LINKED ISSUE #1652: use different strategy to create common Webpack chunk
Creating an issue (for myself mostly) to revisit the Webpack strategy used to create the common Webpack chunk. This was mentioned in the original #1476 Webpack PR (copied below). This is as yet to be investigated, but it seems to have adverse affect on CI, as exposed by @jbeezley in DigitalSlideArchive/large_image#111:
- There still appears to be too much duplicated code going in `girder.main.min.js` and each `plugin.min.js`; this needs to be investigated and fixed by myself (Seb; consider [robertknight/webpack-bundle-size-analyzer](https://github.com/robertknight/webpack-bundle-size-analyzer)), also consider the `DllPlugin`, which would increase performance as well ([Webpack Plugins we been keepin on the DLL](https://medium.com/@soederpop/webpack-plugins-been-we-been-keepin-on-the-dll-cdfdd6cb8cd7), [Optimizing Webpack for Faster React Builds](http://engineering.invisionapp.com/post/optimizing-webpack/), [Optimizing Webpack build times and improving caching with DLL bundles](https://robertknight.github.io/posts/webpack-dll-plugins/), [webpack/examples/dll](https://github.com/webpack/webpack/tree/master/examples/dll)). Also check [trivago/parallel-webpack: Builds multi-config webpack projects in parallel](https://github.com/trivago/parallel-webpack)

zachmullen:
This fixes the duplication of code that we saw inside the plugin
webpack targets. This commit also fixes a small bug in the jobs
web client plugin code.

ping @jbeezley @sebastienbarre

### DIFF
--- a/plugins/jobs/web_client/main.js
+++ b/plugins/jobs/web_client/main.js
@@ -1,3 +1,4 @@
+import $ from 'jquery';
 import './routes';
 
 // Extends and overrides API
--- a/grunt_tasks/webpack.config.js
+++ b/grunt_tasks/webpack.config.js
@@ -60,30 +60,9 @@ module.exports = {
         filename: '[name].min.js'
     },
     plugins: [
-        // See http://stackoverflow.com/a/29087883/250457
-        // TODO: there is still too much code going in each plugin.min.js
-        // new webpack.optimize.CommonsChunkPlugin({
-        //     name: 'girder.app',
-        //     minChunks: function (module) {
-        //         var include = module.resource &&
-        //             module.resource.indexOf(paths.clients_web) !== -1;
-        //         if (include) {
-        //           // console.log('[girder.app] <=', module.resource.replace(paths.node_modules, ''));
-        //         }
-        //         return include;
-        //     }
-        // }),
         new CommonsChunkPlugin({
-            name: 'girder.ext',
-            minChunks: function (module) {
-                var include = module.resource &&
-                    module.resource.indexOf(paths.clients_web) === -1 &&
-                    module.resource.indexOf(paths.plugins) === -1;
-                if (include) {
-                  // console.log('[girder.ext] <=', module.resource.replace(paths.node_modules, ''));
-                }
-                return include;
-            }
+            names: ['girder.app', 'girder.ext'],
+            minChunks: 2
         }),
         // Automatically detect jQuery and $ as free var in modules
         // and inject the jquery library. This is required by many jquery plugins


plugins/jobs/web_client/main.js
```diff
@@ -1,3 +1,4 @@
+import $ from 'jquery';
```
zachmullen: We actually use jQuery in this file, see line 8.

I'm actually not sure we even need that block where we add `window.girder.plugins.jobs`. Was there a reason to have that? If it's no longer needed we can remove it and also this import.

sebastienbarre:
Eh, I don't know about this change @zachmullen, this is not doing the right thing, in my opinion. 

Originally the idea was for `girder.ext.min.js` to store all the third-party libraries, e.g. jQuery & all, and that's what the previous code was meant to do, by instructing Webpack: "if I'm requiring a module that is outside my own source tree (outside `paths.clients_web` or plugins), then consider it third-party and put it in `girder.ext`)". In that way, you do not have to explicitly list all your third-parties, you "just" tell Webpack where your own code is, and the rest is considered external. The end-goal here was for `girder.ext.min.js` not to change "too often", as opposed to `girder.app.min.js`, our own code, which is likely to evolve faster. This strategy would have helped with caching the corresponding javascript file, etc. In that context, `girder.ext` is meant to be the "commons" chunk in `CommonsChunkPlugin`. 

Before your change, I would see:
```
                                 girder.app.min.js     372 kB       0  [emitted]  girder.app
                                 girder.ext.min.js    1.74 MB      19  [emitted]  girder.ext
```
but now:
```
                                 girder.app.min.js    1.35 MB       0  [emitted]  girder.app
                                 girder.ext.min.js     269 kB      19  [emitted]  girder.ext
```
meaning at least some, if not all the third-party code is now likely in `girder.app`, right?

I'm not sure what `CommonsChunkPlugin` does when you use multiple commons chunks in `names`. If the `name` (singular) of the commons chunk is C, and you reference two entry points A, B in your config, and `minChunks` is 2, then if A and B both use a module M, then M is moved to the commons chunk C (per "minChunks: The minimum number of chunks which need to contain a module before it’s moved into the commons chunk"). Now what happen if you have two commons chunk is unclear to me, but since you listed `girder.app` first in `names`, it might explain why it is taking in most of the "common code", as opposed to `girder.ext`. 

Either way, this strategy, from what my experimentation, does not guarantee that you end up with a commons chunk (`girder.ext`) that has all your third-party code, because if only *one* of your entry point uses a third-party module, then that module stays "there", it is not being moved to your commons chunk, and you end up mixing third-party code with your own.

If all of the above is moot, then I would suggest not bothering with either `girder.app` or `girder.ext`, but just generate one big chunk.

plugins/jobs/web_client/main.js
sebastienbarre: Yes it could go away I think, it was mostly to be able to access it from the console...

zachmullen:
If there is a better way to control what modules get put in what files, I'm all ears, my only care is that module code doesn't get duplicated unnecessarily in multiple files, which this solves. If the CommonsChunk plugin is only going to support a single common chunk rather than providing granular control of what modules go in what files, then I'm fine with just combining into one single big common chunk -- the historical reasons for having two separate files no longer really matter in our new build infrastructure.

sebastienbarre:
In other words, for ease of deployment, I would not bother with several chunks. For ease of development on the other hand, to make it easier on our devs, I would use `DllPlugin` as suggested in #1652, where Webpack would *first* bundle all the third-party code and create a manifest file. It would then bundle all your app code in a second pass and resolve the references to the third-party modules using the manifest. That second pass would be the only pass executed during development, because that's the code devs would modify, not the third-party -- that's the code that would be under "watch" by Webpack. But from my preliminary reading it would require the first bundle (`girder.ext`) to explicitly list which third-party module we need to put there, as opposed to letting Webpack decide either based on path (like it was before), or on the number of times a module is being required before it is flagged as being "common" (your change).

zachmullen:
Cool, that sounds good. I'll merge this for now since it does fix the immediate issue, and we can plan for a more robust solution in the future, perhaps using the DLL plugin.

PR #1659, opened 2016-10-31
Export plugins under `girder.plugins`

jbeezley:
This is the only way I could find to make plugins available globally.  Any method using dynamic require or the context module api will result in all plugins being built.

Plugins can export symbols by defining a file called `web_client/index.js`.  Currently only the `jobs` plugin does this, but we should add exports for the other plugins as well.  I can do that here if desired.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -9,3 +9,4 @@ clients/web/static/
 clients/web/test/
 build/
 _build/
+clients/web/src/plugins.js
--- a/clients/web/src/.gitignore
+++ b/clients/web/src/.gitignore
@@ -1,1 +1,2 @@
 version.js
+plugins.js
--- a/clients/web/src/index.js
+++ b/clients/web/src/index.js
@@ -9,6 +9,7 @@ import * as server from './server';
 import * as utilities from './utilities';
 import * as version from './version';
 import * as views from './views';
+import * as plugins from './plugins';
 import events from './events';
 import router from './router';
 
@@ -20,6 +21,7 @@ export {
   events,
   misc,
   models,
+  plugins,
   rest,
   router,
   server,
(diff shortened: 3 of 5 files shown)


zachmullen:
Is there an easy-ish way to build these statements into each plugin.min.js file rather than into the core javascript? If it's not straightforward, this is OK, but long-term, my hope is that we can have no plugin references in the core code, even if there were plugins specified at build time.

jbeezley:
None of the plugin code will go in the main bundle, it is only a reference AFAIK.  I can't really think of a good way to do something similar by injecting code into the plugin.  It would have to be something like what was done in the jobs plugin.  I don't really like that because it defers attaching symbols to the `girder.plugins` object until after the page load.  That is bound to lead to problems.

Perhaps @sebastienbarre can have a quick look?

zachmullen:
Ah, I guess that's true now that the `window.girder` references are no longer used. LGTM

sebastienbarre:
Ha, auto-generated files in the source tree, I find that always risky...

Anyway, this above brought all the jobs plugin code in the main chunks, because of the import, but that was to be expected.
Before:
```
                    plugins/jobs/plugin.min.js    19.5 kB       2  [emitted]  plugins/jobs/plugin
                                 girder.ext.min.js     269 kB      19  [emitted]  girder.ext
                                girder.app.min.js    1.35 MB       0  [emitted]  girder.app
```
after:
```
                        plugins/jobs/plugin.min.js    1.28 kB      17  [emitted]  plugins/jobs/plugin
                                 girder.ext.min.js     269 kB      19  [emitted]  girder.ext
                                 girder.app.min.js    1.37 MB       0  [emitted]  girder.app
```

sebastienbarre:
Difficult to google for the right way to generate plugins with Webpack, since you get all the results related to webpack plugins themselves...

jbeezley:
Yeah, I don't like either, but we were already doing it for `version.js`...

Anyway, it looks like this isn't doing exactly what I was intending.  I would be okay with backing out the changes.

PR #1660, opened 2016-10-31
Add 'mimeType' attribute to the 'linkUrl' file objects.

Xarthisius:
This PR is complementary to PR #1623.

### DIFF
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -107,7 +107,8 @@ class File(Resource):
             return self.model('file').filter(
                 self.model('file').createLinkFile(
                     url=params['linkUrl'], parent=parent, name=params['name'],
-                    parentType=parentType, creator=user, size=params.get('size')), user)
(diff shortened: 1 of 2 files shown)


zachmullen:
So this stores the MIME type if it's passed, but it doesn't look like that value is read anywhere. Where would you expect this info to be used?

Xarthisius:
@zachmullen I'm not sure I follow, where's `mimeType` supposed to be read?

zachmullen:
In this changeset, you made it so we can store this value in the database, but it looks like we don't actually use it anywhere as far as controlling any sort of behavior. So I was just curious why you wanted it there. I'm guessing you have some plugin that is reading this value for some purpose?

Xarthisius:
oh! yes, that's gonna be used not even in a plugin but in an external girder interface.

zachmullen:
Ah OK.

PR #1665, opened 2016-11-04
Dynamic linking of webpack bundles

### LINKED ISSUE #1652: use different strategy to create common Webpack chunk

zachmullen:
Fixes #1652


zachmullen:
* `girder.ext.min.js` no longer exists, we just have one core library bundle containing all the dependencies.
* Because of a detail of how DllPlugin works, bundle names that use the DllPlugin cannot contain the . character. The core library bundle is now called `girder_lib.min.js` and exposes a global var called `girder_lib` that works much like the old library. It is a webpack DLL object with a manifest built alongside it, which enables really fast incremental builds of bundles that want to link against it dynamically, e.g. girder built-in plugins that modify the main web client app.
* There is an additional new file called `girder_app.min.js` that is just the actual entrypoint that starts up the girder app on the HTML page body. It dynamically links against the girder_lib bundle.
* Each plugin file now dynamically links against the girder_lib bundle, so they are as lightweight as possible and can build incrementally, e.g. `npm run build -- --plugins=jobs webpack:plugin_jobs`.
* The magical registrations of plugins containing index.js under the `girder.plugins` global object has been removed since it violated our bundle separation designs. Bundles must once again explicitly register themselves under that namespace if their symbols need to be exposed for, e.g. testing. A future improvement could be to also bundle plugins themselves as Dlls so that downstream plugins could link against them, but that's out of scope for this PR.

zachmullen:
I think I broke `watch`, that should be fixed before this gets merged.

jbeezley:
> The core library bundle is now called girder_lib.min.js and exposes a global var called girder_lib that works much like the old library.

Girder 3.0? :trollface:

zachmullen:
Heh, if it really matters to people, we can re-add some symlink rules so that `girder.ext.min.js` -> `girder_lib.min.js` and `girder.app.min.js` -> `girder_app.min.js`, though the contents of those files aren't really in good correspondence anymore.

zachmullen:
Also worth noting is that the top-level `window.girder` symbol is still in place just like before if you import `main.js`. For downstreams building their own bundles, it won't be there.

zachmullen:
Watch mode is now "working"--the limitation now is that you can only watch a single bundle at a time for each watch process. I'm actually OK with that since it covers the practical use cases with very fast response time.

1. If you're working on girder core and want to watch core:

    npm run watch

2. If you're working on a plugin and want to watch that plugin:

    npm run watch -- --plugins=foo,foo_dependency webpack:plugin_foo

If you are making changes in both core and a plugin, you can have two separate watch processes going to handle both, which would work fine as long as the core module interface wasn't changing.

I considered going down the route of exposing easier watch commands behind `girder-install web`, but for the moment I think we should wait on those until we've tried these out in the wild for a while to see if any problems arise.

AFAIC, this is now ready.

plugins/jobs/web_client/main.js
```diff
@@ -2,3 +2,8 @@ import './routes';
 
 // Extends and overrides API
 import './views/HeaderUserView';
+
+import { registerPluginNamespace } from 'girder/pluginUtils';
+import * as jobs from './index';
+
+registerPluginNamespace('jobs', jobs);
```
jbeezley: That works for me.

clients/web/src/main.js
```diff
@@ -1,3 +1,30 @@
...
+    };
+}
+
+// For testing and convenience, available now because of testUtils.js reliance on $
+window.$ = $;
+window._ = _;
+window.moment = moment;
+window.Backbone = Backbone;
+
+if (!window.disableGirderMainApp) {
```
jbeezley: I would prefer this stay in the mako template rather than the main bundle.  Is there a reason besides convenience for moving it here?

zachmullen: Library bundles aren't able to execute code at import time (i.e. they cannot have an entry point). So it seemed like a good idea to put this inside the main bundle, in the interest of having all of our javascript code in files rather than in the HTML directly.

What's the reason to prefer it in the mako template?

zachmullen:
@jbeezley I think everything is now fixed.

jeffbaumes:
@zachmullen is this new way of linking plugins documented? Is there a canonical externally-built (i.e. not within Girder) plugin that shows the right way to do things in 2.0 for people to go to?

zachmullen:
Fortunately, the details of the linking are actually not exposed to plugin developers. Plugins that are building against the main girder web client will not need to do anything different than they were doing before, except in the case of the `watch` command, which I'm hoping we can expose in a friendlier way on a future topic.

As for custom apps that want to use symbols from the girder library, @ronichoudhury is working on the first major example of that. The recommended way to build those is still via the normal static linking process, and the changes in question are mostly a matter of allowing people to augment the core webpack configuration rules, so this won't affect that case either.

jeffbaumes:
Great, thanks for explaining.

PR #1667, opened 2016-11-08
Bump frozen Pillow version for security patches

zachmullen:
This doesn't really affect our environment since the version in setup.py isn't pinned, but Gemnasium threw up an alert based on the requirements.txt file.

### DIFF
--- a/plugins/thumbnails/requirements.txt
+++ b/plugins/thumbnails/requirements.txt
@@ -1,3 +1,3 @@
-Pillow==3.1.0
+Pillow==3.4.2
 pydicom==0.9.9
 numpy==1.11.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ hachoir-metadata==1.3.3 ; python_version < '3.0'
 hachoir-parser==1.3.4   ; python_version < '3.0'
 
 # thumbnails
-Pillow==3.3.2
+Pillow==3.4.2
 
 # dependencies of celery
 amqp==1.4.9
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ hachoir-metadata==1.3.3 ; python_version < '3.0'
 hachoir-parser==1.3.4   ; python_version < '3.0'
 
 # thumbnails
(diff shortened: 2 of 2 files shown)


manthey:
Should this also be bumped in the thumbnails plugin?

I feel we should use >=(x.y),<(x+1) for requirements, rather than ==, so that we end up with the latest version that all of the plugins and core can work with, rather than them fighting it out.  I can't remember why we chose to pin the requirements exactly.

zachmullen:
We actually don't use these frozen version files in installations. The only time we read the requirements.txt files inside plugins is when using `girder-install plugin` on external plugins, which are allowed to specify their versions how they want. These files remain around for historical reasons that @jbeezley can probably talk about more accurately than I can.

manthey:
For .travis, we use `pip install -r requirements-dev.txt -e .[plugins,sftp] -e clients/python`, and that appears to take the pinned version for the plugins.

jbeezley:
For clarification, the pinned versions in `requirements.txt` are intended to create a repeatable testing environment.  This doesn't work unless it pins all packages (including dependencies of dependencies).

The intention of a requirements file is for it to be the output of `pip freeze` including everything necessary for testing.  In this case, it should contain everything in `requirements.txt`, `requirements-dev.txt`, and those required for the plugins tested by travis.  The complication and the reason this is not done is because we use the plugins' `requirements.txt` files in `girder-install`.

zachmullen:
@manthey that command installs the pinned version of `requirements-dev.txt`, but the setup.py versions of the girder and girder-client package requirements.

manthey:
Yes, but it uses the plugins' requirements.txt files.

zachmullen:
That command does not, the `[plugins]` extra uses the un-versioned specifiers defined [here](https://github.com/girder/girder/blob/master/setup.py#L75-L83) for the plugins.

manthey:
I see -- the pinned version is used in the packaging_plugin_install test, not in the initial install.  See for instance, https://travis-ci.org/girder/girder/jobs/174236779, line 171.

PR #1668, opened 2016-11-09
Access flags

zachmullen:
This change adds a new structure to the core permission model in Girder.

In addition to the `level` allowed on resources for groups and users, customized permission flags may now also be granted. These flags are orthogonal to the permission level, but users may only be granted flags if they have at least read access level on a document.

Girder core does not (at this time) expose any such flags, but plugins may add their own flags with semantics defined by the plugin itself. For example, this could be added in a plugin:

    from girder.constants import registerPermissionFlag

    registerPermissionFlag('execute', 'Execute analyses', description='Use items as analysis descriptors', admin=True)

Then, this flag would appear in the access control widget, and admin users could enable it on specific users and groups, or to the public if the resource is publicly visible. Then, if the user attempts to take an action that requires that flag, the plugin can enforce that behavior via

    self.model('foo', 'bar').requireAccessFlags(doc, user=self.getCurrentUser(), flags='execute'):

or as an argument to `loadmodel`:

    @loadmodel(model='item', level=AccessType.READ, requiredFlags='execute')


zachmullen:
This is now ready for review.

girder/models/model_base.py
```diff
@@ -736,7 +750,96 @@ def setPublic(self, doc, public, save=False):
...
+        :type doc: dict
+        :param flags: Flags or set of flags to add.
+        :type flags: flag identifier, or a list/set/tuple of them
+        :param user: The user performing this action.
+        :type user:
+        :param append: Whether to append to the list or replace it.
+        :type append: bool
+        :param save: Whether to save the document to the database afterward.
+        :type save: bool
+        :param force: Set this to True to disable validation of the current
```
mgrauer: I find this explanation confusing.

How about

`Set this to True if you explicitly want to circumvent validation of the passed in user's ability to set the flags.`

or even simpler

`Set this to True to set the flags regardless of the passed in user's permissions.`

girder/models/model_base.py
```diff
@@ -736,7 +750,96 @@ def setPublic(self, doc, public, save=False):
...
+    def _isFlagEnabled(self, doc, type, entry, flag):
+        """
+        Check whether a specific flag is enabled for a specific ACL entry in the document.
+        """
+        id = entry['id']
+        perms = doc.get('access', {})
+
+        if type == 'group':
+            return self._hasGroupAccessFlag(perms.get('groups', ()), id, flag)
+        else:  # 'user'
```
mgrauer: I find this comment surprising, for some reason I wouldn't have expected it from you here.  No action required.

😮

girder/models/model_base.py
```diff
@@ -736,7 +750,96 @@ def setPublic(self, doc, public, save=False):
...
+            return self._hasUserAccessFlag(perms.get('users', ()), id, flag)
+
+    def _validateFlags(self, doc, user, type, entry):
```
mgrauer: A bit of the same issue here with silently dropping disallowed flags.  This docstring is better for that case though.  Maybe you could amend the docstring to be 

```
Coerces a flag or set/list/tuple of flags into a valid form,
returning a list that only contains the valid flags that the passed in user has permission to set.
```

Could this be combined with setPublicFlags as there is a lot of similarity? Perhaps `force` and `append` make that hard to do.

girder/models/model_base.py
```diff
@@ -736,7 +750,96 @@ def setPublic(self, doc, public, save=False):
...
+        if user and user['admin']:
+            return list(flags)
+
+        allowedFlags = []
+        for flag in flags:
+            info = PERMISSION_FLAGS[flag]
+
+            # If this is an admin-only flag, we only allow it if it's already enabled
+            # for this specific group or user rule
+            if info['admin'] and self._isFlagEnabled(doc, type, entry, flag):
```
mgrauer: For setting flags, are there only `admin` and `non-admin` permission levels on flags?  This logic assumes such.

docs/plugin-development.rst
```diff
@@ -199,6 +199,44 @@ could access it using ::
 
 Where the second argument to ``model`` is the name of your plugin.
 
+Adding a custom permission flag
```
mgrauer: These are great docs.  Please add a discussion of the fact that a flag can be `admin` or not, and if there are any other permission levels for flags.

girder/models/folder.py
```diff
@@ -798,11 +799,18 @@ def setAccessList(self, doc, access, save=False, recurse=False, user=None,
         :param setPublic: Pass this if you wish to set the public flag on the
             resources being updated.
         :type setPublic: bool or None
+        :param publicFlags: Pass this if you wish to set the public flag list on
+            resources being updated.
+        :type publicFlags: list or None
```
mgrauer: flag identifier, or a list/set/tuple of them, or None

girder/models/model_base.py
```diff
@@ -806,9 +911,13 @@ def setGroupAccess(self, doc, group, level, save=False):
                      Set this to False if you want to wait to save the
                      document for performance reasons.
         :type save: bool
+        :param flags: List of special access flags to grant to the group.
```
mgrauer: call this `permission flags` instead of `special access flags` for consistency.

zachmullen:
>  It could be that there is a user/codebase divide on access vs permission, which is fine, but in code sometimes permission is used and sometimes access.

We actually have always used these two terms interchangeably, but you're right, I'll go through and refactor to just use "access" in this PR instead of "permission".

girder/models/model_base.py
zachmullen: Yes, just those two.

zachmullen:
@mgrauer I think this is ready for pass # 2.

mgrauer:
@zachmullen 

Starting pass 2.

I noticed you didn't mention public flags in the documentation, could you add that in?

girder/models/model_base.py
```diff
@@ -684,11 +717,10 @@ def _hasUserAccess(self, perms, userId, level):
                 return True
         return False
 
-    def _setAccess(self, doc, id, entity, level, save):
+    def _setAccess(self, doc, id, entity, level, save, flags=None, user=None):
```
mgrauer: by the same logic that `setPublicFlags` has a `force` param, this method should too

girder/models/model_base.py
```diff
@@ -703,10 +735,13 @@ def _setAccess(self, doc, id, entity, level, save):
...
         if level is not None:
-            doc['access'][entity].append({
+            entry = {
                 'id': id,
-                'level': level
-            })
+                'level': level,
+                'flags': flags
+            }
+            entry['flags'] = self._validateFlags(doc, user, entity, entry)
```
mgrauer: pass `force` to `_validateFlags`

girder/models/model_base.py
```diff
@@ -736,7 +771,97 @@ def setPublic(self, doc, public, save=False):
...
+        """
+        id = entry['id']
+        perms = doc.get('access', {})
+
+        if type == 'group':
+            return self._hasGroupAccessFlag(perms.get('groups', ()), id, flag)
+        else:  # 'user'
+            return self._hasUserAccessFlag(perms.get('users', ()), id, flag)
+
+    def _validateFlags(self, doc, user, type, entry):
```
mgrauer: by the same logic that `setPublicFlags` has a `force` param, this method should too

girder/models/model_base.py
```diff
@@ -736,7 +771,97 @@ def setPublic(self, doc, public, save=False):
...
+            # If this is an admin-only flag, we only allow it if it's already enabled
+            # for this specific group or user rule
+            if info['admin'] and self._isFlagEnabled(doc, type, entry, flag):
+                allowedFlags.append(flag)
+            elif not info['admin']:
+                allowedFlags.append(flag)
+
+        return allowedFlags
+
+    def setAccessList(self, doc, access, save=False, user=None):
```
mgrauer: by the same logic that `setPublicFlags` has a `force` param, this method should too

girder/models/model_base.py
```diff
@@ -791,7 +918,7 @@ def setAccessList(self, doc, access, save=False):
 
         return doc
 
-    def setGroupAccess(self, doc, group, level, save=False):
+    def setGroupAccess(self, doc, group, level, save=False, flags=None, user=None):
```
mgrauer: by the same logic that `setPublicFlags` has a `force` param, this method should too

girder/models/model_base.py
```diff
@@ -892,7 +1023,7 @@ def getFullAccessList(self, doc):
 
         return acList
 
-    def setUserAccess(self, doc, user, level, save=False):
+    def setUserAccess(self, doc, user, level, save=False, flags=None, currentUser=None):
```
mgrauer: by the same logic that `setPublicFlags` has a `force` param, this method should too

girder/models/model_base.py
```diff
@@ -907,9 +1038,57 @@ def setUserAccess(self, doc, user, level, save=False):
...
                      document for performance reasons.
         :type save: bool
+        :param flags: List of access flags to grant to the group.
+        :type flags: specific flag identifier, or a list/tuple/set of them
+        :param currentUser: The user performing this action. Only required if attempting
+            to set admin-only flags on the resource.
         :returns: The modified resource document.
         """
-        return self._setAccess(doc, user['_id'], 'users', level, save)
+        return self._setAccess(doc, user['_id'], 'users', level, save, flags, currentUser)
```
mgrauer: pass `force`

girder/models/model_base.py
```diff
@@ -806,9 +933,13 @@ def setGroupAccess(self, doc, group, level, save=False):
...
                      document for performance reasons.
         :type save: bool
+        :param flags: List of access flags to grant to the group.
+        :type flags: specific flag identifier, or a list/tuple/set of them
+        :param user: The user performing this action. Only required if attempting
+            to set admin-only flags on the resource.
         :returns: The updated resource document.
         """
-        return self._setAccess(doc, group['_id'], 'groups', level, save)
+        return self._setAccess(doc, group['_id'], 'groups', level, save, flags, user)
```
mgrauer: pass `force`

girder/models/model_base.py
```diff
@@ -747,39 +872,41 @@ def setAccessList(self, doc, access, save=False):
...
-                                               AccessType.WRITE,
-                                               AccessType.ADMIN):
+                if not userAccess['level'] in allowedLevels:
                     raise ValidationException('Invalid access level', 'access')
 
                 acList['users'].append({
                     'id': ObjectId(userAccess['id']),
-                    'level': userAccess['level']
+                    'level': userAccess['level'],
+                    'flags': self._validateFlags(doc, user, 'user', userAccess)
```
mgrauer: pass `force`

girder/models/model_base.py
```diff
@@ -747,39 +872,41 @@ def setAccessList(self, doc, access, save=False):
...
-                                                AccessType.WRITE,
-                                                AccessType.ADMIN):
+                if not groupAccess['level'] in allowedLevels:
                     raise ValidationException('Invalid access level', 'access')
 
                 acList['groups'].append({
                     'id': ObjectId(groupAccess['id']),
-                    'level': groupAccess['level']
+                    'level': groupAccess['level'],
+                    'flags': self._validateFlags(doc, user, 'group', groupAccess)
```
mgrauer: pass `force`

girder/models/model_base.py
```diff
@@ -907,9 +1038,57 @@ def setUserAccess(self, doc, user, level, save=False):
...
+        if user and user['admin']:
+            # Short-circuit the case of admins
+            return True
+
+        if not flags:
+            # Special case if no flags are passed
+            return True
+
+        if not isinstance(flags, (list, tuple, set)):
+            flags = {flags}
```
mgrauer: here `flags` is a set, in other similar places it is a list and a tuple.  same nit as above, though maybe this one has a better claim to be made a set.

zachmullen:
I think that's everything, ready for round 3

tests/cases/system_test.py
```diff
@@ -434,3 +456,144 @@ def testLogRoute(self):
             '=== Last 0 bytes of %s/info.log: ===\n\n' % logRoot)
 
         del config.getConfig()['logging']
+
+    def testAccessFlags(self):
```
mgrauer: can you add the following test cases?

 * acl_mixin hasAccessFlags
 * acl_mixin requireAccessFlags
 * setPublicFlags with a non-admin user where force=True
 * setAccessList with a non-admin user where force=True
 * setUserAccess with a non-admin user where force=True
 * setGroupAccess with a non-admin user where force=True
 * setPublicFlags passing in a single flag instead of a list/set/tuple
 * an admin-only access flag that is set by the admin on a group, and then set by a non-admin (the flag should still remain because it already exists on the group, you tested this case for user but not for group)
 * model_base hasAccessFlags called with flags=None

girder/models/model_base.py
```diff
@@ -892,7 +1029,8 @@ def getFullAccessList(self, doc):
 
         return acList
 
-    def setUserAccess(self, doc, user, level, save=False):
+    def setUserAccess(self, doc, user, level, save=False, flags=None, currentUser=None,
```
zachmullen: Since there is already a `user` parameter to `setUserAccess`, if we wanted consistency we'd have to change the one in `setGroupAccess` to be called `currentUser`. Is that OK?

clients/web/src/views/widgets/AccessWidget.js
```diff
@@ -133,11 +160,55 @@ var AccessWidget = View.extend({
...
+                    content: function () {
+                        return $(this).parent().find('.g-flags-popover-container').html();
+                    }
+                }).click(function () {
+                    $(this).popover('toggle');
+                });
+            }
+        });
+
+        _.each(this.$('.g-action-manage-public-flags'), el => {
```
mgrauer: can you repeat that comment from above here, in case these two blocks get separated?

zachmullen:
@mgrauer I couldn't reproduce that client-side error, does it persist after you do a fresh `girder-install web`?

girder/api/rest.py
```diff
@@ -345,6 +368,10 @@ def wrapped(*args, **kwargs):
                     raise RestException(
                         'Invalid %s id (%s).' % (model.name, str(id)))
 
+                if self.requiredFlags and not self.force:
```
mgrauer: Thanks for adding those other tests.  Can you add a test for the case where you require flags for the loadmodel decoration and also set force?  It is kind of a degenerate case, because the client would be saying "require this specific level of access, and also ignore access control", but I still want to check it works correctly.

    @loadmodel(model='user', level=AccessType.READ, force=True, requiredFlags='my_key')

zachmullen: Actually that is a weird case, I don't think that even makes sense. I'm more inclined to remove `and not self.force` from the line, which allows the user a little more power, albeit a very odd one I don't think anybody would want.

girder/models/model_base.py
```diff
@@ -806,9 +937,17 @@ def setGroupAccess(self, doc, group, level, save=False):
...
                      document for performance reasons.
         :type save: bool
+        :param flags: List of access flags to grant to the group.
+        :type flags: specific flag identifier, or a list/tuple/set of them
+        :param currentUser: The user performing this action. Only required if attempting
+            to set admin-only flags on the resource.
+        :type currentUser: dict or None
         :returns: The updated resource document.
+        :param force: Set this to True to set the flags regardless of the passed in
+            user's permissions (only matters if flags are passed).
```
mgrauer: I think it'll be clearer to change the docstring `user's` => `currentUser's`.  More for consistency with a parallel change below, where it matters more.

girder/models/model_base.py
```diff
@@ -901,15 +1041,65 @@ def setUserAccess(self, doc, user, level, save=False):
...
-                     Set this to False if you want to wait to save the
-                     document for performance reasons.
+            Set this to False if you want to wait to save the document for performance reasons.
         :type save: bool
+        :param flags: List of access flags to grant to the group.
+        :type flags: specific flag identifier, or a list/tuple/set of them
+        :param currentUser: The user performing this action. Only required if attempting
+            to set admin-only flags on the resource.
+        :param force: Set this to True to set the flags regardless of the passed in
+            user's permissions (only matters if flags are passed).
```
mgrauer: Change this docstring `user's` => `currentUser's`, because there is also a `user` param.

zachmullen:
🎉

mgrauer:
You have my ~~permission~~ access to merge.

🏆

PR #1673, opened 2016-11-11
Fix front-end bugs (undefined dereferences)

zachmullen:


zachmullen:
@mgrauer this fixes the two errors you noticed on data.kitware.com, PTAL

PR #1674, opened 2016-11-11
Circle ci config

zachmullen:
Wanted to open this as a PR just to test the hooks I just tried to configure on circle.


circle.yml
```diff
@@ -0,0 +1,34 @@
...
+    - pyenv rehash
+
+dependencies:
+  cache_directories:
+    - "~/cache"
+  override:
+    - pyenv rehash
+    - sudo apt-get update; sudo apt-get install libgif-dev
+    - ./scripts/install_cmake.sh
+    - npm install; npm update; npm prune
```
zachmullen: Do we need this step? `girder-install web` will run `npm install` for us.

jbeezley:
This is ready now.  There are a few differences from travis that developers should be aware of:

1. Every line in the circle configuration file is run in a different shell, so you can't set an environment variable or change directories and have it persist.  (It is possible to append lines to `~/.circlerc` for this purpose, but that is undocumented.)
2. Commands in the "test" section are by default only run on the first container.  You have to use the [`parallel`](https://circleci.com/docs/parallel-manual-setup/) directive for them to run on all containers.
3.  Getting cmake to detect the correct installed python scripts is a bit tricky.  It's best to pass the full path (`pyenv which flake8`) and to use `pyenv exec` to execute commands in the shell.
4.  All databases, services, and programming languages are available in all containers whether you want them or not.

There are also a number of improvements from travis:

1.  `sudo` can be used in all containers without restriction.
2.  They have an Ubuntu 14.04 container available (which this config uses).
3.  You can store and serve build artifacts just by copying files to `$CIRCLE_ARTIFACTS`.  I have added the html coverage output as build artifacts so you can [browse them](https://103-17380122-gh.circle-artifacts.com/2/tmp/circle-artifacts.v7CBZqu/coverage/js/index.html) there rather than at codecov.  This also allows you to see coverage for each environment (python 2 vs 3) separately rather than combined as it is on codecov.
4.  You can ssh into the test containers directly.
5.  The build environment claims to have 16 cpu cores and about 50GB of memory.

The one thing I removed in this PR is testing under mongo 2.6.  This is mostly out of laziness because the containers already have mongo 3.0 running, but I can add it back if anyone thinks it is important.

circle.yml
jbeezley: I removed the `install`, but the `update` and `prune` steps are technically necessary to get a clean environment.  Subsequent `npm install` executions only take a few seconds.

mgrauer:
We claim Girder runs on [Mongo 2.6+](http://girder.readthedocs.io/en/latest/prerequisites.html) so I think it's important to be tested on 2.6.

zachmullen:
@mgrauer since this is blocking all of our topics from getting merged, my vote would be that we merge this and figure out how to test on mongo 2.6 on a future topic.

mgrauer:
works for me

zachmullen:
Great job @jbeezley , thanks for your quick response on this.

msmolens:
Great, CircleCI looks like an improvement. What necessitated the quick switchover?

Also, there are lingering references to Travis, at least some of which should definitely be updated if CircleCI is the sole CI provider for Girder:

    $ git grep -c -i travis
    CONTRIBUTING.md:6
    README.rst:2
    cmake/travis_continuous.cmake:4
    devops/ansible/roles/girder/README.md:1
    girder/utility/system.py:1
    scripts/vmtest.py:1

zachmullen:
Travis recently upgraded their infrastructure, and in so doing caused our python builds to consistently run out of disk space partway through the tests. Circle is a definite improvement over travis, so this was an opportune time to switch over.

Good catch on those references, I'll remove those.

PR #1676, opened 2016-11-14
Fix docker build

jbeezley:
The command `npm run build -- --env=prod` still tries to run development tasks.  See the build on [dockerhub](https://hub.docker.com/r/girder/girder/builds/bvvjr9rpzwi9nsvzzj7dhlm/).


zachmullen:
Might be nice to have a `--no-plugins` option to girder-install web. I'll add that on another topic, though.

PR #1677, opened 2016-11-14
Fix warnings from multiple errorReponse bindings to the same code

zachmullen:
This fixes warnings like the following at server startup time:

    WARNING: Error response for code '400' is already defined (old: '['Path is invalid.']', new: 'Path refers to a resource that does not exist.')

We just add some structure to the description text to support multiple description bindings to the same error code. Swagger-ui renders these as markdown, so they appear as separate paragraph elements.

### DIFF
--- a/girder/api/describe.py
+++ b/girder/api/describe.py
@@ -268,14 +268,11 @@ class Description(object):
             reason = '\n\n'.join(reason)
 
         if code in self._responses:
-            print(TerminalColor.warning(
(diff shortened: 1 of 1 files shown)


mgrauer:
Warnings have gone away, and multiple response codes are concatenated.

See e.g. `/resource/lookup`

before the fix

screen shot 2016-11-15 at 11 24 32 am

after the fix

screen shot 2016-11-15 at 11 24 43 am

PR #1678, opened 2016-11-14
Support anonymous sftp when check_auth_none is not called

mgrauer:
The Cyberduck app doesn't pass the password separately (or maybe doesn't pass anything) when trying to connect with anonymous sftp, as `check_auth_password` is never called.

This fix allows Cyberduck to use anonymous sftp.


mgrauer:
@zachmullen I didn't feel good without this being tested.

:face_with_thermometer:

PR #1681, opened 2016-11-16
Add methods to set apiRoot and staticRoot

ronichoudhury:
Setting these values via special elements in application code is unusual and unintuitive. This PR adds `setApiRoot()` and `setStaticRoot()` functions to the REST module, and sets reasonable defaults for these values when the elements are missing and the functions aren't called. The magic elements are still queried so this change is fully backward compatible.

### DIFF
--- a/clients/web/test/spec/setApiSpec.js
+++ b/clients/web/test/spec/setApiSpec.js
@@ -10,11 +10,9 @@ describe('Test setApiRoot() and setStaticRoot() functions', function () {
         });
 
         runs(function () {
-            var host = 'http://localhost:30019';
-
             // Test the default values.
-            expect(girder.rest.apiRoot).toBe(host + '/api/v1');
-            expect(girder.rest.staticRoot).toBe(host + '/static');
+            expect(girder.rest.apiRoot.slice(girder.rest.apiRoot.indexOf('/', 7))).toBe('/api/v1');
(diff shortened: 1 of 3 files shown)


clients/web/test/spec/setApiSpec.js
```diff
@@ -0,0 +1,28 @@
...
+girderTest.startApp();
+
+describe('Test setApiRoot() and setStaticRoot() functions', function () {
+    it('Check for default values and mutation', function () {
+        waitsFor(function () {
+            return $('.g-frontpage-body').length > 0;
+        });
+
+        runs(function () {
+            var host = 'http://localhost:30019';
```
danlamanna: I'm fuzzy on how we actually assign ports to tests, but will this break the next time someone adds a test, or something equally as brittle?

ronichoudhury: If there's a `port` variable available in the testing environment I should use that. Anyone know how to get to it?

manthey: I don't think you need to know it.  Just `expect(girder.rest.apiRoot.slice(girder.rest.apiRoot.indexOf('/', 7))).toBe('/api/v1')`

ronichoudhury: Nice :nerd_face:

PR #1683, opened 2016-11-16
Fixes #1682. Exceeding data quota should not be a 500 error

### LINKED ISSUE #1682: Upload that exceeds quota throws 500 error
The upload widget gives me a nice error message and no bytes are transferred, but this should be a different error type.

screen shot 2016-11-16 at 1 28 51 pm

I saw this on authorized upload, but it happens any time the upload widget is used.

zachmullen:
ping @mgrauer

### DIFF
--- a/plugins/user_quota/plugin_tests/user_quota_test.py
+++ b/plugins/user_quota/plugin_tests/user_quota_test.py
@@ -69,7 +69,7 @@ class QuotaTestCase(base.TestCase):
             public=True, creator=self.admin)
 
     def _uploadFile(self, name, parent, parentType='folder', size=1024,
-                    error=None, partial=False):
+                    error=None, partial=False, validationError=None):
         """
         Upload a random file to an item.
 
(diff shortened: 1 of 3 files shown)


manthey:
Would a 413 error code be appropriate?

zachmullen:
I'm leaning towards no; 413 typically indicates to user agents that the request body of the just-sent request is too large. In this case, that's not what happened. The body of the request that was just sent is fine, this simply tells the client that the process they are about to start would not be accepted.

Handling of status codes at that level of granularity is typically left to the guts of the user agent rather than being exposed to the users themselves, and user agents are not aware of girder-level policy semantics.

mgrauer:
@zachmullen back at you for QA.  Your changes LGTM.  I fixed an issue in the config view where there should be whitespace but wasn't.  e.g. "individualusers".

before

screen shot 2016-11-16 at 5 17 41 pm

after

screen shot 2016-11-16 at 5 21 11 pm

zachmullen:
Good catch. LGTM.

PR #1687, opened 2016-11-29
Only keep EventStream open if the window is visible

zachmullen:
@brianhelba this addresses the issue you saw with many girder tabs open in chrome with the event stream socket causing new connections to girder to block. Now, once a tab is backgrounded, the event stream is shut off after 5 seconds. Once the tab is visible again, the stream is re-opened.

### DIFF
--- a/clients/web/src/utilities/EventStream.js
+++ b/clients/web/src/utilities/EventStream.js
@@ -66,6 +66,7 @@ prototype.open = function () {
             }
             stream.trigger('g:event.' + obj.type, obj);
         };
+        this._stopHeartbeat = false;
         this._heartbeat();
     } else {
         console.error('EventSource is not supported on this platform.');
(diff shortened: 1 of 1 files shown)


manthey:
I suspect the problem is when the `requestAnimationFrame` callback is called, it calls `window.clearTimeout` with a value that will be undefined initially.  WebKit (at least the version used by phantomjs), expects an integer as an argument to this, though it will ignore values that aren't valid window timeout ids.  The solution might just be to change ` window.clearTimeout(this._closeTimeout);` to ` window.clearTimeout(this._closeTimeout || 0);`.  If this fixes the problem, I'd check to see if we clear any other timeouts where the value might not be an integer.

zachmullen:
Added some debug statements, it turns out that `requestAnimationFrame` isn't being called in phantom, which makes a lot of sense actually since it's a headless browser...

I'll figure out some way to polyfill it or workaround it.

zachmullen:
Turns out I was following a red herring. `requestAnimationFrame` works just fine in phantom, the test was finding a legitimate bug in the code which I fixed on 6ff7747.

zachmullen:
@brianhelba any time to review this?

PR #1689, opened 2016-11-30
Add user-friendly watch commands for web code

zachmullen:
As promised, here are some handy watch commands built into `girder-install web` now that the process has stabilized.

ping @fogleman

### DIFF
--- a/girder/utility/install.py
+++ b/girder/utility/install.py
@@ -155,6 +155,13 @@ def runWebBuild(wd=None, dev=False, npm='npm', allPlugins=False, plugins=None, p
                             (' '.join(cmd), proc.returncode))
 
 
+def _runWatchCmd(*args):
+    try:
+        subprocess.Popen(args, cwd=constants.PACKAGE_DIR).wait()
+    except KeyboardInterrupt:
+        pass
+
+
 def install_web(opts=None):
     """
     Build and install Girder's web client. This runs `npm install` to execute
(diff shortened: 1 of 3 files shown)


zachmullen:
@girder/developers any plugin developers interested in QAing this small change? I've been using it on some plugin development and it's made life a lot easier.

girder/utility/install.py
```diff
@@ -277,6 +290,10 @@ def main():
     web.add_argument('--all-plugins', action='store_true',
                      help='build all available plugins rather than just enabled ones')
     web.add_argument('--plugins', default='', help='comma-separated list of plugins to build')
+    web.add_argument('--watch', action='store_true',
+                     help='watch for changes and rebuild girder core library in dev mode')
```
mgrauer: what does `dev mode` here and below mean?  I don't see that this functions differently with a dev flag or that it only works with a dev flag.

zachmullen: It means that the web client build process is built in dev mode instead of prod, since `watch` is an operation that only really happens in a dev environment.

* dev mode: source maps are built, no minification postprocessing
* prod mode: no source maps, includes minification step

girder/utility/install.py
```diff
@@ -155,13 +155,26 @@ def runWebBuild(wd=None, dev=False, npm='npm', allPlugins=False, plugins=None, p
...
+def _runWatchCmd(*args):
+    try:
+        subprocess.Popen(args, cwd=constants.PACKAGE_DIR).wait()
+    except KeyboardInterrupt:
+        pass
+
+
 def install_web(opts=None):
     """
     Build and install Girder's web client. This runs `npm install` to execute
```
mgrauer: Can you update this docstring?  `npm install` is no longer right, at least not on it's own, and maybe it's better to not mention it here.  Also, under certain conditions this does not build and install Girder's web client (when watching a plugin).  I'm not sure of a good docstring, maybe

   Build and install Girder's web client, or perform a watch on the web client or a specified plugin.

Otherwise I tested this out for main client and a plugin, it works for js, stylus and pug files.

PR #1693, opened 2016-12-05
Worker tasks plugin

zachmullen:


plugins/worker_tasks/server/rest.py
```diff
@@ -64,7 +64,9 @@ def _validateTask(self, item):
             ioSpec['format'] = ioSpec.get('format', 'none')
             ioSpec['type'] = ioSpec.get('type', 'none')
 
-        return spec
+        handler = item.get('meta', {}).get('workerTaskHandler') or 'worker_handler'
```
kotfic: @zachmullen would it be possible to move this up earlier in the execution and then trigger a handler specific event (e.g. worker_task.{{ handler }}.validate)?  Currently the _validateTask method puts some requirements on input/output types,  mode etc that might not be applicable in all cases (as an extreme example, I could see them message being in some binary format).

zachmullen: Great idea

jeffbaumes:
Does this plugin handle outputs, and specifically file outputs? How would I correctly set up an output file that gets stored back into Girder?

zachmullen:
@jeffbaumes I am working on the output side next

jeffbaumes:
@zachmullen I see you are adding support for self-describing containers. I think it would make sense to add this for the JSON format as well. Should there be a special input argument, like `--describe` that when passed to compatible containers outputs the JSON?

Also, should be the strategy for multiple analyses per container? I see some code for listing the analyses in a container but I assume those still need to be split out into multiple task items, one for each analysis, in order to run them? Should containers report themselves as single-task or multi-task, or should they always be assumed to be multi-task?

I think there are many reasonable answers to these, but I'm just wondering where you are headed with these open questions.

plugins/item_tasks/server/rest.py
```diff
@@ -0,0 +1,319 @@
...
+                    'format': 'none'
+                }
+            else:
+                raise ValidationException('Invalid input mode: %s.' % v['mode'])
+
+        return transformed
+
+    def _transformOutputs(self, outputs, token):
+        """
+        Validates and sanitizes the input bindings. If they are Girder inputs, adds
```
jeffbaumes: Should refer to outputs in this comment.

zachmullen:
> I think it would make sense to add this for the JSON format as well. Should there be a special input argument, like --describe that when passed to compatible containers outputs the JSON?

These self-describing executable are defined as a specific standard (Slicer CLI) that to my knowledge only supports XML output right now. If we want to add JSON output support, that's probably something we'd need to talk with the Slicer folks to see what challenges that would bring.

> Also, what should be the strategy for multiple analyses per container?

This is a question @cdeepakroy has done a lot of work to address. Deepak, has the architecture you devised for this become part of the official standard? I have some thoughts on how Girder could handle this, but would want to hear Deepak's input before we try to formally support multi-CLI images.

jeffbaumes:
I think `ctk_cli` needs to be added to a `requirements.txt` somewhere. I got an error when trying to use the latest version of this branch.

zachmullen:
If you `pip install -e .[item_tasks]` it'll install the plugin dependencies. We only use `requirements.txt` for non-builtin plugins at this point.

jeffbaumes:
Ah, maybe this gets pulled in through the dependent plugin slicer_cli_web? There is nothing to say it should get installed as part of this plugin that I can find.

zachmullen:
We encode builtin plugin python dependencies in [setup.py](https://github.com/girder/girder/blob/worker-tasks-plugin/setup.py#L76-L85)

jeffbaumes:
Got it. I searched the PR diff for `ctk_cli` so didn't catch `ctk-cli` in `setup.py`.

jeffbaumes:
I have the following issue. I run a task specified as:

```json
{
  "container_args": [
    "-m",
    "analytics_tasks.statistics.foldchange",
    "--data=$input{data}",
    "--threshold=$input{threshold}",
    "--output=/mnt/girder_worker/data/output"
  ],
  "description": "Perform a fold change analysis.",
  "docker_image": "analytics_tasks",
  "inputs": [
    {
      "description": "",
      "id": "data",
      "name": "The input data table",
      "required": true,
      "target": "filepath",
      "type": "file"
    },
    {
      "default": {
        "data": 2
      },
      "description": "",
      "id": "threshold",
      "min": 0,
      "name": "Fold change threshold",
      "required": true,
      "type": "number"
    }
  ],
  "mode": "docker",
  "name": "Fold change",
  "outputs": [
    {
      "description": "",
      "id": "output",
      "name": "The fold change table",
      "target": "filepath",
      "type": "new-file"
    }
  ],
  "pull_image": false
}
```

It completes successfully, but for the output it states "Not created - view destination". But the output is actually created. I think I have the latest. Any suggestions?

zachmullen:
This is probably in a good state to review and merge. We can add further improvements on future branches.

jeffbaumes:
This branch is quite massive, but I'll give a big LGTM from a practical usage point of view. I'm using this plugin to good effect, and am seeing it function without a hitch. I've also seen that it can be extended to inspect and run tasks inside other types of self-describing Docker containers, which would come in a later PR after this gets merged. I have also utilized the UI generation components that were moved here from slicer-cli-web, so I can attest that these were built with reusability in mind.

If someone can vet the source and tests, it would be great to get this merged. Maybe @jbeezley since a good amount of this is the UI code?

jbeezley:
I'm up against some hard deadlines right now, but I can take a look at it early next week.

jbeezley:
In looking at the code here, the ctk-cli stuff looks a little out of place.  It feels like the autodescribing docker containers should output a json spec similar (or identical) to that in the item metadata.  I think the `cli_parser.py` file would be better placed inside `slicer_cli_web` to act as an intermediary between slicer containers and generic item task containers.

zachmullen:
I'm torn; I agree that that would be cleaner, however I was hoping we could avoid having a runtime dependency on the slicer_cli_web plugin when we deploy this, since we don't really want to enable any of its other functionality.

jbeezley:
If I understand correctly, the dependency should go the other way.  That is slicer_cli_web would depend on this.

My point is the cli schema passed from docker containers for general jobs shouldn't be based on slicer.  I think something more like the json schema actually stored as metadata would be more appropriate.  In this case, the cli_parser code would be an adapter that is only used by slicer containers.

zachmullen:
> If I understand correctly, the dependency should go the other way. That is slicer_cli_web would depend on this.

I meant I was hoping we would not have to enable the slicer_cli_web plugin when we deploy this, since we don't want to use most of its functionality.

> My point is the cli schema passed from docker containers for general jobs shouldn't be based on slicer.

Agreed. When discussing this with @jeffbaumes , the idea is that this is just one mode for auto-configuration, and we would support others (e.g. #1764). My vision was that clicking "Configure task" in the future would open a dialog containing multiple tabs, of which one would be "Slicer CLI", another would be "JSON", and then we can add any others we wish. Does that make sense?

jbeezley:
I guess I'm kind of cold on the idea of supporting multiple auto-describing schemas in general.  It seems like something that is better handled at the docker/application level.

As an example from HistomicsTK, the slicer cli spec uses numeric vectors for colors.  From a user interface perspective, it would be much better to display this with a color picker, but there is no generic way to determine when that is the case.  However, the author of the docker container defining the CLI could easily make the conversion to the "color" type manually.

There are other places where the slicer schema would benefit from additional hints from the docker image author that can't be done automatically.  Should a file be provided as an item or file id?  Should there be a default path?  Are there limitations on the file extension or mime type?  We can engineer around these problems, but it seems to me the better solution is to provide a reference schema that analyses adapt to.

zachmullen:
That's fair, and we should make it easy for image authors to output a more robust JSON spec if they want. That said, this feature is already written, tested, and documented, and we have actual ongoing projects where we have pre-existing docker images that adhered to the Slicer CLI spec, and therefore required no changes at all to work in this system, which was a compelling feature for our collaborators on those projects. So, unless there is some risk in having this feature there for site admins to use, I think we should allow it in.

jbeezley:
In the interest of progress, I'll give this a stamp of approval because it's working as is.  I would like to see us move towards a more formalized description of the task item metadata.  Perhaps #1764 will help guide us in that direction.

zachmullen:
A JSON-schema for it would be great 😁

PR #1694, opened 2016-12-06
Render all hierarchy descriptions as markdown

zachmullen:
In the worker tasks plugin, items represent analyses, and it makes sense that they may have rich descriptions that we want to display when configuring the task for execution.

This makes the item description consistent with folder descriptions in the web client.

### DIFF
--- a/clients/web/src/stylesheets/body/itemPage.styl
+++ b/clients/web/src/stylesheets/body/itemPage.styl
@@ -1,6 +1,7 @@
 .g-item-header
   border-bottom 1px solid #ddd
-  margin-bottom 15px
+  margin-bottom 12px
+  padding-bottom 8px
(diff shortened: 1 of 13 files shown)


zachmullen:
@jcfr PTAL, this fixes part of the issues you noticed on data.kitware.com. Namely, the displaying of ` != renderMarkdown` bug has been fixed. Also, this makes the folder, item, and collection descriptions all render consistently as markdown.

jcfr:
Nice. Is there a test instance where I could have a look ... my dev env is not up-to-date (or a docker container I could use to quickly check ?)

zachmullen:
This PR isn't showing the latest state of this branch for some reason, I am going to re-open it to try to fix.

PR #1695, opened 2016-12-08
Collections view rendering

zachmullen:
@jcfr this contains the other changes you requested, namely improving the rendering of the collection list. Long collection names are now gracefully truncated, and descriptions are shown as markdown by clicking on a toggle. Hopefully this will encourage people to use more rich descriptions for their collections.

Depends on #1696

### DIFF
--- a/clients/web/src/stylesheets/layout/global.styl
+++ b/clients/web/src/stylesheets/layout/global.styl
@@ -111,6 +111,20 @@ textarea
 /* Stuff that applies across the top level object lists */
 .g-top-level-list-title-container
   display inline-block
+  overflow hidden
+  white-space nowrap
+  width calc(100% - 350px)
+  position relative
+
+  .g-text-fade
+    height 30px
+    width 45px
+    position absolute
+    right 0
+    background linear-gradient(left, rgba(250, 250, 252, 0), rgba(250, 250, 252, 1))
+
+    &[public="false"]
+      background linear-gradient(left, rgba(255, 249, 234, 0), rgba(255, 249, 234, 1))
 
 .g-list-public-status-icon
   display inline-block
--- a/clients/web/src/views/body/CollectionsView.js
+++ b/clients/web/src/views/body/CollectionsView.js
@@ -8,8 +8,7 @@ import router from 'girder/router';
 import SearchFieldWidget from 'girder/views/widgets/SearchFieldWidget';
 import View from 'girder/views/View';
 import { cancelRestRequests } from 'girder/rest';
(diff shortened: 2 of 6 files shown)


clients/web/src/views/body/CollectionsView.js
```diff
@@ -94,6 +96,25 @@ var CollectionsView = View.extend({
         collection.set('_id', result.id).on('g:fetched', function () {
             router.navigate('/collection/' + collection.get('_id'), {trigger: true});
         }, this).fetch();
+    },
+
+    _toggleDescription: function (e) {
+        const link = $(e.currentTarget);
+        const cid = link.attr('cid');
+        const dest = this.$(`.g-collection-description[cid="${cid}"]`);
```
msmolens: Is it intended that this line uses backticks instead of single quotes?

zachmullen:
Here is how it looks now:

screen shot 2016-12-08 at 10 25 53 am

clients/web/src/views/body/CollectionsView.js
zachmullen: If you are interested in seeing all the greatness of modern ES, [this](http://es6-features.org/) is a really nice resource.

clients/web/src/stylesheets/layout/global.styl
```diff
@@ -111,6 +111,20 @@ textarea
 /* Stuff that applies across the top level object lists */
 .g-top-level-list-title-container
   display inline-block
+  overflow hidden
+  white-space nowrap
+  width calc(100% - 350px)
```
jcfr: I know this out of the scope of this PR, but is using this type of rule "mobile" friendly ?

zachmullen: Yes, this is a CSS3 feature and all modern browsers should support it. If we want to change the view to look different on smaller device sizes, we'd need to use media queries.

jcfr:
Based on limited understanding of "pug et al." and what @zachmullen showed me, this looks good :+1:

clients/web/src/stylesheets/layout/global.styl
jcfr: Makes sense. For future reference, see https://www.w3.org/TR/css3-mediaqueries/#media0

msmolens:
I noticed that the descriptions inherit a bold font weight:
image

I think it would be better to render the markdown with its intended styles, like:
image

zachmullen:
Oops, good catch, I forgot to remove the `font-weight: 700` style rule from `.g-collection-description`.

PR #1696, opened 2016-12-08
Render all descriptions in the data hierarchy as markdown

zachmullen:
Reopened #1694 since that PR had bugged out.


jcfr:
Based on limited understanding of "pug et al.", this looks good :+1:

PR #1700, opened 2016-12-09
In prod environment, only install prod npm dependencies

zachmullen:
This makes it so that we call `npm install --production` within `girder-install web`, skipping the devDependencies section of npm packages.

There is a bit of a hack here; the `grunt-webpack` package unconditionally includes a `webpack-dev-server` task that we never used anyway, and in the prod environment, that package it tries to import does not exist. It didn't break the build, but it did cause an error message to appear, so in order to suppress the inclusion of that task, we now just delete the included task as a postinstall step.

### DIFF
--- a/scripts/postNpmInstall.js
+++ b/scripts/postNpmInstall.js
@@ -1,10 +1,13 @@
 // Remove the webpack-dev-server grunt task since we do not use it
+
 var fs = require('fs');
-var path = require('path');
+var taskFile = null;
 
-var taskFile = path.join(__dirname, '..', 'node_modules', 'grunt-webpack', 'tasks', 'webpack-dev-server.js');
+try {
+    taskFile = require.resolve('grunt-webpack/tasks/webpack-dev-server.js');
+} catch(e) {}
 
-if (fs.existsSync(taskFile)) {
+if (taskFile) {
     console.log('Removing unused task file ' + taskFile);
     fs.unlinkSync(taskFile);
 }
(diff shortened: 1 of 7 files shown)


zachmullen:
Any @girder/developers have time to QA this? It reduces the response time of our circle builds by a noticeable amount.

scripts/postNpmInstall.js
```diff
@@ -0,0 +1,10 @@
+// Remove the webpack-dev-server grunt task since we do not use it
+var fs = require('fs');
+var path = require('path');
+
+var taskFile = path.join(__dirname, '..', 'node_modules', 'grunt-webpack', 'tasks', 'webpack-dev-server.js');
```
jbeezley: Does `require.resolve('grunt-webpack')` work here?  That would be a more resilient method of finding it's path.

zachmullen: Just tried, no luck. Because `grunt-webpack` does not specify a `main` inside its package.json, require.resolve fails on it. If there's some way to only get the path to a module directory without looking up its entry point, we could use that instead.

zachmullen: Tried a bit more, and `require.resolve('grunt-webpack/tasks/webpack-dev-server.js')` *does* work. I'll just have to either try/catch around it or pass some option to turn off exception throwing if the file isn't found. 👍

jbeezley: Yeah, that's better.

PR #1701, opened 2016-12-10
Description decorator that also controls behavior

zachmullen:
This is extremely similar to describeRoute, but in addition to
building the swagger documentation, it also actually validates
and coerces the request parameters to ensure that they adhere
to the description docs.


zachmullen:
This is ready for review. I've ported all of the core routes and most of the plugins.

cjh1:
@mgrauer sure i would be happy to take a look at it;

girder/api/describe.py
```diff
@@ -304,7 +304,7 @@ def modelParam(self, name, description=None, model=None, destName=None, paramTyp
 
         return self
 
-    def jsonParam(self, name, description, paramType='query', required=True, default=None,
+    def jsonParam(self, name, description, paramType='query', dataType='string', required=True, default=None,
```
zachmullen: This is always a string, no?

cjh1:
@zachmullen I took to liberty of pushing another commit! It adds the dataType parameter to jsonParam(...). This allows a model to be specified so we see that is the swagger doc. Hope that makes sense?

zachmullen:
Ah, that makes perfect sense. Will that be easy to tie in with jsonschema?

cjh1:
@zachmullen That is a good question. So currently we have addModel(...) that allows a model to be added to the definitions section of the swagger file that can then be used when documenting endpoints. As far as I  understand it the swagger object definitions are a subset of JSON schema. So it would be nice if we could use the same schema to document the endpoint as well as providing validation. I think there are two possibilities in term of validation.
- Use the existing addModel(...) call to allow users to add "swaggers subset" of JSON schema and then add a "validate" options to jsonParam(...) which would then use the model specified in the dataType parameter to do the validation. ( this case would give use documentation and validation )
- Add a new parameter to jsonParam(...) maybe 'schema' that could be any valid JSON schema and in the case the mere presence of the parameter would switch on validation. ( just validation )

I guess we could do both or something completely different.

cjh1:
Whilst migrating to autoDescribeRoute(...) I discovered that jsonParam(...) doesn't extract out the request body if paramType='body' (there was a TODO for this case). I have a fix that I can push next week, I just need to add a test case.

cjh1:
@zachmullen I have pushed another commit for processing JSON bodies. I would appreciate you  passing a critical eye over it.

zachmullen:
Just took a look, I approve 👍

zachmullen:
BTW, it doesn't show up in the test output unfortunately, but the tests are failing when they try to set the metadata via JSON body (the old mechanism, not your new one). The response is:

    {"message": "Parameter \"body\" is required.", "type": "rest"}

zachmullen:
I think that's because we handle the case where it's a body param and in jsonParams, but not the case of a body param that is *not* in jsonParams.

cjh1:
I will take a look

cjh1:
@zachmullen I have pushed a fix. However, do you think if paraType is set to 'body' we could extract out the request body into a parameter ( in the non json case? )

zachmullen:
We shouldn't attempt to read it into memory, but it would be good to set it to `cherrypy.request.body` so that it could be read like a file.

cjh1:
Goodpoint, that would be a bad idea for large requests ...

girder/api/describe.py
```diff
@@ -535,10 +535,11 @@ def wrapped(*args, **kwargs):
...
                     kwargs['params'].pop(name, None)  # Remove from form/query params
-                elif descParam['in'] == 'body' and name in self.description.jsonParams:
-                    info = self.description.jsonParams[name].copy()
-                    info['required'] = descParam['required']
-                    kwargs[name] = self._loadJsonBody(name, info)
+                elif descParam['in'] == 'body':
+                    if name in self.description.jsonParams:
+                        info = self.description.jsonParams[name].copy()
+                        info['required'] = descParam['required']
+                        kwargs[name] = self._loadJsonBody(name, info)
```
zachmullen:
```
else:
    kwargs[name] = cherrypy.request.body
```

cjh1: Thanks, sorry got distracted ... will push

girder/api/describe.py
```diff
@@ -218,6 +245,87 @@ def param(self, name, description, paramType='query', dataType='string',
...
+        """
+        This should be used in lieu of ``param`` if the parameter is a model ID
+        and the model should be loaded and passed into the route handler.
+
+        :param name: The name passed in via the request, e.g. 'id'.
+        :type name: str
+        :param description: The description of the parameter. If not passed, defaults
+            to "The ID of the <model>."
+        :type description: str
+        :param destName: The kwarg name after model loading, e.g. 'folder'. Deafults
```
mgrauer: Defaults

girder/api/describe.py
```diff
@@ -218,6 +245,87 @@ def param(self, name, description, paramType='query', dataType='string',
         self._params.append(param)
         return self
 
+    def modelParam(self, name, description=None, model=None, destName=None, paramType='path',
+                   plugin='_core', level=None, required=True, force=False, exc=True,
+                   requiredFlags=None, **kwargs):
+        """
+        This should be used in lieu of ``param`` if the parameter is a model ID
+        and the model should be loaded and passed into the route handler.
```
mgrauer: It's hard for me to read through the docstring of the params and figure out how to use this with the defaults.  Can a few examples be provided in the docstring, using `name` and `map` and `model`?  With one example using `id` and one using something like `folderId` perhaps.

girder/api/describe.py
```diff
@@ -218,6 +245,87 @@ def param(self, name, description, paramType='query', dataType='string',
...
+
+        :param name: The name passed in via the request, e.g. 'id'.
+        :type name: str
+        :param description: The description of the parameter. If not passed, defaults
+            to "The ID of the <model>."
```
mgrauer: Unclear to me what this means.

Like does "The ID of the <model>" get converted to e.g. `folderId`?

girder/api/describe.py
```diff
@@ -377,3 +490,190 @@ def routeHandler(...)
...
+class autoDescribeRoute(describeRoute):  # noqa: class name
+    def __init__(self, description, hide=False):
+        """
+        Like describeRoute, but this decorator also controls behavior of the
+        underlying method. It handles parameter validation and transformation
+        based on the Description object passed.
+
+        :param description: The description object.
+        :type description: Description
+        :param hide: If this route should be hidden from the list, set this to True.
```
mgrauer: which list?

docs/plugin-development.rst
```diff
@@ -129,42 +131,79 @@ will default to being restricted to administrators.
...
+sent to the method as kwargs (so the order you declare them in the header doesn't matter).
+Any additional parameters that were passed but not listed in the ``Description`` object
+will be contained in the ``params`` kwarg as a dictionary. The validation of required
+parameters, coercion to the correct data type, and setting default values is all
+handled automatically for you based on the parameter descriptions in the ``Description``
+object passed. Two special methods of the ``Description`` object can be used for
+additional behavior control: :py:func:`girder.api.describe.Description.modelParam` and
+:py:func:`girder.api.describe.Description.jsonParam`.
+
+The ``modelParam`` method is used to convert parameters passed in as IDs to the corresponding
```
mgrauer: Is it worth giving an example with two different models as params, so `id` and some other param need to be used together?  Because `id` does some syntactic sugar/magic for you that e.g. `folderId` wouldn't.

Also, maybe to make the docstring a bit more clear?

    The ``modelParam`` method is used to convert parameters passed in as IDs to the model document corresponding to those IDs,

or even `hydrated model document`.

🐃

girder/api/describe.py
zachmullen: It uses the `model` param, not the name, so it would say "The ID of the folder" in that case.

girder/api/rest.py
```diff
@@ -903,8 +903,9 @@ def _matchRoute(self, path, route):
                 return False
         return wildcards
 
-    def requireParams(self, required, provided):
+    def requireParams(self, required, provided=None):
         """
+        This method has two modes, in the first mode # TODO document,
```
mgrauer: This needs to be completed.

girder/api/rest.py
```diff
@@ -914,35 +915,34 @@ def requireParams(self, required, provided):
...
 
-        for param in required:
-            if param not in provided:
-                raise RestException("Parameter '%s' is required." % param)
+            for param in required:
+                if provided is None or param not in provided:
+                    raise RestException('Parameter "%s" is required.' % param)
 
-    def boolParam(self, key, params, default=None):
+    @staticmethod
```
mgrauer: When reading through the class, why are these two (this and `getPagingParams`) `staticmethod` but no other methods, this smells a bit off.  At least a comment on each of these explaining why would help.

Also, in the `boundHandler` docstring at the bottom of this module, it has the following text which needs to be updated with this `staticmethod` change.

    this to gain access to bound convenience methods like ``self.model``,
        ``self.boolParam``, ``self.requireParams``, etc.

girder/api/v1/assetstore.py
```diff
@@ -53,33 +52,22 @@ def getAssetstore(self, assetstore, params):
...
 
     @access.admin
-    @describeRoute(
+    @autoDescribeRoute(
         Description('Create a new assetstore.')
         .responseClass('Assetstore')
         .notes('You must be an administrator to call this.')
         .param('name', 'Unique name for the assetstore.')
-        .param('type', 'Type of the assetstore.')
+        .param('type', 'Type of the assetstore.', dataType='integer')
```
mgrauer: Wouldn't this be better as an `enum`?

girder/api/v1/file.py
```diff
@@ -345,87 +304,77 @@ def cancelUpload(self, upload, params):
...
-    @loadmodel(model='item', map={'itemId': 'item'}, level=AccessType.WRITE)
     @filtermodel(model='file')
-    @describeRoute(
+    @autoDescribeRoute(
         Description('Copy a file.')
-        .param('id', 'The ID of the file.', paramType='path')
-        .param('itemId', 'The item to copy the file to.', required=True)
+        .modelParam('id', model='file', level=AccessType.READ)
+        .modelParam('itemId', description='The ID of the item to copy the file to.',
+                    level=AccessType.READ, paramType='formData')
```
mgrauer: Shouldn't this be `AccessType.WRITE`?

girder/api/v1/group.py
```diff
@@ -204,59 +178,46 @@ def joinGroup(self, group, params):
...
                'administrators and can be used to bypass the invitation process'
                ' and instead add the user directly to the group.')
-        .param('id', 'The ID of the group.', paramType='path')
-        .param('userId', 'The ID of the user to invite or accept.')
-        .param('level', 'The access level the user will be given when they '
-               'accept the invitation.',
-               required=False, dataType='int', default=AccessType.READ)
+        .modelParam('id', model='group', level=AccessType.WRITE)
+        .modelParam('userId', 'The ID of the user to invite or accept.',
+                    destName='userToInvite', level=AccessType.READ, paramType='formData')
```
mgrauer: Are we starting to use access level's on users now?

girder/api/v1/item.py
```diff
@@ -276,46 +249,35 @@ def deleteItem(self, item, params):
...
     )
     def rootpath(self, item, params):
         return self.model('item').parentsToRoot(item, self.getCurrentUser())
 
     @access.user(scope=TokenScope.DATA_WRITE)
-    @loadmodel(model='item', level=AccessType.READ)
     @filtermodel(model='item')
-    @describeRoute(
+    @autoDescribeRoute(
         Description('Copy an item.')
```
mgrauer: 'Copy an item, into the parent folder of the source item if no folderId is specified.'

girder/api/v1/system.py
```diff
@@ -171,103 +142,108 @@ def getVersion(self, params):
...
                'specific user.', required=False)
         .param('parentId', 'Restrict listing uploads to those within a '
                'specific folder or item.', required=False)
-        .param('assetstoreId', 'Restrict listing uploads within a specific '
-               'assetstore.', required=False)
+        .param('assetstoreId', 'Restrict listing uploads within a specific assetstore.',
+               required=False)
         .param('minimumAge', 'Restrict listing uploads to those that are at '
-               'least this many days old.', required=False)
+               'least this many days old.', dataType='float', required=False)
```
mgrauer: Better as an integer (yes, from the caller's perspective it is just clearer what you are getting)?  Would that break backwards compatibility?

docs/plugin-development.rst
```diff
@@ -129,42 +131,80 @@ will default to being restricted to administrators.
...
+will be contained in the ``params`` kwarg as a dictionary. The validation of required
+parameters, coercion to the correct data type, and setting default values is all
+handled automatically for you based on the parameter descriptions in the ``Description``
+object passed. Two special methods of the ``Description`` object can be used for
+additional behavior control: :py:func:`girder.api.describe.Description.modelParam` and
+:py:func:`girder.api.describe.Description.jsonParam`.
+
+The ``modelParam`` method is used to convert parameters passed in as IDs to the model document
+corresponding to those IDs, and also can perform access checks to ensure that the user calling the
+endpoint has a requisite access level on the resource. For example, we can convert the above
```
mgrauer: has the requisite

girder/api/v1/assetstore.py
zachmullen: The problem there is that we don't statically know all of the possible assetstore types, some of which may be defined in plugins. I think a future improvement to this infrastructure would be to add some methods to augment existing descriptions via plugins, at which point we could change this to use enum.

girder/api/v1/group.py
zachmullen: Users have always been ACModels, we just don't expose a lot of control over their ACLs at this point.

girder/api/v1/file.py
zachmullen: Good catch 🇫🇯

girder/api/v1/system.py
zachmullen: I think having a float is better, gives more control, and also is a superset of integers :)

mgrauer: So far that holds true for all the integers I've thought of, but I'll keep checking.

manthey: Floats are not necessarily supersets of integers, e.g., `int(float(int(17**17))) != int(17**17)`, but I don't think we care about those integers anyway.  :smirk_cat:

mgrauer:
Please add tests for the following cases.  I'm going off of the coverage report for the patch and am trying to stick to the most important stuff.  Also there might be something that is already covered but I misread the report (or maybe there was a break in the line and half of it was covered).

`girder/api/v1/resource.py`

- [ ] `_validateResourceSet` with an invalid resource (one that is not in `allowedModels`)
- [ ] `delete` on a non-access controlled resource (is that even possible? there is a code path for it)
- [ ] `_getResource` on a non-access controlled resource (same ?)

`girder/api/describe.py`

- [ ] `_validateParamInfo` with a `dataType` of `File`
- [ ] `_validateParamInfo` with a `paramType` of `Form`
- [ ] `_validateParamInfo` that will trigger the `if dataType == 'file' and paramType != 'formData':` at the bottom
- [ ] @autoDescribeRoute that will trigger `if 'type' not in descParam and 'schema' not in descParam`
- [ ] `loadModel` passing None as id for a required model
- [ ] `loadModel` passing required flags

`girder/api/v1/system.py`

- [ ] some paths through partial uploads could be tested better, not required for this PR

`girder/api/v1/file.py`

- [ ] some paths through upload endpoints could be tested better, not required for this PR

`girder/api/v1/user.py`

- [ ] changing a user status from a non-admin user
- [ ] `sendVerificationEmail` is not covered well, not required for this PR

`girder/api/rest.py`

- [ ] `requireParams` passing a required param that is not provided

zachmullen:
@mgrauer I believe this is ready to go.

girder/api/describe.py
```diff
@@ -122,75 +130,82 @@ def responseClass(self, obj, array=False):
...
         # Legacy data type conversions
         if dataType == 'int':
             dataType = 'integer'
-        elif dataType == 'File':
-            print(TerminalColor.warning(
-                "WARNING: dataType 'File' should be updated to 'file'"))
-            dataType = 'file'
+
+        # Parameter Object spec:
+        # If type is "file", then consumes MUST be either
```
mgrauer: What is `consumes`?

girder/api/describe.py
```diff
@@ -377,3 +494,190 @@ def routeHandler(...)
...
+        except ValueError:
+            raise RestException('Parameter %s must be valid JSON.' % name)
+
+        self._validateJsonType(name, info, val)
+
+        return val
+
+    def _loadModel(self, name, info, id):
+        if id is None:
+            if info['required']:
```
mgrauer: Add a test for this case, where you load a required model through autodescribe route, but id is None.

tests/cases/resource_test.py
```diff
@@ -336,8 +336,17 @@ def testDownloadResources(self):
         self.assertEqual(zip.namelist(), [])
 
     def testDeleteResources(self):
+        # Make sure an invalid resource type
```
zachmullen: Yep, cruft

girder/api/describe.py
zachmullen: My understanding of it is that it's the swagger specification of how request data should be encoded. We only use this in one place in core, our file chunk upload endpoint, to specify that the data should be multipart encoded.

girder/api/describe.py
zachmullen: I thought I did that in 7a2a815, but that isn't `None`, it's actually the string `'None'`. I'm not sure it's possible to get this value to be `None` through the HTTP API, so I will remove this check altogether.

girder/api/describe.py
mgrauer: Can you add a mention that consumes comes from the swagger spec in that case?  The context wasn't clear to me as I read it.

mgrauer:
🍾

PR #1702, opened 2016-12-12
Improve log output during web client build

zachmullen:
@ronichoudhury PTAL, this makes some minor changes to the log messages you added in your recent branch related to plugin building. These changes were made to improve the visualization of the log output when configuring plugins for building.

* For the new config options, output will only appear if you select a non-default option. If you turn on verbose mode, you will get all the messages as before.
* Status messages about plugin config options are now indented with a leading " >>" to make it visually apparent which plugin config they pertain to.

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -35,6 +35,8 @@ module.exports = function (grunt) {
         return;
     }
 
+    require('colors');
+
     grunt.config.merge({
         default: {
             plugin: {}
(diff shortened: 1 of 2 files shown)


zachmullen:
As for the grunt upgrade included in this, that resulted from me misreading an error message and thinking that we needed a newer version of grunt to get the verbose log output. I left that change in because it seems like a good idea to move from 0.x to 1.x of grunt.

Issue #1703, opened 2016-12-13
Static plugin interdependencies cause duplicate code loading


zachmullen:
When a plugin statically builds against another plugin's code, and both plugins are enabled, the same code can be loaded multiple times since the dependencies are pulled into both target bundles.

The solution is to use webpack's DllReference for the plugin bundle dependencies in the same way that we do on the core package dependency.

Concretely, this is currently causing a bug where the job timeline isn't rendering worker plugin's statuses since there are now two JobStatus objects competing with each other.

zachmullen:
After digging through this for a while, it gets incredibly hairy very quickly. The particular issue I ran into can be solved by just modifying the global object dynamically, instead of trying to do it statically across DLL boundaries, so for now I'm going to close this as wontfix since it's more trouble than it's worth at the moment.

PR #1704, opened 2016-12-13
Cannot modify plugin globals via static imports

zachmullen:
This is the issue I referred to in #1703.

The problem is, when statically importing symbols from one plugin to another, at runtime these objects are copies, not the same global object. The point of this line in the worker plugin is to modify a shared object (JobStatus), whereas it was only modifying its own copy, which broke the job rendering page.

I'm not sure of the best pattern to address cases like these where we want to mutate some global state in an upstream plugin, but we can't do it via static imports. My attempts to make plugins generate their own DLLs and link against one another didn't yield great results, and seemed to be leading to a nightmare of engineering maintenance.

### DIFF
--- a/plugins/worker/web_client/JobStatus.js
+++ b/plugins/worker/web_client/JobStatus.js
@@ -1,3 +1,6 @@
+// Since plugins are not dynamically linked against other plugin libraries,
+// we have to modify the runtime global JobStatus, rather than importing it
+// here statically, which would only modify a local copy.
 girder.plugins.jobs.JobStatus.registerStatus({
     WORKER_FETCHING_INPUT: {
         value: 820,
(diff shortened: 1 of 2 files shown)


mgrauer:
Is [this](https://github.com/girder/girder/blob/master/plugins/worker/web_client/main.js#L5) import still needed with this change?

It is, in `girder/plugins/worker/web_client/main.js`

    // Extends and overrides API
    import './JobStatus';


Additional points if you can decide what needs to be done to remove the nearby comment

    // unclear what it is really used for

mgrauer:
Now that I think of it, maybe a comment should be added to the definition of `girder.plugins.jobs.JobStatus.registerStatus` describing this?  I don't particularly like that idea, as it isn't anything inherent to that function, but I don't want someone else to be surprised by this.

Any ideas on surprise prevention?

zachmullen:
I removed the comments from main.js. All the imports there are necessary. Also added a comment explaining why we modify the object this way.

PR #1705, opened 2016-12-14
Filter results by flags

zachmullen:
@mgrauer you should probably be the one to QA this since you QA'd the original permission flags work. This fills in a gap in our access flag functionality, namely within `filterResultsByPermission`.

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -1264,9 +1264,8 @@ class AccessControlledModel(Model):
                                   removeKeys=(), flags=None):
         """
         Given a database result cursor, this generator will yield only the
-        results that the user has the given level of access on, respecting the
-        limit and offset specified. It can also only yield results for which the
-        user has certain access flags.
+        results that the user has the given level of access and specified access flags on,
+        respecting the limit and offset specified.
 
         :param cursor: The database cursor object from "find()".
         :param user: The user to check policies against.
@@ -1281,6 +1280,7 @@ class AccessControlledModel(Model):
                            matching document.
         :type removeKeys: list
         :param flags: A flag or set of flags to test.
+        :type flags: flag identifier, or a list/set/tuple of them
         """
         if flags:
             def hasAccess(doc):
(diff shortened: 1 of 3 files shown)


girder/models/model_base.py
```diff
@@ -1261,11 +1261,12 @@ def copyAccessPolicies(self, src, dest, save=False):
         return dest
 
     def filterResultsByPermission(self, cursor, user, level, limit=0, offset=0,
-                                  removeKeys=()):
+                                  removeKeys=(), flags=None):
         """
         Given a database result cursor, this generator will yield only the
         results that the user has the given level of access on, respecting the
-        limit and offset specified.
+        limit and offset specified. It can also only yield results for which the
```
mgrauer: I find this language a bit confusing (specificaly the 'also only yield results' part).  Maybe

    Given a database result cursor, this generator will yield only the
    results that the user has the given level of access and specified access flags on, respecting the
    limit and offset specified.

Also, do you want to support the case where someone only searches by flags and not by level, or do you always want someone to specify a level?

girder/models/model_base.py
```diff
@@ -1279,8 +1280,14 @@ def filterResultsByPermission(self, cursor, user, level, limit=0, offset=0,
         :param removeKeys: List of keys that should be removed from each
                            matching document.
         :type removeKeys: list
+        :param flags: A flag or set of flags to test.
```
mgrauer: :type flags: flag identifier, or a list/set/tuple of them

girder/models/model_base.py
```diff
@@ -1279,8 +1280,14 @@ def filterResultsByPermission(self, cursor, user, level, limit=0, offset=0,
...
         """
-        hasAccess = functools.partial(self.hasAccess, user=user, level=level)
+        if flags:
+            def hasAccess(doc):
+                return (self.hasAccess(doc, user=user, level=level) and
```
mgrauer: and and "and"

girder/models/model_base.py
zachmullen: Users can only see resources to which they have at least READ access, so we always filter by level.

I'll change the docstring to be clearer.

PR #1706, opened 2016-12-14
Support building web client plugins that are symlinked into plugins dir

zachmullen:
@ronichoudhury PTAL. This fixes the issue where certain plugins failed to build if they were symlinked into the directory, namely those that want to modify the core girder web client using the standard girder webpack loaders, i.e. they have no custom webpack configuration.

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -205,7 +205,18 @@ module.exports = function (grunt) {
                 for (var i = 0; i < numLoaders; i++) {
                     var selector = 'webpack.options.module.loaders.' + i + '.include';
                     var loaders = grunt.config.get(selector);
-                    grunt.config.set(selector, loaders.concat([path.resolve(dir)]));
+                    var pluginPath = path.resolve(dir);
+                    var realPath = fs.realpathSync(dir);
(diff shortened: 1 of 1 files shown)


zachmullen:
Are we sure it's in node? My first guess would be that the issue would be in webpack, but I haven't dug into it.

ronichoudhury:
Right. I shouldn't have said "node" - I just mean we should kick it upstairs if we can, first by finding a minimal example that demonstrates the problem (including whether it's Webpack or something else). You are right, it's more likely to be a Webpack issue.

jbeezley:
Module resolution for symlinked directories is complicated in node-land in general.  Problems with it are well known and treated as a WONTFIX.  https://github.com/nodejs/node/issues/3402

ronichoudhury:
:sigh:

zachmullen:
Welp, next to getting involved with that whole controversy, this fix seems pretty painless!

ronichoudhury:
Agreed! Ship it!

jbeezley:
There is also this: https://github.com/webpack/webpack/issues/985#issuecomment-261497772 I don't know if it will help this particular issue or not.

PR #1708, opened 2016-12-15
Support multiple bundles in the same plugin using config option

zachmullen:
@ronichoudhury PTAL, this addresses the issue we discussed this morning. I tested this on COVALIC and it works.

Generally speaking, this supports cases like COVALIC where a plugin needs to build both a plugin.min.* to augment the main girder app, as well as its own custom front-end that uses the same loaders.

### DIFF
--- a/docs/plugin-development.rst
+++ b/docs/plugin-development.rst
@@ -508,6 +508,25 @@ above, the value of this property is ``web_client/main.js``):
         }
     }
 
+You may also set ``main`` to an object that maps bundle names to entry points, which is
+helpful for plugins that want to build multiple targets using the same loaders. For example:
+
+.. code-block:: json
+
+    {
+        "name": "MY_PLUGIN",
+        "webpack": {
+            "main": {
+                "plugin": "web_client/main.js",
+                "external": "web_external/main.js"
+            }
+        }
+    }
+
+That will cause both ``plugin.min.*`` and ``external.min.*`` files to appear in the
+built directory. The file paths of the entry points should be specified relative to the
+plugin directory.
+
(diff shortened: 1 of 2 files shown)


zachmullen:
In particular, it would be good to hear comments on how multiple bundles per plugin should mesh with the existing `defaultLoaders` and `configHelper` features.

ronichoudhury:
This looks good to me (and works with Resonant Lab).

I have a couple of questions. if you want to build multiple bundles with the same overall config, then you could simply supply the `main` object with different outputs listed in the keys. How does this "generalized" method of supplying the output name interact with what I had originally added, namely the `webpack.output` config property? Is there a case for getting rid of that config option, and force the "object style" for controlling the output name?

What happens if you wish to build two bundles with different configs? Is that something a reasonable developer would even want to do?

zachmullen:
The old behavior of `webpack.output` and `webpack.main` (as a string) will still work as before. However, if `main` is specified as an object, `webpack.output` has no effect.

docs/plugin-development.rst
```diff
@@ -508,6 +508,25 @@ above, the value of this property is ``web_client/main.js``):
...
+                "plugin": "web_client/main.js",
+                "external": "web_external/main.js"
+            }
+        }
+    }
+
+That will cause both ``plugin.min.*`` and ``external.min.*`` files to appear in the
+built directory. The file paths of the entry points should be specified relative to the
+plugin directory.
+
```
ronichoudhury: I think we should mention that using the object format for `webpack.main` means `webpack.output` will be ignored.

PR #1709, opened 2016-12-16
Add progress option to file move

cjh1:


girder/models/upload.py
```diff
@@ -349,7 +349,7 @@ def createUpload(self, user, name, parentType, parent, size, mimeType=None,
         upload = adapter.initUpload(upload)
         return self.save(upload)
 
-    def moveFileToAssetstore(self, file, user, assetstore):
+    def moveFileToAssetstore(self, file, user, assetstore, progress=None):
```
zachmullen: Let's add a docstring for this new parameter. Also, if you prefer, you could use `girder.utility.progress.noProgress` as the default value for this param; it's an immutable progress context on which every operation is a no-op, which prevents having to have `if progress` conditional blocks inside the method body. Whether that's better or not is a matter of personal preference, though.

PR #1710, opened 2016-12-18
Missing _ import in dialog.js

zachmullen:
This isn't a problem for the main girder app since `window._` is set there. However, for other apps that link against `girder_lib`, it's problematic if _ is not in global scope.

### DIFF
--- a/clients/web/src/dialog.js
+++ b/clients/web/src/dialog.js
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import _ from 'underscore';
 import Backbone from 'backbone';
 
 import router from 'girder/router';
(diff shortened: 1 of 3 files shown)


jbeezley:
There's a similar problem in `girder/clients/web/src/views/layout/LoginView.js`.  I put in a PR on the eslint config to remove all the globals.  It looks like these are the only two things that break.  What do you think about pushing a new release there and updating?

zachmullen:
Just approved that branch, let me know when the update goes through and I'll kick circle.

jbeezley:
It's up now.  You'll have to update the package.json to 2.0 here.

PR #1712, opened 2016-12-19
Browser widget improvements

jbeezley:
This contains some bug fixes and enhancements for the Browser widget.
* Added CSS to limit the size of the modal dialog and make the hierarchy widget scrollable.
* Fixed page limit parameters for collection in the root selector widget.
* Fixed some race conditions in the tests.
* Added a mode for selecting items rather than folders.
* Added a setting to preset a root collection/user.

### DIFF
--- a/clients/web/src/stylesheets/widgets/browserWidget.styl
+++ b/clients/web/src/stylesheets/widgets/browserWidget.styl
@@ -1,0 +1,4 @@
+.g-hierarchy-widget-container
+  max-height calc(100vh - 400px)
+  min-height 200px
+  overflow-y scroll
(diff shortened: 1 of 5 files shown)


clients/web/src/stylesheets/widgets/browserWidget.styl
```diff
@@ -0,0 +1,4 @@
+.g-hierarchy-widget-container
+  max-height calc(100vh - 400px)
+  min-height 200px
```
zachmullen: Is the min-height necessary?

jbeezley: Not strictly.  I added it in case somebody has a very small screen, then the hierarchy widget will still appear.  It's not ideal, but there are many bigger problems small screens anyway.

PR #1716, opened 2016-12-20
Must declare global girder for eslint

zachmullen:


zachmullen:
We changed our upstream eslint-config-girder prior to my recent change to this file, and the eslint test failed on master as a result.

jbeezley:
Actually, I'm a little confused how this works.  The `girder` global is exposed in [main.js](https://github.com/girder/girder/blob/master/clients/web/src/main.js#L7) in the same place as `_`.  If you build a plugin in a such as way that `_` isn't exposed how then is `girder` set globally?

zachmullen:
It's not, this is not for the same plugin. This plugin is one that modifies the main girder app, the other one is its own standalone app (COVALIC).

jbeezley:
Wouldn't an external app that uses this plugin break then because girder wouldn't be defined globally?

jbeezley:
Anyway, that is probably an issue for a different PR.  This LGTM.

zachmullen:
I guess so, I'm not sure of a good solution to the problem in general. We could dynamically link our plugins against each other, but I'd still want to be able to have a static-only dependency on a plugin's web code. We could architect such a solution, but it would involve more options in plugin.json probably.

PR #1718, opened 2016-12-21
Move local node_modules under main node_modules dir

zachmullen:
This moves the local node_modules under the main node_modules dir. This has the advantage of allowing the search path to be visible via relative paths at config time, and also makes these directories automatically gitignored.


ronichoudhury:
This is a good idea. I assume this is premised sufficiently independently of my proposal in #1721 to merge this now and update that one if necessary?

zachmullen:
Yep, I think so, it will probably have merge conflicts, but they should be trivial.

PR #1721, opened 2016-12-22
Implement an "npm install" option for Webpack-built plugins

ronichoudhury:
The idea is to let the Webpack build process install the plugin's dependencies in its own directory, via an ordinary `npm install` invocation (much the same as what is already recommended for Gruntfile-built plugins).

I believe this approach may be strictly superior to the `node_modules_<plugin>` approach previously implemented:
- It uses no non-standard directories, and does only standard-looking things
- It eliminates the in-plugin need to use a bulky node alias (e.g. `import foo from girder_plugins/plugin/node/foo')
    - (the build process still supplies the alias, so sibling plugins will be able to gain access to the plugin's sources and dependencies as needed)
- It enables "dependency cascading" or whatever the correct term is (i.e., the plugin can invoke its own dependencies, but also have access to Girder's by invoking them in the same form)

The one downside: symlinking directories into a Node-built project doesn't work the way Unix programmers might expect. For my own use case I bit the bullet and decided to only support the non-symlink mode of plugin installation. If Node ever fixes this problem in a principled way, this downside will disappear.

Depends on #1720 to prevent Girder loaders from being applied to sources in plugin's `node_modules` directory.

### DIFF
--- a/docs/plugin-development.rst
+++ b/docs/plugin-development.rst
@@ -729,6 +729,18 @@ Each type needs to be installed differently due to how node manages external pac
   isolated. By default, the ``localNodeModules`` is set to ``false`` and the
   dependencies will be installed to Girder's own ``node_modules`` directory.
 
+  The final alternative for Webpack-built plugins is to set the ``npm.install``
+  configuration property to ``true``; this will cause the build system to run
+  ``npm install`` in the plugin directory. This may have certain benefits for
+  plugin development, such as allowing plugin sources to import modules without
+  the alias prefix as described above (though, this alias would still be
+  available for use by other plugins that want to access your plugin's
+  dependencies). Additionally, if your plugin is installed without using
+  symlinks, then you will still have access to Girder's Node dependencies (see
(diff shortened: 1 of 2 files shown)


ronichoudhury:
This may not be necessary. I could just explain in the documentation for my plugin that it must be installed without the symlink option, and that it's necessary to walk into the plugin directory and run `npm install` manually.

Thoughts?

ronichoudhury:
I rebased against master (including the `node_modules` change from #1718).

zachmullen:
If I use this new option and want to import something from my plugin's local node_modules dir, what import path do I use?

ronichoudhury:
Same as you would otherwise - `girder_plugins/foobar/node/...`.

zachmullen:
Cool, let's just mention that in the new docs, and then fix the eslint failures, and then this LGTM.

zachmullen:
@ronichoudhury what's the status on this? I know you'd had ideas recently for how to achieve this, does that make this PR obsolete?

PR #1722, opened 2016-12-27
Only admins should be able to select destination assetstore

zachmullen:


zachmullen:
@girder/developers  Could someone review this? I'm planning to cut a release and would like this bugfix to go in.

Thanks!

zachmullen:
Thanks @mgrauer !

PR #1725, opened 2017-01-02
Clarify ansible testing

danlamanna:
@zachmullen Sorry for the confusion, let me know if this clears things up or not.


zachmullen:
Looks good, one thing I'm unclear on: is this testing suite just meant to test the role, or does it also include the testing of the girder ansible module?

danlamanna:
Both. The first label tests the role, and the second tests the module/client.

PR #1729, opened 2017-01-03
add more widget options and client-side filtering

opadron:
Adds options to the HierarchyWidget and other related widgets that allow client-side filtering of the items/folders that are to be displayed. Also added a "folder selection mode".

In order to get the filtering working, I had to extend the base Collections class with additional filtering logic.  As much as possible, I tried to keep the filtering completely transparent to any subclasses.

The addition of transparent filtering gets tricky when considering the case of pagination.  The fetch logic now makes multiple round-trips as needed to fill up an entire "page", filtering as it goes.  The trickier issue, however, was in dealing with going backwards (fetchPreviousPage()).  The best solution I came up with was to maintain a stack of offsets from which pages where successfully fetched.  In any case, all the functional changes should be entirely opt-in, and should still work same as before by default.

 - Added new per-instance attribute to
   Collections: "filterFunc".

   - Set to a function to perform client-side
     filtering of fetched collection items.

     ```
     /* set this attribute before fetching data */
     myFolderListView.collection.filterFunc = (
         (model) => model.name !== 'test');
     ```

 - Added new options to FolderListWidget

   - folderFilter: collection filter function for
     folders

 - Added new options to ItemListWidget

   - itemFilter: collection filter function for
     items

 - Added new options to HierarchyWidget

   - itemFilter: collection filter function for
     items and folders

   - selectFolder: set to true to enable folder
     selection mode

   - onFolderSelect: callback triggered when
     selecting a folder in folder selection mode

     - selectFolder and onFolderSelect must both
       be set to enable folder selection mode.  In
       this mode, a button is created at the top
       of every folder view with the text "Select
       this folder/user/collection."  Useful for
       prompting a user to select a girder folder.

### DIFF
--- a/clients/web/src/collections/Collection.js
+++ b/clients/web/src/collections/Collection.js
@@ -105,6 +105,10 @@ var Collection = Backbone.Collection.extend({
     /**
      * Return the 0-indexed page number of the current page. Add 1 to this
      * result when displaying it to the user.
+     *
+     * If this collection hasn't been fully initialized (i.e.: before any pages
+     * have been fetched), pageNum() may return a page number < 0 to signal that
+     * there is no "current page".
      */
     pageNum: function () {
         if (this.filterFunc) {
--- a/clients/web/src/collections/Collection.js
+++ b/clients/web/src/collections/Collection.js
@@ -159,7 +159,9 @@ var Collection = Backbone.Collection.extend({
                     sort: this.sortField,
                     sortdir: this.sortDir
                 }, this.params)
-            }).done(_.bind(function (list) {
+            });
+
+            var result = $.when(xhr).then(_.bind(function (list) {
                 if (!_.isArray(list)) {
                     list = [list];
                 }
@@ -247,7 +249,7 @@ var Collection = Backbone.Collection.extend({
                         if (this.append && !reset) {
                             this.add(finalList);
                         } else {
-                            this.reset(finalList);
+                            this.reset(finalList, this.offset);
                         }
                     }
 
@@ -255,7 +257,7 @@ var Collection = Backbone.Collection.extend({
                 }
             }, this));
             xhr.girder = {fetch: true};
-            return xhr;
+            return result;
         }
 
         return fetchListFragment.apply(this);
--- a/clients/web/src/collections/Collection.js
+++ b/clients/web/src/collections/Collection.js
@@ -161,7 +161,7 @@ var Collection = Backbone.Collection.extend({
                 }, this.params)
             });
 
-            var result = $.when(xhr).then(_.bind(function (list) {
+            var result = $.when(xhr).then((list) => {
                 if (!_.isArray(list)) {
                     list = [list];
                 }
@@ -255,7 +255,7 @@ var Collection = Backbone.Collection.extend({
 
                     this.trigger('g:changed');
                 }
-            }, this));
+            });
             xhr.girder = {fetch: true};
             return result;
         }
--- a/clients/web/src/collections/Collection.js
+++ b/clients/web/src/collections/Collection.js
@@ -40,7 +40,7 @@ var Collection = Backbone.Collection.extend({
      * property should not be changed in the definition of collections, but
      * after they are instantiated.
      */
-    filter: null,
+    filterFunc: null,
 
     /**
      * If filtering and not appending, this stack is used to record the offsets
@@ -53,7 +53,7 @@ var Collection = Backbone.Collection.extend({
      * i.e. if the offset of the current page start is > 0
      */
     hasPreviousPage: function () {
-        if (this.filter) {
+        if (this.filterFunc) {
             if (this.append) { return false; }
             this.pageOffsetStack = this.pageOffsetStack || [];
             return (this.pageOffsetStack.length > 1);
@@ -74,8 +74,8 @@ var Collection = Backbone.Collection.extend({
      * Fetch the previous page of this collection, emitting g:changed when done.
      */
     fetchPreviousPage: function (params) {
-        if (this.filter) {
-            if(this.append) {
+        if (this.filterFunc) {
+            if (this.append) {
                 this.offset = 0;
             } else {
                 this.pageOffsetStack = this.pageOffsetStack || [];
@@ -107,7 +107,7 @@ var Collection = Backbone.Collection.extend({
      * result when displaying it to the user.
      */
     pageNum: function () {
-        if (this.filter) {
+        if (this.filterFunc) {
             if (this.append) { return 0; }
             this.pageOffsetStack = this.pageOffsetStack || [];
             return this.pageOffsetStack.length - 1;
@@ -129,12 +129,12 @@ var Collection = Backbone.Collection.extend({
             return;
         }
 
-        if (this.filter && !this.append) {
+        if (this.filterFunc && !this.append) {
             this.pageOffsetStack = this.pageOffsetStack || [];
         }
 
         if (reset) {
-            if (this.filter && !this.append) {
+            if (this.filterFunc && !this.append) {
                 this.pageOffsetStack = [];
             }
             this.offset = 0;
@@ -142,7 +142,7 @@ var Collection = Backbone.Collection.extend({
             this.params = params || {};
         }
 
-        if (this.filter && !this.append) {
+        if (this.filterFunc && !this.append) {
             this.pageOffsetStack.push(this.offset);
         }
 
@@ -150,7 +150,7 @@ var Collection = Backbone.Collection.extend({
 
         var finalList = []; /* will be built up in pieces */
 
-        function fetchListFragment () {
+        function fetchListFragment() {
             var xhr = restRequest({
                 path: this.altUrl || this.resourceName,
                 data: _.extend({
@@ -160,6 +160,10 @@ var Collection = Backbone.Collection.extend({
                     sortdir: this.sortDir
                 }, this.params)
             }).done(_.bind(function (list) {
+                if (!_.isArray(list)) {
+                    list = [list];
+                }
+
                 if (this.pageLimit > 0 && list.length > this.pageLimit) {
                     // This means we have more pages to display still. Pop off
                     // the extra that we fetched.
@@ -175,8 +179,8 @@ var Collection = Backbone.Collection.extend({
                  * If filtering, decorate the list with their pre-filtered
                  * indexes.  The index will be needed when adjusting the offset.
                  */
-                if (this.filter) {
-                    var filter = this.filter;
+                if (this.filterFunc) {
+                    var filter = this.filterFunc;
                     list = (
                         list
                             .map(function (x, index) { return [index, x]; })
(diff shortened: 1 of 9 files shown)


opadron:
Looks like

    60 - web_client_routing

Is timing out at some point.  But, I am not able to reproduce this timeout locally.

opadron:
@zachmullen feel free to pass this along to anyone else who might be more familiar with the code being altered.

jbeezley:
Good work on the tests.  One thing that makes me nervous is not making any use of the filtering mechanism anywhere outside of the tests.  I believe there is code in the [jobs plugin](https://github.com/girder/girder/blob/master/plugins/jobs/web_client/views/JobListWidget.js) that performs the same kind of filtering on collections.  Would it be possible to refactor the `JobListWidget` to use this instead?

opadron:
@jbeezley Good point.  The truth is I was working on these features for my own project, which is why I didn't seek to use it anywhere else in core.

I did notice the jobs plugin was doing some kind of its own filtering, which is actually why I ended up renaming my attribute to `filterFunc` (thus avoiding a collision).  I'm not too familiar with the details of the plugin, though, and the filtering looks a little more complicated.

jbeezley:
From what I can tell, the jobs list widget is just filters by job status.  Since @cjh1 wrote that part, maybe he can weigh in.

The code here is well documented and the testing is good.  My concern here is that this adds a lot of cyclomatic complexity to `collection.fetch`.  It also causes a core asynchronous function to make multiple requests in serial.  There is a good chance race conditions will crop up that we will have to deal with later.

I think something like this needs to be justified by a clear use case, which means making use of it somewhere.  To me the primary benefit for client side filtering would be to handle multiple filter functions  reactively in response to input from the user.  E.g. filtering by jobs status or regex on item names.  It looks like this change can't handle that use case because modifying the filter function requires reseting the collection state.

opadron:
@jbeezley that's right.  The filtering is meant to be transparent to users of the collection.  Our intended use case (as hinted by the other changes in this PR) is to use the HierarchyWidget like a file/folder-selection dialog, and to be able to, for example, prompt the user to select a csv file (showing only girder items with names ending in '.csv' as the user browses).

I added the filtering functionality to work transparently at the collection level because laying the ground work for our use case anywhere else would have required more involved changes, run into non-trivial problems with pagination, and would have been of less potential value in other hypothetical scenarios where one might want to filter client-side.  The consequence of this decision is that the chosen filter (if any) is an inseparable part of the collection state, and use cases that require filters that change over the lifetime of the collection are not served by this PR.  Those use-cases are better off implemented at the View level, such as in the jobs plugin.

The change to the workings of the asynchronous fetch are least of our concerns, imo.  Recall that the entire change should be entirely opt-in, and in the default case only one fetch should ever be needed for a given page (just like how it works, now).  Since we already fetch collection data in discrete pages, anyway, there's nothing inherently inferior about staggering these fetches.  In other words, whatever race conditions you might be worried about are already there -- excluding these changes will not save us from having to deal with them.

clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -117,6 +118,18 @@ var HierarchyWidget = View.extend({
 
         this._showActions = _.has(settings, 'showActions') ? settings.showActions : true;
         this._showItems = _.has(settings, 'showItems') ? settings.showItems : true;
+
+        /*
+         * require that selectFolder be set *and* that a folder select callback
+         * be provided
```
jbeezley: Why require both settings to be present to enable this?  If there really has to be two options here, I think an or condition would be more appropriate (making the selection callback a noop).

opadron: Good point.  I'm trying to remember what my thought process was here, but I think I agree with you.  This part also reminds me of a "foldersOnly" option I meant to add; where the view would only show folders (and not items).  I wonder if maybe I somehow got that mixed up here?

opadron: Ack!  I just notice there's already a showItems option.

clients/web/src/collections/Collection.js
```diff
@@ -54,12 +74,29 @@ var Collection = Backbone.Collection.extend({
...
+            if (this.append) {
+                this.offset = 0;
+            } else {
+                this.pageOffsetStack = this.pageOffsetStack || [];
+                if (this.pageOffsetStack.length) {
+                    this.pageOffsetStack.pop();
+                }
+
+                this.offset = (
+                    this.pageOffsetStack.length ? this.pageOffsetStack.pop() : 0
```
jbeezley: Why does this end up popping two elements off the stack?

clients/web/src/collections/Collection.js
```diff
@@ -70,6 +107,11 @@ var Collection = Backbone.Collection.extend({
      * result when displaying it to the user.
      */
     pageNum: function () {
+        if (this.filterFunc) {
+            if (this.append) { return 0; }
+            this.pageOffsetStack = this.pageOffsetStack || [];
+            return this.pageOffsetStack.length - 1;
```
jbeezley: If this is called before `this.pageOffsetStack`, the `pageNum` returned will be `-1`.  Shouldn't it be `0`?

clients/web/src/collections/Collection.js
opadron: That's because the fetch ends up pushing the offset back on.  So, the second pop is more of a peek.

clients/web/src/collections/Collection.js
```diff
@@ -87,46 +129,138 @@ var Collection = Backbone.Collection.extend({
...
+                    offset: this.offset,
+                    sort: this.sortField,
+                    sortdir: this.sortDir
+                }, this.params)
+            });
 
-            if (list.length > 0 || reset) {
-                if (this.append && !reset) {
-                    this.add(list);
+            var result = $.when(xhr).then(_.bind(function (list) {
```
jbeezley: I would use an arrow function here.

clients/web/src/collections/Collection.js
```diff
@@ -87,46 +129,138 @@ var Collection = Backbone.Collection.extend({
...
+                if (!_.isArray(list)) {
+                    list = [list];
+                }
```
jbeezley: Why would a collection fetch not return an array?  It doesn't look like the old code did anything like this.

clients/web/src/collections/Collection.js
opadron: I can change this if you want, but I think it does make some sense.  Similar to in the two-pops question, you only run into this -1 result if you call pageNum before ever fetching any pages.  It's only after having fetched the first page that you have the information you need to display pageNum() === 0.  One could interpret pageNum() === -1 as signaling that you're not even *on* a page which is distinct from being on the pageNum() === 0 page.

Anyway, let me know if you still want me to change this.

clients/web/src/collections/Collection.js
```diff
@@ -87,46 +129,138 @@ var Collection = Backbone.Collection.extend({
...
+                 * indexes.  The index will be needed when adjusting the offset.
+                 */
+                if (this.filterFunc) {
+                    var filter = this.filterFunc;
+                    list = (
+                        list
+                            .map(function (x, index) { return [index, x]; })
+                            .filter(function (tuple) {
+                                return filter(tuple[1]);
+                            })
```
jbeezley: I think this can just be
```javascript
list = list.filter(filter)
```

clients/web/src/collections/Collection.js
opadron: This is one change I don't have a good answer for, admittedly.  Before I added this check, tests where failing when trying to login/authenticate.  It seemed in those limited cases a collection fetch was returning a single scalar, which messed up the fetching algorithm.  After coercing scalar-returning fetches into length-1 arrays like I did above, everything started working again. so I left it in.

Maybe someone familiar with the user login tests could chime in and help me understand what was going on or possibly help me dig deeper to be absolutely sure if the array check is something we want to keep?

clients/web/src/collections/Collection.js
jbeezley: I guess you are using the `map` below.  I would avoid having the variable list mean two different things depending on if `filterFunc` is defined.  Maybe you could store the indices in a separate array?

opadron: This is not so, because the list is one of tuples, the second elements of which we need to filter by.  The result array needs to preserve the tuple structure.

opadron: I just saw your second comment.

While I could keep the lists separate, I would need to build an index so that I can cross reference items between the two.  I think not having to go through that manually in exchange for some variable name reuse is the better tradeoff, here.

jbeezley: How about setting this to a new, temporary variable and reassigning to list at the end?  To me, overloading the `list` variable is very confusing.  You also might want to call the filter function so it matches the standard JS api:
```javascript
return filter(tuple[1], tuple[0], list);
```

clients/web/src/collections/Collection.js
jbeezley: That makes sense.  Could you add a comment mentioning that here?

clients/web/src/collections/Collection.js
jbeezley: As a personal preference, something unexpected should raise an exception rather than passing a nonsense response back to the user.  In this case, I'm willing accept `-1` as a valid response if it is documented as such.

clients/web/src/collections/Collection.js
opadron: Alright, I'll add another variable.  If you absolutely cannot *stand* to have an overloaded variable name. :grin:

I don't know what you mean by "standard JS api" in this context; AFAIK, my usage is standard while the line of code in your comment is not?

edit: Oh, I think I see what you were getting at (maybe ;P).  No, the filter function provided does not conform to the standard Array.filter callback signature and that is intentional.  The filter function has online semantics, which is the major characteristic that sets it apart from a standard Array.filter callback.  The filter is applied to data as it is fetched, so it doesn't even make sense to consider the collection nor any internal index on the temporarily fetched data as part of the filtering criteria (the former because the collection may be incomplete, and the latter because the index has no meaning outside of the fetch and is subject to change from one fetch to the next).

jbeezley: Good point, the filter call is fine as is.

opadron: Also, I took a swing at the two-variables change and feel that I've gotta push back again on it, yet again.  While `list` may have a different structure in two cases (list of [index, value] pairs vs list of values), the common "list" structure is used in several points in the algorithm.  These points would all need branches added just to tell them which variable name to use, which I think unnecessarily complicates the code more than having the list take on two different shapes.

jbeezley: Okay, we can leave it then.

clients/web/src/collections/Collection.js
jbeezley: The test fails is one that I added that mocks rest calls.  I'll look into fixing it so we can remove this bit.

jbeezley:
@opadron See the commit I pushed.  It should remove the hack coercing the response into an array.

opadron:
LGTM, thanks so much for your help!

PR #1730, opened 2017-01-03
LDAP authentication plugin

zachmullen:
This is sufficient for me to be able to register/login against Kitware's ActiveDirectory.


jeffbaumes:
Looks like this depends on #1701? Will be good to get that merged first to be able to look at these changes more cleanly.

jeffbaumes:
Also: yay for this PR.

mgrauer:
This needs testing, front end development and documentation.

@zackgalbreath is this something you would be interested in taking on?

zackgalbreath:
> @zackgalbreath is this something you would be interested in taking on?

👍

zackgalbreath:
Apparently I don't have push permission here yet.  So instead, I pushed a draft of the config UI for this plugin [to my fork](https://github.com/girder/girder/compare/ldap-plugin...zackgalbreath:ldap-plugin).  PTAL @zachmullen.

I'll work on adding tests once we decide this is in good shape.

zachmullen:
@zackgalbreath you should now have write access to push those changes onto this topic 👍

mgrauer:
Aha, I had added @zackgalbreath to the Girder organization, but didn't realize I needed to make him a member of the developer team for this access.

danlamanna:
I can take a stab at reviewing this if no one else wants to, but I'm not incredibly familiar with LDAP. What are the steps for configuring this to work with our ActiveDirectory?

plugins/ldap/server/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+        # Append one if the user did not specify.
+        uri = server['uri']
+        if '://' not in uri:
+            uri = 'ldap://' + uri
+        conn = ldap.initialize(uri)
+        conn.set_option(ldap.OPT_TIMEOUT, _CONNECT_TIMEOUT)
+        conn.set_option(ldap.OPT_NETWORK_TIMEOUT, _CONNECT_TIMEOUT)
+
+        try:
+            conn.bind_s(server['bindName'], server['password'], ldap.AUTH_SIMPLE)
```
zachmullen: I assume you were using non-SSL (ldap://) mode. If you were sending these credentials over an insecure network, you would want to use LDAPS (ldaps://..., default port 636).

plugins/ldap/server/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+    try:
+        login = attrs[server['searchField']][0].split('@')[0]
+        return ModelImporter.model('user').createUser(
+            login, password=None, firstName=first, lastName=last, email=email)
+    except ValidationException as e:
+        if e.field != 'login':
+            raise
+
+    # Fall back to deriving login from user's name
+    for i in six.moves.range(_MAX_NAME_ATTEMPTS):
```
zachmullen: It's the OAuth plugin. Maybe so, could be moved to the user model.

plugins/ldap/server/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+def _defaultServers():
+    return []
+
+
+@setting_utilities.validator(PluginSettings.LDAP_SERVERS)
+def _validateLdapServers(doc):
+    try:
+        jsonschema.validate(doc['value'], _serversSchema)
+    except jsonschema.ValidationError as e:
+        raise ValidationException('Invalid LDAP servers list: ' + e.message)
```
zachmullen: In this case I believe bindName is being passed, but it's the empty string, and minLength is the rule used to enforce a non-empty string.

circle.yml
```diff
@@ -18,7 +18,7 @@ dependencies:
   cache_directories:
     - "~/cache"
   override:
-    - sudo apt-get -q update; sudo apt-get install -q -y libgif-dev
+    - sudo apt-get -q update; sudo apt-get install -q -y libgif-dev libsasl2-dev
```
zachmullen: The pyldap library is a wrapper around the native openLDAP library, which this installs. Do you want me to add this to the Dockerfile?

plugins/ldap/server/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+            conn.bind_s(server['bindName'], server['password'], ldap.AUTH_SIMPLE)
+        except ldap.LDAPError:
+            logger.exception('LDAP connection failed (%s).' % server['uri'])
+            continue
+
+        searchStr = '%s=%s' % (server['searchField'], login)
+        results = conn.search_s(server['baseDn'], ldap.SCOPE_ONELEVEL, searchStr, _LDAP_ATTRS)
+
+        if results:
+            entry, attrs = results[0]
```
zachmullen: We are assuming that the searchField should identify at most one entry; if it doesn't, that is a problem that we should probably log.

plugins/ldap/web_client/views/ConfigView.js
```diff
@@ -0,0 +1,125 @@
...
+        }));
+
+        this.breadcrumb.setElement(this.$('.g-config-breadcrumb-container')).render();
+        sortable.create(this.$('.g-ldap-server-accordion')[0]);
+        return this;
+    },
+
+    _saveSettings: function () {
+        const servers = _.map(this.$('.g-ldap-server-panel'), panel => {
+            const server = {};
```
zachmullen: `const` in es6 doesn't refer to internal mutability of the value, it only forbids reassignment of the variable. I have no idea why that would be valuable, maybe some hinting to the GC, but I've been trying to use it where it makes sense since it's been such a heavily documented feature in ES2015.

circle.yml
danlamanna: Our Dockerfile promises that Girder + all core plugins will install/function correctly, so if that library is necessary then we have to.

plugins/ldap/server/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+@autoDescribeRoute(
+    Description('Test connection status to a LDAP server.')
+    .notes('You must be an administrator to call this.')
+    .param('uri', 'The URI of the server.')
+    .param('bindName', 'The LDAP identity to bind with.')
+    .param('password', 'Password to bind with.')
+    .errorResponse('You are not an administrator.', 403)
+)
+def _ldapServerTest(self, uri, bindName, password, params):
+    conn = ldap.initialize(uri)
```
zachmullen: Probably a good idea, but I'm inclined to defer that to future topics.

zachmullen:
@danlamanna I think I addressed all the comments, this is ready for another pass.

danlamanna:
This LGTM other than Python3 throwing this exception when logging in:
```
  File "/home/dan/projects/girder/prs/1730/girder/api/rest.py", line 567, in endpointDecorator
    val = fun(self, args, kwargs)
  File "/home/dan/projects/girder/prs/1730/girder/api/rest.py", line 1120, in GET
    return self.handleRoute('GET', path, params)
  File "/home/dan/projects/girder/prs/1730/girder/api/rest.py", line 890, in handleRoute
    val = handler(**kwargs)
  File "/home/dan/projects/girder/prs/1730/girder/api/describe.py", line 601, in wrapped
    return fun(*args, **kwargs)
  File "/home/dan/projects/girder/prs/1730/girder/api/v1/user.py", line 120, in login
    user = self.model('user').authenticate(login, password)
  File "/home/dan/projects/girder/prs/1730/girder/models/user.py", line 142, in authenticate
    'password': password
  File "/home/dan/projects/girder/prs/1730/girder/events.py", line 291, in trigger
    handler['handler'](e)
  File "/home/dan/projects/girder/prs/1730/plugins/ldap/server/__init__.py", line 157, in _ldapAuth
    conn.bind_s(attrs['distinguishedName'][0], password, ldap.AUTH_SIMPLE)
  File "/home/dan/.virtualenvs/girder-1730/lib/python3.6/site-packages/ldap/ldapobject.py", line 400, in bind_s
    msgid = self.bind(who,cred,method)
  File "/home/dan/.virtualenvs/girder-1730/lib/python3.6/site-packages/ldap/ldapobject.py", line 394, in bind
    return self.simple_bind(who,cred)
  File "/home/dan/.virtualenvs/girder-1730/lib/python3.6/site-packages/ldap/ldapobject.py", line 379, in simple_bind
    return self._ldap_call(self._l.simple_bind,who,cred,RequestControlTuples(serverctrls),RequestControlTuples(clientct
rls))
  File "/home/dan/.virtualenvs/girder-1730/lib/python3.6/site-packages/ldap/ldapobject.py", line 265, in _ldap_call
    result = func(*args,**kwargs)
TypeError: argument 1 must be str, not bytes
```

zachmullen:
Good catch, our mocked LDAP service was returning text type, but the real library returns serialized strings.

zachmullen:
@danlamanna any other changes needed?

zachmullen:
Thanks for the review!

manthey:
I should have noticed sooner, but we need to add the appropriate apt-get modules to the ansible role (adding `libsasl2-dev` to devops/ansible/roles/girder/tasks/girder.yml).  Do we also need `libldap2-dev`?

zachmullen:
We don't on CircleCI, not sure about other environments.

danlamanna:
I think we do need libldap, since our Docker build is now throwing an error about missing `lber.h`: https://hub.docker.com/r/girder/girder/builds/bewwlmf5cpel3osdmclikec/

manthey:
I have a provisional branch that should address this.  I'll make a PR once I've run some more tests.

PR #1731, opened 2017-01-03
Fix broken web client build for vega plugin

zachmullen:
An upstream of the vega plugin (datalib) changed to require
loading of a json file. This change extends our loaders so that
the JSON loading rule also applies under node_modules, which
seems pretty harmless.

This is blocking our CI for the moment, could any @girder/developers give it a quick QA so we can get builds rolling again?


grunt_tasks/webpack.config.js
```diff
@@ -106,7 +106,7 @@ module.exports = {
             // JSON files
             {
                 test: /\.json$/,
-                include: loaderPaths,
+                include: [...loaderPaths, ...loaderPathsNodeModules],
```
jbeezley: Did we ever decide on a minimum node version for development?  This requires v4, I think.

zachmullen: v4 is now what our docs prescribe for installation. (We are actually already using ES2015 features in our build code besides this).

jbeezley: Okay, LGTM then.

zachmullen:
thanks @jbeezley !

PR #1732, opened 2017-01-03
Use JAVASCRIPT_STYLE_TESTS flag to control running eslint/puglint tests

zachmullen:
This allows downstreams (in my case COVALIC) to run static analysis
without having to run the setup/teardown tests for the web client
runtime.


zachmullen:
See https://github.com/girder/girder/blob/master/CMakeLists.txt#L19 for its definition. I think it was just a typo/bug that we were using the other variable for this.

PR #1733, opened 2017-01-03
Support serialization of hash states in python 2.7.13

zachmullen:


zachmullen:
CircleCI doesn't have python 2.7.13 available. @dorukozturk could you test this fix on your 2.7.13?

zachmullen:
I installed 2.7.13 locally and verified this fix. @girder/developers PTAL

girder/utility/hash_state.py
```diff
@@ -35,40 +38,46 @@ def _getHashStateDataPointer(hashObject):
...
+        # this expands to:
+        #     typedef struct {
+        #         Py_ssize_t ob_refcnt;      // Word 0
+        #         PyTypeObject *ob_type;     // Word 1
+        #         PyObject *name;            // Word 2
+        #         struct env_md_ctx_st {
+        #             const EVP_MD *digest;  // Word 3
+        #             ENGINE *engine;        // Word 4
+        #             unsigned long flags;   // Word 5
+        #             void *md_data;         // Word 6   << this guy is what we want
```
mgrauer: How do you know its gender?

👩‍🚀

zachmullen:
What, you don't like coding in C... *in python*?

I also hate that we have this in the code. If an alternative implementation exists that exposes this state in the python API, or allows pickling, we should switch, although a pure python implementation is probably too slow.

dorukozturk:
Works like a charm !!! LGTM

girder/utility/hash_state.py
```diff
@@ -25,6 +25,9 @@
 import hashlib
 import ctypes
 import binascii
+import sys
+
+_HASHLIB_INLINE_EVP_STRUCT = (sys.version_info < (2, 7, 13) or sys.version_info >= (3,))
```
msmolens: It looks like the commits to address [Python Issue #26470](https://bugs.python.org/issue26470) are in 2.7, 3.5, and 3.6, so I doubt `sys.version_info >= (3,)` is correct.

zachmullen: Good catch, I'll constrain this further and open another PR.

PR #1734, opened 2017-01-04
Hash struct was changed back to containing pointer in v3.5

zachmullen:
@msmolens PTAL

@mgrauer This has the added bonus that it will now let us test both conditions in CI.

### DIFF
--- a/girder/utility/hash_state.py
+++ b/girder/utility/hash_state.py
@@ -27,7 +27,8 @@ import ctypes
 import binascii
 import sys
 
-_HASHLIB_INLINE_EVP_STRUCT = (sys.version_info < (2, 7, 13) or sys.version_info >= (3,))
+_ver = sys.version_info
+_HASHLIB_INLINE_EVP_STRUCT = _ver < (2, 7, 13) or (_ver >= (3,) and _ver < (3, 5, 3))
 
 
 def _getHashStateDataPointer(hashObject):


zachmullen:
Nevermind, no python 3.5 on circleCI either 😢

jbeezley:
It has the following:

```
$ pyenv versions
  system
  2.7.10
* 2.7.11 (set by /opt/circleci/.pyenv/version)
  2.7.12
  3.1.4
  3.1.5
  3.2.5
  3.2.6
  3.3.5
  3.3.6
  3.4.3
  3.4.4
  3.5.1
  3.5.2
  pypy-1.9
  pypy-2.6.1
  pypy-4.0.1
```

You can also `pyenv install` other versions if necessary.

girder/utility/hash_state.py
```diff
@@ -27,7 +27,8 @@
 import binascii
 import sys
 
-_HASHLIB_INLINE_EVP_STRUCT = (sys.version_info < (2, 7, 13) or sys.version_info >= (3,))
+_ver = sys.version_info
+_HASHLIB_INLINE_EVP_STRUCT = (_ver < (2, 7, 13) or _ver >= (3,)) and _ver < (3, 5)
```
msmolens: For consistency with checking the minor version of 2.7, let's check the minor version of 3.5. The changelog shows that the relevant commit is in 3.5.3: https://docs.python.org/3.5/whatsnew/changelog.html#python-3-5-3.

zachmullen:
Excellent! I'll use 3.5.1 then.

girder/utility/hash_state.py
```diff
@@ -27,7 +27,8 @@
 import binascii
 import sys
 
-_HASHLIB_INLINE_EVP_STRUCT = (sys.version_info < (2, 7, 13) or sys.version_info >= (3,))
+_ver = sys.version_info
+_HASHLIB_INLINE_EVP_STRUCT = (_ver < (2, 7, 13) or _ver >= (3,)) and _ver < (3, 5, 3)
```
msmolens: The logic looks right but the grouping is confusing to me. Would it be better to group the `_ver >= (3,)` and `_ver < (3, 5, 3)` checks?

zachmullen:
Ok, I tried to use python 3.6 beta version, which added 3 minutes to the build due to not being installed by default, and also broke most of the python tests for some strange reason, looks like `six` might not even support python that new yet.

For now it might be best just to wait until either 3.5.3+ or 2.7.13 is added as a pre-installed option. I'm running 2.7.13 locally, so that's something at least...

PR #1740, opened 2017-01-06
Add instructions for troubleshooting upgrade

fedorov:

PR #1741, opened 2017-01-06
Update migration guide

zachmullen:
Reopened against origin since CircleCI was not building @fedorov's original submission


zachmullen:
refs #1740

PR #1742, opened 2017-01-07
[WIP] Dynamic linking between plugin bundles

zachmullen:
This solves the problem I've been running into where plugin code imported statically gets a copy of a module rather than referencing a single global version of it. This is because webpack does not normally link modules across multiple bundles. The way to do so is to use the DllPlugin and DllReferencePlugin together.

As before, plugins containing `web_client/main.js` will have `plugin.min.js` and `plugin.min.css` built as an entrypoint bundle from main.js. In addition to that behavior, plugins containing `web_client/index.js` will have a second bundle built as a DLL module. If this DLL module is built, the `main.js` bundle will import plugin symbols from the DLL module

Benefits of this approach:

1. Only one copy of all shared modules.
2. We already do this in core, so this behavior is consistent.

Downsides:

1. Webroot files could previously simply import `plugin.min.*` and that would be sufficient to load the plugin. Now, they must also import `plugin_lib.min.*` if they exist. So, this could break some existing custom webroots that rely on loading plugins.
2. The `girder-install web --watch-plugin` command, and our watch commands in general, can only watch a single webpack bundle at a time. If a developer is attempting to watch a plugin that contains both a DLL bundle and an entry bundle, they'll have to figure out which of the two bundles to watch, or run two processes at once. So, instead of `girder-install web --watch-plugin jobs`, they'll have to run some modified version of that command, either `girder-install web --watch-plugin lib_jobs`, or we could add a new option like `girder-install web --watch-plugin-lib jobs`.

TODO

- [ ] Merge #1738 and rebase this.
- [ ] Write docs.
- [ ] Refactor new plugin.js code to reduce duplication.
- [ ] Don't require explicit listing of DLL references between plugins in plugin.json.
- [ ] Add webroot utility to abstract the process of importing plugin code into the template.

### DIFF
--- a/girder/utility/webroot.mako
+++ b/girder/utility/webroot.mako
@@ -8,6 +8,9 @@
     <link rel="stylesheet" href="${staticRoot}/built/fontello/css/animation.css">
     <link rel="stylesheet" href="${staticRoot}/built/girder_lib.min.css">
     <link rel="icon" type="image/png" href="${staticRoot}/img/Girder_Favicon.png">
+    % for plugin in pluginLibCss:
+    <link rel="stylesheet" href="${staticRoot}/built/plugins/${plugin}/plugin_lib.min.css">
+    % endfor
     % for plugin in pluginCss:
     <link rel="stylesheet" href="${staticRoot}/built/plugins/${plugin}/plugin.min.css">
     % endfor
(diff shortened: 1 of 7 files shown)


zachmullen:
I've opened https://github.com/webpack/webpack/issues/4376 , which if it gets fixed will alleviate a lot of the problems with this.

mgrauer:
Closing this, wait for someone to complain or webpack to fix before taking any action.

zachmullen:
Reopening on account of https://github.com/webpack/webpack/issues/4376#issuecomment-301592662

PR #1744, opened 2017-01-07
Fixes #1743. Use modern API for ExtractTextPlugin

### LINKED ISSUE #1743: webpack ExtractTextPlugin is not pulling the CSS out of the JS module
All of the CSS is still inlined inside our .js files, in addition to being extracted into the .css files. We should find some way to remove this duplication, we might just be misusing ExtractTextPlugin.

@sebastienbarre I banged my head on this for a while, but the docs for ExtractTextPlugin leave a lot to be desired, and I couldn't get it working... any experience with this issue?

zachmullen:


zachmullen:
Anyone have time to review this bugfix?

PR #1748, opened 2017-01-11
New core setting for server root

zachmullen:


jbeezley:
Does this overlap the intent of the proxy settings that cherrypy uses?  For example in the [deployment docs](http://girder.readthedocs.io/en/latest/deploy.html#girder-settings), the api root is overridden in the settings.

For a more pathological question, wouldn't this make it impossible to host two girder instances from the same database or to proxy a single girder instance to multiple mount points?

zachmullen:
> Does this overlap the intent of the proxy settings that cherrypy uses? 

This is for special cases like @manthey 's where the girder instance may have a different URL on one machine than another (e.g. a VM). This shouldn't interfere with the proxy settings.

> wouldn't this make it impossible to host two girder instances from the same database or to proxy a single girder instance to multiple mount points?

If this setting is used, yes. However this is a configurable behavior and off by default; if not set, the behavior will be the same as before, which would enable those cases.

zachmullen:
@manthey this is now ready for review

zachmullen:
Actually sorry, I need to add front-end for it still

zachmullen:
Ok, actually ready now.

manthey:
Although this works for girder_worker, it breaks Swagger.  Would it be better to have this as a setting for the remote worker rather than core?  Or could we have Swagger not use the override logic?

zachmullen:
Ah right, because in your case there is no one address where girder is visible to all the relevant machines... In that case, it makes sense to have this in the worker plugin as a setting.

zachmullen:
I'll keep this setting too for whatever exotic use case my emerge. It's under advanced settings anyway, so it's probably not harmful.

girder/conf/girder.dist.cfg
```diff
@@ -18,7 +18,7 @@ replica_set = None
 [server]
 # Set to "production" or "development"
 mode = "development"
-api_root = "api/v1"
+api_root = "/api/v1"
```
manthey: Since you lstrip the `/`, what is the virtue of adding it here?

zachmullen:
I was planning to add the worker setting to this branch, would you prefer it on a separate PR?

manthey:
Add away.  I'll re-review it then.

manthey:
One minor point of confusion:  In the core setting, a value like `http://127.0.0.1:8080` automatically gets `/api/v1` appended to it (or whatever the config setting for api_url is).  In the remote worker setting, the value is the entire string (e.g., `http://127.0.0.1:8080/api/v1`).  We should document this somewhere.

PR #1752, opened 2017-01-17
add new flag to girder-install web: "--plugin-prefix"

opadron:


opadron:
@zachmullen PTAL.

I was able to use this flag by running

```
girder-install web --plugin osumo --plugin-prefix index
```

It seems like this, and other plugin-specific options are only used when singling out a particular plugin?  I have not figured out how to handle `girder-install web --all-plugins ...` when the prefixes differ among the plugins to be installed.

zachmullen:
I think that's fine, this really just refers to watching, which happens on a single bundle at a time.

zachmullen:
To fix that test, you just need to add this new option to our parser mock: https://github.com/girder/girder/blob/master/tests/cases/install_test.py#L33-L46

opadron:
@zachmullen don't you think there should be a little more intelligence, though?  It seems that with this option the way it is, now the officially sanctioned way to build plugins whose bundles aren't prefixed with "plugin" is to run girder-install web, let it fail, and then run girder-install web --watch-plugin PLUGIN_NAME --plugin-prefix PLUGIN_PREFIX for each plugin, ctrl-c-ing out of the watch process once the plugin is built.

Not the cleanest solution, imo.

zachmullen:
girder-install web won't fail, that just builds all the targets

opadron:
I thought that if you install a plugin and then run girder-install web, it will at some point try to run a webpack:plugin_$NAME task?

zachmullen:
`girder-install web` does not specifically enumerate targets to run, it just runs them all. The watch modes, however, attempt to run specific target names, which is why this new option is needed, but it will only affect watch mode.

opadron:
I just tested this and it looks like the intelligence I'm referring to is already there.  `girder-install web` can automatically infer when a plugin has a different prefix than the default.

So, everything here LGTM.

PR #1753, opened 2017-01-17
fix missing initialization parameters in hierarchy-widget

opadron:
Fixes a minor oversight from #1729 where subcomponents were being reinitialized without passing some of the newly added options along, causing incorrect behavior.

### DIFF
--- a/clients/web/src/views/widgets/HierarchyWidget.js
+++ b/clients/web/src/views/widgets/HierarchyWidget.js
@@ -539,7 +539,8 @@ var HierarchyWidget = View.extend({
                     checkboxes: this._checkboxes,
                     downloadLinks: this._downloadLinks,
                     viewLinks: this._viewLinks,
-                    itemFilter: this._itemFilter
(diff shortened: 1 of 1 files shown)


clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -535,7 +536,10 @@ var HierarchyWidget = View.extend({
             if (this.itemListView) {
                 this.itemListView.initialize({
                     folderId: parent.get('_id'),
-                    checkboxes: this._checkboxes
+                    checkboxes: this._checkboxes,
+                    downloadLinks: this._downloadLinks,
+                    viewLinks: this._viewLinks,
+                    itemFilter: this._itemFilter
```
jbeezley: Would it be worthwhile to add `showSizes` here as well?

opadron: may<sup>beeeee... <sup>(yes)</sup></sup></sup>

PR #1755, opened 2017-01-18
Imported file move

cjh1:
If an imported file is moved to another assetstore we need to reset the imported flag. Otherwise, in the case of the filesystem assetstore the download will fail as the appropriate path will not be generated.

### DIFF
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -178,6 +178,10 @@ class Upload(Model):
             file['created'] = datetime.datetime.utcnow()
             file['assetstoreId'] = assetstore['_id']
             file['size'] = upload['size']
+            # If the file was previously import, it is no longer.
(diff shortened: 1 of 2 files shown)


cjh1:
@manthey PTAL

girder/models/upload.py
```diff
@@ -178,6 +178,10 @@ def finalizeUpload(self, upload, assetstore=None):
             file['created'] = datetime.datetime.utcnow()
             file['assetstoreId'] = assetstore['_id']
             file['size'] = upload['size']
+            # If the file was previously import, it is no longer.
```
manthey: import -> imported

cjh1:
@manthey Thanks for the review

cjh1:
@TristanWright This should resolve the issue we were seeing yesterday.

PR #1759, opened 2017-01-20
Import S3 upload handler via FileModel

zachmullen:
This makes it so that any downstreams that use the FileModel or
the UploadWidget will also support direct-to-S3 uploads, since
unfortunately this module has to be imported for a side effect.

Refs https://github.com/girder/covalic/issues/220


zachmullen:
@msmolens I deployed this branch onto challenge.kitware.com and it fixed the issue.

PR #1760, opened 2017-01-25
fontello.zip sometimes fails to download

manthey:
When running the grunt tasks (such as via `npm install`), fontello.zip is downloaded from data.kitware.com.  Frequently, this fails, and an error of ` Corrupted zip : can't find end of central directory` is emitted when the file is unzipped.  Rerunning the task will often succeed (it has the same chance of failure as the initial run).

Anecdotally, this seems to happen more often when running in a VM or docker environment.

This appears to be caused by nodejs's `request` module emitting a finished event before the file is fully flushed to disk via a stream made with `fs.createWriteStream`.  This happens in the `grunt-curl` task (which, misleadingly, doesn't use `curl`), and also happens with the same frequency if the `request` module is used alone.

As a solution, we can shell out to the actual `curl` command.  We already require curl as a dependency, according to our docs.

If there is a preferred way to do this, I'd be happy to switch to a different method.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -70,7 +70,6 @@ module.exports = function (grunt) {
     grunt.loadNpmTasks('grunt-contrib-copy');
     grunt.loadNpmTasks('grunt-contrib-symlink');
     grunt.loadNpmTasks('grunt-gitinfo');
-    grunt.loadNpmTasks('grunt-curl');
     grunt.loadNpmTasks('grunt-zip');
     grunt.loadNpmTasks('grunt-file-creator');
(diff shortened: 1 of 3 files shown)


jbeezley:
One thing you might try when encountering timeouts is to update node.  There have been a lot of changes since about 4.10 that attempt to solve these timeouts. I would recommend trying the latest v6.

manthey:
I see this a lot on node 6.9.4.

jbeezley:
Okay, I guess they haven't fixed them all yet.  :)  It is on their radar though.

PR #1761, opened 2017-01-26
Revert to working version of extract-text-plugin

zachmullen:
@mgrauer PTAL

This fixes our build breakage, which is caused by an upstream breakage in extract-text-webpack-plugin as of v2.0.0-rc.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "css-loader": "^0.23.1",
     "eonasdan-bootstrap-datetimepicker": "~4.15",
     "event-source": "0.1.1",
-    "extract-text-webpack-plugin": "^2.0.0-beta.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "css-loader": "^0.23.1",
     "eonasdan-bootstrap-datetimepicker": "~4.15",
     "event-source": "0.1.1",
-    "extract-text-webpack-plugin": "^2.0.0-beta.2",
+    "extract-text-webpack-plugin": "2.0.0-beta.5",
```
jbeezley: I think you need `2.0.0-beta.4`.

jbeezley: Nevermind, beta 5 works for me.

mgrauer:
This works using npm 4.1.2.

I'll try it with nvm npm 4.2 next.

Issue #1170, opened 2016-01-13
Considering removing CherryPy version from response headers


danlamanna:
@ronichoudhury brought this up on Kitware/tangelo#533, and Girder also returns CherryPy/version in its response headers.

zachmullen:
There's probably a simple way to disable this in cherrypy, it's just a matter of tracking that down in the docs. We can certainly turn it off, from what I can tell there is no reason we should be sending the Server header at all.

cjh1:
response.headers.server = "" will do it

zachmullen:
Looking at our deployments, nginx actually rewrites the Server header to specify itself, so this is probably more of a deployment issue than anything.

brianhelba:
This is a nice "onramp" (or whatever we end up calling it) issue. Adding the label.

zachmullen:
It's onramp on the cherrypy side, but in real life we will be deferring the SSL handling to a proxy server that will need to be configured independently, which might not be as clear-cut for newbies.

PR #1762, opened 2017-01-26
Obfuscate underlying HTTP Server

zachmullen:
Details of our underlying HTTP server were leaking in two known
places; the Server header, and the error pages e.g. 404. This
addresses both of those, replacing cherrypy references with
Girder references.

### DIFF
--- a/girder/utility/server.py
+++ b/girder/utility/server.py
@@ -18,15 +18,33 @@
 ###############################################################################
 
 import cherrypy
+import mako
 import mimetypes
 import os
 import six
 
 import girder.events
 from girder import constants, logprint
+from girder.api.describe import API_VERSION
 from girder.utility import plugin_utilities, model_importer
 from girder.utility import config
 from . import webroot
+
+# TODO make a pretty error page someday.
+_errorTemplate = """
+<h1>${status}</h1>
+<p>${message}</p>
+<p><i>Girder server version ${version}</i></p>
+"""
+
+
+def _errorDefault(status, message, *args, **kwargs):
+    """
+    This is used to render error pages outside of the normal Girder app, such as
+    404's. This overrides the default cherrypy error pages.
+    """
+    return mako.template.Template(_errorTemplate).render(
+        status=status, message=message, version=API_VERSION)
 
 
 def configureServer(test=False, plugins=None, curConfig=None):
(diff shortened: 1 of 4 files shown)


girder/utility/server.py
```diff
@@ -18,16 +18,34 @@
...
+from girder.api.describe import API_VERSION
 from girder.utility import plugin_utilities, model_importer
 from girder.utility import config
 from . import webroot
 
+# TODO make a pretty error page someday.
+_errorTemplate = """
+<h1>${status}</h1>
+<p>${message}</p>
+<p><i>Girder server version ${version}</i></p>
```
cjh1: Do we what to expose the version information?

mgrauer: Yes, we already do in other places.

zachmullen: Yeah, it's already exposed through the web API and the front page of the web GUI.

cjh1: OK

danlamanna:
re #1170

girder/utility/server.py
mgrauer: After some further thinking, I'd rather we not expose the server version in more places for now.  We may want to retreat from doing it more generally.

So I say let's remove it from here, I don't think it is urgent that we have it here regardless.

girder/utility/server.py
```diff
@@ -18,16 +18,34 @@
...
 
 import girder.events
 from girder import constants, logprint
+from girder.api.describe import API_VERSION
 from girder.utility import plugin_utilities, model_importer
 from girder.utility import config
 from . import webroot
 
+# TODO make a pretty error page someday.
+_errorTemplate = """
```
mgrauer: I'd rather see this broken out to a static mako template for consistency with our other makos.  I'd be ok with a TODO comment or a GH issue tracking that though.

girder/utility/server.py
```diff
@@ -50,7 +68,9 @@ def configureServer(test=False, plugins=None, curConfig=None):
         '/': {
             'request.dispatch': cherrypy.dispatch.MethodDispatcher(),
             'request.show_tracebacks': test,
-            'request.methods_with_bodies': ('POST', 'PUT', 'PATCH')
+            'request.methods_with_bodies': ('POST', 'PUT', 'PATCH'),
+            'response.headers.server': 'Girder ' + API_VERSION,
```
mgrauer: Another option would be to send an empty string for this header value.  If we want to generally start reducing the number of places this is exposed.

girder/utility/server.py
zachmullen: I'm not convinced that hiding the version provides any meaningful security. Even if we remove the version output everywhere, plenty of other public-facing aspects of a girder server can be used to gain information about what version it's running; front-end text, swagger specs, new endpoints that were added, the content of static files being served... Given that the source code is open, the task of figuring out the server version is pretty simple, and hence I don't think we can ever really consider Girder server version to be secure information. (Versions of other tools in our underlying stack can and should be hidden, though.)

The argument about search engines indexing version info actually makes *our* job easier more than anything, by making it easier for us to discover outdated versions deployed in the wild.

If this is a sticking point, let's discuss it offline and come to a consensus.

mgrauer:
LSTM.  Sufficient that is, as it's 🦄 free.

screen shot 2017-01-27 at 1 08 52 pm

mgrauer:
Good catch

PR #1764, opened 2017-01-27
Add self-describing JSON

jeffbaumes:
Adds the ability for a JSON-based self-describing container to add all of its tasks to a folder. See https://github.com/Kitware/pysciencedock#usage-through-girder to see how to use this to import a set of tasks. I just pushed a `kitware/pysciencedock` up so it should work without needing to build the docker image manually.

### DIFF
--- a/plugins/item_tasks/server/rest.py
+++ b/plugins/item_tasks/server/rest.py
@@ -389,7 +389,8 @@ class ItemTask(Resource):
     @autoDescribeRoute(
         Description('Create item tasks under a folder using a list of JSON specifications.')
         .modelParam('id', model='folder', force=True)
-        .jsonParam('json', 'The JSON specifications as a list or a single specification object.', paramType='body')
+        .jsonParam('json', 'The JSON specifications as a list or a single object.',
+                   paramType='body')
         .param('image', 'The docker image name.', required=True, strip=True)
         .param('pullImage', 'Whether the image should be pulled from a docker registry. ' +
                'Set to false to use local images only.',
--- a/plugins/item_tasks/server/rest.py
+++ b/plugins/item_tasks/server/rest.py
@@ -23,7 +23,7 @@ class ItemTask(Resource):
         self.route('POST', (':id', 'slicer_cli_description'), self.runSlicerCliDescription)
         self.route('PUT', (':id', 'slicer_cli_xml'), self.setSpecFromXml)
         self.route('POST', (':id', 'json_description'), self.runJsonTasksDescription)
-        self.route('PUT', (':id', 'add_json_specs'), self.addJsonSpecs)
+        self.route('POST', (':id', 'json_specs'), self.addJsonSpecs)
 
     @access.public
     @autoDescribeRoute(
@@ -296,7 +296,7 @@ class ItemTask(Resource):
         .modelParam('id', model='item', force=True)
         .param('xml', 'The Slicer CLI XML spec.', paramType='body')
         .param('setName', 'Whether item name should be changed to the title of the CLI.',
-               dataType='boolean', required=False, default=True)
+               dataType='boolean', required=True)
         .param('setDescription', 'Whether the item description should be changed to the '
                'description of the CLI.', dataType='boolean', required=False, default=True),
         hide=True
@@ -325,6 +325,7 @@ class ItemTask(Resource):
         })
 
     @access.admin(scope=constants.TOKEN_SCOPE_AUTO_CREATE_CLI)
+    @filtermodel(model='job', plugin='jobs')
     @autoDescribeRoute(
         Description('Create item task specs based on a docker image.')
         .notes('This operates on an existing folder, adding item tasks '
@@ -361,10 +362,10 @@ class ItemTask(Resource):
                 'outputs': {
                     '_stdout': {
                         'mode': 'http',
-                        'method': 'PUT',
+                        'method': 'POST',
                         'format': 'text',
                         'url': '/'.join((getApiUrl(), self.resourceName,
-                                         str(folder['_id']), 'add_json_specs')),
+                                         str(folder['_id']), 'json_specs')),
                         'headers': {'Girder-Token': token['_id']},
                         'params': {
                             'image': image,
@@ -388,7 +389,7 @@ class ItemTask(Resource):
     @autoDescribeRoute(
         Description('Create item tasks under a folder using a list of JSON specifications.')
         .modelParam('id', model='folder', force=True)
-        .jsonParam('json', 'The JSON specs.', paramType='body')
+        .jsonParam('json', 'The JSON specifications as a list or a single specification object.', paramType='body')
         .param('image', 'The docker image name.', required=True, strip=True)
         .param('pullImage', 'Whether the image should be pulled from a docker registry. ' +
                'Set to false to use local images only.',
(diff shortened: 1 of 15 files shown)


zachmullen:
Rebased onto latest master.

zachmullen:
@jeffbaumes did you mean to check in the candela plugin onto this topic?

jeffbaumes:
It was here for convenience. I should split it into another WIP PR. You are welcome to do that for me if you feel inclined, since I probably won't get to it today.

zachmullen:
Done, the other topic is here: https://github.com/girder/girder/tree/candela-plugin

zachmullen:
@jeffbaumes if you merge latest master into this, it should fix the test flakiness.

jeffbaumes:
I've rebased on current master. I'm getting the following in one of the four builds. @zachmullen do you think this is from this branch or something else? Or is it transient?

```
142: . Create an item task -> create item
142: 
142: Create an item task: 4 of 4 passed.
142: . Auto-configure the item task -> navigate to the item
142: . Auto-configure the item task -> run configuration job
142: 
142: Auto-configure the item task: 0 of 0 passed.
142: Could not normalize value ""
142: Could not normalize value ""
142: Could not normalize value "1"
142: Could not normalize value "1"
142: Could not normalize value "20"
142: Could not normalize value "20"
142: Could not normalize value "40"
142: Could not normalize value "40"
142: Could not normalize value "3"
142: Could not normalize value "3"
142: Could not normalize value "5000"
142: Could not normalize value "5000"
142: Could not normalize value "30"
142: Could not normalize value "30"
142: Could not normalize value "3"
142: Could not normalize value "3"
142: Could not normalize value "false"
142: Could not normalize value "false"
142: Could not normalize value ""
142: Could not normalize value ""
142: ERROR: TypeError: undefined is not a constructor (evaluating '['number-vector', 'number-enumeration-multiple', 'string-vector', 'string-enumeration-multiple'].includes(this.get('type'))')
142: TRACE:
142:  -> undefined: 4012 (in function "isVector")
142:  -> : 3710 (in function "validate")
```

jbeezley:
`Array.prototype.includes` is an ES6 feature not supported by phantomjs.  I would check that we are correctly including babel's polyfill library in our bundles.

jeffbaumes:
Ok yeah it looks like this is me. I changed some `_.contains` to `[...].includes` on this branch. For now I'll go back to `contains`, but I'll also open an issue for including the ES6 polyfill.

plugins/item_tasks/server/rest.py
```diff
@@ -320,3 +323,94 @@ def setSpecFromXml(self, item, xml, setName, setDescription, params):
...
+
+        job = jobModel.save(job)
+        jobModel.scheduleJob(job)
+        return job
+
+    @access.token
+    @autoDescribeRoute(
+        Description('Create item tasks under a folder using a list of JSON specifications.')
+        .modelParam('id', model='folder', force=True)
+        .jsonParam('json', 'The JSON specs.', paramType='body')
```
zachmullen: We can now use the `schema` kwarg to validate the JSON if that's something you are interested in adding.

plugins/item_tasks/server/rest.py
```diff
@@ -320,3 +323,94 @@ def setSpecFromXml(self, item, xml, setName, setDescription, params):
...
+                creator=user,
+                folder=folder,
+                description=itemTaskSpec.get('description', ''),
+                reuseExisting=True)
+            logger.info(pullImage)
+            logger.info(params)
+            itemTaskSpec['docker_image'] = image
+            itemTaskSpec['pull_image'] = pullImage
+            self.model('item').setMetadata(item, {
+                'itemTaskSpec': itemTaskSpec
```
zachmullen: You'll also need to add `'isItemTask': True` in order for this to appear in the list of tasks (this is due to a limitation of mongo indexing where we cannot index for existence only).

plugins/item_tasks/server/rest.py
```diff
@@ -320,3 +323,94 @@ def setSpecFromXml(self, item, xml, setName, setDescription, params):
...
+        .param('pullImage', 'Whether the image should be pulled from a docker registry. ' +
+               'Set to false to use local images only.',
+               dataType='boolean', required=False, default=True),
+        hide=True
+    )
+    def addJsonSpecs(self, folder, json, image, pullImage, params):
+        self.ensureTokenScopes('item_task.set_task_spec.%s' % folder['_id'])
+        token = self.getCurrentToken()
+        user = self.model('user').load(token['userId'], force=True)
+        for itemTaskSpec in json:
```
zachmullen: I think it would be useful to also support single specs: if the value passed is an Object rather than an Array, we can assume it's just a single spec.

plugins/item_tasks/server/rest.py
```diff
@@ -320,3 +323,94 @@ def setSpecFromXml(self, item, xml, setName, setDescription, params):
...
+            logger.info('Configuring item "%s"' % itemTaskSpec['name'])
+            item = self.model('item').createItem(
+                name=itemTaskSpec['name'],
```
zachmullen: Having a default value for this name is probably a good idea. Maybe use `enumerate` in the loop and fall back to `<image_id> <i>`?

plugins/item_tasks/server/rest.py
```diff
@@ -21,6 +22,8 @@ def __init__(self):
         self.route('POST', (':id', 'execution'), self.executeTask)
         self.route('POST', (':id', 'slicer_cli_description'), self.runSlicerCliDescription)
         self.route('PUT', (':id', 'slicer_cli_xml'), self.setSpecFromXml)
+        self.route('POST', (':id', 'json_description'), self.runJsonTasksDescription)
+        self.route('PUT', (':id', 'add_json_specs'), self.addJsonSpecs)
```
zachmullen: This should probably be a `POST` since it's creating documents rather than updating them, and the path should be `json_specs` instead of `add_json_specs` (we avoid verbs in the path when possible, which is a convention of REST).

plugins/item_tasks/server/rest.py
jeffbaumes: The image id might be messy if it is using a specific docker sha, but yes a fallback could be good.

plugins/item_tasks/server/rest.py
jeffbaumes: This is now changed to `POST`.

plugins/item_tasks/server/rest.py
jeffbaumes: I have not done this, since this schema seemed somewhat complicated (can be array or single object). If you know how to put a basic check like this in I'd welcome that, otherwise we may want to add item task schema specification on another PR.

plugins/item_tasks/server/rest.py
jeffbaumes: Done.

plugins/item_tasks/server/rest.py
jeffbaumes: This now falls back to `<image_id> <i>`.

plugins/item_tasks/web_client/views/ConfigureTasksDialog.js
```diff
@@ -0,0 +1,42 @@
...
+                data.image = image;
+            }
+
+            restRequest({
+                path: `item_task/${this.model.id}/json_description`,
+                type: 'POST',
+                data,
+                error: null
+            }).done((job) => {
+                console.log(job);
```
jeffbaumes: Gone now.

jeffbaumes:
Ready for another review.

@zachmullen I've addressed all but the JSON schema issue. I've also added server and client testing to test the JSON endpoints for getting the JSON description from the container and configuring items. I did not re-test the execution of the created tasks because at that point it should function identically to the Slicer CLI item tasks and that was already tested.

plugins/item_tasks/web_client/main.js
```diff
@@ -47,6 +48,29 @@ ItemView.prototype.events['click .g-configure-item-task'] = function () {
     this.configureTaskDialog.render();
 };
 
+import hierarchyMenuModTemplate from './templates/hierarchyMenuMod.pug';
+wrap(HierarchyWidget, 'render', function (render) {
+    render.call(this);
+    this.$('.g-folder-actions-menu .dropdown-header').after(hierarchyMenuModTemplate({
+        _: _,
```
zachmullen: We don't use `_` in this template, so we don't need to pass it in.

PR #1766, opened 2017-01-28
Upgrade to webpack 2.2.0

zachmullen:
This unpins us from the beta version we were relying on. It also
allows us to get rid of our hacked-up custom DllReferencePlugin,
because the built-in version now supports our use case out of the box.

This also now works against the latest release candidate of the extract-text-webpack-plugin.


ronichoudhury:
Shiny! :spider_web: :package:

zachmullen:
Anybody interested in QAing? @ronichoudhury ? @sebastienbarre ? @jbeezley ?

jbeezley:
I'll give it a try.

package.json
```diff
@@ -19,14 +19,14 @@
...
     "css-loader": "^0.23.1",
     "eonasdan-bootstrap-datetimepicker": "~4.15",
     "event-source": "0.1.1",
-    "extract-text-webpack-plugin": "2.0.0-beta.5",
-    "file-loader": "^0.x",
+    "extract-text-webpack-plugin": "^2.0.0-rc.0",
+    "file-loader": "^0.9.0",
     "grunt": "^1.0.1",
-    "grunt-cli": "~0.1",
+    "grunt-cli": "^1.2.0",
```
jbeezley: With grunt updated, I think we can remove `nopt-grunt-fix`.

jbeezley:
With these changes, I'm seeing new problems with symlinked plugins.  For example, trying to build the latest large_image symlinked into place, I get this: <https://gist.github.com/jbeezley/933b461a9cd11b4ff4bd0a7cf244d666>

For some reason, I can't reproduce this problem by moving any of the core plugins to symlinks.  I'm not sure what is done different in large_image... a custom `Gruntfile.js` maybe?

zachmullen:
That is strange, the error message is from a failure to resolve node_modules, so I wouldn't think it's to do with the gruntfile. If you disable the custom gruntfile does the error still occur?

zachmullen:
Also worth checking, does reverting to extract-text-webpack-plugin 2.0.0-beta.5 fix that error?

jbeezley:
Nope, removing the custom gruntfile and downgrading `extract-text-webpack-plugin` doesn't fix it.

jbeezley:
It turns out the problem occurs when the plugin is symlinked parallel to the `girder` directory.  I'm not sure what to make of it.  It resolves modules from javascript fine, but gets tripped up loading javascript imported through the pug and css loaders.

I'm not sure what we should do here.  I spent quite a while trying to debug it.  I know of several deployments that would break due to this.  We could just go nuclear and say that symlinked plugins is not supported...

ronichoudhury:
When you say "through the loaders" what do you mean exactly?

jbeezley:
```
ERROR in ../large_image/web_client/templates/largeImage_fileAction.pug
Module not found: Error: Can't resolve './../../../girder/node_modules/pug-runtime/index.js' in '/home/jbeezley/tmp/large_image/web_client/templates'
```

The pug runtime is being injected by the pug loader.  Modules referenced by `large_image` directly as `import { restRequest } from 'girder/rest'`, etc. resolve correctly.

zachmullen:
The gist contains references to the `package.json` inside large_image, does removing that change anything?

jbeezley:
There is no `package.json` in that directory.

zachmullen:
On [master](https://github.com/girder/large_image/blob/master/package.json) I'm seeing one, and it's mentioned in your gist:

    using description file: /home/jbeezley/tmp/large_image/package.json (relative path: ./web_client/templates)

jbeezley:
Nevermind. there is.  Let me try it.

ronichoudhury:
There is also a separate set of Webpack options for resolving loader
modules. I wonder if that's at play here too.

jbeezley:
Good catch.  Removing `package.json` resolved the problem.  I think we can just remove `package.json` files from plugins now that dependencies can be installed via `plugin.json`.

package.json
```diff
@@ -19,14 +19,14 @@
...
     "bootstrap-switch": "3.3.2",
-    "colors": "0.6.2",
+    "colors": "1.1.2",
     "css-loader": "^0.23.1",
     "eonasdan-bootstrap-datetimepicker": "~4.15",
     "event-source": "0.1.1",
-    "extract-text-webpack-plugin": "2.0.0-beta.5",
-    "file-loader": "^0.x",
+    "extract-text-webpack-plugin": "^2.0.0-rc.0",
+    "file-loader": "^0.9.0",
```
jbeezley: Do you want to update all the loaders (`css-loader@0.26.1` and `file-loader@0.10.0`)?  I tried updating them in my debugging adventure, so I think they work.

zachmullen: Sure thing, that new file-loader was just released since I made this PR :)

zachmullen:
There is an actual problem here, which is shown

    aliased with mapping 'girder': '/home/jbeezley/tmp/girder/clients/web/src' to '/home/jbeezley/tmp/girder/clients/web/src/node_modules/pug-runtime/index.js'

It's looking for the pug runtime file inside girder/clients/web/src/node_modules instead of girder/node_modules. Not sure exactly what is causing this particular problem, could be we have to tweak our pug loader configuration options.

zachmullen:
Closes #1768

zachmullen:
@jbeezley after 7097a39, I can now build symlinked master of large_image that is a sibling of girder. Could you test that it works for you and that everything is OK in its runtime?

zachmullen:
Thanks @jbeezley

PR #1767, opened 2017-01-30
Fix triggerJobClick option of JobListWidget

jbeezley:
I think the class name on the event was inadvertently changed.  This restores the `triggerJobClick` option and adds a regression test.


zachmullen:
This was changed in https://github.com/girder/girder/commit/ba0206950df014aa73513a60fb24be692b97f305, but I'm not certain as to why. @cjh1 could you please review this PR?

PR #1769, opened 2017-01-31
Filter file models returned from upload process

zachmullen:


zachmullen:
@girder/developers could someone review this change?

PR #1777, opened 2017-02-06
Replace ItemSelectorWidget with BrowserWidget

jbeezley:
@zachmullen This is targeted at your WIP branch.  It removes a bunch of custom code and adds the ability to select items and folders from any collection.  I'm not sure I like the interface this creates, but it's the best I could do without making changes to the BrowserWidget itself.


zachmullen:
Thanks, I killed this circle build due to setting the parallel level to 4 on Circle which will currently fail. Once #1776 goes in we can rebase this and the CI should be green again.

jbeezley:
I think you can, but the UI just might not be obvious.  The way it works is, you choose a (valid) path first, then the text input unlocks and you can type a file name.

I played with generating two input elements (one writable for the file name, the other readonly for the path) but I couldn't get it to look right.  I do think we need to iterate on this, but I didn't want to wait too long because merge conflicts.

An additional note, I should add testing for the new validation paths.

zachmullen:
Maybe I've configured the task wrong, it's using type `new-file` which worked for me prior to this. When I navigate to a folder I don't see a text input appear, just this:

screen shot 2017-02-07 at 10 18 24 am

jbeezley:
It's not in the browser widget.  When you click save there, the main control widget will turn to writable and the placeholder will say `Choose a file name...`.

zachmullen:
Thanks, I see it now. We can iterate on this UI later, it gives us more power for now.

zachmullen:
One other small improvement we could make to the UI is to change "Save" to something more accurate like "Select".

PR #1779, opened 2017-02-08
Worker plugin signals

kotfic:
This moves JobInfoSpec into the headers of the task that is sent to girder-worker.  This should prevent job models from having to put this information into task kwargs.


zachmullen:
Great! Is there a corresponding change in the worker that this depends on?

kotfic:
@zachmullen Yes,  this will need to go in around the same time that https://github.com/girder/girder_worker/pull/101 is merged

zachmullen:
I think we need to add a deprecation warning log message to `utils.jobInfoSpec` when it's called, and add a `deprecated` field to its docstring.

zachmullen:
@kotfic since we are planning to merge this soon, could you make `jobInfoSpec` a no-op and add `:deprecated:` to the docstring?

kotfic:
@zachmullen not sure i understand - we still rely on jobInfoSpec to generate the dict with the necessary info to instantiate the JobManager in girder_worker

plugins/worker/plugin_tests/worker_test.py
```diff
@@ -112,7 +111,6 @@ def testWorker(self):
                 'girder_worker.run', job['args'], job['kwargs']))
 
             self.assertTrue('headers' in sendTaskCalls[0][2])
-            self.assertTrue('jobInfoSpec' in sendTaskCalls[0][2]['headers'])
```
zachmullen: Is this assertion no longer true?

zachmullen:
It seems the item tasks test was doing some mocking that relied on the jobInfoSpec being set in the kwargs. For now, let's back out a19345e and we can address this issue later.

PR #1786, opened 2017-02-15
Add girder-hardened role to Ansible examples

danlamanna:
This adds an experimental variable to the ansible role called `girder_hardened` which is disabled by default (nothing in this PR breaks existing code).

What this does is utilize some of the sandbox features of systemd in our existing systemd service file, as well as add a sample playbook which implements many of the headers from the [OWASP Secure Headers Project](https://www.owasp.org/index.php/OWASP_Secure_Headers_Project#tab=Main) to largely reduce the attack surface of things browsers can control (XSS for example).

Some additional details are in the relevant commit messages.

### DIFF
--- a/devops/ansible/examples/girder-hardened/site.yml
+++ b/devops/ansible/examples/girder-hardened/site.yml
@@ -1,0 +1,64 @@
+---
+
+- hosts: all
+  vars:
+    girder_update: no
+    girder_force: no
+    nginx_vhosts:
+      - listen: "8080 default_server"
+        extra_parameters: |
+          location / {
+            proxy_set_header Host $proxy_host;
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+            proxy_set_header X-Forwarded-Host $host;
+            proxy_set_header X-Forwarded-Proto $scheme;
+            proxy_pass http://localhost:8888/;
+            # Must set the following for SSE notifications to work
+            proxy_buffering off;
+            proxy_cache off;
+            proxy_set_header Connection '';
+            proxy_http_version 1.1;
+            chunked_transfer_encoding off;
+            proxy_read_timeout 600s;
+            proxy_send_timeout 600s;
+          }
+
+  pre_tasks:
+    - name: Update package cache
+      apt:
+        update_cache: yes
+      become: yes
+      become_user: root
+
+  roles:
+    - role: Stouts.mongodb
+      become: yes
+      become_user: root
+    - role: girder
+    - role: geerlingguy.nginx
+      become: yes
+      become_user: root
+
+  post_tasks:
+    - name: Tweak Girder server configurations
+      ini_file:
+        dest="{{ girder_path }}/girder/conf/girder.local.cfg"
+        backup=yes
+        section="{{ item.section }}"
+        option="{{ item.option }}"
+        value="{{ item.value }}"
+      with_items:
+        - { section: "global", option: "server.thread_pool", value: "1000" }
+        - { section: "global", option: "server.socket_port", value: "8888" }
+        - { section: "global", option: "tools.proxy.on", value: "True" }
+        - { section: "server", option: "mode", value: '"production"' }
+
+    - name: restart girder/nginx
+      service:
+        state=restarted
+        name={{ item }}
+      with_items:
+        - girder
+        - nginx
+      become: yes
+      become_user: root
--- a/devops/ansible/examples/girder-hardened/site.yml
+++ b/devops/ansible/examples/girder-hardened/site.yml
@@ -14,6 +14,19 @@
           ssl_certificate_key /etc/ssl/default_server/default_server.key;
           ssl_certificate /etc/ssl/default_server/default_server.pem;
 
+          add_header X-Content-Type-Options nosniff;
+          add_header X-Frame-Options DENY;
+          add_header Content-Security-Policy "style-src 'self' https://fonts.googleapis.com 'sha256-I08hTk63KqFj9IjGtMeORROxIYJYsBEazDOd2hkXWZY='
+                                                                                            'sha256-zB4+dg/iQaGd+pLwCZo03/lJhn/b6UNYzJppzv3EEAc='
+                                                                                            'sha256-h8Ck0CS7cKiFtPMm/8DjdkPXgP8gxqL7G997urcLS28='
+                                                                                            'sha256-MGNMaa1KDEaQ0N9k9XVIdXWtkb9PIkCbTCFGKcQTs7Q=';
+                                              font-src 'self' data: https://fonts.gstatic.com;
+                                              script-src 'self' 'sha256-zB4+dg/iQaGd+pLwCZo03/lJhn/b6UNYzJppzv3EEAc='
+                                                                'sha256-Z36Z11zZSKHfI2n01+6JycFJBmzRbTWWGsVvkXtuvaY=';";
+
+          # add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
+          # add_header Public-Key-Pins
+
           location / {
             proxy_set_header Host $proxy_host;
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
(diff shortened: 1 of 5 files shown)


devops/ansible/examples/girder-hardened/site.yml
```diff
@@ -0,0 +1,88 @@
...
+          add_header Content-Security-Policy "style-src 'self' https://fonts.googleapis.com 'sha256-I08hTk63KqFj9IjGtMeORROxIYJYsBEazDOd2hkXWZY='
+                                                                                            'sha256-zB4+dg/iQaGd+pLwCZo03/lJhn/b6UNYzJppzv3EEAc='
+                                                                                            'sha256-h8Ck0CS7cKiFtPMm/8DjdkPXgP8gxqL7G997urcLS28='
+                                                                                            'sha256-MGNMaa1KDEaQ0N9k9XVIdXWtkb9PIkCbTCFGKcQTs7Q=';
+                                              font-src 'self' data: https://fonts.gstatic.com;
+                                              script-src 'self' 'sha256-zB4+dg/iQaGd+pLwCZo03/lJhn/b6UNYzJppzv3EEAc='
+                                                                'sha256-Z36Z11zZSKHfI2n01+6JycFJBmzRbTWWGsVvkXtuvaY=';";
+
+          # add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
+          # add_header Public-Key-Pins
```
danlamanna: These are headers we would want to support in the future, but require a valid certificate chain (use of the ansible letsencrypt role may be useful here).

devops/ansible/roles/girder/templates/daemon/girder.service.j2
```diff
@@ -10,7 +10,19 @@ Restart=always
...
+{% endif %}
+
+{% if girder_hardened %}
+# Hardened systemd options
+ProtectSystem=full
+ProtectHome=read-only
+ReadWriteDirectories={{ girder_path }} -{{ ansible_user_dir }}/.girder {{ girder_virtualenv|default("") }}
+
+#PrivateTmp=yes
+#PrivateDevices=yes
```
danlamanna: These are additional, useful restrictions that I couldn't get working at the moment.

devops/ansible/roles/girder/templates/daemon/girder.service.j2
```diff
@@ -10,7 +10,19 @@ Restart=always
...
+
+# As of systemd >= 232, use PrivateUsers option and set ProtectSystem=strict
```
danlamanna: User namespacing was added to systemd services in v232, and a more strict `ProtectSystem` setting was introduced which effectively turns the sandboxing into a whitelist instead of a blacklist.

devops/ansible/examples/girder-hardened/site.yml
```diff
@@ -0,0 +1,88 @@
...
+        ssl_certificate_key: "/etc/ssl/default_server/default_server.key"
+        ssl_certificate: "/etc/ssl/default_server/default_server.pem"
+        extra_parameters: |
+          ssl on;
+          ssl_certificate_key /etc/ssl/default_server/default_server.key;
+          ssl_certificate /etc/ssl/default_server/default_server.pem;
+
+          add_header X-Content-Type-Options nosniff;
+          add_header X-Frame-Options DENY;
+          add_header Content-Security-Policy "style-src 'self' https://fonts.googleapis.com 'sha256-I08hTk63KqFj9IjGtMeORROxIYJYsBEazDOd2hkXWZY='
```
danlamanna: Removing our reliance on an external service for our font would make the CSP much simpler, but it's not a goal of this PR.

devops/ansible/examples/girder-hardened/site.yml
```diff
@@ -0,0 +1,88 @@
...
+                                                                                            'sha256-zB4+dg/iQaGd+pLwCZo03/lJhn/b6UNYzJppzv3EEAc='
+                                                                                            'sha256-h8Ck0CS7cKiFtPMm/8DjdkPXgP8gxqL7G997urcLS28='
+                                                                                            'sha256-MGNMaa1KDEaQ0N9k9XVIdXWtkb9PIkCbTCFGKcQTs7Q=';
+                                              font-src 'self' data: https://fonts.gstatic.com;
```
danlamanna: JSONEditor is responsible for the `data:` directive I believe, it's worth looking into if this can be mitigated.

danlamanna:
This is ready for review, keep in mind this is an example playbook that doesn't change any of the existing behavior.

I've added (quite a few) additional components to a separate issue that can be addressed in the future. See #2197.

danlamanna:
@brianhelba Mind reviewing this?

brianhelba:
Sure.

devops/ansible/examples/girder-hardened/site.yml
```diff
@@ -0,0 +1,97 @@
...
+      become_user: root
+    - role: geerlingguy.nginx
+      become: yes
+      become_user: root
+
+  post_tasks:
+    - name: Tweak Girder server configurations
+      ini_file:
+        dest="{{ girder_path }}/girder/conf/girder.local.cfg"
+        backup=yes
```
brianhelba: These lines should use YAML syntax: `backup: yes`, etc.

devops/ansible/examples/girder-hardened/site.yml
```diff
@@ -0,0 +1,97 @@
...
+    girder_force: no
+    girder_always_build_assets: yes
+    girder_web_extra_args: "--all-plugins"
+    girder_hardened: yes
+    nginx_vhosts:
+      - listen: "8080 default_server"
+        ssl_certificate_key: "/etc/ssl/default_server/default_server.key"
+        ssl_certificate: "/etc/ssl/default_server/default_server.pem"
+        server_name: "girder-hardened"
+        extra_parameters: |
```
brianhelba: Could we store this in a file and load it into `extra_parameters` using a [lookup](http://docs.ansible.com/ansible/playbooks_lookups.html)? That would be a lot cleaner than this inline string.

Additionally, I wonder if some of this code duplication (particularly for setting up the proxy) could be reduced with the [`assemble` module](http://docs.ansible.com/ansible/assemble_module.html).

devops/ansible/examples/girder-hardened/site.yml
```diff
@@ -0,0 +1,97 @@
...
+        value="{{ item.value }}"
+      with_items:
+        - { section: "global", option: "server.thread_pool", value: "1000" }
+        - { section: "global", option: "server.socket_port", value: "8888" }
+        - { section: "global", option: "tools.proxy.on", value: "True" }
+        - { section: "server", option: "mode", value: '"production"' }
+
+    - name: restart girder/nginx
+      service:
+        state=restarted
```
brianhelba: Change this to YAML syntax.

devops/ansible/examples/girder-hardened/site.yml
```diff
@@ -0,0 +1,97 @@
...
+          add_header X-Content-Type-Options nosniff;
+          add_header X-Frame-Options DENY;
+          add_header Content-Security-Policy "style-src 'self'
+                                                        'sha256-I08hTk63KqFj9IjGtMeORROxIYJYsBEazDOd2hkXWZY='
+                                                        'sha256-zB4+dg/iQaGd+pLwCZo03/lJhn/b6UNYzJppzv3EEAc='
+                                                        'sha256-h8Ck0CS7cKiFtPMm/8DjdkPXgP8gxqL7G997urcLS28='
+                                                        'sha256-MGNMaa1KDEaQ0N9k9XVIdXWtkb9PIkCbTCFGKcQTs7Q=';
+                                              font-src 'self';
+                                              script-src 'self'
+                                                         'sha256-MKRiXCHIn6830x4q3pvDhTPRHXpThPnEVF6dkhVBqm8='";
```
brianhelba: Do you have any thoughts for how to remember to keep this updated in the event that these inline scripts / styles ever change?

danlamanna: Eliminating inline scripts/styles as much as possible is the best idea I have (which is almost done). The JSON metadata editor brings in ace-editor which is what's dynamically creating the inline style tags.

zachmullen:
Should this be merged?

danlamanna:
It's passed back to @brianhelba.

danlamanna:
tree-rings-diagram

PR #1787, opened 2017-02-16
Allow to specify the type of the elements in an array passed as an API arg

Xarthisius:
This PR allows to properly specify the type of elements for a query argument which is an `array`. Moreover, the argument is automatically converted into a python list whenever `autoDescribeRoute` is used.

### DIFF
--- a/girder/api/describe.py
+++ b/girder/api/describe.py
@@ -22,6 +22,7 @@ import dateutil.parser
 import os
 import six
 import cherrypy
+import functools
 
 from girder import constants
 from girder.api.rest import getCurrentUser, RestException, getBodyJson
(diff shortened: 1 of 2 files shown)


zachmullen:
@Xarthisius I think this functionality is a subset of the jsonParam functionality, and we have plans in the near future to incorporate [JSON schema](http://json-schema.org/documentation.html)-based validation into that method. Is there a reason to prefer this format, or would using JSON with schema validation cover your use case?

Xarthisius:
Nope, `jsonParam` and JSON schema-based validation would also work. However, girder do accept `dataType='array'` now and I was confused that I got a string instead of python list.

zachmullen:
Let's hold off on this, I'll implement the json schema support hopefully this week, which will supersede this format.

PR #1791, opened 2017-02-17
Fix AccessControlledModel.prefixSearch signature

Xarthisius:
As per `AccessControlledModel.prefixSearch`'s docstring it should accept the same arguments as `Model.prefixSearch`. Currently it's missing `prefixSearchFields`

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -1330,7 +1330,7 @@ class AccessControlledModel(Model):
             cursor, user=user, level=level, limit=limit, offset=offset)
 
     def prefixSearch(self, query, user=None, filters=None, limit=0, offset=0,
-                     sort=None, fields=None, level=AccessType.READ):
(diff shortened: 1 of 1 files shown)


girder/models/model_base.py
```diff
@@ -1330,7 +1330,7 @@ def textSearch(self, query, user=None, filters=None, limit=0, offset=0,
             cursor, user=user, level=level, limit=limit, offset=offset)
 
     def prefixSearch(self, query, user=None, filters=None, limit=0, offset=0,
-                     sort=None, fields=None, level=AccessType.READ):
+                     sort=None, fields=None, prefixSearchFields=None, level=AccessType.READ):
```
zachmullen: Let's put this new kwarg last since otherwise it would theoretically break backward compatibility.

zachmullen:
thanks @Xarthisius !

PR #1792, opened 2017-02-21
Adds a new job listing page under Admin Console for admin users to see all jobs

matthewma7:
Update: See new comment for new implmenetation

------

Adds a checkbox on jobs page for admins users. When checked, all user jobs and anonymous jobs will be shown.

Adds three helper functions in misc module for URL hash parameter manipulation.

--------

For admin users, the existing **My Jobs** drop down will say **Jobs** instead. A checkbox will be shown at the top of the page. 
admin user 1

When checked, all user jobs and anonymous jobs will be shown. The check state will be preserved in the URL as hash parameter.
admin user 2

For non-admin user, the page remains functionally the same.
non-admin user


plugins/jobs/plugin_tests/jobsSpec.js
```diff
@@ -454,5 +457,44 @@ $(function () {
               expect(fired).toBe(true);
           });
       });
+
+        it('Test show all jobs option', function () {
```
jeffbaumes: `it('Job list widget show all jobs.', ...` would be a more consistent test name.

jeffbaumes:
When "Show all jobs" is checked, it seems there should be a (sortable/filterable) column for the user who submitted the job, which is a link to that person's homepage. Would that be difficult to add to this PR?

mgrauer:
For the UI text, I think it isn't clear what "show all jobs" means when unchecked.  Is it only the first 10, or only my jobs, or only jobs that have finished?

I'd prefer "show only my jobs" as the text, which has the opposite meaning when checked to "show all jobs" (just saying that you'd need to reverse the filter for this text).

fogleman:
I don't love the checkbox UI either - nor the URL including user/<ID> even though it's actually showing all the jobs.

I'd suggest an entirely different approach. `Admin Console > Jobs` And perhaps this new page can later include sorting/filtering as Jeff suggested.

matthewma7:
Jeff suggestion to put the all job page under the  **Admin console**, alongside with the **Server configuration** and **Plugins**. I think that is a much cleaner way to organize admin and non-admin stuff. 
I'll work on moving the page to the **Admin console** and leave the existing **my jobs** intact.

matthewma7:
A new page as been added to the Admin Console and it will list all jobs. 

@manthey and I think providing user information of the jobs would be really helpful for the new jobs page but because it may involve few other things we would create a separate issue to elaborate on it.

------

A new option available under Admin Console
chrome_2017-02-23_11-28-52

A separate job list page that list all jobs
chrome_2017-02-23_11-45-59

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -24,21 +24,28 @@ var JobListWidget = View.extend({
...
+        if (this.showAllJobs) {
+            this.collection.resourceName = 'job/all';
+        }
         this.collection.sortField = settings.sortField || 'created';
         this.collection.sortDir = settings.sortDir || SORT_DESC;
         this.collection.pageLimit = settings.pageLimit || this.collection.pageLimit;
 
         this.collection.on('g:changed', function () {
+            this.typeFilter = {};
+            this.statusFilter = {};
```
manthey: It would be better to not reset the filters on a page change.  Just commenting out these two `= {}` lines seems to behave  more like I'd expect.  If a type or status isn't on the new page, it will be reenabled when you navigate back (it would ideally remember the setting, but that doesn't have to change in this PR).

matthewma7: Hi @manthey, I see what you mean with the behavior. I made this change is because I thought I was experiencing missing records (but it was actually filtered out). Maybe it is the filter checkbox needs to be fixed instead.

manthey:
@matthewma7 Thanks for this feature; I'm already finding it useful.

matthewma7:
Thanks for the discussion and suggestion. 😊

PR #1801, opened 2017-02-23
Fix race condition in item_tasks.tasks test

zachmullen:
We were clobbering our job model by saving it on the test
main thread after it had been saved on the async events thread.


zachmullen:
I re-ran this several times and they all passed. PTAL

PR #1805, opened 2017-02-24
Require that the static root be routeable.

manthey:
Before, if you tried to run the server routetable test twice, teh second time would fail since there would be no initial static route.

### DIFF
--- a/girder/models/setting.py
+++ b/girder/models/setting.py
@@ -308,8 +308,15 @@ class Setting(Model):
             if key not in doc['value'] or not doc['value'][key]:
                 raise ValidationException('Girder and static root must be routeable.')
 
-        if not all(route.startswith('/') for route in nonEmptyRoutes):
-            raise ValidationException('Routes must begin with a forward slash.')
+        for key in doc['value']:
+            if (key != GIRDER_STATIC_ROUTE_ID and doc['value'][key] and
+                    not doc['value'][key].startswith('/')):
+                raise ValidationException('Routes must begin with a forward slash.')
+        if doc['value'].get(GIRDER_STATIC_ROUTE_ID):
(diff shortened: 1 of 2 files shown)


zachmullen:
The original intent of the static root setting in the config file was to make it easy to offload our content to a CDN. With all these recent changes, is that still simple to do?

manthey:
That should still work as before.  Though, perhaps, we should create explicit tests for it (in a different PR).

This PR was to address that if you ran the server_routetable test twice, the second time it would fail, as the module utility/server.py setup function expects the static root to be set in the route table, but it had been cleared in the test database (we drop test databases, but only *after* setUpModule is called, which is too late in this instance).

@danlamanna pointed out that we may not want to require a static root for cases where Girder is used just as server and not for the web client.  If that is the case, we can change the behavior wherever ` routeTable[GIRDER_STATIC_ROUTE_ID]` is expected to exist and change this test.  I'll leave that for him to comment on -- this PR continues to be appropriate until such a change is done.

danlamanna:
I looked at the old code, and it seems the `routeTable` implementation maintained the previous behavior, which was to mount `/static` even if Girder wasn't being built with client-side assets. For proof of this, run Girder **without doing** a `girder-install web` and go to http://localhost:8080/static/built/.gitignore. 

This PR only fixes things, so I say merge it and we'll open an issue about how to deal with when Girder should mount static files (which is trickier than it sounds).

manthey:
Any idea what is up with codecov's statement of "No report found to compare against"?

jbeezley:
https://docs.codecov.io/docs/error-reference#section-missing-base-commit

It's probably because of the test failures on master.  Now that #1801 is merged, I suspect a rebase or merge would fix it.

manthey:
I broadened the scope of the PR to address issue #1806.

PR #1795, opened 2017-02-22
Support JSON Schema for REST input validation

zachmullen:
Fixes #1745.

PR #1807, opened 2017-02-24
Add new requirements into requirements.txt

zachmullen:
@cjh1 this is the commit I forgot to push on #1795

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,8 @@ bcrypt==2.0.0
 boto==2.40.0
 cffi==1.8.3
 CherryPy==6.0.1
+functools32==3.2.3.post2
+jsonschema==2.6.0
 Mako==1.0.4
 pymongo==3.2.2
 PyYAML==3.11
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -3,6 +3,8 @@ bcrypt==2.0.0
 boto==2.40.0
 cffi==1.8.3
 CherryPy==6.0.1
+functools32==3.2.3.post2
```
cjh1: Whats this needed for?

zachmullen: Transitive dependency of jsonschema. This file is the dump of a `pip freeze` in a known working state.

cjh1: OK, that is how we keep this up to date?

zachmullen: This file is not actually used in CI or during pip install, the authoritative listing is in setup.py. @jbeezley can better explain the purpose of requirements.txt.

jbeezley: I'm planning to do a D&A lightning talk about it some time.

PR #1810, opened 2017-02-24
Disable extract-text-loader-plugin

zachmullen:
It turns out it was roughly doubling our build time, and also
causing builds to hang on certain plugins for reasons we
could not discover. Disabling it seems to be the best option.

### DIFF
--- a/clients/web/test/testUtils.js
+++ b/clients/web/test/testUtils.js
@@ -653,15 +653,11 @@ girderTest.addScripts = function (scripts) {
 };
 
 /**
- * Import a CSS file into the runtime context.
+ * Import a CSS file into the runtime context. Now that we
+ * no longer have CSS files per plugin, this is a no-op.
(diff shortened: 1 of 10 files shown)


jbeezley:
I'm not seeing much improvement from this.  I updated to this branch and reran the build, and I was still seeing ~3 minute build time for HistomicsTK.  On a whim, I tried to delete node_modules and update to node v6.10 and npm 4.3.  Enabling profiling gave me the following:

| bundle | master | this branch |
| ------ | ------ | ------ |
| core_lib | 15.5 s | 15.1 s |
| HistomicsTK | 2.63 s | 2.62 s |

I'm not sure why HistomicsTK is now fast, but I don't get any measurable difference in speed either way.

zachmullen:
Weird, wonder why I was seeing such drastic differences on my machine. I'm going to dig further into this, it is probably a red herring.

PR #1813, opened 2017-02-27
Add codecov configuration disabling comments

danlamanna:


jbeezley:
@danlamanna I added an extra option to account for the small error in the coverage per run due to asynchronous tests.

@zachmullen The +/-1% value is arbitrary.  Feel free to suggest other changes.  I found the best place to look for documentation is here: <https://docs.codecov.io/docs/commit-status>.

PR #1814, opened 2017-02-27
DB permission should be update when starting  mangoDB on  Ubuntu16.04

LBTINGLB:
You need to ensure that your ‘mongo’ user should has ownership of that file
Job for mongod.service failed because the control process exited with error code.

### DIFF
--- a/docs/prerequisites.rst
+++ b/docs/prerequisites.rst
@@ -81,6 +81,7 @@ With Ubuntu 16.04, create a systemd init script at
 
 and start it with: ::
 
+  sudo chmod 600 var/lib/mongodb/
   sudo service mongod start
 
 Enable the Node.js APT repository: ::
--- a/docs/prerequisites.rst
+++ b/docs/prerequisites.rst
@@ -81,7 +81,7 @@ With Ubuntu 16.04, create a systemd init script at
 
 and start it with: ::
 
-  sudo chmod 600 var/lib/mongodb/
+  sudo chmod 600 /var/lib/mongodb/
   sudo service mongod start
 
 Enable the Node.js APT repository: ::


LBTINGLB:
@jeffbaumes @manthey Would you please kindly review this MR? 

Haocheng

docs/prerequisites.rst
```diff
@@ -81,6 +81,7 @@ With Ubuntu 16.04, create a systemd init script at
 
 and start it with: ::
 
+  sudo chmod 600 var/lib/mongodb/
```
manthey: Shouldn't this be `/var/lib/mongodb` (with a leading slash)?

And, if we are updating information on installing Mongo, should we update the key server and apt repository information to the current values?

LBTINGLB: Nice catch Manthey. I was using this command from root path so it worked.

docs/prerequisites.rst
```diff
@@ -81,6 +81,7 @@ With Ubuntu 16.04, create a systemd init script at
 
 and start it with: ::
 
+  sudo chmod 600 /var/lib/mongodb/
```
zachmullen: Is there a rationale for removing all permissions from other users of the system? If not, I'd suggest just

    sudo chmod u+rw /var/lib/mongodb/

Also, does this need to be applied recursively? If so add `-R` to the command.

zachmullen:
@LBTINGLB we are planning to remove low-level specifics about mongo installation from our own docs due to that being out of scope, so I'm closing this for now.

docs/prerequisites.rst
LBTINGLB: If I remember correctly, I think so.

LBTINGLB:
@zachmullen Sorry for the late reply. Got it.

PR #1815, opened 2017-02-27
Fix stylus loader bug that caused builds to hang

zachmullen:
This was fixed in https://github.com/shama/stylus-loader/pull/166 but
has not yet been merged and released. This is the only way I
discovered to workaround this issue.

A longer term solution would be to change our webpack configuration to
the new structure, but that's more than I want to undertake right now.


msmolens:
Thanks @zachmullen! 🎷

PR #1818, opened 2017-03-01
Add Box to Oauth providers

sutartmelson:
Expanding the Oauth Plugin, this will add Box to the list of default providers.

### DIFF
--- a/plugins/oauth/web_client/stylesheets/oauthLoginView.styl
+++ b/plugins/oauth/web_client/stylesheets/oauthLoginView.styl
@@ -101,3 +101,16 @@
 
   .g-oauth-button-icon
     border-right-color darken($brandColor, 30%)
+
+.g-oauth-button-box
+  $brandColor = #205081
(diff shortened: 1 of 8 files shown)


zachmullen:
Just tested this and it works. Thanks @sutartmelson !

plugins/oauth/web_client/stylesheets/oauthLoginView.styl
```diff
@@ -101,3 +101,16 @@
 
   .g-oauth-button-icon
     border-right-color darken($brandColor, 30%)
+
+.g-oauth-button-box
+  $brandColor = #205081
```
zachmullen: Is this their official color? It looks different from the one I see predominantly on their website.

sutartmelson: Doesn't look like it. I'll try to find their official brand color and change that.

PR #1819, opened 2017-03-01
Re-implements the GUI for Collection Creation Policy on System Configuration page

matthewma7:
@zachmullen pointed out the current GUI for editing Collection Creation Policy (CCP) is could use a more user-friendly GUI. Since the current AccessWidget is handling adding users and groups to a list, as suggested by @zachmullen and @manthey, I added few more options to AccessWidget to support the use case for CCP.
A server-side API is added for AccessWidget for retrieving corresponding names of user id and group id.

When CCP enabled
chrome_2017-03-01_13-46-52

When it's not enabled
chrome_2017-03-01_13-47-19

### DIFF
--- a/clients/web/src/models/CollectionCreationPolicyModel.js
+++ b/clients/web/src/models/CollectionCreationPolicyModel.js
@@ -1,0 +1,25 @@
+import _ from 'underscore';
+import AccessControlledModel from 'girder/models/AccessControlledModel';
+import { restRequest } from 'girder/rest';
+
+var CollectionCreationPolicyModel = AccessControlledModel.extend({
+    resourceName: 'system/setting/collection_creation_policy',
+    fetchAccess: function () {
+        return restRequest({
+            path: this.resourceName + '/access',
+            type: 'GET'
+        }).done(_.bind(function (resp) {
+            if (resp.access) {
+                this.set(resp);
+            } else {
+                this.set('access', resp);
+            }
+            this.trigger('g:accessFetched');
+            return resp;
+        }, this)).error(_.bind(function (err) {
+            this.trigger('g:error', err);
+        }, this));
+    }
+});
+
+export default CollectionCreationPolicyModel;
--- a/clients/web/src/models/CollectionCreationPolicyModel.js
+++ b/clients/web/src/models/CollectionCreationPolicyModel.js
@@ -9,11 +9,7 @@ var CollectionCreationPolicyModel = AccessControlledModel.extend({
             path: this.resourceName + '/access',
             type: 'GET'
         }).done(_.bind(function (resp) {
-            if (resp.access) {
-                this.set(resp);
-            } else {
-                this.set('access', resp);
-            }
+            this.set('access', resp);
             this.trigger('g:accessFetched');
             return resp;
         }, this)).error(_.bind(function (err) {
--- a/clients/web/src/models/CollectionCreationPolicyModel.js
+++ b/clients/web/src/models/CollectionCreationPolicyModel.js
@@ -1,4 +1,3 @@
-import _ from 'underscore';
 import AccessControlledModel from 'girder/models/AccessControlledModel';
 import { restRequest } from 'girder/rest';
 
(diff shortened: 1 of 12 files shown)


clients/web/src/templates/body/systemConfiguration.pug
```diff
@@ -55,11 +55,16 @@ form.g-settings-form(role="form")
...
           defaults['core.add_to_group_policy']) === "yesmod"))
           | Yes, by group administrators and moderators unless disabled per group
-    .g-collection-create-policy-container.g-settings-form-container
-      label(for="g-core-collection-create-policy") Collection creation policy
-      textarea#g-core-collection-create-policy.form-control
-        = JSON.stringify(settings['core.collection_create_policy'] || defaults['core.collection_create_policy'], null, 4)
-      .g-search-container
+    .g-collection-create-policy-container
+      label 
+        | Collection creation policy
```
zachmullen: There's a shortcut for this in pug, which is just

    label Collection creation policy

clients/web/src/templates/body/systemConfiguration.pug
```diff
@@ -55,11 +55,16 @@ form.g-settings-form(role="form")
...
+      p.
+        Allow additional users and groups to create collection.
+      input.g-plugin-switch(type="checkbox",
+          checked="checked", data-size="mini", data-on-color="primary")
+      .access-widget-container
+      textarea#g-core-collection-create-policy(style={'display':'none'})
```
zachmullen: We can use the handy `.hide` class instead of setting the style explicitly.

zachmullen:
Just tested this and it works great. Thanks @matthewma7 !

matthewma7:
I think this PR is ready for review. 
@zachmullen, I didn't change much code after you tested, only committed the change you suggested and other small change. 
@manthey Thank you for help me tracking down the timing issue I have with frontend testing.

clients/web/src/stylesheets/body/systemConfig.styl
```diff
@@ -28,6 +32,14 @@
   // this class is deprecated
   @extend .g-settings-form-container
 
+  .access-widget-container
+    width: 420px
+    margin-top: 15px;
+
+    .g-grant-access-container
+      div.form-group
+        margin-bottom: 0
```
zachmullen: For consistency, let's remove the `:` separators for stylus files.

matthewma7: Oh, thanks. I didn't notice the stylus syntax at all.

clients/web/src/models/CollectionCreationPolicyModel.js
```diff
@@ -0,0 +1,21 @@
...
+import AccessControlledModel from 'girder/models/AccessControlledModel';
+import { restRequest } from 'girder/rest';
+
+var CollectionCreationPolicyModel = AccessControlledModel.extend({
+    resourceName: 'system/setting/collection_creation_policy',
+    fetchAccess: function () {
+        return restRequest({
+            path: this.resourceName + '/access',
+            type: 'GET'
+        }).done(_.bind(function (resp) {
```
zachmullen: Since we have ES6 features, the use of `_.bind` is (usually) not required, instead we should just use arrow functions.

clients/web/src/models/CollectionCreationPolicyModel.js
```diff
@@ -0,0 +1,21 @@
...
+            this.set('access', resp);
+            this.trigger('g:accessFetched');
+            return resp;
+        }, this)).error(_.bind(function (err) {
```
zachmullen: Same here

matthewma7: I see. I just copied from existing code. Good to know we want to keep close to the standard.

zachmullen: Makes sense. We didn't replace all the existing occurrences of it, however it's good to clean them up when you encounter them, and not add any new ones unless they're necessary. 😄

zachmullen:
@matthewma7 you may click the Big Green Button at your convenience. Make sure to delete the branch afterward. Thanks!

PR #1823, opened 2017-03-02
Improve girder cli

jcfr:
This PR aims at improving the user overall user experience associated with the Girder CLI.

### Why ?

As a user, I initially got frustrated with the behaviour of the CLI. I couldn't figure out the correct combination of scheme, host, ... and had to do some guess work.

Here is my take at improving the user experience.

### Details

* Documentation
  * Reference Girder CLI documentation from the user documentation
  * Add a "Specifying Credentials" section to the python client documentation
* girder client
  * allow api-key to be specified using `GIRDER_API_KEY` env. variable
  * api-root can be specified as `api/v1` or `/api/v1`
  * scheme defaults to `https` allowing to simply specify the host when using production instance supporting https
  * display default value in help string
  * support auto-completion when using bash
  * report download and upload progress (for both multipart and non-multipart implementation)
  * download local folder now default to current directory

* girder client (deprecated functionalities)
  * specifying URL by part is now deprecated

* Internal
  * Introduce dependency to [click python library](http://pocco-click.readthedocs.io/en/latest/). A modern and flexible command line parser.
  * Introduce dependency to [toolbelt](https://toolbelt.readthedocs.io/en/latest/uploading-data.html#monitoring-your-streaming-multipart-upload) so that progress for multipart upload is reported.

### What is next ? 

* <s>Progress reporting leveraging the facilities offered by [click](http://pocco-click.readthedocs.io/en/latest/utils.html#showing-progress-bars)</s>
* Better error message in case of error. No need to report a stack trace when authentication failed or a folder if do not exist

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -173,7 +173,8 @@ class GirderClient(object):
             if not apiRoot:
                 apiRoot = self.DEFAULT_API_ROOT
             # If needed, prepend '/'
-            apiRoot = '/' + apiRoot if apiRoot[0] != '/' else apiRoot
+            if not apiRoot.startswith('/'):
+                apiRoot = '/' + apiRoot
 
             self.scheme = scheme or self.DEFAULT_SCHEME
             self.host = host or self.DEFAULT_HOST
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -107,6 +107,30 @@ class _NoopProgressReporter(object):
         pass
 
 
+# Used for fast non-multipart upload
+class _ProgressBytesIO(six.BytesIO):
+    def __init__(self, *args, **kwargs):
+        self.reporter = kwargs.pop('reporter')
+        six.BytesIO.__init__(self, *args, **kwargs)
+
+    def read(self, _size):
+        _chunk = six.BytesIO.read(self, _size)
+        self.reporter.update(len(_chunk))
+        return _chunk
+
+
+# Used for deprecated multipart upload
+class _ProgressMultiPartEncoder(MultipartEncoder):
+    def __init__(self, *args, **kwargs):
+        self.reporter = kwargs.pop('reporter')
+        MultipartEncoder.__init__(self, *args, **kwargs)
+
+    def read(self, _size):
+        _chunk = MultipartEncoder.read(self, _size)
+        self.reporter.update(len(_chunk))
+        return _chunk
+
+
 class GirderClient(object):
     """
     A class for interacting with the Girder RESTful API.
@@ -796,20 +820,6 @@ class GirderClient(object):
 
         with self.progressReporterCls(label=uploadObj['name'], length=size) as reporter:
 
-            # Used for fast non-multipart upload
-            class _ProgressBytesIO(six.BytesIO):
-                def read(self, _size):
-                    _chunk = six.BytesIO.read(self, _size)
-                    reporter.update(len(_chunk))
-                    return _chunk
-
-            # Used for deprecated multipart upload
-            class _ProgressMultiPartEncoder(MultipartEncoder):
-                def read(self, _size):
-                    _chunk = MultipartEncoder.read(self, _size)
-                    reporter.update(len(_chunk))
-                    return _chunk
-
             while True:
                 chunk = stream.read(min(self.MAX_CHUNK_SIZE, (size - offset)))
 
@@ -822,7 +832,7 @@ class GirderClient(object):
                 if self.getServerVersion() >= ['2', '2']:
                     uploadObj = self.post(
                         'file/chunk?offset=%d&uploadId=%s' % (offset, uploadId),
-                        data=_ProgressBytesIO(chunk))
+                        data=_ProgressBytesIO(chunk, reporter=reporter))
                 else:
                     # Prior to version 2.2 the server only supported multipart uploads
                     parameters = {
@@ -831,6 +841,7 @@ class GirderClient(object):
                     }
 
                     m = _ProgressMultiPartEncoder(
+                        reporter=reporter,
                         fields={'chunk': ('chunk', chunk, 'application/octet-stream')},
                     )
 
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -166,7 +166,9 @@ class GirderClient(object):
             to pass 443 for the port
         :param cacheSettings: Settings to use with the diskcache library, or
             None to disable caching.
-        :param progressReporterCls: the progress reporter class to instantiate.
+        :param progressReporterCls: the progress reporter class to instantiate. This class
+            is expected to be a context manager with a constructor accepting `label` and
+            `length` keyword arguments and an `update` method accepting a `chunkSize` argument.
             This defaults to :class:`_NoopProgressReporter`.
         """
         if apiUrl is None:
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -196,7 +196,9 @@ class GirderClient(object):
             None to disable caching.
         :param progressReporterCls: the progress reporter class to instantiate. This class
             is expected to be a context manager with a constructor accepting `label` and
-            `length` keyword arguments and an `update` method accepting a `chunkSize` argument.
+            `length` keyword arguments, an `update` method accepting a `chunkSize` argument and
+            a class attribute `reportProgress` set to True (It can conveniently be
+            initialized using `sys.stdout.isatty()`).
             This defaults to :class:`_NoopProgressReporter`.
         """
         if apiUrl is None:
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -782,41 +782,51 @@ class GirderClient(object):
         """
         offset = 0
         uploadId = uploadObj['_id']
-        while True:
-            chunk = stream.read(min(self.MAX_CHUNK_SIZE, (size - offset)))
 
-            if not chunk:
-                break
+        with self.progressReporterCls(label=uploadObj['name'], length=size) as reporter:
 
-            if isinstance(chunk, six.text_type):
-                chunk = chunk.encode('utf8')
+            class _ProgressBytesIO(six.BytesIO):
+                def read(self, _size):
+                    _chunk = super(_ProgressBytesIO, self).read(_size)
+                    reporter.update(len(_chunk))
+                    return _chunk
 
-            if self.getServerVersion() >= ['2', '2']:
-                uploadObj = self.post(
-                    'file/chunk?offset=%d&uploadId=%s' % (offset, uploadId), data=chunk)
-            else:
-                # Prior to version 2.2 the server only supported multipart uploads
-                parameters = {
-                    'offset': offset,
-                    'uploadId': uploadId
-                }
+            while True:
+                chunk = stream.read(min(self.MAX_CHUNK_SIZE, (size - offset)))
 
-                uploadObj = self.post('file/chunk', parameters=parameters, files={
-                    'chunk': chunk
-                })
+                if not chunk:
+                    break
 
-            if '_id' not in uploadObj:
-                raise Exception(
-                    'After uploading a file chunk, did not receive object with _id. Got instead: ' +
-                    json.dumps(uploadObj))
+                if isinstance(chunk, six.text_type):
+                    chunk = chunk.encode('utf8')
 
-            offset += len(chunk)
+                if self.getServerVersion() >= ['2', '2']:
+                    uploadObj = self.post(
+                        'file/chunk?offset=%d&uploadId=%s' % (offset, uploadId),
+                        data=_ProgressBytesIO(chunk))
+                else:
+                    # Prior to version 2.2 the server only supported multipart uploads
+                    parameters = {
+                        'offset': offset,
+                        'uploadId': uploadId
+                    }
 
-            if callable(progressCallback):
-                progressCallback({
-                    'current': offset,
-                    'total': size
-                })
+                    uploadObj = self.post('file/chunk', parameters=parameters, files={
+                        'chunk': chunk
+                    })
+
+                if '_id' not in uploadObj:
+                    raise Exception(
+                        'After uploading a file chunk, did not receive object with _id. Got instead: ' +
+                        json.dumps(uploadObj))
+
+                offset += len(chunk)
+
+                if callable(progressCallback):
+                    progressCallback({
+                        'current': offset,
+                        'total': size
+                    })
 
         if offset != size:
             self.delete('file/upload/' + uploadId)
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -30,6 +30,8 @@ import shutil
 import six
 import tempfile
 
+from requests_toolbelt import MultipartEncoder
+
 __version__ = '2.0.0'
 __license__ = 'Apache 2.0'
 
@@ -327,7 +329,8 @@ class GirderClient(object):
 
         return self._serverApiDescription
 
-    def sendRestRequest(self, method, path, parameters=None, data=None, files=None, json=None):
+    def sendRestRequest(self, method, path, parameters=None,
+                        data=None, files=None, json=None, headers=None):
         """
         This method looks up the appropriate method, constructs a request URL
         from the base URL, path, and parameters, and then sends the request. If
@@ -351,6 +354,8 @@ class GirderClient(object):
         :type files: dict
         :param json: A JSON object to send in the request body.
         :type json: dict
+        :param headers: If present, a dictionary of headers to encode in the request.
+        :type headers: dict
         """
         if not parameters:
             parameters = {}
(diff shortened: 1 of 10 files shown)


zachmullen:
@jcfr we introduced the `--api-url` option to alleviate the pain of passing host/scheme/api root/port, we just kept the old options around for backward compatibility.

jcfr:
> option to alleviate the pain of passing host/scheme/api root/port, we just kept the old options around for backward compatibility.

Now that the `scheme` defaults to `https`, it is quite convenient to only give `--host data.kitware.com` without having to deal with `port` and `api-root`. In my first experience with the CLI, since I couldn't see the expected default values ... I had few failed attempt.

That said, it is also nice to be explicit and always give the full api-url.

I will add a commit removing `scheme`, 'api-root', `port` and `host` from the help text while still accepting them. That way, it is backward compatible and we will have a clear and unambiguous way of specifying the full api url

jcfr:
Et voila, the help of the CLI is now simpler:

```
Usage: girder-cli [OPTIONS] COMMAND [ARGS]...

  Perform common Girder CLI operations.

  The CLI is particularly suited to upload (or download) large, nested
  hierarchy of data to (or from) Girder from (or into) a local directory.

  The recommended way to use credentials is to first generate an api key and
  then specify the ``api-key`` argument or set the ``GIRDER_API_KEY``
  environment variable.

  The client also supports ``username`` and ``password`` args. If only the
  ``username`` is specified, the client will prompt the user to
  interactively input his/her password.

Options:
  --api-url TEXT   RESTful API URL (e.g https://girder.example.com:443/api/v1)
  --api-key TEXT   [default: GIRDER_API_KEY env. variable]
  --username TEXT
  --password TEXT
  -h, --help       Show this message and exit.

Commands:
  download   Download files from Girder
  localsync  Synchronize local folder with remote Girder folder
  upload     Upload files to Girder
```

For backward compatibility, the port/host/scheme/api-root are still supported but not documented in the help string.

danlamanna:
+1 for click, it creates really nice composable CLI programs and has a number of helpers along with it (progress bars, plugin system, etc). I believe @kotfic was thinking of doing a lightning talk on it.

zachmullen:
@jcfr is this ready for review?

jcfr:
@zachmullen  Here is is, I just pushed what I did for the download progress.

As we discussed, I will wait for the refactoring of the upload before working on the upload progress.

Here are some pointers that could be useful for the upload:
* https://gist.github.com/yejianye/3b08730d4ed31ae18c7d
* https://toolbelt.readthedocs.io/en/latest/uploading-data.html#monitoring-your-streaming-multipart-upload
* http://stackoverflow.com/questions/13909900/progress-of-python-requests-post#14953090

clients/python/girder_client/__init__.py
```diff
@@ -148,12 +168,15 @@ def __init__(self, host=None, port=None, apiRoot=None, scheme=None, apiUrl=None,
             None to disable caching.
         """
         if apiUrl is None:
-            if apiRoot is None:
-                apiRoot = '/api/v1'
+            if not apiRoot:
+                apiRoot = self.DEFAULT_API_ROOT
+            # If needed, prepend '/'
+            apiRoot = '/' + apiRoot if apiRoot[0] != '/' else apiRoot
```
zachmullen: How about this instead:

    if not apiRoot.startswith('/'):
        apiRoot = '/' + apiRoot

clients/python/girder_client/__init__.py
```diff
@@ -88,6 +88,21 @@ def __str__(self):
         return super(HttpError, self).__str__() + '\nResponse text: ' + self.responseText
 
 
+class DownloadProgressReporter(object):
```
zachmullen: We should make this an internal symbol, and change its name to something like `_NoopProgressReporter` to indicate that it does nothing.

clients/python/girder_client/__init__.py
```diff
@@ -123,8 +138,13 @@ class GirderClient(object):
...
     MAX_CHUNK_SIZE = 1024 * 1024 * 64
 
+    DEFAULT_API_ROOT = 'api/v1'
+    DEFAULT_HOST = 'localhost'
+    DEFAULT_PORT = 443
+    DEFAULT_SCHEME = 'https'
+
     def __init__(self, host=None, port=None, apiRoot=None, scheme=None, apiUrl=None,
-                 cacheSettings=None):
+                 cacheSettings=None, progressReporterCls=DownloadProgressReporter):
```
zachmullen: Let's make this default to `None` to avoid mutable default, and use this as the default internally. Also, this parameter needs a docstring.

clients/python/girder_client/cli.py
```diff
@@ -49,78 +52,151 @@ def __init__(self, username, password, host=None, port=None, apiRoot=None,
...
+        except HttpError as exc_info:
+            if exc_info.status == 400:
+                continue
+            if sys.version_info[0] >= 3:
+                raise exc_info.with_traceback(sys.exc_info()[2])
+            else:
+                raise (sys.exc_info()[0], sys.exc_info()[1], sys.exc_info()[2])  # noqa: E999
+
+
+def _common_parameters(path_exists=False, path_writable=True,
```
zachmullen: For consistency, let's stick to camelCase for the function names in this library.

clients/python/girder_client/__init__.py
```diff
@@ -898,12 +923,17 @@ def downloadFile(self, fileId, path, created=None):
                 return
 
         # download to a tempfile
+        progressFileName = os.path.basename(path) if isinstance(path, six.string_types) else fileId
```
zachmullen: I find these ternary operators hard to read in python, could we expand it to a normal if/else block?

docs/python-client.rst
```diff
@@ -112,15 +136,53 @@ separated list to the ``--blacklist`` arg ::
...
+
+To download the file(s) associated with a Girder Item with if `58b8eb798d777f0aef5d0f78` under
+the local folder `download_folder`::
+
+    girder-cli download --parent-type item 8b8eb798d777f0aef5d0f78 download_folder
+
+Auto-detecting parent-type
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Both download and upload commands accept a `--parent-type` argument allowing to specify
```
zachmullen: Both the download and upload commands accept a `--parent-type` argument allowing users to specify...

docs/python-client.rst
```diff
@@ -112,15 +136,53 @@ separated list to the ``--blacklist`` arg ::
...
+the type (folder, collection, user, or item) associated with the chosen object id.
+
+If the argument is omitted, the client will conveniently tries to autodetect the type
```
zachmullen: try*

docs/python-client.rst
```diff
@@ -112,15 +136,53 @@ separated list to the ``--blacklist`` arg ::
...
+by iteratively invoking the `resource/%id/path?type=%type` API end point and checking
+if a resource is found.
+
+Note that relying on auto-detection induces an extra time cost.
```
zachmullen: Note that relying on auto-detection incurs extra network requests, which will slow down the script, so it should be avoided for time-sensitive operations.

jcfr:
@zachmullen Thanks for the review. Suggested changes have been integrated.

To avoid merge conflicts, I will add progress reporting for the upload after #1830 is integrated.

jcfr:
Would it make sense to update [uploadFileToItem](https://github.com/girder/girder/blob/55c9410a73eac8ec2c99d68646b34e134e45df42/clients/python/girder_client/__init__.py#L635) method to make use of `_uploadContents` ? That we would have a single pathway to process upload.

It also look the method `uploadFile` could be directory used from within `uploadFileToItem`

jcfr:
@zachmullen This can be integrated as it is. I will publish the upload progress reporting in a different PR after #1830 is also integrated.

jcfr:
@zachmullen @mgrauer Now #1830 is integrated, I suggest we get this one in.

clients/python/girder_client/__init__.py
```diff
@@ -146,14 +166,20 @@ def __init__(self, host=None, port=None, apiRoot=None, scheme=None, apiUrl=None,
             to pass 443 for the port
         :param cacheSettings: Settings to use with the diskcache library, or
             None to disable caching.
+        :param progressReporterCls: the progress reported class to instantiate.
```
zachmullen: reporter*

Let's also add a `:type:` specifier that says the class must be a context manager, and also must have an `update` method that receives a quantity. It's unclear right now whether the quantity is the current amount or an incremental amount; maybe we could document that inside the _NoopProgressReporter class if that's easier.

clients/python/girder_client/__init__.py
```diff
@@ -88,6 +88,21 @@ def __str__(self):
         return super(HttpError, self).__str__() + '\nResponse text: ' + self.responseText
 
 
+class _NoopProgressReporter(object):
+    def __init__(self, label="", length=0):
```
zachmullen: single quotes

clients/python/girder_client/cli.py
```diff
@@ -49,78 +52,151 @@ def __init__(self, username, password, host=None, port=None, apiRoot=None,
...
+        try:
+            client.get('resource/%s/path' % object_id, parameters={'type': parent_type})
+            return parent_type
+        except HttpError as exc_info:
+            if exc_info.status == 400:
+                continue
+            if sys.version_info[0] >= 3:
+                raise exc_info.with_traceback(sys.exc_info()[2])
+            else:
+                raise (sys.exc_info()[0], sys.exc_info()[1], sys.exc_info()[2])  # noqa: E999
```
zachmullen: How is this different from just a simple `raise`?

clients/python/girder_client/cli.py
```diff
@@ -49,78 +52,151 @@ def __init__(self, username, password, host=None, port=None, apiRoot=None,
...
+
+
+def _CommonParameters(path_exists=False, path_writable=True,
+                      additional_parent_types=['collection', 'user']):
```
zachmullen: Avoid mutable default parameters (change to tuple would work)

docs/python-client.rst
```diff
@@ -50,16 +63,27 @@ you wish to connect to. The easiest way to do so is to pass the full URL to the
 REST API of the Girder instance you wish to connect to using the ``api-url``
 argument to ``girder-cli``. For example: ::
 
-    girder-cli --api-url http://localhost:8080/api/v1 <subcommand> ...
+    girder-cli --api-url http://localhost:8080/api/v1 <command> ...
 
-You may also specify the URL in parts, using the ``host`` argument, and optional
-``scheme``, ``port``, and ``api-root`` args. ::
+For backward compatibility, you may also specify the URL in parts, using the ``host`` argument,
```
zachmullen: This isn't really related to backward compatibility, just a matter of preference.

docs/python-client.rst
```diff
@@ -50,16 +63,27 @@ you wish to connect to. The easiest way to do so is to pass the full URL to the
...
+and optional ``scheme``, ``port``, and ``api-root`` args.
 
-    girder-cli --host girder.example.com ...
+Specifying credentials
+^^^^^^^^^^^^^^^^^^^^^^
 
-Or... ::
+The recommended way is to generate an :ref:`api key <api_keys>` and specify
```
zachmullen: API key*

tests/cases/py_client/lib_test.py
```diff
@@ -449,7 +450,9 @@ def testDownloadCache(self):
         # create another client with caching enabled
         cacheSettings = {'directory': os.path.join(self.libTestDir, 'cache')}
         client = girder_client.GirderClient(
-            port=os.environ['GIRDER_PORT'], cacheSettings=cacheSettings)
+            port=os.environ['GIRDER_PORT'],
+            scheme="http",
```
zachmullen: quotes

docs/python-client.rst
jcfr: Since the parameter `--host`, `--port`, ... ... are not displayed anymore in the help. Indeed, the promoted approach is `--api-url`. May be I should just remove that sentence ...

clients/python/girder_client/cli.py
jcfr: In that particular case, it is the same. I will update. My understand is that form makes sense only in the case of nested try/except block ...

jcfr:
@zachmullen Et voila, this is ready for final review. Thanks

zachmullen:
It would be great to also show the file size for the file being currently uploaded/downloaded in the progress output, e.g. "2.5 KB / 100 KB".

jcfr:
> It would be great to also show the file size for the file being currently uploaded/downloaded in the progress output, e.g. "2.5 KB / 100 KB".

Makes sense. I will then rebase this topic on top of master and integrate the remaining of my improvement related to upload/download progress reporting.

clients/python/girder_client/__init__.py
```diff
@@ -327,9 +367,13 @@ def sendRestRequest(self, method, path, parameters=None, data=None, files=None,
         url = self.urlBase + path
 
         # Make the request, passing parameters and authentication info
+        _headers = {'Girder-Token': self.token}
+        if type(headers) is dict:
```
zachmullen: `if isinstance(headers, dict)`

clients/python/girder_client/__init__.py
```diff
@@ -749,41 +793,62 @@ def _uploadContents(self, uploadObj, stream, size, progressCallback=None):
...
-                parameters = {
-                    'offset': offset,
-                    'uploadId': uploadId
-                }
+            # Used for deprecated multipart upload
+            class _ProgressMultiPartEncoder(MultipartEncoder):
+                def read(self, _size):
+                    _chunk = MultipartEncoder.read(self, _size)
+                    reporter.update(len(_chunk))
+                    return _chunk
```
zachmullen: Could we move these classes out of the function and up to the module level instead? Makes them easy to mock, and slightly less overhead when running this function.

clients/python/girder_client/cli.py
```diff
@@ -39,8 +41,38 @@ def __init__(self, username, password, host=None, port=None, apiRoot=None,
...
+            def formatPos(_self):
+                pos = formatSize(_self.pos)
+                if _self.length_known:
+                    pos += '/%s' % formatSize(_self.length)
+                return pos
+
+            bar.format_pos = types.MethodType(formatPos, bar)
+            return bar
+
+        _progressBar.reportProgress = isatty(_default_text_stdout())
```
zachmullen: `sys.stdout.isatty()`

jcfr:
:fireworks:

mgrauer:
Great contribution, thanks @jcfr.

🧀 🍷

ronichoudhury:
Thanks for this work, @jcfr.

I realize I'm late to the party but can I ask why we need to deprecate the `--scheme`, `--port`, etc. options? Even if the preferred way is to supply `--api-url` it seems counterintuitive to remove options for simply modifying the defaults. Without them, you force the user to know what the defaults are, and then supply them explicitly, even if they just want to change which port Girder is running on, etc.

To be clear, my proposal is to undeprecate those options; whether to leave them undocumented is a separate issue but personally, I see no problem with documenting them somewhere, after specifying that `--api-url` is the preferred way to do this.

Thanks!

zachmullen:
Addendum: there is a small issue related to specifying the URL by parts via the CLI, which I don't know if it existed before or not. In `GirderClient.__init__`, if no port is explicitly passed and we are constructing the URL by parts, the default port is chosen based on the scheme. However, in the CLI, we set the default port unconditionally to be 443. So if the user specifies just `--scheme=http`, it will still try to use 443 as the port even though that's probably not what the user wants.

jcfr:
> To be clear, my proposal is to undeprecate those options; whether to leave them undocumented is a separate issue but personally, I see no problem with documenting them somewhere, after specifying that --api-url is the preferred way to do this.

There are still supported but not documented. Would it work if we list them at the end of the help in an "Advanced Group" ? 

> if no port is explicitly passed and we are constructing the URL by parts, the default port is chosen based on the scheme.

We should indeed allow the port to be overridden.

Otherwise, [here is the comment](https://github.com/girder/girder/pull/1823#issuecomment-283833409) from Zach from few week ago:

> @jcfr we introduced the --api-url option to alleviate the pain of passing host/scheme/api root/port, we 
> just kept the old options around for backward compatibility.

@zachmullen How do we stand ? Should we maintain (and properly test) these options .. and document them in an advanced group in the CLI help ?

zachmullen:
I want the help docs to be as simple as possible so users aren't confused or overwhelmed. An advanced section is fine, or I'd be ok with leaving them undocumented as they are now.

jcfr:
Sounds good. For now, I will keep them undocumented then. 

@ronichoudhury Out of curiosity, what is the use case for using the "older" API ?

ronichoudhury:
On Tue, Mar 28, 2017 at 3:01 PM Jean-Christophe Fillion-Robin
notifications@github.com <http://mailto:notifications@github.com> wrote:

@ronichoudhury <https://github.com/ronichoudhury> Out of curiosity, what is
> the use case for using the "older" API ?
>
Essentially, this:
https://en.wikipedia.org/wiki/Convention_over_configuration.

If I know that girder-cli “just works” in the common case (which it more or
less does), and I also know that I’m using it in http mode, on port 8080, I
expect to only have to override the “just works” situation with --scheme
http and --port 8080.

If, on the other hand, I have to use --api-url to make *any* change to
these values, then I need to figure out what the “just works” default value
is for *all* of them. This seems like a UI anti-pattern to me.

Hopefully that makes sense?
​

jcfr:
Roger that. I will propose a change.

PR #1825, opened 2017-03-03
Simplify Globus OAuth provider module

Xarthisius:
Instead of parsing the JWT identity token, we now explicitly call an endpoint providing a basic user information. This drops the additional requirement on *cryptography* and *pyjwt*.


setup.py
```diff
@@ -81,7 +81,6 @@ def run(self, *arg, **kw):
     'item_tasks': ['ctk-cli'],
     'thumbnails': ['Pillow', 'pydicom', 'numpy'],
     'worker': ['celery'],
```
zachmullen: Remove comma from line above

zachmullen:
Thanks!

PR #1826, opened 2017-03-03
Store OAuth headers in the user document

Xarthisius:
Up till now, the authentication token obtained in the OAuth2 flow wasonly used to get the user's email/name and was immediately discarded. With this change, auth token is stored in the user model metadata and can be conveniently reused for accessing the issuer's resources.

### DIFF
--- a/plugins/oauth/plugin_tests/oauth_test.py
+++ b/plugins/oauth/plugin_tests/oauth_test.py
@@ -274,10 +274,60 @@ class OauthTest(base.TestCase):
 
         # Try callback for the 'existing' account, which should succeed
         existing = doOauthLogin('existing')
-
         # Try callback for the 'new' account, with open registration
         self.model('setting').set(SettingKey.REGISTRATION_POLICY, 'open')
         new = doOauthLogin('new')
+
+        # Enable storing access token
+        params = {
+            'list': json.dumps([
+                {
+                    'key': PluginSettings.PROVIDERS_ENABLED,
+                    'value': [providerInfo['id']]
+                }, {
+                    'key': providerInfo['client_id']['key'],
+                    'value': providerInfo['client_id']['value']
+                }, {
+                    'key': providerInfo['client_secret']['key'],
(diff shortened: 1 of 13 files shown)


zachmullen:
I see that we are storing this info, but it isn't read anywhere in this PR. How do you plan to use this after storing it? What happens if the cached token is expired?

brianhelba:
Indeed, all of the tokens do or should an expiration date.

In general, I'm not comfortable with apps that gather and store excess user information without a compelling (and transparently communicated) use case for it.

Xarthisius:
My use case specifically: for Globus it allows to access/import external data. As for the expiration date that's going to be checked in the plugin that actually uses the token.

zachmullen:
Ok, sounds good. I restarted the circle build, there was a network error when fetching some npm packages.

brianhelba:
@Xarthisius In keeping with the [Principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege), would it be acceptable to only store this token for the Globus provider? Unless other providers have a particular use case for requiring tokens after login has completed, I see retention of the access token for longer than strictly necessary as a troublesome practice, for both security and privacy.

Xarthisius:
I have additional use cases for github/bitbucket. @brianhelba how about making this behavior configurable in plugin settings'? 

Maybe I'm doing it wrong. What I'm really after is give user the ability to link his/her girder account to external services (once) and be able to access them after each login.

brianhelba:
IMO, a setting to enable this behavior would be perfect.

At the end of the day, an OAuth user does grant the technical permission for Girder to do things on their behalf via a token. Of course, individual instances and external plugins may use or abuse this however they see fit. My concern is just to ensure that, for a less sophisticated administrator setting up a new deployment of Girder+OAuth, Girder is still respecting the principle of least privilege.

zachmullen:
I'm going to close this, because I think #1889 solves this more generally.

PR #1827, opened 2017-03-03
Enable esdoc for automated web client documentation

zachmullen:
@Purg your question earlier made me look into this, and this library seems like a good choice for handling this.

@girder/developers once this is merged to master, we *should* have our JS docs built and hosted for us on https://doc.esdoc.org/

### DIFF
--- a/.esdoc.json
+++ b/.esdoc.json
@@ -1,0 +1,4 @@
+{
+  "source": "./clients/web/src",
+  "destination": "./docs/_build/esdoc"
+}
--- a/circle.yml
+++ b/circle.yml
@@ -31,3 +31,11 @@ test:
   override:
     - ~/girder/cmake/run_circleci.sh:
         parallel: true
+
+deployment:
+  master:
+    branch: master
+    owner: girder
+    commands:
+      - npm run esdoc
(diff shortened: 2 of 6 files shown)


zachmullen:
This also does linting and coverage computation for our JS docs out of the box. I fixed the cases it warned about.

zachmullen:
And for anybody wishing to test locally, run `npm run esdoc && open docs/_build/esdoc/index.html`

jbeezley:
That looks like a nice service.  I do like that it adds easy integration into hosting and checking documentation coverage.

The current output from this tool isn't incredibly useful in my opinion.  It doesn't support backbone style inheritance, so you don't get class level documentation of inherited properties.  It is also  not organizing functions according to modules, only directories.  There are other edge cases that it doesn't (or can't) detect properly like `restRequest`.  It is showing the docstring from `restRequestMock` rather than real method.

I know JSDoc suffers these same problems as well, but it is flexible enough in how it handles tags that you can always override the automatic parsing.  It isn't obvious to me that ESDoc provides the same level of flexibility, especially given it is so young.

That said, ESDoc does use (mostly) the same syntax and tags as JSDoc, so I don't see any harm in enabling it.  The most important issue we have is filling in the docstrings consistently anyway.  We should be able to switch seamlessly to JSDoc if necessary.

Just my 2¢

zachmullen:
Half a loaf!

PR #1828, opened 2017-03-03
Fix minor XSS vulnerability in human-readable REST HTML output

zachmullen:
Thanks to @jjomier for finding this with a pen-testing tool. This is not a major security vulnerability since none of our end user clients would ever encounter it (except perhaps via following a malicious link), but it is a legitimate XSS vector. To see the problem, using your web browser, go to, e.g.

    http://localhost:8080/api/v1/item/some%3Cb%3Efoo%3C/b%3E%3Ci%3Ebar%3C/i%3E

Scope: only when using a web browser to invoke an API route directly (i.e. the `Accept` header is set to `text/html`.

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -17,6 +17,7 @@
 #  limitations under the License.
 ###############################################################################
 
+import cgi
 import cherrypy
 import collections
 import datetime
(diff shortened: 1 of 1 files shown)


danlamanna:
This LGTM, a couple of notes:

1) A good CSP would have stopped this from occurring in the wild (at the deployment level). The CSP included in #1786 would work (though I have to tweak an nginx setting).
2) We should probably enforce that PRs labeled "security" have to be backported to 1.x moving forward.

mgrauer:
Have we specified somewhere for how long we'll support 1.x?

zachmullen:
+1 for CSP. Do you think that should be in girder code or in deployment scripts e.g. nginx config?

danlamanna:
@zachmullen There was some discussion about that in my lightning talk, I lean towards [deployment scripts](https://github.com/girder/girder/pull/1786/files#diff-a4a256b8335e3e2a2d256949ade1ccf9R17). I generally don't want to put a GUI in front of too many security related things.

zachmullen:
Sounds good, sad I could not make your talk. Do you have slides from it?

PR #1829, opened 2017-03-03
Do not build esdoc on Circle, only on ESDoc hosting

zachmullen:
It uses ES6 features that are too new for the version of
node we use for testing on Circle.


zachmullen:
Gonna merge this to see if it fixes master

PR #1831, opened 2017-03-06
Teach UploadWidget to provide newly created file IDs

zackgalbreath:
The g:uploadFinished event passes an array of files to its callbacks.
These files now contain their newly created IDs.


zachmullen:
@msmolens make sure this accommodates your use case. If so, go ahead and merge away.

PR #1832, opened 2017-03-06
Make logs more configurable.

manthey:
Allow setting the size of log files, the number of logs kept, and the log level in the config file.  Also provide an option to suppress all output to stdout.

Make it easier to log cherrypy accesses to the info log.  You can, in the [global] section, add
  log.screen: None
to disable stdout access logging, and
  log.access_file: (path to info log)
to log accesses to the info log (however this doesn't honor rotation properly).

By adding a log_access parameter in the [logging] configure section, this can be controlled more easily.  This can take a list, tuple, set, or single string from the list ('screen', 'info').  An empty list will not log accesses at all (pedantically cherrypy logs to the null handler in any case).  When logging to the info log, our usual format is not applied, as this adds the time stamp a second time (though with milliseconds) and the INFO level tag, which seems unneeded for accesses.

This has the virtue that you don't need to repeat the log file name or path (whereas to use the [global] log.access_file, you do), and that logs are rotated as per our rules.

When reading logs via the endpoint, handle recent rotations.

Add an endpoint for setting log levels.

Convert some bare print statements to logprint statements.  Except for installation we should make it a habit of NEVER leaving print statements in the code -- logging should always preferred.  That way, information can be accessed from the system log endpoint.


manthey:
My motives for this branch are several:
- the logs should show everything (it shouldn't be necessary to view stdout and stderr)
- the log size and rotation should be configurable
- the log level should be adjustable, that way we can start adding debug logs which are normally not shown

girder/__init__.py
```diff
@@ -86,11 +96,20 @@ def _setupLogger():
...
     Sets up the Girder logger.
     """
+    global Quiet
+
     logger = logging.getLogger('girder')
-    logger.setLevel(logging.DEBUG)
+    cfg = config.getConfig()
+    logCfg = cfg.get('logging', {})
+    import sys  # ##DWM::
+    sys.stderr.write('logging %r\n' % logCfg)  # ##DWM::
```
zachmullen: Did you mean to leave this here?

girder/__init__.py
```diff
@@ -35,6 +35,9 @@
 __license__ = 'Apache 2.0'
 
 
+Quiet = False
```
zachmullen: Is this meant to be an internal symbol? Why did you choose to capitalize it?

girder/__init__.py
manthey: No -- debug residue.

girder/__init__.py
manthey: It is a symbol within the module that is used to stop logprint from printing.  It could be manipulated other code, but is more likely just set by the configuration.  We could read it from the config whenever logprint is called, but that seems expensive.

I capitalized it because apparently my ingrained coding habits are module-globals are capitalized.  I'd be happy to change it.

zachmullen: If it's meant to cache a config value, let's make it internal for now: `_quiet`. We can easily open a symbol up for external use later if needed.

PR #1838, opened 2017-03-08
Add file-like interface for Girder files

zachmullen:
Prior to this change, downloading a file involved using the generator interface. This exposes a read-only file-like interface for Girder files, e.g.

```
with self.model('file').open(file) as f:
    while True:
        buf = f.read(BUF_LEN)
        if not buf:
            break
        ...
```

A naive (but still relatively good) implementation of the file-like interface was added to the abstract assetstore adapter, so all assetstore types will get this for free. This PR switches the Girder SFTP service to use this new API, which caused a speedup of more than 2x in non bandwidth-limited environments (on my local machine from a filesystem assetstore, I went from about 30MB/sec to 80MB/sec transfer speeds).

I plan to create an implementation special to the filesystem assetstore that will likely be even faster by just returning a native file handle.

### DIFF
--- a/girder/utility/abstract_assetstore_adapter.py
+++ b/girder/utility/abstract_assetstore_adapter.py
@@ -46,6 +46,8 @@ class FileHandle(object):
     def __init__(self, file, adapter):
         self._file = file
         self._adapter = adapter
+        self._pos = None
+
         self.seek(0)
 
     def __enter__(self):
@@ -78,7 +80,7 @@ class FileHandle(object):
                 self._prev = []
                 length += chunkLen
 
-                if length + chunkLen == size:
+                if length == size:
                     break
             else:
                 chunkLen = min(size - length, chunkLen)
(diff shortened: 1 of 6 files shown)


zachmullen:
@manthey this change does not contain any FUSE support, but does make such support trivial, as you can see in Girder's SFTP code.

manthey:
Very nice.

girder/utility/abstract_assetstore_adapter.py
```diff
@@ -25,6 +26,83 @@
...
+                length += chunkLen
+                break
+
+        self._pos += length
+        return data.getvalue()
+
+    def tell(self):
+        return self._pos
+
+    def seek(self, offset):
```
manthey: We probably want seek to handle a whence option and to be a no op if the _pos is the same as the offset.  The standard python file-object seek has no return value.

zachmullen: Good point

girder/utility/abstract_assetstore_adapter.py
```diff
@@ -25,6 +26,87 @@
...
+
+            if chunkLen == 0:
+                break
+
+            if length + chunkLen <= size:
+                data.write(chunk)
+                self._prev = []
+                length += chunkLen
+
+                if length + chunkLen == size:
```
manthey: Since you just added to length, this should just be `if length == size:`

girder/utility/abstract_assetstore_adapter.py
```diff
@@ -25,6 +26,87 @@
...
+
+    def seek(self, offset, whence=os.SEEK_SET):
+        if whence == os.SEEK_SET:
+            self._pos = offset
+        elif whence == os.SEEK_CUR:
+            self._pos += offset
+        elif whence == os.SEEK_END:
+            self._pos = self._file['size'] - offset
+        self._prev = []
+        self._stream = self._adapter.downloadFile(self._file, offset=self._pos, headers=False)()
```
manthey: Because creating a new stream could have significant overhead, for seeks that don't change the file position, I think it would be better to have a check that if `self._pos` doesn't change then we don't recreate the `self._stream` value.

girder/utility/abstract_assetstore_adapter.py
zachmullen: Good catch

zachmullen:
Are further changes needed to this branch, or is it ready to go?

PR #1839, opened 2017-03-08
Branding in README

zachmullen:


README.rst
```diff
@@ -13,6 +19,10 @@ We'd love for you to `contribute to Girder <CONTRIBUTING.md>`_.
 
 .. |logo| image:: clients/web/static/img/Girder_Favicon.png
 
+.. |kitware-logo| image:: https://www.kitware.com/img/small_logo_over.png
+    :target: https://kitware.com
+    :alt: Kitware Logo
+
```
jcfr: I would suggest to include a different logo/diagram, something that illustrates the role of girder within the resonant ecosystem.

May be could create a Kitware badge, something like

PR #1840, opened 2017-03-08
Fix downloading GridFS files with offsets.

manthey:
If the offset within a GridFS file download is within the second to last chunk, a partial download may result because the last chunk would never be accessed.


manthey:
We should add a test for this.  We don't currently have any testing which exercises the gridFS download file with an offset.

PR #1844, opened 2017-03-09
Combining Rebuild and Restart button on Admin Plugin page

matthewma7:
Earlier, I sent out an email proposing to combine the Rebuild and Restart button on the admin page based the following assumptions:
- The buttons are designed mostly for non-developer and non-development use. 
- If there is a newly enabled plugin that has client side code, rebuilding web client is needed before restarting the server (both buttons are needed and the order matters). 
- A not-so-familiar-with-girder user would be too afraid to press any button or would press both buttons to ensure everything would work.
- It's relatively quick to restart the server

A few people have given feedbacks and it sounds like this is not a bad idea, so I created this PR for it.
I'd create a PR to update documentation after this PR.

--------

Also, as Promise become first class member of JavaScript (Native promise support and some native APIs has been revised to Promise style from callback style), I used some Promise in this PR, it helped me with creating better execution flow and also unit testing (Monkey patching the web client rebuild API to return a resolved promise instead of actually rebuild web client). PhantomJS doesn't support Promise natively, so I had to add a minimal polyfill to the unit test environment. Thank you @zachmullen for your help with the questions I have on this topic.

----------------------
The is the how the new button looks. Existing style and functionality are kept.


jbeezley:
I think using promises is fine, but the `restRequest` method already returns a promise-compatible object.  I'm not sure it's necessary to wrap them here.

mgrauer:
@jjomier 

You mentioned having some confusion about these buttons, which helped to spawn this work.

matthewma7:
@jbeezley I was thinking the restRequest returns something Promise-like, but it doesn't have the .catch() and relies on .fail() and .done(). So, I wrapped it into a standard Promise, so the potential consumer of the restartServer() doesn't need to know what actually happened.

jbeezley:
@matthewma7  You're right, the jquery deferred object (< version 3) is not promise compliant, but it does have the `then` method that can be used to catch rejections.  If we ever get around to upgrading to jquery 3, that issue will go away.  I'm inclined to err toward API consistency here, either `restRequest` should return an actual promise or this code should return a jquery deferred.

On a related note, I think we are targeting some browsers which don't have full ES6 support, which is the reason why we use Babel.  If we do add promises to the interface, the polyfill should go in library bundle rather than just included in the tests.

clients/web/src/server.js
```diff
@@ -10,22 +10,28 @@ import { restRequest } from 'girder/rest';
...
+                }).done(_.bind(function (resp) {
+                    if (resp.serverStartDate !== restartServer._lastStartDate) {
+                        resolve();
+                        restartServer._reloadWindow();
+                    } else {
+                        window.setTimeout(wait, 1000);
+                    }
+                })).error(_.bind(function () {
+                    window.setTimeout(wait, 1000);
+                }));
```
jbeezley: I realize that you didn't write this code, but the `_.bind` here and a few lines above don't actually bind a context.  It would be better to just change these to arrow functions.

matthewma7:
@jbeezley Correct me if I misinterpreted what you meant. Do you mean restRequest() should remain the same? I didn't change restRequest(), I made restartServer() return a Promise so the consumer of restartSever() don't need to know if it is making a bunch of requests or just a mock function that returns a resolved Promise. 
About the polyfill, @zachmullen told me it is ok to not include a polyfill in the web client, but I am totally ok to include a minimal polyfill or even [bluebirdjs](http://bluebirdjs.com), which provides many utility functions and great performance (Even faster than native Promise implementation on certain [problematic browsers](https://www.danyow.net/edge-promise-perf/))

jbeezley:
Sorry for not being clear.  I'm trying to say that you *could* accomplish the code in this PR using `$.Deferred` rather than promises.  Most other async methods in girder return `$.Deferred` objects by virtue of the fact that they tend to return the result of `restRequest`.  My opinion is that having this one async function return a promise instead, even though it may be better or easier to use, breaks the consistency of the library's API.

That said I don't think the issue is extremely important, so if you and @zachmullen are okay with it as is, I won't get in the way.

matthewma7:
@jbeezley Thank for your explanation. When I write this code, I had some difficulty return an async result for the waitForServer() with deferred (Probably because I didn't use it beyond the ajax with .done() .fail() .complete()). Since Zach once told me we want to keep close to the standard such as the arrow function when writing new stuff and changing existing code and he is ok with the standard Promise, so I implemented the functions with Promise.

zachmullen:
In case it's hard to find due to the crazy CI turnover:

```
/home/ubuntu/girder/clients/web/src/server.js
  1:8  error  '_' is defined but never used  no-unused-vars
```

mgrauer:
I know this is after the fact, but will `Promise` work on IE?  A quick search shows it may not without a polyfill.

jbeezley:
No, it won't.  I thought we had come to the conclusion we weren't going to support IE.  It would be pretty easy to include a promise polyfill if necessary.

zachmullen:
My vote would be to just use jquery promises as we have been, and remove the polyfill.

matthewma7:
I am personally thinking, moving forward, if we are going to gradually add new features to girder, we should start to use standard Promise because it is the standard as native JavaScript APIs and other libraries use Promise (I see Candela is using Promise and converting Girder's deferred into Promise). It helps provide unified async interfaces. (In this PR here, I created a mock function returns a resolved Promised without caring the original function uses jquery deferred or q.defferd). Also, it will help us transit into future features such as [async await](https://medium.com/@bluepnume/learn-about-promises-before-you-start-using-async-await-eb148164a9c8#.hnjfv97k2),
Overall, I think the benefit of using Promise greatly outweighs the trouble of having a 3KB Promise polyfill or even an es6 shim (especially as we are bundling).

zachmullen:
Sounds good, we can leave this and just look to remove the phantomjs polyfill in the future whenever they add support for it.

mgrauer:
@jbeezley 

I thought we had not come to any firm conclusion wrt supporting IE, I might have misheard though. 

Either way making our support more explicit is a good goal.

zachmullen:
We don't officially support IE (and never have), though if given the choice of knowingly breaking support for it or not, we'll choose not to unless there's a good reason. You seem convinced of the value of doing so in this case, and I'm sure support for the Promise will come along to IE eventually, so it's fine.

PR #1845, opened 2017-03-09
Updates documentation for mainly development

matthewma7:
I am new to girder development and I did have some questions when I started to develop Girder, so I was asked to update the documentation from a new developer's perspective. 
Also, there are changes to `girder-install web`, other code changes, and clarification requests. 
So, I created this PR to address all above.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -10,7 +10,7 @@ the tools used to develop Girder.
 Configuring Your Development Environment
 ----------------------------------------
 
-In order to develop Girder, you should first refer to :doc:`prerequisites <prerequisites>`, `virtual environment <installation.html#creating-a-virtual-environment>`__, `install from Git <installation.html#install-from-git-checkout>`__, and `run <installation.html#run>`__ sections to setup a basic local development environment.
+In order to develop Girder, you should first refer to :doc:`prerequisites <prerequisites>`, `virtual environment <installation.html#creating-a-virtual-environment>`__, `install from Git <installation.html#install-from-git-repository>`__, and `run <installation.html#run>`__ sections to setup a basic local development environment.
 
 Next, you should install the Python development dependencies with pip, to
 provide helpful development tools and to allow the test suite to run: ::
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -10,29 +10,21 @@ the tools used to develop Girder.
 Configuring Your Development Environment
 ----------------------------------------
 
-In order to develop Girder, you should first refer to the :doc:`prerequisites`
-and :doc:`installation` sections to setup a basic local environment.
+In order to develop Girder, you should first refer to :doc:`prerequisites <prerequisites>`, `virtual environment <installation.html#creating-a-virtual-environment>`__, `install from Git <installation.html#install-from-git-checkout>`__, and `run <installation.html#run>`__ sections to setup a basic local development environment.
 
-Next, you should install the Python development dependencies, to
+Next, you should install the Python development dependencies with pip, to
 provide helpful development tools and to allow the test suite to run: ::
 
     pip install -r requirements-dev.txt
 
-During development, once Girder is started via ``python -m girder``, the server
-will reload itself whenever a Python file is modified.
+Install front-end web client development dependencies. This will install npm modules eslint and pug-lint, which are needed to run unit tests. This will also build the web client code: ::
 
-If you are doing front-end development, it's much faster to use a *watch* process to perform
-automatic fast rebuilds of your code whenever you make changes to source files. If you are making
-changes to Girder's core web client, run the following watch command: ::
+    girder-install web --dev
 
-    girder-install web --watch
+For more options for building web client, run: ::
 
-If you are developing a plugin and want to rebuild when you change its source, run: ::
+    girder-install web --help
 
-    girder-install web --watch-plugin my_plugin_name
-
-When you want to end the watch process, press Ctrl+C (or however you would normally terminate a
-process in your terminal).
 
 Vagrant
 ^^^^^^^
@@ -48,6 +40,28 @@ dependencies.
 .. seealso::
 
    For more information on provisioning Girder, see :doc:`provisioning`.
+
+
+During Development
+------------------
+
+Once Girder is started via ``python -m girder``, the server
+will reload itself whenever a Python file is modified.
+
+If you are doing front-end development, it's much faster to use a *watch* process to perform
+automatic fast rebuilds of your code whenever you make changes to source files. 
+
+If you are making changes to Girder's core web client, run the following watch command: ::
+
+    girder-install web --watch
+
+If you are developing a web client of a plugin, run: ::
+
+    girder-install web --watch-plugin your_plugin_name
+
+With ``watch`` option, *sourcemap* will be generated, which helps debugging front-end code in browser.
+When you want to end the watch process, press Ctrl+C (or however you would normally terminate a
+process in your terminal).
 
 
 Utilities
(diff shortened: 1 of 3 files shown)


docs/development.rst
```diff
@@ -10,29 +10,21 @@ the tools used to develop Girder.
...
-If you are doing front-end development, it's much faster to use a *watch* process to perform
-automatic fast rebuilds of your code whenever you make changes to source files. If you are making
-changes to Girder's core web client, run the following watch command: ::
+Install front-end web client development dependencies. This will install npm modules eslint and pug-lint, which are needed to run unit tests. This will also build the web client code: ::
 
-    girder-install web --watch
+    girder-install web --dev
 
-If you are developing a plugin and want to rebuild when you change its source, run: ::
+For more options for building web client, run: ::
```
zachmullen: the web client*

docs/development.rst
```diff
@@ -50,6 +42,28 @@ dependencies.
    For more information on provisioning Girder, see :doc:`provisioning`.
 
 
+During Development
+------------------
+
+Once Girder is started via ``python -m girder``, the server
```
zachmullen: Let's recommend the `girder-server` command rather than `python -m girder`.

docs/development.rst
```diff
@@ -50,6 +42,28 @@ dependencies.
...
+will reload itself whenever a Python file is modified.
```
zachmullen: Let's add the following:

> This auto-reload behavior only happens when the server mode is set to `development`, not `production` mode.

docs/development.rst
```diff
@@ -50,6 +42,28 @@ dependencies.
...
+
+If you are making changes to Girder's core web client, run the following watch command: ::
+
+    girder-install web --watch
+
+If you are developing a web client of a plugin, run: ::
+
+    girder-install web --watch-plugin your_plugin_name
+
+With ``watch`` option, *sourcemap* will be generated, which helps debugging front-end code in browser.
```
zachmullen: With the watch modes, *sourcemaps* will be generated

docs/development.rst
```diff
@@ -180,14 +198,17 @@ Client Side Testing
 -------------------
 
 Using the same setup as above for the Server Side Tests, your environment will be set up
-to run client side tests. Running ::
+to run client side tests. Running client side is the same as running server side test. Run ::
```
zachmullen: > ~~Running client side is the same as running server side test. Run~~

> The client side tests and server side tests are both harnessed with CTest, so use the following commands to run both:

docs/development.rst
```diff
@@ -180,14 +198,17 @@ Client Side Testing
...
 
     cd girder-build
     ctest
 
 will run all of the tests, which include the client side tests.  Our client tests use the
 Jasmine JS testing framework.
 
+If you encounter errors regarding ESLINT or PUG_LINT, there is a chance you missed certain steps for setting up development dependencies.
+You could use ``ccmake`` to change ``CMake`` configuration. Or, it might be easier to recreate the environment from the beginning.
```
zachmullen: The solution to such errors is likely to run `girder-install web --dev` followed by running `cmake` on the build directory, so we should prescribe that here.

docs/development.rst
```diff
@@ -275,7 +296,7 @@ Code Review
 -----------
 
 Contributions to Girder are done via pull requests with a core developer
-accepting a PR by saying it "Looks good to me" or LGTM. At this point, the
+approving the PR with GitHub review system. At this point, the
```
zachmullen: I think we should remove the whole section on Code Review since it is redundant with our CONTRIBUTING.md file.

docs/installation.rst
```diff
@@ -39,8 +39,14 @@ environments should consider using other packages that help manage them such as
 `pyenv-virtualenvwrapper <https://github.com/yyuu/pyenv-virtualenvwrapper>`_.
 
 
+Sources
+-------
+
+Girder can be installed from two sources, **pip** or **git repository**.
```
zachmullen: Both of these actually should be installed using pip, the real distinction is between **pypi** and **git**.

docs/installation.rst
```diff
@@ -138,17 +144,27 @@ Then open http://localhost:8080/ in your web browser, and you should see the app
...
+++++++++++++++
+
 The next recommended action is to enable any plugins you want to run on your server.
 Click the ``Admin console`` navigation link, then click ``Plugins``. Here, you
 can turn plugins on or off. Whenever you change the set of plugins that are
-enabled, you must restart the `CherryPy <http://www.cherrypy.org>`_ server for
-the change to take effect. For information about specific plugins, see the
-:ref:`Plugins <plugins>` section.
+enabled, you need to press the **Rebuild and restart** button at the top of the 
+Plugins page to rebuild web client and restart the server to apply the change.
```
zachmullen: the web client*

docs/plugin-development.rst
```diff
@@ -57,7 +57,14 @@ The contents of plugin.json for our example will be:
 This information will appear in the web client administration console, and
 administrators will be able to enable and disable it there. Whenever plugins
 are enabled or disabled, a server restart is required in order for the
-change to take effect.
+change to take effect. 
+
+If you are developing a plugin for girder, sometimes using the Rebuild and restart button 
+on Plugins page may be undesirable as it will rebuild core and all enabled plugins
```
zachmullen: the Plugins page*

docs/plugin-development.rst
```diff
@@ -57,7 +57,14 @@ The contents of plugin.json for our example will be:
 This information will appear in the web client administration console, and
 administrators will be able to enable and disable it there. Whenever plugins
 are enabled or disabled, a server restart is required in order for the
-change to take effect.
+change to take effect. 
+
+If you are developing a plugin for girder, sometimes using the Rebuild and restart button 
+on Plugins page may be undesirable as it will rebuild core and all enabled plugins 
+in production mode, which will take some time and doesn't provide sourcemap.
```
zachmullen: sourcemaps*

docs/plugin-development.rst
```diff
@@ -57,7 +57,14 @@ The contents of plugin.json for our example will be:
...
+Rebuild specific plugin restart the server may be a better choice. See `During Development <development.html#during-development>`__ for detail.
```
zachmullen: "Rebuilding specific plugins and restarting the server manually may be a better choice."

zachmullen: details*

docs/development.rst
matthewma7: Should we write how to set the server mode? I personally don't know yet.

docs/development.rst
matthewma7: I think I tried that you suggested. Without nuke the configuration file, the configuration won't be updated. Is there an option of `cmake` to forcefully update the configuration?

docs/development.rst
matthewma7: I remember you suggested that. But I saw the Code Review section has few bullet points specifically on what to review. The CONTRIBUTING.md looks like focuses more on the process to make a contribution. Look like these two has different emphasizes.

docs/installation.rst
matthewma7: En, I personally don't know how to install from git with pip, I always have been cloning with git. 
Would you share some details on this? or make the change directly if you prefer.

docs/plugin-development.rst
matthewma7: @zachmullen I think I wasn't sure `watch` would be considered as automatic or manual when I wrote this. Do you see what I meant? Do you other words in mind? `specifically`?

docs/installation.rst
zachmullen: I'm happy to make the change, but the difference is:

    pip install girder   # this downloads from pypi

vs.

    git clone git://github.com/girder/girder.git
    cd girder && pip install -e .

Both install modes use pip, but one installs from the local directory and one from the pypi package.

zachmullen:
This is ready to go, as long as you approve of my latest commit @matthewma7

zachmullen:
Good catch. LGTM

PR #1846, opened 2017-03-10
Doc reference python client api and misc warn fixes

jcfr:


docs/deploy.rst
```diff
@@ -96,13 +96,13 @@ example, we have the following:
 .. code-block:: ini
 
     [global]
-    server.socket_host: "0.0.0.0"
-    server.socket_port: 9000
-    tools.proxy.on: True
+    server.socket_host= "0.0.0.0"
```
danlamanna: Nitpick, I would add a space on both sides of the `=` since that matches what is currently in our cfg file.

jcfr: Thanks for reviewing. Last commit should take care of this.

PR #1850, opened 2017-03-10
Config file support for girder_client

zachmullen:
Reopened from #1464 against latest master


clients/python/girder_client/cli.py
```diff
@@ -39,14 +40,27 @@ def __init__(self, username, password, host=None, port=None, apiRoot=None,
...
+        elif config:
+            configApiKey = config.get('girder_client', 'apiKey')
+            if configApiKey:
+                self.authenticate(apiKey=configApiKey)
+            elif config.get('girder_client', 'username'):
+                self.authenticate(config.get('girder_client', 'username'),
+                                  config.get('girder_client', 'password'),
+                                  interactive=False)
+        else:
+            print('Unable to authenticate. Insufficient options')
```
zachmullen: Is this an error condition? Should we raise an exception?

Xarthisius: No, actually the `else` statement should be removed. You can still download public folder if you are not authenticated user.

zachmullen:
We are going to close this for now -- after some initial investigation, it looks like there's an easy way to do this using the click library, which we now are using in the girder client. We'll attempt to re-implement this using the appropriate click features.

PR #1852, opened 2017-03-11
Upgrade remarkable version

zachmullen:
The version of remarkable we were using contained an XSS vulnerability documented here: https://nodesecurity.io/advisories/319

I've verified that this was broken before and that this upgrade fixes the issue.


manthey:
I see there is a version 1.7.1.  As long as we are upgrading the version, should we go to the latest?

zachmullen:
Done

PR #1855, opened 2017-03-14
Move some path functions from the resource api to the path utilities module.

manthey:
Add a force option to lookUpPath.

Note that our resource paths are not guaranteed to be unique: an item can have multiple files with the same name.  Perhaps we should also allow resource paths that use resource IDs.

### DIFF
--- a/girder/utility/path.py
+++ b/girder/utility/path.py
@@ -20,7 +20,8 @@
 """This module contains utility methods for parsing girder path strings."""
 
 import re
-from girder.models.model_base import AccessException, ValidationException
+from ..constants import AccessType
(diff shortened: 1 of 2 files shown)


girder/utility/path.py
```diff
@@ -195,3 +200,50 @@ def lookUpPath(path, user=None, test=False, filter=True):
...
+
+
+def getResourcePath(type, doc, user=None, force=False):
+    """
+    Get the path for a resource.
+
+    :param type: the resource model type.
+    :param doc: the resource document.
+    :param user: user with correct privileges to access path
+    :param force: if True, don't validate the access.
```
zachmullen: Let's add `:type:` specifiers to these docstrings

PR #1856, opened 2017-03-14
Fixes #1854. Pass unbound arguments via params in route handlers

### LINKED ISSUE #1854: Error create filesystem assetstore with hdfs plugin enable
To recreate, enable hdfs assetstore plugin and then type to create a filesystem assetstore, the following error will occur:

```python
{
  "message": "TypeError: TypeError(\"createAssetstore() got an unexpected keyword argument 'webHdfsPort'\",)",
  "trace": [
    [
      "girder/api/rest.py",
      546,
      "endpointDecorator",
      "val = fun(self, args, kwargs)"
    ],
    [
      "girder/api/rest.py",
      1103,
      "POST",
      "return self.handleRoute(method, path, params)"
    ],
    [
      "girder/api/rest.py",
      870,
      "handleRoute",
      "val = handler(**kwargs)"
    ],
    [
      "girder/api/access.py",
      39,
      "wrapped",
      "return args[0](*iargs, **ikwargs)"
    ],
    [
      "girder/api/describe.py",
      521,
      "wrapped",
      "return fun(*args, **kwargs)"
    ]
  ],
  "type": "internal"
}```

This is result of changes made to support autoDescribeRoute, in particular [here](https://github.com/girder/girder/blame/master/girder/api/describe.py#L568). We are now passing all describe parameters to the wrapped function regardless of if they were provided.

zachmullen:


cjh1:
@zachmullen I restarted circle, seems like it had been cancelled?

zachmullen:
Yeah, there's a test that's failing now as a result of this change that I need to fix. Once I do that, I'll merge this.

zachmullen:
@cjh1 5075499 fixes the issue, which was a result of `@autoDescribeRoute` not being the first decorator applied. The relevant endpoint was declared as follows:


    @access.public
    @autoDescribeRoute(
        Description('List all available tasks that can be executed.')
    )
    @filtermodel(model='item')
    def listTasks(self, limit, offset, sort, params):

PR #1857, opened 2017-03-14
Use file-like object for reading file in DICOM Viewer

fogleman:
I left in the file-size check.


zachmullen:
We may have large dicom files in the future, I would vote to remove the size limit since we'll either succeed or fail on the early bytes of the file either way.

fogleman:
OK I'll make sure it fails early on a large file.

PR #1858, opened 2017-03-15
Allow null values in metadata, as well as metadata deletion via new endpoint

### LINKED ISSUE #1628: Add endpoint for deleting metadata keys
The way it's done right now is to use the endpoint for setting keys, but set ones to `null` that should be deleted. This is wonky, and it would be more intuitive to just have a separate `DELETE` metadata endpoint to remove a list of keys.

ronichoudhury:
This PR does two related things:
- Adds a `DELETE /item/<id>/metadata` endpoint, whose request body is a JSON list of key names to delete from the metadata
- Adds a boolean `allowNull` parameter to `PUT /item/<id>/metadata`, which allows for `null`-valued metadata fields

To be clear, this does not change the default behavior of the `PUT` endpoint.

I bundled these together to prepare for an eventual breaking change where the `allowNull` behavior is made the only mode of operation for `PUT`, and the `DELETE` endpoint is always used to remove metadata (as discussed in #1628).

Thanks @zachmullen for discussing this with me, including providing the idea for `allowNull`.

Closes #1628

### DIFF
--- a/girder/api/v1/item.py
+++ b/girder/api/v1/item.py
@@ -156,6 +156,8 @@ class Item(Resource):
         .modelParam('id', model='item', level=AccessType.WRITE)
         .jsonParam('metadata', 'A JSON object containing the metadata keys to add',
                    paramType='body')
+        .param('allowNull', 'Whether "null" is allowed as a metadata value.', required=False,
+               dataType='boolean', default=False)
         .errorResponse(('ID was invalid.',
                         'Invalid JSON passed in request body.',
                         'Metadata key name was invalid.'))
@@ -173,7 +175,7 @@ class Item(Resource):
                 raise RestException(
                     'Invalid key %s: keys must not start with the "$" character.' % k)
 
-        return self.model('item').setMetadata(item, metadata)
+        return self.model('item').setMetadata(item, metadata, params['allowNull'])
 
     @autoDescribeRoute(
         Description('Delete metadata fields on an item.')
--- a/girder/api/v1/item.py
+++ b/girder/api/v1/item.py
@@ -163,7 +163,7 @@ class Item(Resource):
                         'Metadata key name was invalid.'))
         .errorResponse('Write access was denied for the item.', 403)
     )
-    def setMetadata(self, item, metadata, params):
+    def setMetadata(self, item, metadata, allowNull, params):
         # Make sure we let user know if we can't accept a metadata key
         for k in metadata:
             if not k:
@@ -175,7 +175,7 @@ class Item(Resource):
                 raise RestException(
                     'Invalid key %s: keys must not start with the "$" character.' % k)
 
-        return self.model('item').setMetadata(item, metadata, params['allowNull'])
+        return self.model('item').setMetadata(item, metadata, allowNull)
 
     @autoDescribeRoute(
         Description('Delete metadata fields on an item.')
--- a/girder/api/v1/item.py
+++ b/girder/api/v1/item.py
@@ -182,7 +182,7 @@ class Item(Resource):
         .responseClass('Item')
         .modelParam('id', model='item', level=AccessType.WRITE)
         .jsonParam('fields', 'A JSON list containing the metadata fields to delete',
-                   paramType='body')
+                   paramType='body', requireArray=True)
         .errorResponse(('ID was invalid.',
                         'Invalid JSON passed in request body.',
                         'Metadata key name was invalid.'))
--- a/girder/api/v1/item.py
+++ b/girder/api/v1/item.py
@@ -24,11 +24,14 @@ from girder.constants import AccessType, TokenScope
 from girder.api import access
 
 
-def validate_field(f):
-    if '.' in f:
-        raise RestException('Invalid key %s: keys must not contain the "." character.' % f)
-    if f[0] == '$':
-        raise RestException('Invalid key %s: keys must not start with the "$" character.' % f)
+def _validateFields(fields):
+    for f in fields:
+        if not f:
+            raise RestException('Key names must not be empty.')
+        if '.' in f:
+            raise RestException('Invalid key %s: keys must not contain the "." character.' % f)
+        if f[0] == '$':
+            raise RestException('Invalid key %s: keys must not start with the "$" character.' % f)
 
 
 class Item(Resource):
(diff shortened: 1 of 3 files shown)


danlamanna:
I'm going to manually run the ansible library tests against this branch just to make sure our use of the settings API still works correctly.

ronichoudhury:
@zachmullen this is ready for review.

girder/api/v1/item.py
```diff
@@ -172,7 +175,32 @@ def setMetadata(self, item, metadata, params):
                 raise RestException(
                     'Invalid key %s: keys must not start with the "$" character.' % k)
 
-        return self.model('item').setMetadata(item, metadata)
+        return self.model('item').setMetadata(item, metadata, params['allowNull'])
```
zachmullen: Let's add `allowNull` as an argument to `setMetadata` so we don't need to look it up in the params dict.

girder/api/v1/item.py
```diff
@@ -172,7 +175,32 @@ def setMetadata(self, item, metadata, params):
...
+        if not fields:
+            raise RestException('Key names must not be empty.')
+
+        for k in fields:
+            if '.' in k:
+                raise RestException(
+                    'Invalid key %s: keys must not contain the "." character.' % k)
+            if k[0] == '$':
+                raise RestException(
+                    'Invalid key %s: keys must not start with the "$" character.' % k)
```
zachmullen: Since this logic exists twice in this file, we should move it to a helper function.

girder/models/item.py
```diff
@@ -304,9 +304,23 @@ def setMetadata(self, item, metadata):
...
+            toDelete = [k for k, v in six.viewitems(metadata) if v is None]
+            for key in toDelete:
+                del item['meta'][key]
+
+        item['updated'] = datetime.datetime.utcnow()
+
+        # Validate and save the item
+        return self.save(item)
+
+    def deleteMetadata(self, item, fields):
```
zachmullen: Let's add a docstring.

girder/models/item.py
```diff
@@ -304,9 +304,23 @@ def setMetadata(self, item, metadata):
...
+        if 'meta' not in item:
+            item['meta'] = {}
+
+        for field in fields:
+            if field in item['meta']:
+                del item['meta'][field]
```
zachmullen:
```
for field in fields:
    item['meta'].pop(field, None)
```

tests/cases/item_test.py
```diff
@@ -352,6 +352,15 @@ def testItemMetadataCrud(self):
         self.assertStatusOk(resp)
         item = resp.json
 
+        # Try to delete metadata from an item that doesn't have any set on it
+        # yet.
+        resp = self.request(path='/item/%s/metadata' % (item['_id']),
+                            method='DELETE', user=self.users[0],
+                            body=json.dumps(['foobar']), type='application/json')
```
zachmullen: Minor nit: this isn't codified in our style rules yet, but we like to avoid visual indents usually.

girder/api/v1/item.py
```diff
@@ -172,7 +175,32 @@ def setMetadata(self, item, metadata, params):
...
 
-        return self.model('item').setMetadata(item, metadata)
+        return self.model('item').setMetadata(item, metadata, params['allowNull'])
+
+    @autoDescribeRoute(
+        Description('Delete metadata fields on an item.')
+        .responseClass('Item')
+        .modelParam('id', model='item', level=AccessType.WRITE)
+        .jsonParam('fields', 'A JSON list containing the metadata fields to delete',
+                   paramType='body')
```
zachmullen: If this `fields` param is expected to be a JSON array, let's add `requireArray=True`. We could validate even further by passing a `schema` dict to indicate it's an Array of Strings and may not be empty.

girder/api/v1/item.py
ronichoudhury: 451a62f7a6ec17f815d7ccbd683d21d32c8950ae

girder/api/v1/item.py
ronichoudhury: 90fb0ce87d278f07995f5238c40ebcd2ca850227

ronichoudhury: I didn't see any other examples of using a schema to enforce a list structure; need some help with that part if we're doing it now.

zachmullen:
Are you planning to change the front-end on this topic so that it uses the new endpoint and passes `allowNull=true`?

girder/api/v1/item.py
ronichoudhury: a167161cba4222c88c2b51401bfaf8a8839b4df5

girder/api/v1/item.py
```diff
@@ -24,6 +24,13 @@
 from girder.api import access
 
 
+def validate_field(f):
```
zachmullen: We could also move the loop into this, and let's call it `_validateField`.

girder/models/item.py
ronichoudhury: 3478b33d0a7a45fc41ab5de73590f20a48793ab0

girder/models/item.py
ronichoudhury: Just out of curiosity, what's wrong with using `del` here?

ronichoudhury: Also, the deletion logic in `setMetadata()` also uses `del`. Should I change that to `pop()` as well?

zachmullen: Nothing wrong with it per se, but the formulation I suggested has the opportunity to perform the operation using only one hashtable lookup per key instead of two. It's also one less line of code and one less conditional in our code, which is always nice.

zachmullen: Within `setMetadata` there is no such conditional, so it's not the same issue.

ronichoudhury: Ohhh I see. I misunderstood what `pop()` is for.

ronichoudhury: 327f65a063f4a96978f371d702e871e3c71ce378

tests/cases/item_test.py
ronichoudhury: I was following the style in the rest of this file. What would you like me to do here?

girder/api/v1/item.py
ronichoudhury: 68618c5c7a1e36ced4ff9eaf0d29e2d34d9fb866

ronichoudhury:
@zachmullen this is ready for review again.

zachmullen:
Should I make an issue to change the front-end to use these new features?

ronichoudhury:
@zachmullen that sounds like a good idea - I think there might be a bit of discussion over just how to change the frontend.

PR #1861, opened 2017-03-17
handle DICOM files without a valid header

fogleman:
@zachmullen This fixes the Emory issue. Some of their DICOM files don't have a valid header but can still be parsed with a `force=True` option. But this also meant extra steps needed to be taken to avoid reading entire files that aren't actually DICOM. This is done with `defer_size=1024`.


plugins/dicom_viewer/server/__init__.py
```diff
@@ -65,12 +65,22 @@ def coerce(x):
         return None
 
 
-def process_file(f):
+def parse_file(f):
```
zachmullen: Any reason to change this name? It's a public symbol, so this is technically a compatibility break.

fogleman: `process_file` is still there but now calls `parse_file`. I did it to have an early return.

zachmullen: Whoops, totally missed that.

fogleman:
@zachmullen Why do I see changes unrelated to my PR in the codecov details?

https://codecov.io/gh/girder/girder/compare/e34549d653c63c7fc4a5ae49b5f9b773974e8122...7d5523fa8e747a18cada234e5cdb3a51457936ba/changes

zachmullen:
That may be diffing against master, the diff vis is here: https://codecov.io/gh/girder/girder/compare/e34549d653c63c7fc4a5ae49b5f9b773974e8122...7d5523fa8e747a18cada234e5cdb3a51457936ba/diff

PR #1865, opened 2017-03-21
Use only 3 parallel slots in Circle

zachmullen:
We now have 6 containers on Circle. Scaling back to three parallel containers per build will allow us to run two builds at once; by moving the static analysis tests to be with the browser tests, this also shouldn't increase response time for an individual build.


zachmullen:
@girder/developers can we merge this? Our static analysis tests won't run into this makes it into master

zachmullen:
BTW, behold:

screen shot 2017-03-21 at 11 47 24 am

PR #1867, opened 2017-03-21
girder_cli: Report progress with 2 digit precision when using python 2.7

jcfr:
This commit explicitly divides by a float number to support python version
that do not implement PEP238 by default.

See https://www.python.org/dev/peps/pep-0238/

Reported-by: "zach.mullen" <zach.mullen@kitware.com>

### DIFF
--- a/clients/python/girder_client/cli.py
+++ b/clients/python/girder_client/cli.py
@@ -51,7 +51,8 @@ class GirderCli(GirderClient):
                 if length == 0:
                     return '%.2f' % length
                 unit = ''
-                units = ['K', 'M', 'G', 'P']
+                # See https://en.wikipedia.org/wiki/Binary_prefix
+                units = ['k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y']
                 while True:
                     if length <= 1024 or len(units) == 0:
                         break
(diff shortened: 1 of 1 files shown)


jcfr:
This Pr follow up #1823 

Cc: @zachmullen

zachmullen:
Thanks for the quick fix, I've already verified that this fixes the issue.

clients/python/girder_client/cli.py
```diff
@@ -53,10 +53,10 @@ def formatSize(length):
                 unit = ''
                 units = ['K', 'M', 'G', 'P']
```
manthey: Am I missing something here, or shouldn't there be a `'T'` between `'G'` and `'P'`?

manthey: And, pedantically, should `K` be lowercase?

jcfr: Ooops. Indeed

jcfr:
@manthey PTAL

clients/python/girder_client/cli.py
```diff
@@ -51,12 +51,13 @@ def formatSize(length):
                 if length == 0:
                     return '%.2f' % length
                 unit = ''
-                units = ['K', 'M', 'G', 'P']
+                # See https://en.wikipedia.org/wiki/Binary_prefix
+                units = ['k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y']
```
manthey: That will do it...  I don't want to wait for the yottabyte file, though.

mgrauer: I think that answers the question of whether Girder can handle big data.

PR #1868, opened 2017-03-22
Refactor AccessControlMixin to fully support "attachedToId" behavior

brianhelba:
The `attachedToId` / `attachedToType` properties have been allowed on `File` models since #857. In #1591, partial support was added to create new uploads for these types of `File`s, but plugin-defined `attachedToType` models did not work.

This PR now adds first-class support for `attachedToId` / `attachedToType` properties and behavior to all `AccessControlMixin` types. This also means that Files using `attachedToId` / `attachedToType` now properly do access checking against their "resource" models, which was previously not enforced.

This contains 3 commits, for which it may be helpful to review independently:
* Add testing to AccessControlMixin
* Allow upload of Files with a plugin-model in attachedToType
* Refactor AccessControlMixin to fully support "attachedToId" behavior

### DIFF
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -70,8 +70,11 @@ class Upload(Model):
         :param size: The total size of
         :param name: The name of the file to create.
         :type name: str
-        :param parentType: The type of the parent: "folder" or "item".
-        :type parentType: str
+        :param parentType: The type of the parent being uploaded into. If attachParent is False,
+                           this should be 'folder' or 'item'. If attachParent is True, this may be
+                           a string with the lowercase name of any model, or a list of the form
+                           ['model_name', 'plugin_name'].
+        :type parentType: str or list
         :param parent: The parent (item or folder) to upload into.
         :type parent: dict
         :param user: The user who is creating the file.
@@ -332,8 +335,11 @@ class Upload(Model):
         :type user: dict
         :param name: The name of the file being uploaded.
         :type name: str
-        :param parentType: The type of the parent being uploaded into.
-        :type parentType: str ('folder' or 'item')
+        :param parentType: The type of the parent being uploaded into. If attachParent is False,
+                           this should be 'folder' or 'item'. If attachParent is True, this may be
+                           a string with the lowercase name of any model, or a list of the form
+                           ['model_name', 'plugin_name'].
+        :type parentType: str or list
         :param parent: The document representing the parent.
         :type parent: dict.
         :param size: Total size in bytes of the whole file.
@@ -377,7 +383,7 @@ class Upload(Model):
             upload['reference'] = reference
 
         if parentType and parent:
-            upload['parentType'] = parentType.lower()
+            upload['parentType'] = parentType
             upload['parentId'] = parent['_id']
         else:
             upload['parentType'] = None
--- a/girder/utility/acl_mixin.py
+++ b/girder/utility/acl_mixin.py
@@ -20,32 +20,103 @@
 import itertools
 import six
 
-from ..models.model_base import Model, AccessException
+from ..models.model_base import Model, AccessException, GirderException
 from ..constants import AccessType
 
 
 class AccessControlMixin(object):
     """
-    This mixin is intended to be used for resources which aren't access
-    controlled by default, but resolve their access controls through other
-    resources. As such, the overridden methods retain the same parameters and
-    only alter functionality related to access control resolution.
+    This mixin is intended to be used for Model types which aren't access controlled by default,
+    but resolve their access controls through another "resource" model object. As such, the
+    overridden methods retain the same parameters and only alter functionality related to access
+    control resolution.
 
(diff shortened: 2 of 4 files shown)


brianhelba:
The behavior of `File`'s `attachedToId` and `attachedToType` keeps breaking. I'm going to add some testing.

girder/models/upload.py
```diff
@@ -355,7 +355,7 @@ def createUpload(self, user, name, parentType, parent, size, mimeType=None,
             upload['reference'] = reference
 
         if parentType and parent:
-            upload['parentType'] = parentType.lower()
```
zachmullen: Why are we no longer lowering?

brianhelba: If this is a plugin-defined model, then `parentType` refers to a list. I look at the various places that call `createUpload`, and they all sanitize the model type (typically by only allowing enumerated values) before passing it in.

zachmullen: In that case, we should update the docstring of this function and mention that it accepts a list/tuple with model and plugin.

zachmullen:
We should rebase this on latest master to grab the circleCI config changes.

brianhelba:
`_isInclusionProjection` is currently only half-implemented, and needs to be completed in order for this to work. ASAP, I'll round out the implementation of `_isInclusionProjection` and add unit tests for it. After that, this PR can be rebased on top and should work.

brianhelba:
To facilitate review, I've scoped this down to a simpler set of commits, which adds unit tests to demonstrate an existing bug (see the PR description), then adds new code to resolve it.

@girder/developers PTAL.

brianhelba:
Actually, this is going to need #2366 for the tests to pass.

brianhelba:
@manthey @girder/developers This should finally be ready for review, as all prereqs have been merged.

girder/utility/acl_mixin.py
```diff
@@ -20,32 +20,103 @@
...
+            ...
+
+        itemInstance = {
+            'folderId': folderInstance['_id'],
+            ...
+        }
+        Item.hasAccess(itemInstance, ...)  # This will check permissions for folderInstance
+    ```
+
+    Second (taking precedence over the first):
```
manthey: Since this way takes precedence, would it be conceptually better for it to appear first?

girder/utility/acl_mixin.py
```diff
@@ -113,51 +186,58 @@ def requireAccess(self, doc, user=None, level=AccessType.READ):
...
                 userid = None
-            raise AccessException("%s access denied for %s %s (user %s)." %
-                                  (perm, self.name, doc.get('_id', 'unknown'),
-                                   userid))
+            raise AccessException('%s access denied for %s %s (user %s).' %
+                                  (perm, self.name, doc.get('_id', 'unknown'), userid))
 
     def requireAccessFlags(self, doc, user=None, flags=None):
         """
         See the documentation of AccessControlledModel.requireAccessFlags, which this wraps.
```
manthey: This no longer wraps `AccessControlledModel.requireAccessFlags`, so the documentation should be updated.

girder/utility/acl_mixin.py
```diff
@@ -113,51 +186,58 @@ def requireAccess(self, doc, user=None, level=AccessType.READ):
...
 
         def hasAccess(_result):
-            resourceId = _result[self.resourceParent]
-
-            # return cached check if it exists
-            if resourceId in resourceAccessCache:
-                return resourceAccessCache[resourceId]
+            # Calling "_resourceModelandId" with "returnType=True" returns the type of the resource
+            # as a hashable string or list, and doesn't call ModelImporter (so it's fast)
+            resourceType, resourceId = self._resourceModelandId(_result, returnType=True)
```
manthey: Isn't `resourceId` guaranteed (by Mongo) to be distinct regardless of the `resourceType`?  As such, isn't `resourceId` sufficient for the cache key?

zachmullen:
Now that I'm using this behavior in another plugin, I'm taking a look at your branch.

I can't find anywhere that these fields were actually indexed. Seems like they should be, no?

girder/utility/acl_mixin.py
brianhelba: I think so, but since this is supposed to also handle plugin models, I see no reason to make that assumption unnecessarily, given that there's no performance benefit to also including the model type in the cache key.

girder/utility/acl_mixin.py
brianhelba: The first method is actually more common (e.g. used for `File`) and cleaner for downstreams to use, so I wanted to introduce it first.

brianhelba:
@manthey @zachmullen I just rebased this on master, and responded to feedback. I'm going to make one more commit to further unify some method documentation, then this should be ready for review again.

brianhelba:
Closing in favor of https://github.com/girder/girder/issues/2513

Issue #1863, opened 2017-03-21
Update frontend to use `allowNull` feature of `PUT /item/<id>/metadata` endpoint


ronichoudhury:
Now that we are able to store actual `null` values in metadata, we should allow it to happen via the frontend.

What tasks would this entail? I believe this list will cover it:
- set `allowNull` to true when the metadata update call happens (i.e., when the "save" button is clicked)
- exclusively use the new `DELETE /item/<id>/metadata` endpoint to remove metadata entries that disappear from the metadata record when the "save" button is clicked

zachmullen:
I've added the "bug" label here, even though this was initially a conscious design decision. I think it's counterintuitive when you add a JSON metadata field and write "null" in the value, and then when you save the field and reload, it just disappears altogether.

ronichoudhury:
Agreed, especially because there's already a correct way to delete a metadata field in the UI.

PR #1869, opened 2017-03-22
Make folder metadata backend consistent with item

zachmullen:
This also moves the key validation logic into the model
base since it could be generally useful.

We should merge #1865 before this gets reviewed.


zachmullen:
Closes #1863

zachmullen:
ping @ronichoudhury @jbeezley

PR #1871, opened 2017-03-22
Document setting_utilities

zachmullen:


girder/utility/setting_utilities.py
```diff
@@ -74,14 +74,20 @@ def getDefaultFunction(key):
...
+    >>> def validateMySetting(doc):
+    >>>     if not doc['value']:
+    >>>         raise ValidationException('This key must not be empty.')
 
-        :param key: The key(s) that this function validates.
-        :type key: str or iterable of str
-        """
+    :param key: The key(s) that this function validates.
+    :type key: str or iterable of str
+    """
```
mgrauer: It isn't clear what `replace` does, can you add that to the docstring?

PR #1874, opened 2017-03-23
Ensure that there is an index on the created field for job.

manthey:
With the admin list-all-jobs option, the list will fail if the job collection is too big.  This fixes that.


manthey:
In many places we query mongo documents using find() without any filter only to immediately filter the results using `filterResultsByPermission` or `filter`.  I think it would be better practice to have a flag for the model `find` method that would filter by the union of `_filterKeys` at least, so that we can transfer less data between Mongo and Python.  This would also push back when we run into sort memory limitations.

PR #1876, opened 2017-03-23
Don't show old events when reopening an event stream

### LINKED ISSUE #1866: Refresh causes re-notification of past events
@manthey mentioned this in the gitter channel last week:
> When I switch away from a Girder tab in a browser and switch back, I get renotified of various events. This is probably due to PR #1687 which detaches the event stream. It would be nice if it remembered what had been shown so as to not reshow messages 

I just recently encountered the same behavior.

jbeezley:
There are two components to this:

1. The addition of a `since` parameter to `/notification/stream`.  This is unix time stamp that filters out notifications before this time.
2. The server time is injected into all the notification messages and stored into `localStorage`.  This value (if present) is passed to the server when opening a new stream.

The second component is a little convoluted, but is designed to get around the issue of synchronization between the client and server clocks.  This fixes the immediate issue of being bombarded by notifications when reconnecting the stream, but further improvements could be made to make notifications more easily dismissable.

Fixes #1866

### DIFF
--- a/clients/web/src/utilities/EventStream.js
+++ b/clients/web/src/utilities/EventStream.js
@@ -47,12 +47,21 @@ prototype._heartbeat = function () {
 prototype.open = function () {
     if (window.EventSource) {
         var stream = this,
-            url = apiRoot + this.settings.streamPath;
+            url = apiRoot + this.settings.streamPath,
+            params = {};
 
         if (this.settings.timeout) {
-            url += '?timeout=' + this.settings.timeout;
+            params.timeout = this.settings.timeout;
         }
 
+        // Set the "since" argument to filter out notifications
+        // that have already been sent to this client.
+        var since = parseInt(localStorage.getItem('sseTimestamp'));
+        if (since >= 0) {
+            params.since = since;
+        }
+
+        url += '?' + $.param(params);
         this._eventSource = new window.EventSource(url);
 
         this._eventSource.onmessage = function (e) {
@@ -64,6 +73,7 @@ prototype.open = function () {
                 stream.trigger('g:error', e);
                 return;
             }
+            localStorage.setItem('sseTimestamp', obj._girderTime);
             stream.trigger('g:event.' + obj.type, obj);
         };
         this._stopHeartbeat = false;
--- a/clients/web/src/utilities/EventStream.js
+++ b/clients/web/src/utilities/EventStream.js
@@ -56,7 +56,7 @@ prototype.open = function () {
 
         // Set the "since" argument to filter out notifications
         // that have already been sent to this client.
-        var since = parseInt(localStorage.getItem('sseTimestamp'));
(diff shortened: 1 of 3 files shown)


clients/web/src/utilities/EventStream.js
```diff
@@ -47,12 +47,21 @@ prototype._heartbeat = function () {
...
+            params = {};
 
         if (this.settings.timeout) {
-            url += '?timeout=' + this.settings.timeout;
+            params.timeout = this.settings.timeout;
         }
 
+        // Set the "since" argument to filter out notifications
+        // that have already been sent to this client.
+        var since = parseInt(localStorage.getItem('sseTimestamp'));
```
zachmullen: Could we refer to this as `window.localStorage`? I know we don't do this for all global symbols, but I think for any that aren't immediately recognizable as browser globals, it helps with readability.

clients/web/src/utilities/EventStream.js
```diff
@@ -64,6 +73,7 @@ prototype.open = function () {
                 stream.trigger('g:error', e);
                 return;
             }
+            window.localStorage.setItem('sseTimestamp', obj._girderTime);
```
msmolens: For robustness, I'd consider catching exceptions from `setItem()`. From https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem:

> setItem() may throw an exception if the storage is full. Particularly, in Mobile Safari (since iOS 5) it always throws when the user enters private mode (Safari sets quota to 0 bytes in private mode, contrary to other browsers, who allow storage in private mode, using separate data containers).
> Hence the developers should make sure to **always catch the possible exceptions from setItem()**.
>

danlamanna: +1 - I also think there are addons/settings for various browsers to explicitly limit/disable local storage on a per site basis.

jbeezley: Sounds like a good idea.  I'll update later today.

zachmullen: My understanding is that localStorage is less likely to be allowed than cookies, would it be safer to store as a cookie? (If it isn't much more effort for us, that is)

manthey: There are pros and cons to both.  Unless it is something we want transmitted on most requests or needs some sort of security, then localStorage would be preferred, I think.  Cookies get transmitted on every request.  localStorage is supported by all modern browsers (and if we keep track of the value in a local variable as well, then when it isn't, such as in private mode, we just fail to remember state when we reload the whole page).

jbeezley: My latest commits catches errors in the localstore API and falls back on an "in memory" timestamp.  Local storage is pretty ubiquitous these days, I don't think it is worth using a cookie, but if someone feels strongly about it, I can make the switch.

PR #1877, opened 2017-03-23
Add the API url for Girder inside of Celery headers

danlamanna:
@zachmullen

### DIFF
--- a/plugins/worker/server/__init__.py
+++ b/plugins/worker/server/__init__.py
@@ -27,7 +27,7 @@ from girder.utility import setting_utilities
 from girder.utility.model_importer import ModelImporter
 
 from .constants import PluginSettings
-from .utils import jobInfoSpec
+from .utils import getWorkerApiUrl, jobInfoSpec
 
 _celeryapp = None
 
@@ -81,7 +81,8 @@ def schedule(event):
         # Send the task to celery
         asyncResult = getCeleryApp().send_task(
             task, job['args'], job['kwargs'], queue=job.get('celeryQueue'), headers={
-                'jobInfoSpec': jobInfoSpec(job, job.get('token', None))
+                'jobInfoSpec': jobInfoSpec(job, job.get('token', None)),
+                'apiUrl': getWorkerApiUrl()
             })
 
         # Set the job status to queued and record the task ID from celery.


plugins/worker/server/utils.py
```diff
@@ -134,9 +134,12 @@ def jobInfoSpec(job, token=None, logPrint=True):
     if isinstance(token, dict):
         token = token['_id']
 
+    apiUrl = getWorkerApiUrl()
+
     return {
         'method': 'PUT',
-        'url': '/'.join((getWorkerApiUrl(), 'job', str(job['_id']))),
+        'url': '/'.join((apiUrl, 'job', str(job['_id']))),
+        'apiUrl': apiUrl,
```
manthey: We call this api_url in the `girderInputSpec` and `girderOutputSpec` functions.

zachmullen:
This integrates with https://github.com/girder/girder_worker/pull/104 if I understand correctly.

This isn't really to do with job management per se, right? Does it belong in jobInfoSpec?

plugins/worker/server/utils.py
danlamanna: I either have to be inconsistent with those, or inconsistent with the keys in this returning dictionary.

danlamanna:
> Does it belong in jobInfoSpec?

The alternative is to add the api url as a separate header [when we send tasks to celery](https://github.com/girder/girder/blob/0e4382ff43c7ccdba0a9be227945632fb99c8d54/plugins/worker/server/__init__.py#L84), maybe that's the right decision here.

The jobInfoSpec as it is exposes the URL a job should call back to to report status updates, I'm simply adding the URL of the spawning Girder instance so the job can call back for other things.

For context, keep in mind that the use case here is using Girder worker *without* input or output specs.

zachmullen:
Separate header seems like the right choice semantically.

danlamanna:
Okay, @manthey @zachmullen, this should be ready to review.

zachmullen:
LGTM on the Girder side, it looks like the worker side hasn't changed to use this new header yet.

danlamanna:
@zachmullen I probably won't get to the worker side today (though it will still continue to function fine). I just wanted to get this in under the wire for the next release.

PR #1878, opened 2017-03-23
Don't render some Pug comments in HTML

msmolens:
As a follow up to #1873, convert some buffered Pug comments to
unbuffered Pug comments so that they don't appear in the rendered HTML.

See: https://pugjs.org/language/comments.html


zachmullen:
Good call, I didn't know about this distinction and assumed it just stripped them all out.

PR #1882, opened 2017-03-23
Python client v2.1.0

zachmullen:


zachmullen:
@jcfr could you please mark this as approved if you are OK with cutting the release of girder-client

jcfr:
I didn't have a chance to test the source and wheel distribution. Do you confirm they work as expected ?

zachmullen:
Yep, all tested

PR #1885, opened 2017-03-24
Add a method for authenticating Girder Client with an existing token

### LINKED ISSUE #803: girder_client: token kwarg for `authenticate()`

danlamanna:
This change seems pretty pedantic looking at it now, but I would rather have a setter that we guarantee backwards compatibility support for in the event that the logic around the `token` property ever changes. It also serves as documentation for the correct way to use Girder client given the user already has a token.

Fixes #803

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -323,7 +323,7 @@ class GirderClient(object):
             resp = self.post('api_key/token', parameters={
                 'key': apiKey
             })
-            self.token = resp['authToken']['token']
+            self.setToken(resp['authToken']['token'])
         else:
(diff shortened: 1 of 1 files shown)


zachmullen:
If we are going to call this `authenticateWithToken`, I think we should just add this as an argument to the existing `authenticate` method. However, if we define authentication as the process of obtaining a token, this new functionality isn't part of that function. So if we have a separate function, I'd vote we just name it `setToken`.

PR #1887, opened 2017-03-25
Add Docker Hub instructions to release documentation

### LINKED ISSUE #786: Add tagged version to dockerhub as part of release process
This will make it possible to do:

```
docker run girder/girder:v1.2.1
```

etc.

Tagging _should_ be as easy as

```
docker build -t girder/girder:<version> .
docker push girder/girder
```

danlamanna:
Fixes #786

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -388,6 +388,14 @@ recommended process for generating a new release is described here.
 
         python setup.py sdist upload
 
+10. Ensure that a built Docker image is tagged on
+    `Girder's Docker Hub page <https://hub.docker.com/r/girder/girder/tags/>`_.
+    Such a build can be added by following the instructions for
+    `adding and running a new build <https://docs.docker.com/docker-hub/builds/#add-and-run-a-new-build>`_.
+
+    The tag name should be the version number only, for example Girder ``v2.2.0`` is tagged on Docker Hub
+    as ``2.2.0``.
+
 Releasing the python client package
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 


docs/development.rst
```diff
@@ -388,6 +388,14 @@ recommended process for generating a new release is described here.
 
         python setup.py sdist upload
 
+10. Ensure that a built Docker image is tagged on
+    `Girder's Docker Hub page <https://hub.docker.com/r/girder/girder/tags/>`_.
+    Such a build can be added by following the instructions for
+    `adding and running a new build <https://docs.docker.com/docker-hub/builds/#add-and-run-a-new-build>`_.
```
zachmullen: Could we just add the commands here to make this easier? It's just

```
docker build -t girder/girder:<version> .
docker push girder/girder
```

zachmullen: Maybe, though I just followed that for the most recent release and it worked (@jbeezley recommended it on #786)

jbeezley: Yeah, I don't actually know how `docker push` works.  That's partially why I didn't try to do the documentation. :)

danlamanna: A couple of thoughts on that:

1) This could result in different dependencies based on the users image cache. This could be solved by adding the `--no-cache=true` flag to the build instructions.
2) Users could potentially leak the contents/configs of their (dirty) working tree into the built/distributed image.

Generally, I would prefer CI machines building the images rather than one of us. That being said, I've definitely exceeded my bikeshedding allowance recently... so I'll fold pretty easily.

manthey: Docker hub automatically builds master.  If you pull the docker image before any other commits are merged to master, you can just tag it with the release information.  This ensures that the built image is as clean as docker hub makes it.

jbeezley: It sounds like clicking the tag button on dockerhub is equivalent to pull/tag/push from the terminal.  It seems like there is more room for error doing it on the terminal though if, for example, you forgot to pull or accidentally built to the tag locally.

zachmullen: Is it possible to have docker hub automatically tag our releases based on github tags?

danlamanna: Apparently it is: https://docs.docker.com/docker-hub/builds/#create-an-automated-build

zachmullen: It supports regexes, so that gets my vote. If the tag matches `^v[0-9]+\.[0-9]+\.[0-9]+$`, that should work I think.

danlamanna:
Automated builds have been setup on Dockerhub using regular expressions, there's no need to include instructions in our release docs.

PR #1889, opened 2017-03-27
Allow extension of the OAuth plugin

zachmullen:


zachmullen:
This is a more general solution to the problem presented in #1826. This allows use of more features from the various OAuth providers besides just their common function of principal authentication. Plugins may now depend on the OAuth plugin and hook into it to request more auth scopes from specific providers, and may listen to events when OAuth users log in to add custom behavior.

zachmullen:
ping @jamiesnape

docs/plugins.rst
```diff
@@ -240,6 +240,35 @@ secret values into the plugin's configuration page, and hit **Save**. Users shou
 then be able to log in with their Google account when they click the log in page
 and select the option to log in with Google.
 
+Extension
```
mgrauer: Can you add to the OAuth Login section above in the docs, an explanation of registration vs login with OAuth, and when Girder Users are created as part of this flow, and what is stored or not stored in the Girder User that distinguishes the OAuth from say a password based login?

docs/plugins.rst
```diff
@@ -240,6 +240,35 @@ secret values into the plugin's configuration page, and hit **Save**. Users shou
...
+to authorize during the OAuth login process.
+
+.. code-block:: python
+
+    from girder.plugins.oauth.providers.google import Google
+    Google.addScopes(['https://www.googleapis.com/auth/plus.circles.read'])
+
+Then, you can hook into the event of a user logging in via OAuth. You can
+hook in either before the Girder user login has occurred, or afterward. In
+our case, we want to do it after the Girder user has been established:
```
mgrauer: can you be more precise about what `established` means here, related to registration vs login?

mgrauer:
@Xarthisius Can you check if this works for your use case?

Xarthisius:
@mgrauer yup! now I can just bind to an event and utilize an oauth token. Thanks!

Issue #1739, opened 2017-01-06
Pressing enter when selecting new user/group in ACL dialog causes page to refresh


zachmullen:
Go to an access control widget, go to the search text input, type a name that matches a user or group, scroll down to one using the arrow keys, and hit enter/return. Rather than adding the user/group to the list (as it does when you click the result), the page is refreshed.

We likely just need to `preventDefault` on an event somewhere.

mgrauer:
I saw the same thing in the Girder header bar site search.

matthewma7:
Fixed with PR https://github.com/girder/girder/pull/1897

PR #1897, opened 2017-03-30
Fixes enter key causing page refresh in Search Control

matthewma7:
This PR is for issue https://github.com/girder/girder/issues/1739.
The issue is caused by the fact there could be `<form>` wrapping the Search Control, Pressing enter key inside a form triggers a form submit an event, which, by default, causes the page to have a form POST refresh. 
As @zachmullen said in the issue, `preventDefault` will fix it.


zachmullen:
Thanks!

Issue #1817, opened 2017-02-28
Make group updates atomic to avoid clobbering ACL


zachmullen:
@TristanWright

zachmullen:
If multiple modification are made concurrently via the REST API, they may clobber one another since we call `save` on the group document rather than using `$push`.

PR #1898, opened 2017-03-30
_setAccess uses update() instead of save() to avoid clobbering

TristanWright:
Addresses #1817

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -760,10 +760,6 @@ class AccessControlledModel(Model):
         if entity not in doc['access']:
             doc['access'][entity] = []
 
-        # First remove any existing permission level for this entity.
-        doc['access'][entity] = [perm for perm in doc['access'][entity]
-                                 if perm['id'] != id]
-
         key = 'access.' + entity
         update = {}
         # Add in the new level for this entity unless we are removing access.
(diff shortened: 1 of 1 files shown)


zachmullen:
@TristanWright thanks for the PR. Two of the server-side tests are failing, meaning this likely contains some logic that breaks assumptions about document state.

girder/models/model_base.py
```diff
@@ -772,10 +770,30 @@ def _setAccess(self, doc, id, entity, level, save, flags=None, user=None, force=
                 'flags': flags
             }
             entry['flags'] = self._validateFlags(doc, user, entity, entry, force)
-            doc['access'][entity].append(entry)
+            # becuase we're iterating this operation is not necesarily atomic
```
zachmullen: because* necessarily*

girder/models/model_base.py
```diff
@@ -772,10 +770,30 @@ def _setAccess(self, doc, id, entity, level, save, flags=None, user=None, force=
...
+                    doc['access'][entity].remove(perm)
 
         if save:
-            doc = self.save(doc)
+            if '_id' not in doc:
+                doc = self.save(doc)
+            else:
+                doc = self.collection.find_one_and_update(
+                    {'_id': ObjectId(doc['_id'])}, update,
+                    return_document=pymongo.ReturnDocument.AFTER)
```
zachmullen: This breaks a subtle part of this function's contract, which is that prior to this change, calling with `save=True` could be used to update other fields of the document besides just `access`.

To fix this, inside this block we could iterate over all the top-level keys of this dictionary *besides* "access", and add them to the `$update` list.

girder/models/model_base.py
```diff
@@ -791,6 +791,15 @@ def _setAccess(self, doc, id, entity, level, save, flags=None, user=None, force=
             if '_id' not in doc:
                 doc = self.save(doc)
             else:
+                updateType = update.keys()[0]
+                # copy all other (potentially updated) fields to the update list
+                if updateType != '$set':
```
zachmullen: For readability, let's invert this conditional since double negation can be confusing.

girder/models/model_base.py
```diff
@@ -791,6 +791,15 @@ def _setAccess(self, doc, id, entity, level, save, flags=None, user=None, force=
...
+                    update['$set'] = {prop: doc[prop] for prop in doc
+                                      if prop != 'access'}
```
zachmullen: Slightly better would be:

    update['$set'] = {k: v for k, v in six.viewitems(doc) if k != 'access'}

(Saves a hash lookup for each key)

girder/models/model_base.py
TristanWright: The combination of `if _id not in doc ... if updateType !='set'` or just the `... != $set`?

girder/models/model_base.py
TristanWright: Neat, that does look better, thanks for reviewing this.

girder/models/model_base.py
zachmullen: Just meant the one you added here, let's do `if updateType == '$set' ... else`

zachmullen:
Great work, thanks for the fix! 👍

zachmullen:
Hm, this broke on python3 but was not reported by our CI because for some reason CircleCI only ran it with a parallel level of 1. Not sure why that is, but I'll put in a quick fix.

PR #1901, opened 2017-03-31
Implements two graphs for jobs

matthewma7:
This PR implements two graphs for jobs. 

Some other notes:
- Job list API endpoint is updated to support server side filtering.
- Another helper endpoint called meta on jobs entity is introduced to query all type and status, type and status is indexed
- When records don't fit one screen, the graphs will grow horizontally.
- All date on the two graph is the updated date time.
- This feature applies to my jobs and the new admin all jobs page.


plugins/jobs/web_client/templates/jobCheckBoxMenu.pug
```diff
@@ -2,10 +2,16 @@ a.g-job-checkbox-menu-link(data-toggle="dropdown", data-target=`#g-job-checkbox-
...
+        input.g-job-filter-checkall(type="checkbox", data-toggle="tooltip", checked=checkAllChecked)
+        | Check all
+    hr
+    ul
+      each checked, name in values
+        li(role="presentation")
+          .checkbox
+            label
+              input.g-job-filter-checkbox(type="checkbox", data-toggle="tooltip", id=name, title=name, checked=checked)
+              | #{name}
```
zachmullen: There's a shorthand syntax for this in pug, this is equivalent to:

    = name

plugins/jobs/web_client/templates/JobList.pug
```diff
@@ -0,0 +1,27 @@
...
+  span(style="margin-right:10px") Page size
+  select.g-page-size
+    each val in pageSizes
+      option(selected=val==pageSize)= val
+
+.filter-container.clearfix
+  .type
+  .status
+
+ul.jobs.nav.nav-tabs
```
zachmullen: For classes that are specific to Girder, let's preface them with `g-` to avoid name collision with third party library selectors.

plugins/jobs/web_client/templates/JobList.pug
```diff
@@ -0,0 +1,27 @@
+.page-size-container
+  span(style="margin-right:10px") Page size
```
zachmullen: Is there a reason not to put this style rule into the style files?

plugins/jobs/web_client/templates/jobCheckBoxMenu.pug
danlamanna: Can pug-lint catch these sorts of things? Should we file an issue about it maybe?

plugins/jobs/web_client/templates/jobCheckBoxMenu.pug
```diff
@@ -2,10 +2,16 @@ a.g-job-checkbox-menu-link(data-toggle="dropdown", data-target=`#g-job-checkbox-
...
-        label
-          input.g-job-filter-checkbox(type="checkbox", data-toggle="tooltip",
-              id=name, title=name, checked=checked)
-          span #{name}
+  .dropdown-menu(role="menu")
+    .checkbox
+      label
+        input.g-job-filter-checkall(type="checkbox", data-toggle="tooltip", checked=checkAllChecked)
+        | Check all
+    hr
```
zachmullen: Minor nit, but my understanding is that the `hr` tag is no longer supposed to be used for presentation, only semantics. If this is just meant to provide a visual divider between sections, my vote would be to use a `border` style rule on one of the surrounding elements or even a `div` with a fixed height and background color.

plugins/jobs/web_client/routes.js
```diff
@@ -6,25 +6,26 @@ import JobDetailsWidget from './views/JobDetailsWidget';
...
 
 router.route('jobs/user/:id', 'jobList', function (id) {
     events.trigger('g:navigateTo', JobListWidget, {
-        filter: {userId: id}
+        filter: { userId: id }
     });
 });
 
-router.route('jobs', 'allJobList', function () {
+router.route('jobs(/:view)', 'allJobList', function (view) {
```
zachmullen: Does this break the old route? We should preserve the old route with some default behavior, in case anyone downstream has links to it.

plugins/jobs/server/job_rest.py
```diff
@@ -42,30 +46,41 @@ def __init__(self):
...
+        .param('statusesArray', 'Filter for type', required=False)
         .pagingParams(defaultSort='created', defaultSortDir=SortDir.DESCENDING)
     )
-    def listAllJobs(self, limit, offset, sort, params):
+    def listAllJobs(self, typesArray, statusesArray, limit, offset, sort, params):
         currentUser = self.getCurrentUser()
-        return list(self.model('job', 'jobs').listAll(
-            offset=offset, limit=limit, sort=sort, currentUser=currentUser))
+        types = json.loads(typesArray) if typesArray is not None else None
+        statuses = json.loads(statusesArray) if statusesArray is not None else None
```
zachmullen: We could use the `jsonParam` instead of `param` for these two new parameters since they are JSON.

plugins/jobs/web_client/templates/jobCheckBoxMenu.pug
zachmullen: Would be nice, it's fine by me to file such an issue.

plugins/jobs/web_client/routes.js
matthewma7: @zachmullen, Thank you for reviewing this PR. 
This route doesn't break the old behavior because the view part is inside a pair of parentheses. Actually, the default landing URL to the page from other pages remained the same.

matthewma7:
@zachmullen @danlamanna, thank you for your reviewing and the comments. I made the changes. This PR is ready for further review.

zachmullen:
It looks like the time series and bar charts are ordered with the oldest job being on the right. My expectation was that it would be the opposite, with the jobs being chronologically ordered with the oldest on the left.

A great feature to add would be to make clicking a bar in the bar chart or a point in the time series chart navigate to the job details page for that particular job. If that's easy, we can do it on this branch, but if not it's fine to open an issue to add in a future topic.

zachmullen:
One other usability request: when clicking the "type" checkboxes in the filter area, the list of types goes away each time I check or uncheck a box. We should make it so the box stays visible until the user clicks outside of it.

matthewma7:
@zachmullen, Thank you for your feedbacks. :)
The drop down menu closing is caused by excessive DOM recreation of the render model. Frankly, I am wondering if we plan to incorporate a new mechanism for views. But for now, I'll see if I can avoid the excessive DOM recreation.

Open detail record on click is a good idea, I'll implement it with page number persevered in the URL (Otherwise when user navigating back, the paging state will be lost).

Our paging and list view is designed as right or down means older. Frankly, I feel it is more consistent to put oldest to the right. And, since people mostly interested in most recent stuff, position the latest consistently on the left of the page is more useful. Otherwise, when the page size is large, the position of the most recent record will be changing and pushed to the far right.

matthewma7:
@zachmullen 
The not closing when checking on the filter drop down involve hacks and makes the code is not clean, so I didn't implement it. 
I realized preserving page number without preserving filter state is not enough, so I open job details in a new tab when clicking on the graph bar and line.

zachmullen:
@matthewma7 a situation like that can usually be remedied by breaking up a larger view up and having it contain smaller sub-views within it, then only the sub-view needs to be re-rendered. Does that make sense?

matthewma7:
@zachmullen I understand. The page is largely dividable. The tricky part is the dropdown. The bootstrap dropdown will close on click by default, a preventdefault could prevent closing, but would also prevent the backbone handler from getting the event, but seems some hacks may be used. Within the dropdown, without re-rendering, detailed jQuery DOM manipulation or further division of view is needed when dropdown state changed. Overall, it's still possible without some smarter view engine. So, taking the side effects into consideration, should we do it?

matthewma7:
Hi @zachmullen, I implement the feature to preserve the opening state of the filter dropdowns.

matthewma7:
This is ready for further review ;)

mgrauer:
Overall this is a very nice improvement, thanks @matthewma7.  

Here are some specific impressions.  I don't want to hold up merging for most of these, they can be completed on later revisions.

Most of the time series I've seen have the newest stuff on the right, see e.g. this Google cloud console graph.  I'm fine either way as this is more of a preference, perhaps the ability to switch between left and right would be useful?

screen shot 2017-04-07 at 1 20 43 pm

I don't like the scrolling off screen as more elements come in, especially with new items on the right side.  I'd rather see new elements always centered, and if we have to scroll, scroll the older items.

- [ ] Either way, left or right for newer items, I'd want to see something like an axis label or other indicator about where newer items are (or how the jobs are ordered left to right).  This is important enough that I'd want to see it before merging.

- [ ] Is the color legend on the time series graph filled in with actual colors?  In the GIF attached to this PR, it has the names but not the colors in the legend.

mgrauer:
`Timing history` sounds like the best choice, to replace the label for the middle tab `Phases`.

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -57,27 +79,72 @@ var JobListWidget = View.extend({
...
+                statusTextToStatusCode[statusText] = status;
+                return statusText;
+            }).reduce((obj, statusText) => {
+                obj[statusText] = true;
+                return obj;
+            }, {});
+            this.statusFilterWidget.setValues(statusFilter);
+        });
+
+        this.phaseFilterWidget = new CheckBoxMenu({
```
mgrauer: ah, there will have to be a lot of changes throughout the code to update from `phase` to `Timing history`.  Sorry about that!

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -90,99 +157,246 @@ var JobListWidget = View.extend({
...
+                }
+            };
+        };
+
+        if (this.currentView === 'phase') {
+            new JobStatusSegmentizer().segmentize(jobs);
+            let vegaData = this._prepareDataForChart(jobs);
+
+            let config = jQuery.extend(true, {}, phaseChartConfig);
+            config.width = Math.min(Math.max(this.$el.width() - 50, jobs.length * 16 + 100), jobs.length * 30 + 400);
```
mgrauer: where do these numbers come from?  @zachmullen is there a better way to do this?

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -90,99 +157,246 @@ var JobListWidget = View.extend({
...
+            let positivePhases = _.clone(this.phasesFilter);
+            this.phaseFilterWidget.setValues(positivePhases);
         }
 
-        tr.addClass('g-highlight');
+        if (this.currentView === 'time') {
+            new JobStatusSegmentizer().segmentize(jobs);
+            let vegaData = this._prepareDataForChart(jobs);
+            let config = jQuery.extend(true, {}, timeChartConfig);
+            config.width = Math.min(Math.max(this.$el.width() - 50, jobs.length * 16 + 100), jobs.length * 30 + 400);
```
mgrauer: here too, the magic numbers

plugins/jobs/web_client/views/JobListWidget.js
zachmullen: Ideally we could do this with static CSS3 (e.g. `calc`)...

plugins/jobs/server/models/job.py
```diff
@@ -467,3 +463,11 @@ def _computeFields(self, kwargs, includeLogDefault=False):
         if fields is None and not kwargs.pop('includeLog', includeLogDefault):
             fields = {'log': False}
         return fields
+
+    def getAllTypesAndStatuses(self, user=None):
```
matthewma7: Hi @mgrauer, Thank you very much for all your review and comments. 
About this comment and the comment above, the whole story was like this. Before I implemented the admin jobs page, the jobs rest endpoint uses a string 'none' to indicate for anonymous jobs. Since I was just started, not wanting to break other's code. I created a dedicated listAll to list all jobs. 
Now, when I come back to this for the graph, as I am more familiar with the code, I found it redundant to have two sets of methods for list and list all. So, I removed the listall() and learned existing user='none' string and user = 'all' for filtering jobs. 
But I agree it is kind of confusing, I'll do some refactoring or adding comments to clear thing up.

plugins/jobs/server/models/job.py
```diff
@@ -63,24 +68,15 @@ def list(self, user=None, limit=0, offset=0, sort=None, currentUser=None):
...
         """
-        userId = user['_id'] if user else None
-        cursor = self.find({'userId': userId}, sort=sort)
-
-        for r in self.filterResultsByPermission(cursor=cursor, user=currentUser,
-                                                level=AccessType.READ,
-                                                limit=limit, offset=offset):
-            yield r
-
-    def listAll(self, limit=0, offset=0, sort=None, currentUser=None):
```
matthewma7: Please read the below comment first for the background story. 
So, since I added this listall() recently myself, should I add listall() back with a deprecated mark?

plugins/jobs/web_client/views/JobListWidget.js
matthewma7: These numbers are some numbers I use for sizing the charts. As I am changing the sizing behavior of the charts, I'll use comment or variable name to clear things up.  
I should comment more in general.

plugins/jobs/web_client/JobStatus.js
```diff
@@ -41,42 +45,55 @@ var JobStatus = {
...
             this._map[info.value] = {
                 text: info.text,
-                icon: info.icon
+                icon: info.icon,
+                color: info.color
             };
         }, this);
+    },
+
+    getAllStatus: function () {
```
danlamanna: This can be called `getAllStatuses` or just `getAll`. Also, I think this can just be `return _.values(this._map)`.

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -90,99 +157,246 @@ var JobListWidget = View.extend({
...
-        this.filterStatusMenuWidget.setElement(this.$('.g-job-status-header')).render();
+        this.$el.html(JobListTemplate(this));
+
+        this.typeFilterWidget.setElement(this.$('.filter-container .type')).render();
+        this.statusFilterWidget.setElement(this.$('.filter-container .status')).render();
+
+        this.$('a[data-toggle="tab"]').on('shown.bs.tab', e => {
+            this.currentView = $(e.target).attr('name');
+            if (this.userId) {
+                router.navigate(`jobs/user/${this.userId}/${this.currentView}`);
```
danlamanna: The fact that we're doing navigation to a property called `currentView`, coupled with the large blocks of code dependent on which "view" we're in seems to indicate these tabs should be separated out into actual backbone views.

plugins/jobs/web_client/views/CheckBoxMenu.js
```diff
@@ -1,27 +1,63 @@
...
 
 var CheckBoxMenu = View.extend({
     events: {
-        'click input': function (e) {
-            var checkBoxStates = {};
-            $.each(this.$('input'), function (i, input) {
-                checkBoxStates[input.id] = input.checked;
+        'click input.g-job-filter-checkbox': function (e) {
+            e.stopPropagation();
+            _.keys(this.values).forEach(key => {
```
matthewma7: Thanks, otherwise, the id property is kind of wasted.

plugins/jobs/web_client/views/JobStatusSegmentizer.js
```diff
@@ -0,0 +1,37 @@
+import _ from 'underscore';
+import JobStatus from '../JobStatus';
+
+export default class JobStatusSegmentizer {
+    segmentize(jobs) {
```
matthewma7: The logic is extracted let's keep it here for now and extract it to other places when the logic is useful elsewhere.

plugins/jobs/web_client/views/CheckBoxMenu.js
```diff
@@ -1,27 +1,63 @@
...
+        },
+        'click .g-job-checkall input': function (e) {
+            e.stopPropagation();
+            var checked = !this.allItemChecked();
+            this.checkAllChecked = checked;
+            _.keys(this.values).forEach(key => {
+                this.values[key] = checked;
+            });
+            this._renderContent();
+            this.trigger('g:triggerCheckBoxMenuChanged', this.values);
```
matthewma7: I just kept the logic of 'g:triggerCheckBoxMenuChanged'. Is this event not scoped to only its subscriber? Extend values sounds like a good idea, will the event be emitted automatically when values change?

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -90,99 +157,246 @@ var JobListWidget = View.extend({
...
+            filter.userId = this.userId;
+        }
+        if (this.typeFilter) {
+            filter.types = JSON.stringify(this.typeFilter);
+        }
+        if (this.statusFilter) {
+            filter.statuses = JSON.stringify(this.statusFilter);
+        }
+        this.collection.params = filter;
+        this.collection.fetch({}, true);
```
matthewma7: When the page first starts, the filter doesn't exist. I found those specific logics make more sense on the view than the model. So, I kept it in the view.

plugins/jobs/web_client/views/JobListWidget.js
matthewma7: There are some interactions between this three views, such as paging, filter etc, so I found it more easy to work with to keep a flat structure.

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -19,18 +28,28 @@ var JobListWidget = View.extend({
...
+        this.userId = (settings.filter && !settings.allJobsMode) ? (settings.filter.userId ? settings.filter.userId : currentUser.id) : null;
+        this.typeFilter = null;
+        this.statusFilter = null;
+        this.phasesFilter = JobStatus.getAllStatus().reduce((obj, status) => {
+            obj[status.text] = true;
+            return obj;
+        }, {});
+
+        this.pageSizes = [25, 50, 100, 250, 500, 1000];
+        this.pageSize = 25;
```
matthewma7: Do you suggestion on how to handle for the job/all situation?

This logic for page limit is simple enough to reuse what's in the collection.
Overall, I think I have a tendency to keep view layer and model layer away from each other. I mean, the pug doesn't need to know if there is one collection or two collections, the pug only need to know there is a page limit, the logic is kept within the View Model.

plugins/jobs/web_client/views/CheckBoxMenu.js
```diff
@@ -1,27 +1,63 @@
...
     },
     setValues: function (values) {
-        this.params.values = values;
-        this.render();
+        this.values = values;
+        this.checkAllChecked = this.allItemChecked();
+        this._renderContent();
+    },
+    allItemChecked: function () {
+        return _.find(_.keys(this.values), key => !this.values[key]) === undefined;
```
matthewma7: I agree. all sounds more logically straightforward with what the function do, the performance benefit of the find() is negligible.  Weird, I didn't find all() on the underscore website.
Totally, the terminology should be unified.

plugins/jobs/web_client/JobStatus.js
matthewma7: I see. getAll() sounds great. values() should totally be used 👍.

plugins/jobs/web_client/views/CheckBoxMenu.js
zachmullen: _.all is an alias for [_.every](http://underscorejs.org/#every)

plugins/jobs/web_client/views/CheckBoxMenu.js
jbeezley: I think triggering an event on the view object is appropriate here.

plugins/jobs/web_client/views/JobStatusSegmentizer.js
jbeezley: :+1: to moving this.  It isn't a view.  Until we find a use for it elsewhere, it could be moved as just a simple function inside `JobListWidget.js`.

plugins/jobs/web_client/views/JobListWidget.js
jbeezley: I agree this should be refactored into multiple views.  The common components can be extracted into a base class or a set of helper functions.

matthewma7: @danlamanna @jbeezley Frankly, the only problem I can see is a slightly long render() method. What you are suggesting to extract specifically and what benefits will that provide?

plugins/jobs/web_client/views/JobStatusSegmentizer.js
matthewma7: @jbeezley, I see. Moving it into JobListWidget to indicate that it's used as a helper. I that case, the timeChartConfig phaseChartConfig file problem should be moved too. Will creating a subdirectory for all three helper file for the JobList help?

plugins/jobs/web_client/views/CheckBoxMenu.js
```diff
@@ -1,27 +1,63 @@
...
+        // When a label wrap a input and the label is clicked, there will be two event being triggered, stop the first one from bubbling up to prevent unexpected behavior
+        'click label': function (e) {
+            e.stopPropagation();
+        },
+        'click .dropdown-menu': e => {
+            e.stopPropagation();
+        },
+        'click .g-job-checkall input': function (e) {
+            e.stopPropagation();
+            var checked = !this.allItemChecked();
```
matthewma7: Thank you. the property truly is redundant for the ViewModel, removed.

plugins/jobs/web_client/views/JobListWidget.js
matthewma7: Is there a similar situation I can take a look?

danlamanna: It's nothing to do with performance and everything to do with maintainability/readability imo. We have a view that has different "types" of views within it, that's not very intuitive to the next person who will work on this code. For example, to change something on the list view the developer now has to wade through the logic of multiple graphs. If this were a view that created child views which housed that logic, that would be more intuitive.

matthewma7:
@danlamanna, @jbeezley I extracted Graph stuff into a subview.

This PR is ready for review. 
But I think I need to make some changes once https://github.com/girder/girder/pull/1916#issuecomment-293314083 is merged.

plugins/jobs/server/models/job.py
mgrauer: Yes, please add `listAll` back in, to maintain backwards compatibility.

mgrauer:
Yikes, what happened to the diff?  Merging in master added over 4000 files to the diff.  I haven't seen that when merging in master before.

Looking at [`getAllTypesAndStatuses`](https://github.com/girder/girder/blob/ec4b70bbdfb05f50bf7d66d577384d9e18735dc6/plugins/jobs/server/models/job.py#L483-L491) in the job model in the latest commit, this looks problematic.  If you allow the default of `None` for `user` it will throw an error when accessing `user['_id']`.

```
    def getAllTypesAndStatuses(self, user=None):
        query = {}
        if user == 'all':
            pass
        else:
            query['userId'] = user['_id']
```

danlamanna:
@matthewma7 It looks like you accidentally committed the directory containing your virtualenv?

You'll have to do some history rewriting here I think. At a minimum do a `git rm --cached env` to delete `env` from git, but keep it in your local working tree.

matthewma7:
The merge commit is fixed, don't know how that happened.
Ready for review.

clients/web/test/testEnv.pug
```diff
@@ -14,7 +14,7 @@ html(lang="en")
     each js in jsFiles
       script(src=js)
     
-    script(src="../../../test/lib/promise.min.js")
```
zachmullen: Do we still need promise.min.js at all? It's still included in this PR

matthewma7: No, I use Core.min.js to have a super polyfill for Phantom.js now because Vega uses something about Symbol. Did I forget to remove it from some place?

zachmullen: My bad, I saw the file in the diff list but it was collapsed so I didn't notice it was being removed instead of added.

plugins/jobs/web_client/templates/JobList.pug
```diff
@@ -0,0 +1,29 @@
...
+
+ul.g-jobs.nav.nav-tabs
+  li(class=(currentView=='list'?'active':''))
+    a(data-toggle="tab", name="list")
+      i.icon-th-list
+      |  List
+  li(class=(currentView=='timing-history'?'active':''))
+    a(data-toggle="tab", name="timing-history")
+      i.icon-chart-bar
+      |  Timing history chart
```
zachmullen: Let's drop the "chart" and just leave it at "Timing history". 👍

mgrauer:
@matthewma7 The job model method `listAll` was included in release 2.2.0, so you'll need to add it back in, to avoid breaking backwards compatibility.

plugins/jobs/server/models/job.py
```diff
@@ -89,6 +89,19 @@ def list(self, user, types=None, statuses=None,
...
 
+    def listAll(self, limit=0, offset=0, sort=None, currentUser=None):
+        """
+        List all jobs.
+
+        :param limit: The page limit.
+        :param offset: The page offset
+        :param sort: The sort field.
+        :param currentUser: User for access filtering.
+        :deprecated: use list method instead
```
mgrauer: [This](https://github.com/girder/girder/blob/e34549d653c63c7fc4a5ae49b5f9b773974e8122/girder/api/v1/file.py#L204) is the only other instance of deprecation I could find in our Python. 
 I don't care which style we use, but I'd like to be consistent.

    Multipart uploads are @deprecated as of v2.2.0.

zachmullen: I think the one we want for sphinx is `.. deprecated ::`

http://www.sphinx-doc.org/en/stable/markup/para.html#directive-deprecated

matthewma7:
image
This is an example to show sphinx deprecated syntax style.

girder/api/v1/file.py
```diff
@@ -201,7 +201,7 @@ def readChunk(self, upload, offset, params):
         well as passing offset and uploadId as query parameters and passing
         the chunk as the body, which is the recommended method.
 
-        Multipart uploads are @deprecated as of v2.2.0.
+        .. deprecated :: v2.2.0
```
zachmullen: Since the output already contains the word "version", let's remove the "v" here

plugins/jobs/server/job_rest.py
```diff
@@ -130,3 +141,19 @@ def updateJob(self, job, log, overwrite, notify, status, progressTotal, progress
     )
     def deleteJob(self, job, params):
         self.model('job', 'jobs').remove(job)
+
+    @access.admin
+    @autoDescribeRoute(
+        Description('Get metadata of all jobs')
```
mgrauer: I didn't pay close enough attention to this before, but this method talks about the metadata of the job, though it only returns the types and statuses.  There is a separate metadata field in the job model, which is not returned here, so I think this is confusing.

Better to call it `allJobsTypesAndStatuses` or something along those lines.

plugins/jobs/server/job_rest.py
```diff
@@ -130,3 +141,19 @@ def updateJob(self, job, log, overwrite, notify, status, progressTotal, progress
...
+        .errorResponse('Admin access was denied for the job.', 403)
+    )
+    def allJobsMeta(self, params):
+        return self.model('job', 'jobs').getAllTypesAndStatuses(user='all')
+
+    @access.user
+    @autoDescribeRoute(
+        Description('Get metadata of jobs of current user')
+    )
+    def jobsMeta(self, params):
```
mgrauer: Here too, `jobsTypesAndStatuses`.  Or maybe there is a word that means 'types and statuses' but is different from 'metadata', good luck if you want to find it.

plugins/jobs/server/models/job.py
```diff
@@ -28,19 +28,23 @@
...
+            ('status', SortDir.ASCENDING)
         )
         self.ensureIndices([(compoundSearchIndex, {}), 'created'])
 
         self.exposeFields(level=AccessType.READ, fields={
             'title', 'type', 'created', 'interval', 'when', 'status',
             'progress', 'log', 'meta', '_id', 'public', 'async', 'updated', 'timestamps'})
 
-        self.exposeFields(level=AccessType.SITE_ADMIN, fields={'args', 'kwargs'})
+        self.exposeFields(level=AccessType.SITE_ADMIN,
```
mgrauer: Why this change?  Is your editor somehow set to only allow 80 chars or to reformat lines with more than 80 chars?  We allow 100 chars in Girder Python.

It isn't a big deal, but I wanted to point it out because whitespace changes can make the diff take longer to read through and can be a distraction while reviewing.

plugins/jobs/server/models/job.py
```diff
@@ -52,19 +56,33 @@ def _validateStatus(self, status):
             raise ValidationException(
                 'Invalid job status %s.' % status, field='status')
 
-    def list(self, user=None, limit=0, offset=0, sort=None, currentUser=None):
+    def list(self, user, types=None, statuses=None,
```
mgrauer: I didn't pay close enough attention here either.  You've changed the default for the `user` param, which breaks backwards compatibility.  Can you reset it to `None`? Is there a reason you needed to use the string value `none` instead of the Python `None`?

It is possible you did this because of review feedback I gave you before, I apologize if I've created any confusion.

plugins/jobs/server/models/job.py
```diff
@@ -79,13 +97,11 @@ def listAll(self, limit=0, offset=0, sort=None, currentUser=None):
         :param offset: The page offset
         :param sort: The sort field.
         :param currentUser: User for access filtering.
+        .. deprecated :: 2.3.0
+           Use :func:`list` instead
```
mgrauer: is there a way to specify that this is the `list` method of the `Job` class, instead of say, Python's `list` function?  

It is probably fine as is, but if it is easy to add in that additional info, please do.

plugins/jobs/server/job_rest.py
matthewma7: Definitely, I didn't aware there is an existing concept of **meta**.

plugins/jobs/server/models/job.py
matthewma7: Got it. I changed my autopep8 settings to 100 chars.

danlamanna: Depending on your IDE, you should be able to just specify the flake8 config which is at `tests/flake8.cfg`.

plugins/jobs/server/models/job.py
matthewma7: Originally, the user parameter can only be a user dict or `None` to filter by user or anonymous job. I added a separate `listAll()` method to list all jobs when I implement the admin jobs page. However, when I create the `jobsmeta`(Now called JobsTypesAndStatuses), I feel it too redundant to have a series of a separate list and listAll on the model, so I combined them and used string `'all'` to indicate the difference. I tried to make it more clear. I thought `'none'` is better than `None` semantically, especially when put aside with  `'all'`. 
So, maybe I keep the `"none"` and add back `None` and make them behave the same?

mgrauer: Why do you think `'none'` is better than `None`?, I want to understand your thinking here.

For me `None` seems fine, because IIUC `user` says which jobs you want based on which users are attached to the job.  `None` means anonymous jobs, `'all'` means all users, and a specific user document gets you jobs tied to that user.  

If you prefer to include `'none'`, I'm ok with that, you have definitely thought about this more than me, as long as the interface goes back to `user=None` as default.

matthewma7: This could be purely my personal way of thinking. I see the `user` parameter as a filter, so `None` could mean filter value is `None` or no user filter (don't filter by user) because None has two level of meaning, in programming language it means something does exist, in our business logic, it means anonymous user.

mgrauer:
I approve on the server side.  I lost track of who was QAing most recently on the client, but this is RFR.

plugins/jobs/server/models/job.py
mgrauer: We talked about this outside of the PR.  It is confusing, the best path forward while still maintaining backwards compatibility seemed to be a slightly awkward implementation sprinkled liberally with explanatory comments.

plugins/jobs/server/models/job.py
matthewma7: @danlamanna Thanks for the tip 👍, but seems the formatter I use doesn't support flake8 setting, but I'll keep the flake8.cfg in mind in future.

matthewma7:
Hi all, could we approve this PR if there are no new changes needed?

plugins/jobs/plugin_tests/jobsSpec.js
```diff
@@ -188,305 +170,138 @@ $(function () {
...
-                    status: i,
-                    type: t,
-                    updated: '2015-01-12T12:00:0' + i,
-                    created: '2015-01-12T12:00:0' + i,
-                    when: '2015-01-12T12:00:0' + i
+                return $($('.g-jobs-list-table>tbody>tr').get(2)).find('td.g-job-status-cell').text() === 'Error';
+            }, 'Third row status change to Error');
+
+            runs(function () {
+                console.log('trigger job created');
```
zachmullen: Remove debug statement

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -1,190 +0,0 @@
-import _ from 'underscore';
```
zachmullen: I must have missed this before, but we need to move this file back to its original name and location so that we don't break backward compatibility. If someone was importing this file into their own bundle, this would break their build.

jbeezley:
There's another issue with backwards compatibility here.  In slicer_cli_web, we [import `JobListWidget`](https://github.com/girder/slicer_cli_web/blob/master/web_client/views/JobsPanel.js#L4), which is now gone.  Would it be a problem to name `JobList.js` back to `JobListWidget.js`?

plugins/jobs/web_client/views/JobListWidget.js
jbeezley: You officially beat me by 2 minutes.

matthewma7:
@jbeezley @zachmullen, I see, changing the filename back is no problem.

jbeezley:
Another potential compatibility issue (depending on how you define it) is that the jobs panel now doesn't work well inside the slicer_cli_web panels.  It's functional, but it doesn't fit well in the space.  Is there a way to make the new tabs optional and just show the list view as before?

jobs_list
test

zachmullen:
A solution to what @jbeezley ran into might be to keep JobListWidget as it was before, and have a higher-level view that contains the three tabs, and have the `jobs/user/:id` and `jobs` routes point to the new view. I.e.

```
girder.views.body.JobsView
|
|__ girder.views.widgets.JobListWidget
|__ girder.views.widgets.JobTimingHistoryWidget
|__ girder.views.widgets.JobTimeSeriesWidget
```

matthewma7:
@zachmullen, there are some other changes such as the filters and paging. I think It's not possible to keep JobListWidget unchanged and wraps it into another view. I think making the graph optional is cleaner.

zachmullen:
Sounds good.

matthewma7:
PR is ready for review.

plugins/jobs/web_client/stylesheets/jobListWidget.styl
```diff
@@ -49,3 +34,46 @@
...
   display inline
+
+a.g-job-checkbox-menu-link
+  display inline-block
+  overflow hidden
+  text-overflow ellipsis
+  white-space nowrap
+  color: #333
+
+.dropdown
```
jbeezley: This should be a more specific selector.  As is, it will modify properties of any dropdown in the page.

plugins/jobs/web_client/stylesheets/jobListWidget.styl
```diff
@@ -49,3 +34,46 @@
...
+
+.graph-filter-container,.filter-container
+  margin-top 7px
+  > div
+    float left
+    margin-right 10px
+  > .checkbox
+    margin 0 0 0 10px
+
+.page-size-container
```
jbeezley: Girder uses a convention that custom class names are prefixed with `g-`.

matthewma7:
@jbeezley Thanks! Code has been updated.

jbeezley:
One last thing we might want to do to preserve the old behavior is make the filter and page size dropdowns optional.  Before they were included in the header, which can be disabled.

plugins/jobs/web_client/templates/JobList.pug
```diff
@@ -34,5 +34,3 @@ table.g-jobs-list-table
           td= job.get('type')
         if columns & columnEnum.COLUMN_STATUS
           td.g-job-status-cell= JobStatus.text(job.get('status'))
-
-.g-job-pagination
```
zachmullen: Let's move this file back to its original name in case any downstreams were importing it statically.

plugins/jobs/web_client/templates/JobsGraphWidget.pug
```diff
@@ -0,0 +1,8 @@
+.g-job-filter-container.clearfix
+  .timing
+  .checkbox
+    label
+      input.linear-scale(type="checkbox", checked=yScale=='linear')
+      | Use linear scale
+
+.g-jobs-graph
```
zachmullen: Let's use a consistent naming convention in our pug files, the rest of them start with lowercase.

plugins/jobs/web_client/templates/JobList.pug
matthewma7: This file has been basically split into two files. So, I don't think any code were using it could use it without change.

matthewma7:
@jbeezley I see. The filters were able to be disabled because it was part of the header row, right? Which solution should we choose for the current situation?
1. Provide two new options and default to false
2. Provide two new options and default to true
3. Respect the showHeader option and use it to determine show or not filters and page size dropdown

jbeezley:
My preference would be option 1 if it isn't too much work.

mgrauer:
Way to go @matthewma7, I appreciate your persistence in pushing this through the lengthy and choral QA process.

👨‍🎤

matthewma7:
Thank you @mgrauer, and thank you all @mgrauer, @zachmullen, @danlamanna, @jbeezley for catching things I overlooked or didn't aware before. I learned a lot in the process. 😃

PR #1905, opened 2017-04-04
Trigger global events in ControlWidget

jbeezley:
This ports the changes from [slicer_cli_web](https://github.com/girder/slicer_cli_web/pull/32/commits/54936dd09f26513630b5d3ce8fa0122a2af4d89f) to provide global lifecycle events for item_task control widgets.  This allows disconnected components to respond to changes on the views without having store references to the widget models.

### DIFF
--- a/plugins/item_tasks/plugin_tests/widget.js
+++ b/plugins/item_tasks/plugin_tests/widget.js
@@ -653,5 +653,117 @@ girderTest.promise.then(function () {
             expect(message).toBe('Invalid widget type "invalid"');
             console.warn = _warn;
         });
+
+        describe('events', function () {
+            it('g:itemTaskWidgetSet', function () {
+                var w = new itemTasks.views.ControlWidget({
+                    parentView: parentView,
+                    el: $el.get(0),
+                    model: new itemTasks.models.WidgetModel({
+                        type: 'number',
+                        title: 'Title',
+                        id: 'number-widget',
+                        value: 0
+                    })
+                });
+                w.render();
+
+                girder.events.trigger('g:itemTaskWidgetSet:' + w.model.id, 1);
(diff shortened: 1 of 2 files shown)


zachmullen:
This breaks with the normal Model-sharing pattern normally used in Backbone, is it a big hassle to share the models between multiple views?

jbeezley:
I don't entirely disagree; however, global events is a common girder client pattern.  E.g. it is the standard way in which we respond to a user logging in.  The problem I have here is similar to the login/logout pattern.  As an example in HistomicsTK, we have a toplevel view with two children.  One contains the image viewer, the other contains the control panels.  I need to image viewer to respond to setting properties on a particular widget (i.e. to draw a rectangle showing a region of interest).

Neither the control widget nor the image viewer are direct descendants of one another, so to share models it would either have to be globally or through the common descendant.  IMO, having to pass models through parent views like this limits reusability of the components because it unnecessarily couples the view with it's parent.  I also think storing the models globally is roughly equivalent to triggering global events.  Doing things this way has the benefit that there is no assumption that an object or function lives anywhere in particular, if no one is listening to the event, triggering it is a noop.

Finally, nothing here prevents you from sharing models across views, it just adds the option of listening to global events.  I can sympathize with the argument that encourages an antipattern, but this isn't all that uncommon in the backbone universe.  [Marionette's radio](https://marionettejs.com/docs/master/backbone.radio.html) is an example that is basically just global events but namespaced to avoid collisions.

zachmullen:
Everything's global at the end of the day, just wanted to make sure we're avoiding an actual problem 👍

jbeezley:
No problem.  I'm just trying to add the flexibility that will be necessary to use the item_tasks plugin in an application like HistomicsTK.  I'm happy to explore other approaches.

PR #1911, opened 2017-04-07
Unpin cherrypy version in setup.py

zachmullen:
If this fails, I'll close this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ with open('README.rst') as f:
 install_reqs = [
     'bcrypt',
     'boto',
-    'CherryPy<8',  # see https://github.com/girder/girder/issues/1615
+    'CherryPy',
     'jsonschema',
     'Mako',
     'pymongo>=3',


manthey:
This had repercussions on the slicer_cli_web plugin.  See https://github.com/girder/slicer_cli_web/pull/34.

zachmullen:
thanks @manthey , I think it would be a good idea to handle the encoding of text types in Girder's handling of raw responses, can you think of any reason not to do so?

manthey:
I think that would be good.  CherryPy < 8 was using the code:
```
if isinstance(chunk, six.text_type):
   chunk = chunk.encode('ISO-8859-1')
```

I think that if we are going to do that, it would be better to use UTF-8 than ISO-8859-1, as (to me) that would be more expected.

zachmullen:
Yeah, UTF8 for sure since that's what we use throughout Girder.

PR #1913, opened 2017-04-07
Only run packaging tests on master and release branches

zachmullen:
This will make CI response time a lot faster.

### DIFF
--- a/cmake/circle_continuous.cmake
+++ b/cmake/circle_continuous.cmake
@@ -4,8 +4,9 @@ set(CTEST_BINARY_DIRECTORY "$ENV{HOME}/build")
 include(${CTEST_SOURCE_DIRECTORY}/CTestConfig.cmake)
 
 set(test_group $ENV{TEST_GROUP})
+set(branch $ENV{CIRCLE_BRANCH})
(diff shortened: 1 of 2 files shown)


cmake/circle_continuous.cmake
```diff
@@ -20,15 +21,22 @@ ctest_start("Continuous")
...
   ctest_test(
     PARALLEL_LEVEL 4 RETURN_VALUE res
-    INCLUDE_LABEL girder_python
+    INCLUDE_LABEL "${_label}"
   )
 elseif(test_group STREQUAL browser)
   ctest_test(
     PARALLEL_LEVEL 4 RETURN_VALUE res
-    EXCLUDE_LABEL girder_python
+    EXCLUDE_LABEL "girder_python|girder_package"
```
mgrauer: would `EXCLUDE_LABEL "${_label}"` work here?  if so, it's a bit more consistent with the INCLUDE_LABEL

zachmullen: Sadly no, it would cause the packaging tests to be run in the browser group on all branches that didn't match.

mgrauer:
Now we get to see if the packaging tests are run on master.

PR #1914, opened 2017-04-07
fix #1756 - reset plugin webroot entries only when disabling the plugin

### LINKED ISSUE #1756: Plugin routing gets reset on syntax error.
- Begin developing your own plugin
 - Set the system-wide plugin routing setting so that your plugin serves its contents from "/" and so that girder-core is served from "/girder".
 - While working on your server-side code, "accidentally" introduce a syntax error.

At this point, Girder auto-restarts, sees that your plugin is broken, and responds with a number of actions that seem to include resetting the plugin's routing.

Therefore, when you fix your code and go back, your plugin is serving on "/PLUGIN_NAME" instead of where you left it.

I propose that this behavior be regarded as a bug as the less surprising behavior, imo, would be to leave the routing alone and having your plugin work again once you fix your syntax error.  Correction of such a bug would greatly improve the plugin developer experience.

opadron:
fixes #1756

### DIFF
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -157,6 +157,7 @@ class System(Resource):
         .errorResponse('You are not a system administrator.', 403)
     )
     def enablePlugins(self, plugins, params):
+        # Determine what plugins have been disabled and remove their associated routes.
         setting = self.model('setting')
         routeTable = setting.get(SettingKey.ROUTE_TABLE)
         oldPlugins = setting.get(SettingKey.PLUGINS_ENABLED)
@@ -174,7 +175,8 @@ class System(Resource):
         if routeTableChanged:
             setting.set(SettingKey.ROUTE_TABLE, routeTable)
 
-        return self.model('setting').set(SettingKey.PLUGINS_ENABLED, plugins)
+        # Route cleanup is done; update list of enabled plugins.
+        return setting.set(SettingKey.PLUGINS_ENABLED, plugins)
 
 
     @access.admin
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -27,7 +27,8 @@ import os
 import logging
 
 from girder.api import access
-from girder.constants import SettingKey, TokenScope, ACCESS_FLAGS, VERSION
+from girder.constants import GIRDER_ROUTE_ID, GIRDER_STATIC_ROUTE_ID, \
+                             SettingKey, TokenScope, ACCESS_FLAGS, VERSION
 from girder.models.model_base import GirderException
 from girder.utility import install, plugin_utilities, system
 from girder.utility.progress import ProgressContext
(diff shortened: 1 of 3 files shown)


girder/api/v1/system.py
```diff
@@ -156,8 +157,35 @@ def getVersion(self, params):
...
         return self.model('setting').set(SettingKey.PLUGINS_ENABLED, plugins)
 
+    def reconcileRouteTable(routeTable):
+        hasChanged = False
+
+        # GIRDER_ROUTE_ID is a special route, which can't be removed
+        ## for name in routeTable.keys():
+        ##     if name not in reservedRoutes and name not in pluginWebroots:
+        ##         del routeTable[name]
+        ##         hasChanged = True
```
zachmullen: Let's remove this commented code, and this function now seems to be a no-op.

girder/utility/server.py
```diff
@@ -174,10 +174,10 @@ def reconcileRouteTable(routeTable):
...
 
         # GIRDER_ROUTE_ID is a special route, which can't be removed
-        for name in routeTable.keys():
-            if name not in reservedRoutes and name not in pluginWebroots:
-                del routeTable[name]
-                hasChanged = True
+        ## for name in routeTable.keys():
+        ##     if name not in reservedRoutes and name not in pluginWebroots:
+        ##         del routeTable[name]
+        ##         hasChanged = True
```
zachmullen: Same here

girder/api/v1/system.py
```diff
@@ -156,8 +157,27 @@ def getVersion(self, params):
         .errorResponse('You are not a system administrator.', 403)
     )
     def enablePlugins(self, plugins, params):
+        pluginWebroots = plugin_utilities.getPluginWebroots()
+        setting = self.model('setting')
+        routeTable = setting.get(SettingKey.ROUTE_TABLE)
+        oldPlugins = setting.get(SettingKey.PLUGINS_ENABLED)
+        reservedRoutes = (GIRDER_ROUTE_ID, GIRDER_STATIC_ROUTE_ID)
```
zachmullen: If we just make `reservedRoutes = {GIRDER_ROUTE_ID, GIRDER_STATIC_ROUTE_ID}`, we won't have to copy it into a set further down.

girder/utility/server.py
```diff
@@ -173,12 +173,6 @@ def loadRouteTable(reconcileRoutes=False):
     def reconcileRouteTable(routeTable):
         hasChanged = False
 
-        # GIRDER_ROUTE_ID is a special route, which can't be removed
```
danlamanna: The docstring for `loadRouteTable` needs to be updated to reflect that it's no longer dealing with "no longer active" plugins.

girder/api/v1/system.py
```diff
@@ -156,8 +157,27 @@ def getVersion(self, params):
...
+
+        routeTableChanged = False
+        removedRoutes = (
```
danlamanna: This method deserves a docstring (or at least inline comments).

opadron:
@danlamanna let me know if those comments I added seem reasonable.

opadron:
@zachmullen @danlamanna PTAL

opadron:
Pinging again on this.  Let me know if I should add anything else.  It will be a great convenience for plugin developers to get this merged.

PR #1915, opened 2017-04-10
Improve error checking when user attempts to upload a directory

### LINKED ISSUE #1489: Dragging and dropping a directory onto uploader causes issue
1. Open finder/browser, select a folder, drag onto upload dialog. A message appears indicating everything is OK (even though this isn't supported).
2. Hit the start upload button. Error message appears, though not a good one.

msmolens:
Add functionality to the upload widget to check whether every dropped
item is a file. Display an error message when the check fails.

Note that the implementaion uses the DataTransferItem.webkitGetAsEntry()
method, which has not yet been standardized, but first tries to use the
suggested future standard method name getAsEntry(). See
https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem.

Fixes #1489.

### DIFF
--- a/clients/web/src/views/widgets/UploadWidget.js
+++ b/clients/web/src/views/widgets/UploadWidget.js
@@ -176,7 +176,15 @@ var UploadWidget = View.extend({
         this.$('.g-drop-zone')
             .removeClass('g-dropzone-show')
             .html(`<i class="icon-docs"/> ${this._browseText}`);
-        this.files = e.originalEvent.dataTransfer.files;
+
+        var dataTransfer = e.originalEvent.dataTransfer;
+
+        // Require all dropped items to be files
+        if (!_.every(dataTransfer.items, _.bind(this._isFile, this))) {
+            this.$('.g-upload-error-message').html('Only files may be uploaded.');
+            return;
+        }
+        this.files = dataTransfer.files;
 
         if (!this.multiFile && this.files.length > 1) {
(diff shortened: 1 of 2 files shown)


clients/web/src/views/widgets/UploadWidget.js
```diff
@@ -176,7 +176,15 @@ var UploadWidget = View.extend({
         this.$('.g-drop-zone')
             .removeClass('g-dropzone-show')
             .html(`<i class="icon-docs"/> ${this._browseText}`);
-        this.files = e.originalEvent.dataTransfer.files;
+
+        var dataTransfer = e.originalEvent.dataTransfer;
+
+        // Require all dropped items to be files
+        if (!_.every(dataTransfer.items, _.bind(this._isFile, this))) {
```
zachmullen: Nowadays we favor using arrow functions instead of _.bind:

`_.bind(this._isFile, this)` -> `() => this._isFile()`

zachmullen:
Works for me in chrome, nice fix!

clients/web/src/views/widgets/UploadWidget.js
msmolens: That doesn't work for me. How does it pass the item to the method?
How about:
`(item) => this._isFile(item)`

zachmullen: Yeah sorry, yours is correct, didn't realize it had params

PR #1916, opened 2017-04-10
Newly created jobs show on the jobs page immediately

### LINKED ISSUE #1833: Jobs page should automatically show new jobs
The eventstream is used for progress bars of new jobs, but the actual new job doesn't display on the jobs page until it's manually refreshed.

danlamanna:
Fixes #1833

### DIFF
--- a/plugins/jobs/plugin_tests/jobsSpec.js
+++ b/plugins/jobs/plugin_tests/jobsSpec.js
@@ -150,14 +150,33 @@ $(function () {
             }, 'job list to auto-reload when collection is updated');
 
             runs(function () {
+                girder.utilities.eventStream.trigger('g:event.job_created', {
+                    data: {
+                        _id: 'foo' + 4,
+                        title: 'My batch job ' + 4,
+                        status: 4,
+                        updated: '2015-01-12T12:00:0' + 4,
+                        created: '2015-01-12T12:00:0' + 4,
+                        when: '2015-01-12T12:00:0' + 4
+                    }
+                });
+            });
+
+            waitsFor(function () {
+                return $('.g-jobs-list-table>tbody>tr').length === 4;
+            }, 'job list to auto-reload when job_created is triggered');
+
+            runs(function () {
                 // Make sure we are in reverse chronological order
                 rows = $('.g-jobs-list-table>tbody>tr');
-                expect($(rows[0]).text()).toContain('My batch job 3');
-                expect($(rows[0]).text()).toContain('Success');
-                expect($(rows[1]).text()).toContain('My batch job 2');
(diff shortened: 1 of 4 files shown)


danlamanna:
Whoever has some time feel free to review. I ran into a weird issue where I discovered the JS tests already existed to test the feature this PR adds... and they've always passed (I have no idea how this is the case).

zachmullen:
Could you link to the web client tests you mentioned?

danlamanna:
@zachmullen https://github.com/girder/girder/blob/a801587ee2fa225d48f74966add9350186081d01/plugins/jobs/plugin_tests/jobsSpec.js#L138-L146

The issue I was encountering on this branch that I fixed, is that `collection.add` doesn't trigger a render on the widget, so I have no clue how this test is triggering render.

zachmullen:
https://github.com/girder/girder/blob/master/plugins/jobs/web_client/views/JobListWidget.js#L43-L45

Doesn't that explain it...?

zachmullen:
Or not, I guess `add()` doesn't trigger `g:changed`... We may want to add some screenshots before and after that to make sure the test isn't doing something unexpected.

danlamanna:
It definitely **is** doing something wrong, I just couldn't figure out how to really debug in the environment aside from console.log calls. The `console.log` calls show it doing what it asserts, but that doesn't happen in my browser.

matthewma7:
@danlamanna, I think I noticed the weirdness you are talking about, and @manthey help me figured it out. 
https://github.com/girder/girder/blob/a801587ee2fa225d48f74966add9350186081d01/plugins/jobs/plugin_tests/jobsSpec.js#L136-L136, the render() at the end is triggering the render()

I changed it to this in my jobs graph branch, which I think making more sense to me, meaning, ignore the first render() and explicitly trigger one with `widget.collection.trigger('g:changed');`
https://github.com/girder/girder/blob/jobs_graph/plugins/jobs/plugin_tests/jobsSpec.js#L121-L151

Now I am looking at it, I think the test is not meant for this feature specifically. The code snippet has been used a few times in the jobspec.js

zachmullen:
As a sanity check, I did verify that the jobs appeared in the list in the jobsSpec test. This is on latest master branch. Before the call to collection.add:



And then immediately after the `waitsFor`:

danlamanna:
@matthewma7 Not really sure how your code block works, but it did fix the issue (meaning master is broken and my branch fixes it).

danlamanna:
@matthewma7 I switched the job list widget in this PR to listen to the `add` event, so you should be able to remove the `g:changed` trigger in your test when you rebase your branch.

This is ready for review.

PR #1918, opened 2017-04-11
Add job link to progress notifications

### LINKED ISSUE #1853: Job notifications should link to job details page
For example, a notification about a job should have some link referring to `#job/:jobId`.

danlamanna:
Fixes #1853 

@manthey Do you have any thoughts for a more elegant (backwards compatible) solution than this?

### DIFF
--- a/clients/web/src/templates/widgets/taskProgress.pug
+++ b/clients/web/src/templates/widgets/taskProgress.pug
@@ -6,6 +6,11 @@
           = percentText
         .progress-left
           = timeLeftText
-  .g-task-progress-title(title=progress.data.title)= progress.data.title
+  .g-task-progress-title(title=progress.data.title)
+    if progress.data.resource && progress.data.resourceName
+      a(title=progress.data.title, href='#' + progress.data.resourceName + '/' + progress.data.resource._id)
+        = progress.data.title
+    else
+      = progress.data.title
 .g-task-progress-bottom
   .g-task-progress-message= progress.data.message
--- a/clients/web/src/templates/widgets/taskProgress.pug
+++ b/clients/web/src/templates/widgets/taskProgress.pug
@@ -7,7 +7,7 @@
         .progress-left
           = timeLeftText
   .g-task-progress-title(title=progress.data.title)
-    if progress.data.resource && progress.data.resourceName
+    if progress.data.resource && progress.data.resource._id && progress.data.resourceName
       a(title=progress.data.title, href='#' + progress.data.resourceName + '/' + progress.data.resource._id)
         = progress.data.title
     else
--- a/clients/web/test/spec/widgetsSpec.js
+++ b/clients/web/test/spec/widgetsSpec.js
@@ -5,15 +5,25 @@
  */
 girderTest.startApp();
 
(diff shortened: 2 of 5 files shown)


girder/models/notification.py
```diff
@@ -92,7 +92,7 @@ def createNotification(self, type, data, user, expires=None, token=None):
         return self.save(doc)
 
     def initProgress(self, user, title, total=0, state=ProgressState.ACTIVE,
-                     current=0, message='', token=None, estimateTime=True):
+                     current=0, message='', token=None, estimateTime=True, job_id=None):
```
zachmullen: jobId

manthey:
Since jobs are a plugin, I think we'd rather abstract this at least a bit.  For instance, make the notification initProgress take an id and a resource type.  That way, if the progress were about an item, it could link to that in the same manner.

manthey:
Since jobs are a plugin, they shouldn't be mentioned explicitly in the non-plugin core.

danlamanna:
That's right, I keep forgetting jobs is a plugin.

danlamanna:
@manthey If this generally looks okay, I'll write some tests for it tomorrow.

manthey:
@danlamanna Looks reasonable; proceed with testing.

danlamanna:
This is ready for review.

It seems like `resourceType` is the preferred naming in Girder Client/Worker, but `resourceName` is throughout Girder core.

clients/web/test/spec/widgetsSpec.js
```diff
@@ -5,15 +5,25 @@
...
+     *     is associated with.
+     * : param resourceName: optional resource type that the progress notification
+     *     is associated with.
      */
     girder.rest.restRequest({path: 'webclienttest/progress', type: 'GET',
-        data: {test: test, duration: duration}});
+        data: {
+            test: test,
+            duration: duration,
+            resourceId: resourceId,
```
manthey: Why do we use `resourceId` here, but elsewhere we use an object `resource` and have to get its id via `resource._id`?

zachmullen:
Just verified that this works.

clients/web/test/spec/widgetsSpec.js
danlamanna: The API supports a partial or complete resource, as long as an id is provided (so we have somewhere to link). The reason for this is different resources might have different requirements, i.e. a job wants to pass along its id but maybe not the `logs` field which can have a lot of data.

For that reason, I opted to just provide the id when testing.

clients/web/src/templates/widgets/taskProgress.pug
```diff
@@ -6,6 +6,11 @@
           = percentText
         .progress-left
           = timeLeftText
-  .g-task-progress-title(title=progress.data.title)= progress.data.title
+  .g-task-progress-title(title=progress.data.title)
+    if progress.data.resource && progress.data.resourceName
```
manthey: Since in the test you demonstrate having the resource the simple ID rather than an object with an _id element, perhaps this test should be `if progress.data.resource && progress.data.resource._id && progress.data.resourceName`.

PR #1920, opened 2017-04-12
Small documentation improvements on entering virtual environment

matthewma7:
The documentation was like this
chrome_2017-04-12_14-40-30

I changed to this
chrome_2017-04-12_14-46-59

The entering virtual environment wasn't inside a code block, which is a little bit inconsistent with other steps. And, I personally tend to overlook stuff that is not inside code blocks. So I made this change, hoping it would improve the documentation.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -27,7 +27,7 @@ Enter the virtual environment:
 
 .. code-block:: none
 
-   source ~/girder_env/bin/activate
+   . ~/girder_env/bin/activate
    
 The ``(girder_env)`` prepended to your prompt indicates you have *entered* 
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -23,8 +23,14 @@ not be moved.  The following command will generate a new directory called
 
    virtualenv ~/girder_env
 
-Now you can run ``source ~/girder_env/bin/activate`` to *enter* the new
-virtual environment.  Inside the virtual environment you can use ``pip``,
+Enter the virtual environment: 
+
+.. code-block:: none
+
+   source ~/girder_env/bin/activate
```
zachmullen: This is a good change -- while we're at it, I vote we change from `source` to `.` for shell portability.

PR #1922, opened 2017-04-13
Remove the Webpack url-loader

brianhelba:
The 'url-loader' provides a wrapper around 'file-loader', so the two loaders should not be chained together. This bug was not discovered, as Girder was not loading any '.woff', '.ttf', or '.svg' files with Webpack.

Rather than fix the chaining bug, this just removes the url-loader, as its use is probably a premature optimization. Using url-loader will theoretically improve first-load performance, but makes it more difficult for user agents to cache assets and will interfere with the effectiveness of HTTP2 in the future.

### DIFF
--- a/grunt_tasks/webpack.config.js
+++ b/grunt_tasks/webpack.config.js
@@ -36,20 +36,6 @@ function fileLoader() {
     };
 }
 
-function urlLoader(options) {
-    options = options || {};
-    var loader = {
-        loader: 'url-loader',
-        query: {
-            limit: 4096
-        }
-    };
-    if (options.mimetype) {
-        loader.query.mimetype = options.mimetype;
-    }
-    return loader;
-}
-
 function _coverageConfig() {
     try {
         var istanbulPlugin = require.resolve('babel-plugin-istanbul');
@@ -151,7 +137,6 @@ module.exports = {
                 test: /\.woff(\?v=\d+\.\d+\.\d+)?$/,
                 include: loaderPathsNodeModules,
                 loaders: [
-                    urlLoader({ mimetype: 'application/font-woff' }),
                     fileLoader()
                 ]
             },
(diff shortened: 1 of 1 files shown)


sebastienbarre:
Yes, the reason for this is to indeed improve first-load performance, since it can return a Data Url if the file is smaller than a byte limit. It's valid case *today*, since HTTP2 is not widely used. I would recommend it over separate, individual requests to small files. My preference nowadays would be to directly embed any icon in SVG format (something you can do easily with React for example, or this loader). Fonts will not be embedded by the url loader and should be cached by the user agent (Girder wasn't using fonts?), but my preference nowadays is to use system fonts (everything old is new again).

Your call.

package.json
```diff
@@ -52,7 +52,6 @@
     "swagger-ui": "2.2.10",
     "toposort": "^1.0.0",
     "underscore": "~1.8.3",
-    "url-loader": "^0.5.7",
```
zachmullen: We may want to leave this here, simply in case downstream plugins are expecting it to be present as a loader they can refer to.

zachmullen:
Aside from being an anti-pattern w.r.t. HTTP 2, there are a couple other downsides to embedding certain types of files as data URLs:

1. Base-64 encoding the data causes it to increase in size by 1/3.
2. If they're embedded as data URLs, the browser cannot delay fetching them until they are needed by, for instance, CSS rules that may not be applicable as soon as the app starts.

Since this only affected small files, neither of those is a big deal, but removing this loader from our build does reduce complexity a bit, which is a net benefit IMO.

sebastienbarre:
OK.
Speaking of HTTP2, I was reading [A Comprehensive Guide To HTTP/2 Server Push](https://www.smashingmagazine.com/2017/04/guide-http2-server-push/), and it sure is going to introduce some complexity as well. The benefits are not quite obvious at the moment.

brianhelba:
My optimization concern was less about HTTP/2 and more about preventing modern user agents from being able to aggressively cache these resources. Since these resources are truly static (if they do internally change, they will end up with a different filename), we should cooperate and bundle them separately from JS code that may change relatively more frequently.

@zachmullen 's point about delayed fetching is also more relevant to my thinking.

brianhelba:
I'll re-include the NPM package. We should drop it in the next major release though.

PR #1923, opened 2017-04-13
Attempt to support non-encoded raw responses

zachmullen:
If we are given a rawResponse that is a text type, we have to try and guess how to send it along the wire. The safest bet is to use UTF-8.


tests/cases/rest_decorator_test.py
```diff
@@ -94,3 +94,7 @@ def testRawResponse(self):
         resp = self.request('/other/rawInternal', isJson=False)
         self.assertStatusOk(resp)
         self.assertEqual(self.getBody(resp), 'this is also a raw response')
+
+        resp = self.request('/other/rawReturningText', isJson=False)
+        self.assertStatusOk(resp)
+        self.assertEqual(self.getBody(resp), 'this is not encoded')
```
manthey: We want to test this with actual calls, so that we'd see the failure condition in the test.  The test has to specify BIND_SERVER in the CMakeLists file and set a port at the top.

The actual test would be something like:
```
import requests

def testEndpointUnicode(self):
    req = requests.get(
            'http://127.0.0.1:%d/api/v1/other/rawReturningText' % (
                int(os.environ['GIRDER_PORT'])))
        self.assertEqual(req.status_code, 200)
        self.assertEqual(req.text, 'this is not encoded')
```

I'd also add an upper code point character in the string to verify that it comes through as UTF-8 (say `\U0001F44D` and come through as `\xf0\x9f\x91\x8d`.

zachmullen:
@manthey PTAL

tests/cases/rest_decorator_test.py
```diff
@@ -19,16 +19,20 @@
 
 import json
 import os
+import requests
```
manthey: Do we need to add requests to the requirements-dev.txt file (it is in the client/python/requirements.txt file)?

zachmullen: It's already a core requirement, so we don't strictly need to, but maybe for explicitness...? I'll let you decide.

tests/test_plugins/test_plugin/server.py
```diff
@@ -77,6 +78,13 @@ def rawWithDecorator(self, params):
         return b'this is a raw response'
 
     @access.public
+    @rawResponse
+    @describeRoute(None)
+    def rawReturningText(self, params):
+        setResponseHeader('Content-Type', 'text/plain')
```
manthey: What occurs if an the endpoint doesn't set the `Content-Type` header?

zachmullen: cherrypy defaults to `text/html;charset=utf-8`, at least in its current version.

tests/cases/rest_decorator_test.py
manthey: Somehow I missed seeing it in the core requirements.

PR #1928, opened 2017-04-14
Allow OAuth plugin to override closed core registration policy

zachmullen:


mgrauer:
This is to close registration except for those who have OAuth access?

Seems reasonable generally, though I have a nagging feeling about allowing a plugin to override a system setting.  Do we do this general thing (override a system setting through a plugin) anywhere else?  

I'm probably overly worried about it, but it would make me feel better if you added something to the core system setting UI for closing registration (the text in the drop down) to say it can be overridden by OAuth.  What I don't want is for someone to get a false impression by only looking at the system setting and thinking registration is closed when it is not.

zachmullen:
The default for this setting is that the core setting is honored, so site admins will have to explicitly override this.

> Do we do this general thing (override a system setting through a plugin) anywhere else?

Not sure, but it's definitely something we want to allow. Generally speaking, we don't go out of our way to artificially limit the power of plugins.

But, if people feel strongly that this needs a warning on the core settings page, we can add one.

mgrauer:
>The default for this setting is that the core setting is honored, so site admins will have to explicitly override this.

This makes sense, and once I went through the workflow of setting this on my site I see exactly what you mean, hence my concern has disappeared.

When the site is closed, you can enter all your info before it tells you that the site is closed, so it would be nice to see that beforehand.  I looked at how the RegisterView works, and I see it just delegates all this to the server, so it can stay simple, and to display a message at the outset would add a bit of complexity.  I'm fine with this as, but it would be nice, i.e., a lukewarm ask.

I need to do a bit more testing in the field, but so far this seems good.

zachmullen:
> Mention closed registration on register UI, if it's easy.

I'll open an issue for this, I think it's something that will be simple once we convert the Setting model to be an AccessControlledModel -- the registration setting can be made public at that point, and the UI can do a lookup for its value.

PR #1931, opened 2017-04-17
Always include API version in system/version response

zachmullen:
This fixes a bug when the girder web client does not exist. In that
case, the API_VERSION value is set to null, even though there's no
reason for us not to report it since it's known in the backend.

This commit changes the behavior by always ensuring API_VERSION
is set to the package version, regardless of the existence of
girder-version.json.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -332,7 +332,9 @@ class GirderClient(object):
         :type useCached: bool
         :return: The API version as a list (e.g. ``['1', '0', '0']``)
         """
-        return self.getServerAPIDescription(useCached)["info"]["version"].split('.')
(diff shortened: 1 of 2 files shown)


brianhelba:
If it's possible, +1 to removing the usage of Grunt / `girder-version.json` to set server parameters. It's messy for the case where the web client isn't built or installed, and I've noticed that in some installation configurations, the file isn't getting written to the correct location.

On this PR, I think you should remove https://github.com/girder/girder/blob/master/grunt_tasks/version.js#L35 , to prevent the `apiVersion` from getting set multiple times.

zachmullen:
Agreed. For now I'm leaving girder-version.json in place in case anybody is relying on it, but I'll open an issue for this.

PR #1932, opened 2017-04-18
Add client test for homepage plugin

### LINKED ISSUE #1373: Add client side tests for homepage plugin
follow up to #1370 .

Have the test set markdown and then test that the homepage renders that markdown.

msmolens:
Fixes #1373.

### DIFF
--- a/plugins/homepage/plugin_tests/homepageSpec.js
+++ b/plugins/homepage/plugin_tests/homepageSpec.js
@@ -1,0 +1,110 @@
+/* globals girder, girderTest, describe, expect, it, runs, waitsFor */
+
+girderTest.addCoveredScripts([
+    '/clients/web/static/built/plugins/homepage/plugin.min.js'
+]);
+
+girderTest.startApp();
+
+function _goToHomepagePluginSettings() {
+    waitsFor(function () {
+        return $('a[g-target="admin"]:visible').length > 0;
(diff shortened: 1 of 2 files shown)


plugins/homepage/plugin_tests/homepageSpec.js
```diff
@@ -0,0 +1,110 @@
...
+    waitsFor(function () {
+        return girder.rest.numberOutstandingRestRequests() === 0;
+    }, 'rest requests to finish');
+}
+
+function _verifyMarkdownContent(elem) {
+    expect(elem.find('p:contains("It\'s very easy")').length).toBe(1);
+    expect(elem.find('strong:contains("bold")').length).toBe(1);
+    expect(elem.find('em:contains("italic")').length).toBe(1);
+    expect(elem.find('a[href="https://girder.readthedocs.io/"]:contains("link to Girder!")').length).toBe(1);
```
zachmullen: For what it's worth, there is a `toContain()` jasmine function that we could use instead of using the selector + length property.

msmolens: Yes. Is that preferable here?

zachmullen: Not sure, your call. This is good to merge as is if you are OK with how it's written now.

PR #1934, opened 2017-04-19
When piping install output, strip out ANSI escape codes.

manthey:
This is intended to make the web client rebuilding messages somewhat less ugly.

### DIFF
--- a/girder/utility/install.py
+++ b/girder/utility/install.py
@@ -24,6 +24,7 @@ be restarted for these changes to take effect.
 
 import os
 import pip
+import re
 import select
 import shutil
 import six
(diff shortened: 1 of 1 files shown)


girder/utility/install.py
```diff
@@ -104,6 +105,8 @@ def _pipeOutputToProgress(proc, progress):
                 buf = os.read(pipe.fileno(), 1024)
                 if buf:
                     buf = buf.decode('utf8', errors='ignore')
+                    # Remove ANSI escape sequences
+                    buf = re.sub('(\x9B|\x1B\[)[0-?]*[ -\/]*[@-~]', '', buf)
```
zachmullen: Is there a link to where this regex comes from, or did you invent it just from reading the standard?

zachmullen: Also, should we be acting on a string in this case or on the raw bytes?

manthey: There is a discussion here: http://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python and here: https://en.wikipedia.org/wiki/ANSI_escape_code .

It shouldn't matter whether this is done on the bytes or the unicode string, as all of the characters involved get passed through without modification.  It would be mildly more efficient to operate on the bytes, since any length reduction would reduce the decoding cost.  I can change it, if you like.

Issue #1919, opened 2017-04-12
Capture all stdout and stderr into the logs


manthey:
Currently, there are a few exceptions that can be emitted on stderr that aren't logged.  Generally, it would be nice if anything emitted to stdout and stderr would be sent to the logs as well, as that would give a single location to check.

Similarly, we currently divide information between the info and error logs by default.  We can merge these together by specifying a single file for both, but I think it might be handier to log everything to info while still logging errors to the error log (to have separate levels for each of the logs), that way you can tail the info log to watch everything that is happening but check the error log for recent errors.

danlamanna:
>  I think it might be handier to log everything to info while still logging errors to the error log (to have separate levels for each of the logs), that way you can tail the info log to watch everything that is happening but check the error log for recent errors.

+10

manthey:
Added in PR #1935.

PR #1935, opened 2017-04-19
More logging options.

manthey:
This adds another logging configuration setting to allow the info log to be a super-set of the error log.  If log_max_info_level is set to 'CRITICAL', for instance, then all log messages will go to the info log while errors and above will also go to the error log.

Additionally, all output from stdout will be logged at the INFO level (except things sent via the logprint method), and all output from stderr will be logged at the ERROR level.

With these changes, setting log_max_info_level = 'CRITICAL' and log_access to include 'info', then ALL output should appear in the info log, giving a single point that needs to be monitored to observe the system.


manthey:
This resolves issue #1919.

mgrauer:
QAing this will be good to get @zackgalbreath deeper into Girder core.  

I'm sure @zachmullen @danlamanna @manthey can help answer any questions along the way.

girder/__init__.py
```diff
@@ -74,6 +76,28 @@ def format(self, record, *args, **kwargs):
...
+    Redirect a file-like stream to a logger.
+    """
+    def __init__(self, stream, logger, level):
+        self.stream = stream
+        self.logger = logger
+        self.level = level
+        self.logger._girderLogHandlerOutput = False
+        for key in dir(stream):
+            if key != 'write' and not key.startswith('_') and callable(getattr(stream, key)):
+                setattr(self, key, getattr(stream, key))
```
manthey: We are overriding the default `sys.stdout` and `sys.stderr` with the logger method.  However, I want to preserve as much functionality for stdout and stderr as possible, so that other modules that send data to them can do so without a problem.  The only method we really need to override is `write`, but if we cannot mutate the `write` method on the stream itself, so we replace the stream with this custom class.  To preserve the stream methods, all of them get added to our class instance *except* private and built-in methods, which, in python, begin with `_`.

Fundamentally, it lets our stream replacement handle functions `flush`, `writeline`, and others without having to enumerate them individually.

zackgalbreath: Makes sense, thanks for the clarification.

manthey:
I'll add some more tests for `logPrint`.

girder/__init__.py
mgrauer: @manthey Your explanation here was really useful in understanding what wasn't obvious to me otherwise.  Can you add this info as a comment in the code?

manthey:
@zackgalbreath I've added some tests for `logprint`.  @mgrauer I've added comments to explain the `StreamToLogger` class in more detail.

PR #1937, opened 2017-04-19
Allow coverage paths to be overridden by add_python_test

jbeezley:
See the discussion in https://github.com/DigitalSlideArchive/HistomicsTK/pull/291 for background.

### DIFF
--- a/docs/developer-cookbook.rst
+++ b/docs/developer-cookbook.rst
@@ -457,7 +457,8 @@ core tests, the following paths are used:
 For tests created within a plugin, the plugin's ``server`` directory is
 added to this list.  If you wish to report coverage on files residing outside
 of one of these directories, you can add a ``COVERAGE_PATHS`` argument
-to the ``add_python_test`` call.  For example, to add coverage reporting only
+to the ``add_python_test`` call.  This argument accepts one or more
+comma-separated paths.  For example, to add coverage reporting to
 to python files in a plugin's ``utils`` directory for a given test:
 
 .. code-block:: cmake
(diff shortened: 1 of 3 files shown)


docs/developer-cookbook.rst
```diff
@@ -442,6 +442,31 @@ by the environment variable ``GIRDER_TEST_DATA_PREFIX`` as follows
...
+will only report coverage for python files within certain paths.  For
+core tests, the following paths are used:
+
+- ``girder``
+- ``clients/python/girder_client``
+
+For tests created within a plugin, the plugin's ``server`` directory is
+added to this list.  If you wish to report coverage on files residing outside
+of one of these directories, you can add a ``COVERAGE_PATHS`` argument
+to the ``add_python_test`` call.  For example, to add coverage reporting only
```
zachmullen: May be worth mentioning that multiple directories can be supplied as a comma-separated list.

tests/PythonTests.cmake
```diff
@@ -103,6 +103,12 @@ function(add_python_test case)
     set(other_covg "")
   endif()
 
+  if(fn_COVERAGE_PATHS)
+    set(coverage_source_paths "${fn_COVERAGE_PATHS}")
```
zachmullen: Should this overwrite the existing entries, or append to them?

jbeezley: I could go either way.  I did this to make it easy to uninclude default paths.  I think the same could be achieve with an `omit` in the coveragerc file.

PR #1938, opened 2017-04-19
Show plugin load failures on plugins page

### LINKED ISSUE #817: Show plugin loading failures in UI
When a plugin fails to load due to an exception, girder prints the error to the console and continues running, just without the plugin. However, if you go to the plugin page it will show the plugin as enabled, and users may be confused as to why the functionality isn't there. We should remember which plugins failed to load and the stack trace and put a red "X" or something next to them to indicate they failed to load, and when the user clicks the X, the stack trace will be shown.

msmolens:
Plugins that fail to load are displayed with a reddish background and have an
"Error" tag. Clicking the tag shows the failure traceback.

Fixes #817.

image


msmolens:
@zachmullen I'm not sure how to meet the codecov/patch requirements here. I see a couple problems:
- Packaging tests are not run for PR branches (https://github.com/girder/girder/pull/1913), so CI won't test some changes in plugin_utilities.py, and omitting those tests counts against line coverage.
- `findEntryPointPlugins()` isn't showing as being covered when viewing codecov on master, but that should be covered by the `packaging_install` test, at least.

zachmullen:
The packaging tests aren't run with coverage.  A unit test should be good for covering the code you've added to the backend, and we can mock things if necessary.

msmolens:
@zachmullen Sounds good. In that case, how can I satisfy the codecov/patch CI check?

zachmullen:
You could mock `pkg_resources.iter_entry_points` in a unit test.

msmolens:
@zachmullen I rebased to bring in the `web_client_jobs.jobs` fix and added additional tests to meet the Python test coverage requirements.

clients/web/src/views/body/PluginsView.js
```diff
@@ -109,11 +115,24 @@ var PluginsView = View.extend({
...
             container: this.$el,
             animation: false,
             delay: { show: 100 }
         });
+        this.$('.g-plugin-list-item-failed-notice').popover({
+            container: this.$el,
+            template: '<div class="popover g-plugin-list-item-failed-notice-popover" role="tooltip">' +
+                      '<div class="arrow"></div><h3 class="popover-title"></h3>' +
+                      '<div class="popover-content g-plugin-list-item-failed-notice-popover-content"></div>' +
+                      '</div>'
```
zachmullen: Should we put this into a template and import it?

zachmullen:
Thanks @msmolens , great feature!

PR #1939, opened 2017-04-19
Item tasks improvements

zachmullen:


jeffbaumes:
This is very nice. The provenance information is great and very discoverable.

I think rerunning a task could be made just a bit more discoverable. Maybe a "visit setup page for this task" text that is linked - I am not being very creative this morning, I'm sure there's a better/shorter link text.

jeffbaumes:
Looks like input files show as their Girder IDs instead of item name when re-running. It would be nice if it showed the item name like when you originally set up a task.

jeffbaumes:
Link to re-run looks much better. Maybe it could be "with these same parameters" instead of "from these same parameters"?

zachmullen:
> Maybe it could be "with these same parameters" instead of "from these same parameters"?

This was actually somewhat deliberate, I was trying to convey that it isn't necessarily going to run with the exact same parameters, but that these parameters are just a starting point in the configuration. I struggled with this text for a while, can you think of something good that conveys that behavior?

jeffbaumes:
I see your reasoning @zachmullen. Let's keep that text as-is for now. Any potential solutions for the input file display issue?

zachmullen:
Yep, that's coming soon.

jeffbaumes:
I can confirm the file name is being set correctly now. This is still set as WIP - is it ready for final review?

zachmullen:
I'll add a quick test for the new route then it should be good.

zachmullen:
Ready for review

PR #1940, opened 2017-04-19
Fix conditional so it works in python 3

zachmullen:
@TristanWright this fixes a bug on python 3, resulting from the fact that dict_keys are not iterable.

I changed the condition to simply check whether '$set' was in the dict, which is really what we want to know anyway.


zachmullen:
Refs #1898

PR #1943, opened 2017-04-20
Fix race condition in web_client_jobs.jobs

zachmullen:


matthewma7:
@zachmullen, I think I have addressed some of this in the jobs graph PR https://github.com/girder/girder/pull/1901

zachmullen:
This has succeeded several times now, I'm fairly confident that it has fixed the issue: creating a JobListWidget causes an XHR to be sent, but we were not waiting for it to finish before calling `add()` on the collection.

zachmullen:
Sorry for the line changes, I ran `eslint --fix` on jobsSpec.js, it had a ton of style issues. The essence of the fix is captured by the first commit, and that same pattern was being repeated in several other places, so we actually had several race conditions.

PR #1945, opened 2017-04-21
Allow non-JSON responses via the python client methods

zachmullen:
This fixes an issue @dgutman ran into with the python client. There was no way to use its core methods to receive a non-JSON response from any endpoints besides the core download ones.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -414,7 +414,7 @@ class GirderClient(object):
             return getattr(requests, method.lower())
 
     def sendRestRequest(self, method, path, parameters=None,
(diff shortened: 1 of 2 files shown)


manthey:
It would be handy in this context to have the response Content-Type available.

zachmullen:
Would it be better to just `return response` and let downstreams deal with it?

danlamanna:
It seems like it would be more useful to allow returning of the raw response object instead? That way the caller can do whatever they want with it (such as view the response headers like @manthey mentioned).

manthey:
I think so.

It would be nice to explicitly mention using `.content` on the return values to get the body..

Issue #1912, opened 2017-04-07
Support poll option for 'girder-install web --watch'


msmolens:
When developing using Vagrant `girder-install web --watch` doesn't recognize changes to files made outside the VM. Supporting the `poll` webpack watch option should allow this workflow. 
https://webpack.js.org/configuration/watch/#watchoptions-poll

ronichoudhury:
Just as an FYI, I recently ran into this problem for plugin development, and added the option to my plugin's webpack.helper.js. It worked nicely so I think we should mention this in the plugin dev docs, and also add it as an option to the core build.

brianhelba:
Fixed by #2214 ! You'll have to `vagrant provision` again, once that code is merged in.

PR #1948, opened 2017-04-24
FIX automatically rebuild for --watch

MayeulChassagnard:
Working on a plugin, I noticed that doing:
```
girder-install web --watch-plugin your_plugin_name
```
wasn't working on my side.

I had to add this commit to make it work.

Note: Now weirdly, I do not need this line anymore to make it work... But anyway, I still open a PR in case it may help people.


zachmullen:
We should make this optional. Generally, system file watches are preferable to polling since they'll use a lot less resources. We could base its value on an environment variable, and then expose `girder-install web --watch --poll`.

msmolens:
Note that there is already an issue about implementing an option for polling: #1912.

zachmullen:
Closing, we can re-open a new PR to implement `--poll` option.

Issue #1649, opened 2016-10-21
Item preview plugin doesn't work properly for items with multiple files


manthey:
If I have an item named `foo.jpg`, and that item contains two files, `foo.jpg` and `bar.other`, the item preview plugin will try to render the item as an image, but uses the download endpoint to access it.

It should either properly detect that this is not a simple image, or should always try to handle the first file in an item (if such a file exists).

hwrod:
Thanks @zachmullen! Unfortunately, I won't be able to get to this until February 2017, but if this ticket is open then, I will take a look. Feel free to reassign, or wait until I have availability and I will be glad to take a look. Thanks!

zachmullen:
After discussing with @MayeulChassagnard , I think the best path forward is to introduce a `nFiles` field on each item that tracks/caches the number of files within it. This can be used to quickly determine the number of files in an item without needing to do an extra request for the file list.

That's a significant core change and also involves a migration path to compute this value on legacy items.

hwrod:
@zachmullen I was working a bit on this, though I see @MayeulChassagnard has a good solution as well in PR #1949 . Glad to see it's working out. I had a simple solution of replacing the image src attribute directly (below) on `ItemPreviewWidget` render, but I like his collection update method better. 

Implementing nFiles is beyond anything I can contribute right now, but at least this strategy can preview multiple-file items as images, only requesting a file list when hinted by file extension.

`ItemPreviewWidget.render:`

        ...
        supportedItems.filter(this._isImageItem).forEach(function(item) { 
            restRequest({
              path: `item/${id}/files`,
            })
            .done(function(files) {
              if (files.length > 1) {
                var image = files.find(file => file.mimeType.match(/image/))
                image && view.$(`img[data-id="${item.id}"]`)
                  .attr('src', `/api/v1/item/${image._id}/download`)
              }
        })

manthey:
This was fixed some time ago.

PR #1949, opened 2017-04-24
BUG: Don't display items containing multiple files in item_preview plugin

MayeulChassagnard:
Since an item containing multiple files is going to be download as a zip file,
and the item_preview do not support a display inline of a zip file, it produces
a display error.

Then, one solution is to check all items and remove the ones containing
multiple files from `supportedItems`.

Cons:
* Do not preview:
  - supported item containing multiple supported files
  - item named `example` containing a supported file (indeed, for now,
    an item is defined as supported if its name contain .png at the end
    for instance)
* Error still shown:
  - if an item considered as supported (with .png at the end of its name for
    instance) and the file inside is not supported (a .txt file for example)

### DIFF
--- a/plugins/item_previews/web_client/views/ItemPreviewWidget.js
+++ b/plugins/item_previews/web_client/views/ItemPreviewWidget.js
@@ -28,10 +28,11 @@ var ItemPreviewWidget = View.extend({
 
     initialize: function (settings) {
         this.collection = new ItemCollection();
+        this.itemCollection = {};
 
         this.collection.on('g:changed', function () {
             this.trigger('g:changed');
-            this.render();
+            this.loadFileList();
         }, this).fetch({
             folderId: settings.folderId
         });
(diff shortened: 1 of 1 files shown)


MayeulChassagnard:
This PR fix the issue #1649

jcfr:
Thanks for the PR.

Before we move forward with the review process, consider fixing the failing test. 

Assuming girder sources are in `~/Projects/girder`, you can do the following to run the tests:

```bash
$ mkdir ~/Projects/girder-build
$ cd ~/Projects/girder-build
$ workon girder_env # This assumes you have a virtual env named girder_env
$ cmake ../girder
$ ctest -R item_previews
Test project /home/jcfr/Projects/girder-build
    Start 128: eslint_item_previews
1/2 Test #128: eslint_item_previews .............   Passed    0.59 sec
    Start 129: puglint_item_previews
2/2 Test #129: puglint_item_previews ............   Passed    0.13 sec
```

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -53,9 +54,38 @@ var ItemPreviewWidget = View.extend({
         return /\.(jpg|jpeg|png|gif)$/i.test(item.get('name'));
     },
 
+    _isSingleFileInItem: function (item) {
```
jcfr: `_isSingleFileItem`

That said this is more explicit: `_isMultifileFileItem`

MayeulChassagnard: I used this name previously indeed but what if there is no files in an item? We will have to return the same answer that if there was a single item, which could be annoying I think

jcfr: Makes sense. Then, I would suggest a function name like `_isSingleFileItem`

zachmullen:
Let's reimplement this once we cache the `nFiles` value on items, which will save us from doing O(n) XHRs on the list view.

PR #1953, opened 2017-04-25
Add version option for girder client

### LINKED ISSUE #1936: Add python client option to return the version of the python client
like a -v or --version CLI option.

/cc @bpaniagua

dorukozturk:
Fixes #1936

### DIFF
--- a/clients/python/girder_client/cli.py
+++ b/clients/python/girder_client/cli.py
@@ -20,7 +20,7 @@
 import click
 import sys
 import types
-from girder_client import GirderClient, HttpError
+from girder_client import GirderClient, HttpError, __version__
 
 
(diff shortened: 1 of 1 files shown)


clients/python/girder_client/cli.py
```diff
@@ -146,6 +146,7 @@ def format_options(self, ctx, formatter):
                    '[default: %s]' % GirderClient.DEFAULT_API_ROOT,
               show_default=True,
               cls=_AdvancedOption)
+@click.version_option()
```
zachmullen: We should probably pass the version explicitly to avoid the auto-discovery logic.

```
from girder_client import __version__

@click.version_option(version=__version__, prog_name='Girder command line interface')
```

PR #1956, opened 2017-04-25
BUG: wrapPreviews didn't hide items in dicom_preview plugin

MayeulChassagnard:
Previously, clicking on the arrow icon on the right of the Item Previews
didn't hide items.

There is maybe a better solution.

Maybe there is something to change [here](https://github.com/MayeulChassagnard/girder/blob/8209a602d0bbcf80f6573d6efb21ca974d672065/plugins/item_previews/web_client/views/ItemPreviewWidget.js#L18-L26) (upper in the file) instead?

### DIFF
--- a/plugins/item_previews/web_client/views/ItemPreviewWidget.js
+++ b/plugins/item_previews/web_client/views/ItemPreviewWidget.js
@@ -54,7 +54,12 @@ var ItemPreviewWidget = View.extend({
     },
 
     render: function () {
-        var supportedItems = this.collection.filter(this._isSupportedItem.bind(this));
+        var supportedItems = {};
+
+        // Do not display items if this.wrapPreviews = True
+        if (!this.wrapPreviews) {
+            supportedItems = this.collection.filter(this._isSupportedItem.bind(this));
+        }
 
         var view = this;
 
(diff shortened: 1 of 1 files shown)


plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -56,6 +56,11 @@ var ItemPreviewWidget = View.extend({
     render: function () {
         var supportedItems = this.collection.filter(this._isSupportedItem.bind(this));
 
+        // Do not display items if this.wrapPreviews = True
+        if(this.wrapPreviews) {
```
zachmullen: eslint is failing, it wants a space after `if`.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -56,6 +56,11 @@ var ItemPreviewWidget = View.extend({
     render: function () {
         var supportedItems = this.collection.filter(this._isSupportedItem.bind(this));
```
zachmullen: We should probably only do this logic if `!this.wrapPreviews` to avoid doing computation.

MayeulChassagnard: Ok, I'll give a try! Thanks for the review

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -66,27 +71,29 @@ var ItemPreviewWidget = View.extend({
...
-            // Ajax request the JSON files to display them.
-            restRequest({
-                path: 'item/' + id + '/download',
-                type: 'GET',
-                error: null // don't do default error behavior (validation may fail)
-            }).done(function (resp) {
-                view.$('.json[data-id="' + id + '"]').text(JSON.stringify(resp, null, '\t'));
-            }).error(function (err) {
-                console.error('Could not preview item', err);
+        if (!this.wrapPreviews) {
```
zachmullen: Let's remove this conditional. If `this.wrapPreviews`, the block below would be a no-op anyway.

MayeulChassagnard: Removing this leads to an error since `supportedItems` is not a function and we call `supportedItems.filter' ligne 76

MayeulChassagnard:
Closing this since I misunderstood the behaviour of this button (I thought it was hiding the preview content instead of toggling display from line to column).

The icon may have confused me, would it make sense to modify this icon by icon-level-down from `/girder/clients/web/static/built/fontello/demo.html`?

PR #1960, opened 2017-04-27
Doc intent for python client versioning

### LINKED ISSUE #1959: Sync Python client version with server version
I realize this might be tricky to pull off (dealing with semantic API versions of both), and controversial, but it is generally much simpler for developers and users if the versions are synced.

mgrauer:
Fixes #1959.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -396,7 +396,7 @@ Releasing the python client package
 The design intent behind the python client package is to work with as many
 versions of the Girder server as possible; think carefully before breaking this
 compatibility. Whenever the main Girder package is released, the python client
-package should also be versioned and released if is has changed since the last
+package should also be versioned and released if it has changed since the last
 Girder release or the last time it was released. Normal semantic versioning is
 not in use for the python client package because its version is partially
 dependent on the Girder server package version. The rules for versioning the
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -395,12 +395,13 @@ Releasing the python client package
 
 The design intent behind the python client package is to work with as many
 versions of the Girder server as possible; think carefully before breaking this
-compatibility. Whenever the main Girder package is released, the python client
-package should also be versioned and released if it has changed since the last
-Girder release or the last time it was released. Normal semantic versioning is
-not in use for the python client package because its version is partially
-dependent on the Girder server package version. The rules for versioning the
-python client package are as follows:
+compatibility. There isn't a formal rule for releasing versions of the python
+client package, releases tend to be made when a significant change is made to
+the client that people want to use in production.
+
+Normal semantic versioning is not in use for the python client package because
+its version is partially dependent on the Girder server package version. The
+rules for versioning the python client package are as follows:
 
 * The major version of the python client should be the same as the major version
   of the Girder server package, assuming it is compatible with the server API.
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -388,14 +388,19 @@ recommended process for generating a new release is described here.
 
         python setup.py sdist upload
 
+.. _releasepythonclientpackage:
+
 Releasing the python client package
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Whenever the main Girder package is released, the python client package should also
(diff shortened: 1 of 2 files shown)


docs/python-client.rst
```diff
@@ -11,6 +11,16 @@ much easier to upload a large, nested hierarchy of data from a local directory
 to Girder, and also makes it much easier to download a large, nested hierarchy
 of data from Girder to a local directory.
 
+Versioning
+----------
+
+The Girder python client is semanticly versioned separately from the Girder
```
zachmullen: semantically*

docs/python-client.rst
```diff
@@ -11,6 +11,16 @@ much easier to upload a large, nested hierarchy of data from a local directory
...
+server, but not fully independently because the client and server are coupled.
+The design intent is that every version of the client should work with every
+version of the server, modulo bugfixes, though this is not an explicit
```
zachmullen: "modulo bugfixes" may not be a great phrase to use for the audience. I think we can remove the phrase here, since the line "though this is not an explicit guarantee" kind of captures that sentiment anyway.

mgrauer:
I didn't know about that section, I can point to it for more details but that is beyond the level I expect someone who just wants to use the python client will want to see.  I'll revise accordingly.

After reading the details on the python client versioning, I think my description of it being 'semanticly' versioned was correct! 😸

mgrauer:
This is RFR.

docs/development.rst
```diff
@@ -388,14 +388,19 @@ recommended process for generating a new release is described here.
...
 
-Whenever the main Girder package is released, the python client package should also
-be versioned and released if is has changed since the last Girder release or the last
-time it was released. Normal semantic versioning is not in use for the python client
-package because its version is partially dependent on the Girder server package
-version. The rules for versioning the python client package are as follows:
+The design intent behind the python client package is to work with as many
+versions of the Girder server as possible; think carefully before breaking this
+compatibility. Whenever the main Girder package is released, the python client
+package should also be versioned and released if is has changed since the last
```
msmolens: if *it

mgrauer: This was actually a reference to the famous quote `It depends upon what the meaning of the word 'is' is`, so when 'is' has changed...

Naw, good catch.

docs/development.rst
```diff
@@ -388,14 +388,19 @@ recommended process for generating a new release is described here.
...
 
-Whenever the main Girder package is released, the python client package should also
-be versioned and released if is has changed since the last Girder release or the last
-time it was released. Normal semantic versioning is not in use for the python client
-package because its version is partially dependent on the Girder server package
-version. The rules for versioning the python client package are as follows:
+The design intent behind the python client package is to work with as many
+versions of the Girder server as possible; think carefully before breaking this
+compatibility. Whenever the main Girder package is released, the python client
+package should also be versioned and released if it has changed since the last
```
zachmullen: We actually don't do this in practice, and given the independence of the two libraries, maybe we shouldn't prescribe this policy at all, unless you think there's a good reason to do so.

mgrauer:
For clarity, I didn't change that block in development.rst other than to add the initial sentence.  It looks like I changed the whole block because I updated the spacing.

That is to say, I didn't write most of what is in the diff in that file, and I haven't been paying attention in detail to how the python client is released, so it isn't like I added that in today based on some perception of how we do things or should do things.  I.e., I'm not prescribing what is written there because I didn't write it and I don't know what we do! 

🍹 

We *should* clarify our policy to ourselves, and then write that here.

Since the design intent of the client is to be maximally compatible with different server versions, and that is now captured here in the docs (in two places), and we don't want to link the versions of client and server (beyond the major version, mentioned in how to come up with the client version number, and which I think should stay) I don't think we need to version the client when we version the server.  This is also supported by the fact that we don't actually do it.

We seem to version the client by roughly the same rules as the server, something like whenever a big enough change occurs or when people ask for it.  That might be a bit vague to put into docs, but it does approximate reality, and half a loaf is better than none.

🥖

mgrauer:
Thanks to you both for reading this so carefully.

PR #1962, opened 2017-04-27
Circleci display test summary

jcfr:
This PR updates `run_circleci.sh` so that it convert CTest results into a format usable by the CircleCI test summary.

It makes use of [ctest-junit-formatter](http://scikit-ci-addons.readthedocs.io/en/latest/addons.html#ctest-junit-formatter) provided by `scikit-ci-addons`.

Here is an example of test summary:

ctest-within-circleci

### DIFF
--- a/cmake/run_circleci.sh
+++ b/cmake/run_circleci.sh
@@ -21,6 +21,11 @@ esac
 mkdir $HOME/build
 touch $HOME/build/test_failed
 ctest -VV -S $HOME/girder/cmake/circle_continuous.cmake
+
+# Convert CTest output to Junit and ensure CircleCI will include it in its summary
+pip install scikit-ci-addons==0.15.0
+mkdir ${CIRCLE_TEST_REPORTS}/CTest
+ci_addons ctest_junit_formatter $HOME/build > ${CIRCLE_TEST_REPORTS}/CTest/JUnit-${CIRCLE_NODE_INDEX}.xml || exit 1
 if [ -f $HOME/build/test_failed ] ; then
 	exit 1
 fi


jcfr:
Cc: @mgrauer

mgrauer:
Magnifique!  Is there a reason you prefer adding `set -e` to `#!/bin/bash -e`?

zachmullen:
> Is there a reason you prefer adding `set -e` to `#!/bin/bash -e`?

The second option wouldn't work if the script was invoked as `bash run_circleci.sh`, right?

tests/packaging/pip_install_girder.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
```
zachmullen: This breaks some behavior in the script. If you look below, we explicitly test return values and print human readable error messages.

tests/packaging/test_plugin_install.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
```
zachmullen: same here

jcfr:
I reverted the bash script changes and will push them in a different PR

zachmullen:
In the success case, there is not a test summary shown in CircleCI, is that expected?

cmake/run_circleci.sh
```diff
@@ -21,6 +21,12 @@ esac
 mkdir $HOME/build
 touch $HOME/build/test_failed
 ctest -VV -S $HOME/girder/cmake/circle_continuous.cmake
+
+# Convert CTest output to Junit and ensure CircleCI will include it in its summary
+pip install scikit-ci-addons==0.15.0
+mkdir ${CIRCLE_TEST_REPORTS}/CTest
+ci_addons ctest_junit_formatter $(pwd) > ${CIRCLE_TEST_REPORTS}/CTest/JUnit-${CIRCLE_NODE_INDEX}.xml
```
zachmullen: ` || exit 1` ?

jcfr:
Et voila :)

PR #1963, opened 2017-04-27
Add Vagrantfile/ansible for item_tasks fullstack

mgrauer:
I think the item_tasks plugin is great and want to enable more people to use it.  The goal here is to make it easier to play with item_tasks and focus on building a Docker image, without having to know as much about Girder/Girder-Worker.

This relies on the heavy lifting done by those who built the girder ansible client, the girder ansible role, and the worker ansible role.


mgrauer:
How to test this, each of these steps should work.

`vagrant up` and ensure provisioning happens.
Go to http://localhost:9080 on your host.
Log in as girder/girder.
Go into a user folder, say private.
From the folder actions menu, add tasks.  Add `kitware/pysciencedock`, which will start a worker job to ingest the Docker image.
It may take a little while as the Docker image is built.
You should have the Docker image ingested by a worker job, resulting in several new tasks (see them via the tasks menu at the left, they are like "normalize", "t-test", and there are more than 10).
Upload the [csv file](https://raw.githubusercontent.com/Kitware/pysciencedock/master/tests/study.csv) to a location in your Girder instance.
Run a normalize task (go to tasks menu, click on normalize), pointing to that csv file stored in Girder as input.  Select sum/none/none.  Choose an output path in Girder.
Ensure that the job runs correctly and that you get an output that looks like
```
a,b
1,1
1,2
```
I'm not sure if this is actually correct for normalize, but it will test that the plumbing of item_tasks is correct, which is what this PR is about.

kotfic:
@mgrauer I was able to run the following test without a problem.  I'm wondering if there is a way to set up some kind of integration test here though?  It would be nice to have a nightly or something that could let us know if this stopped working.

mgrauer:
I agree @kotfic .  Is that something we could add to the set of nightly dashboards for testing ansible on various platforms @danlamanna ?  Would Packer be able to help us here in some way using CI tools?

danlamanna:
We could add nightly tests for this.

Also @mgrauer, I created a branch off of this using a very popular upstream Docker role.. take a look and tell me what you think: https://github.com/girder/girder/compare/item_tasks_fullstack_vagrant...use-docker-role-for-item-tasks-vagrant

mgrauer:
@danlamanna we get to net remove many lines of code, not support that Docker nonsense ourselves, and pull in a repo named `angstwad`.   What's not to like?

plugins/item_tasks/devops/README
```diff
@@ -0,0 +1,38 @@
+The goal of this Vagrantfile and Ansible setup is to provide a fully
```
danlamanna: This should be in markdown/rst so GitHub can render it.

plugins/item_tasks/devops/README
```diff
@@ -0,0 +1,38 @@
...
+
+`vagrant up`
+
+## ssh into the box as vagrant/vagrant, with passwordless sudo
+
+`vagrant ssh`
+
+## Halt the box
+
+`vagrant halt`
```
danlamanna: I don't know how much of this is necessary, this is basic vagrant usage - I suppose it's a matter of the target audience.

danlamanna:
`pre-tasks/docker.yml` needs to be deleted now.

plugins/item_tasks/devops/README
mgrauer: Yeah, I'm not sure how much to keep in, I want to add a bit in case people aren't familiar.

I'll leave it in unless someone really squawks.

🐦

mgrauer:
Got to remove even more LOC because girder.worker supplied a new ability.  This is RFR.

mgrauer:
There is a problem that the last commit fixes, but probably a better solution is available.

The issue is that the worker user (when the girder_worker service was started) wouldn't be in the Docker group after provisioning, so e.g. importing a task would fail.  The last commit restarts the girder_worker service after adding the worker user to the Docker group, so that user has a chance to pick up the additional group.

What I don't like about this is that it directly copies a restart out of the girder.worker role, which doesn't seem like an approach that reuses existing work.  I also call `when: girder_worker_start`, and I don't need to.  Finally it is a repeat of the girder_worker service restart, which is slightly inefficient.

We do a couple tasks as post_tasks (creating the docker group and restarting the docker daemon) that happen after all the roles.  Is there a way to have those tasks be triggered either at the end of the docker role or before the girder.worker role?  Another option is to have the girder_worker service restart be triggered as a handler, but @danlamanna mentioned that was a pain.

mgrauer:
Maybe just a local role inserted in between docker and girder_worker where I move the two post tasks?  Anything wrong with that approach?

manthey:
I've had this as an issue in the HistomicsTK deployment of multiple docker containers.  The girder worker container can't be assigned to right group until it is started as an image, as the image may be started on a different computer with a different docker gid than on the machine where the image was built.

Rather than restart, is it possible for the start up to join the correct group and then start the worker in a subshell that would be a member of the group?

mgrauer:
>  is it possible for the start up to join the correct group

Do you mean the start up service script of the girder_worker?  In that case, I think I would need some ability to pass a groupname or list into the service script (or other part) of the girder.worker role.

You did give me another idea (which might be cheating), that is to create a docker group and add the worker user to that group as pre_tasks.

kotfic:
@manthey This deployment script doesn't look like it is intended to be used in any context other than deploying to a single vagrant machine for the purpose of demo'ing the item_task full stack of technologies.

@mgrauer it looks like you can use the [docker_group_members](https://github.com/angstwad/docker.ubuntu/blob/master/tasks/main.yml#L282) variable on the angstwad role to add the worker user to the docker group.  Not sure if this will help (do you need a docker restart before you run into trouble?) Otherwise that will insure the worker user is apart of the docker group before running the girder.worker role.

mgrauer:
@kotfic , that worked, thanks for the pointer.  I'm still not in the mode of looking for the external roles to do things for me, I need to learn to think like that.

You are correct about the purpose of this Vagrantfile/ansible playbook, it intended for a single VM to make it easy for someone who wants to develop `item_tasks` Dockerfiles to get a working Girder stack without having to deal with setting up the stack.

This is RFR.

PR #1964, opened 2017-04-27
Add more arguments to girder's add_test ctest wrappers

jbeezley:
Adds `ENVIRONMENT` to both javascript and python tests and `EXTERNAL_DATA` to javascript tests.


tests/JavascriptTests.cmake
```diff
@@ -145,6 +146,15 @@ function(add_web_client_test case specFile)
     set(test_module "tests.web_client_test")
   endif()
 
+  if(fn_EXTERNAL_DATA)
```
zachmullen: Do we need to list EXTERNAL_DATA in the `_args`?

jbeezley: Yes, added.

PR #1967, opened 2017-04-29
Secure cookie setting

zachmullen:


jbeezley:
When enabling the setting, existing non-secure cookies are still valid.  Should we delete all session tokens when enabling this feature?

zachmullen:
I figure the vast majority of deployments won't explicitly set this setting, but will use the default. For the scope of this PR, I'd say we can leave the decision to blow away tokens up to individual site admins, and any features related to that can be implemented on future PRs.

jbeezley:
Sounds good to me.

PR #1968, opened 2017-05-01
Pass through a data option to the model fetch.

manthey:
Document the fetch options.

This adds unit tests for some of the base Model class.


clients/web/test/spec/modelSpec.js
```diff
@@ -0,0 +1,113 @@
...
+            } else {
+                resp.resolve({});
+            }
+            // The jqXHR response of a rest request needs an error function
+            resp.error = resp.fail;
+            return resp;
+        });
+
+        windowAlert = window.alert;
+        window.alert = function (msg) {
```
zachmullen: Why do we need to mock alert here?

manthey: It is just a way to get a counter to see if we called it.

zachmullen: IIRC, calling `alert` in phantom (at least in older versions) would hang the test indefinitely :)

zachmullen: Hence, maybe we should remove the `alert` from our model code in favor of a `console.error`.

manthey: Yes.  So in the tests, we normally just mock `window.alert` and have it call `console.log` to prevent that.

manthey: We use alert in a number of places: Collection, AccessControlledModel, GroupView, and Model.  All of them are really for a failure to set up a model properly (thus to inform the developer and not the user).  Converting them to `console.error` seems fine to me.  Would you like that on this or a different PR?

manthey: The alert in the GroupView has a note to `don't alert, show something useful`.

zachmullen: Yeah we can do the alert squashing on a separate PR, and at the same time add something in the test env to mock it out and cause it to fail the test if it's called.

PR #1969, opened 2017-05-01
Fix log errors from item_tasks plugin

zachmullen:
Any time a file was uploaded that did not contain a reference
field, the item_tasks plugin was throwing an exception. This
was not visible to end users since it's an async task, but was
appearing in the admin logs.


zachmullen:
Not sure why codecov report is failing, but this is ready for review.

jbeezley:
It's because the exception condition is not tested.  I don't think it's necessary here though.

PR #1970, opened 2017-05-01
Compute all missing file hashes when files are uploaded

zachmullen:
@jamiesnape @jcfr @mgrauer the first commit here demonstrates the basic functionality of async computing hashes on files at upload time. TODO:

- [x] Admin endpoint to compute missing hashes on existing files
- [x] Setting to control this behavior (off by default)
- [x] Tests

### DIFF
--- a/plugins/hashsum_download/plugin_tests/hashsum_download_test.py
+++ b/plugins/hashsum_download/plugin_tests/hashsum_download_test.py
@@ -288,6 +288,11 @@ class HashsumDownloadTest(base.TestCase):
         self.assertIn('sha256', file)
         self.assertEqual(file['sha256'], expected.hexdigest())
 
+        expected = hashlib.sha512()
+        expected.update(self.userData)
+        self.assertIn('sha512', file)
+        self.assertEqual(file['sha512'], expected.hexdigest())
+
         hashsum_download.SUPPORTED_ALGORITHMS = old
 
     def testManualComputeHashes(self):
--- a/plugins/hashsum_download/plugin_tests/hashsum_download_test.py
+++ b/plugins/hashsum_download/plugin_tests/hashsum_download_test.py
@@ -17,9 +17,11 @@
 #  limitations under the License.
 ###############################################################################
 
-import six
 import hashlib
+import six
+import time
 
+from girder.models.model_base import ValidationException
 from tests import base
 
 
@@ -260,3 +262,60 @@ class HashsumDownloadTest(base.TestCase):
         resp = self.request(template % (self.privateFile['_id'], 'sha512'))
         self.assertStatus(resp, 401)
         six.assertRegex(self, resp.json['message'], '^Read access denied')
+
+    def testAutoComputeHashes(self):
+        from girder.plugins import hashsum_download
+        with self.assertRaises(ValidationException):
+            self.model('setting').set(hashsum_download.PluginSettings.AUTO_COMPUTE, 'bad')
+
+        old = hashsum_download.SUPPORTED_ALGORITHMS
+        hashsum_download.SUPPORTED_ALGORITHMS = {'sha512', 'sha256'}
+        self.model('setting').set(hashsum_download.PluginSettings.AUTO_COMPUTE, True)
+
+        file = self.model('upload').uploadFromFile(
+            obj=six.BytesIO(self.userData), size=len(self.userData), name='Another file',
+            parentType='folder', parent=self.privateFolder, user=self.user)
+
+        start = time.time()
+        while time.time() < start + 15:
+            file = self.model('file').load(file['_id'], force=True)
+            if 'sha256' in file:
+                break
+            time.sleep(0.2)
+
+        expected = hashlib.sha256()
+        expected.update(self.userData)
+        self.assertIn('sha256', file)
+        self.assertEqual(file['sha256'], expected.hexdigest())
+
+        hashsum_download.SUPPORTED_ALGORITHMS = old
+
+    def testManualComputeHashes(self):
+        from girder.plugins import hashsum_download
+        self.model('setting').set(hashsum_download.PluginSettings.AUTO_COMPUTE, False)
+        old = hashsum_download.SUPPORTED_ALGORITHMS
+        hashsum_download.SUPPORTED_ALGORITHMS = {'sha512', 'sha256'}
+
+        self.assertNotIn('sha256', self.privateFile)
+
+        expected = hashlib.sha256()
+        expected.update(self.userData)
+
+        # Running the compute endpoint should only compute the missing ones
+        resp = self.request(
+            '/file/%s/hashsum' % self.privateFile['_id'], method='POST', user=self.user)
+        self.assertStatusOk(resp)
+        self.assertEqual(resp.json, {
+            'sha256': expected.hexdigest()
+        })
+
+        # Running again should be a no-op
+        resp = self.request(
+            '/file/%s/hashsum' % self.privateFile['_id'], method='POST', user=self.user)
+        self.assertStatusOk(resp)
+        self.assertEqual(resp.json, None)
+
+        file = self.model('file').load(self.privateFile['_id'], force=True)
+        self.assertEqual(file['sha256'], expected.hexdigest())
+
+        hashsum_download.SUPPORTED_ALGORITHMS = old
--- a/plugins/hashsum_download/server/__init__.py
+++ b/plugins/hashsum_download/server/__init__.py
@@ -33,6 +33,7 @@ from girder.utility.model_importer import ModelImporter
 from girder.utility.progress import ProgressContext, noProgress
 
 SUPPORTED_ALGORITHMS = {'sha512'}
+_CHUNK_LEN = 65536
 
 
 class PluginSettings(object):
@@ -150,15 +151,23 @@ class HashedFile(File):
 
 def _computeHashHook(event):
     """
-    Computes all supported checksums on a given file. Downloads the
-    file data and stream-computes all required hashes on it, saving
-    the results in the file document.
+    Event hook that computes the file hashes in the background after
+    a completed upload. Only done if the AUTO_COMPUTE setting enabled.
     """
     if ModelImporter.model('setting').get(PluginSettings.AUTO_COMPUTE, default=False):
         _computeHash(event.info['file'])
 
 
 def _computeHash(file, progress=noProgress):
+    """
+    Computes all supported checksums on a given file. Downloads the
+    file data and stream-computes all required hashes on it, saving
+    the results in the file document.
+
+    In the case of assetstore impls that already compute the sha512,
+    and when sha512 is the only supported algorithm, we will not download
+    the file to the server.
+    """
     toCompute = SUPPORTED_ALGORITHMS - set(file)
     toCompute = {alg: getattr(hashlib, alg)() for alg in toCompute}
 
@@ -168,7 +177,7 @@ def _computeHash(file, progress=noProgress):
     fileModel = ModelImporter.model('file')
     with fileModel.open(file) as fh:
         while True:
-            chunk = fh.read(65536)
+            chunk = fh.read(_CHUNK_LEN)
             if not chunk:
                 break
             for digest in six.viewvalues(toCompute):
--- a/plugins/hashsum_download/server/__init__.py
+++ b/plugins/hashsum_download/server/__init__.py
@@ -17,6 +17,10 @@
 #  limitations under the License.
 ###############################################################################
 
+import hashlib
+import six
+
+from girder import events
 from girder.api import access
 from girder.api.describe import autoDescribeRoute, Description
 from girder.api.rest import RestException, setRawResponse, setResponseHeader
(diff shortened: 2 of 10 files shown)


plugins/hashsum_download/server/__init__.py
```diff
@@ -17,19 +17,24 @@
...
+SUPPORTED_ALGORITHMS = {'sha256'}
+
 
 class HashedFile(File):
 
-    supportedAlgorithms = {
-        'sha512'
-    }
+    # deprecated, use module-level constant instead
+    supportedAlgorithms = SUPPORTED_ALGORITHMS
```
jcfr: Generally speaking, when attribute are deprecated, I wonder if we should add property with the name `supportedAlgorithms`  and report a deprecation warning ...

plugins/hashsum_download/server/__init__.py
```diff
@@ -118,7 +123,37 @@ def _getFirstFileByHash(self, algo, hash, user=None):
...
+    }, multi=False)
+
+
 def load(info):
     HashedFile(info['apiRoot'].file)
     ModelImporter.model('file').exposeFields(
-        level=AccessType.READ, fields=HashedFile.supportedAlgorithms)
+        level=AccessType.READ, fields=SUPPORTED_ALGORITHMS)
+
+    events.bind('data.process', info['name'], _computeHash)
```
jcfr: Should the [handler name](http://girder.readthedocs.io/en/latest/api-docs.html#girder.events.bind) be prefixed by the name of the plugin ?

zachmullen: Yes, it is here.

jcfr: Nevermind, `info['name']` is the name of the plugin already.

jcfr: Moving forward, it would be nice to document what data are expected with which event ...

plugins/hashsum_download/server/__init__.py
```diff
@@ -118,7 +123,37 @@ def _getFirstFileByHash(self, algo, hash, user=None):
...
 
+def _computeHash(event):
+    """
+    Computes all supported checksums on a given file. Downloads the
+    file data and stream-computes all required hashes on it, saving
+    the results in the file document.
+    """
+    file = event.info['file']
+
+    toCompute = SUPPORTED_ALGORITHMS - set(file)
```
jcfr: Not sure what the line above is doing.

zachmullen: set difference of the supported algorithms and the keys in the file document

plugins/hashsum_download/server/__init__.py
```diff
@@ -118,7 +129,46 @@ def _getFirstFileByHash(self, algo, hash, user=None):
...
+    if not toCompute:
+        return
+
+    fileModel = ModelImporter.model('file')
+    with fileModel.open(file) as fh:
+        while True:
+            chunk = fh.read(65536)
+            if not chunk:
+                break
+            for hash in six.viewvalues(toCompute):
```
mgrauer: nit: can you use a different name here for `hash`, because elsewhere in this module `hash` is a var name for the actual hash digest value.  maybe `hasher` or `hashComputation` or `hashCalculation`, `ongoing`, `digester`?

plugins/hashsum_download/server/__init__.py
```diff
@@ -118,7 +129,46 @@ def _getFirstFileByHash(self, algo, hash, user=None):
...
+                hash.update(chunk)
+
+    fileModel.update({'_id': file['_id']}, update={
+        '$set': {alg: hash.hexdigest() for alg, hash in six.viewitems(toCompute)}
```
mgrauer: same here

plugins/hashsum_download/server/__init__.py
jcfr: Thanks for the clarification.

I see the `file` you use is from https://github.com/girder/girder/blob/41b3e3be68df34e55fbc88676bb81540d1cbfc43/girder/models/upload.py#L228-L241

And it contains keys like these ones ... 

```
{
    "_id" : ObjectId("5906bfba4154cd7dfbe47cfe"),
    "relpath" : "/trigirderdemo-assetstore/77/47/7747937d98b34aaab0a54e96395ddfc1",
    "mimeType" : "application/octet-stream",
    "assetstoreId" : ObjectId("590613464154cd7dfbe47cf7"),
    "size" : 8794800,
    "creatorId" : ObjectId("574305a44154cd4e67c1d835"),
    "dicom" : {},
    "exts" : [ 
        "obj"
    ],
    "s3Key" : "77/47/7747937d98b34aaab0a54e96395ddfc1",
    "itemId" : ObjectId("5906bf884154cd7dfbe47cfc"),
    "created" : ISODate("2017-05-01T04:55:22.129Z"),
    "name" : "baby_dragon.obj"
}
```

plugins/hashsum_download/server/__init__.py
```diff
@@ -39,9 +40,12 @@ class PluginSettings(object):
...
 class HashedFile(File):
-
-    # deprecated, use module-level constant instead
-    supportedAlgorithms = SUPPORTED_ALGORITHMS
+    @property
+    def supportedAlgorithms(self):  # pragma: no cover
+        warnings.warn(
+            'HashedFile.supportedAlgorithms is deprecated, use the module-level '
+            'SUPPORTED_ALGORITHMS instead.', DeprecationWarning)
+        return SUPPORTED_ALGORITHMS
```
jcfr: 👍

zachmullen:
This is ready for final review

plugins/hashsum_download/web_client/templates/config.pug
```diff
@@ -0,0 +1,10 @@
+.g-config-breadcrumb-container
+form#g-hashsum-download-config-form(role="form")
+  .checkbox
+    label.control-label(for="hashsum_download_auto_compute")
+      input#hashsum_download_auto_compute(type="checkbox",
```
msmolens: Nit: Usually IDs use dash separators, not underscores.

plugins/hashsum_download/server/__init__.py
```diff
@@ -118,7 +148,50 @@ def _getFirstFileByHash(self, algo, hash, user=None):
...
+    toCompute = SUPPORTED_ALGORITHMS - set(file)
+    toCompute = {alg: getattr(hashlib, alg)() for alg in toCompute}
+
+    if not toCompute:
+        return
+
+    fileModel = ModelImporter.model('file')
+    with fileModel.open(file) as fh:
+        while True:
+            chunk = fh.read(65536)
```
mgrauer: You should probably move this to the top as a module level constant.

Maybe we should centralize this value, as we have it in a few places.  Is it well determined, or semi-arbitrary, both?

As [BUF_LEN](https://github.com/girder/girder/blob/70902befc45b165b6d5e9a34828c15ad6281814e/girder/utility/s3_assetstore_adapter.py#L33), [BUF_SIZE](https://github.com/girder/girder/blob/70902befc45b165b6d5e9a34828c15ad6281814e/girder/utility/filesystem_assetstore_adapter.py#L36), and [MAX_CHUNK_SIZE](https://github.com/girder/girder/blob/70902befc45b165b6d5e9a34828c15ad6281814e/clients/python/girder_client/__init__.py#L158).  Actually, it's probably better to keep them independent, and this isn't terribly important anyway.

plugins/hashsum_download/plugin_tests/hashsum_download_test.py
```diff
@@ -260,3 +262,60 @@ def testKeyFile(self):
...
+            parentType='folder', parent=self.privateFolder, user=self.user)
+
+        start = time.time()
+        while time.time() < start + 15:
+            file = self.model('file').load(file['_id'], force=True)
+            if 'sha256' in file:
+                break
+            time.sleep(0.2)
+
+        expected = hashlib.sha256()
```
mgrauer: can you also test that sha512 was also calculated?

zachmullen: I can add that assertion if you like; I omitted it because the test uses a filesystem assetstore, in which the sha512 is automatically computed already.

plugins/hashsum_download/server/__init__.py
zachmullen: > Maybe we should centralize this value, as we have it in a few places. Is it well determined, or semi-arbitrary, both?

semi-arbitrary. I haven't profiled running different-sized chunks through hash algs; I assume we'd be waiting on the network socket anyway, so probably not a huge deal.

zachmullen:
Your assumptions are correct, except for this one:

> A file is fully [all bytes] stream downloaded a 64M chunk at a time, to have the hashes calculated on it. File chunks going through the upload process will not have partial digests calculated on them as they are uploaded. The server therefore will see a chunk twice in the case of some assetstores such as filesystem, once as the file is uploaded, and again when the file is downloaded to calculate hashes. The downside is seeing every chunk twice. The upside is a simple design that will work for every assetstore type and client type combination.

* The file is downloaded all in the same stream, not 64MB chunks (you are probably thinking of the upload process.)
* In the case of assetstore impls that already compute the sha512, we will not download the file to the server at all. The conditional `if not toCompute` means we don't ever download a file that we already know the hashes of, in the post-upload auto hook or via the manual request.

zachmullen:
> Is it worth some UI text (and code comments) mentioning that calculating a hash on an existing file could take a while for a large file (and tie up a request thread) ?

We could add some text to that effect. It does create a progress notification so that users are aware of how long it will take *after* they have started the process, but they don't have advance warning.

plugins/hashsum_download/plugin_tests/hashsum_download_test.py
mgrauer: I would like the test to more explicitly assert the expected behavior, I think it would make the test code clearer and easier to read.  Please do add it.

mgrauer:
I wasn't being careful enough with my wording.

> * The file is downloaded all in the same stream, not 64MB chunks (you are probably thinking of the upload process.)

All in one stream, but only 64MB in memory at a time.

> * In the case of assetstore impls that already compute the sha512, we will not download the file to the server at all. The conditional if not toCompute means we don't ever download a file that we already know the hashes of, in the post-upload auto hook or via the manual request.

Right, when only using sha512 for those assetstore types that calculate it, the server only sees the bytes once, but to calculate any additional algorithms, the server will download the bytes.

I think it is worth adding something like this as a code comment.

    In the case of assetstore impls that already compute the sha512, and when sha512 is the only supported algorithm, we will not download the file to the server.

zachmullen:
> All in one stream, but only 64MB in memory at a time.

Oh, you are referring to the chunking in this new code. That's actually 64 KB :)

mgrauer:
Off by 1024X is well within my usual margin of error.

mgrauer:
The API endpoint to calculate hashes only returns the hashes calculated with that request.  It could easily return all hashes on that file (those calculated before and at the time), I'm not sure which makes more sense (return all or only those calculated by the request) and I don't have a strong opinion, but the API endpoint description should be clarified if we keep the current behavior as it implies to me it would return all hashes.

    Manually compute the checksum values for a given file.

I tried the hash calculation endpoint for both filesystem assetstore and S3, and it works in both cases for calculating all hashes (I added sha256 to the SUPPORTED_ALGORITHMS for this testing).  

The UI and logic seems very 'sha512' centric, which is probably by design, but here are a couple ways that this limits the ability to use other hashes.

If 'sha512' and other hashes are calculated, the only hash shown in the UI is 'sha512', and only that keyfile can be downloaded from the UI.  In the below screenshot 'sha256' also exists for this file.

screen shot 2017-05-03 at 7 20 29 pm

If 'sha512' is not calculated, then you can request hashes to be calculated from the UI, but if 'sha512' is calculated, and other hashes are SUPPORTED but not yet calculated, there is no way to request that they be calculated from the UI.  

screen shot 2017-05-03 at 7 39 16 pm

You can do everything with the alternative hashes from the API, and those work correctly
  * requesting to calculate the additional hashes even when sha512 exists
  * downloading a keyfile for alternate hashes
  * downloading the file itself via alternate hash algos

It may not be worth it to change the UI to be less 'sha512' centric, but I'd want to hear from you explicitly about this.

I also noticed some weird behavior from swagger, when I went to a download endpoint and executed a download via fileid, swagger shows you a response body link that you can click on to download the file, but if you click the link the file gets downloaded as '-filename.ext-' instead of 'filename.ext'.  I've seen this for both `/file/hashsum/{algo}/{hash}/download` and `/file/{id}/download`.

screen shot 2017-05-03 at 7 55 50 pm

zachmullen:
Yeah, the UI can definitely be improved. The intent of this PR isn't to take the hashsum_download plugin into a "final form", but simply to get us a bit closer in a limited scope of effort. If we think we shouldn't merge this until it generally supports multiple hash algorithms, we can do that, but I'm happy to defer that to future topics.

zachmullen:
Not sure about the swagger thing, might be worth reporting as an issue.

mgrauer:
Since this is the scope you planned, I'm fine with the UI as is, to be improved later as determined by what we actually need it to do.

This works as is and is tested, so I'm good with it.

zachmullen:
Thanks for the QA @mgrauer !

jcfr:
@thewtex This may be of interest in regards of ITK testing data.

@mgrauer Do you think we could enable the plugin and support md5 on data.kitware.com ?  Or should we wait the associated UI is improved ?

thewtex:
:+1:

zachmullen:
I'd rather not enable md5 on DKC since the algorithm has been so thoroughly broken as a cryptographic hash...

PR #1972, opened 2017-05-01
Remove alerts from the web client.

manthey:
See the discussion in PR #1968.

Add more unit tests to the base Model class.

### DIFF
--- a/clients/web/test/spec/modelSpec.js
+++ b/clients/web/test/spec/modelSpec.js
@@ -1,11 +1,20 @@
 /**
+ * Intercept window.location.assign calls so we can test the behavior of,
+ * e.g. download directives that occur from js.
+ */
+(function () {
+    window.location.assign = function (url) {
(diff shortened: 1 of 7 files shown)


manthey:
ping @zachmullen

clients/web/test/spec/modelSpec.js
```diff
@@ -1,11 +1,20 @@
 /**
+ * Intercept window.location.assign calls so we can test the behavior of,
+ * e.g. download directives that occur from js.
+ */
+(function () {
+    window.location.assign = function (url) {
+        girderTest._redirect = url;
+    };
+}(window.location.assign));
```
zachmullen: Why are we passing `window.location.assign` to a function that does not take any arguments?

manthey: Good question.  I copied this from `clients/web/test/spec/dataSpec.js`.

PR #1975, opened 2017-05-03
Make catch-all "params" kwarg optional for route handlers

zachmullen:
This value was previously required on every route handler and
was a dict containing any additional params passed to the request
that were not captured by described params.

Note, this only works when using `autoDescribeRoute`.


girder/api/describe.py
```diff
@@ -540,6 +540,20 @@ def _passArg(self, fun, kwargs, name, val):
...
+        Combines the sort/sortdir params appropriately for consumption by the model
+        layer, and only passes the "params" catch-all dict if there is a corresponding
+        kwarg for it in the wrapped function.
+        """
+        if self.description.hasPagingParams and 'sort' in kwargs:
+            sortdir = kwargs.pop('sortdir', None) or kwargs['params'].pop('sortdir', None)
+            kwargs['sort'] = [(kwargs['sort'], sortdir)]
+
+        if 'params' not in fun._fnArgs and not fun._fnKeywds:
+            kwargs.pop('params', None)
```
danlamanna: Is there any way that `params` could contain information that isn't being collected by args/kwargs?

zachmullen: It's a dict that will contain any parameters that were passed but were not listed in the description. Is that the question you were asking...?

brianhelba:
I'll QA this.

zachmullen:
Your changes LGTM

brianhelba:
I've removed all the remaining unused "params" parameter from autoDescribeRoute-decorated endpoints.

I do see that using `inspect.getargspec` to get function signatures, has been [deprecated since Python 3.0](https://docs.python.org/3.5/library/inspect.html#inspect.getargspec). The [new replacement method](https://docs.python.org/3.5/library/inspect.html#inspect.signature) in Python 3.5 is able to look through call-signature-modifying decorators, which would allow us to move this functionality to a more general place than just inside `autoDescribeRoute`. I'd like to take a little more time to ensure there's no way to work around this in Python 2.

I also want to port a few holdout routes from `describeRoute` to `autoDescribeRoute`, so we can drop their `params` too.

jbeezley:
You can use funcsigs on 2.7.  I'm using it in my girder_worker PR at https://github.com/girder/girder_worker/blob/a03d0f92466eb180f50942ef99bc38f929f53611/girder_worker/describe.py

brianhelba:
@zachmullen I'd still like to have `params` be optional for all REST handlers, not just those decorated by `@autoDescribeRoute`. However, while `funcsigs` may provide the necessary code reflection capabilities, it looks like it will take some more work to make the upstream logic not break the existing behavior of `@autoDescribeRoute`.

So, I think we should merge now, as we can always generalize this functionality later, without changing the behavior as seen by implementing handlers.

I'll look into the failing tests now.

plugins/item_tasks/server/rest.py
```diff
@@ -36,7 +36,7 @@ def __init__(self):
         .pagingParams(defaultSort='name')
     )
     @filtermodel(model='item')
```
brianhelba: It looks like the presence of this `@filtermodel` decorator before the `@autoDescribeRoute` decorator is causing the tests to fail, when `listTasks` does not have a .params`. Moving `@filtermodel` to run after `@autoDescribeRoute` fixes the issue.

However, Girder has never made a requirement that its decorators be applied in a particular order, so I think this bug should be fixed before merging. Perhaps we should also modify one of our test cases to feature a decorator between `@autoDescribeRoute` and the base function, since `@autoDescribeRoute` has so much complicated reflection behavior now.

brianhelba:
@zachmullen I've added two more commits. The first updates the tests to demonstrate the previously-failing behavior. The second adds a fix.

Commit messages:
* Test autoDescribeRoute with no params and other prior decorators
* Use inspect.signature to get parameters of autoDescribeRoute-wrapped functions
  * This ensures that the underlying wrapped function is always being examined, not any intermediary decorators.
  * Also, this stores arguments within the decorator instance, instead of adding additional attributes to the wrapped function.

PTAL at this.

brianhelba:
@jbeezley Thanks for the tip, I ended up using that package.

clients/web/girder
```diff
@@ -0,0 +1 @@
+src
```
jbeezley: Was adding this symlink intentional?

jbeezley:
Should we also remove the `params` argument for the examples in the[documentation](http://girder.readthedocs.io/en/latest/plugin-development.html#adding-a-new-route-to-the-web-api) as well?

requirements.txt
```diff
@@ -5,6 +5,7 @@ cffi==1.8.3
 CherryPy==6.0.1
 filelock==2.0.8
 functools32==3.2.3.post2; python_version <= '2.7'
+funcsigs==1.0.2 ; python_version < '3.5'
```
zachmullen: Shouldn't we also include this in setup.py?

clients/web/girder
brianhelba: That was a mistake. I'll fix.

brianhelba:
Add new commits:
* Ensure that the "funcsigs" Python library is installed when appropriate
* Document that "params" is optional with "@autoDescribeRoute"

I also removed the mistakenly-committed symlink.

PTAL.

setup.py
```diff
@@ -103,6 +103,8 @@ def run(self, *arg, **kw):
             'hachoir-parser'
         ]
     })
+if sys.version_info[0:2] <= (3, 5):
```
zachmullen: This can just be `if sys.version_info < (3, 5)`

brianhelba: You're right. I'll fix.

brianhelba: @zachmullen Fixed.

brianhelba:
@jbeezley @zachmullen PTAL again.

Issue #1971, opened 2017-05-01
Support 'du' type API endpoint on assetstore


mgrauer:
This is to answer the question: "what is taking up all the space in my assetstore?"

It gets a bit tricky because all of a collection or user folder may not be rooted in the same assetstore (perhaps not all the files in a given item may be in the same assetstore), which is why I'm requesting on this as an endpoint on assetstore rather than an instance.

The endpoint should require site admin access, which is good, because this might be slow.

The endpoint should return a descending sorted list of Collections and Users by cumulative file size, based on the files that are stored in the given Assetstore for those Collections and Users.

I'm outlining the approach in case anyone has questions/comments/further requests, and I expect some further optimizations are possible.  

For each User/Collection

1. get the items rooted at the given User or Collection
2. get the files in those items that live in the assetstore in question
3. accumulate the sizes of those files

This will potentially be complicated by copied items, as those could add a multiple to the size of files.

zachmullen:
This wouldn't be horribly expensive to compute if you want granularity of root-level objects. We can do an indexed lookup of all files contained in a given assetstore: O(lg n). Then we can iterate over the result set, reducing to a dict that maps (root type, id) -> total size. Unfortunately files don't directly store their root pointers, but their items do, so it'll just be one extra lookup per file to build that hashtable.

mgrauer:
Thanks, I didn't realize that file was [indexed](https://github.com/girder/girder/blob/master/girder/models/file.py#L38) on assetstore.

PR #1979, opened 2017-05-03
[WIP] an Assetstore space usage endpoint

matthewma7:
For the issue https://github.com/girder/girder/issues/1971

This is an initial implementation. Open to any suggestions. @mgrauer 
The size is currently calculated without considering copied files or files have the same hash.

The API endpoint returns result in the following format.

```json
{
  "collections": [
    {
      "collection": {
        "_accessLevel": 2,
        "_id": "5907a19f1d0e5c5498610964",
        "_modelType": "collection",
        "created": "2017-05-01T20:59:11.734000+00:00",
        "description": "",
        "name": "collection2",
        "public": true,
        "publicFlags": [],
        "size": 5120,
        "updated": "2017-05-01T20:59:11.734000+00:00"
      },
      "size": 4096
    },
    {
      "collection": {
        "_accessLevel": 2,
        "_id": "5907a1841d0e5c5498610963",
        "_modelType": "collection",
        "created": "2017-05-01T20:58:44.113000+00:00",
        "description": "",
        "name": "collection1",
        "public": false,
        "size": 2048,
        "updated": "2017-05-01T20:58:44.113000+00:00"
      },
      "size": 1024
    }
  ],
  "users": [
    {
      "size": 5120,
      "user": {
        "_accessLevel": 2,
        "_id": "59079e101d0e5c4dc1ad6dba",
        "_modelType": "user",
        "admin": true,
        "created": "2017-05-01T20:44:00.163000+00:00",
        "email": "girder@email.com",
        "emailVerified": true,
        "firstName": "girder",
        "groupInvites": [],
        "groups": [],
        "lastName": "girder",
        "login": "girder",
        "public": true,
        "size": 0,
        "status": "enabled"
      }
    }
  ]
}
```

### DIFF
--- a/girder/api/v1/assetstore.py
+++ b/girder/api/v1/assetstore.py
@@ -29,6 +29,7 @@ class Assetstore(Resource):
     """
     API Endpoint for managing assetstores. Requires admin privileges.
     """
+
     def __init__(self):
         super(Assetstore, self).__init__()
         self.resourceName = 'assetstore'
@@ -39,6 +40,7 @@ class Assetstore(Resource):
         self.route('PUT', (':id',), self.updateAssetstore)
         self.route('DELETE', (':id',), self.deleteAssetstore)
         self.route('GET', (':id', 'files'), self.getAssetstoreFiles)
+        self.route('GET', (':id', 'spaceusage',), self.getSpaceUsageStatistics)
 
     @access.admin
     @autoDescribeRoute(
@@ -248,3 +250,73 @@ class Assetstore(Resource):
         return list(self.model('file').find(
             query={'assetstoreId': assetstore['_id']},
             offset=offset, limit=limit, sort=sort))
+
+    @access.admin
+    @autoDescribeRoute(
+        Description('get space usage statistics by collection and by user.')
+        .modelParam('id', model='assetstore')
+        .errorResponse()
+        .errorResponse('You are not an administrator.', 403)
+    )
+    def getSpaceUsageStatistics(self, assetstore, params):
+        filesInStore = list(self.model('file').find(
+            query={'assetstoreId': assetstore['_id']}))
+
+        itemsUsingStore = list(self.model('item').find(query={
+            '_id': {
+                '$in': [x['itemId'] for x in filesInStore]
+            },
+            'baseParentType': 'collection'
+        }))
+        itemsUsingStoreMap = dict((x['_id'], x) for x in itemsUsingStore)
+
+        collectionUsingStore = list(self.model('collection').find(
+            query={
+                '_id': {
+                    '$in': [x['baseParentId'] for x in itemsUsingStore]
+                }
+            }))
+        collectionUsingStoreMap = dict((x['_id'], x) for x in collectionUsingStore)
+
+        collectionFileMap = {}
+        for file in filesInStore:
+            collectionId = itemsUsingStoreMap[file['itemId']]['baseParentId']
+            if collectionId not in collectionFileMap:
+                collectionFileMap[collectionId] = [file]
+            else:
+                collectionFileMap[collectionId].append(file)
+
+        usersUsingStore = list(self.model('user').find(
+            query={
+                '_id': {
+                    '$in': [x['creatorId'] for x in filesInStore]
+                }
+            }))
+        userUsingStoreMap = dict((x['_id'], x) for x in usersUsingStore)
+        userFileMap = {}
+        for file in filesInStore:
+            userId = file['creatorId']
+            if userId not in userFileMap:
+                userFileMap[userId] = [file]
+            else:
+                userFileMap[userId].append(file)
+
+        # based on collected data stored in dictionary, aggregate size and format result
+        currentUser = self.getCurrentUser()
+        collections = list({
+            'collection': self.model('collection')
+            .filter(collectionUsingStoreMap[collectionId], currentUser),
+            'size': sum(file['size'] for file in files)
+        } for collectionId, files in collectionFileMap.items())
+        collections.sort(key=lambda x: x['size'], reverse=True)
+
+        users = list({
+            'user': self.model('user').filter(userUsingStoreMap[userId], currentUser),
+            'size': sum(file['size'] for file in files)
+        } for userId, files in userFileMap.items())
+        users.sort(key=lambda x: x['size'], reverse=True)
+
+        return {
+            'collections': collections,
+            'users': users
+        }
--- a/girder/api/v1/assetstore.py
+++ b/girder/api/v1/assetstore.py
@@ -16,7 +16,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 ###############################################################################
-
+from collections import defaultdict
+import six
 from ..describe import Description, autoDescribeRoute
 from ..rest import Resource, RestException
 from girder import events
@@ -264,11 +265,11 @@ class Assetstore(Resource):
 
         collectionItemsInStore = list(self.model('item').find(query={
             '_id': {
-                '$in': list(set([x['itemId'] for x in filesInStore]))
+                '$in': list(set([file['itemId'] for file in filesInStore if 'itemId' in file]))
             },
             'baseParentType': 'collection'
         }))
-        collectionItemsInStoreMap = dict((x['_id'], x) for x in collectionItemsInStore)
+        collectionItemsInStoreMap = dict((item['_id'], item) for item in collectionItemsInStore)
 
         collectionUsingStore = list(self.model('collection').find(
             query={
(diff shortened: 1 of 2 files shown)


girder/api/v1/assetstore.py
```diff
@@ -248,3 +250,73 @@ def getAssetstoreFiles(self, assetstore, limit, offset, sort, params):
...
+    @access.admin
+    @autoDescribeRoute(
+        Description('get space usage statistics by collection and by user.')
+        .modelParam('id', model='assetstore')
+        .errorResponse()
+        .errorResponse('You are not an administrator.', 403)
+    )
+    def getSpaceUsageStatistics(self, assetstore, params):
+        filesInStore = list(self.model('file').find(
+            query={'assetstoreId': assetstore['_id']}))
```
zachmullen: This could be many thousands of files; we shouldn't pull them all into memory via `list()`. We'll want to iterate over the cursor.

girder/api/v1/assetstore.py
```diff
@@ -248,3 +250,73 @@ def getAssetstoreFiles(self, assetstore, limit, offset, sort, params):
...
+
+        itemsUsingStore = list(self.model('item').find(query={
+            '_id': {
+                '$in': [x['itemId'] for x in filesInStore]
```
zachmullen: This will need to change also, so that the size of the query sent to mongo is not O(n) with the number of files in the assetstore.

mgrauer:
>I'm wondering if it would be a better idea to simply copy the baseParentType & baseParentId info onto all files so we could do this without O(n) item lookups...

This will be slow without that change for sure, but that would involve some core changes, add to the amount we store in the DB for all files, and require a migration, for what is a rarely used site-admin-only maintenance action, so I'd rather not make that change.

girder/api/v1/assetstore.py
matthewma7: Changed.

matthewma7:
This is ready for review

The output format has been simplified to

```json
{
  "collections": [
    {
      "_id": "5907a19f1d0e5c5498610964",
      "name": "collection2",
      "spaceUsage": 4096
    },
    {
      "_id": "5907a1841d0e5c5498610963",
      "name": "collection1",
      "spaceUsage": 1024
    }
  ],
  "users": [
    {
      "_id": "59079e101d0e5c4dc1ad6dba",
      "email": "girder@email.com",
      "firstName": "girder",
      "lastName": "girder",
      "login": "girder",
      "spaceUsage": 5120
    }
  ]
}
```

girder/api/v1/assetstore.py
```diff
@@ -248,3 +250,96 @@ def getAssetstoreFiles(self, assetstore, limit, offset, sort, params):
         return list(self.model('file').find(
             query={'assetstoreId': assetstore['_id']},
             offset=offset, limit=limit, sort=sort))
+
+    @access.admin
+    @autoDescribeRoute(
+        Description('get space usage statistics by collection and by user.')
```
zachmullen: Capitalize description

girder/api/v1/assetstore.py
```diff
@@ -39,6 +40,7 @@ def __init__(self):
         self.route('PUT', (':id',), self.updateAssetstore)
         self.route('DELETE', (':id',), self.deleteAssetstore)
         self.route('GET', (':id', 'files'), self.getAssetstoreFiles)
+        self.route('GET', (':id', 'spaceusage',), self.getSpaceUsageStatistics)
```
zachmullen: I'd vote for renaming this to `GET assetstore/:id/stats`, that way we can add additional info under the same endpoint in the future.

zachmullen:
I got an error when I ran this via swagger:

```
{
  "message": "KeyError: KeyError(None,)",
  "trace": [
    [
      "/Users/zach/dev/girder/girder/api/rest.py",
      567,
      "endpointDecorator",
      "val = fun(self, args, kwargs)"
    ],
    [
      "/Users/zach/dev/girder/girder/api/rest.py",
      1120,
      "GET",
      "return self.handleRoute('GET', path, params)"
    ],
    [
      "/Users/zach/dev/girder/girder/api/rest.py",
      890,
      "handleRoute",
      "val = handler(**kwargs)"
    ],
    [
      "/Users/zach/dev/girder/girder/api/access.py",
      39,
      "wrapped",
      "return args[0](*iargs, **ikwargs)"
    ],
    [
      "/Users/zach/dev/girder/girder/api/describe.py",
      601,
      "wrapped",
      "return fun(*args, **kwargs)"
    ],
    [
      "/Users/zach/dev/girder/girder/api/v1/assetstore.py",
      339,
      "getSpaceUsageStatistics",
      "users = [userMapper(x, y) for x, y in userFileMap.items()]"
    ],
    [
      "/Users/zach/dev/girder/girder/api/v1/assetstore.py",
      326,
      "userMapper",
      "'user': userModel.filter(userUsingStoreMap[userId], currentUser),"
    ]
  ],
  "type": "internal"
}
```

girder/api/v1/assetstore.py
matthewma7: I see. In that case, did you also think we should push space usage one level down in the JSON result?

zachmullen: Sure, that's a good call.

matthewma7:
This is ready for review. 🐴

girder/api/v1/assetstore.py
```diff
@@ -248,3 +250,96 @@ def getAssetstoreFiles(self, assetstore, limit, offset, sort, params):
...
+            userId = file['creatorId']
+            if userId not in userFileMap:
+                userFileMap[userId] = [file]
+            else:
+                userFileMap[userId].append(file)
+
+        # based on collected data stored in dictionary, aggregate size and format
+        # result. Ignore the fact there will be only one copy for the same file
+        # in the assetstore.
+        currentUser = self.getCurrentUser()
```
brianhelba: Per my comments below, I don't think the output should depend on the user calling this endpoint (since it's admin-only).

girder/api/v1/assetstore.py
```diff
@@ -248,3 +250,96 @@ def getAssetstoreFiles(self, assetstore, limit, offset, sort, params):
...
+        def collectionMapper(collectionId, files):
+            collection = collectionModel.filter(collectionUsingStoreMap[collectionId], currentUser)
+            collection = {
+                '_id': collection['_id'],
+                'name': collection['name']
+            }
+            collection['spaceUsage'] = sum(file['size'] for file in files)
+            return collection
+        collections = [
+            collectionMapper(x, y) for x, y in collectionFileMap.items()]
```
brianhelba: Should use `six.viewitems()` here, though pretty much none of the code is using iterators.

girder/api/v1/assetstore.py
```diff
@@ -248,3 +250,96 @@ def getAssetstoreFiles(self, assetstore, limit, offset, sort, params):
...
+        collections.sort(key=lambda x: x['spaceUsage'], reverse=True)
```
brianhelba: You can create and sort a list in one shot, by passing a generator comprehension to the `sorted` function.

matthewma7:
@brianhelba, Thank you for the review. 
Yes, the number of files is a problem, but I think @mgrauer asked this API at this moment as a pure API endpoint for admin use, so adding additional database field or special handling was not chosen. Do we have some knowledge of the number of files of the largest Assetstore we have? If there are more than half a million files, I could change this to a batch process fashion.

girder/api/v1/assetstore.py
```diff
@@ -248,3 +250,96 @@ def getAssetstoreFiles(self, assetstore, limit, offset, sort, params):
...
+        def userMapper(userId, files):
+            if userId not in userUsingStoreMap:
+                return None
+            user = userModel.filter(userUsingStoreMap[userId], currentUser)
+            user = {
+                '_id': user['_id'],
+                'login': user['login'],
+                'firstName': user['firstName'],
+                'lastName': user['lastName'],
+                'email': user['email']
```
zachmullen: We are already calling `filter` on the user document above -- if we are going to just pull out some select fields from the user document, there's no need to filter, it should be one or the other.

girder/api/v1/assetstore.py
```diff
@@ -248,3 +250,96 @@ def getAssetstoreFiles(self, assetstore, limit, offset, sort, params):
...
+            }
+            user['spaceUsage'] = sum(file['size'] for file in files)
+            return user
+        users = [z for z in (userMapper(x, y) for x, y in userFileMap.items()) if z]
```
matthewma7: I see, variable names updated.

girder/api/v1/assetstore.py
matthewma7: I wanted to be safe, but just casting one time works fine for the case.

girder/api/v1/assetstore.py
```diff
@@ -248,3 +250,96 @@ def getAssetstoreFiles(self, assetstore, limit, offset, sort, params):
...
+                    collectionFileMap[collectionId].append(file)
+
+        usersUsingStore = list(self.model('user').find(
+            query={
+                '_id': {
+                    '$in': list(set([x['creatorId'] for x in filesInStore]))
+                }
+            }))
+        userUsingStoreMap = dict((x['_id'], x) for x in usersUsingStore)
+        userFileMap = {}
```
matthewma7: Thank you for letting me know about defaultdict, it's very useful.

matthewma7:
Ready for further review.

brianhelba:
I'll post my aggregation pipeline. Let's see what the scalability looks like with that.

Xarthisius:
If you consider downstream projects: https://girder.hub.yt/#collection/5719870753a1a2000166a615  this one has over 850k files :)

mgrauer:
@brianhelba Would you be able to post your aggregation pipeline?

@Xarthisius comment reminds me, we should add a warning to the swagger docs/description of this endpoint mentioning that it is for admin site maintenance only and could potentially be quite slow.

zachmullen:
Conclusion: we should do this with an aggregation pipeline. Closing until someone can adopt this.

Alternatively: if we don't want to go the whole 9 yards, we can just turn this logic into a script rather than a user-facing feature.

mgrauer:
There is some current research work into DICOM metadata search that will involve use of aggregation pipelines.  The outcome of that research could inform a revised solution here.

PR #1981, opened 2017-05-04
grunt_tasks/webpack: Fix depreacted warnings

jcfr:
This commit update webpack config to address these warnings:

fallbackLoader option has been deprecated - replace with "fallback"
loader option has been deprecated - replace with "use"
fallbackLoader option has been deprecated - replace with "fallback"
loader option has been deprecated - replace with "use"


jcfr:
Thanks for the review.

PR #1983, opened 2017-05-04
Add girder_worker job cancellation

cjh1:
The brother PR in girder_worker is girder/girder_worker#128


plugins/worker/server/__init__.py
```diff
@@ -91,6 +94,32 @@ def schedule(event):
...
+
+        if celeryTaskId is None:
+            msg = ("Unable to cancel Celery task. Job '%s' doesn't have a Celery task id."
+                   % job['_id'])
+            logger.warn(msg)
+            return
+
+        # Send the task to celery
+        asyncResult = AsyncResult(celeryTaskId)
+        asyncResult.revoke()
```
cjh1: @kotfic I'm not appose to added a terminate and signal parameter to this endpoint, which I think is all that would required for your use case? With the appropriate warnings of course :smile:

cjh1:
Still todo:

- [ ] Update canceling icon to rotating orange
- [ ] Make canceling jobs in the following states a noop: ['success', 'error', 'canceling']

zachmullen:
The rotation of the icon is super easy: https://github.com/girder/girder/blob/master/plugins/jobs/web_client/JobStatus.js#L75

cjh1:
Thanks, super easy sounds good :smile:

plugins/jobs/web_client/views/JobDetailsWidget.js
```diff
@@ -112,6 +120,19 @@ var JobDetailsWidget = View.extend({
...
+
+    _cancelJob: function () {
+        const jobId = this.job.id;
+        restRequest({
+            path: `job/${jobId}/cancel`,
+            type: 'PUT',
+            error: null
+        }).done(_.bind(function () {
+            console.log('success');
+        }, this)).error(_.bind(function (err) {
```
brianhelba: Per the changes in https://github.com/girder/girder/pull/2032 , this should be changed to `.fail`.

zachmullen:
@cjh1 The failing test appears to be a result of a bad request, see this screenshot: http://my.cdash.org/testDetails.php?test=32164915&build=1223795

zachmullen:
[The JSON](https://www.celebsclothing.com/product_images/e/183/Happy_Days_Jacket__82354_std.jpg)

cjh1:
@zachmullen Thanks that helps alot, I will fix up those tests.

cjh1:
@girder/developers This is ready for review, PTAL

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -260,7 +307,33 @@ var JobListWidget = View.extend({
             filter.statuses = JSON.stringify(this.statusFilter);
         }
         this.collection.params = filter;
+
         return this.collection.fetch({}, true);
+    },
+
+    _cancelJobs: function () {
+        Promise.all(
```
brianhelba: Per the consensus in #2068, we need to stick with jQuery Promise until at least Girder 3.

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -260,7 +307,33 @@ var JobListWidget = View.extend({
...
+            Object.keys(this.jobCheckedStates)
+                .filter(jobId => {
+                    var status = this.jobs.find(job => job.id === jobId).get('status');
+                    return [JobStatus.CANCELED, JobStatus.SUCCESS, JobStatus.ERROR].indexOf(status) === -1;
+                })
+                .map(jobId => Promise.resolve(restRequest({
```
brianhelba: Per the consensus in #2068, we need to stick with jQuery Promise until at least Girder 3.

plugins/jobs/plugin_tests/jobsSpec.js
```diff
@@ -59,12 +59,28 @@ $(function () {
...
 
                 expect($('.g-timeline-segment').length).toBe(3);
                 expect($('.g-timeline-point').length).toBe(4);
                 expect($('.g-timeline-start-label').text()).toBe('0 s');
                 expect($('.g-timeline-end-label').text()).toBe('12 s');
-                expect($('.g-timeline-point')[3].className).toContain('g-job-color-success');
+
+                function toHex(c) {
+                    var hex = parseInt(c, 10).toString(16);
+                    return hex.length == 1 ? '0' + hex : hex;
```
brianhelba: Always use `===` in JS.

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -45,6 +69,17 @@ var JobListWidget = View.extend({
             obj[status.text] = true;
             return obj;
         }, {});
+        this.jobCheckedStates = {};
+        this.jobHighlightStates = {};
+        Object.defineProperty(this, 'jobs', {
```
brianhelba: Backbone Views don't expose their methods this way. Why is this being used?

plugins/jobs/web_client/views/JobListWidget.js
matthewma7: I have another PR that has the Promise.all like $.whenAll, I am hoping that PR could go in before this one, then I'll merge from master to change this to jQuery Promise.

plugins/jobs/web_client/views/JobListWidget.js
matthewma7: This is not directly related to Backbone, this is the es5 way to define property without ES6 class syntax.

plugins/worker/server/__init__.py
```diff
@@ -94,6 +117,34 @@ def schedule(event):
...
+            logger.warn(msg)
+            return
+
+        if job['status'] not in [JobStatus.SUCCESS, JobStatus.ERROR]:
+            # Set the job status to canceling
+            ModelImporter.model('job', 'jobs').updateJob(job, status=CustomJobStatus.CANCELING)
+
+            # Send the revoke request.
+            asyncResult = AsyncResult(celeryTaskId)
+            asyncResult.revoke()
```
kotfic: ```signal``` and ```terminate``` kwargs are not being passed through from REST endpoint.

plugins/worker/server/__init__.py
```diff
@@ -94,6 +117,34 @@ def schedule(event):
...
 
 
+def cancel(event):
+    """
+    This is bound to the "jobs.cancel" event, and will be triggered any time
+    a job is canceled. This handler will process any job that has the
+    handler field set to "worker_handler".
+    """
+    job = event.info
+    if job['handler'] == 'worker_handler':
```
kotfic: ```worker_handler``` is the handler that manages traditional ```app.send_message(...)``` task production.  Please also white list ```"celery_handler"``` which is the handler type we've assigned to traditional celery based task distribution (e.g. with ```my_task.delay(...)``` and ```my_task.apply_async(...)```)

plugins/worker/server/__init__.py
cjh1: Correctly, I planned on adding that in a later PR if that makes sense?

plugins/worker/server/__init__.py
cjh1: Ok, will do

cjh1:
This is ready for another look, tests are passing. Not sure how to read the coverage report, is this a fault positive?

plugins/jobs/web_client/views/JobListWidget.js
brianhelba: Unless there's an important reason for these attributes to be assigned this way, I'd strongly prefer that these be created using the normal Backbone syntax of `extend`. Also, these should be private (prefixed with a `_`).

plugins/jobs/web_client/views/JobDetailsWidget.js
brianhelba: This still needs to be fixed.

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -232,20 +277,25 @@ var JobListWidget = View.extend({
...
         this._renderData();
-        this._highlightRecordIfOnList(event.data._id);
+        this.trySetHighlightRecord(event.data._id);
     },
 
     _jobCreated: function (event) {
         this._fetchWithFilter()
-            .done(() => {
-                this._highlightRecordIfOnList(event.data._id);
+            .then(() => {
```
brianhelba: This should use `.done`, with no return value.

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -261,7 +311,34 @@ var JobListWidget = View.extend({
...
+                .filter((jobId) => {
+                    var jobModel = this.jobs.find((job) => job.id === jobId);
+                    return JobStatus.isCancelable(jobModel);
+                })
+                .map((jobId) => restRequest({
+                    path: `job/${jobId}/cancel`,
+                    type: 'PUT',
+                    error: null
+                }))
+        ).then((results) => {
```
brianhelba: This should use `.done`, with no return value.

plugins/jobs/web_client/views/JobListWidget.js
matthewma7: @brianhelba Could you detail what do you mean by using Backbone `extend`?

matthewma7: @brianhelba Could you illustrate your point on `extend`?

plugins/jobs/web_client/views/JobListWidget.js
matthewma7: I have to say the `done` and `then` mix use is really confusing to me. 😵

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -233,20 +269,24 @@ var JobListWidget = View.extend({
...
+                this.trySetHighlightRecord(event.data._id);
             });
     },
 
-    _highlightRecordIfOnList: function (jobId) {
-        if (this.currentView === 'list') {
-            var tr = this.$('tr[g-job-id=' + jobId + ']').addClass('g-highlight');
-            setTimeout(() => tr.removeClass('g-highlight'), 1000);
+    trySetHighlightRecord: function (jobId) {
+        if (this.collection.find((job) => job.id === jobId)) {
```
brianhelba: [`Collection.get`](http://backbonejs.org/#Collection-get) is the correct (and more efficient) way to look up a Model by id:
```
if (this.collection.get(jobId)) {
```

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -217,7 +249,11 @@ var JobListWidget = View.extend({
                 triggerJobClick: this.triggerJobClick,
                 JobStatus: JobStatus,
                 formatDate: formatDate,
-                DATE_SECOND: DATE_SECOND
+                DATE_SECOND: DATE_SECOND,
+                jobCheckedStates: this.jobCheckedStates,
+                jobHighlightStates: this.jobHighlightStates,
+                anyJobChecked: _.find(_.values(this.jobCheckedStates), (status) => status === true),
```
brianhelba: There's no need to use `_.values` here, Underscore's [`_.find`](http://underscorejs.org/#find) works just fine on objects:
```
anyJobChecked: _.find(this.jobCheckedStates, (status) => status === true),
```

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -28,6 +29,29 @@ var JobListWidget = View.extend({
         'change select.g-page-size': function (e) {
             this.collection.pageLimit = parseInt($(e.target).val());
             this.collection.fetch({}, true);
+        },
+        'change input.g-job-checkbox': function (e) {
+            var jobId = $(e.target).closest('tr').attr('g-job-id');
+            if ($(e.target).is(':checked')) {
```
brianhelba: You usually want to use `e.currentTarget`, to ensure that you're selecting the element this handler was bound to, if the original event fired on a sub-element.

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -28,6 +29,29 @@ var JobListWidget = View.extend({
...
+            } else {
+                delete this.jobCheckedStates[jobId];
+            }
+            this._renderData();
+        },
+        'click input.g-job-checkbox-all': function (e) {
+            e.stopPropagation();
+        },
+        'change input.g-job-checkbox-all': function (e) {
+            if ($(e.target).is(':checked')) {
```
brianhelba: Same here with `e.currentTarget`.

plugins/jobs/web_client/views/JobDetailsWidget.js
```diff
@@ -119,6 +126,19 @@ var JobDetailsWidget = View.extend({
...
+    },
+
+    _cancelJob: function () {
+        const jobId = this.job.id;
+        restRequest({
+            path: `job/${jobId}/cancel`,
+            type: 'PUT',
+            error: null
+        }).done(_.bind(function () {
+            console.log('success');
```
brianhelba: These debugging statements should be removed, or converted to UI notifications.

plugins/jobs/web_client/JobStatus.js
```diff
@@ -47,12 +53,20 @@ var JobStatus = {
...
+
+            let statusInfo = {
                 text: info.text,
                 icon: info.icon,
                 color: info.color,
+                textColor: 'white',
                 finished: _.has(info, 'finished') ? info.finished : false
             };
+
+            if ('textColor' in info) {
```
brianhelba: `in` is a bit tricky to use, since it checks up the entire prototype chain. Maybe you want [`_.has`](http://underscorejs.org/#has)?

clients/web/src/misc.js
```diff
@@ -213,7 +213,7 @@ function capitalize(str) {
 
 function splitRoute(route) {
     if (!route) {
-        return {base: '', name: ''};
+        return { base: '', name: '' };
```
brianhelba: Not sure why this is changed.

cjh1:
@zachmullen @kotfic Conflicts fixed

plugins/jobs/server/models/job.py
```diff
@@ -384,12 +395,23 @@ def updateJob(self, job, log=None, overwrite=False, status=None,
             job['updated'] = now
             updates['$set']['updated'] = now
 
-            self.update({'_id': job['_id']}, update=updates, multi=False)
+            update_result = self.update(query, update=updates, multi=False)
```
zachmullen: camelCase

plugins/jobs/web_client/JobStatus.js
cjh1: How is it a problem in this situation?

plugins/jobs/web_client/views/JobDetailsWidget.js
cjh1: Can you provide an example of a UI notification? Thanks

plugins/jobs/web_client/JobStatus.js
zachmullen: I think it goes like this:

If you intend to check up the entire prototype chain to see if the object has any property with the given name, use `in`.

If you only want to check if the object itself has the given property (which is likely what you want here) you should use `hasOwnProperty`, which is equivalent to `_.has`.

plugins/jobs/web_client/views/JobDetailsWidget.js
zachmullen: If you're looking for a global notification, it looks like this:

```
events.trigger('g:alert', {
                    icon: 'ok',
                    text: 'Folder info updated.',
                    type: 'success',
                    timeout: 4000
                });
```

clients/web/src/misc.js
cjh1: Not sure @matthewma7?, seem harmlesss

plugins/jobs/server/models/job.py
cjh1: Fixed

plugins/jobs/web_client/views/JobListWidget.js
cjh1: Ok, fixed

plugins/jobs/web_client/views/JobListWidget.js
cjh1: fixed

plugins/jobs/web_client/views/JobListWidget.js
cjh1: Ok, again harmless, but  fixed

plugins/jobs/web_client/views/JobListWidget.js
cjh1: fixed

plugins/jobs/web_client/JobStatus.js
cjh1: Sure, but I am guessing 'textColor' is not in the prototype of {}. I will change it.

plugins/jobs/web_client/views/JobDetailsWidget.js
cjh1: @zachmullen Thanks

cjh1: Done

plugins/jobs/web_client/JobStatus.js
cjh1: Fixed

zachmullen: Thanks -- I think this is just something that was hammered into those of us who read Crockford's treatises on javascript: "only use `in` when you really mean it", to the point that when we do see it now, it raises a red flag, even if it probably doesn't matter.

brianhelba: Yeah, it's not a big deal in this context, but IMO, given that it now raises a mental red flag for me and other Crockford disciples, it decreases the readability of the code every time I see `in` used, because I have to stop and think about whether it's intentional or a bug.

cjh1:
I think I have responded to the review comments

zachmullen:
@kotfic you are the last one with a change request review, can you respond with any additional feedback here?

zachmullen:
I'm going to merge this before it gets out of date again 😎

cjh1:
Awesome, thanks :relieved:

mgrauer:
Nice work, Love Tractor.

manthey:
@cjh1 How do I mark that an ordinary worker job is allowed to be cancelled?  I have a job I wrote previous to this PR that occasionally checks if it was deleted or cancels and terminates early.  The `cancel` function in `plugins/worker/server/__init__.py` prevents this from working, since it cancels the cancel.

PR #1984, opened 2017-05-04
Indicate presence of permission flags visually

zachmullen:
In the Access editor UI, users would previously have to
open the flags popover to see whether any flags were granted
to a user, group, or the public. Now, the presence of permission
flags is visible right away.

In the below screenshot, there is one public flag enabled, and one enabled for the user zachmullen.

screen shot 2017-05-04 at 3 39 49 pm

### DIFF
--- a/clients/web/src/stylesheets/widgets/accessWidget.styl
+++ b/clients/web/src/stylesheets/widgets/accessWidget.styl
@@ -34,7 +34,7 @@ i.g-flag-icon
 .g-flag-count-indicator
   display inline-block
   border-radius 3px
-  background-color rgba(200, 0, 0, 0.75)
+  background-color #c80000
   color white
   position absolute
   font-size 9px
--- a/clients/web/src/stylesheets/widgets/accessWidget.styl
+++ b/clients/web/src/stylesheets/widgets/accessWidget.styl
@@ -28,6 +28,21 @@
   font-weight bold
   padding 3px 10px
 
+i.g-flag-icon
+  position relative
+
+.g-flag-count-indicator
+  display inline-block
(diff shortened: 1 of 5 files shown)


zachmullen:
@jeffbaumes PTAL

mgrauer:
@msmolens could you QA this please?

msmolens:
@mgrauer Yes, I will take a look.

clients/web/src/stylesheets/widgets/accessWidget.styl
```diff
@@ -28,6 +28,21 @@
   font-weight bold
   padding 3px 10px
 
+i.g-flag-icon
+  position relative
+
+.g-flag-count-indicator
+  display inline-block
+  border-radius 3px
+  background-color rgba(200, 0, 0, 0.75)
```
msmolens: What's the rationale for transparency here? Usually I'd expect an overlay like this to be opaque.

zachmullen: In case the dot was covering up the icon, I figured it might be helpful to show the shape of the icon behind. Might not look good, though. Should I change it?

msmolens: I'm not sure that there are enough pixels at these sizes to easily see the flag behind a transparent badge, so yes, I'd suggest removing the alpha.

msmolens:
@zachmullen What do you think about a "dot" overlay instead of flag count?

zachmullen:
I like the information conveyed by the count. Open to changing it later, but I think this is OK for now.

PR #1985, opened 2017-05-05
Fail gracefully if script is missing in phantom env

zachmullen:
Prior to this change, a missing plugin.min.js would cause the
test environment to hang indefinitely. Now, we fail gracefully
with an error message indicating a missing script.

ping @jeffbaumes @zackgalbreath since you both encountered this issue separately.

### DIFF
--- a/clients/web/test/specRunner.js
+++ b/clients/web/test/specRunner.js
@@ -126,6 +126,9 @@ page.onCallback = function (data) {
             break;
         case 'exit':
             // The "Testing Finished" string is magical and causes web_client_test.py not to retry
+            if (data.errorMessage) {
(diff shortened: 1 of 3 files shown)


zachmullen:
ping @girder/developers could someone review this? I just ran into this issue, it caused our tests to hit the 600s CTest timeout on Circle on a different topic 😛

jbeezley:
The changes LGTM.  Were the missing script errors caused by race conditions or were they actually missing?

zachmullen:
They were actually missing, and we simply didn't take any action if they failed to load. Phantom's behavior is to just sit there happily until you tell it explicitly to exit.

PR #1987, opened 2017-05-08
Simplify WidgetModel creation and add docstrings

jbeezley:
This replaces the logic of setting task inputs and outputs from a previous job with a single function that is more compact and easier to understand.  It also moves the logic for setting default attributes into the widget model.


jbeezley:
Codecov's status reporting is being flaky.  The real diff coverage is [100%](https://codecov.io/gh/girder/girder/commit/3d4bfb22e4f448ce95fb8c9fc67859e6d8d9d4cf).

PR #1988, opened 2017-05-08
Update Webpack config to use 2.0 syntax

brianhelba:
* Migrate Webpack config's "module.loaders" to "module.rules"
  * This is the preferred syntax for Webpack 2.0: https://webpack.js.org/guides/migrating/#module-loaders-is-now-module-rules
* Migrate Webpack config's "Rule.loaders" / "Rule.loader" to "Rule.use"
  * This is the preferred syntax for Webpack 2.0: https://webpack.js.org/configuration/module/#rule-loaders
* Migrate Webpack config to use "Rule.resource"
  * This is the canonical syntax for Webpack 2.0, and provides better encapsulation of Rule Conditions: https://webpack.js.org/configuration/module/#rule-resource
* Migrate Webpack config to use "UseEntry.options"
  * This is the canonical syntax for Webpack 2.0, and more clearly specifies which Rule the options apply to: https://webpack.js.org/configuration/module/#rule-options-rule-query
* Use a more explicit version of ExtractTextPlugin, and clean up its config
  * The 'extract-text-webpack-plugin' was already being used at version '2.1.0' (hence the changes in #1981 ), so this commit updates the package dependency to ensure that older versions are not used with the new option syntax.
  * This new version should now properly pass on options to child loaders ( https://github.com/webpack-contrib/extract-text-webpack-plugin/pull/266 ), so the options to 'stylus-loader' can now use a clearer syntax. This also adds a comment to explain a poorly documented 'stylus-loader' option.
* Explicitly mark some Webpack-chain packages as having newer versions
  * This should not actually upgrade packages, as the "^" specifier already allows minor version upgrades.
* Add compatibility code for Webpack helper files still using "module.loaders"


zachmullen:
We have downstream plugins that are using the webpack helper file and are modifying the `config.module.loaders` object. Would this break those?

ping @ronichoudhury since he could provide a test case

msmolens:
We should verify that this doesn't resurrect the problem worked around by https://github.com/girder/girder/pull/1815.

brianhelba:
@zachmullen I already share your concerns, and agree that we need to address this before merging. I at least wanted to get the changes out in a public PR first.

@msmolens Thanks for the heads up. This looks like a good time to upgrade `stylus-loader` too (in addition to just `extract-text-webpack-plugin`). I'll look into it.

ronichoudhury:
I had a discussion with Brian yesterday evening about this. I’ll try to
summarize here.

In some sense, this change (while necessary in the long term) represents a
breaking change to Girder’s API. I say that because it involves a part of
Girder that I must necessarily interact with in building, say, my Resonant
Lab plugin for Girder. Even if that’s true, however, it’s somewhat
mitigated by the fact that I may be the only person using Webpack helpers
in this way (as a reminder, my helper function installs the webpack loaders
I need for Resonant Lab code, tightly scoped to *just* my code, so that my
loaders don’t ever run on Girder core code, and Girder’s loaders - even if
shared in principle with my code - never run on my modules).

So, depending on how much the build system is considered as part of
Girder’s public API:

   1. If this is going to really be considered a breaking change, I’d say
   hold this PR until Girder 3.
   2. If it’s more of a secret, undocumented feature of Girder and we don’t
   want to consider it a breaking change, then merge it now; I won’t mind
   updating my client to be compliant.

(2) is the easier choice, but it assumes that I’m really the only one using
webpack helpers.
​

zachmullen:
The build system is considered part of the API surface (in fact it was the biggest breaking change from version 1 to 2). So, let's make this a Girder 3 thing. I'll make a label in github for it.

ronichoudhury:
Oh I completely forgot (3)! (Drinking my coffee now.)

3. Keep the old behavior but deprecate it (allowed by marking the
deprecation and bumping minor version number) - Brian's idea was to check
for a "loaders" property on the config and, if present, convert it to the
"rules" style and append it. This results in more complexity, and it might
be tough to guarantee via testing that the stopgap functionality works, but
it would let us merge this now, and make use of the "deprecation" feature
of semver for its intended purpose.

zachmullen:
Ah I like 3, this is what I was hoping for. If we can swing that, let's do it.

brianhelba:
@ronichoudhury PTAL, and let me know if this now continues to work with your use of a Webpack helper file.

brianhelba:
@girder/developers PTAL, this is ready for broader review.

To follow up on the earlier discussion, I did indeed take "option 3", and implement a bit of compatibility code for plugins with old Webpack helper files, which we'll remove in Girder 3.

@ronichoudhury is not going to have time to spin up his Webpack helper-using plugins to test against this branch, but if any other @girder/developers have a plugin that uses the Webpack helper, please chime in, so we can ensure it's compatible, and ideally update your files to use the new format.

grunt_tasks/plugin.js
```diff
@@ -309,6 +309,12 @@ module.exports = function (grunt) {
             }
 
             var newConfig = webpackHelper(grunt.config.getRaw('webpack.options'), helperConfig);
+            if (_.has(newConfig.module, 'loaders')) {
+                grunt.log.writeln(`  >> "module.loaders" is deprecated, use "module.rules" in ${webpackHelperFile} instead.`.yellow);
+                newConfig.module.rules = newConfig.module.rules || [];
+                newConfig.module.rules = newConfig.module.rules.concat(newConfig.module.loaders);
+                delete newConfig.module.loaders;
+            }
```
jbeezley: I tried using the [old webpack helper](https://github.com/girder/girder/blob/master/plugins/dicom_viewer/webpack.helper.js) for the dicom_plugin hit the following error:
```javascript
  >> Loading webpack helper from /home/jbeezley/test/girder/plugins/dicom_viewer/webpack.helper.js
Loading "plugin.js" tasks...ERROR
>> TypeError: Cannot read property 'push' of undefined
```

Could we define `loaders` as an empty array in the config passed to this function?

grunt_tasks/webpack.config.js
```diff
@@ -72,52 +72,78 @@ module.exports = {
...
-                test: /\.json$/,
-                include: loaderPaths.concat(loaderPathsNodeModules),
-                loader: 'json-loader'
+                resource: {
+                    test: /\.json$/,
+                    include: loaderPaths.concat(loaderPathsNodeModules)
+                },
+                use: [
+                    'json-loader'
+                ]
```
jbeezley: According to https://webpack.js.org/guides/migrating/#json-loader-is-not-required-anymore this shouldn't be necessary any more.

brianhelba:
@jbeezley Good comments, thanks. Fixed both of your issues in new commits on the end. PTAL.

PR #1990, opened 2017-05-09
Item tasks demo and tests

jbeezley:
This adds a new `item_tasks` container that serves to demonstrate the different widgets available.  It is also used to generate an integration test case.  While working on this, I fixed a number of bugs and inconsistencies in the UI:

1. Integer types now display as a numeric input
2. Fixed the height on select boxes so that descending elements in the text is no longer cut off
3. Limiting the size of a multiselect box to the number of values
4. Added linting tests to the item task testing specs
5. Fixed a small offset in the range slider tooltip

### DIFF
--- a/plugins/item_tasks/web_client/models/WidgetModel.js
+++ b/plugins/item_tasks/web_client/models/WidgetModel.js
@@ -102,7 +102,7 @@ var WidgetModel = Backbone.Model.extend({
         }
 
         /*
-         * Integers are special numeric types where adjecent values differ
+         * Integers are special numeric types where adjacent values differ
          * by exactly 1.  Setting the "step" field to one, ensures that
          * clicking the input element arrows increment or decrement the
          * value by one.
--- a/plugins/item_tasks/web_client/models/WidgetModel.js
+++ b/plugins/item_tasks/web_client/models/WidgetModel.js
@@ -101,6 +101,9 @@ var WidgetModel = Backbone.Model.extend({
             attrs.value = attrs.default.data;
         }
 
+        if (attrs.type === 'integer') {
+            attrs.step = 1;
+        }
         this.set(attrs);
     },
 
@@ -208,8 +211,6 @@ var WidgetModel = Backbone.Model.extend({
         var out;
         if (this.isNumeric()) {
             out = this._validateNumeric(value);
-        } else if (this.isInteger()) {
-            out = this._validateInteger(value);
         }
         if (this.isEnumeration() && !_.contains(this.get('values'), this._normalizeValue(value))) {
             out = 'Invalid value choice';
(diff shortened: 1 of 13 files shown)


plugins/item_tasks/web_client/models/WidgetModel.js
```diff
@@ -101,6 +101,9 @@ var WidgetModel = Backbone.Model.extend({
             attrs.value = attrs.default.data;
         }
 
+        if (attrs.type === 'integer') {
```
mgrauer: can you add a comment here explaining what this is doing?

mgrauer:
This is nice, and I really like the demo Dockerfile which enumerates (all?) of the UI possibilities and provides a quick way to test end-to-end values, from UI, to Docker, back to the Job.

Some questions/suggestions, these might already be baked into item_tasks or maybe out of scope for this PR.

Here's the `file_input` type of input from the demo task.  What is confusing to me is that this is actually selecting an item, instead of an individual file.  Can you change the name of the type to item_input instead of file_input?  If not, can you change the name and the description here so that the example task won't be misleading--at least when users see the demo task the UI would say item.  If you could do that plus change the hint text in the UI to say "choose an item" instead of "choose a file" this would help without having to change the name of `file_input`, but then of course you have a disconnect between the actual type name and what the demo task text says.  Same thing goes for the outputs.

```
{
         "id": "file_input",
         "name": "File input",
         "description": "Choose an existing file that will be used by the task",
         "type": "file",
         "target": "filepath"
}
```

After selecting an input and output file, you get to see their name in the text box.  Would it be possible to show their full path?
screen shot 2017-05-10 at 6 21 22 pm

jbeezley:
It would make more sense to select files rather than items, but the hierarchy widget used doesn't support showing files.  In retrospect, it would probably be better to call it an "item" type, and add a "file" type later when we extend the file chooser interface.  We would have to get a consensus on that due to the compatibility implications.  For now, I can change the text in the example.

In regards to showing the full path, I agree, but it would fit in better with a future PR.  One of my plans is to add first class support for default paths, which would likely mean supporting path strings from the `/resource/lookup` and `/resource/:id/path` endpoints.  With that change, it would be straightforward to display the full path.

manthey:
With regard to using resource paths, we currently can have two files with identical names in the same item, which means there is no way to distinguish them with a resource path.

jbeezley:
Indeed, my plans aren't entirely fleshed out.  For defaults, it makes most sense to talk about outputs though.  I'm not sure the usefulness of a default input outside providing a default folder.  I suppose there might be ambiguity about folders being identically named, but at some point we have to tell users "don't do that, or things will break" otherwise I don't see how we set nontrivial default's that work across databases.

mgrauer:
The only thing I need for this PR to go in is to add a comment on the `attrs.step` part.

Otherwise, I'm fine with your judgement on whatever else you want to do here vs another PR.

plugins/item_tasks/web_client/models/WidgetModel.js
```diff
@@ -101,6 +101,23 @@ var WidgetModel = Backbone.Model.extend({
             attrs.value = attrs.default.data;
         }
 
+        /*
+         * Integers are special numeric types where adjecent values differ
```
mgrauer: adjacent

mgrauer:
spelling nit, otherwise looks good

msmolens:
The `web_client_item_tasks.tasks` test needs updates along the lines of:
https://github.com/girder/girder/pull/1996/commits/cb56787f348f52316772ac382e883fc3d21f4789
Or just update the following line to use `toContain`:
````
expect($('.g-execute-task-link').eq(0).text()).toBe('item_tasks widget types demo');
````

PR #1992, opened 2017-05-09
Fix web client leaks

### LINKED ISSUE #1530: Web client leaks Backbone views
On the web client front page, simply navigating between the Collections, Users, Group leaks Backbone views. This can be confirmed in Chrome by taking heap snapshots and examining 'Detached DOM tree' entries. [Backbone Debugger](https://github.com/Maluen/Backbone-Debugger) also shows the leaking objects.

Update: Backbone Debugger view list isn't compatible with how we manage views, so it shouldn't be trusted in this case.

msmolens:
Fix various DOM tree leaks in the web client.

Fixes #1530.

### DIFF
--- a/clients/web/src/dialog.js
+++ b/clients/web/src/dialog.js
@@ -74,7 +74,9 @@ function confirm(params) {
     }, params);
     $('#g-dialog-container').html(ConfirmDialogTemplate({
         params: params
-    })).girderModal(false);
+    })).girderModal(false).one('hidden.bs.modal', function () {
+        $('#g-confirm-button').off('click');
+    });
 
     var el = $('#g-dialog-container').find('.modal-body>p');
     if (params.escapedHtml) {
--- a/clients/web/src/views/body/AssetstoresView.js
+++ b/clients/web/src/views/body/AssetstoresView.js
@@ -44,6 +44,7 @@ var AssetstoresView = View.extend({
 
     initialize: function (settings) {
         cancelRestRequests('fetch');
+        this.plots = [];
         this.assetstoreEdit = settings.assetstoreEdit || false;
         this.importableTypes = [
             AssetstoreType.FILESYSTEM,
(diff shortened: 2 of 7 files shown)


clients/web/src/dialog.js
```diff
@@ -74,7 +74,9 @@ function confirm(params) {
     }, params);
     $('#g-dialog-container').html(ConfirmDialogTemplate({
         params: params
-    })).girderModal(false);
+    })).girderModal(false).one('hidden.bs.modal', function () {
```
zachmullen: Is `.one()` an actual function, or is that a typo?

zachmullen: To answer my own question, this is [a jquery function](http://api.jquery.com/one/), not a backbone function.

msmolens: Right:

> The .one() method is identical to .on(), except that the handler [...] is unbound after its first invocation.

clients/web/src/views/body/AssetstoresView.js
```diff
@@ -61,11 +62,24 @@ var AssetstoresView = View.extend({
...
     },
 
+    destroy: function () {
+        this._destroyPlots();
+        View.prototype.destroy.call(this);
+    },
+
+    _destroyPlots: function () {
+        for (let plot of this.plots) {
+            plot.data('jqplot').destroy();
```
zachmullen: So jqplots bind themselves to global scope in some way?

msmolens: Without this the views leave detached DOM trees with HTML canvas elements, created by jqplot, at the roots.
https://github.com/jqPlot/jqPlot/blob/460412a84e603a62eea0ecd90057cbaac3e2d9c8/src/jqplot.core.js#L2843

zachmullen: Ew, gross. It is a rather dated library.

zachmullen:
@msmolens I had tried investigating this some time ago and discovered one source of leaking was through tooltips, did you notice that as well?

msmolens:
@zachmullen Yes, in https://github.com/girder/girder/pull/1992/commits/b5d1c4c27b3bd740fad984c7a8a947a0c8300c26 it mentions this:

> DOM elements created by Bootstrap tooltips tended to exhibit this behavior

zachmullen:
Great, sorry I missed that message.

msmolens:
@zachmullen Thanks for the review.

zachmullen:
Thanks for the fixes! 🥂

PR #1995, opened 2017-05-10
Add ansible tests for item_tasks plugin

danlamanna:
@mgrauer 
This should exercise the ability of vagrant up/provision/destroy the
included Vagrantfile for the item tasks plugin. These tests are only
added if CMake is configured with ANSIBLE_TESTS=ON as they are long running
and shouldn't be run during CI.

### DIFF
--- a/plugins/item_tasks/devops/README.md
+++ b/plugins/item_tasks/devops/README.md
@@ -17,7 +17,7 @@ Logs for girder-worker go to /var/log/upstart/girder_worker.log.
  
 # Basic operations
 
-These commands should be run from the same director as this README file,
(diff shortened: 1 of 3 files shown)


mgrauer:
I had to change the network adapter type to `virtio` for better [performance](https://www.virtualbox.org/manual/ch06.html#network_performance), otherwise the provision test kept failing due to timeouts.

I noticed that installing girder role's [virtualenv](https://github.com/girder/girder/blob/master/devops/ansible/roles/girder/tasks/main.yml#L29-L37) takes weirdly long.

@danlamanna I approve your changes, but I've added commits that need QA.

PR #1996, opened 2017-05-10
Improve presentation of item_tasks task list

msmolens:
image

### DIFF
--- a/plugins/item_tasks/plugin_tests/tasks.js
+++ b/plugins/item_tasks/plugin_tests/tasks.js
@@ -89,7 +89,10 @@ describe('Run the item task', function () {
 
         runs(function () {
             expect($('.g-execute-task-link').length).toBe(1);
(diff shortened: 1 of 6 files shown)


jeffbaumes:
This is great, but there doesn't seem to be pagination, even if there are 4 screens of analyses (I reloaded pysciencedock into several folders). Is the pagination set to a large number per page?

msmolens:
Indeed, the default page size for ItemCollection is 100: https://github.com/girder/girder/blob/36970b8afe3fbf0211d86347ac4c2a58400dbcf4/clients/web/src/collections/ItemCollection.js#L8.

I'll improve this behavior.

jeffbaumes:
Looks way better, though I'd maybe increase the page size to something like 20. Seems like you should be sure to fill screens on larger monitors so there is no blank space.

msmolens:
I increased the page size to 20. If users need more flexibility we can always expose this parameter in the UI.

PR #1998, opened 2017-05-12
Fix item_tasks client test

jbeezley:
As pointed out by @msmolens, this should fix the failure on master.

### DIFF
--- a/plugins/item_tasks/plugin.cmake
+++ b/plugins/item_tasks/plugin.cmake
@@ -13,6 +13,10 @@ add_web_client_test(
     tasks "${PROJECT_SOURCE_DIR}/plugins/${PLUGIN}/plugin_tests/tasks.js" PLUGIN ${PLUGIN}
     SETUP_MODULES "${CMAKE_CURRENT_LIST_DIR}/plugin_tests/mock_worker.py")
 
+if (BUILD_JAVASCRIPT_TESTS)
+    set_property(TEST web_client_item_tasks.tasks PROPERTY RUN_SERIAL ON)
+endif()
+
 add_eslint_test(${PLUGIN} "${PROJECT_SOURCE_DIR}/plugins/${PLUGIN}/web_client")
 add_eslint_test(${PLUGIN}-tests "${PROJECT_SOURCE_DIR}/plugins/${PLUGIN}/plugin_tests")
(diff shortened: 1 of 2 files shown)


jbeezley:
I'm getting this error on circle, but not locally.  Not sure what changed since my PR.  Does the timeout need to be increased?

```
Resource timed out.  (#20): {"errorCode":408,"errorString":"Network timeout on resource.","headers":[{"name":"Accept","value":"text/event-stream"},{"name":"Cache-Control","value":"no-cache"},{"name":"Referer","value":"http://localhost:30031/static/built/testing/testEnv.html"},{"name":"User-Agent","value":"Mozilla/5.0 (Unknown; Linux x86_64) AppleWebKit/538.1 (KHTML, like Gecko) PhantomJS/2.1.1 Safari/538.1"}],"id":20,"method":"GET","time":"2017-05-12T13:19:36.043Z","url":"http://localhost:30031/api/v1/notification/stream?timeout=10"}
PHANTOM_TIMEOUT
```

jbeezley:
@msmolens I couldn't merge until the CI passed.  I think I have it working now.  Want to have another look?

plugins/item_tasks/plugin.cmake
```diff
@@ -13,6 +13,10 @@ add_web_client_test(
     tasks "${PROJECT_SOURCE_DIR}/plugins/${PLUGIN}/plugin_tests/tasks.js" PLUGIN ${PLUGIN}
     SETUP_MODULES "${CMAKE_CURRENT_LIST_DIR}/plugin_tests/mock_worker.py")
 
+if (BUILD_JAVASCRIPT_TESTS)
+    set_property(TEST web_client_item_tasks.tasks PROPERTY RUN_SERIAL ON)
+endif()
+
```
msmolens: Is this change still necessary?

jbeezley: I think so, that's how I fixed the event stream timeout.  It doesn't occur (as far as I can tell) when running serially.  There were a few other occasional race conditions that I fixed with the other commits.

msmolens: In that case, can you add a comment that we run this serially to avoid an event stream timeout?

PR #2000, opened 2017-05-12
WIP: Enable text search for tasks on item_tasks task list

msmolens:


plugins/item_tasks/server/rest.py
```diff
@@ -40,6 +41,30 @@ def listTasks(self, limit, offset, sort, params):
...
+        Description('Search for tasks that can be executed in the system.')
+        .param('q', 'The search query.')
+        .pagingParams(defaultSort=None, defaultLimit=20)
+    )
+    def search(self, q, limit, offset, params):
+        user = self.getCurrentUser()
+        itemModel = self.model('item')
+
+        filters = {
+            'meta.isItemTask': {'$exists': True}
```
jeffbaumes: It seems that this should check that it has a True value, and not just that the flag exists. That is, if isItemTask is set to False, it shouldn't be interpreted as a task.

msmolens: This plugin ignores the value of `isItemTask`, probably as an optimization. This is actually documented: https://girder.readthedocs.io/en/latest/plugins.html#item-tasks.

Note that this same query is performed in `listTasks()`.

jeffbaumes: Good enough for me to leave as-is, since we would need to update the docs and listTasks function to change it at this point.

msmolens:
Superseded by #2004.

PR #2001, opened 2017-05-15
Wait for bootstrap transitions.

manthey:
We use bootstrap transitions on dialogs.  These are meant to trigger the appropriate browser event, which, if it doesn't fire in the expected amount of time, also uses a window timeout as a fallback to trigger the event.

Unfortunately, there isn't currently any official way in Bootstrap 3 to know if you are in a transition state, and some of our tests rely on the previous activities being completed before starting the next activities.   Bootstrap 4 is adding an internal flag to determine if it is in a transition, but it has not been released yet.

This adds our own testing event for transition end, ignoring the browser's css transitionend event.  It allows a flag to specify that we are in a transition, and therefore allows waitForDialog and waitForLoad to detect if we are still in a transition.

### DIFF
--- a/clients/web/test/testUtils.js
+++ b/clients/web/test/testUtils.js
@@ -588,6 +588,9 @@ girderTest.waitForLoad = function (desc) {
         return !$('.modal').data('bs.modal').$backdrop;
     }, 'any modal dialog to be hidden' + desc);
     waitsFor(function () {
+        return !girder._inTransition;
+    }, 'transitions to finish');
+    waitsFor(function () {
         return girder.rest.numberOutstandingRestRequests() === 0;
     }, 'rest requests to finish' + desc);
     waitsFor(function () {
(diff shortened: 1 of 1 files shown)


manthey:
It is possible (though completely unproven) that this stop the routing tests from failing arbitrarily.  It definitely stops the large_image plugin's geojs tests from failing much of the time.

clients/web/test/testUtils.js
```diff
@@ -1219,6 +1225,25 @@ $(function () {
 girderTest.startApp = function () {
     var defer = new $.Deferred();
     girderTest.promise.then(function () {
+        /* Track bootstrap transitions using our own test event. */
+        $.support.transition = {end: 'girdertest_transitionend'};
```
jbeezley: It's possible I'm not understanding this correctly, but looking at what bootstrap does with this, I think it is expecting this to be a browser specific event name.  I don't think you should be changing it.

clients/web/test/testUtils.js
```diff
@@ -1219,6 +1225,25 @@ $(function () {
...
+        $.fn.emulateTransitionEnd = function (duration) {
+            girder._inTransition = true;
+            var $el = this;
+            window.setTimeout(function () {
+                girder._inTransition = false;
+                $($el).trigger($.support.transition.end);
+            }, duration);
+            return this;
+        };
```
jbeezley: The original source of this code is described here: https://blog.alexmaccaw.com/css-transitions  While I think this works for our purposes, it does break the original intent of the code that the browser specific event is *always* triggered after a transition.  I wonder if this has side effects elsewhere.

I wonder if a better approach would be to bind to `$.support.transition.end` setting `girder._inTransition = false` and to wrap this function setting `girder._inTransition = true` before calling the original.

PR #2003, opened 2017-05-16
Pull image option on slicer CLI task config

zachmullen:
@cdeepakroy PTAL, this allows pulling the image to be optional when configuring a slicer CLI XML item task.

screen shot 2017-05-16 at 1 29 48 pm

### DIFF
--- a/plugins/item_tasks/server/rest.py
+++ b/plugins/item_tasks/server/rest.py
@@ -2,7 +2,6 @@ import json
 import six
 
 from girder import events
-from girder import logger
 from girder.api import access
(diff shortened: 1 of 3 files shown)


mgrauer:
Does `Docker Hub` mean the same thing as `docker registry`?  We should standardize our usage.

kotfic:
late to the party but :clap: for screenshot!

zachmullen:
Good point @mgrauer , I copied that comment from the other endpoint. Docker Hub is the one we should use.

PR #2004, opened 2017-05-16
WIP: Item tasks search and tags

msmolens:
Supersedes #2000.

Enhancements to item_tasks task list:
- Add search box to search for tasks
  - Search results are displayed in a separate view
- Item tasks support a string array of tags in the `meta.itemTaskSpec.tags` field
  - The task list page shows the tags
  - Clicking a tag navigates to a view that lists all tasks with that tag
  - That view supports filtering by multiple tags, but currently there is no user interface to select multiple tags
  - Future enhancements could include a view listing all tags for tasks executable by the current user

@mgrauer @zachmullen @jeffbaumes Before I continue, please let me know what parts of this look useful.

### DIFF
--- a/plugins/item_tasks/plugin_tests/tasks_test.py
+++ b/plugins/item_tasks/plugin_tests/tasks_test.py
@@ -87,6 +87,14 @@ class TasksTest(base.TestCase):
         self.assertStatusOk(resp)
         self.assertEqual(len(resp.json), 2)
 
+        resp = self.request('/item_task/search', params={'q': 'Task'})
+        self.assertStatusOk(resp)
+        self.assertEqual(resp.json, [])
+
+        resp = self.request('/item_task/search', user=self.admin, params={'q': 'Task'})
+        self.assertStatusOk(resp)
+        self.assertEqual(len(resp.json), 2)
+
         # Test adding single task spec
         folder2 = self.model('folder').createFolder(
             name='placeholder2', creator=self.admin, parent=self.admin, parentType='user')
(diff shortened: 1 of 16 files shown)


kotfic:
Just a general comment: We may want to consider a system for tag completion to ensure a consistent tag vocabulary - a feature like that could be really useful across more than just item_tasks.

jeffbaumes:
I agree with @kotfic that tags could be generalized, possibly into a special model field (with its own DB indexing and UI for showing, editing tags) and not in the metadata object. Even though this is not needed now, I could see tags being needed on items, folders, etc. in the future.

That being said, I can see a path from your current approach to one that uses generalized tagging support when/if it's available.

mgrauer:
@msmolens After some discussion, this would be useful to get merged in, in a specific implementation that could be generalized later, assuming this isn't a lot of work, say less than 20 hours.  I can talk with you over email about it if it seems like more work than that or for other details.

jeffbaumes:
I've recently talked to @sutartmelson about adding an additional way to search tasks (by number of file inputs). Is this implemented as additional optional flags the item_tasks GET endpoint? That is how I anticipated searching to be implemented.

brianhelba:
Is there a reason why this couldn't be implemented via the `resource/search` endpoint (adding new features there, if necessary)?

msmolens:
In its current form, `resource/search` is designed to be extensible for new model types. Here, we search a built-in model type (`item`) but need to apply restrictions:
- The collection query must include `meta.isItemTask` to restrict to "task" items
- `filterResultsByPermission` must apply the `ACCESS_FLAG_EXECUTE_TASK` access flag

For the first point, I don't think we'd want to allow adding arbitrary query parameters to the search endpoint. For the second point, the search endpoint, and the model search methods, could be extended to take a list of flags.

zachmullen:
@jbeezley has an upcoming need for this and plans to take it over.

jbeezley:
For anyone watching this, I have resolved most of the conflicts and fixed the tests, but UI is still broken.  I'll make a note when it's ready for review.

zachmullen:
The web client code in this will be superseded before this can be merged. The server-side code may be superseded by more general tagging and searching features exposed via core.

PR #2005, opened 2017-05-17
Docker registry -> Docker Hub

zachmullen:
@mgrauer PTAL


mgrauer:
Sorry, didn't see this ping before.  

Can you also change https://github.com/girder/girder/blob/master/plugins/item_tasks/web_client/templates/configureTasks.pug#L22?

zachmullen:
Done

PR #2006, opened 2017-05-17
Dynamic linking between dependent plugins

zachmullen:
At long last, we have it! If plugin A explicitly lists plugin B in its runtime dependencies, any code imported from plugin B into plugin A will no longer be copied, but shared at runtime. This means you can augment symbols from one plugin inside a downstream plugin, as shown here in the example of modifying the JobStatus object.

### DIFF
--- a/girder/utility/install.py
+++ b/girder/utility/install.py
@@ -69,7 +69,8 @@ def fix_path(path):
 
 def _getPluginBuildArgs(buildAll, plugins):
     if buildAll:
-        return ['--all-plugins']
+        sortedPlugins = plugin_utilities.getToposortedPlugins()
+        return ['--plugins=%s' % ','.join(sortedPlugins)]
     elif not plugins:  # build only the enabled plugins
         settings = model_importer.ModelImporter().model('setting')
         plugins = settings.get(constants.SettingKey.PLUGINS_ENABLED, default=())
--- a/girder/utility/plugin_utilities.py
+++ b/girder/utility/plugin_utilities.py
@@ -109,11 +109,14 @@ def loadPlugins(plugins, root, appconf, apiRoot=None, buildDag=True):
     return root, appconf, apiRoot
 
 
-def getToposortedPlugins(plugins, ignoreMissing=False, keys=('dependencies',)):
+def getToposortedPlugins(plugins=None, ignoreMissing=False, keys=('dependencies',)):
     """
     Given a set of plugins to load, construct the full DAG of required plugins
     to load and yields them in toposorted order.
 
+    :param plugins: A set of plugins that must be in the output list. If you want
+        to toposort all available plugins, omit this parameter.
+    :type plugins: iterable of str
     :param ignoreMissing: Normally if one of the plugins specified does not exist,
         this raises a ValidationError. Set this to False to suppress that and instead
         print an error message and continue.
@@ -121,11 +124,14 @@ def getToposortedPlugins(plugins, ignoreMissing=False, keys=('dependencies',)):
     :param keys: Keys that should be used to determine dependencies.
     :type keys: list of str
     """
-    plugins = set(plugins)
-
     allPlugins = findAllPlugins()
     dag = {}
     visited = set()
+
+    if plugins is None:
+        plugins = set(allPlugins.keys())
+    else:
+        plugins = set(plugins)
 
     def addDeps(plugin):
         if plugin not in allPlugins:
--- a/grunt_tasks/build.js
+++ b/grunt_tasks/build.js
@@ -96,7 +96,6 @@ module.exports = function (grunt) {
                 },
                 output: {
                     library: '[name]'
-
                 },
                 plugins: [
                     // Remove this if it turns out we don't want to use it for every bundle target.
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -32,8 +32,8 @@ module.exports = function (grunt) {
     var plugins = grunt.option('plugins');
 
     if (grunt.option('all-plugins')) {
-        throw new Error(
-            'The --all-plugins option no longer works, use `girder-install web --all-plugins` instead.')
+        grunt.fail.warn(
+            'The --all-plugins option no longer works, use `girder-install web --all-plugins` instead.');
     }
 
     if (_.isString(plugins) && plugins) {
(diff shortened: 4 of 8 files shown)


jbeezley:
I think we need to ensure the plugins are built in the correct order. I tried running the build from a new clone, and I got:
```
...
Running "webpack:core_lib" (webpack) task                                                                                                 10% building modules(node:626) DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaceVersion: webpack 2.2.1                                                                                                            
                                             Asset      Size  Chunks                    Chunk Names
              assets/jsoneditor-icons-bf23347e.png   14.4 kB          [emitted]         
  assets/glyphicons-halflings-regular-f4769f9b.eot   20.1 kB          [emitted]         
  assets/glyphicons-halflings-regular-89889688.svg    109 kB          [emitted]         
 assets/glyphicons-halflings-regular-fa277232.woff   23.4 kB          [emitted]         
assets/glyphicons-halflings-regular-448c34a5.woff2     18 kB          [emitted]         
  assets/glyphicons-halflings-regular-e18bbf61.ttf   45.4 kB          [emitted]         
                                 girder_lib.min.js   4.17 MB       0  [emitted]  [big]  girder_lib
                                girder_lib.min.css    229 kB       0  [emitted]         girder_lib
                             girder_lib.min.js.map   5.65 MB       0  [emitted]         girder_lib
                            girder_lib.min.css.map  95 bytes       0  [emitted]         girder_lib

Running "webpack:core_app" (Version: webpack 2.2.1                                                                                                            
                Asset     Size  Chunks             Chunk Names
    girder_app.min.js  4.75 kB       0  [emitted]  girder_app
girder_app.min.js.map  5.31 kB       0  [emitted]  girder_app

Running "test-env-html" task

Running "webpack:plugin_authVersion: webpack 2.2.1                                                                                                            
             Asset      Size  Chunks             Chunk Names
     plugin.min.js   24.1 kB       0  [emitted]  plugins/authorized_upload/plugin
    plugin.min.css   1.93 kB       0  [emitted]  plugins/authorized_upload/plugin
 plugin.min.js.map   37.3 kB       0  [emitted]  plugins/authorized_upload/plugin
plugin.min.css.map  91 bytes       0  [emitted]  plugins/authorized_upload/plugin

Running "webpack:plugin_autoVersion: webpack 2.2.1                                                                                                            
             Asset       Size  Chunks             Chunk Names
     plugin.min.js    29.6 kB       0  [emitted]  plugins/autojoin/plugin
    plugin.min.css  145 bytes       0  [emitted]  plugins/autojoin/plugin
 plugin.min.js.map    48.3 kB       0  [emitted]  plugins/autojoin/plugin
plugin.min.css.map   91 bytes       0  [emitted]  plugins/autojoin/plugin

Running "webpack:plugin_candVersion: webpack 2.2.1                                                                                                            
                               Asset       Size  Chunks                    Chunk Names
           assets/alpha-5278808a.png  488 bytes          [emitted]         
assets/alpha-horizontal-1b3c0ada.png  557 bytes          [emitted]         
  assets/hue-horizontal-34a3768e.png  478 bytes          [emitted]         
             assets/hue-e75b2a9c.png  504 bytes          [emitted]         
      assets/saturation-668d0a0e.png    4.14 kB          [emitted]         
                       plugin.min.js    10.5 MB       0  [emitted]  [big]  plugins/candela/plugin
                      plugin.min.css    14.7 kB       0  [emitted]         plugins/candela/plugin
                   plugin.min.js.map    12.8 MB       0  [emitted]         plugins/candela/plugin
                  plugin.min.css.map   91 bytes       0  [emitted]         plugins/candela/plugin

Running "webpack:plugin_celery_jobs" (webpack) task
Warning: ENOENT: no such file or directory, open '/home/jbeezley/test/girder/clients/web/static/built/plugins/jobs/plugin-manifest.json' Use --force to continue.
```

zachmullen:
Was that using `--all-plugins`?

jbeezley:
Yes

zachmullen:
I see two options here:

1. Add toposorting logic into the js build code
2. Remove the buildAll logic from the js build code and make `girder-install web --all-plugins` toposort all the plugins and pass them down as a list.

#2 seems easier since we already have the sorting logic readily available, but it would break an undocumented interface (which is non-API-breaking, but it wouldn't surprise me if somebody out there is still using it).

jbeezley:
2. would just break `npm run build --all-plugins` correct?  My vote would be to do that rather than implement the plugin dependency logic in two places.

zachmullen:
That's correct.

zachmullen:
Strangely, raising an Error during `grunt.loadTasks` does not actually cause grunt to fail, it still pushes through and exits with 0. Not sure how to make it fail, the docs don't say.

zachmullen:
Nevermind, figured it out, just have to use `grunt.fail.warn` instead

jbeezley:
Inside the item_tasks plugin bundle, the `JobModel` is still defined.  I suspect the reason is because it doesn't list jobs as an explicit dependency, rather it includes it through the worker plugin.  I imagine trying to resolve the full dependency tree would mean duplicating the plugin toposorting code.  I'm inclined to think that plugins should list as dependencies any plugin they import directly.  In this case, item_tasks should be modified to depend on jobs.

Are there any other cases of this behavior elsewhere in the code?

zachmullen:
> I'm inclined to think that plugins should list as dependencies any plugin they import directly.

Yep, that's exactly the policy we should adopt. I'll add jobs to the item_tasks dependencies.

I couldn't find any other occurrences of this stuff in built-in plugins.

zachmullen:
Thanks for the review @jbeezley

mgrauer:
Is this a breaking change?  I think we have a few non-built in plugins that depend on worker but not jobs directly.  Will they still build, and just get the old behavior of copying?

jbeezley:
No, it shouldn't break anything.  If you don't explicitly list something as a dependency, it will copy the module into the bundle as it did before.  You just don't get the benefit of this change.

mgrauer:
In that case, 🥇

manthey:
I see an error in a travis build on the large_image plugin (https://travis-ci.org/girder/large_image/jobs/232438328) when it runs `girder-install web --plugins=large_image,worker,jobs --dev`.  There is an error:  `Warning: ENOENT: no such file or directory, open '/home/travis/build/girder/large_image/build/girder/clients/web/static/built/plugins/worker/plugin-manifest.json'`.

Is the failure related to this change, or just an unrelated issue?

zachmullen:
Yes... if you list the plugins out explicitly, they are built in the order listed. The fix would be to invert the order of the listed plugins.

zachmullen:
We could also fix this by having `girder-install web` reorder the passed in plugin list. That's probably the right way to go.

manthey:
Yes, we should.  Since otherwise, this *is* a breaking change.

PR #2008, opened 2017-05-17
Set private item background color to yellow

### LINKED ISSUE #1899: Show yellow background for items in private folders
Currently, sub-folders show as yellow in folder listings, but items in private folders are not shaded yellow. For example, this is the listing for the user's Private folder, which shows a private sub-folder but items are not marked as private.

image

dorukozturk:
Fixes #1899

### DIFF
--- a/clients/web/src/templates/widgets/itemList.pug
+++ b/clients/web/src/templates/widgets/itemList.pug
@@ -1,6 +1,6 @@
 ul.g-item-list
   each item in items
-    li.g-item-list-entry
+    li.g-item-list-entry(public=(isParentPublic ? 'true' : 'false'))
       if checkboxes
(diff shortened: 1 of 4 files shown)


zachmullen:
I'm encountering a bug with this.

1. Navigate to a private folder. Items correctly show up yellow.
2. Use the breadcrumb link to navigate to the parent folder.
3. Click a public folder inside the parent folder. The items show up as yellow even though they are public. Refreshing the page fixes the issue.

dorukozturk:
@zachmullen Thank you very much for the review. I will try to reproduce that ASAP.

dorukozturk:
@zachmullen I was able to reproduce and fix the bug your were encountering. The view was not getting initialized before the render method was called which was causing attributes to be 'undefined'. Then undefined was evaluating to public = 'false', which was changing the background color to yellow.

PR #2009, opened 2017-05-17
Ingest item_task Docker images as part of Vagrant VM construction

mgrauer:
This adds to the item_tasks Vagrant/Ansible to ingest Docker images as part of the construction of the VM, then waits for those Docker images to be fully ingested.  At the current time it is not working for local Docker images, but does work for images on DockerHub.

To test this, after following the instructions in the README on this branch to build the local demo Docker image on the VM, you can add the following variables to the site.yml near the top where other girder and girder_worker vars are set.

```
    item_tasks_local_images:
      - item-tasks-demo
    item_tasks_dockerhub_images:
      - kitware/pysciencedock
```

### DIFF
--- a/plugins/item_tasks/devops/README.md
+++ b/plugins/item_tasks/devops/README.md
@@ -15,8 +15,23 @@ The Girder admin user is girder/girder.
 Logs for Girder go to /var/log/upstart/girder.log.
 Logs for girder-worker go to /var/log/upstart/girder_worker.log.
 
-To ingest dockerhub docker images, set the variable `item_tasks_images` with
+To ingest dockerhub docker images, set the variable `item_tasks_dockerhub_images` with
 a list of dockerhub image paths.
+
+To ingest locally built docker images, set the variable `item_tasks_local_images` with
+a list of local image paths.
+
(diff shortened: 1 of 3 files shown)


mgrauer:
@danlamanna was able to get this to work with local docker images, but I haven't yet.

In thinking more about it, I'm going to back out the local docker images part because it doesn't work for me, but more importantly, I don't provide support through Ansible to create the local image on the VM in the first place.

danlamanna:
@mgrauer I added a commit which stops it from flat out failing when `item_tasks_dockerhub_images` isn't set, but this should at least work with `kitware/pysciencedock` before we merge it. (It doesn't for me currently)

zachmullen:
Has merge conflicts in some documentation, @mgrauer could you fix and merge this PR when you get a chance?

danlamanna:
@zachmullen Did you test this worked with at least one dockerhub image?
Otherwise I think it should still be blocked.

mgrauer:
I'm closing this one unmerged.  It only worked for Dockerhub images, not local images, and it didn't even work in the Dockerhub case for Dan.  It isn't that useful of an addition to the item_tasks playbook.

Issue #1999, opened 2017-05-12
Uploading and deleting identical files in a filesystem assetstore can result in a missing file


manthey:
If you have a file in a filesystem assetstore and you upload an identical file (the same sha512 hash), if the original file is deleted while the upload is being finalized, the file will be deleted, but not copied from the temp upload location to the assetstore.

There is only a small window where this can occur (unless an event handler on `model.file.finalizeUpload.before`, `model.file.save`, etc. takes a long time).

manthey:
Fixed in PR #2010.

PR #2010, opened 2017-05-17
Prevent deleting a file that is being uploaded.

manthey:
If an upload occurs where the file has the same hash as an existing file, we don't duplicate the file.  There was a window of time where if a delete occurred during the upload, we would detect that the file already existed, but then allow the delete to occur.  This fixes this by checking to see if there are any pending uploads with the file and using a lock to ensure that the check and deletes cannot conflict with one another.

This fixes issue #1999.

### DIFF
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -37,6 +37,7 @@ class Upload(Model):
     """
     def initialize(self):
         self.name = 'upload'
+        self.ensureIndices(['sha512'])
 
     def _getChunkSize(self, minSize=32 * 1024**2):
         """
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -37,7 +37,7 @@ class Upload(Model):
     """
     def initialize(self):
         self.name = 'upload'
-        self.ensureIndices(['sha512'])
+        self.ensureIndex('sha512')
 
     def _getChunkSize(self, minSize=32 * 1024**2):
         """
--- a/girder/utility/filesystem_assetstore_adapter.py
+++ b/girder/utility/filesystem_assetstore_adapter.py
@@ -109,7 +109,6 @@ class FilesystemAssetstoreAdapter(AbstractAssetstoreAdapter):
         self.tempDir = os.path.join(self.assetstore['root'], 'temp')
         # Use a filelock at the root level of the assetstore; this should work
         # between multiple servers.
-        self.deleteLock = filelock.FileLock(os.path.join(self.assetstore['root'], '_deleteLock'))
         try:
             mkdir(self.tempDir)
         except OSError:
@@ -230,7 +229,7 @@ class FilesystemAssetstoreAdapter(AbstractAssetstoreAdapter):
         # Only maintain the lock which checking if the file exists.  The only
         # other place the lock is used is checking if an upload task has
         # reserved the file, so this is sufficient.
-        with self.deleteLock:
+        with filelock.FileLock(abspath + '.deleteLock'):
             pathExists = os.path.exists(abspath)
         if pathExists:
             # Already have this file stored, just delete temp file.
@@ -317,7 +316,7 @@ class FilesystemAssetstoreAdapter(AbstractAssetstoreAdapter):
         if matching.count(True) == 1:
             path = os.path.join(self.assetstore['root'], file['path'])
             if os.path.isfile(path):
-                with self.deleteLock:
+                with filelock.FileLock(path + '.deleteLock'):
                     if self.model('upload').findOne(q) is None:
                         try:
                             os.unlink(path)
--- a/girder/utility/filesystem_assetstore_adapter.py
+++ b/girder/utility/filesystem_assetstore_adapter.py
@@ -25,6 +25,7 @@ import six
 from six import BytesIO
 import stat
 import tempfile
+import threading
 
 from girder import events, logger
 from girder.api.rest import setResponseHeader
@@ -106,6 +107,7 @@ class FilesystemAssetstoreAdapter(AbstractAssetstoreAdapter):
         # happens to existing assetstores that no longer can access their temp
         # directories.
         self.tempDir = os.path.join(self.assetstore['root'], 'temp')
+        self.deleteLock = threading.Lock()
         try:
             mkdir(self.tempDir)
         except OSError:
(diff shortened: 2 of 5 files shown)


girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -216,11 +218,21 @@ def finalizeUpload(self, upload, file):
...
-        if os.path.exists(abspath):
+        # Only maintain the lock which checking if the file exists.  The only
+        # other place the lock is used is checking if an upload task has
+        # reserved the file, so this is sufficient.
+        with self.deleteLock:
+            pathExists = os.path.exists(abspath)
+        if pathExists:
             # Already have this file stored, just delete temp file.
-            os.remove(upload['tempFile'])
+            os.unlink(upload['tempFile'])
```
zachmullen: Why change from remove to unlink?

zachmullen:
> This fixes this by checking to see if there are any pending uploads with the file

What do you mean uploads with the file?

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -106,6 +107,7 @@ def __init__(self, assetstore):
         # happens to existing assetstores that no longer can access their temp
         # directories.
         self.tempDir = os.path.join(self.assetstore['root'], 'temp')
+        self.deleteLock = threading.Lock()
```
zachmullen: A new assestore adapter instance is created in each request thread, but it looks like this lock only protects against concurrent access on the same adapter instance.

My understanding of the race condition was that it would only happen across multiple request threads, is that not the case?

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -216,11 +218,21 @@ def finalizeUpload(self, upload, file):
         path = os.path.join(dir, hash)
         abspath = os.path.join(self.assetstore['root'], path)
 
+        # Store the hash in the upload so that deleting a file won't delete
+        # this file
+        upload['sha512'] = hash
+        upload = self.model('upload').save(upload, validate=False, triggerEvents=False)
```
zachmullen: This might be better done with `update` rather than `save`.

girder/utility/filesystem_assetstore_adapter.py
manthey: We use `unlink` in some spots in Girder and `remove` in others.  I think we should be consistent, even though they are the same thing.  It looks like we mildly prefer `unlink`, so I figured I would change this as a move for more consistency.

girder/utility/filesystem_assetstore_adapter.py
manthey: Somehow I thought we were creating singletons of the adapters for each assetstore (I don't know why I thought this, since it is wrong).

manthey:
By uploads with the file, I meant uploads that have a matching file hash and therefore are using a physical file in the assetstore (this is the usually brief period of time when a physical file is completely uploaded but the bookkeeping hasn't finished creating a Girder file).

girder/utility/filesystem_assetstore_adapter.py
zachmullen: Hm, this should still help shorten the critical section, but we can remove all the lock stuff.

In general, we can't use this locking primitive to assume we are locked across threads, since we have to assume there could be many app servers.

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -303,7 +315,12 @@ def deleteFile(self, file):
         if matching.count(True) == 1:
             path = os.path.join(self.assetstore['root'], file['path'])
             if os.path.isfile(path):
-                os.remove(path)
+                with self.deleteLock:
+                    if self.model('upload').findOne(q) is None:
```
zachmullen: Will this query use an index?

manthey: We can add an index, but is it necessary?  Usually uploads are relatively transient, so we don't typically have a huge number in the system at a time.  Obviously, you can create a scenario where this is otherwise.

zachmullen: Fair point. The number of uploads in practice will increase over time since we don't have a built-in process for cleaning up partial ones automatically, but we do have an endpoint that does so.

girder/utility/filesystem_assetstore_adapter.py
manthey: We can use a file lock, instead.  This is heavier, of course. 

Do we have deployed instances with multiple app servers?  Do we test for other resource contention issues?

zachmullen: > We can use a file lock, instead. This is heavier, of course.

Probably the best idea, since that uses the resource that is explicitly shared across all servers, and we have atomic primitives to use with the filesystem.

> Do we have deployed instances with multiple app servers? Do we test for other resource contention issues?

External users have deployed on elastic beanstalk with multiple app servers, but we do not test these sorts of things.

zachmullen:
Should we have a timeout on the lock acquisition? I'm trying to consider the catastrophic case where the server dies cold while the file lock is held.

We could also implement per-blob locks rather than a global one, i.e. a locks directory that contains multiple file locks, each named by the sha512.

manthey:
Nicely, the file lock is opened by maintaining an open file handle, so it will not survive the owner process termination.  This works properly on both linux-like environments and Windows.

girder/models/upload.py
```diff
@@ -37,6 +37,7 @@ class Upload(Model):
     """
     def initialize(self):
         self.name = 'upload'
+        self.ensureIndices(['sha512'])
```
zachmullen: Minor nit, this is equivalent to `self.ensureIndex('sha512')`

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -303,7 +316,12 @@ def deleteFile(self, file):
         if matching.count(True) == 1:
             path = os.path.join(self.assetstore['root'], file['path'])
             if os.path.isfile(path):
-                os.remove(path)
+                with filelock.FileLock(path + '.deleteLock'):
```
zachmullen: I think we may need to move this lock acquisition up above the file collection query -- there's another race condition here. If we do the file query, but someone uploads a duplicate after the file query returns 1 result, we'd delete the file erroneously.

manthey: When we have an upload, we obtain the lock just long enough to check if the file exists.  If so, then the file exists and was not being actively deleted, so we are safe to delete the temp file -- we've already marked that the upload is using the file.

When we are remove a Girder file, if it is the only girder file using the physical file because no girder file lists it, the we get a lock, make sure no upload is using it, and then remove it.  I don't see an order when we would accidentally delete an upload.  However, if an upload has completed, our file could could change to 2 in the interim between checking how many files are using the physical file and the delete.

So, yes, we need to move the lock, but I think for different reasons.

zachmullen: Oh right, forgot about the other place you use the lock.

mgrauer:
Just wanted to second the idea of supporting multiple app servers, we definitely have some Elastic Bean Stalk deployments of Girder lined up for the near future.

🦅

PR #2015, opened 2017-05-22
Item preview plugin improvements

matthewma7:
The following improvements has been implemented.
- Handle multiple files. Use HTTP HEAD request to determine if the item could potentially have multiple files and if the item or file is able to be previewed. Then iterate each file and preview them.
- Lazy load. More items will be loaded as user scroll near to the bottom to improve performance.
- Improved life cycle. Sharing the item collection with ItemListWidget and improving widget life cycle to avoid duplicate calls to API endpoints caused by render().
- Support ItemListWidget pagination.


matthewma7:
@dorukozturk, I think the change I made won't ItemListWidget change you made. 
Could you help me double check it?

dorukozturk:
@matthewma7 I tried and your changes did not affect what I fixed.

matthewma7:
@dorukozturk, That's great. Thanks for checking it for me. 😃

zachmullen:
Let's wait until the `nFiles` backend support lands to merge this; that will make this acceptably performant.

matthewma7:
@zachmullen This only load enough item to fill the page then lazy load the rest items. Both functionality and performance should have improved than the current implementation.

clients/web/src/utilities/jquery/promise.js
```diff
@@ -0,0 +1,6 @@
+import $ from 'jquery';
+
+// $.when behave differently with zero, one or more argument. This helper method minics Promise.all() behavior
+$.whenAll = function (promises) {
```
brianhelba: Let's call this `_whenAll`.

Per @zachmullen 's comment, consider putting it as a general Girder utility, rather than attaching it to `$`..

danlamanna:
re #1649

matthewma7:
@jbeezley, I fixed the issue after merging with master. This is ready for review.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
+            items.forEach((item, i) => {
+                var contents = results[i].filter((result) => result);
+                if (contents.length) {
+                    $container.append(ItemPreviewItemTemplate({
+                        item: item,
+                        contents: contents
+                    }));
+                }
             });
+            this.addingMoreItems = false;
```
jbeezley: If any of the prior promises are rejected, this variable will never get reset.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
-                type: 'GET',
-                error: null // don't do default error behavior (validation may fail)
-            }).done(function (resp) {
-                view.$('.json[data-id="' + id + '"]').text(JSON.stringify(resp, null, '\t'));
-            }).fail(function (err) {
-                console.error('Could not preview item', err);
+        var $container = this.$('.g-widget-item-prevews-wrapper');
+        _whenAll(items.map((item) => {
+            return restRequest({
+                path: `item/${item._id}/files`
```
jbeezley: This could potentially queue up 50 files to be downloaded, which would block anything else from loading until it is done.  Maybe use a smaller limit here?

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
-
-            // Don't process JSON files that are too big to preview.
-            var size = item.get('size');
-            if (size > this._MAX_JSON_SIZE) {
-                return view.$('.json[data-id="' + id + '"]').text('JSON too big to preview.');
-            }
+        if (!this.initialized) {
+            this.initialized = true;
+            this.renderedIndex = 0;
+            this.$el.html(ItemPreviewWidgetTemplate(this));
```
jbeezley: This template doesn't require any arguments.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
+            this.debouncedAddMoreItem();
+        }
+    },
+
+    isNotFull: function () {
+        return this.$('.g-widget-item-prevews-wrapper').height() <= this.$('.g-widget-item-previews-container').height();
+    },
+
+    allItemAdded: function () {
+        return this.renderedIndex === this.supportedItems.length;
```
jbeezley: I would make this `>=` just to be safe.

plugins/item_previews/web_client/templates/itemPreviewWidget.pug
```diff
@@ -1,30 +1,6 @@
...
-
-      - else if (isJSONItem(item))
-        .panel.panel-default
-          .panel-heading
-            h4
-              a.g-item-preview-link(data-id=id)
-                = item.name()
-          .panel-body
-            textarea.form-control.json(data-id=id)
+  .g-widget-item-prevews-wrapper
```
matthewma7: Thanks.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
+import { _whenAll } from 'girder/misc';
 
 import ItemPreviewWidgetTemplate from '../templates/itemPreviewWidget.pug';
+import ItemPreviewItemTemplate from '../templates/itemPreviewItem.pug';
 import '../stylesheets/itemPreviewWidget.styl';
 
 /**
  * The Item Preview widget shows a preview of items under a given folder.
- * For now, the only supported item previews are image previews.
+ * For now, this only support json and images file types.
```
matthewma7: Thanks

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
-        return /\.(json)$/i.test(item.get('name'));
+    _isImageItem: function (str) {
+        return /(jpg|jpeg|png|gif)$/i.test(str);
     },
 
-    _isImageItem: function (item) {
-        return /\.(jpg|jpeg|png|gif)$/i.test(item.get('name'));
+    setElement: function ($el) {
+        this.initialized = false;
+        View.prototype.setElement.apply(this, arguments);
```
matthewma7: I see.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
-            // Don't process JSON files that are too big to preview.
-            var size = item.get('size');
-            if (size > this._MAX_JSON_SIZE) {
-                return view.$('.json[data-id="' + id + '"]').text('JSON too big to preview.');
-            }
+        if (!this.initialized) {
+            this.initialized = true;
+            this.renderedIndex = 0;
+            this.$el.html(ItemPreviewWidgetTemplate(this));
+            this.$('.g-widget-item-previews-container').on('scroll', (e) => this.onScroll(e));
```
matthewma7: The scroll event is not bubbling to the parent. I didn't find a way to tell backbone to go with the event capture route, so I had to bind manually.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
+            this.debouncedAddMoreItem();
```
matthewma7: I remember I found that the HierarchyWidget may call the ItemPreviewWidget too frequently or too early, causing duplicate calls so I have the debounce, but internal calls to addMoreItem are better managed.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
+        }
+    },
+
+    isNotFull: function () {
+        return this.$('.g-widget-item-prevews-wrapper').height() <= this.$('.g-widget-item-previews-container').height();
+    },
+
+    allItemAdded: function () {
```
matthewma7: Thanks again.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
matthewma7: Is the number 50 the max number of files in an item? The code currently loads two items at a time, but I didn't put the effort to limit the files loading based on a maybe totally incorrect assumption that normally there won't be too many files for an item.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
+            return restRequest({
+                path: `item/${item._id}/files`
+            }).then((files) => {
+                return _whenAll(files.map((file) => {
+                    return this.tryGetFileContent(file, file.mimeType, 'file');
+                }));
+            });
+        })).then((results) => {
+            items.forEach((item, i) => {
+                var contents = results[i].filter((result) => result);
```
zachmullen: What's this `filter` for?

plugins/item_previews/web_client/views/ItemPreviewWidget.js
matthewma7: I was assuming if loading an item failed, there could be a bigger problem (server is down) in the system than loading the items, but putting this kind of logic inside a .always() or .finally() make sense.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
zachmullen: 50 is the default page size... if all we need to know is whether the file has more than 1 item, we can set limit=2.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
+    },
+
+    isNearBottom: function () {
+        var $container = this.$('.g-widget-item-previews-container');
+        return $container.scrollTop() + $container.innerHeight() >= $container[0].scrollHeight - 200;
+    },
+
+    onScroll: function (e) {
+        if (this.isNearBottom()) {
+            if (!this.addingMoreItems && !this.allItemAdded()) {
```
matthewma7: Totally, onscroll was called too frequently. 😁

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
+        }
+    },
 
-        this.$('.g-widget-item-previews-wrap').tooltip({
-            container: this.$el,
-            placement: 'left',
-            animation: false,
-            delay: {show: 100}
+    tryGetFileContent: function (file, contentType, type) {
+        if (this._isJSONItem(file.name) && contentType === 'application/octet-stream') {
```
matthewma7: I thought it should be `application/json` too, but it is `application/octet-stream` for both .json and .geojson I uploaded, did I miss something here?

zachmullen: That content type is set based on whatever the client's system believes the type to be, and they typically base that on looking up the file extension against a central registry.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,175 @@
...
+            if (file.size > this._MAX_JSON_SIZE) {
```
matthewma7: agreed, I tried a couple of large images and set a limit on images as well.

plugins/item_previews/web_client/stylesheets/itemPreviewWidget.styl
```diff
@@ -16,33 +11,28 @@ $maxPreviewHeight = 400px
...
-.g-widget-item-previews-container .panel .panel-body
-  padding 5px
-
 .g-widget-item-previews-container .json
   background-color #f5f5f5
   border-radius 4px
   border 1px solid #ccc
   color #333
-  height 300px
+  height $maxPreviewHeight
```
matthewma7: I had some trouble setting a good height for the textarea, without a `height`, it always shows as a single line for me.

plugins/item_previews/web_client/stylesheets/itemPreviewWidget.styl
```diff
@@ -16,33 +11,28 @@ $maxPreviewHeight = 400px
...
+  width 30%
```
matthewma7: Yea, it's kind of too small. The reason I didn't make it 100% is that a scrollable textarea will make it harder to scroll the container. (The user will scrolll the textarea instead of the div because the mouse cursor is on the textrea). I changed it to a `calc(100% - 100px)` to leave some space for scrolling at both ends.

plugins/item_previews/web_client/stylesheets/itemPreviewWidget.styl
```diff
@@ -16,33 +11,28 @@ $maxPreviewHeight = 400px
...
+  resize both
   white-space pre-wrap
```
matthewma7: Agreed. Updated.

plugins/item_previews/web_client/views/HierarchyWidget.js
```diff
@@ -7,20 +7,24 @@ import ItemPreviewWidget from './ItemPreviewWidget';
...
+        this.itemPreviewView.setElement($element);
+        this.itemPreviewView.setCollection(this.itemListView.collection);
+        this.itemPreviewView.render();
+    }
+
+    return this;
+});
 
-        // Add the item preview widget into the container.
+wrap(HierarchyWidget, '_initFolderViewSubwidgets', function (base) {
```
matthewma7: Logically, I think the `_initFolderViewSubwidgets` is the right place to initialize folder view specific stuff, but it works to move the initialization inside the render().

plugins/item_previews/web_client/views/ItemPreviewWidget.js
matthewma7: This is striping the out the `null` results coming from too large json or not supported content types.

matthewma7:
Hi @jbeezley, Is this ready to be merged?

plugins/item_previews/web_client/views/ItemPreviewWidget.js
jbeezley: It depends on how you upload.  If you upload through the browser, you should get `application/json`.  In any case, the correct mime type is `application/json`.

Since you are using the file extension to detect the type for both images and json, this extra condition is unnecessary.  I would just remove it.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,180 @@
...
+        _whenAll(items.map((item) => {
+            return restRequest({
+                path: `item/${item._id}/files`,
+                data: { limit: 5 }
+            }).then((files) => {
+                return _whenAll(files.map((file) => {
+                    return this.tryGetFileContent(file, file.mimeType);
+                }));
+            });
+        })).then((results) => {
```
brianhelba: You can simplify this by using `.done(` instead of `.then(` here, and removing the `return undefined;` at the end of the function.

plugins/item_previews/web_client/templates/itemPreviewItem.pug
```diff
@@ -0,0 +1,12 @@
...
+  h4
+    a.g-item-preview-link(data-id=item._id)= item.name
+  
+  each content in contents
+    if content.type==='image'
+      div
+        img(src=content.value)
+
+    else if content.type==='json'
+      textarea.json(data-id=item._id)= JSON.stringify(content.value, null, '\t')
```
brianhelba: This file should end with a newline.

plugins/item_previews/web_client/views/HierarchyWidget.js
```diff
@@ -7,20 +7,18 @@ import ItemPreviewWidget from './ItemPreviewWidget';
...
-            el: element
-        })
-        .render();
+        if (!this.itemPreviewView) {
+            this.itemPreviewView = new ItemPreviewWidget({
+                collection: this.itemListView.collection,
+                parentView: this
+            });
+        }
+        var $element = $('<div class="g-item-previews-container">');
```
brianhelba: Girder variable names typically don't start with a `$`.

brianhelba: Also, it's probably best to start using `let` to declare variables in new code.

plugins/item_previews/web_client/views/HierarchyWidget.js
```diff
@@ -7,20 +7,18 @@ import ItemPreviewWidget from './ItemPreviewWidget';
...
-
-        // Add the item preview widget into the container.
-        this.itemPreviewView = new ItemPreviewWidget({
-            folderId: this.parentModel.get('_id'),
-            parentView: this,
-            el: element
-        })
-        .render();
+        if (!this.itemPreviewView) {
+            this.itemPreviewView = new ItemPreviewWidget({
```
brianhelba: The binding of this sub-view to the DOM can be simplified. Just pass `class: 'g-item-preview-container',` as an argument when instantiating the `ItemPreviewWidget`. The new `this.itemPreviewView` will be created with it's own `<div>`element (that's the default type) of the correct class, making the element creation and call to `.setElement(` here unnecessary.

Then you can just do ` this.$el.append(this.itemPreviewView.$el);`.

brianhelba: See http://backbonejs.org/#View-el

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,179 @@
...
 
-        this.wrapPreviews = false;
+        this.supportedItems = [];
+    },
+
+    // Because the parent-child life cycle is not clean, this method is needed.
+    setCollection: function (collection) {
+        this.collection = collection;
+        this.collection.on('g:changed', () => {
+            this.supportedItems = this.collection.toJSON().filter(this._isSupportedItem);
```
brianhelba: Backbone Collections natively have a `.filter(` method. Using `.toJSON` will cause the contained Models to be converted to ordinary objects, losing their special methods.

See  http://backbonejs.org/#Collection-Underscore-Methods

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,179 @@
...
-                this.$('.g-widget-item-previews-container').addClass('g-wrap');
-            } else {
-                this.$('.g-widget-item-previews-container').removeClass('g-wrap');
-            }
         }
     },
 
     initialize: function (settings) {
-        this.collection = new ItemCollection();
+        this._isSupportedItem = this._isSupportedItem.bind(this);
```
brianhelba: This self-modifying behavior is too tricky, IMO. Most Backbone methods are smart enough to call things with the correct `this` bound. If there are truly problematic places, `_.bind` should be used inline.

matthewma7:
@brianhelba, I don't have the allocation to look into your suggestion. If you feel strongly about the suggestions you made, could you make the changes? 
Otherwise, I see the PR works perfectly as a bug fix and feature improvement.

plugins/item_previews/web_client/views/ItemPreviewWidget.js
```diff
@@ -1,103 +1,179 @@
...
+        this.addingMoreItems = false;
 
-        this.wrapPreviews = false;
+        this.supportedItems = [];
+    },
+
+    // Because the parent-child life cycle is not clean, this method is needed.
+    setCollection: function (collection) {
+        this.collection = collection;
+        this.collection.on('g:changed', () => {
```
brianhelba: Use `this.listenTo` here, to avoid a memory leak.

brianhelba:
I'll be free on Wednesday. I'll be happy then to help ensure that things are using Backbone / Girder conventions.

Some of the issues (particularly memory leaks) should be addressed before merging.

zachmullen:
Conclusion: the item_previews plugin suffers from design problems and we may remove its built-in status in 3.0.

zachmullen:
Conclusion: @matthewma7 will spend some time to address issues in this PR to merge it since it is strictly an improvement.

matthewma7:
@jbeezley Thank you for the reviewing!

PR #2021, opened 2017-05-23
Upgrade dicom_viewer dependency to vtk.js 2.24.1

zachmullen:
This simplifies our build process and reduces the size of
the plugin, as well as reducing npm install time by removing
the dependency on kw-web-suite.


mgrauer:
@MayeulChassagnard Can you test this please?

PR #2023, opened 2017-05-24
WIP: Improve S3 assetstore error message

danlamanna:
This also fixes an issue where we were leaving around a file created by our S3 assetstore initialization code.

To quote the boto docs:
```
If the default ``validate=True`` is passed, a request is made to the
service to ensure the bucket exists. Prior to Boto v2.25.0, this fetched
a list of keys (but with a max limit set to ``0``, always returning an empty
list) in the bucket (& included better error messages), at an
increased expense. As of Boto v2.25.0, this now performs a HEAD request
(less expensive but worse error messages).

If you were relying on parsing the error message before, you should call
something like::

bucket = conn.get_bucket('<bucket_name>', validate=False)
bucket.get_all_keys(maxkeys=0)
```

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -152,7 +152,8 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
         else:
             try:
                 bucket = conn.get_bucket(bucket_name=doc['bucket'],
-                                         validate=True)
+                                         validate=False)
+                bucket.get_all_keys(maxkeys=0)
                 testKey = boto.s3.key.Key(
                     bucket=bucket, name='/'.join(
                         filter(None, (doc['prefix'], 'test'))))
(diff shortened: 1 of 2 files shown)


danlamanna:
@zachmullen It seems like [this test](https://github.com/girder/girder/blob/improve-s3-assetstore-errors/tests/cases/assetstore_test.py#L699) relied on that 'test' item existing to throw that error. Do you know the moto/boto way to mock an existing item in the bucket?

zachmullen:
If you're using moto you can just write the key using the normal API, and it should behave identically as though you were talking to a real S3 service.

danlamanna:
@zachmullen That actually didn't work for me, I had to mock `_getBucket` to return the same mocked bucket I added the key to. Let me know if that looks good to you or if there's a simpler way.

Tests won't pass for me on Circle, web_client_admin was failing pretty often, and now I'm getting an npm error regarding "No matching version found for eslint-config-girder@^3.0.0". @brianhelba

zachmullen:
Did you rebase on latest master?

zachmullen:
Also, did you test this with a real S3 assetstore? I can give you credentials if you don't have some.

danlamanna:
I did test it on real S3 assetstores. I rebased on master with no success, I just rebased again now though so we'll see.

zachmullen:
This PR will become obsolete due to a forthcoming switch to boto3. We can re-apply this fix if necessary on top of that change.

PR #2026, opened 2017-05-25
Use jQuery Deferred and Promise, instead of ES6 Promise

brianhelba:
* Use jQuery Deferred and Promise, instead of ES6 Promise, in core
  * The ES6 promises cannot be transpiled by Babel, and requires a polyfill for some browsers (mostly IE) and the test environment. All other web code in Girder uses jQuery promises, which are not interoperable with ES6 promises (particularly in jQuery < v3).
* Reorder and move some private functions to a more logical grouping
* Use jQuery Deferred and Promise, instead of ES6 Promise, in plugins

Updates code written in https://github.com/girder/girder/pull/1844.

### DIFF
--- a/clients/web/src/server.js
+++ b/clients/web/src/server.js
@@ -1,49 +1,87 @@
+import $ from 'jquery';
+
 import { confirm } from 'girder/dialog';
 import events from 'girder/events';
 import { restRequest } from 'girder/rest';
 
 /**
- * Restart the server, wait until it has restarted, then reload the current
- * page.
+ * Periodically re-run a promise-returning function, until it is fulfilled.
+ * @param func A promise-returning function, taking no arguments.
+ * @param delay A delay, in milliseconds, to wait before re-running a rejected "func".
+ * @returns {$.Promise} A promise, which fulfills with the same values as the fulfilled "func".
+ */
+function _retryUntilFulfilled(func, delay) {
+    const resolution = $.Deferred();
+    const loop = () => {
+        // Run the function
+        func()
+            .done(function () {
+                // If it's fulfilled, then fulfill the resolution with the same values
+                resolution.resolve.apply(resolution, arguments);
+            })
+            .fail(() => {
+                // If it's rejected, check again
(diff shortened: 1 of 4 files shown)


brianhelba:
@matthewma7 PTAL. This is the code we discussed, which is now ready for review.

We can address the broader matter of a jQuery -> ES6 Promise transition separately, but this change just needs a review on it's technical merits.

clients/web/src/server.js
```diff
@@ -1,50 +1,100 @@
...
+        } else {
+            checkResolution.reject();
+        }
+        return checkResolution.promise();
+    });
+};
+
+/* Having these as object properties facilitates testing */
+restartServer._callSystemRestart = function () {
+    restRequest({
```
zachmullen: Do we need to be returning this value? I notice above that `restartServer` asynchronously returns the return value from this function, which is always undefined as written.

zachmullen:
I'm generally in favor of this change. Should we remove the Promise polyfill from our phantom environment?

clients/web/src/server.js
brianhelba: @zachmullen Fixed. That change was lost in my rebase. The old behavior is actually to not wait on the restart to confirm, but the new (now fixed) behavior includes it in the pipeline.

brianhelba:
This PR is an important dependency of my upgrade to jQuery3. Right now, it's blocking me from moving forward on that task.

This PR also fixes some minor bugs with the current implementation, fixes the return type of a public method, and (IMHO) now has more readable code.

I'd really like to merge this now. We can already change parts of it back, if the community consensus pushes us towards ES6 Promises.

@matthewma7 Is that ok?

matthewma7:
@brianhelba Sounds good. 
I am preparing to create a separate post for ES6 promise based on the drawback of jQuery deferred we found yesterday and benefits of ES6 promises.

zachmullen:
FYI, you are ok to merge even though that diff coverage status failed.

Issue #2018, opened 2017-05-23
Creation of girder job models via REST endpoints in worker plugin


kotfic:
### Deliverables:
+ Post to endpoint that creates a girder job
### Technical Notes:
+ Parent-child is one to many
+ How will we implement access control on creating jobs?
+ How should access control be implemented for job creation?
+ Depending on how restrictive we want to be this to be we may want require that jobs can only be created if they are children of existing jobs
### Depends on: 
+ Allow parent-child relationships between girder job models (https://github.com/girder/girder/issues/2017 - see note above)

zachmullen:
Re: access control, for the first pass we can require a special token scope that allows job creation, and create such a token just for the worker to use. Thus, ordinary users could not just create jobs arbitrarily.

PR #2031, opened 2017-05-25
Create REST endpoint for POSTing job models

kotfic:
Does not allow posting to ```/job``` unless jobs.rest.create_job is set as a scope
on the current token - regardless of whether TokenScope.USER_AUTH is set.

### DIFF
--- a/girder/models/token.py
+++ b/girder/models/token.py
@@ -21,6 +21,7 @@ import datetime
 import six
 
 from girder.constants import AccessType, SettingKey, TokenScope
+from girder.models.model_base import AccessException
 from girder.utility import genToken
 from .model_base import AccessControlledModel
 
@@ -130,6 +131,20 @@ class Token(AccessControlledModel):
         if isinstance(scope, six.string_types):
             scope = (scope,)
         return set(scope).issubset(set(self.getAllowedScopes(token)))
+
+    def requireScope(self, token, scope):
+        """
+        Raise an error if given set of scopes are not included.
+
+        :param token: The token object.
+        :type token: dict
+        :param scope: A scope or set of scopes that will be tested as a subset
+            of the given token's allowed scopes.
+        :type scope: str or list of str
+        """
+
+        if not self.hasScope(token, scope):
+            raise AccessException('Invalid token scope, required: %s.' % (scope))
 
     def clearForApiKey(self, apiKey):
         """
--- a/plugins/jobs/server/__init__.py
+++ b/plugins/jobs/server/__init__.py
@@ -21,10 +21,6 @@ import importlib
 
 from girder import events
 from . import constants, job_rest
-from girder.api import access
-from girder.api.rest import getCurrentToken
-from girder.models.model_base import AccessException
-from girder.utility.model_importer import ModelImporter
 
 
 def scheduleLocal(event):
@@ -46,18 +42,7 @@ def scheduleLocal(event):
         fn = getattr(module, job.get('function', 'run'))
         fn(job)
 
-@access.token(scope=constants.REST_CREATE_JOB_TOKEN_SCOPE)
-def _authorizeRestJobCreation(event):
-    tokenModel = ModelImporter.model('token')
-    token = getCurrentToken()
-
-    if not tokenModel.hasScope(token, constants.REST_CREATE_JOB_TOKEN_SCOPE):
-        raise AccessException(
-            'Invalid token scope.\n'
-            'Required: %s.\n' % (constants.REST_CREATE_JOB_TOKEN_SCOPE))
-
 
 def load(info):
     info['apiRoot'].job = job_rest.Job()
-    events.bind('rest.post.job.before', 'jobs', _authorizeRestJobCreation)
     events.bind('jobs.schedule', 'jobs', scheduleLocal)
--- a/plugins/jobs/server/__init__.py
+++ b/plugins/jobs/server/__init__.py
@@ -21,6 +21,10 @@ import importlib
 
 from girder import events
 from . import constants, job_rest
+from girder.api import access
+from girder.api.rest import getCurrentToken
+from girder.models.model_base import AccessException
+from girder.utility.model_importer import ModelImporter
 
 
 def scheduleLocal(event):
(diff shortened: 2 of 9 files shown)


kotfic:
@zachmullen can you comment on whether or not this is the right/best approach? thanks!

kotfic:
See: https://github.com/girder/girder/issues/2018

kotfic:
btw,  I found this to be a little bit confusing - i expected when setting ```@access.token(scope='FOOBAR')``` that the endpoint would throw a 403 if the token didn't have scope 'FOOBAR' set on it.  

However,  because [ensureTokenScopes](https://github.com/girder/girder/blob/master/girder/api/rest.py#L622) returns immediately if the user is logged in,  it effectively means a logged in user has ALL custom scopes defined. I didn't see a built-in way to handle a situation where i wanted to protect an endpoint with a custom scope regardless of user authentication status.

plugins/jobs/server/job_rest.py
```diff
@@ -63,6 +64,32 @@ def listJobs(self, userId, types, statuses, limit, offset, sort, params):
             user=user, offset=offset, limit=limit, types=types,
             statuses=statuses, sort=sort, currentUser=currentUser))
 
+
+    @filtermodel(model='job', plugin='jobs')
+    @access.token(scope=constants.REST_CREATE_JOB_TOKEN_SCOPE)
+    @autoDescribeRoute(
+        Description('Create a job model via a RESTful endpoint')
```
zachmullen: No need to put "via a RESTful endpoint" :)

plugins/jobs/server/__init__.py
```diff
@@ -42,7 +46,18 @@ def scheduleLocal(event):
...
+
+    if not tokenModel.hasScope(token, constants.REST_CREATE_JOB_TOKEN_SCOPE):
+        raise AccessException(
+            'Invalid token scope.\n'
+            'Required: %s.\n' % (constants.REST_CREATE_JOB_TOKEN_SCOPE))
+
 
 def load(info):
     info['apiRoot'].job = job_rest.Job()
+    events.bind('rest.post.job.before', 'jobs', _authorizeRestJobCreation)
```
zachmullen: Since this is all in the same plugin, let's just move this logic into the REST route handler itself. That will improve readability via lexical proximity of the logic, and won't preclude behavioral modification by downstreams.

plugins/jobs/server/__init__.py
```diff
@@ -42,7 +46,18 @@ def scheduleLocal(event):
...
 
+@access.token(scope=constants.REST_CREATE_JOB_TOKEN_SCOPE)
+def _authorizeRestJobCreation(event):
+    tokenModel = ModelImporter.model('token')
+    token = getCurrentToken()
+
+    if not tokenModel.hasScope(token, constants.REST_CREATE_JOB_TOKEN_SCOPE):
+        raise AccessException(
+            'Invalid token scope.\n'
+            'Required: %s.\n' % (constants.REST_CREATE_JOB_TOKEN_SCOPE))
```
zachmullen: This pattern appears in a few places, it's probably worth adding a standardized `requireScope` method into the token model that raises an exception like this.

plugins/jobs/server/models/job.py
```diff
@@ -24,7 +24,9 @@
 from girder import events
 from girder.constants import AccessType, SortDir
 from girder.models.model_base import AccessControlledModel, ValidationException
-from girder.plugins.jobs.constants import JobStatus, JOB_HANDLER_LOCAL
+from girder.plugins.jobs.constants import (JobStatus,
+                                           JOB_HANDLER_LOCAL,
+                                           REST_CREATE_JOB_TOKEN_SCOPE)
```
zachmullen: This change doesn't seem necessary.

zachmullen:
> btw, I found this to be a little bit confusing - i expected when setting @access.token(scope='FOOBAR') that the endpoint would throw a 403 if the token didn't have scope 'FOOBAR' set on it.

Right, that's the behavior we use to support OAuth-provider-like authorization control. One way to support these sorts of use cases is to add another flag to the access decorators, e.g.

    @access.token(scope=FOO, required=True)

The `required` flag (for lack of a more expressive name) would default to False, but if True would actually require the listed scopes. This PR would be a good place to add that, and also the token model `requireScope` method that would underlie it.

plugins/jobs/server/__init__.py
mgrauer: Would this also help with the recent issue we found where invalidly scoped tokens where throwing 401s and saying "not logged in" when really it was the case that the token didn't have the correct scope?  If we can improve that error handling as a part of this PR that would be rad.

zachmullen: It wouldn't help with that particular case. I'm not actually sure how to engineer a solution for that at this time.

dorukozturk:
@kotfic, @zachmullen This is ready for a review.

girder/models/token.py
```diff
@@ -131,6 +132,21 @@ def hasScope(self, token, scope):
...
+        :param token: The token object.
+        :type token: dict
+        :param scope: A scope or set of scopes that will be tested as a subset
+            of the given token's allowed scopes.
+        :type scope: str or list of str
+        """
+
+        if not self.hasScope(token, scope):
+            raise AccessException('Invalid token scope.'
+                                  'Required: %s.' % (scope))
```
zachmullen: I think we can fit this all on one line, but if not, we'll need to add a space between the sentences.

plugins/jobs/server/job_rest.py
```diff
@@ -72,6 +74,36 @@ def listJobs(self, userId, parentJob, types, statuses, limit, offset, sort,
             statuses=statuses, sort=sort, currentUser=currentUser,
             parentJob=parent))
 
+    @filtermodel(model='job', plugin='jobs')
+    @access.token(scope=constants.REST_CREATE_JOB_TOKEN_SCOPE, required=True)
+    @autoDescribeRoute(
+        Description('Create a job model')
+        .param('title', '', required=True)
+        .param('type', '', required=True)
+        .modelParam('parentId', 'Id of the parent job.', model='job',
```
zachmullen: In user-facing text, we should use "ID" rather than "Id".

dorukozturk: @zachmullen I applied those changes.

PR #2032, opened 2017-05-26
Deprecate restRequest.error

brianhelba:
* Deprecate restRequest.error
  * This method has been deprecated since jQuery 1.8. Since it's still used frequently in Girder, patch it to issue a warning.
* Remove use of deprecated restRequest.error method

### DIFF
--- a/clients/web/src/rest.js
+++ b/clients/web/src/rest.js
@@ -30,7 +30,7 @@ function setStaticRoot(root) {
 /**
  * Make a request to the REST API. Bind a "done" handler to the return
  * value that will be called when the response is successful. To bind a
- * custom error handler, bind an "error" handler to the return promise,
+ * custom error handler, bind a "fail" handler to the return promise,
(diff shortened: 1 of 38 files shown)


clients/web/src/rest.js
```diff
@@ -113,7 +113,13 @@ function __restRequest(opts) {
         opts.headers = opts.headers || {};
         opts.headers['Girder-Token'] = token;
     }
-    return Backbone.ajax(opts);
+
+    let jqXHR = Backbone.ajax(opts);
+    jqXHR.error = function () {
+        console.warn('Use of restRequest.error is deprecated, use restRequest.fail instead.');
+        jqXHR.fail.apply(jqXHR, arguments);
```
jbeezley: You should return the result of this call.

brianhelba:
Fixed the return issue, and rebased on master to handle merge conflicts. No other changes.

@jbeezley Ready for re-review.

PR #2033, opened 2017-05-26
Replace deprecated jQuery functions

brianhelba:
* Use $(fn) as a document-ready handlers
  * The other forms are deprecated in jQuery 3.
* Replace jQuery .unbind with .off
  * The .unbind method is deprecated in jQuery 3.
* Use JSON.parse instead of $.parseJSON
  * The jQuery method is deprecated in jQuery 3.

### DIFF
--- a/clients/web-external/main.js
+++ b/clients/web-external/main.js
@@ -1,4 +1,4 @@
-$(document).ready(function () {
+$(function () {
     girder.apiRoot = '/girder/api/v1';
     girder.router.enabled(false);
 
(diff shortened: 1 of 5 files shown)


brianhelba:
PTAL:
@jeffbaumes @jbeezley : You were blamed for code that uses `$(document).ready`
@zachmullen : You were blamed for code that uses `$.parseJSON`
@manthey : You were blamed for code that uses `$(...).unbind`

brianhelba:
@girder/developers PTAL, it should be an easy review.

PR #2035, opened 2017-05-26
Use $.girderEnable to enable and disable widgets

brianhelba:
Setting an element's "disabled" *attribute* is typically not correct. The element's *property* should be set instead.


brianhelba:
@girder/developers PTAL.

PR #2036, opened 2017-05-26
Don't use $.when to re-wrap promises

brianhelba:
This is redundant.

Given that the multi-argument version of `$.then` has different behavior, `$.when` should not be used with a single argument.

### DIFF
--- a/clients/web/src/collections/Collection.js
+++ b/clients/web/src/collections/Collection.js
@@ -164,7 +164,7 @@ var Collection = Backbone.Collection.extend({
                 }, this.params)
             });
 
-            var result = $.when(xhr).then((list) => {
+            var result = xhr.then((list) => {
                 if (this.pageLimit > 0 && list.length > this.pageLimit) {
                     // This means we have more pages to display still. Pop off
                     // the extra that we fetched.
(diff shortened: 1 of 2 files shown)


brianhelba:
PTAL:
@opadron : You were blamed for code that uses `$.when` to re-wrap promises

manthey:
`collection.fetch()` can return something other than a promise, seemingly only in what should be an uncommon error condition.  Since this is so, don't you need to wrap the results to ensure that you have a promise, since otherwise the results might not have a `then` method?

brianhelba:
@manthey That's a good point, though the old behavior is possibly worse: `$.when(...fetch(...))` could just create a resolved Promise (with a value of `undefined`) and continue to the handlers without error.

I think `Collection.fetch` should `throw` in that error case. It's fine if it's uncaught, as it's intended as a guard for new subclasses, and should never trigger in a non-development runtime.

brianhelba:
@manthey Thanks for the tip. I've fixed all three places.

brianhelba:
@manthey See https://github.com/girder/girder/pull/2049, since the fix was actually getting to be outside the scope of this PR.

brianhelba:
Rebased, now that #2049 is merged.

PR #2039, opened 2017-05-29
Allow usage of a complete URL for static root

danlamanna:
Prior to this, the static route accepted URLs (e180d73) but would still pass
them to os.path.relpath leading to a malformed static url. Here, we
leave the static route alone if it's a URL.

This PR allows us to use a CDN when deploying on Beanstalk (or anywhere).

### DIFF
--- a/girder/utility/server.py
+++ b/girder/utility/server.py
@@ -22,6 +22,7 @@ import functools
 import mako
 import mimetypes
 import os
+import posixpath
 import six
 
 import girder.events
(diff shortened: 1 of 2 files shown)


girder/utility/server.py
```diff
@@ -125,24 +125,32 @@ def configureServer(test=False, plugins=None, curConfig=None):
...
+    else:
+        # Make the staticRoot relative to the api_root, if possible.  The api_root
+        # could be relative or absolute, but it needs to be in an absolute form for
+        # relpath to behave as expected.  We always expect the api_root to
+        # contain at least two components, but the reference from static needs to
+        # be from only the first component.
+        apiRootBase = os.path.split(os.path.join('/', curConfig['server']['api_root']))[0]
+        apiStaticRoot = os.path.relpath(routeTable[constants.GIRDER_STATIC_ROUTE_ID],
+                                        apiRootBase)
+        staticRoot = os.path.relpath(routeTable[constants.GIRDER_STATIC_ROUTE_ID],
```
zachmullen: Shouldn't we use `posixpath` instead of `os.path`?

danlamanna: I'd never actually heard of this, but this is the docstring for `posixpath`:
> Instead of importing this module directly, import os and refer to
this module as os.path.  The "os.path" name is an alias for this
module on Posix systems; on other systems (e.g. Mac, Windows),
os.path provides the same operations in a manner specific to that
platform, and is an alias to another module (e.g. macpath, ntpath).

zachmullen: Yep, that's true if you're referring to the local fileystem, but in this case we know we are in an environment that exclusively supports forward slash separators, so using `posixpath` is more defensive in case we ever want to deploy on windows.

danlamanna:
@manthey If this method of solving my problem makes sense/looks good to you, I'll move it out into a separate `configureStaticRoutes` method to keep `configureServer` simpler.

manthey:
@danlamanna I like the approach.

PR #2040, opened 2017-05-29
Add database initialization for testing

### LINKED ISSUE #2022: RFC: Prepopulating the database for client tests
A common pattern when writing client tests for girder is to create a bunch of set up code that creates users, uploads files, etc.  We have many functions in [testUtils](https://github.com/girder/girder/blob/master/clients/web/test/testUtils.js) to reduce the amount of boilerplate code necessary to do this; however, I've had a few problems with this approach.
1. The helper functions don't work for testing external plugins that don't use the main girder ui.
2. It relies on navigating the page using synthetic click events that are difficult to get right in the headless environment.
3. It creates many opportunities for race conditions in the tests and generally slows down the testing.

I'm considering adding some code to the test runner that can initialize the database to some repeatable state before starting the web server, and I want to collect comments about it before I attempt to implement it.  The idea would be to add a new optional argument to the `add_web_client_test` where you could pass in the path to a json file that would contain the desired state.

I considered the idea of just loading a mongo backup, but that would require regenerating them every time the document model changes.  Instead, I think it would be best to derive a custom format containing the information necessary to initialize the documents using girder's model API.  A possible json format would be structured as follows:
```json
{
  "users": [{
    "admin": true,
    "email": "...",
    "firstName": "...",
    "lastName": "...",
    "login": "...",
    
    "folders": [{
      "name": "...", 
      "description": "...",
      "public": true,

      "folders": [],

      "items": [{
        "name": "...", 
        "description": "...",
        "files": [{
          "name": "...",
          "mimeType": "...",
          "path": "/path/to/file/content"
        }]
      }]
    }]
  }],
  
  "collections": [
  ]
}
```
Each object would roughly contain the arguments passed to the `create*` method for each model type.  The objects would be nested to declare their intended topology.

Are there any problems with this that I'm missing?  Is this something others would be interested in using?

jbeezley:
This largely follows the description in #2022 along with the suggestions about simplified representations for common use cases.  This solves my current need, but there are a number of things that could be added in future PR's:
* Item and folder metadata
* Settings
* Groups
* Access control lists

Closes #2022


jbeezley:
In back channel conversations, it has been noted that this PR overlaps with the purpose of the girder ansible client.  This code interacts with the model API directly, while the ansible client calls the rest API.  There is also the unfortunate reality that this introduces an alternate "spec" for bootstrapping girder.  An argument in favor of this PR is that it is intended for a different use case than the ansible client and that they are sufficiently different to merit a custom solution.  In this case, it would be important to emphasize that this code should *not* be used to provision a deployment.

I'm curious if anyone else has any thoughts about this.  I'm willing to reconsider how it is implemented to unify the approaches if necessary.

docs/development.rst
```diff
@@ -245,6 +245,57 @@ You will find many useful methods for client side testing in the ``girderTest``
 defined at ``/clients/web/test/testUtils.js``.
 
 
+Initializing the Database for a Test
+------------------------------------
+
+When running tests in Girder, the database will initially be empty.  Often times, you want to be able to start the test with the database in a
+particular state.  To avoid repetitive initialization code, Girder provides a way to import a folder hierarchy from the file system
+using a simple initialization file.  This file is in YAML (or JSON) format and provides a list of objects to insert into the database
```
danlamanna: Is the JSON part true, where does it import that?

danlamanna:
1) Are there any tests in Girder that we should move to this new method as another example to have, and to reduce the boilerplate?
2) With the merging of #2047, does it make sense to force some standard path/naming for these sorts of files and have some level of integration with `add_standard_plugin_tests`?

docs/development.rst
jbeezley: It's a minor point.  The yaml module supports parsing json files as well.

zachmullen: JSON is YAML :)

danlamanna: :face_with_head_bandage: :boom:

tests/JavascriptTests.cmake
```diff
@@ -181,6 +186,7 @@ function(add_web_client_test case specFile)
     "GIRDER_PORT=${web_client_port}"
     "MONGOD_EXECUTABLE=${MONGOD_EXECUTABLE}"
     "GIRDER_TEST_DATA_PREFIX=${GIRDER_EXTERNAL_DATA_ROOT}"
+    "${setup_database_env}"
```
jbeezley: You mean `"GIRDER_TEST_DATABASE_CONFIG=${fn_SETUP_DATABASE}"` here?  The problem with that is the environment variable would be set.  I would have to modify the code [here](https://github.com/girder/girder/pull/2040/files/0f1fb6929af47fad4073a14d6c0f3d51e7778fed#diff-44b2005fe01184a1089900dedd126fd4R196) to detect if the var exists but is empty.  I'm willing to do so if you think it's better.

danlamanna: I think it's easier to read if the variables are consistent.

jbeezley:
> Are there any tests in Girder that we should move to this new method as another example to have, and to reduce boilerplate?

Probably many, but I don't this is the correct place to do a large-scale refactoring of test cases.  I would propose picking a single core test to modify as an example for this PR.  Unless someone wants to nominate a test, I'll choose one to port over.

> With the merging of #2047, does it make sense to force some standard path/naming for these sorts of files and have some level of integration with add_standard_plugin_tests?

At one point, I was thinking something along the lines of autodetecting from file names, e.g. `testSpec.js` would look for `testSpec.(yml|json)`.  I haven't decided if that is a good idea or not.  In either case, I do need to integrate this with the changes in #2047.

zachmullen:
> I would propose picking a single core test to modify as an example for this PR. Unless someone wants to nominate a test, I'll choose one to port over.

I'll nominate search_test

jbeezley:
>> With the merging of #2047, does it make sense to force some standard path/naming for these sorts of files and have some level of integration with add_standard_plugin_tests?

> At one point, I was thinking something along the lines of autodetecting from file names, e.g. testSpec.js would look for testSpec.(yml|json). I haven't decided if that is a good idea or not. In either case, I do need to integrate this with the changes in #2047.

In https://github.com/girder/girder/pull/2040/commits/f461c494aa87c4257d247935e83c7d76d2c9132d and https://github.com/girder/girder/pull/2040/commits/51958aef8fed38f5968e62376424ccac2c1ec756, I added a default location for the initialization file.  With this change, I don't think any extra parameters need to be added to `add_standard_plugin_tests`.

jbeezley:
I'll make one last call for comment on this, otherwise I'll go ahead and merge it later tonight.

Issue #2027, opened 2017-05-25
Vagrant up doesn't work on Windows


manthey:
When `vagrant up` is used on Windows (and possibly on linux when the girder repository is on certain file systems), it fails.  The current directory is mounted by Vagrant to /home/vagrant/girder, and node attempts to create symlinks within this directory.  However, the ability to make symlinks is not exposed through the vboxsf, so this fails.

VirtualBox says the ability to create symlinks in a mounted directory is a security issue (https://www.virtualbox.org/ticket/10085).  It looks like a won't fix (or at least not any time soon), though there are various work-arounds.

One solution would be to *not* install in the mounted local directory, but rather copy the current directory to some other path, and use ansible_local from there (`/home/vagrant/not_girder`).  This has the drawback that it wouldn't use the current girder code for the vagrant, but the virtue it doesn't spatter the local file system with node modules.

msmolens:
An upstream solution for this would be welcome, especially one that still allows editing Girder code directly on the host. This would make it easier to develop core Girder when only a Windows machine is available, or for participants in hackfests who might only have a Windows machine.

manthey:
At least as a workaround, one could bind mount a local directory to the girder/node_modules directory in the VirtualBox.  The build then works (since it is only npm that creates symlinks to npm binaries).

I know there are better and more official ways to do this, but, crudely, adding this near the end of devops/ansible/roles/girder/tasks/girder.yml works:
```
- name: Bind mount node_modules
  shell: |
    mkdir -p /home/vagrant/girder/node_modules
    mkdir -p /home/vagrant/girder_node_modules
    sudo mount --bind /home/vagrant/girder_node_modules /home/vagrant/girder/node_modules
```
If we want to use that approach, we'd want the bind mount to automatically be done when `vagrant up` is done on an already provisioned VM, too.

manthey:
Resolved in PR #2041.

PR #2041, opened 2017-05-29
When installing via Vagrant, bind mount node_modules directory.

manthey:
This works around the issue that npm symlinks files in the node_modules directory, and that VirtualBox's vboxfs doesn't expose symlinks on Windows.

Fixes issue #2027.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -29,9 +29,26 @@ Vagrant.configure("2") do |config|
     provisioner_type = "ansible"
   end
 
+  client_testing = ENV["ANSIBLE_CLIENT_TESTING"] || false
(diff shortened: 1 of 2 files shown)


danlamanna:
I think this is too deployment-site specific to be the responsibility of the Ansible role. I would rather see this added to the Vagrantfile through some other means, like another provision block.

manthey:
Okay.  I'll refactor it.

danlamanna:
This currently causes all of our provision steps to fail because git is trying to checkout Girder into `/home/vagrant/girder` and it already has `node_modules`, see: http://my.cdash.org/viewTest.php?onlyfailed&buildid=1218587

manthey:
Because the provisioning tests use the special ANSIBLE_TESTING value, it means we aren't actually testing the Vagrantfile for the expected use case of a typical user (running from the clone girder repo).

PR #2044, opened 2017-05-30
Use npm install --no-save for npm v5

### LINKED ISSUE #2043: Incompatibility with npm v5
npm version 5 was recently released and made some changes that break how we install plugins in girder. `npm install <package>` now saves the package into `package.json` so when we install plugin dependencies, girder's `package.json` is modified.

jbeezley:
I didn't version guard the flag because it doesn't seem to break any old versions of npm, and it is difficult to detect in the script which version of npm is being used (you can't `require('npm')`).  I tested it on versions 3, 4, and 5.

Fixes #2043

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-package-lock.json
+/package-lock.json
 *.py[cod]
 node_modules
 docs/_build
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+package-lock.json
 *.py[cod]
 node_modules
 docs/_build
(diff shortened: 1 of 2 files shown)


zachmullen:
Is there any reason we should also add this flag in install.py?

jbeezley:
I don't think it's necessary.  We are just doing a bare `npm install` there; this change only affects the case when you are doing `npm install <package>`.

jbeezley:
I also added `package-lock.json` to the gitignore file because this is now created by `npm install`.

.gitignore
```diff
@@ -1,3 +1,4 @@
+package-lock.json
```
zachmullen: This will cause this file to be ignored in all subdirectories, is that intentional?

PR #2046, opened 2017-05-30
Parent Child Jobs

### LINKED ISSUE #2017: Allow parent-child relationships between girder job models
### Deliverables:
+ Add attribute to job model in girder that references parrent
+ Add utilities to query job collection to get list of child tasks
### Technical Notes:
+ In the case of groups of tasks,  how do we handle the failure of an individual task? In some cases we want the group to fail,  in others we want the group to continue. 
+ How to visualise the parent child relationship (keeping an eye towards supporting canvas features). 
### Depends on: None

dorukozturk:
Fixes #2017


dorukozturk:
@kotfic and @danlamanna PTAL.

plugins/jobs/server/job_rest.py
```diff
@@ -33,6 +33,8 @@ def __init__(self):
         self.route('GET', ('all',), self.listAllJobs)
         self.route('GET', (':id',), self.getJob)
         self.route('PUT', (':id',), self.updateJob)
+        self.route('GET', (':id', 'child',), self.getChildJobs)
```
kotfic: should this be 'children'  instead of 'child'?

plugins/jobs/server/job_rest.py
```diff
@@ -33,6 +33,8 @@ def __init__(self):
...
+        self.route('PUT', (':id', 'parent',), self.setParentJob)
```
kotfic: Let's remove this route for now and implement it in https://github.com/girder/girder/pull/2031 once we have a better idea on how we want to handle access control for an endpoint like this.

cjh1: I would also expect setParentJob(...) be a single shot thing ( as in real life :smile: ), once you have a parent, that is it!

plugins/jobs/server/job_rest.py
jbeezley: As discussed in person, a more consistent way of getting at "child resources" would be to add a `parentId` parameter to the `GET /job` endpoint.  This would follow the conventions of other model types.

dorukozturk:
@kotfic @cjh1 @jbeezley 
These are the changes that I made based on you reviews:

1) Removed the setParentJob endpoint.
2) Refactored getChildJobs endpoint. Now child jobs are getting filtered by the job endpoint.
3) Refactored the tests to make them smaller and added some extra tests.
4) Cannot overwrite the parentId once it is set now

plugins/jobs/server/job_rest.py
```diff
@@ -58,10 +61,13 @@ def listJobs(self, userId, types, statuses, limit, offset, sort, params):
         else:
             user = self.model('user').load(
                 userId, user=currentUser, level=AccessType.READ)
+        if parentId:
+            parentId = self.model('job', 'jobs').load(parentId, force=True)['_id']
```
kotfic: Can we do this with .modelParam() as apart of autoDescribeRoute? Thanks

plugins/jobs/server/models/job.py
```diff
@@ -55,8 +56,17 @@ def _validateStatus(self, status):
             raise ValidationException(
                 'Invalid job status %s.' % status, field='status')
 
+    def _validateParent(self, parentJobId):
+        self.model('job', 'jobs').load(parentJobId, force=True)
```
kotfic:
```python
try:
     self.model('job', 'jobs').load(parentJobId, force=True, exc=True)
except ValidationException:
     raise ValidationException("Parent Id is not valid" % parentJobId)

```

dorukozturk:
@kotfic @zachmullen PTAL

I changed everything that you have requested.

zachmullen:
> 100644 → 100755

Were the file mode changes here intentional? If not, let's change them back.

plugins/jobs/server/models/job.py
```diff
@@ -55,8 +56,20 @@ def _validateStatus(self, status):
             raise ValidationException(
                 'Invalid job status %s.' % status, field='status')
 
+    def _validateParent(self, parentJobId):
+        try:
+            self.model('job', 'jobs').load(parentJobId, force=True, exc=True)
+        except ValidationException:
+            raise ValidationException('%s is not a valid Job Id' % parentJobId)
```
zachmullen: Is there a good reason to catch and re-raise with a different error message? The default one is "No such job: [id]", which seems just as good to me.

plugins/jobs/server/models/job.py
```diff
@@ -151,7 +167,7 @@ def createLocalJob(self, module, function=None, **kwargs):
 
     def createJob(self, title, type, args=(), kwargs=None, user=None, when=None,
                   interval=0, public=False, handler=None, async=False,
-                  save=True, otherFields=None):
+                  save=True, parentId=None, otherFields=None):
```
zachmullen: I think a more common pattern is to pass in the parent document rather than just its ID, which removes the need to do a `load` on the parent inside this method.

plugins/jobs/server/models/job.py
```diff
@@ -513,3 +534,30 @@ def getAllTypesAndStatuses(self, user):
         types = self.collection.distinct('type', query)
         statuses = self.collection.distinct('status', query)
         return {'types': types, 'statuses': statuses}
+
+    def setParentJob(self, job, parentId):
```
zachmullen: Same here, if we pass in the parent document rather than the ID, it leaves it up to the caller to load the document, which they most likely have already done anyway.

plugins/jobs/server/models/job.py
```diff
@@ -55,8 +56,14 @@ def _validateStatus(self, status):
...
 
+    def _validateChild(self, parentJob, childJob):
+        if str(parentJob['_id']) == str(childJob['_id']):
+            raise ValidationException('Child Id cannot be equal to Parent Id')
+        if childJob['parentId']:
+            raise ValidationException('Cannot overwrite the Parent Id')
+
     def list(self, user=None, types=None, statuses=None,
-             limit=0, offset=0, sort=None, currentUser=None):
+             limit=0, offset=0, sort=None, currentUser=None, parentId=None):
```
zachmullen: Let's also change this one to `parentJob` for consistency.

zachmullen:
job.py still has the permission change applied

dorukozturk:
I am using Prelude with Emacs and they have a functionality which changes the mode to executable if there is a shebang in a Python file.  Dan helped me figure this out which was very annoying. Sorry for the dirty commits.

zachmullen:
I mean, it does make sense that if we have a shebang we should have the execute flag, but if we change it, it's best to do it consistently across the whole codebase in a separate PR

dorukozturk:
I agree 100%. My changes were done by my editor and unintentional. But in the future maybe we get rid off them since we don't execute them directly, in a separate PR. Other than that the rest looks good to you?

zachmullen:
This LGTM, would be good to have @kotfic also mark final approval before merge.

kotfic:
LGTM,  pull the lever!

PR #2049, opened 2017-05-30
Throw an error when some Model / Collection methods are called incorrectly

brianhelba:
This now ensures that these functions will always return a Promise (when they do return).


brianhelba:
@girder/developers PTAL.

PR #2050, opened 2017-05-31
Improve promise usage conventions

brianhelba:
* Don't use `$.when` to create resolved promises
  * Given that the multi-argument version of `$.then` has different behavior, `$.when` should not be used with zero or one arguments. Instead, use the more explicit construction to create resolved promises.
* Use `.done` instead of `.then` for handlers without a return value
  * `.then` creates a new promise, with the fulfillment value of the handler's return value. Using `.then` without returning a new value causes the new fulfillment value to become `undefined`; if this is desired, then an explicit `return undefined` should be used instead.
  *  In cases where the promise is not returned or chained past the last `.then`, `.done` should still be used for consistency.
* Ensure that Collection.fetch explicitly has a resolution value

### DIFF
--- a/clients/web/test/spec/browserSpec.js
+++ b/clients/web/test/spec/browserSpec.js
@@ -18,7 +18,7 @@ describe('Test the hierarchy browser modal', function () {
             if (onRestRequest) {
                 return onRestRequest.apply(this, arguments);
             }
-            return $.when(returnVal);
+            return $.Deferred().resolve(returnVal).promise();
         });
         transition = $.support.transition;
         $.support.transition = false;
@@ -99,12 +99,12 @@ describe('Test the hierarchy browser modal', function () {
             onRestRequest = function (params) {
                 if (params.path === '/user/authentication') {
                     // The return value for the initial login call
-                    return $.when(user);
+                    return $.Deferred().resolve(user).promise();
                 }
 
                 // After login return an empty array for collection fetches
                 // on the RootSelector
-                return $.when([]);
+                return $.Deferred().resolve([]).promise();
             };
 
             girder.auth.login('johndoe', 'password');
--- a/clients/web/src/auth.js
+++ b/clients/web/src/auth.js
@@ -108,9 +108,8 @@ function login(username, password, cors) {
         events.trigger('g:login', response);
 
         return response.user;
-    }, function (jqxhr) {
+    }).fail(function (jqxhr) {
         events.trigger('g:login.error', jqxhr.status, jqxhr);
-        return jqxhr;
     });
 }
 
@@ -118,13 +117,13 @@ function logout() {
     return restRequest({
         method: 'DELETE',
         path: '/user/authentication'
-    }).then(function () {
+    }).done(function () {
         setCurrentUser(null);
         setCurrentToken(null);
 
         events.trigger('g:login', null);
         events.trigger('g:logout.success');
-    }, function (jqxhr) {
+    }).fail(function (jqxhr) {
         events.trigger('g:logout.error', jqxhr.status, jqxhr);
     });
 }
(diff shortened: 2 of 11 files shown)


clients/web/test/spec/browserSpec.js
```diff
@@ -18,7 +18,7 @@ describe('Test the hierarchy browser modal', function () {
             if (onRestRequest) {
                 return onRestRequest.apply(this, arguments);
             }
-            return $.when(returnVal);
+            return $.Deferred().resolve(returnVal).promise();
```
jbeezley: I don't understand the motivation for this change.  The jquery documentation specifically this use case for created a resolved promise.  It also follows the conventions of `Promise.resolve`.

brianhelba:
My proposal (which I'm hoping this PR will provide feedback on) is to discourage using the zero- or one-argument forms of `$.when` in favor of more explicit construction, resolution, and demotion when creating pre-resolved promises.

My concern here is that the behavior of `$.when` is rather confusing, as it has essentially two different behaviors (equivalent to either `Promise.resolve` or `Promise.all`) depending on how it's called. See the superfluous usage in https://github.com/girder/girder/pull/2036

I'd contend that syntax here being a bit longer isn't even a bad thing, as creating a pre-resolved promise from scratch (as opposed to chaining onto an existing asynchronous event) is somewhat atypical, and we should want developers to be explicit about exactly what they're doing. Also note that to create a pre-rejected promise, you have no option but to use `$.Deferred().reject(val).promise()`.

I've talked to few Girder developers now who are not totally comfortable with jQuery's promise model. Until we can transition (hopefully) to use ES6 Promises in Girder 3, I think it'd be healthy to encourage a more explicit usage style. The shortcut functions seem to be confusing people. The same rationale applies to `.then` vs `.done`, as some Girder developers weren't previously understanding the differences in their behavior.

jbeezley:
I see your motivation.  In this case while the meaning of `$.when` may be confusing, it is still an official use of JQuery that hasn't been deprecated.  On the other hand, it isn't a huge burden, so I can see the benefit.  I think I land on the side of not making the change, but I won't stand in the way if others feel strongly about it.

In regards to the other changes in this PR, I have tended to prefer using one of the following in code I write:
```javascript
promise.then(onResolve, onError);
promise.then(onResolve)
  .then(undefined, onError);
```
over any of the following
```javascript
promise.done(onResolve).fail(onError);
promise.then(onResolve).catch(onError);
```
The reason being that it works with both A+ promises and any version (>=1.5) of JQuery.  The change you made looks fine, but it does mean all these lines of code will have to change if and when we switch to native promises.

zachmullen:
Clearly we need to write our own Promise library, none of these existing ones got it right!

brianhelba:
@jbeezley A final motivation is that I intend to be working on adding promise returns to many of Girder's model and collection methods. In that new code, my strong preference would be to use `$.Deferred().resolve(val).promise()` when necessary. I think the enhancements to make Girder use promises more holistically will be better received if we have a consistent convention everywhere. Indeed, we've now got both jQuery Promises and ES6 Promises (which I'm working to resolve in https://github.com/girder/girder/pull/2026) due to the previous lack of standards around Promise usage.

As for `.done` vs `.then`, I'd actually be ok in theory with simply scrapping `.done` entirely, as we'll need to do it as part of the migration to ES6. However, Girder already uses `.done` and `.fail` all over the place (in 193 places, vs 17 usages of `.then`). Since `.done` doesn't have any special rules when used for chaining (regarding return values), I think it's easier to stick with it as the consistent default, as we push the broader codebase and development community to buy into promises. `.then` is also not Promises/A+ compliant until jQuery3, so I'd strongly suggest at least waiting until we upgrade before using it more than absolutely necessary.

PR #2053, opened 2017-05-31
Trigger save event when updating model ACL

zachmullen:
This fixes a (maybe breaking?) change in behavior as a result
of how we update a document's ACL; until recently, we called
save() during this process if save was set to True, which
caused the save events for the model to trigger.

Specifically, this broke some expectations of COVALIC where we
were expecting the ACL updates to trigger save events.


manthey:
Do you want a test to check for the new events?

zachmullen:
The COVALIC test is now passing, that's good enough for me :)

PR #2055, opened 2017-05-31
Use core-js from npm in the testing environment

brianhelba:
`core-js` is a dependency of Babel (and could be fetched from npm, regardless). There's no reason to have a static copy of it in our repo.

Also, move some test resources into Grunt, to make the test environment HTML simpler and more robust.


brianhelba:
@girder/developers PTAL.

Issue #1896, opened 2017-03-30
Support download/upload of multiple entities in girder-client


danlamanna:
It would be nice to do something like:
```
girder-cli download [PARENT_ID...]
```
or
```
girder-cli upload PARENT_ID [folder|file...]
```

manthey:
To that end, rather then using PARENT_ID, it would be nice to specify the resource via the path routing we use in the SFTP client.  That way you could say `girder-cli download user/manthey/Public/dataset`

jcfr:
(Please, discard that xref, I referenced the wrong issue.)

zachmullen:
@danlamanna I've read this a few times but can't quite figure out what change you are proposing. Could you elaborate?

danlamanna:
I just want the ability to pass multiple arguments to upload/download. If you wanted to download 2 separate items it looks like you would have to run `girder-cli download` twice.

It might get tricky because of positional arguments, but this is typically handled with a separate usage line which includes the `-t` flag like mv/cp/ln etc.

PR #2057, opened 2017-06-01
Allow girder-client to accept multiple files for upload.

manthey:
This allows uploading files using shell wildcards, such as:
```
girder-cli upload /collection/mycollection/myfolder *.png
```

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -1522,7 +1522,8 @@ class GirderClient(object):
         create a hierarchy on the server under the parentId.
 
         :param filePattern: a glob pattern for files that will be uploaded,
-            recursively copying any file folder structures.
+            recursively copying any file folder structures.  If this is a list,
+            each item in the list will be used in turn.
         :type filePattern: str
         :param parentId: Id of the parent in Girder or resource path.
         :type parentId: ObjectId or Unix-style path to the resource in Girder.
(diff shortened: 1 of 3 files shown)


danlamanna:
re #1896

manthey:
@danlamanna It is outside of my needs at the moment, but I could add the multiple resources for download case, too.  I'd implement that as all resources are of the same parent type OR are specified as resource paths.  Oddly, we seem to only be able to download folders by resource path.

danlamanna:
@manthey No worries, I was just trying to ensure the issue<->pr relationship existed.

danlamanna:
I just realized that we're handling globs within the upload code itself. Perhaps a follow on issue/pr should be removing this since the shell will handle globs and the multiple argument code will support it.

Thanks for this PR, it's nice to be able to start using these commands with find/xargs and the like.

manthey:
We do handle globs, but because in cli.py, the `_upload` function has `@_CommonParameters(path_exists=True, ...`, we couldn't actually pass globs from the command line to the upload function.  If `path_exists` were set to `False`, then one could have quoted a glob as an argument (e.g., `girder-cli upload /collection/mycollection/myfolder '*.png'`), but this is unconventional -- it is better to let the shell do the globbing.

I could see having the internal function do globbing could still be useful for programs that import and use the upload function.

danlamanna:
> I could see having the internal function do globbing could still be useful for programs that import and use the upload function.

I disagree, but I don't think it matters since it would be breaking backwards compatibility with girder-cli (as a library). 

In general I would rather have the API surface take an iterable, in which case the user could decide to make that iterable be the return value of `glob.iglob` if they wanted.

PR #2059, opened 2017-06-01
Drop test databases at the end of the test.

manthey:
Currently, we reference a mongo database, usually one per test.  Some tests that use gridfs reference additional databases.  With mongo, storing values in a database that doesn't yet exist creates it.  We try to drop the databases as we start tests, but there are conditions where we read the database before we drop it (such as determining the static root).  By removing the databases at the end of the tests, this eliminates behavioral changes on different test runs depending on where the database persisted or not.  Also, it avoids cluttering mongo with databases when one isn't running tests.

This can be overridden with EXTRADEBUG=keepdb.

If EXTRADEBUG includes 'newdb', a failure will be triggered if the database already exists (so running tests using EXTRADEBUG='keepdb newdb' twice should always result in a failure).

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -117,7 +117,7 @@ def dropTestDatabase(dropModels=True):
     if 'girder_test_' not in dbName:
         raise Exception('Expected a testing database name, but got %s' % dbName)
     if dbName in db_connection.database_names():
-        if (dbName not in usedDBs and 'newdb' in os.environ.get('EXTRADEBUG', '').split()):
+        if dbName not in usedDBs and 'newdb' in os.environ.get('EXTRADEBUG', '').split():
             raise Exception('Warning: database %s already exists' % dbName)
         db_connection.drop_database(dbName)
     usedDBs[dbName] = True
@@ -132,7 +132,7 @@ def dropGridFSDatabase(dbName):
     """
     db_connection = getDbConnection()
     if dbName in db_connection.database_names():
-        if (dbName not in usedDBs and 'newdb' in os.environ.get('EXTRADEBUG', '').split()):
+        if dbName not in usedDBs and 'newdb' in os.environ.get('EXTRADEBUG', '').split():
             raise Exception('Warning: database %s already exists' % dbName)
         db_connection.drop_database(dbName)
     usedDBs[dbName] = True
--- a/tests/base.py
+++ b/tests/base.py
@@ -47,6 +47,7 @@ remote = cherrypy.lib.httputil.Host('127.0.0.1', 30001)
 mockSmtp = mock_smtp.MockSmtpReceiver()
 mockS3Server = None
 enabledPlugins = []
+usedDBs = {}
 
 
 def startServer(mock=True, mockS3=False):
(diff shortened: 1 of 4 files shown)


tests/base.py
```diff
@@ -99,8 +116,11 @@ def dropTestDatabase(dropModels=True):
 
     if 'girder_test_' not in dbName:
         raise Exception('Expected a testing database name, but got %s' % dbName)
-    db_connection.drop_database(dbName)
-
+    if dbName in db_connection.database_names():
+        if (dbName not in usedDBs and 'newdb' in os.environ.get('EXTRADEBUG', '').split()):
```
zachmullen: We can remove the outer parens

tests/base.py
```diff
@@ -111,7 +131,11 @@ def dropGridFSDatabase(dbName):
     :param dbName: the name of the database to drop.
     """
     db_connection = getDbConnection()
-    db_connection.drop_database(dbName)
+    if dbName in db_connection.database_names():
+        if (dbName not in usedDBs and 'newdb' in os.environ.get('EXTRADEBUG', '').split()):
```
zachmullen: same here

tests/base.py
```diff
@@ -86,6 +91,18 @@ def stopServer():
...
     cherrypy.engine.exit()
     mockSmtp.stop()
+    dropAllTestDatabases()
+
+
+def dropAllTestDatabases():
+    """
+    Unless otherwise requested, drop all test databases.
+    """
+    if 'keepdb' not in os.environ.get('EXTRADEBUG', '').split():
```
danlamanna: Should this function be generalized and parameterized with `usedDBs`, and only called in the case of `keepdb` (from `stopServer`)?

tests/base.py
```diff
@@ -635,3 +659,7 @@ def _sigintHandler(*args):
 
 
 signal.signal(signal.SIGINT, _sigintHandler)
+# If we insist on test databases not existing when we start, make sure we
+# check right away.
+if 'newdb' in os.environ.get('EXTRADEBUG', '').split():
```
danlamanna: Should we have a separate env variable for keepdb/newdb, rather than loading extra options into `EXTRADEBUG`? Maybe something like `KEEP_TEST_DATABASES`?

zachmullen: I like re-using EXTRADEBUG personally, it alleviates the need to add more CMake code for each additional debugging feature.

tests/base.py
manthey: It needs to be callable from tests that don't start the server in the ordinary manner (like the sftp test).  We already have a function `dropTestDatabase` for dropping one database which can also reload models if appropriate.

mgrauer:
Is there some documentation that would help here, maybe either in the developer's guide or around the EXTRADEBUG in CMake to say what set of things are involved?

Or is the work in this PR the sort of thing most developers can safely ignore most of the time?

manthey:
This is definitely something most developers can ignore.  I don't think most of us ever use Mongo to look in the test databases after a test is run.

PR #2061, opened 2017-06-01
Refactor how plugin dependencies are installed for npm 5

### LINKED ISSUE #2045: Plugin deps fail to install with npm v5
I saw this when submitting #2043, but it took me a while to validate what was going on.  This seems to apply only to npm v5 and only on the invocation of `girder-install web --all-plugins`.  The output from the initial install is [here](https://gist.github.com/jbeezley/d40fc53421cbb05852c3c788533ef85e).  After running, the dependencies (vega in this case) are not installed in `node_modules`.  Here, the failure is in the jobs plugin, but similar behavior appears for any plugin installing extra dependencies.

What is weird is that I have found that you can run `girder-install web --plugins jobs` to fix the error building the jobs plugin.  After that, running with `--all-plugins` will fail on the ldap plugin due to the missing sortablejs.  And again, you can fix it by building that plugin alone.  To get, `--all-plugins` to succeed, you have to first install each plugin individually.

jbeezley:
@girder/developers This is a fairly invasive change, but I attempted to maintain compatibility.  The core of the change is to group all of the `npm install` commands into a single grunt task which is executed before all others.  There is some fairly complex logic that goes into how and where the dependencies are installed, and I think I covered them, but it would be worthwhile for others to test this change on their plugins to be sure.

Fixes #2045

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -514,6 +514,11 @@ module.exports = function (grunt) {
         if (buildPlugin) {
             processPluginDependencies(plugin, dir, config);
 
+            // Make sure the npm task runs before plugin build tasks
+            grunt.config.set(`default.plugin:${plugin}`, {
+                dependencies: ['npm-install-plugins']
+            });
+
             // For backward compatibility, we still generate the old 'npm-install'
             // task as a noop, but ensure that the real npm install task is
             // executed as a prerequisite.
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -537,8 +537,11 @@ module.exports = function (grunt) {
         configurePluginForBuilding(path.resolve(grunt.config.get('pluginDir'), name), true);
     });
 
-    // This is noop that exists for backwards compatibility in case any plugins rely
-    // on its existence.
+    /**
+     * This task is noop that exists for backwards compatibility in case any plugins rely
+     * on its existence.
+     * @deprecated: remove in v3
+     */
     grunt.registerTask(
         'npm-install', 'Install plugin NPM dependencies (deprecated)',
         _.constant(true)
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -104,7 +104,7 @@ module.exports = function (grunt) {
         });
 
         if (child.status !== 0) {
-            throw new Error(`"npm ${args.join(' ')}" failed`);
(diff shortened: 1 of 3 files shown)


zachmullen:
Thanks for tackling this so quickly!

grunt_tasks/plugin.js
```diff
@@ -61,6 +61,168 @@ module.exports = function (grunt) {
...
+                })
+            );
+        }
+
+        var child = child_process.spawnSync('npm', args, {
+            stdio: 'inherit'
+        });
+
+        if (child.status !== 0) {
+            throw new Error(`"npm ${args.join(' ')}" failed`);
```
zachmullen: I'm run into a case where throwing a new Error inside one of these grunt modules does not actually cause a build failure, but I can't remember if it was at config time or task run time. Does this cause the build to fail? If not, `grunt.fail()` will work.

jbeezley: That's a good point.  I'll check.

grunt_tasks/plugin.js
```diff
@@ -468,6 +537,26 @@ module.exports = function (grunt) {
         configurePluginForBuilding(path.resolve(grunt.config.get('pluginDir'), name), true);
     });
 
+    // This is noop that exists for backwards compatibility in case any plugins rely
+    // on its existence.
+    grunt.registerTask(
```
zachmullen: Let's add `@deprecated: remove in v3` here. I grep -i for "deprecated" to find things to remove :)

zachmullen:
This works in my testing. It still prunes the tree, so there is some performance downside to rapidly switching between building different sets of plugins, but that's not strictly a problem.

jbeezley:
I suspect that npm will add an option to avoid autoprunning eventually, but for now we just have to deal with it.  Thankfully, installation is quite a bit faster now.

zachmullen:
> Thankfully, installation is quite a bit faster now.

Awesome -- will we get the benefits of this caching in CircleCI already, or do we need to add additional cache dirs?

zachmullen:
Also, should we switch CircleCI to use npm5? If it's easy, I'd vote to do it.

jbeezley:
That seems like a good idea.  It follows how we test with the latest version of pip as well.  The cache is in a different place now, but it is still in `~/.npm`, so circle should still save it.

jbeezley:
@girder/developers Since I haven't heard of any breakages, I'll go ahead and merge.  Let me know if there are any issues.

PR #2063, opened 2017-06-02
Upgrade to jQuery 3

### LINKED ISSUE #1892: Review and address js issues reported by snyk
This report is related to current master.

See https://snyk.io/test/github/girder/girder

brianhelba:
Fixes #1892

* Retrieve HTML data attributes with camel case
  * Per the HTML spec, data attributes should be set on elements as kebab-case attributes, but stored and accessed in Javascript via transformed camel-case variables.
  * https://www.w3.org/TR/html5/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes
* Upgrade to jQuery 3
* Utilize Promise/A+ compliant behavior of `.then`
  * Since thrown values in .then or .catch are converted into Promise rejection values (which may legitimately contain any value), the eslint `no-throw-literal` rule must be disabled now.
* Refactor `RootSelectorWidget` test to respect async events
  * As documented in new code comments, most of the lifecycle of the `RootSelectorWidget` occurs asynchronously. Although the `browserSpec` tests explicitly disables Bootstrap transitions, tests should still not assume that events will occur instantly and synchronously. In jQuery3, Ajax callbacks happen truly asynchronously, in the main event loop queue, so these tests would not pass.
* Fix broken tests, now that events may be truly asynchronous
* Add new static analysis rules for promises
* Fix the thumbnails test to wait for some asynchronous events to complete
* Attempt to fix possible race conditions in the `jobsSpec` test
* Fix race condition in `web_client_empty_layout` test
  * Just because a new layout is loaded, doesn't mean the `CollectionsView` has finished rendering itself, as that depends on the collection list to fetch.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,8 @@
 {
     "extends": "girder",
     "rules": {
-        "no-throw-literal": "off"
+        "no-throw-literal": "off",
+        "promise/always-return": "error",
+        "promise/no-nesting": "error"
     }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,3 +1,6 @@
 {
     "extends": "girder",
+    "rules": {
+        "no-throw-literal": "off"
+    }
 }
--- a/clients/web/src/views/body/SystemConfigurationView.js
+++ b/clients/web/src/views/body/SystemConfigurationView.js
@@ -34,7 +34,7 @@ var SystemConfigurationView = View.extend({
                     return {
                         key: key,
                         value: _.object(_.map($('.g-core-route-table'), function (el) {
-                            return [$(el).data('webroot-name'), $(el).val()];
+                            return [$(el).data('webrootName'), $(el).val()];
                         }))
                     };
                 }
--- a/clients/web/test/testUtils.js
+++ b/clients/web/test/testUtils.js
@@ -632,17 +632,11 @@ girderTest.promise = $.Deferred().resolve().promise();
  * Import a javascript file and.
  */
 girderTest.addScript = function (url) {
-    var defer = new $.Deferred();
-    girderTest.promise.done(function () {
-        $.getScript(url).done(function () {
-            defer.resolve();
-        }).fail(function () {
-            defer.reject('Failed to load script: ' + url);
+    girderTest.promise = girderTest.promise
+        .then(_.partial($.getScript, url))
+        .catch(function () {
+            throw 'Failed to load script: ' + url;
         });
-    }).fail(function () {
-        defer.reject.apply(defer, arguments);
-    });
-    girderTest.promise = defer.promise();
 };
 
 /**
--- a/clients/web/src/server.js
+++ b/clients/web/src/server.js
@@ -73,13 +73,11 @@ restartServer._checkServer = function (lastStartDate) {
         error: null
     })
     .then((resp) => {
-        const checkResolution = $.Deferred();
         if (resp.serverStartDate !== lastStartDate) {
-            checkResolution.resolve();
+            return undefined;
         } else {
-            checkResolution.reject();
+            throw undefined;
         }
-        return checkResolution.promise();
     });
 };
 
--- a/clients/web/test/spec/browserSpec.js
+++ b/clients/web/test/spec/browserSpec.js
@@ -33,28 +33,61 @@ describe('Test the hierarchy browser modal', function () {
     describe('root selection', function () {
         it('defaults', function () {
             returnVal = [];
-            var view = new girder.views.widgets.RootSelectorWidget({el: testEl, parentView: null});
-            view.render();
-            var select = view.$('select#g-root-selector');
-            expect(select.length).toBe(1);
-            expect(select.find('option:eq(0)').text()).toBe('Select a root...');
-            expect(select.find('optgroup[label="Collections"]').length).toBe(1);
-            expect(select.find('optgroup[label="Users"]').length).toBe(1);
+
+            var view;
+            runs(function () {
+                view = new girder.views.widgets.RootSelectorWidget({
+                    el: testEl,
+                    parentView: null
+                });
+                // RootSelectorWidget will self-render, as soon as its initial 'fetch()' completes
+
+                // We should be able to attach a spy to 'view.render' without a race condition, as
+                // long as we do it before returning to the main event loop (where async results
+                // from 'fetch()' might already be pending)
+                spyOn(view, 'render').andCallThrough();
+            });
+
+            waitsFor(function () {
+                // 'view.render' will be called once for each of 'view.groups', and 'view.groups'
+                // has 2 collections here, so rendering should only be considered finished once all
+                // complete
+                return view.render.callCount >= 2;
+            });
+
+            runs(function () {
+                var select = view.$('select#g-root-selector');
+                expect(select.length).toBe(1);
+                expect(select.find('option:eq(0)').text()).toBe('Select a root...');
+                expect(select.find('optgroup[label="Collections"]').length).toBe(1);
+                expect(select.find('optgroup[label="Users"]').length).toBe(1);
+            });
         });
 
         it('display order', function () {
             returnVal = [];
-            var view = new girder.views.widgets.RootSelectorWidget({
-                el: testEl,
-                parentView: null,
-                display: ['Users']
+
+            var view;
+            runs(function () {
+                view = new girder.views.widgets.RootSelectorWidget({
+                    el: testEl,
+                    parentView: null,
+                    display: ['Users']
+                });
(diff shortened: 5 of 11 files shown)


clients/web/src/views/body/SystemConfigurationView.js
```diff
@@ -34,7 +34,7 @@ var SystemConfigurationView = View.extend({
                     return {
                         key: key,
                         value: _.object(_.map($('.g-core-route-table'), function (el) {
-                            return [$(el).data('webroot-name'), $(el).val()];
+                            return [$(el).data('webrootName'), $(el).val()];
```
zachmullen: This doesn't appear to be related to the rest of this, where is this referenced?

brianhelba: See the commit message. I'll add that info to the PR description, once this is ready for review.

brianhelba:
@girder/developers PTAL

This is actually a pretty painless upgrade, and I'm not anticipating that downstream compatibility will be affected in any breaking way.

A few notes:
* Some promise callbacks now fire as truly asynchronous calls, directly from the main event loop. This broke some tests, which were (incorrectly) assuming that callbacks would fire via callback dispatching on the same stack that they were triggered, particularly in cases where Ajax functionality was mocked by the tests (which also differs from actual usage).
* [Breaking change: Behavior of :hidden and :visible](https://jquery.com/upgrade-guide/3.0/#breaking-change-behavior-of-hidden-and-visible) could theoretically have an impact, but even after searching, I didn't find any places that depended on the specific behavior here.
* [Breaking change: .show(), .hide(), and .toggle() methods now respect more stylesheet changes](https://jquery.com/upgrade-guide/3.0/#breaking-change-show-hide-and-toggle-methods-now-respect-more-stylesheet-changes) this could also theoretically have an impact, but I can't find any places that depend on this behavior, particularly after https://github.com/girder/girder/pull/2035 went in

I've attempted to update any other deprecated behavior via other PRs: 
* https://github.com/girder/girder/pull/2032
* https://github.com/girder/girder/pull/2033
* https://github.com/girder/girder/pull/2035
* https://github.com/girder/girder/pull/2036

.eslintrc
```diff
@@ -1,3 +1,6 @@
 {
     "extends": "girder",
+    "rules": {
+        "no-throw-literal": "off"
```
jbeezley: Is this only for the `throw undefined` in server.js?

zachmullen: This is a good question, @brianhelba could you link to some docs about the use of `throw undefined` in jquery promises? I am having trouble finding them with google.

jbeezley: Throwing an error in a promise will cause it to be rejected.  This is actually one of the incompatible changes in jquery 3 to conform to the A+ specification.  I think @brianhelba is throwing `undefined` there to be compatible with the previous behavior (the promise was rejected with no argument); however, I don't think there is any real chance anyone would rely on that behavior.  If that is indeed the motivation, I would prefer to just throw an actual error instead.

brianhelba: I'll respond down in the main comments.

brianhelba:
I'm not particularly happy about having to globally disable the `no-throw-literal` rule. However, `throw`ing values within a `.then` handler is now the correct way to cause the returned Promise to have a rejection value (to be clear, this thrown rejection value is not received by us in a `catch`, but rather as the argument to a `.fail` handler). Since rejection values can be anything (string, object, array, etc.) and will be received via a normal argument in a `.fail` handler, I don't think it's reasonable to insist that they are always an [`Error` object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error). Since ESLint does not provide a mechanism to except `throw` inside a `.then` function for `no-throw-literal`, we have to disable the rule globally.

For full disclosure, there is another school of thought, manifested as the [`prefer-promise-reject-errors` ESLint rule](http://eslint.org/docs/rules/prefer-promise-reject-errors), which does try to insist that all rejection values be wrapped in `Error`. However, to the best of my knowledge, wrapping a non-string value inside an `Error` makes the original value irretrievable, as [`Error.prototype.message`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message) always returns a string (even if it's `object Object`). Additionally, if we went this route, all of our existing promises that resolve to a rejection value (i.e. anything we attach a `.fail` (previously `.error`) to) should be re-wrapped within an `Error` object, or else it'd be very difficult to predict what the rejection type of chained `.then` Promises would be, which defeats many of the benefits of being able to chain Promises via `.then`.

As for using `throw undefined` within a `.then` at all, this results in the exact same thing as `return $.Deferred.reject().promise()`, just with an easier (and now standards-compliant) syntax. Since `.then` always results in the creation of a new Promise, it's a good practice (soon to be enforced by [an ESLint rule, `promise/always-return`](https://github.com/xjamundx/eslint-plugin-promise)) to always `return` or `throw` something from `.then`, even if it's `undefined`. If you don't want to create a new Promise via return value of your handler, you should use `.done`, `.fail`, or `.always` instead, which only execute for some side effects, and return the original promise for chaining.

jbeezley:
I see.  Being able to reject with an arbitrary value seems a good enough reason to disable the rule.

zachmullen:
LGTM, but there is ostensibly a new race condition in the web_client_thumbnails test now.

brianhelba:
Yes, I will be re-running these tests a few times, to see if I can make the web_client_thumbnails test fail again. I'm unable to reproduce it locally.

brianhelba:
@zachmullen I believe that I've fixed the issue, which was a bug with the test. PTAL at my latest commit.

zachmullen:
`web_client_routing` is failing now, which I hadn't seen in a while...

brianhelba:
The cause on that one is a modal dialog not being hidden. I haven't seen that failure before on this branch, and my intuition is that it's not related to this change. I'm going to run it again a few times.

zachmullen:
We used to see that same failure all the time, but hadn't for a while, we put in some kludge to workaround it a while back.

brianhelba:
I still have no clue about `web_client_routing`, though it just passed on the last run.

[`web_client_jobs.jobs` just broke](http://my.cdash.org/testDetails.php?test=32147445&build=1223196), but I've added a commit in the place that it failed, which should hopefully make the test a bit safer.

Unfortunately, I'm not able to reproduce any of these failures locally. I'll keep re-running, and adding fixes in the places that I can.

As an aside, perhaps a "nice" side effect of the fact that more promise handlers execute asynchronously in jQuery3 is that more of our tests that are mocking out asynchronous calls will start to fail appropriately, which actually just reveals bugs in the tests themselves (as these places would always fail in a real environment).

jcfr:
@brianhelba Thanks for working on this. :+1:  

Based on snyk, all critical issues impacting client side have been addressed, see https://snyk.io/test/github/girder/girder/jquery3  (what are left are *issues* with ` grunt-contrib-watch` and ` grunt-fontello`)

brianhelba:
Just fixed another race condition in the web_client_empty_layout test

@mgrauer PTAL @ https://github.com/girder/girder/pull/2063/commits/6640ca0745d508013ab379c45f08f2114987f344 , as this touches code you were blamed for.

brianhelba:
For full disclosure, this just had [an error on `web_client_autojoin.autojoin`](http://my.cdash.org/testDetails.php?test=32150082&build=1223282), but the logs indicate it's a server-side failure. I can't reproduce locally or on CircleCI. I'm going to ignore it for now.

clients/web/test/testUtils.js
```diff
@@ -632,17 +632,11 @@ girderTest.promise = $.Deferred().resolve().promise();
  * Import a javascript file and.
  */
 girderTest.addScript = function (url) {
```
jcfr: May be worth fixing the docstring.

brianhelba: I'll fix it in another PR. I don't want any more commits at this point, as I'm trying to ensure that the tests repeatably pass.

brianhelba:
@girder/developers Ok, I'm going to merge this soon.

I've re-run the tests 10 times in a row on this same final commit. I've had failures on 2/10 times ([see here](http://my.cdash.org/viewTest.php?onlyfailed&buildid=1223933)), on the tests:
* `web_client_jobs.jobs`: This may be some sort of client issue, but I've been unable to reproduce it.
* `web_client_autojoin.autojoin`: This looks like a server error, so it's out of scope of this PR.

I'm satisfied that this branch isn't the problem *per se*, just that it may be making race conditions due to mocking in tests more likely to occur. Many test bugs have already been fixed here, and if we're able to reproduce any additional test failures, I'd propose that we fix them at that point.

PR #2065, opened 2017-06-02
Upgrade static analysis libraries, for access to new rules

brianhelba:
This is necessary to get access to new rules from `eslint-plugin-promise` (which is now part of `eslint-config-semistandard`).


brianhelba:
@jbeezley We discussed this in-person. Can you please mark it as approved?

PR #2067, opened 2017-06-05
GridFS assetstore speed improvements.

manthey:
Speed up GridFS deletes by adding the right index.  We check if a gridfs chunk is used by querying for the chunkUuid.  There was no index on this, so deletes were very slow with even a modest number of GridFS files.

Record if we have used a Mongo connection recently.  If so, don't ensure indices on it.  On my local system using a remote mongo GridFS server, ensuring indices that already exist consumes an average of around 6 ms.

### DIFF
--- a/girder/utility/gridfs_assetstore_adapter.py
+++ b/girder/utility/gridfs_assetstore_adapter.py
@@ -22,6 +22,7 @@ from hashlib import sha512
 import pymongo
 import six
 from six import BytesIO
+import time
 import uuid
 
 from girder import logger
@@ -35,6 +36,10 @@ from .abstract_assetstore_adapter import AbstractAssetstoreAdapter
 # 2MB chunks. Clients must not send any chunks that are smaller than this
 # unless they are sending the final chunk.
 CHUNK_SIZE = 2097152
+
+# Cache recent connections so we can skip some start up actions
+RECENT_CONNECTION_CACHE_TIME = 600  # seconds
+_recentConnections = {}
 
 
 def _ensureChunkIndices(collection):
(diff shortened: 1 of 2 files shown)


girder/utility/gridfs_assetstore_adapter.py
```diff
@@ -79,31 +84,45 @@ def validateInfo(doc):
...
+            # MongoClient automatically reuses connections from a pool, however
+            # getting the connection takes time, so use our own cache.  We also
+            # avoid redoing ensureChunkIndices, which is also takes time.
+            self.chunkColl = getDbConnection(*connectionArgs)[self.assetstore['db']].chunk
+            if not recent:
+                _ensureChunkIndices(self.chunkColl)
+                if key is not None:
+                    _recentConnections[key] = {
+                        'created': time.time()
+                    }
```
jeffbaumes: I cannot find a time where entries in the `_recentConnections` are removed. Will this cache continue growing until the server is restarted?

manthey: Yes.  I could add a simplistic purge when it gets to a certain size.  Since the individual savings is small, just clearing the cache when it hits a size threshold would probably be sufficient (something like `if len(_recentConnections) > MAX_CACHE_SIZE: _recentConnections.clear()`).

zachmullen:
@manthey is this change sufficient to address the recent issues our user was experiencing, or is this only a partial fix?

girder/utility/gridfs_assetstore_adapter.py
jeffbaumes: That logic for cache purging seems reasonable to me.

manthey:
This is a partial fix that seems to resolve much of the performance issue.  Additional PRs will follow, but this is the most important one.

girder/utility/gridfs_assetstore_adapter.py
```diff
@@ -79,31 +84,45 @@ def validateInfo(doc):
...
+        recent = False
+        connectionArgs = (self.assetstore.get('mongohost', None),
+                          self.assetstore.get('replicaset', None))
+        try:
+            # Guard in case the connectionArgs is unhashable
+            key = connectionArgs
+            if key in _recentConnections:
+                recent = (time.time() - _recentConnections[key]['created'] <
+                          RECENT_CONNECTION_CACHE_TIME)
+        except Exception:
```
zachmullen: If we are just catching unhashability, let's just `except TypeError`.

girder/utility/gridfs_assetstore_adapter.py
```diff
@@ -79,31 +84,45 @@ def validateInfo(doc):
...
+            key = None
         try:
-            self.chunkColl = getDbConnection(
-                self.assetstore.get('mongohost', None),
-                self.assetstore.get('replicaset', None)
-            )[self.assetstore['db']].chunk
-            _ensureChunkIndices(self.chunkColl)
+            # MongoClient automatically reuses connections from a pool, however
+            # getting the connection takes time, so use our own cache.  We also
+            # avoid redoing ensureChunkIndices, which is also takes time.
```
zachmullen: which ~~is~~ also takes time

girder/utility/gridfs_assetstore_adapter.py
```diff
@@ -79,31 +84,45 @@ def validateInfo(doc):
...
+        except Exception:
+            key = None
         try:
-            self.chunkColl = getDbConnection(
-                self.assetstore.get('mongohost', None),
-                self.assetstore.get('replicaset', None)
-            )[self.assetstore['db']].chunk
-            _ensureChunkIndices(self.chunkColl)
+            # MongoClient automatically reuses connections from a pool, however
+            # getting the connection takes time, so use our own cache.  We also
```
zachmullen: > however getting the connection takes time, so use our own cache

It doesn't look like we actually cache the connections themselves, is this accurate?

manthey: Yes.  I ran a test where I upload ~100,000 files across 10 connections onto a remote sharded mongo instance.  When I cache the connections, there *might* be a 1 or 1.5 % speed up, but the results seem a bit inconclusive.  Whereas, skipping the index checking saved around 10% of the time.

manthey: On a local Mongo non-sharded instance, remove the index check doesn't save nearly so much.

zachmullen: If we are not caching the connections, let's change that comment since it makes it sounds like we are.

PR #2070, opened 2017-06-06
Refactor RootSelectorWidget test to respect async events

brianhelba:
As documented in new code comments, most of the lifecycle of the RootSelectorWidget occurs asynchronously. Although the browserSpec tests explicitly disables Bootstrap transitions, tests should still not assume that events will occur instantly and synchronously. In jQuery3, Ajax callbacks happen truly asynchronously, in the main event loop queue, so these tests would not pass.


brianhelba:
Closing this, to just roll it into #2063 . It's too difficult to attach spies to `RootSelectorWidget.render` under both synchronous and asynchronous callbacks without a "race condition" under one of the two conditions.

PR #2071, opened 2017-06-06
Enable additional static analysis rules

brianhelba:
* Update developer documentation to use JS import syntax
* Fix eslint Backbone tests, ensuring Backbone inheritance is set correctly
* Activate a style rule for JS import ordering
* Enable additional Underscore static analysis rules
* Enable additional Backbone static analysis rules
* Enable additional Import and Promise static analysis rules
* Relax the "import/order" rule, allowing relative imports in any order

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,7 +12,7 @@
         "backbone/render-return": "error",
         "import/named": "error",
         "import/order": ["error", {
-            "groups": ["builtin", "external", "parent", "sibling", "index"],
+            "groups": ["builtin", "external", ["parent", "sibling", "index"]],
             "newlines-between": "always-and-inside-groups"
         }],
         "promise/always-return": "error",
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,6 +10,7 @@
         "backbone/no-view-model-attributes": "error",
         "backbone/no-view-qualified-jquery": "error",
         "backbone/render-return": "error",
+        "import/named": "error",
         "import/order": ["error", {
             "groups": ["builtin", "external", "parent", "sibling", "index"],
             "newlines-between": "always-and-inside-groups"
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,7 +19,6 @@
         "promise/no-nesting": "error",
         "underscore/identity-shorthand": ["error", "always"],
         "underscore/jquery-each": ["error", "never"],
-        "underscore/jquery-extend": ["error", "never"],
         "underscore/jquery-proxy": ["error", "never"],
         "underscore/no-return-value-from-each-iteratee": "error"
     },
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,14 @@
     "extends": "girder",
     "rules": {
         "no-throw-literal": "off",
+        "backbone/defaults-on-top": "error",
+        "backbone/no-changed-set": "error",
+        "backbone/no-collection-models": "error",
+        "backbone/no-el-assign": "error",
+        "backbone/no-silent": "error",
+        "backbone/no-view-model-attributes": "error",
+        "backbone/no-view-qualified-jquery": "error",
+        "backbone/render-return": "error",
         "promise/always-return": "error",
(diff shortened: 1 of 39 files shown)


grunt_tasks/build.js
```diff
@@ -15,10 +15,11 @@
  */
 
 var path = require('path');
-var _ = require('underscore');
 
+var _ = require('underscore');
```
zachmullen: Are the changes in this file meaningful in some way?

grunt_tasks/webpack.config.js
```diff
@@ -20,8 +20,8 @@
  * that can extend these.
  */
 var path = require('path');
-var webpack = require('webpack');
 
+var webpack = require('webpack');
```
zachmullen: Same question here

zachmullen:
This could cause a *lot* of style test failures for downstream external plugins. Should we tag this for 3.0 to avoid that disruption?

brianhelba:
I think this is a combination of bug fixes and new features, appropriate for a new minor release. That said, I still think it could be a good idea to relax some of these rules in specific ways. I pushed this here as a WIP branch to start evaluating the broader impact.

However, the style rules are not part of Girder's API surface, they are a tool to improve the reliability and maintainability of code. If there are "new" warnings, I see this as a feature for plugin developers to take advantage of, pointing out hazards in pre-existing parts of their code. If they don't want certain warnings, plugin developers are free to silence them via [the well-documented mechanism here].(https://girder.readthedocs.io/en/latest/plugin-development.html#customizing-static-analysis-of-client-side-code)

Note that the bug (which this fixes) in the `Fix eslint Backbone tests, ensuring Backbone inheritance is set correctly` commit meant that almost no Backbone static analysis checks have been getting run since the 2.0 release. By fixing this bug, indeed some new recently-added code did now fail the style checker.

Some new rules, (e.g. `promise/always-return`) are truly a source of dangerous bugs. I think it's critical that we enable these rules now.

In other cases, (e.g. `underscore/jquery-extend`) Girder was already following these rules implicitly, with enforcement via the community code review process. The only code that's flagged by them was added very recently by new developers who are less familiar with Girder's code practices. IMO, explicit is better than implicit.

Finally, some things, in particular. `import/order`, reflect a practice that Girder code was mostly already attempting to follow, but which we lacked any strict enforcement mechanism for. This particular rule could be a source of frustration for plugin developers, even given our good intentions. I think it's going to make sense to relax that rule configuration to allow some of the common mix-ups in ordering to pass, while still flagging the ones that basically everyone was already following.

brianhelba:
Note, that I left a number of good rules our of this PR, because adding them would indeed have caused too much code to be flagged.

clients/web/src/models/ApiKeyModel.js
```diff
@@ -22,10 +22,10 @@ var ApiKeyModel = AccessControlledModel.extend({
     save: function () {
         // Scope needs to be sent to the server as JSON
         var scope = this.get('scope');
-        this.attributes.scope = JSON.stringify(scope);
+        this.attributes.scope = JSON.stringify(scope);  // eslint-disable-line backbone/no-model-attributes
```
brianhelba: It's better, though ideally we'd have a more seamless serialization mechanism (or really just POST everything as JSON...). I'll change it.

.eslintrc
```diff
@@ -2,7 +2,39 @@
     "extends": "girder",
     "rules": {
         "no-throw-literal": "off",
+        "backbone/defaults-on-top": "error",
```
jbeezley: You may have answered this already, but what is the rational behind putting these rules here rather than in our eslint config repo?

brianhelba: This is only a staging ground, since it's much easier to discuss and apply rules locally here. Once this is approved, I'll generate another PR to `eslint-config-girder`, merge that, release that, update the dependency here in Girder's `package.json`, then remove these explicit local rules.

jbeezley: Right, now I remember.

Issue #1675, opened 2016-11-13
Add Mongo 2.6 to circle ci


mgrauer:
Follow up to #1674

danlamanna:
[Do we really want to still support 2.6?](https://www.mongodb.com/blog/post/mongodb-2-6-end-of-life)

> We are now approaching the end of life of this major version, and we will sunset support for MongoDB 2.6 at the end of October, 2016.

zachmullen:
My thought as well. It's unlikely we'd actually break compatibility with 2.6 given the mongo feature set that we use in Girder, but I'd be fine with dropping it from our official support and CI testing. If that is acceptable to people, this would just be a change in our documentation, perhaps including Dan's link as a justification.

PR #2073, opened 2017-06-06
Shard a gridfs collection and test working with shards.

manthey:
This adds the ability to ask a gridfs assetstore to shard based on chunk uuid.  As part of this, add a test for a mongo sharding server.

The web_client_admin test is not long run by itself, as recent changes have reduced the timing issues.

### DIFF
--- a/girder/utility/gridfs_assetstore_adapter.py
+++ b/girder/utility/gridfs_assetstore_adapter.py
@@ -27,6 +27,7 @@ import uuid
 
 from girder import logger
 from girder.api.rest import setResponseHeader
+from girder.external.mongodb_proxy import MongoProxy
 from girder.models import getDbConnection
 from girder.models.model_base import ValidationException
 from . import hash_state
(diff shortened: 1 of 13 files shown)


girder/utility/gridfs_assetstore_adapter.py
```diff
@@ -46,13 +47,49 @@
...
+
+    :param collection: a connection to a mongo collection.
     """
     collection.create_index([
         ('uuid', pymongo.ASCENDING),
         ('n', pymongo.ASCENDING)
     ], unique=True)
 
 
+def _setupSharding(collection, keyname='uuid'):
```
zachmullen: The default keyname is 'uuid', but in the adapter it's called 'chunkUuid', and you are using the default value when this is called.

manthey: On the file model, we store the reference in `chunkUuid`, and in the chunk collection, we call it `uuid`.  We are sharding the chunk collection, not the file collection.

zachmullen: Withdrawn!

manthey:
@zachmullen I had to make some changes to work with Mongo 3.0 (which is used in our Circle testing).  Locally, I've tested with Mongo 3.4.  Do we need to test with Mongo 2.6?

zachmullen:
If our CI environment isn't testing 2.6, we probably don't need to here.

manthey:
In which case, this is ready for review.

brianhelba:
Are we still formally supporting MongoDB 2.6?

manthey:
We list 2.6 as the minimum version in the prerequisites.  I'll at least test it manually...

mgrauer:
We discussed it [here](https://github.com/girder/girder/issues/1675) and seemed to converge on dropping support for Mongo 2.6.

If people are good with that as an explicit decision, I can rename that issue to "Drop formal support for Mongo 2.6".

manthey:
I've confirmed that the sharding tests run and pass on Mongo 2.6 (nice to have a Girder vagrant where we can just add `mongodb_version: 2.6.12` to the vars).

girder/utility/gridfs_assetstore_adapter.py
```diff
@@ -93,11 +130,11 @@ def __init__(self, assetstore):
...
         super(GridFsAssetstoreAdapter, self).__init__(assetstore)
         recent = False
-        connectionArgs = (self.assetstore.get('mongohost', None),
-                          self.assetstore.get('replicaset', None))
         try:
             # Guard in case the connectionArgs is unhashable
-            key = connectionArgs
+            key = (self.assetstore.get('mongohost', None),
+                   self.assetstore.get('replicaset', None),
+                   self.assetstore.get('shard', None))
```
zachmullen: `None` is the default for dict.get.

tests/base.py
```diff
@@ -195,12 +197,22 @@ def setUp(self, assetstoreType=None, dropModels=True):
...
                 createGridFsAssetstore(
                 name='Test', db=gridfsDbName,
                 mongohost='mongodb://127.0.0.1:27070,127.0.0.1:27071,'
                 '127.0.0.1:27072', replicaset='replicaset')
+        elif assetstoreType == 'gridfsshard':
+            gridfsDbName = 'girder_test_%s_shard_assetstore_auto' % assetstoreName
+            self.replicaSetConfig = mongo_replicaset.makeConfig(
+                port=27073, shard=True, sharddb=None)
+            mongo_replicaset.startMongoReplicaSet(self.replicaSetConfig)
+            self.assetstore = self.model('assetstore'). \
```
zachmullen: Could we remove the bacskslash and move the function name up to this line for readability

zachmullen:
LGTM other than the minor style nits. Full disclosure, I glossed over the code in `mongo_replicaset.py` -- it seems quite complicated, but since it's just testing code I'm not too worried about that.

tests/base.py
manthey: I cleaned up other places where we used the backslash.

Issue #2060, opened 2017-06-01
Allow girder-cli uploads to include "reference"


zachmullen:
request from @jcfr , e.g.

    girder-cli upload --reference=foo my_file dest_id

PR #2074, opened 2017-06-07
Allow girder-cli uploads to include reference

sutartmelson:
see issue: #2060 

@zachmullen Let me know if this is the right idea regarding the issue.

### DIFF
--- a/tests/cases/py_client/cli_test.py
+++ b/tests/cases/py_client/cli_test.py
@@ -282,6 +282,8 @@ class PythonCliTestCase(base.TestCase):
 
         # Test if reference is sent with each file upload
         fileList = os.listdir(localDir)
+        self.assertTrue(queryList)
+        self.assertTrue(fileList)
         self.assertEqual(len(queryList), len(fileList))
         for query in queryList:
             self.assertIn('reference', query)
--- a/tests/cases/py_client/cli_test.py
+++ b/tests/cases/py_client/cli_test.py
@@ -269,15 +269,21 @@ class PythonCliTestCase(base.TestCase):
         self.assertEqual(ret['exitVal'], 0)
 
         # Test uploading with reference
+        queryList = []
+
         @httmock.urlmatch(netloc='localhost', path='/api/v1/file$', method='POST')
         def checkParams(url, request):
-            query = six.moves.urllib.parse.parse_qs(url[3])
-            self.assertIn('reference', query)
-            self.assertIn('reference_string', query['reference'])
+            # Add query for every file upload request
+            queryList.append(six.moves.urllib.parse.parse_qs(url[3]))
 
         with httmock.HTTMock(checkParams):
             ret = invokeCli(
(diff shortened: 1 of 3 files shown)


zachmullen:
Yep, this looks right 👍

danlamanna:
Could you change your git author configuration to include your last name?

tests/cases/py_client/cli_test.py
```diff
@@ -267,6 +268,17 @@ def testUploadDownload(self):
                          downloadDir), username='mylogin', password='password')
         self.assertEqual(ret['exitVal'], 0)
 
+        # Test uploading with reference
+        @httmock.urlmatch(netloc='localhost', path='/api/v1/file$', method='POST')
+        def checkParams(url, request):
+            query = six.moves.urllib.parse.parse_qs(url[3])
+            self.assertIn('reference', query)
+            self.assertIn('reference_string', query['reference'])
```
zachmullen: This is good -- one thing I also like to do is add a sanity check to make sure these asserts are actually getting called, since we actually have no way of knowing based on this. So, what I do is create a mutable structure like a list outside the `checkParams` function, then inside `checkParams` do `my_list.append(query)`. Then, right below the `invokeCli` call, you can assert that the list has one entry, and move the current assertions down below that.

sutartmelson: That's a great approach, thanks for the tip!

brianhelba:
@sutartmelson This needs to be rebased before merging.

Your earlier commits also don't include your full name, and use your personal email address (which I'm not sure if you want to use). If you'd like some help fixing the earlier commits, feel free to ping me.

tests/cases/py_client/cli_test.py
```diff
@@ -267,6 +268,23 @@ def testUploadDownload(self):
...
+        def checkParams(url, request):
+            # Add query for every file upload request
+            queryList.append(six.moves.urllib.parse.parse_qs(url[3]))
+
+        with httmock.HTTMock(checkParams):
+            ret = invokeCli(
+                args + ['--reference', 'reference_string'], username='mylogin', password='password')
+
+        # Test if reference is sent with each file upload
+        for query in queryList:
```
zachmullen: This loop body won't be entered if queryList remains empty, so we should also add an assertion of the expected length of queryList.

tests/cases/py_client/cli_test.py
```diff
@@ -281,6 +281,8 @@ def checkParams(url, request):
                 args + ['--reference', 'reference_string'], username='mylogin', password='password')
 
         # Test if reference is sent with each file upload
+        fileList = os.listdir(localDir)
+        self.assertEqual(len(queryList), len(fileList))
```
zachmullen: Just as one additional sanity check to assist us in the future, let's also assert that the list is not empty.

jcfr:
Well done :+1:

PR #2076, opened 2017-06-07
Remove unused old files

brianhelba:
* Remove unused JSHint files
* Remove unused and outdated Heroku files
* Remove unused Bower file
  * Girder is not and probably should not be published on Bower.
* Remove outdated files for garant dashboard
  * Even if it was up-to-date (which it isn't) these machine-specific dashboard files probably shouldn't all be published and maintained in the upstream Girder repo.


Procfile
```diff
@@ -1 +0,0 @@
-web: python -m girder --port ${PORT}
```
zachmullen: There is a section in our deployment docs about Heroku, and it mentions these files and kind of makes it sound like they are important to deploy onto Heroku. @cpatrick might know whether these are still required.

cmake/garant_nightly.cmake
```diff
@@ -1,38 +0,0 @@
-set(CTEST_SOURCE_DIRECTORY "$ENV{HOME}/Dashboards/girder")
```
zachmullen: @jbeezley is garant nightly now defunct?

jbeezley: Sadly, I haven't gotten around to setting it up again.  If I need the files, I can get them from the history.

danlamanna:
FWIW, I'm going to be adding Beanstalk configurations to the repo soon under the directory `devops/beanstalk`, and our instructions will tell users to copy these files into their Girder checkout. I think that approach makes more sense than keeping configurations for every possible platform in our root directory.

Procfile
brianhelba: Given that the `Procfile` is so obviously out of date, I'm going to consider Heroku to now be unsupported via our own neglect, and thus mercilessly remove that section of the docs.

If anyone wants to pick up the baton and fix Heroku deployment (preferably with some sort of testing mechanism), that'd be fine, but until then, we shouldn't (IMHO) document and have cruft for a broken configuration.

brianhelba: If we do care about Heroku deployment, we should file a bug for the issue.

Regardless, right now, the config is broken, and should be removed. If anyone wants to work on the bug, they can resurrect the old docs from Git's history.

brianhelba: Ok, I've filed https://github.com/girder/girder/issues/2078 , in case anyone wants to ever come back to it.

brianhelba:
@girder/developers PTAL.

PR #2079, opened 2017-06-08
Make NodeJS 6 the minimum supported version

### LINKED ISSUE #1848: Explicitly specify NPM and Node version in docs

brianhelba:
Per https://github.com/nodejs/LTS :
* NodeJS 4 went out of active support on April 1, 2017, so only critical bugs, critical security fixes, and documentation updates will be made to it in the future.
* NodeJS 5 is not an LTS release, and currently has no support.
* NodeJS 6 is the supported LTS release, at present, until April 2018.

Per http://node.green/ , NodeJS 6.5 is the first release that generally supports the full ES6 feature set.

Accordingly, Girder should now consider NodeJS 6.5 to be the minimum supported version. Girder has adequate documentation (updated by this commit) on how to install NodeJS 6.x, and Ansible scripts to do the same. For systems without root access, the "nvm" tool can be used to install any newer version of NodeJS.

Since NodeJS 6.5 is released with npm 3.10, per https://nodejs.org/en/download/releases/ , Girder should consider that to be the minimum supported version of npm.

Fixes #1848

### DIFF
--- a/devops/ansible/roles/girder/tasks/npm-RedHat.yml
+++ b/devops/ansible/roles/girder/tasks/npm-RedHat.yml
@@ -1,23 +1,19 @@
 ---
 
-- name: NodeJS | Download repo setup script
-  # Due to issues with this potentially running on python < 2.7.9, validating
-  # certs via Python (ansible) can be an issue.
-  # We could use curl and bypass this, but instead we'll just disable cert validation
-  # and require the output match the known good checksum.
-  get_url:
-    url: https://rpm.nodesource.com/setup_4.x
-    dest: "/tmp/setup_4.x"
-    validate_certs: no
-    checksum: sha256:3359378636868f9fd9f0565a28191ed7e831e91b703b581ab9ff737a4574e5f4
-
 - name: NodeJS | Install repo
-  shell: "bash /tmp/setup_4.x"
+  yum_repository:
+    name: nodesource
+    description: Node.js Packages for Enterprise Linux 7 - $basearch
+    baseurl: https://rpm.nodesource.com/pub_6.x/el/7/$basearch
+    gpgcheck: yes
+    gpgkey: https://rpm.nodesource.com/pub/el/NODESOURCE-GPG-SIGNING-KEY-EL
+    enabled: yes
   become: yes
   become_user: root
 
 - name: NodeJS | Install package
   yum:
     name: nodejs
+    update_cache: yes
   become: yes
   become_user: root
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -11,6 +11,10 @@
     "type": "git",
     "url": "https://github.com/girder/girder.git"
   },
+  "engines": {
+    "node" : ">=6.5",
```
zachmullen: Does this mean we'll error out if a version less than 6.5 is detected?

brianhelba: [It will produce warnings.](https://docs.npmjs.com/files/package.json#engines)

brianhelba:
I'll probably need to fight with CircleCI a bit, to get it on the desired version. Pardon my rebases until I get it working.

devops/ansible/roles/girder/tasks/npm-RedHat.yml
```diff
@@ -1,23 +1,19 @@
...
-  # and require the output match the known good checksum.
-  get_url:
-    url: https://rpm.nodesource.com/setup_4.x
-    dest: "/tmp/setup_4.x"
-    validate_certs: no
-    checksum: sha256:3359378636868f9fd9f0565a28191ed7e831e91b703b581ab9ff737a4574e5f4
-
 - name: NodeJS | Install repo
-  shell: "bash /tmp/setup_4.x"
+  yum_repository:
```
danlamanna: You ran into no issues running this on CentOS with Python < 2.7.9?

brianhelba: Testing it now. That's why it's a WIP. :fearful:

brianhelba: On a fresh RHEL7 instance in AWS EC2 (which has PYthon 2.7.5), I ran:
```
sudo yum update
sudo yum install gcc python-devel openssl-devel
curl -O https://bootstrap.pypa.io/get-pip.py
sudo python get-pip.py
sudo pip install ansible
ansible-playbook npm-RedHat.yml
```
Where I modified the `npm-RedHat.yml` to make it into a standalone playbook by adding a
```
- hosts: localhost
  tasks:
    - ...
```
to the top. This served to install NodeJS v6.10.2.

So, this provisions from RHEL 7 onto RHEL 7 with no problems.

brianhelba:
@girder/developers PTAL.

brianhelba:
@danlamanna I see, thanks.

PR #2082, opened 2017-06-08
Client download fail check

sutartmelson:
Solves issue: #1860
 
If the download api request returns an error, downloadFile raises an HttpError

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -1093,7 +1093,7 @@ class GirderClient(object):
 
         url = '%sfile/%s/download' % (self.urlBase, fileId)
         req = self._requestFunc('get')(url, stream=True, headers={'Girder-Token': self.token})
-        if req.status_code != requests.codes.ok:
+        if not req.ok:
             raise HttpError(req.status_code, req.text, url, 'GET')
         with tempfile.NamedTemporaryFile(delete=False) as tmp:
             with self.progressReporterCls(
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -1090,7 +1090,7 @@ class GirderClient(object):
         progressFileName = fileId
         if isinstance(path, six.string_types):
             progressFileName = os.path.basename(path)
-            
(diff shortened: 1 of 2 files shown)


clients/python/girder_client/__init__.py
```diff
@@ -1090,9 +1090,11 @@ def downloadFile(self, fileId, path, created=None):
         progressFileName = fileId
         if isinstance(path, six.string_types):
             progressFileName = os.path.basename(path)
-        req = self._requestFunc('get')(
-            '%sfile/%s/download' % (self.urlBase, fileId),
-            stream=True, headers={'Girder-Token': self.token})
+
+        url = '%sfile/%s/download' % (self.urlBase, fileId)
+        req = self._requestFunc('get')(url, stream=True, headers={'Girder-Token': self.token})
+        if req.status_code != requests.codes.ok:
```
danlamanna: `if not req.ok`

tests/cases/py_client/lib_test.py
```diff
@@ -555,6 +555,26 @@ def mock(url, request):
...
+
+        @httmock.urlmatch(path=r'.*.*/file/.+/download$')
+        def mock(url, request):
+            resp = request.models.Response()
+            resp.status_code = 500
+            resp.text = "error"
+            return resp
+
+        # Attempt to download file to object stream, should raise HttpError
+        self.assertRaises(girder_client.HttpError, self.client.downloadFile(file['_id'], obj))
```
danlamanna: This isn't actually asserting that the exception is raised since the usage of `self.assertRaises` requires the callable and `*args` and `**kwargs`. You can grep for other usages in Girder, but I would recommend using it as a context manager:
```
with self.assertRaises(girder_client.HttpError):
    self.client.downloadFile(file['_id'], obj)
```

danlamanna: Additionally, just defining the `mock` function won't cause it to be used, but a context manager can help here as well:
```
with httmock.HTTMock(mock):
    self.client.downloadFile(file['_id'], obj)
```

PR #2083, opened 2017-06-08
Do fewer resource path lookups.

manthey:
When uploading multiple items from the CLI to a resource path, pass the tuple of items to upload to the upload function.  This allows the upload patent to only be converted from a resource path to an ID once, thus saving multiple queries.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -1522,8 +1522,8 @@ class GirderClient(object):
         create a hierarchy on the server under the parentId.
 
         :param filePattern: a glob pattern for files that will be uploaded,
-            recursively copying any file folder structures.  If this is a list,
-            each item in the list will be used in turn.
+            recursively copying any file folder structures.  If this is a list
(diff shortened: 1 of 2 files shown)


manthey:
In many instances, we upload files as individual items where we don't reuse the existing items.  Currently, we make a minimum of four rest requests per file upload (5 if we have to resolve a resource path) -- one to create the item, one to check for existing files, and two to upload the file.  When uploading large numbers of files, this is inefficient; we could have an optimized code path that just does the upload without checking for existing items, and, since the upload creates the item, we also don't have to check for existing files.  If we don't have any itemUploadCallbacks, we don't even need to know the ID of the created item.

I can add an optimized upload-to-folder method as part of this PR or as a separate one.

I'll note the motivation for this change is that a customer is trying to upload a large number of items to a GridFS assetstore and saw substantially worse performance than using Mongo's own GridFS implementation.  I'm not sure how they are uploading tens of thousands of files (via girder-client, the web UI, or their own rest calls), but this is the path I've been testing, so I've noticed some inefficiencies here.

PR #2085, opened 2017-06-08
Reduce rest calls when uploading new files.

manthey:
When uploading files as new items, don't bother creating items beforehand and then checking if the file already exists in the item.  This skips two rest calls per file upload in the case where items are not being reused and there are no upload callbacks, substantially reducing overhead when uploading a large number of files.

In one test, this saves 14 minutes when uploading 100,000 files (around 9ms per file transfer), which was roughly 18% of the total upload time (files averaging around 1 Mb and being stored to an external GridFS assetstore).

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -877,7 +877,7 @@ class GirderClient(object):
     def uploadFileToFolder(self, folderId, filepath, reference=None, mimeType=None, filename=None,
                            progressCallback=None):
         """
-        Uploads a file to an folder, creating a new item in the process.  If
+        Uploads a file to a folder, creating a new item in the process.  If
         the file has 0 bytes, no uploading will be performed, and no item will
         be created.
 
@@ -1084,7 +1084,7 @@ class GirderClient(object):
 
     def addMetadataToFolder(self, folderId, metadata):
         """
-        Takes an folder ID and a dictionary containing the metadata
+        Takes a folder ID and a dictionary containing the metadata
 
         :param folderId: ID of the folder to set metadata on.
         :param metadata: dictionary of metadata to set on folder.
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -876,8 +876,8 @@ class GirderClient(object):
                            progressCallback=None):
         """
         Uploads a file to an folder, creating a new item in the process.  If
-        the has 0 bytes, no uploading will be performed, and no item will be
-        created.
+        the file has 0 bytes, no uploading will be performed, and no item will
+        be created.
 
         :param folderId: ID of parent folder for file.
         :param filepath: path to file on disk.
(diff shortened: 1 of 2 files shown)


clients/python/girder_client/__init__.py
```diff
@@ -872,6 +872,57 @@ def uploadFileToItem(self, itemId, filepath, reference=None, mimeType=None, file
         with open(filepath, 'rb') as f:
             return self._uploadContents(obj, f, filesize, progressCallback=progressCallback)
 
+    def uploadFileToFolder(self, folderId, filepath, reference=None, mimeType=None, filename=None,
+                           progressCallback=None):
+        """
+        Uploads a file to an folder, creating a new item in the process.  If
+        the has 0 bytes, no uploading will be performed, and no item will be
```
zachmullen: the file* has

tests/cases/py_client/lib_test.py
```diff
@@ -438,6 +438,17 @@ def processEvent(event):
         file = self.client.uploadFileToItem(item['_id'], testPath)
         self.assertEqual(file['mimeType'], 'text/plain')
 
+        # Test uploading to a folder
+        self.client.uploadFileToFolder(
+            str(self.publicFolder['_id']), path, reference='test4_reference')
+        while (not events.daemon.eventQueue.empty() and
+                time.time() - starttime < 5):
+            time.sleep(0.05)
```
zachmullen: There's a synchronous event we can use to test this, `'model.file.finalizeUpload.after'`. The event receives both the 'upload' and 'file' documents as keys.

zachmullen: Looking at this loop condition which you copied from above, it's not really what we want since it suffers from the ABA problem. We should really be waiting on the state of `eventList`. Could you change it in all the places it's used, or simply use the synchronous event everywhere?

manthey: Will do.

clients/python/girder_client/__init__.py
```diff
@@ -874,6 +874,57 @@ def uploadFileToItem(self, itemId, filepath, reference=None, mimeType=None, file
         with open(filepath, 'rb') as f:
             return self._uploadContents(obj, f, filesize, progressCallback=progressCallback)
 
+    def uploadFileToFolder(self, folderId, filepath, reference=None, mimeType=None, filename=None,
+                           progressCallback=None):
+        """
+        Uploads a file to an folder, creating a new item in the process.  If
```
ronichoudhury: "an folder"

manthey: Thanks.

clients/python/girder_client/__init__.py
```diff
@@ -874,6 +874,57 @@ def uploadFileToItem(self, itemId, filepath, reference=None, mimeType=None, file
         with open(filepath, 'rb') as f:
             return self._uploadContents(obj, f, filesize, progressCallback=progressCallback)
 
+    def uploadFileToFolder(self, folderId, filepath, reference=None, mimeType=None, filename=None,
```
danlamanna: @manthey I'm late to the party, but is there anything precluding `filepath` from being a file-like object similar to other methods in girder-client? I'm dealing with in-memory bytes that I want to upload to Girder as part of a Celery task, so it would be nice to avoid writing it to disk first.

manthey: @danlamanna Currently it doesn't allow that, but it would be easy to change.  We'd want to change both `uploadFileToItem` and the new `uploadFileToFolder`.  The internal function `_uploadContents` takes a file-like object, but doesn't have the logic for creating and updating items.  Also, if we aren't given a mime-type, we try to determine it from the path, so we would end up not being able to do this (though we could use the optional filename for that purpose).

PR #2086, opened 2017-06-09
Add a treeview plugin

jbeezley:
This plugin adds new general purpose views for browsing Girder as a tree using a jquery plugin call [jstree](https://www.jstree.com/).  The majority of the code here is to define a unified interface around Girder model types that is necessary to use it with jstree.  The [types](https://github.com/girder/girder/blob/e0322dd769ff7b3babe5e94efbd1e58990a18cca/plugins/treeview/web_client/types.js) module defines a registry for known model types telling jstree how to load models from the server, normalize the output, and fetch children.  This is designed to be extensible by other plugins by calling the `register` method.  All of the builtin types are [defined](https://github.com/girder/girder/tree/e0322dd769ff7b3babe5e94efbd1e58990a18cca/plugins/treeview/web_client/builtin) using this registry mechanism.

I have this marked as a WIP because it is missing some important features.  However because there is a lot of code here, I wanted to get eyes on it now that it is a "minimally viable" feature.  There are many features I would like to add as a part of this (or future) PR's.

* [ ] Integration testing on the views (depends on #2040)
* [ ] Add plugin documentation
* [ ] Handle pagination (a "show more..." node)
* [ ] Support mutation (rename, create, move, copy, upload, access control, ...)
* [ ] Expose more features from jstree (check boxes, context menus, sorting, searching, events, ...)
* [ ] Add a toolbar to the dialog for interaction (as an alternative to standard keyboard shortcuts)
* [ ] Support drag and drop

test


brianhelba:
I'd like to take a look at this code too. Can you ping me, @jbeezley , once the tests are passing, but before merging?

jbeezley:
@brianhelba I have the CI passing, but I still want to get at least one integration style test in before merging.

Issue #2089, opened 2017-06-09
Girder web core can't be built without access to Mongo


danlamanna:
Our client side code can't be built without access to a mongo database unless passing the `--all-plugins` flag or explicitly specifying the enabled plugins with `--plugins=x,y,z`. Sometimes it's desirable to perform the web build with no additional plugins or access to a mongo database, a workaround is to run `girder-install web --plugins=,` and ignore the warning message.

zachmullen:
I consider it a bug that `--plugins=` doesn't work, and even better would be an explicit `--core-only` option that does the same thing.

brianhelba:
What's the use case for building the client without access to MongoDB? I don't know if this is a case we actually want to commit to supporting, as I could see DB access being necessary for more sophisticated client-side build-time features. (For instance, there are several benefits to having Webpack build the root HTML file too, rather than it being a Python template. In this case, we'd need API root info, perhaps some version info, etc. available at build time. Some of this could come from a config file, but I wouldn't want to commit to this.)

+1 for a sane way to force-build only the core though.

manthey:
At the very least, building without a MongoDB is useful for making Docker images that can be deployed and updated against a database external to Docker.

brianhelba:
Ok, maybe this is overkill, but would it make sense for our web client build system to use the HTTP API to get any and all information that it needs (enabled plugins right now, other stuff in the future). This has the benefit of removing a point of backdoor database access in the web build system. It also potentially allows more flexible build environments, as it should be easier and more reasonable to expose a public HTTP port to a Docker (or similar) environment, rather than a localhost-only-bound MongoDB port.

manthey:
When Girder is running, we *require* access to a Mongo database.  This is trying to allow building a certain way when Girder is *not* running, and *cannot* run because no database is present.  We use Girder's settings model during the install when we want to query a setting.  This accesses the Mongo database as Girder (the database may or may not be local).

brianhelba:
I understand how the build process connects now. My idea (perhaps a bad one) would indeed require an instance of girder-server to be running (which is strictly a greater requirement than just MongoDB) during build-time, in exchange for perhaps some theoretically greater decoupling between the server layers and the client build process.

I am, however, admittedly not totally clear on what the issue is when building a Docker environment. Is it not possible to spin up and connect a temporary MongoDB instance (or in my proposal, a server instance) to the Docker build process?

[...]

PR #2091, opened 2017-06-12
Build Girder web core without Mongo access

sutartmelson:
See issue: #2089 

`--plugins=` as well as `--plugins=,` now results in girder web being built with no additional plugins.
`--core-only` has the same result.

Neither require access to Mongo.

### DIFF
--- a/girder/utility/install.py
+++ b/girder/utility/install.py
@@ -123,7 +123,7 @@ def _pipeOutputToProgress(proc, progress):
 
 
 def runWebBuild(wd=None, dev=False, npm='npm', allPlugins=False, plugins=None, progress=None,
-                coreOnly=False):
+                noPlugins=False):
     """
     Use this to run `npm install` inside the package. Also builds the web code
     using `npm run build`.
@@ -139,10 +139,10 @@ def runWebBuild(wd=None, dev=False, npm='npm', allPlugins=False, plugins=None, p
     :type plugins: str, list or None
     :param progress: A progress context for reporting output of the tasks.
     :type progress: ``girder.utility.progress.ProgressContext`` or None
-    :param coreOnly: Enable this to build the girder web with no additional plugins.
-    :type coreOnly: bool
+    :param noPlugins: Enable this to build the girder web with no additional plugins.
+    :type noPlugins: bool
     """
-    if coreOnly:
+    if noPlugins:
         plugins = []
     elif isinstance(plugins, six.string_types):
         plugins = plugins.split(',')
@@ -211,7 +211,7 @@ def install_web(opts=None):
     else:
         runWebBuild(
             dev=opts.development, npm=opts.npm, allPlugins=opts.all_plugins,
-            plugins=opts.plugins, coreOnly=opts.core_only)
+            plugins=opts.plugins, noPlugins=opts.no_plugins)
 
 
 def install_plugin(opts):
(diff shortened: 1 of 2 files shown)


girder/utility/install.py
```diff
@@ -138,8 +139,12 @@ def runWebBuild(wd=None, dev=False, npm='npm', allPlugins=False, plugins=None, p
     :type plugins: str, list or None
     :param progress: A progress context for reporting output of the tasks.
     :type progress: ``girder.utility.progress.ProgressContext`` or None
+    :param coreOnly: Enable this to build the girder web with no additional plugins.
+    :type coreOnly: bool
     """
-    if isinstance(plugins, six.string_types) and plugins:
+    if coreOnly:
+        plugins = ['']
```
zachmullen: Let's change the conditional on line 74 of this file to `if plugins is None`, that way we can set this to be `[]` and have it build no plugins.

sutartmelson:
Not sure why its failing. It's passing these tests locally. First attempt it only failed web_client_jobs. Second attempt it only failed web_client_routing

zachmullen:
See https://github.com/girder/girder/issues/2093

Just restart the build until it goes green

danlamanna:
One issue here is that we should probably enforce the mutual exclusivity behavior, right now `--core-only --all-plugins --plugins=x,y,z` just builds everything without warning the user. We prevent users from doing similar things in `girder-cli`.  I also think a name like `--no-plugins` would make it more obvious that it's mutually exclusive with `--all-plugins` and `--plugins`.

A separate issue we should file is not handling mutually exclusive behavior with the watch functionality, I could be wrong but I don't think it's possible to do a general `--watch` over Girder core and `--watch-plugin` on a specific plugin.

zachmullen:
>  I could be wrong but I don't think it's possible to do a general --watch over Girder core and --watch-plugin on a specific plugin.

That's correct, for that we would need to spin up two separate subprocesses.

girder/utility/install.py
```diff
@@ -320,19 +320,26 @@ def main():
...
+
     web.add_argument('--watch', action='store_true',
                      help='watch for changes and rebuild girder core library in dev mode')
+
     web.add_argument('--watch-plugin', default='',
                      help='watch for changes and rebuild a specific plugin in dev mode')
 
-    web.add_argument('--plugin-prefix', default='plugin',
-                     help='prefix of the generated plugin bundle')
+    pluginGroup = web.add_mutually_exclusive_group()
```
sutartmelson: This address at least some of the above concerns.

sutartmelson:
@danlamanna I made the plugin arguments mutually exclusive.

jbeezley:
I also think `--no-plugins` is more consistent than `--core-only`.

girder/utility/install.py
```diff
@@ -315,16 +320,26 @@ def main():
...
+
+    pluginGroup.add_argument('--all-plugins', action='store_true',
+                             help='build all available plugins rather than just enabled ones')
+
+    pluginGroup.add_argument('--plugins', default=None,
+                             help='comma-separated list of plugins to build')
+
+    pluginGroup.add_argument('--plugin-prefix', default='plugin',
+                             help='prefix of the generated plugin bundle')
```
danlamanna: Does this belong in the group? I think this flag might only apply to the watch flag (which should probably become a subcommand at some point).

sutartmelson: I'm not entirely sure what the `--plugin-prefix` is doing. Could you conceivably want to run `girder-install web --plugin-prefix --no-plugins` or `girder-install web --plugin-prefix --all-plugins`?

zachmullen: Indeed, `--plugin-prefix` is just for watch. For plugins with multiple or non-standard build targets, this option is necessary.

sutartmelson:
@zachmullen Am I okay to merge? I can change our testing of girder-install to actually test the command line interface in a later PR.

zachmullen:
Yep, good to merge

Issue #2093, opened 2017-06-12
Test flakiness after move to jquery 3


zachmullen:
[These tests](http://my.cdash.org/viewTest.php?onlyfailed&buildid=1226366) are failing somewhat frequently now. I believe the move to jquery 3 has caused some new race conditions to emerge.

ping @brianhelba

brianhelba:
I think the `web_client_autojoin.autojoin` failure is a server-side issue. The API request which should [simply create a new user](https://github.com/girder/girder/blob/971d7b60710dba77da626cdbb0078bb78c951bbb/plugins/autojoin/plugin_tests/autojoinSpec.js#L116) is failing as:
```
  {
    "opts": {
      "dataType": "json",
      "type": "POST",
      "error": null,
      "path": "/user",
      "data": {
        "login": "user1",
        "password": "password",
        "email": "user1@example.com",
        "firstName": "Joe",
        "lastName": "User"
      },
      "url": "http://localhost:30023/api/v1/user"
    },
    "status": "error",
    "errorThrown": "Bad Request"
  }
```
with the error message (see screenshot) `Attempt to load null ObjectId: None`.

brianhelba:
`web_client_jobs.jobs` may be client-side. I'll take a look.

zachmullen:
Hm, that does look strange, I'll check into the autojoin test.

zachmullen:
The autojoin test failure is a result of a client-side race condition, combined with poor validation of the setting on the server side that makes the error message unhelpful.

sutartmelson:
I'm still getting this on #2091 
I rebased after the #2094 fix

zachmullen:
Yeah, @brianhelba is currently fixing the jobs test, so that's expected. If you see the autojoin test still failing, that would be unexpected.

manthey:
I haven't seen these tests flake out in quite a while, so I'm closing this issue.

PR #2094, opened 2017-06-12
Refs #2093. Fix autojoin web client test problems

zachmullen:
We were waiting on incorrect conditions, causing race conditions.
This also adds proper setting validation to the plugin, and fixes
a text bug in the UI for the plugin config.

### DIFF
--- a/plugins/autojoin/web_client/templates/configView.pug
+++ b/plugins/autojoin/web_client/templates/configView.pug
@@ -1,8 +1,9 @@
 .g-config-breadcrumb-container
 
-p.g-autojoin-description
-  | The Auto Join plugin allows you to define rules to automatically assign new users to groups based on their email address domain. Typically, this is used in conjunction with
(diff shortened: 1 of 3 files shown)


plugins/autojoin/web_client/templates/configView.pug
```diff
@@ -1,8 +1,9 @@
 .g-config-breadcrumb-container
 
-p.g-autojoin-description
-  | The Auto Join plugin allows you to define rules to automatically assign new users to groups based on their email address domain. Typically, this is used in conjunction with
-  a(href="#settings") email verification.
+p.g-autojoin-description.
+  The Auto Join plugin allows you to define rules to automatically assign new
+  users to groups based on their email address domain. Typically, this is used in
+  conjunction with #[a(href="#settings")] email verification.
```
manthey: The `a` tag doesn't surround any text, and therefore is unreachable.

PR #2095, opened 2017-06-13
Fixes #2081. Fix excessive item list fetches in hierarchy widget

### LINKED ISSUE #2081: Multiple item list fetches in hierarchy widget
I haven't dug into why this is happening, but when rendering a folder view, I'm seeing multiple repeated calls to the item list fetch, and a weird file download call to a nonexistent file.

zachmullen:
This topic contains two commits that both improve performance and reduce XHRs from the HierarchyWidget. 

5be0cc6 fixes #2081; prior to this change, the item list was being fetched three times on
each folder view. Lazy-instantiating the item list widget fixed this.

7e043bc changes the implementation of fetching the breadcrumb path back to the root resource. Rather than make sequential requests for each subsequent ancestor, they're now all fetched with a single request with a new rootpath endpoint that mirrors the one we have for items. This only affects initial loading of the widget, not descending into subfolders.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -22,7 +22,7 @@ dependencies:
     - ./scripts/install_cmake.sh
     - npm install -g npm
     - npm config set progress false
-    - npm update; npm prune
+    - npm update --no-save
     - set -o pipefail; pyenv exec pip install -U pip setuptools | cat
     - set -o pipefail; pyenv exec pip install -r requirements-dev.txt -e .[plugins,sftp] -e clients/python | cat
     - pyenv rehash
--- a/clients/web/src/models/FolderModel.js
+++ b/clients/web/src/models/FolderModel.js
@@ -2,9 +2,16 @@ import _ from 'underscore';
 
 import AccessControlledModel from 'girder/models/AccessControlledModel';
 import MetadataMixin from 'girder/models/MetadataMixin';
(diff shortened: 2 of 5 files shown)


zachmullen:
Hm, [every eslint test is failing](http://my.cdash.org/viewTest.php?onlyfailed&buildid=1227678) with this message:

```
Oops! Something went wrong! :(

ESLint couldn't find the plugin "eslint-plugin-node". This can happen for a couple different reasons:

1. If ESLint is installed globally, then make sure eslint-plugin-node is also installed globally. A globally-installed ESLint cannot find a locally-installed plugin.

2. If ESLint is installed locally, then it's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm i eslint-plugin-node@latest --save-dev

If you still can't figure out the problem, please stop by https://gitter.im/eslint/eslint to chat with the team.
```

@girder/developers I'm guessing this has to do with something that changed on master, anybody know?

jbeezley:
I would guess it's the `package-lock.json`.  I would try changing [this line](https://github.com/girder/girder/blob/master/circle.yml#L25) to
```js
npm update --no-save
```
The prune is no longer necessary.

zachmullen:
Good call, that was it.

PR #2097, opened 2017-06-13
Support filtering logs by duration as well as message frequency.

manthey:
I found this useful when testing high volumes of rest requests.

### DIFF
--- a/girder/api/filter_logging.py
+++ b/girder/api/filter_logging.py
@@ -20,6 +20,7 @@
 import cherrypy
 import logging
 import re
+import time
 
 
 LoggingFilters = []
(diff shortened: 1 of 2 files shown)


zachmullen:
Am I correct in assuming that this sort of rate-limiting will only work for a single girder app server?

manthey:
Yes.  We already can limit by number of messages; this adds limiting by time duration.

As it is, if two girder servers are using the same log file, does it work properly?

zachmullen:
I'm not sure how it would work now, I just like to keep tabs on the set of Girder features that do not work as expected at scale. This particular one is not problematic since at scale, this feature can still be used to approximate the desired rate limiting behavior.

danlamanna:
I would imagine that scaled installations would aggregate the {error,info}.log files in one place (we're looking at this with CloudWatch), and so if the server configuration does some rate limiting it would apply to the log files per server but obviously not the aggregate.

manthey:
I just ran a test with two Girder servers using the same log files.  Initially, they both log to the same file in an interleaved manner (`info.log` in my example).  This change appropriately reduces the logger per server, but the two servers know nothing of each others log filtering.

When one of the server's rotates the log file, it moves `info.log` to `info.log.1` (and rotates older files) and then starts writing to a new `info.log`.  The second server continue writing to `info.log.1`.  Unless restarted, the two servers will write to different log files, each one rotating when their current log file is full (which may not be the unnumbered file).

If one server is responding to more requests than the other, the logs can rotate enough that the less used server's logs are being written to a file that has been deleted, and therefore cannot be examined.

All of this is unrelated to this PR -- we probably should change how Girder log rotation works (or add support for syslogd, or something else), since this current behavior won't scale very well.  Otherwise, if you only keep *n* log files and have *n+1* girder servers, you are certain to miss logs.

zachmullen:
I imagine in a case with multiple app servers, we will uninstall the default log handler and install one that can aggregate logs across multiple servers e.g. LogStash.

danlamanna:
@zachmullen Agreed, I took a similar approach when testing [GW with Graylog](https://github.com/girder/girder_worker/blob/d17c40f6b99c1e4cb5609628defe8302db06d6e1/girder_worker/app.py#L12-L21).

zachmullen:
The trick will be making such things easy to configure without having to write per-deployment plugins.

PR #2098, opened 2017-06-13
item_tasks: Unify importing tasks from JSON and Slicer CLI

msmolens:
Previously it was possible to:
- Configure an item as a task from a Slicer CLI container.
- Add task items to a folder from a JSON container.

This change adds functionality to do both actions from either type of
container. The new configure tasks dialog allows the user to choose
between a JSON container and a Slicer CLI container by choosing a tab.

On the server, this change adds new endpoints to support the new
functionality. The endpoints extend the existing 'item' and 'folder'
routes so that the model type expected by the id URL parameter is clear.
The old endpoints are deprecated, but not removed.

Note that minor updates may necessary once support for Slicer CLI
containers describing more than one item is finalized.


msmolens:
Rebased to fix conflicts and take advantage of https://github.com/girder/girder/pull/2101 (Support marking a route as deprecated).

plugins/item_tasks/server/json_tasks.py
```diff
@@ -0,0 +1,245 @@
...
+                    'params': {
+                        'image': image,
+                        'pullImage': pullImage
+                    }
+                }
+            },
+            'jobInfo': utils.jobInfoSpec(job),
+            'validate': False,
+            'auto_convert': False,
+            'cleanup': True
```
zachmullen: `'cleanup': True` is the default, so let's remove this line.

plugins/item_tasks/server/json_tasks.py
```diff
@@ -0,0 +1,245 @@
...
+    .param('setDescription', 'Whether the item description should be changed to the '
+           'description of the task.', dataType='boolean', required=False, default=True)
+    .param('pullImage', 'Whether the image should be pulled from Docker Hub. '
+           'Set to false to use local images only.',
+           dataType='boolean', required=False, default=True),
+    hide=True
+)
+def configureItemTaskFromJson(self, item, json, image, taskName, setName, setDescription, pullImage,
+                              params):
+    self.ensureTokenScopes('item_task.set_task_spec.%s' % item['_id'])
```
zachmullen: I just realized this is not the method we want to call, we should add a `requireScopes` to the token model and call that instead. We'll want to make sure that our old endpoints change to use that new method also.

plugins/item_tasks/server/json_tasks.py
```diff
@@ -0,0 +1,245 @@
...
+               paramType='body')
+    .param('image', 'The docker image name.', required=True, strip=True)
+    .param('pullImage', 'Whether the image should be pulled from Docker Hub. ' +
+           'Set to false to use local images only.',
+           dataType='boolean', required=False, default=True)
+    .deprecated(),
+    hide=True
+)
+def createItemTasksFromJson(self, folder, json, image, pullImage, params):
+    self.ensureTokenScopes('item_task.set_task_spec.%s' % folder['_id'])
```
zachmullen: Same here

plugins/item_tasks/server/slicer_cli_tasks.py
```diff
@@ -0,0 +1,254 @@
...
+                        'setName': setName,
+                        'setDescription': setDescription
+                    },
+                    'headers': {'Girder-Token': token['_id']}
+                }
+            },
+            'jobInfo': utils.jobInfoSpec(job),
+            'validate': False,
+            'auto_convert': False,
+            'cleanup': True
```
zachmullen: can remove this line

plugins/item_tasks/server/slicer_cli_tasks.py
```diff
@@ -0,0 +1,254 @@
...
+    .param('xml', 'The Slicer CLI XML spec.', paramType='body')
+    .param('setName', 'Whether item name should be changed to the title of the CLI.',
+           dataType='boolean', required=True)
+    .param('setDescription', 'Whether the item description should be changed to the '
+           'description of the CLI.', dataType='boolean', required=False, default=True)
+    .deprecated(),
+    hide=True
+)
+def configureItemTaskFromSlicerCliXml(self, item, xml, setName, setDescription, params):
+    self.ensureTokenScopes('item_task.set_task_spec.%s' % item['_id'])
```
zachmullen: same here

plugins/item_tasks/server/slicer_cli_tasks.py
```diff
@@ -0,0 +1,254 @@
...
+@access.token
+@boundHandler()
+@autoDescribeRoute(
+    Description('Create item tasks under a folder using a list of Slicer CLI XML specs.')
+    .modelParam('id', model='folder', force=True)
+    .param('xml', 'The Slicer CLI XML spec.', paramType='body')
+    .param('image', 'The docker image name.', required=True, strip=True)
+    .jsonParam('args', 'Arguments to be passed to the docker container to output the '
+               'Slicer CLI spec.', required=False, default=[], requireArray=True)
+    .param('pullImage', 'Whether the image should be pulled from Docker Hub. ' +
```
zachmullen: No need for the `+`

plugins/item_tasks/web_client/views/ConfigureTasksDialog.js
```diff
@@ -1,41 +1,135 @@
...
 import router from 'girder/router';
 import View from 'girder/views/View';
+import '../stylesheets/configureTasks.styl';
 import 'girder/utilities/jquery/girderModal';
 
 import template from '../templates/configureTasks.pug';
 
+// View to configure item tasks. Supports both populating a folder with tasks
+// and configuring an item with a task. In both cases, tabs are presented to
+// allow choosing between a JSON container and a Slicer CLI container.
```
zachmullen: Block comment

plugins/item_tasks/web_client/views/ConfigureTasksDialog.js
```diff
@@ -1,41 +1,135 @@
...
-            }).done((job) => {
-                router.navigate(`job/${job._id}`, {trigger: true});
-            }).fail((resp) => {
-                this.$('.g-validation-failed-message').text(resp.responseJSON.message);
-            });
         }
     },
 
+    initialize: function () {
+        this.isFolder = this.model.get('_modelType') === 'folder';
```
zachmullen: Let's check `this.model.resourceName`, probably a bit safer.

plugins/item_tasks/server/json_tasks.py
msmolens: Based on the documentation for `ensureTokenScopes` (https://github.com/girder/girder/blob/637ab0e919bc355cbcb63a827eaabb8073c94408/girder/api/rest.py#L1026-L1029) the current behavior is unexpected. Indeed, this method implementation already passed review and is on master.

What would the proper implementation of `requireScopes` look like. Maybe it should be an API layer utility instead of on the model?

zachmullen: It would just be

```
if not self.model('token').hasScope(token, scope):
    raise AccessException('Invalid token scopes: ...')
```

I'm happy to expose it in the API layer as well, but there should be an implementation in the token model that it would wrap.

msmolens: I filed https://github.com/girder/girder/issues/2117, since that change seems out of scope (ha!) for this PR.

zachmullen: That's fine, as long as we can make that change ASAP after this goes in.

msmolens: Great, @dorukozturk already implemented this in https://github.com/girder/girder/pull/2031, so I'll merge this now.

jbeezley:
Sorry to bump an old PR, but I'm seeing strange things with the deprecation in the swagger docs.  See the attached screenshot:

test

I assume the new endpoints on `/item` and `/folder` aren't supposed to be deprecated, correct?  What I don't understand is why the slicer endpoint is deprecated on one and json on the other.

msmolens:
Thanks for noticing that. It's a bug that the new endpoints are marked deprecated.

To avoid duplication I attached the old route handling functions to the new routes. Because the old handlers are described as deprecated, the new routes inadvertently inherit that deprecation.

PR #2099, opened 2017-06-14
Jobs plugin: fix tests and make code improvements

brianhelba:
There are a lot of separate changes here. I've attempted to logically group them within individual commits, so I'd strongly recommend reviewing this commit-by-commit.

Some of these changes are more stylistic, but this also fixes several race conditions, memory leaks, and improves performance.

I think there's still some more cleanup to be done with this code, particularly with respect to @danlamanna 's [comments about modularization](https://github.com/girder/girder/pull/1901/files#r110745173) on #1901. It's too far outside the original scope of this PR now, but may eventually be worthwhile, given the importance of the `jobs` plugin.


plugins/jobs/web_client/views/timeChartConfig.js
```diff
@@ -270,7 +270,7 @@ export default {
     'legends': [
         {
             'fill': 'timing',
```
brianhelba: @matthewma7 This plot still renders with no legend icons. What is the purpose of the legend on this page?

matthewma7: @brianhelba no legend or legend icon? do you see anything different from the gif https://github.com/girder/girder/pull/1901

plugins/jobs/web_client/views/JobListWidget.js
```diff
@@ -190,9 +192,12 @@ var JobListWidget = View.extend({
     },
 
     _renderData: function () {
-        var jobs = this.collection.toArray();
+        if (!this.$('.g-main-content').length) {
+            // Do nothing until render has been called
+            return;
+        }
```
matthewma7: render() is called at the end of initialize(), fetch() is async. I don't think renderData could be called before render().

plugins/jobs/web_client/views/JobGraphWidget.js
```diff
@@ -11,7 +11,7 @@ import JobsGraphWidgetTemplate from '../templates/jobsGraphWidget.pug';
 import timingHistoryChartConfig from './timingHistoryChartConfig';
 import timeChartConfig from './timeChartConfig';
 
-export default View.extend({
+const JobGraphWidget = View.extend({
```
matthewma7: Is there a particular reason to have a named variable then export?

plugins/jobs/plugin_tests/jobsSpec.js
```diff
@@ -336,47 +336,6 @@ $(function () {
...
-                    el: $('#g-app-body-container'),
-                    parentView: app,
-                    filter: {},
-                    allJobsMode: true,
-                    showGraphs: true,
-                    showFilters: true,
-                    showPageSizeSelector: true
-                });
-
-                expect(widget.collection.resourceName).toEqual('job/all');
```
matthewma7: Is this line covered by other cases?

plugins/jobs/web_client/views/JobGraphWidget.js
```diff
@@ -110,13 +108,11 @@ const JobGraphWidget = View.extend({
...
-            config.scales[2].range = jobs.map(job => {
-                let datetime = moment(job.get('updated')).format('MM/DD');
-                return datetime;
-            });
-            config.scales[3].domain = jobs.map(job => job.get('_id'));
-            config.scales[3].range = jobs.map(job => job.get('title'));
+            config.scales[2].domain = this.collection.pluck('_id');
+            config.scales[2].range = this.collection.map(
+                (job) => moment(job.get('updated')).format('MM/DD'));
+            config.scales[3].domain = this.collection.pluck('_id');
```
matthewma7: I personally would prefer standard methods instead of backbone/underscore particular way, because I think standard stuff is easier to read for anyone. I also think using a local variable **jobs** looks better than **this.collection** and may has a little better performance. 
But this works fine.

matthewma7:
This PR looks good. It improves in general. I learned some technique from it as well.

plugins/jobs/web_client/models/JobModel.js
```diff
@@ -36,8 +36,9 @@ var JobModel = AccessControlledModel.extend({
             }));
         }
 
-        this.set('segments', segments);
```
zachmullen: This seems to change the behavior of this function. Is there a good reason we shouldn't call this `set()`?

package.json
```diff
@@ -20,7 +20,7 @@
     "babel-core": "^6.24.1",
     "babel-loader": "^6.2.x",
     "babel-preset-es2015": "^6.24.1",
-    "backbone": "~1.1.2",
+    "backbone": "~1.3.3",
```
danlamanna: There seem to be quite a bit of changes between these versions. When looking at the changelog did you see anything that makes sense to file in separate issues?

plugins/jobs/web_client/stylesheets/jobListWidget.styl
```diff
@@ -79,4 +79,4 @@ ul.g-jobs.nav.nav-tabs
   overflow-x auto
 
 .g-no-job-record
-  text-align center
\ No newline at end of file
+  text-align center
```
danlamanna: Does utilizing something like [eol-last](http://eslint.org/docs/rules/eol-last) make sense moving forward?

jeffbaumes: See https://github.com/girder/girder/issues/2104

package.json
brianhelba: Yeah, if you'd like,I'll move it to a separate PR.

However, I did review [the changelog](http://backbonejs.org/#changelog), and I don't think there will be any changes that we need to make to Girder. The separate PR would give this more visibility and scrutiny from other devs though.

plugins/jobs/web_client/models/JobModel.js
brianhelba: Yes, per my [commit comment](), `'segments'` is not a Model attribute. It doesn't exist on the server and doesn't store any unique information. We should not abuse Model attributes to serve as a cache for transformed data, as it will cause problems with `Model.sync`, etc. (should we ever want to start using it property) and make our models harder to reason about (since the value gets populated on-demand, not as part of the natural Model lifecycle).

The old control flow wasn't really pre-computing this value at a more convenient point in the lifecycle or a fewer number of times; it was just calling this once per job, then later accessing the attribute once per job. So, rather than pointlessly cache the value, I just made this into a utility.

plugins/jobs/web_client/stylesheets/jobListWidget.styl
brianhelba: It should already be on, but only checks Javascript files, not Pug or anything else.

plugins/jobs/web_client/views/JobGraphWidget.js
brianhelba: Yes, it's a marker for what other modules should import this name as, so we can consistently import it as the same thing.

plugins/jobs/web_client/views/JobListWidget.js
brianhelba: Yes, but I think it's much safer not to rely on the that idiosyncrasy. Many other non-top-level Views don't self-render, and I wouldn't want this to be a source of bugs if that ever changes here.

plugins/jobs/web_client/models/JobModel.js
zachmullen: Makes sense, this is a bugfix and not a breaking change.

plugins/jobs/web_client/views/timeChartConfig.js
brianhelba: No icon. It looks like your screenshot, but without a set of icon keys, it's just a list of words. I'm not sure what the purpose is on this plot.

package.json
danlamanna: I don't think moving it to a separate PR is necessary. I just wanted to make sure we were aware of any new features we could use and/or cleanup we could do, and it seems like you've done that.

plugins/jobs/web_client/views/JobGraphWidget.js
brianhelba: We use Backbone Collections for a reason, casting to an Array removes all of the utilities and benefits:
* Collections have several internal structures to improve the performance of methods like `.at`
* Casting to an Array takes some constant amount of initial overhead and additional memory
* It makes a copy of the object, making the code harder to reason about and maintain (e.g. if we ever wanted to start mutating the Collection)

IMHO, understanding Backbone is a hard requirement for programming Girder's front end. All notions about developer readability should be predicated on them having adequate Backbone knowledge.

package.json
brianhelba: @danlamanna I totally agree. I'll be trying to incorporate new opportunities in other PRs, but I'll be happy for now if there are no breakages.

plugins/jobs/web_client/views/JobGraphWidget.js
matthewma7: I see, looks like that's the pattern we use in Girder.

brianhelba: Yeah, if we can enforce all imports to use the same names for default-exported objects, it will make it much easier to grep for usages in the whole codebase. I'm hoping to enforce this rule further with [`eslint-plugin-import`](https://www.npmjs.com/package/eslint-plugin-import), but it will probably take a little extra work.

plugins/jobs/web_client/views/timeChartConfig.js
matthewma7: I see what you mean. The legend of on the line chart is an indicator of what phases has been selected by the filter.

plugins/jobs/web_client/views/CheckBoxMenu.js
```diff
@@ -43,17 +43,20 @@ var CheckBoxMenu = View.extend({
...
     },
+
     setItems: function (items) {
         this.items = items;
         this._renderContent();
     },
-    allItemChecked: function () {
-        return _.every(_.values(this.items));
+
+    allItemsChecked: function () {
```
zachmullen: Has this made it into a release yet? If not we can skip adding the alias.

brianhelba: No, this isn't released. I agree, I don't think we need a public method for this.

PR #2100, opened 2017-06-14
Enable two additional static analysis checks

brianhelba:
* Make 1-argument arrow functions use the same syntax and 0 and 2 arguments
* Use a stricter form of === enforcement, to ensure nulls are checked

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
     "extends": "girder",
     "rules": {
+        "arrow-parens": "error",
         "no-throw-literal": "off",
         "backbone/defaults-on-top": "error",
         "backbone/no-changed-set": "error",
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,9 @@
     "extends": "girder",
     "rules": {
         "arrow-parens": "error",
+        "eqeqeq": ["error",
+            "always", {"null": "always"}
+        ],
         "no-throw-literal": "off",
         "backbone/defaults-on-top": "error",
         "backbone/no-changed-set": "error",
--- a/clients/web/src/models/CollectionCreationPolicyModel.js
+++ b/clients/web/src/models/CollectionCreationPolicyModel.js
@@ -7,11 +7,11 @@ var CollectionCreationPolicyModel = AccessControlledModel.extend({
         return restRequest({
             path: this.resourceName + '/access',
             type: 'GET'
-        }).done(resp => {
+        }).done((resp) => {
             this.set('access', resp);
             this.trigger('g:accessFetched');
             return resp;
-        }).fail(err => {
+        }).fail((err) => {
             this.trigger('g:error', err);
         });
     }
--- a/clients/web/src/models/MetadataMixin.js
+++ b/clients/web/src/models/MetadataMixin.js
@@ -50,12 +50,12 @@ var MetadataMixin = {
             data: JSON.stringify(key),
             type: 'DELETE',
             error: null
-        }).done(resp => {
+        }).done((resp) => {
             this.set(opts.field || 'meta', resp.meta);
             if (_.isFunction(successCallback)) {
                 successCallback();
             }
-        }).fail(err => {
+        }).fail((err) => {
             err.message = err.responseJSON.message;
             if (_.isFunction(errorCallback)) {
                 errorCallback(err);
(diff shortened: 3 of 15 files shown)


.eslintrc
```diff
@@ -2,6 +2,9 @@
     "extends": "girder",
     "rules": {
         "arrow-parens": "error",
+        "eqeqeq": ["error",
+            "always", {"null": "always"}
+        ],
```
ronichoudhury: What does this setting do?

brianhelba: [It says that we should always use `===`](http://eslint.org/docs/rules/eqeqeq). The options for this rule that we inherit from `eslint-plugin-standard` allow the use of `==` to compare with `null`, so this overrides to not permit that case either.

ronichoudhury:
That's so weird... do you know why `==` is allowed for comparing to `null`
by default?

brianhelba:
I have no real idea. Maybe because people use it as a method to evaluate truthiness of an object, because `Boolean` is deceptively complicated (since it can be used with `new Boolean`) and using `!!` smells like a dirty hack? Either way, IMHO doing `== null` is worse than both of those, so we should disable the exception to the rule we were already using.

zachmullen:
> !! smells like a dirty hack

Whoa, whoa, I happen to love !!

brianhelba:
@zachmullen Just don't let me find out that you've been hanging out with [the `--> operator`](https://stackoverflow.com/questions/1642028/what-is-the-operator-in-c).

ronichoudhury:
```javascript
import popcorn from 'popcorn';

popcorn.eat();
```

zachmullen:
What a classic.

manthey:
`!!` is no more a dirty hack than `=>` is a dirty hack to avoid having to explicitly bind a function, or `+` to cast to a number.

`null` is loosely equal only to `null` and `undefined`, so `foo == null` is a shortcut to `foo === null || foo ===  undefined` (excepting if `undefined` has been overloaded, of course).  This is useful enough, that many projects allow `== null`.

brianhelba:
@manthey Ah, I stand corrected about the comparison. That makes sense.
the_more_you_know

---
Just to continue the religious war, I'll maintain that `!!` is still slightly dodgy. `=>`, `++`, etc. are all atomic operators, whereas `!!` ~~exploits~~ leverages the grammar to pretend to work as a single operator. I'd agree about `+` coercion between dissimilar types (particularly strings) being dirty too, though.

zachmullen:
`!!` isn't even a stand-in for truthiness testing -- for that I would just do `if (variable)`. You really only need to use it if you must have a boolean type. The alternative is `new Boolean(variable).valueOf()`, which IMO is gross.

brianhelba:
Yeah, I was mistaken. It was fake news.

brianhelba:
@girder/developers Does anyone object to either of these rules? If not, I'd like to merge this and push all these config updates upstream to `eslint-config-girder`?

sebastienbarre:
Or you can go the Go way with [gofmt - The Go Programming Language](https://golang.org/cmd/gofmt/), and have an opiniated tool that reformat your code automatically (on the fly, or at commit time).

Said tool now exists for Javascript (by vjeux of FB/React fame): [prettier/prettier: Prettier is an opinionated code formatter](https://github.com/prettier/prettier) and integrates with ESLint.  You only have a few options like column width, space vs tab, and semi-colon or not. It's a lot less work and has shrunk my `.eslintrc` by a lot:
```
{
  "extends": [
    "airbnb",
    "plugin:flowtype/recommended",
    "plugin:react/recommended",
    "plugin:import/errors",
    "plugin:import/warnings",
    "plugin:jsx-a11y/recommended",
    "prettier",           // https://github.com/prettier/eslint-config-prettier
    "prettier/flowtype",
    "prettier/react"
  ],
  "parser": "babel-eslint",
  "plugins": [
    "flowtype",  // https://github.com/gajus/eslint-plugin-flowtype
    "import",    // https://github.com/benmosher/eslint-plugin-import
    "jsdoc",     // https://github.com/gajus/eslint-plugin-jsdoc
    "jsx-a11y",  // https://github.com/evcohen/eslint-plugin-jsx-a11y
    "react",     // https://github.com/yannickcr/eslint-plugin-react
    "smells",    // https://github.com/elijahmanor/eslint-plugin-smells
    "prettier"   // https://github.com/not-an-aardvark/eslint-plugin-prettier
  ],
  "rules": {
    "prettier/prettier": [
      "error", {
        "printWidth": 100,
        "singleQuote": true,
        "trailingComma": "es5"
      }
    ],
  }
}
```

@jourdain you using this?

sebastienbarre:
I only use the ESlint "smells" plugin because I'm French.

jourdain:
I'm not using it but that seems very interesting...

jourdain:
But I guess, I'll be waiting for
https://github.com/prettier/prettier/issues/187

PR #2102, opened 2017-06-14
Fix problem in web_client_curation test

zachmullen:
waitsFor() and runs() blocks should be nested inside it() blocks.


zachmullen:
See http://my.cdash.org/testDetails.php?test=32293727&build=1227695

brianhelba:
@zachmullen Are there any circumstances under which it's necessary to wrap things inside `$(function () {` within a test file?

zachmullen:
Not that I can think of. If it's being done just for scoping purposes, `(function () { } ())` would be preferable. Certainly in this case, there's no reason to bind to document.ready.

Issue #1989, opened 2017-05-09
Download statistics plugin


mathstuf:
A plugin to track download statistics of files would be nice to have.

I think the best would be to track counts for downloads started (byte 0 sent), downloads completed (final byte sent), and maybe total requests.

Cc: @utkarshayachit @jourdain

mgrauer:
Girder provides an SFTP server in addition to the web-app server, and currently the SFTP server doesn't allow for plugins, meaning we wouldn't be able to track download statistics through a plugin via the SFTP server in its current form.

Download statistics should be tracked in a way that is stateless on the web-app server, so that it would work with multiple app servers running at the same time.

zachmullen:
Eventually I think the SFTP server should support plugins. We'd probably need to mock out the `apiRoot` field during `load`, but otherwise it should work.

sutartmelson:
Is the idea to have the individual stats attached to each item? Or stored to be viewed collectively in the plugin control panel, for example. 

@jeffbaumes

zachmullen:
For now let's keep them per-file and use the `increment` method on the file model to atomically increase the download count.

sutartmelson:
I'm struggling to figure out a way to track downloads completed. Downloads started, just checking when byte 0 is requested, is straightforward. Total requests is also straightforward. I'm not sure if there's a way to check when the last byte was requested?

zachmullen:
Perhaps in the generator where we stream the file, we can trigger an event after the loop terminates, assuming the last byte was requested. That should only get triggered if the server actually sent the last byte, not if the request was canceled prior to that.

manthey:
I think you'd have to indicate this within each assetstore type, and couldn't do it generally.  For the filesystem assetstore, if `endByte` in the `downloadFile` method is equal to `file['size']`, and if `readLen` is 0 in the `stream` generator, then we could trigger an event marking that the entire download has been sent.

zachmullen:
I think we can do it generally, but we'd have to introduce a wrapper generator inside the download method in the file model.

sutartmelson:
If a client requests a file download(starting at byte zero, no end byte), is it possible for that request to turn into multiple requests? Or is counting the total requests for when a client specifically requests a partial file download?

manthey:
[...]

PR #2103, opened 2017-06-14
Download statistics plugin

sutartmelson:
As discussed in #1989, this server side plugin tracks file download activity and stores it in the file model.


plugins/download_statistics/plugin.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name": "Download Statistics",
+    "description": "Record file download statistics",
+    "url": "",
+    "version": "0.1.0",
+    "dependencies": []
+}
```
brianhelba: As a general Unix thing, all files should end with a single newline.

sutartmelson: Good to know. Thanks!

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+###############################################################################
+
+
+from girder import events
+from girder.constants import AccessType
+from girder.models.item import Item
+from girder.models.file import File
```
brianhelba: It's the convention to use `ModelImporter` within a method, to get Model class instances.

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def downloadItemStartEvent(event):
+    # Ony count download as started if offset is 0
+    itemModel = Item()
+    item = itemModel.load(event.info['id'], force=True)
+    firstRequest = 'offset' not in event.info['params'] or event.info['params']['offset'] == '0'
+
+    # Get all files in item to increment their download count
+    files = list(itemModel.childFiles(item=item))
```
brianhelba: No need to coerce this to a list. Just iterate over it directly.

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+            downloadFileStart(file['_id'])
+        downloadFileRequest(file['_id'])
+
+
+def downloadItemCompleteEvent(event):
+    itemModel = Item()
+    item = itemModel.load(event.info['id'], force=True)
+
+    # Get all files in item to increment their download count
+    files = list(itemModel.childFiles(item=item))
```
brianhelba: Don't cast to a list.

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def downloadFileCompleteEvent(event):
+    fileModel = File()
+    file = fileModel.load(event.info['id'], force=True)
+    downloadFileComplete(file['_id'])
+
+
+def downloadFileStart(fileId):
+    File().increment(query={'_id': fileId}, field='downloadsStarted', amount=1, multi=False)
```
brianhelba: Maybe it'd be cleaner to nest all 3 of these new properties inside a parent `dict`, so we only have to expose the 1 parent on the `File` object.

plugins/download_statistics/plugin.cmake
```diff
@@ -0,0 +1 @@
+add_standard_plugin_tests()
```
brianhelba: Newline here.

brianhelba:
@sutartmelson Oh, I thought this was ready for review, I didn't see the WIP flag until just now. Sorry for the premature comments.

sutartmelson:
@brianhelba no I appreciate all this feedback now rather than everything all at once later. This helps me get familiar with convention and best practice. Thanks!

plugins/download_statistics/server/__init__.py
sutartmelson: So just expose a single attribute, a new dictionary, rather than the three separate integers?

plugins/download_statistics/plugin.cmake
jeffbaumes: See https://github.com/girder/girder/issues/2104

plugins/download_statistics/server/__init__.py
brianhelba: Something like
```
file['downloadStats'] = {
    'started': started,
    'requested': requested,
    'completed': completed
}
```
might be cleaner to work with, particularly for [`Model.exposeFields`](https://github.com/girder/girder/blob/master/girder/models/model_base.py#L90).

 I have no strong preference on what the property names are, those were just examples.

plugins/download_statistics/server/__init__.py
brianhelba: Check out using [`boundHandler`](https://github.com/girder/girder/blob/master/girder/api/rest.py#L1164) for these new `Resource`-ish methods.

plugins/download_statistics/server/__init__.py
sutartmelson: How would I actually expose that dictionary? Also, how would I increment the values within?

Is this how to expose the dictionary? 
```
    downloadStatistics = {'downloadsStarted': 0,
                          'downloadsRequested': 0,
                          'downloadsCompleted': 0}
    ModelImporter.model('file').exposeFields(level=AccessType.READ,
                                             fields='downloadStatistics')
```
I'm not sure how to change this to access that dictionary..
```
ModelImporter.model('file').increment(query={'_id': fileId}, field='downloadsStarted',
                                          amount=1, multi=False)
```

brianhelba: [`increment`'](https://github.com/girder/girder/blob/master/girder/models/model_base.py#L447) is implemented via [MongoDB's `$inc` operator](https://docs.mongodb.com/manual/reference/operator/update/inc/). To access nested fields in Mongo, just use a `.`. So
```
.increment(query={'_id': fileId}, field='downloadStats.started', amount=1, multi=False)
```

As an aside, the `multi` argument is not be necessary, as an `_id` query will only ever match a single document.

brianhelba: As for exposing. you'd use the field name of whatever the top-level property on your `file` object was. All nested properties would then just be values of that field.

girder/models/file.py
```diff
@@ -94,10 +95,19 @@ def download(self, file, offset=0, headers=True, endByte=None,
         :type extraParameters: str or None
         """
         if file.get('assetstoreId'):
-            return self.getAssetstoreAdapter(file).downloadFile(
+            fileDownload = self.getAssetstoreAdapter(file).downloadFile(
                 file, offset=offset, headers=headers, endByte=endByte,
                 contentDisposition=contentDisposition,
                 extraParameters=extraParameters)
+            if inspect.isgeneratorfunction(fileDownload):
```
zachmullen: Let's make this `inspect.isgeneratorfunction(fileDownload) or inspect.isgenerator(fileDownload)`, since I believe it's valid to return either type.

girder/models/file.py
```diff
@@ -94,10 +95,19 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
+                def downloadGenerator():
+                    for data in fileDownload():
+                        yield data
+                    if endByte is None or endByte >= file['size']:
+                        events.trigger('download.complete', info={'id': file['_id']})
+                return downloadGenerator
+            else:
+                return fileDownload
```
zachmullen: We should also add a try/except for cherrypy.redirect exceptions. We won't know for sure if the user completed the download, but we can log it, perhaps with a field that says `redirect: True` in the event info.

girder/models/file.py
```diff
@@ -94,10 +95,19 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
+            fileDownload = self.getAssetstoreAdapter(file).downloadFile(
                 file, offset=offset, headers=headers, endByte=endByte,
                 contentDisposition=contentDisposition,
                 extraParameters=extraParameters)
+            if inspect.isgeneratorfunction(fileDownload):
+                def downloadGenerator():
+                    for data in fileDownload():
+                        yield data
+                    if endByte is None or endByte >= file['size']:
+                        events.trigger('download.complete', info={'id': file['_id']})
```
zachmullen: Rather than passing id, we should pass `'file': file` so they have the full document.

girder/models/file.py
sutartmelson: Trigger the download.complete in both occurrences of the HTTPRedirect like this? 
```python
if file.get('assetstoreId'):
    try:
        fileDownload = self.getAssetstoreAdapter(file).downloadFile(
            file, offset=offset, headers=headers, endByte=endByte,
            contentDisposition=contentDisposition,
            extraParameters=extraParameters)
        if inspect.isgeneratorfunction(fileDownload) or inspect.isgenerator(fileDownload):
            def downloadGenerator():
                for data in fileDownload():
                    yield data
                if endByte is None or endByte >= file['size']:
                    events.trigger('download.complete', info={'file': file})
            return downloadGenerator
        else:
            return fileDownload
    except cherrypy.HTTPRedirect as e:
        events.trigger('download.complete', info={'file': file, 'redirect': True})
        raise e
elif file.get('linkUrl'):
    if headers:
        events.trigger('download.complete', info={'file': file, 'redirect': True})
        raise cherrypy.HTTPRedirect(file['linkUrl'])
    else:
        endByte = endByte or len(file['linkUrl'])

        def stream():
            yield file['linkUrl'][offset:endByte]
            if endByte >= len(file['linkUrl']):
                events.trigger('download.complete', info={'file': file})
        return stream
else:  # pragma: no cover
    raise Exception('File has no known download mechanism.')
```

zachmullen: That looks good to me, but let's also pass `'redirect': False` in the other cases so we provide a uniform set of fields to the event handlers.

sutartmelson:
Why might this be causing all the eslint tests to fail?

zachmullen:
Just merge latest master onto this topic and those errors should go away.

jbeezley:
Depending on when the first build was made on this branch, you may need to delete the cache as well.

girder/models/file.py
```diff
@@ -94,18 +95,35 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
+                if inspect.isgeneratorfunction(fileDownload) or inspect.isgenerator(fileDownload):
+                    def downloadGenerator():
+                        for data in fileDownload():
+                            yield data
+                        if endByte is None or endByte >= file['size']:
+                            events.trigger('download.complete', info={'file': file,
+                                                                      'redirect': False})
+                    return downloadGenerator
+                else:
+                    return fileDownload
```
brianhelba: In this case, is the `'ownload.complete'` event getting triggered?

sutartmelson: No. From what I understand, that wont ever happen with the current set of assetstores. It's to allow future assetstores to behave differently. In which case we don't yet know if the download actually completed or not.

sutartmelson:
@zachmullen, I think I've addressed all of your suggestions from your review. Take a look at let me know.

brianhelba:
@girder/developers I'm reviewing this now.

girder/models/file.py
brianhelba: Ok, let's put a comment here explaining that fact (that this case should not be reached with any current assetstores).

girder/models/file.py
```diff
@@ -94,18 +95,35 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
+            except cherrypy.HTTPRedirect as e:
+                events.trigger('download.complete', info={'file': file, 'redirect': True})
+                raise e
```
brianhelba: In cases where we want to add some additional behavior to an exception propagation, but don't want to actually alter the exception or its propagation, it's better to use [a plain `raise` statement](https://docs.python.org/2/reference/simple_stmts.html#raise), without a variable. That will preserve the exception's built-in stack trace all the way down to the original place it was raised.

So, this can be simply:
```
except cherrypy.HTTPRedirect:
    events.trigger('download.complete', info={'file': file, 'redirect': True})
    raise
```
since we no longer need `e`.

docs/plugins.rst
```diff
@@ -158,6 +158,18 @@ If there is a match, the user is added to the group with the specified access
 level.
 
 
+Download Statistics
+-------------------
+
+This plugin tracks and records file download activity. The recorded information,
```
brianhelba: Should use a colon: `The recorded information:`

plugins/download_statistics/plugin_tests/download_statistics_test.py
```diff
@@ -0,0 +1,119 @@
...
+        # Create admin user
+        admin = {'email': 'admin@email.com',
+                 'login': 'adminLogin',
+                 'firstName': 'adminFirst',
+                 'lastName': 'adminLast',
+                 'password': 'adminPassword',
+                 'admin': True}
+        self.admin = self.model('user').createUser(**admin)
+
+        folders = self.model('folder').childFolders(parent=self.admin, parentType='user',
```
brianhelba: All these lines (50-58) could be simply:
```
self.publicFolder = self.model('folder').findOne({
    'name': 'Public',
    'parentId': self.admin['_id'],
    'parentCollection': 'user'
})
```

plugins/download_statistics/plugin_tests/download_statistics_test.py
```diff
@@ -0,0 +1,119 @@
...
+                                                    user=self.admin)
+
+        for folder in folders:
+            if folder['public'] is True:
+                self.publicFolder = folder
+
+        self.filesDir = os.path.join(ROOT_DIR, 'plugins', 'download_statistics',
```
brianhelba: @girder/developers It'd be nice to have a more robust way to lookup non-external test artifacts, but I'm not sure what it would be in this case.

plugins/download_statistics/plugin_tests/download_statistics_test.py
```diff
@@ -0,0 +1,119 @@
...
+        # Path to test files
+        file1Path = os.path.join(self.filesDir, 'txt1.txt')
+        file2Path = os.path.join(self.filesDir, 'txt2.txt')
+
+        # Upload files to item
+        with open(file1Path, 'rb') as fp:
+            self.model('upload').uploadFromFile(fp, os.path.getsize(file1Path),
+                                                'txt1.txt', parentType='item',
+                                                parent=item, user=self.admin)
+            self.model('upload').uploadFromFile(fp, os.path.getsize(file2Path),
```
brianhelba: This second upload should go under its own `with` context. Otherwise you're uploading the `fp` for `file1Path` and calling it `txt2.txt`. Worse, the `fp` will have been exhausted by the first upload, so it's empty at this point.

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def downloadFileComplete(fileId):
+    ModelImporter.model('file').increment(query={'_id': fileId},
+                                          field='downloadStatistics.completed', amount=1)
+
+
+def load(info):
+    # Bind REST events
+    events.bind('rest.get.item/:id/download.before', 'download_statistics',
```
brianhelba: What if a folder or collection is downloaded (as a ZIP)? The files within should all be incremented, as they're getting downloaded.

However, since collection downloads also have a `mimeFilter`, we can't assume that every file in the collection is actually getting downloaded in that case.

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    events.bind('rest.get.file/:id/download.before', 'download_statistics',
+                downloadFileStartEvent)
+    events.bind('download.complete', 'download_statistics',
+                downloadFileCompleteEvent)
+
+    # Add download count fields to file model
+    downloadStatistics = {'started': 0,
+                          'requested': 0,
+                          'completed': 0}
+    ModelImporter.model('file').exposeFields(level=AccessType.READ,
```
brianhelba: The field to expose is just `'downloadStatistics'`. Everything within will be returned as the value of this field.

The fact that these fields are getting exposed via the API should be tested.

plugins/download_statistics/plugin_tests/download_statistics_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+        # Download item, and its files, several times and ensure downloads are recorded
+        files = list(self.model('item').childFiles(item=item))
+        self.assertTrue(files)
+        for n in range(0, 5):
+            self._downloadItem(item['_id'])
+            for file in files:
+                self._downloadFile(file['_id'])
+        for file in files:
+            self._checkDownloadsStarted(file['_id'], 10)
```
brianhelba: These assertions should be tested by retrieving the properties via the API (rather than a database query), to ensure they're being exposed properly.

girder/models/file.py
```diff
@@ -94,18 +95,35 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
         elif file.get('linkUrl'):
             if headers:
+                events.trigger('download.complete', info={'file': file, 'redirect': True})
```
brianhelba: This event should use the same conventions as others in Girder. Perhaps something like `'model.file.download.after'`?

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    firstRequest = 'offset' not in event.info['params'] or event.info['params']['offset'] == '0'
+
+    if firstRequest:
+        downloadFileStart(file['_id'])
+    downloadFileRequest(file['_id'])
+
+
+def downloadFileCompleteEvent(event):
+    fileModel = ModelImporter.model('file')
+    file = fileModel.load(event.info['file']['_id'], force=True)
```
brianhelba: Do you need to re-load the field, if all you needed was its `_id`?

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    for file in files:
+        if firstRequest:
+            downloadFileStart(file['_id'])
+        downloadFileRequest(file['_id'])
+
+
+def downloadFileStartEvent(event):
+    # Ony count download as started if offset is 0
+    fileModel = ModelImporter.model('file')
+    file = fileModel.load(event.info['id'], force=True)
```
brianhelba: Do you need to re-load the field, if all you needed was its `_id`?

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+def downloadItemStartEvent(event):
+    # Ony count download as started if offset is 0
+    itemModel = ModelImporter.model('item')
+    item = itemModel.load(event.info['id'], force=True)
+    firstRequest = 'offset' not in event.info['params'] or event.info['params']['offset'] == '0'
+
+    # Get all files in item to increment their download count
+    files = itemModel.childFiles(item=item)
+    for file in files:
```
brianhelba: Rather than do `n` different increment operations for every file in this item, we could be more efficient by pulling the query logic out of `childFiles`, and using it to increment all the matched files with a single database request.

sutartmelson:
@brianhelba I re-implemented this to trigger all download events in the file model. I believe this solves the collection download case you mentioned above, though I have not tested that specifically. I'm still a little unsure about the field exposure thing. Let me know if I'm doing that correctly now.

docs/plugins.rst
sutartmelson: I'm not sure I agree :) 
Maybe I should use parentheses though. 
`The recorded information
(downloads started, downloads completed, and total requests made) is stored on the
file model:`

plugins/download_statistics/plugin_tests/download_statistics_test.py
sutartmelson: I now check the exposed properties separately as to not have to think about triggering more download events while downloading the file just to check the exposed property.

sutartmelson:
@brianhelba I tested collection download, with and without mimetype filtering. Let me know if I forgot something else we discussed.

girder/models/file.py
```diff
@@ -93,19 +94,42 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
+                if inspect.isgeneratorfunction(fileDownload) or inspect.isgenerator(fileDownload):
+                    def downloadGenerator():
+                        for data in fileDownload():
+                            yield data
+                        if endByte is None or endByte >= file['size']:
+                            events.trigger('model.file.download.after',
+                                           info={'file': file, 'redirect': False})
+                    return downloadGenerator
+                else:
+                    # With our current assetstores (June 2017), this case should not be reached.
```
brianhelba: Maybe this should trigger an `'model.file.download.after'` event, just for safety in the future?

plugins/download_statistics/plugin_tests/download_statistics_test.py
```diff
@@ -0,0 +1,146 @@
...
+                                                   parentType='collection',
+                                                   public=True, )
+        # Create item
+        item = self.model('item').createItem('item1', self.admin, folder)
+
+        # Path to test files
+        file1Path = os.path.join(self.filesDir, 'txt1.txt')
+        file2Path = os.path.join(self.filesDir, 'txt2.txt')
+
+        file1 = ''
```
brianhelba: No need to initialize `file1` and `file2` here. Variables declared in a `with` context persist in the scope outside of it.

plugins/download_statistics/plugin_tests/download_statistics_test.py
```diff
@@ -0,0 +1,146 @@
...
+                                     'plugin_tests', 'files')
+
+    def _downloadItem(self, itemId):
+        # Download item through REST api
+        path = '/item/%s/download' % str(itemId)
+        resp = self.request(path, user=self.admin, isJson=False)
+        self.assertStatusOk(resp)
+
+        # Iterate through generator to trigger download events
+        for data in resp.body:
```
brianhelba: I'm just curious, do you know why this is necessary? Is the `self.request` somehow lazily downloading files in chunks? From the server's perspective, is more than one request being made?

plugins/download_statistics/plugin_tests/download_statistics_test.py
```diff
@@ -0,0 +1,146 @@
...
+                                'id': collection['_id'],
+                                'mimeFilter': json.dumps(['image/jpeg'])
+                            })
+
+        # iterate through generator to trigger download events
+        for data in resp.body:
+            data
+
+        self._checkDownloadsCount(file1['_id'], 11)
+        self._checkDownloadsCount(file2['_id'], 12)
```
brianhelba: How hard would it be to test a download that's started but never completed? Or requested multiple times?

plugins/download_statistics/plugin_tests/download_statistics_test.py
```diff
@@ -0,0 +1,146 @@
...
+            data
+
+    def _checkDownloadsCount(self, fileId, count):
+        # Downloads file info and asserts download statistics are accurate
+        path = '/file/' + str(fileId)
+        resp = self.request(path, isJson=True)
+        self.assertStatusOk(resp)
+        data = resp.json
+
+        # The generator is never iterated as to not trigger additional events
```
brianhelba: I don't understand what this line means.

plugins/download_statistics/plugin_tests/download_statistics_test.py
```diff
@@ -0,0 +1,146 @@
...
+                                                        parentType='item', parent=item,
+                                                        user=self.admin)
+
+        # Download item and its files several times and ensure downloads are recorded
+        # Each file is downloaded 10 times
+        for n in range(0, 5):
+            self._downloadItem(item['_id'])
+            self._downloadFile(file1['_id'])
+            self._downloadFile(file2['_id'])
+
```
brianhelba: Maybe we should download the folder, just to be through?

girder/models/file.py
```diff
@@ -93,19 +94,42 @@ def download(self, file, offset=0, headers=True, endByte=None,
         :type contentDisposition: str or None
         :type extraParameters: str or None
         """
+        if offset is 0:
```
brianhelba: `is` should be used to compare `None`, and to see if a value is explicitly `True` or `False`. Otherwise, use `==` to compare, as `is` looks at reference (location in memory) equality, not value equality.

This bug reveals that this code path is not adequately tested.

girder/models/file.py
```diff
@@ -93,19 +94,42 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
-            return self.getAssetstoreAdapter(file).downloadFile(
-                file, offset=offset, headers=headers, endByte=endByte,
-                contentDisposition=contentDisposition,
-                extraParameters=extraParameters)
+            try:
+                fileDownload = self.getAssetstoreAdapter(file).downloadFile(
+                    file, offset=offset, headers=headers, endByte=endByte,
+                    contentDisposition=contentDisposition,
+                    extraParameters=extraParameters)
+                if inspect.isgeneratorfunction(fileDownload) or inspect.isgenerator(fileDownload):
```
brianhelba: Can `inspect.isgenerator(fileDownload)` ever be true? If so, how does the call to `fileDownload()` below work?

plugins/download_statistics/plugin_tests/download_statistics_test.py
```diff
@@ -0,0 +1,146 @@
...
+        resp = self.request(path, user=self.admin, isJson=False)
+        self.assertStatusOk(resp)
+
+        # Iterate through generator to trigger download events
+        for data in resp.body:
+            data
+
+    def _checkDownloadsCount(self, fileId, count):
+        # Downloads file info and asserts download statistics are accurate
+        path = '/file/' + str(fileId)
```
brianhelba: Use the `%` operator here, for consistency?

plugins/download_statistics/plugin_tests/download_statistics_test.py
sutartmelson: After making the download request, it appears as though another file download has happened, which would mean the download stats would all be one higher. Since this isn't iterating over the returned generator, another download does not happen.

plugins/download_statistics/plugin_tests/download_statistics_test.py
sutartmelson: If the response body is a generator, the data isn't actually downloaded until the response is iterated over. I didn't know what else I could do to ensure the download happens. Multiple requests aren't happening here, just iterating over the data.

plugins/download_statistics/plugin_tests/download_statistics_test.py
sutartmelson: That's nice to know :)

girder/models/file.py
sutartmelson: The abstract_assetstore_adapter.downloadFile specifies it will return a generator function. So I'll remove the other check.

zachmullen: Let's leave both of these -- I'd like to be permissive with what we accept from adapters in this regard. Currently, you can return either the function or the result of calling it and I believe everything still works, so allowing it here will preserve that capability.

sutartmelson: Okay, I don't think the current implementation will work for generators. I'll change that.

plugins/download_statistics/plugin_tests/download_statistics_test.py
sutartmelson: I think that will be easy. I'm on it.

brianhelba:
This LGTM! I think we're testing all cases now.

girder/models/file.py
```diff
@@ -93,23 +94,62 @@ def download(self, file, offset=0, headers=True, endByte=None,
         :type contentDisposition: str or None
         :type extraParameters: str or None
         """
+        if offset == 0:
+            events.trigger('model.file.download.before', info={'file': file})
+        else:
+            events.trigger('model.file.download.during', info={'file': file})
```
zachmullen: This event should contain the `offset` field.

girder/models/file.py
```diff
@@ -93,23 +94,62 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
         if file.get('assetstoreId'):
-            return self.getAssetstoreAdapter(file).downloadFile(
-                file, offset=offset, headers=headers, endByte=endByte,
-                contentDisposition=contentDisposition,
-                extraParameters=extraParameters)
+            try:
+                return self._handleAssetstoreDownload(file, offset=offset, headers=headers,
+                                                      endByte=endByte,
+                                                      contentDisposition=contentDisposition,
+                                                      extraParameters=extraParameters)
```
zachmullen: We prefer to use the next-line indentation style for cases like this.

girder/models/file.py
```diff
@@ -93,23 +94,62 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
+    def _handleAssetstoreDownload(self, file, offset=0, headers=True, endByte=None,
+                                  contentDisposition=None, extraParameters=None):
+        fileDownload = self.getAssetstoreAdapter(file).downloadFile(
+            file, offset=offset, headers=headers, endByte=endByte,
+            contentDisposition=contentDisposition,
+            extraParameters=extraParameters)
+        if inspect.isgeneratorfunction(fileDownload) or inspect.isgenerator(fileDownload):
+            def downloadGenerator():
+                # Get generator object from function if necessary
+                if inspect.isgeneratorfunction(fileDownload):
```
zachmullen: I don't think this test is necessary, won't it work in either case if you use `for data in gen()`?

girder/models/file.py
zachmullen: IMO it would be even better to just have a single event that contained `startByte` and `endByte` fields.

girder/models/file.py
brianhelba: The short answer is no, but this brings up a really important issue. I'm going to respond more fully in a non-inline comment.

zachmullen: Ah, that's true, sorry for the noise. Not sure why I thought that would work.

girder/models/file.py
sutartmelson: Combine the before and during event into one? Or combine before, during, and after into one?

zachmullen: I would combine "before" and "during" into one called "model.file.download.request" which has startByte and endByte keys, and change the other one to be called "model.file.download.complete", to be more expressive of the fact that the file was assumed to be completely downloaded.

sutartmelson: Yeah that's more descriptive. I like that.

brianhelba:
We should differentiate between two distinct types here:
* Generator functions. These are functions that contain a `yield` keyword, and thus will pass the `inspect.isgeneratorfunction` test. Otherwise, they are ordinary [callable objects (they have a `__call__` method)](https://docs.python.org/2/library/functions.html#callable), but are not iterable.
* Generators. These are returned by calling a generator-returning function. They are [iterator objects (they have `next` and `__iter__` methods, implementing the iterator protocol)](https://docs.python.org/2/library/stdtypes.html#iterator-types), but are not callable.

So we must have the check in the above code, as iterating over a generator function is illegal, and calling a generator is illegal.

---

To expand this more generally, while one branch of this code is expecting: "a generator function, which when called, necessarily returns a generator", Python theoretically allows a set of types that are more general: "a callable, which when called, happens to return an iterable). The first case is just a more specific version of the second, as "all generator functions are functions, and all functions are callable" and "all generators are iterators, and all iterators are iterable).

Now, the tricky thing is that it appears Girder has previously been using the more general version of this: [expecting downloadable objects to come as callable objects](https://github.com/girder/girder/blob/69c274fe2657c0e7b061307306f740a144455398/girder/api/rest.py#L573). This is both more general than this plugin, as it allows non-generator-function callables (and it assumes that they'll return something iterable), but also more restrictive, as it does not allow raw generators (since it tries to call them before returning to CherryPy).

IMO, this behavior has always been needlessly complicated. There's no point in requiring returned downloads to be wrapped in a function, if the function is just going to be called with no arguments. However, for now, we cannot change this, and I think it'd be outside the scope of this PR to broaden the contract for `AbstractAssetstoreAdapter.downloadFile`, particularly when the restriction on streamable download return types is coming further up from [the endpoint handler](https://github.com/girder/girder/blob/69c274fe2657c0e7b061307306f740a144455398/girder/api/rest.py#L563).

---

So, I'd propose we just drop the support for raw generators here entirely. I don't even think we should clutter the code with additional defensive type checks here, rather we should improve the documentation of [`AbstractAssetstoreAdapter.downloadFile`](https://github.com/girder/girder/blob/69c274fe2657c0e7b061307306f740a144455398/girder/utility/abstract_assetstore_adapter.py#L242) to indicate that theoretically, a more general iterable-returning-callable could be returned, rather than the more specific generator-function.

In the future, we could improve the endpoint handler to allow raw iterators to be returned from endpoints.

girder/models/file.py
```diff
@@ -93,19 +93,40 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
+                fileDownload = self.getAssetstoreAdapter(file).downloadFile(
+                    file, offset=offset, headers=headers, endByte=endByte,
+                    contentDisposition=contentDisposition,
+                    extraParameters=extraParameters)
+
+                def downloadGenerator():
+                    for data in fileDownload():
+                        yield data
+                    if endByte is None or endByte >= file['size']:
+                        info['redirect'] = False
```
brianhelba: Since we already sent out `info` under the `'model.file.download.request'` event, I'm concerned about directly modifying it afterwards here. I think we should be setting / adding additional properties to a copy of `info`.

Otherwise, this could have implications for debuggability and plugins that stored a copy of the original `info` object.

sutartmelson: You're right, I totally overlooked that. Thanks!

girder/models/file.py
```diff
@@ -93,19 +93,44 @@ def download(self, file, offset=0, headers=True, endByte=None,
...
+                    endByte, 'redirect': True})
                 raise cherrypy.HTTPRedirect(file['linkUrl'])
             else:
                 endByte = endByte or len(file['linkUrl'])
 
                 def stream():
                     yield file['linkUrl'][offset:endByte]
+                    if endByte >= len(file['linkUrl']):
+                        events.trigger('model.file.download.complete', info={
+                            'file': file, 'startByte': offset, 'endByte':
```
zachmullen: Guess we don't have a style rule to enforce this, but could we write this dict with one key per line?

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+def load(info):
+    # Bind REST events
+    events.bind('model.file.download.request', 'download_statistics',
+                downloadFileRequestEvent)
+    events.bind('model.file.download.complete', 'download_statistics',
+                downloadFileCompleteEvent)
+
+    # Add download count fields to file model
+    ModelImporter.model('file').exposeFields(level=AccessType.READ,
+                                             fields='downloadStatistics')
```
zachmullen: We allow 100 chars per line, so this and the above lines will fit on one line.

Generally speaking, if a line break is necessary in the middle of parameters, next-line indented is preferred to the visual indent style.

plugins/download_statistics/server/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def downloadFileRequest(fileId):
+    ModelImporter.model('file').increment(query={'_id': fileId},
+                                          field='downloadStatistics.requested', amount=1)
+
+
+def downloadFileComplete(fileId):
+    ModelImporter.model('file').increment(query={'_id': fileId},
+                                          field='downloadStatistics.completed', amount=1)
```
zachmullen: I don't think these two functions are necessary, let's just put them in the event handler functions. Also, let's make those event handler functions being with _ since they shouldn't be used outside this module.

plugins/download_statistics/server/__init__.py
sutartmelson: Hmm. Is this a new style practice?
so..
```python
ModelImporter.model('file').increment(query={'_id': event.info['file']['_id']},
    field='downloadStatistics.requested', amount=1)
```
is preferred to

```python
ModelImporter.model('file').increment(query={'_id': event.info['file']['_id']},
                                      field='downloadStatistics.requested',
                                      amount=1)
```
?

sutartmelson: Nevermind, @brianhelba  explained it to me. I will change it.

brianhelba:
@sutartmelson Approved by me, but we'll wait on @zachmullen to approve too.

danlamanna:
Just to chime in briefly - users want to track a lot of different statistics about Girder installations, should we commit to the name `download_statistics` as a plugin that could end up tracking more than downloads? Or are we certain this plugin will remain scoped to just that?

zachmullen:
Good question. Either way we need to make it straightforward for admins configure exactly which things should be tracked -- whether that's by breaking it up into multiple plugins or just one big admin interface doesn't make too much difference to me, but I could be convinced that there would be economic reasons to combine these things into one big plugin.

brianhelba:
I totally agree that this plugin provides an important, useful feature to Girder. I see a possible path forward in exposing this information via the GUI, and starting to show which files, collections, etc. are the most popular.

However, I'm concerned about this:
> The next step (on a future PR) should be to add a new collection called e.g. "file_download" that collects more detailed information about a file download, including IP address, timestamp, and maybe user agent string.

Perhaps I'm misunderstanding how this will be structured, but in my opinion, this is not functionality that should ever belong in Girder, as it violates a clear separation of concerns for what a Python web app should do. This sort of information more properly belongs in log files (or could be piped directly to another analysis tool), but it should not be pushed to our application database. However, it would certainly support improvements to send more of this information to the logging pipeline as it's inevitably captured in requests.

Of course, I'm open to other perspectives, and I'm happy to talk about this further in a more appropriate venue.

zachmullen:
> This sort of information more properly belongs in log files

Why would the proposed table belong in a log file, but the data in this PR belong in the database? I'm anticipating that we'll want to be able to generate reports in the GUI itself that might show geolocated downloads for a given file, and gathering that information from a log file would be trickier than generating the report from structured data in mongo.

brianhelba:
@sutartmelson I'll merge this now for you. :tada:

brianhelba:
@zachmullen While I think we should use the logging mechanism to channel request information, we can certainly set up adapters to funnel this data to an arbitrary output location.

However, as a matter of course, routine HTTP request information should not be dumped into the application database. There is a whole stack of tools that exist to do this sort of analysis for any web app; even if we end up needing to write our own tool; it should live outside of Girder, so it could be reused with arbitrary other web apps.

If we really need aggregated geolocation statistics to be associated with individual files, I think it'd be reasonable to selectively write this to files on a periodic or event-driven basis, but the actual raw data and analysis tools should live outside of Girder.

Let's discuss this further via an issue, meeting, etc., so we don't clutter a closed PR.

sebastienbarre:
Speaking of log aggregations, and "a whole stack of tools that exist to do this sort of analysis for any web app", I'd recommend you wait a few weeks, as we (the KWiK team) are going to communicate about that topic internally. 

We have been in need for more powerful ways to aggregate our log streams for a while, to navigate through them, to search for events, etc. We also want to be able to set up alerts that will be sent to us devs by email or to our slack channel when specific patterns are detected, or during sudden "bursts". 

We investigated a few third-party services with free tiers (such as Papertrail, which certainly does the trick, or Sentry), before management *strongly* voiced that our logs should not fly out of Kitware. We have therefore been working with our sysadmins to setup a dedicated machine that will run [Graylog, an Open Source Log Management](https://www.graylog.org/). It will not be dedicated to KWiK though -- we will write about our experience once we are finished connecting our log pipes from KWiK to this new machine, and hopefully you guys can evaluate if this can benefit other projects.

PR #2105, opened 2017-06-15
Add a Link header to paged results

jbeezley:
This follows [RFC5988](https://tools.ietf.org/html/rfc5988) adding a `Link` header for navigating paged results from our rest API when using `pagingParams` in `autoDescribeRoute`.  There is a description of how this works with the Github API here: https://developer.github.com/v3/guides/traversing-with-pagination/.

### DIFF
--- a/girder/api/describe.py
+++ b/girder/api/describe.py
@@ -343,7 +343,7 @@ class Description(object):
 
         return self
 
-    def pagingParams(self, defaultSort, defaultSortDir=1, defaultLimit=50):
+    def pagingParams(self, defaultSort, defaultSortDir=1, defaultLimit=50, linkHeader=False):
         """
         Adds the limit, offset, sort, and sortdir parameter documentation to
         this route handler.
@@ -354,6 +354,10 @@ class Description(object):
         :type defaultSortDir: int
         :param defaultLimit: The default page size.
         :type defaultLimit: int
+        :param linkHeader: Add a `Link` header for paging.  This assumes that the value returned
+            by the endpoint is a list or tuple.  The limit passed to the endpoint function will
+            be artifically incremented by one to detect the last page, and the result sent back
+            to the client will be truncated to the requested size.
         """
         self.param(
             'limit', 'Result set size limit.', default=defaultLimit, required=False, dataType='int')
@@ -368,6 +372,7 @@ class Description(object):
                 required=False, dataType='integer', enum=(1, -1), default=defaultSortDir)
 
         self.hasPagingParams = True
+        self.addLinkHeader = linkHeader
         return self
 
     def consumes(self, value):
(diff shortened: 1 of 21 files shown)


girder/api/describe.py
```diff
@@ -598,14 +598,50 @@ def wrapped(*args, **kwargs):
...
+           not cherrypy.request.method == 'GET':
+            return
+
+        qs = cherrypy.lib.httputil.parse_query_string(cherrypy.request.query_string)
+        offset = params['offset']
+        limit = params['limit']
+        link = []
+
+        qs['offset'] = 0
+        link.append('<%s>; rel="first"' % cherrypy.url(qs=qs))
```
jbeezley: Now that I think of it, I'm not sure this will provide the correct link when proxying the API server unless the proxy rewrites the link header.  Should I use the configured API root here instead?

zachmullen: IIRC cherrypy.url() will work fine with a proxy as long as tools.proxy.on = True and the proxy server passes the proxy headers correctly. If the proxy server is not well-behaved, you can manually specify the value with tools.proxy.base.

manthey:
If we want to know if there is more data on the next page, we could adjust endpoints that take a limit to query for `limit+1` items, but only return `limit` items, using the existence of the last item as an indicator of more data.  We could make

jbeezley:
That sounds reasonable, but makes me a little nervous about backwards compatibility.  Perhaps there are cases where the user of `autoDescribeRoute` was doing something internally that depends on the real limit from the request.  Is that overly paranoid?

manthey:
I don't know -- perhaps we need another parameter for `pagingParams`?

Sort of related, I was thinking that we might want to make a change to the `sort` option as well to allow json lists of sort parameters in general.

jbeezley:
Yes, I suppose I could make the behavior opt-in.  I'll give a try.

danlamanna:
Worth noting that `requests` handles these quite nicely: http://docs.python-requests.org/en/master/user/advanced/#link-headers

jbeezley:
I decided it was overly paranoid to make the last page detection opt-in.  I think if it turns out that it is breaking for anyone, we can treat it as a bug and make it opt-in in a different PR.

jbeezley:
I guess this does break the `/resource/search` endpoint because it doesn't return a list.  I'll have to make it opt-in.  :(

manthey:
`/resource/search` with `limit` is unusual...  Maybe make it opt-out?

jbeezley:
Still, if another endpoint in a plugin somewhere does something similar, it would be a breaking change.

jbeezley:
@girder/developers I think I worked out all the kinks here.  PTAL

brianhelba:
I'll review this now.

Pierre-Assemat:
Do you think I could use this behavior to help with the pagination on [the Search Result page](https://github.com/girder/girder/pull/2347) ? @brianhelba

danlamanna:
@jbeezley @brianhelba What would have to be done for this PR to get merged?

brianhelba:
@danlamanna I think we all agree that the outward-facing behavior is positive, but the implementation in this PR is probably too magical to be generally useful and stable. We need to step back and reconsider the HTTP handler lifecycle, and find a more natural place to declare parameter properties, parse them, and set response headers. This is more of an architecture task than a code one.

danlamanna:
> @danlamanna I think we all agree that the outward-facing behavior is
> positive, but the implementation in this PR is probably too magical to
> be generally useful and stable. 

I'm not sure I agree with this sentiment, the actual logic in this PR is
~30-40 lines of fairly simple code.

> We need to step back and reconsider the HTTP handler lifecycle, and
> find a more natural place to declare parameter properties, parse them,
> and set response headers. This is more of an architecture task than a
> code one.

I understand this, and it's possible using CherryPy tools is the correct
approach here:
http://docs.cherrypy.org/en/latest/extend.html#stateful-tools.

mgrauer:
Closing due to lack of will. Should be rewritten once #2614 goes in.

PR #2106, opened 2017-06-15
Add support for Elastic Beanstalk

danlamanna:
This PR adds pre-packaged configurations for running Girder on Elastic Beanstalk.

I expect these configurations to change a bit in the coming weeks/months, but for now they are working in a stable enough fashion that I think it should be merged.

### DIFF
--- a/docs/deploy.rst
+++ b/docs/deploy.rst
@@ -243,3 +243,71 @@ After everything starts, which may take a few minutes, you should be able
 to visit your Girder instance at ``http://X.X.X.X`` where ``X.X.X.X`` is the
 IP address in the container description above. Congratulations, you
 have a full Girder instance available on Google Container Engine!
+
+Elastic Beanstalk
+-----------------
+
+Girder comes with pre-packaged configurations for deploying onto Elastic Beanstalk's
+`Python platform <http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/concepts.platforms.html#concepts.platforms.python>`_
+(both 2.7 and 3.4).
+
+The configurations live within ``devops/beanstalk`` and are designed to be copied into your working Girder directory
+at deploy time.
+
+The following assumes you have a checked out copy of Girder (using git) and an existing MongoDB instance which
+can be accessed by your Beanstalk application.
+
+.. note:: It is **highly** recommended to perform the following steps in an isolated virtual
+	  environment using pip. For more see the documentation for `Virtualenv <https://virtualenv.pypa.io/en/stable/>`_.
+
+From within the checked out copy of Girder, install and configure the CLI tools: ::
+
+  $ pip install awscli awsebcli
+  $ aws configure
+
+Initialize the Beanstalk application with a custom name. This is an interactive process
+that will ask various questions about your setup (see above for supported platforms): ::
+
+  $ eb init my-beanstalk-app
+
+Build Girder and its client-side assets locally: ::
+
+  $ pip install -e .[plugins]  # optionally build specific plugins
+  $ girder-install web --all-plugins  # optionally build specific plugins with --plugins
+
+.. seealso::
+
+   `Building specific plugins with pip <http://girder.readthedocs.io/en/latest/installation.html#installing-extra-dependencies-with-pip>`_.
+
+.. note:: Since Girder is unable to restart and load plugins in the Beanstalk environment,
+	  plugins may be enabled/disabled but will require a restart of Beanstalk application
+	  servers to take effect. Restarting application servers can be performed from the
+	  `Environment Management Console <http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environments-console.html>`_.
+
+Create a requirements.txt for the Beanstalk application, overwriting the default Girder requirements.txt: ::
+
+  $ pip freeze | grep -v '^girder\|^awscli\|^awsebcli' > requirements.txt
+
+Copy the pre-packaged configurations for Beanstalk into the current directory: ::
+
+  $ cp -r devops/beanstalk/. .
+
+.. note:: These are just the default tested Beanstalk configurations. It's likely that these will have to
+	  be modified to suit individual deployments.
+
+Beanstalk deploys code based on commits, so create a git commit with the newly added configurations: ::
+
+  $ git add . && git commit -m "Add Beanstalk configurations"
+
+Create an environment to deploy code to: ::
+
+  $ eb create my-env-name --envvars \
+    GIRDER_CONFIG=/opt/python/current/app/girder.cfg,GIRDER_MONGO_URI=mongodb://my-mongo-uri:27017/girder
+
+At this point running ``eb open my-env-name`` should open a functioning Girder instance
+in your browser. Additionally, running ``eb terminate`` will terminate the newly created environment.
+
+.. seealso::
+
+   It may be useful when deploying to AWS to make use of the built-in Girder support
+   for `S3 Assetstores <http://girder.readthedocs.io/en/latest/user-guide.html#assetstores>`_.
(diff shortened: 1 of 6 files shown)


docs/deploy.rst
```diff
@@ -243,3 +243,67 @@ After everything starts, which may take a few minutes, you should be able
...
+in your browser.
+
+.. note:: The pre-packaged configurations work with Amazon CloudWatch for aggregating log streams
+	  across many application servers. For this to work, the EC2 instances will need the proper
+	  policy attached to write to CloudWatch.
+
+.. seealso::
+
+   It may be useful when deploying to AWS to make use of the built-in support
+   for `S3 Assetstores <http://girder.readthedocs.io/en/latest/user-guide.html#assetstores>`_.
```
kotfic: Can we add the ```eb``` command for terminating the instance as well?

docs/deploy.rst
```diff
@@ -243,3 +243,67 @@ After everything starts, which may take a few minutes, you should be able
...
+
+Initialize the Beanstalk application with a custom name. This is an interactive process
+that will ask various questions about your setup: ::
+
+  $ eb init my-beanstalk-app
+
+Build Girder and its client-side assets locally: ::
+
+  $ pip install .[plugins]  # optionally only build specific plugins
+  $ girder-install web  # optionally specific plugins with --plugins
```
mgrauer: optionally build specific

docs/deploy.rst
```diff
@@ -243,3 +243,67 @@ After everything starts, which may take a few minutes, you should be able
...
+
+.. note:: Since Girder is unable to restart and load plugins in the Beanstalk environment,
```
mgrauer: need more detail here.  do we need to redeploy and then restart the BS env, or what is the process?

docs/deploy.rst
```diff
@@ -243,3 +243,67 @@ After everything starts, which may take a few minutes, you should be able
...
+  $ aws configure
+
+Initialize the Beanstalk application with a custom name. This is an interactive process
+that will ask various questions about your setup: ::
+
+  $ eb init my-beanstalk-app
+
+Build Girder and its client-side assets locally: ::
+
+  $ pip install .[plugins]  # optionally only build specific plugins
```
mgrauer: mention that plugins is actually a list of plugins to be installed, or point to the relevant docs

docs/deploy.rst
```diff
@@ -243,3 +243,67 @@ After everything starts, which may take a few minutes, you should be able
...
+  $ girder-install web  # optionally specific plugins with --plugins
+
+.. note:: Since Girder is unable to restart and load plugins in the Beanstalk environment,
+	  plugins may be enabled/disabled but will require a restart of Beanstalk application
+	  servers to take effect.
+
+Create a requirements.txt for the Beanstalk application: ::
+
+  $ pip freeze | grep -v '^girder\|^awscli\|^awsebcli' > requirements.txt
```
mgrauer: should you mention working in a virtualenv created for this deployment, since you are pip freezing?

docs/deploy.rst
```diff
@@ -243,3 +243,67 @@ After everything starts, which may take a few minutes, you should be able
...
+Build Girder and its client-side assets locally: ::
+
+  $ pip install .[plugins]  # optionally only build specific plugins
+  $ girder-install web  # optionally specific plugins with --plugins
+
+.. note:: Since Girder is unable to restart and load plugins in the Beanstalk environment,
+	  plugins may be enabled/disabled but will require a restart of Beanstalk application
+	  servers to take effect.
+
+Create a requirements.txt for the Beanstalk application: ::
```
mgrauer: is this still in the girder directory?  that will overwrite the girder requirements.txt, is that what you want?

docs/deploy.rst
mgrauer: Unless Girder is installed via a pip editable install, the client assets will be built inside the virtualenv directory, and therefore won't be inside the Girder repo, and won't be included when the app is deployed to EB.

docs/deploy.rst
```diff
@@ -243,3 +243,67 @@ After everything starts, which may take a few minutes, you should be able
...
+At this point running ``eb open my-env-name`` should open a functioning Girder instance
+in your browser.
+
+.. note:: The pre-packaged configurations work with Amazon CloudWatch for aggregating log streams
+	  across many application servers. For this to work, the EC2 instances will need the proper
+	  policy attached to write to CloudWatch.
+
+.. seealso::
+
+   It may be useful when deploying to AWS to make use of the built-in support
```
mgrauer: clarify this to say `built-in Girder support`

docs/deploy.rst
```diff
@@ -243,3 +243,67 @@ After everything starts, which may take a few minutes, you should be able
...
+Beanstalk deploys code based on commits, so create a git commit with the newly added configurations: ::
+
+  $ git add . && git commit -m "Add Beanstalk configurations"
+
+Create an environment to deploy code to: ::
+
+  $ eb create my-env-name --envvars \
+    GIRDER_CONFIG=/opt/python/current/app/girder.cfg,GIRDER_MONGO_URI=mongodb://my-mongo-uri:27017/girder
+
+At this point running ``eb open my-env-name`` should open a functioning Girder instance
```
mgrauer: also mention that you can find the URL by going through the AWS console Elastic Beanstalk service.

mgrauer:
This worked for me, the docs need some clarifications, but the only thing that needed a specific fix was the editable install part.

danlamanna:
Docs have been updated, PTAL.

docs/deploy.rst
```diff
@@ -243,3 +243,75 @@ After everything starts, which may take a few minutes, you should be able
...
+-----------------
+
+Girder comes with pre-packaged configurations for deploying onto Elastic Beanstalk's
+`Python platform <http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/concepts.platforms.html#concepts.platforms.python>`_
+(both 2.7 and 3.4).
+
+The configurations live within ``devops/beanstalk`` and are designed to be copied into your working Girder directory
+at deploy time.
+
+The following assumes you have a a checked out copy of Girder (using git) and an existing MongoDB instance which
```
mgrauer: a a

docs/deploy.rst
```diff
@@ -243,3 +243,75 @@ After everything starts, which may take a few minutes, you should be able
...
+  $ aws configure
+
+Initialize the Beanstalk application with a custom name. This is an interactive process
+that will ask various questions about your setup (see above for supported platforms): ::
+
+  $ eb init my-beanstalk-app
+
+Build Girder and its client-side assets locally: ::
+
+  $ pip install -e .[plugins]  # optionally build specific plugins
```
mgrauer: You should clarify that plugins should be replaced with a list of plugins that are desired to be installed and have python dependencies.  This reads as commands you want people to be typing into their terminal, but "plugins" isn't something they'd ever type.

Maybe

```
Build Girder and its client-side assets locally: ::

When not installing any plugins with Python dependencies:
    
    $ pip install -e .

An example when installing a plugin with a Python dependency, in this case the `dicom_viewer`.  Use a comma separated list in the case of multiple plugins.

    $ pip install -e .[dicom_viewer]
```

danlamanna: plugins is a realistic value that install dependencies for all plugins, this is in line with the client side step which also installs `--all-plugins`.

mgrauer: ok, thanks for clarifying.  I thought I had tried this before and ran into trouble with it, but I must have been confused about what happened or maybe this was related to the `-e`.

danlamanna:
@mgrauer I addressed the typo, and I was unable to reproduce your issue with enabling plugins.

PR #2109, opened 2017-06-15
Add a new girderTest.importPlugin function

brianhelba:
This simplifies boilerplate code in web client plugin tests.

* Add a new `girderTest.importPlugin` function, to load plugin JS and CSS
* Deprecate `girderTest.addCoveredScript` and `girderTest.addCoveredScripts`
* Serve all plugin test files from `/static/built` URLs
* Add documentation on the use of `girderTest.importPlugin`

### DIFF
--- a/clients/web/test/testUtils.js
+++ b/clients/web/test/testUtils.js
@@ -670,6 +670,14 @@ girderTest.importStylesheet = function (css) {
 };
 
 /**
+ * Import the JS and CSS files for a plugin.
+ */
+girderTest.importPlugin = function (pluginName) {
+    girderTest.addScript('/static/built/plugins/' + pluginName + '/plugin.min.js');
+    girderTest.importStylesheet('/static/built/plugins/' + pluginName + '/plugin.min.css');
+};
+
+/**
  * For the current folder, check if it is public or private and take an action.
  * :param current: either 'public' or 'private': expect this value to match.
  * :param action: if 'public' or 'private', switch to that setting.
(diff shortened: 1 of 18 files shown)


clients/web/test/testUtils.js
```diff
@@ -670,6 +658,32 @@ girderTest.importStylesheet = function (css) {
 };
 
 /**
+ * Import the JS and CSS files for a plugin.
+ */
+girderTest.importPlugin = function (pluginName) {
+    girderTest.addScript('/static/built/plugins/' + pluginName + '/plugin.min.js');
+    girderTest.importStylesheet('/static/built/plugins/' + pluginName + '/plugin.min.css');
```
zachmullen: Some plugins don't have CSS files, is that OK in this case?

brianhelba: Yeah, it will just add a new `<link>` tag at the top of the page, pointing to a non-existent URL, which is harmless.

brianhelba:
@girder/developers PTAL. I'd like to get this reviewed+merged ASAP, as I've got additional PRs queued up, which depend on this one.

clients/web/test/testUtils.js
```diff
@@ -670,6 +658,32 @@ girderTest.importStylesheet = function (css) {
 };
 
 /**
+ * Import the JS and CSS files for a plugin.
+ */
+girderTest.importPlugin = function (pluginName) {
```
zachmullen: It would be nice to allow this to import a list of plugins, either by passing a list, or via `arguments`.

brianhelba: @zachmullen  Done.

Issue #2088, opened 2017-06-09
Allow data to be included as part of POST /file


manthey:
Allow data to included as part of `POST /file` like it is included in `POST /file/chunk`.  This would be equivalent of calling `POST /file` without the data and then calling `POST /file/chunk` with the data and an offset of 0.  If the data is the entire file, `POST /file` would return a file object.  If it is not the entire file, it would return an upload object.

This allows small files to be sent as a single request.  For files large enough that they must be chunked, it can still reduce the number of requests by 1.

Increase the version number and allow the girder-cli to use this more efficient method for the first chunk of a file.

manthey:
We may also want to make a similar change to `PUT /file/{id}/contents`.

manthey:
If the data is the complete data for the file, I'd like to consider avoiding saving the upload to the db (which would be a change to `createUpload` and `createUploadToFile`).  Doing so would skip sending the `model.upload.(validate|save|save.created|save.after)` events, but would speed things up.  This might be undesirable as those events may be monitored.

manthey:
Added ind PR #2111.

zachmullen:
We need to test this with S3 assetstores in the direct upload case, to make sure our clients work in this new mode when dealing with a redirect.

PR #2111, opened 2017-06-15
Support single POST uploads.

manthey:
If the parameters on a `POST /file` endpoint are included in the URL and the body is not processed by cherrypy and exists, then use the body as the first chunk of the upload.  Alternately, if there is a `chunk` parameter, use that as the first chunk of the upload.

If multiple chunks are needed, the first can be sent via `POST /file`, and the rest must be sent via `POST /file/chunk`.

If the file is completely uploaded, `POST /file` will return a file resource.  Otherwise, it will return an upload resource.

Skip saving an upload model when the file is uploaded as a single chunk via the `POST /file` endpoint.  If only the first chunk of multiple chunks is sent via the `POST /file` endpoint, then the upload model will be saved at the end of saving the first chunk.

Once this is merged to Girder master and there exists a new version, the girder client can be modified to use a single POST for small uploads (or always start with a POST for the first part of an upload).

This resolves issue #2088.

### DIFF
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -71,8 +71,8 @@ class File(Resource):
                'The data for the first chunk of the file can be included with '
                'this query by sending it as the body of the request using an '
                'appropriate content-type and with the other parameters as '
-               'part of the query string.  If the entire file is uploaded as '
-               'a single via this call, the resulting file is returned.')
+               'part of the query string.  If the entire file is uploaded via '
+               'this call, the resulting file is returned.')
         .responseClass('Upload')
         .param('parentType', 'Type being uploaded into.', enum=['folder', 'item'])
         .param('parentId', 'The ID of the parent.')
@@ -126,7 +126,7 @@ class File(Resource):
                 chunk = None
 
             try:
-                # This can be made more efficient by adding
+                # TODO: This can be made more efficient by adding
                 #    save=chunk is None
                 # to the createUpload call parameters.  However, since this is
                 # a breaking change, that should be deferred until a major
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -67,6 +67,12 @@ class File(Resource):
     @access.user(scope=TokenScope.DATA_WRITE)
     @autoDescribeRoute(
         Description('Start a new upload or create an empty or link file.')
+        .notes('Use POST /file/chunk to send the contents of the file.  '
+               'The data for the first chunk of the file can be included with '
+               'this query by sending it as the body of the request using an '
+               'appropriate content-type and with the other parameters as '
+               'part of the query string.  If the entire file is uploaded as '
+               'a single via this call, the resulting file is returned.')
         .responseClass('Upload')
         .param('parentType', 'Type being uploaded into.', enum=['folder', 'item'])
         .param('parentId', 'The ID of the parent.')
@@ -197,6 +203,12 @@ class File(Resource):
     @access.user(scope=TokenScope.DATA_WRITE)
     @autoDescribeRoute(
         Description('Upload a chunk of a file.')
+        .notes('The data for the chunk should be sent as the body of the '
+               'request using an appropriate content-type and with the other '
+               'parameters as part of the query string.  Alternately, the '
+               'data can be sent as a file in the "chunk" field in multipart '
+               'form data.  Multipart uploads are much less efficient and '
+               'their use is deprecated.')
         .modelParam('uploadId', paramType='formData')
         .param('offset', 'Offset of the chunk in the file.', dataType='integer',
                paramType='formData')
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -123,7 +123,7 @@ class File(Resource):
                 upload = self.model('upload').createUpload(
                     user=user, name=name, parentType=parentType, parent=parent, size=size,
                     mimeType=mimeType, reference=reference, assetstore=assetstore,
-                    saveUpload=chunk is None)
+                    save=chunk is None)
             except OSError as exc:
                 if exc.errno == errno.EACCES:
                     raise GirderException(
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -111,15 +111,7 @@ class File(Resource):
                 assetstore = self.model('assetstore').load(assetstoreId)
 
             chunk = None
-            if 'chunk' in params:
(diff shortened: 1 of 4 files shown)


girder/api/v1/file.py
```diff
@@ -109,16 +109,38 @@ def initUpload(self, parentType, parentId, name, size, mimeType, linkUrl, refere
                 self.requireAdmin(
                     user, message='You must be an admin to select a destination assetstore.')
                 assetstore = self.model('assetstore').load(assetstoreId)
+
+            chunk = None
+            if 'chunk' in params:
+                chunk = params['chunk']
+                if isinstance(chunk, cherrypy._cpreqbody.Part):
```
zachmullen: Since this is a new function, we don't need to support multipart encoding, and IMO we should not.

girder/models/upload.py
```diff
@@ -305,7 +308,8 @@ def createUploadToFile(self, file, user, size, reference=None,
         return self.save(upload)
 
     def createUpload(self, user, name, parentType, parent, size, mimeType=None,
-                     reference=None, assetstore=None, attachParent=False):
+                     reference=None, assetstore=None, attachParent=False,
+                     saveUpload=True):
```
zachmullen: We should document this parameter. I'm also inclined to just call it `save`, since that's what we call it in most other places.

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -219,8 +219,9 @@ def finalizeUpload(self, upload, file):
 
         # Store the hash in the upload so that deleting a file won't delete
         # this file
-        upload['sha512'] = hash
-        self.model('upload').update({'_id': upload['_id']}, update={'$set': {'sha512': hash}})
+        if '_id' in upload:
```
zachmullen: It seems we are changing some behavior that assetstore implementations may have been relying on -- to be safe, I think we should defer the no-save optimization until 3.0, and save the upload model even in a single-shot case.

manthey: My reasoning, probably faulty, was that we don't save an upload model when we import files, so not saving them when uploading would be okay.

We save the upload multiple times (which is part of why this is a significant speed up), and in none of them do we check to see if the process is cancelled, though a handler could throw an exception.

The user quota plugin does use the initial save to check if there is enough space, but, since multiple uploads can be started concurrently, it also checks when the upload is finalized (and that event still occurs).  Perhaps wrongly, the user quota counts imported files the same as uploaded files.

girder/api/v1/file.py
manthey: Okay.  I wanted to make the endpoints behave mostly identically.  I didn't find any user-facing documentation about support of multipart encoding and the preference to not use it.

girder/utility/filesystem_assetstore_adapter.py
zachmullen: > was that we don't save an upload model when we import files, so not saving them when uploading would be okay.

The import logic and the upload logic go through different pathways in the assetstore code, so there's still risk in making this change that we break downstream implementations that were doing something like the filesystem assetstore was doing.

> We save the upload multiple times (which is part of why this is a significant speed up)

I do think it's a valuable optimization, and would love to have it now, but preserving behavior is more important IMO. My recommendation would be to open an issue and mark it with the **3.0** and **performance** labels. When we do release planning, it will go into the backlog of things to do prior to the release. I'd be OK with starting a 3.0-integration branch now if you'd like to push the code now rather than waiting.

> Perhaps wrongly, the user quota counts imported files the same as uploaded files.

It's nice that the user quota logic can be naive and just use the size totals -- since importing is an admin-only operation, we assume admins can also expand the destination quota as needed.

girder/api/v1/file.py
zachmullen: Hm, you're right -- I thought we had documented this preference somewhere, but I can't find anything. I don't think we've deprecated it per se since we probably want to keep it around for a while for very old clients, but we definitely need to officially discourage it. I'll open a ticket.

girder/utility/filesystem_assetstore_adapter.py
manthey: I've left the bulk of the code in place, but now it doesn't ever ask to not save uploads.  I can change the message about future changes to something that has a "for Girder 3" text string in it if we have such a thing.  Once this goes in, I'll create the 3.0/performance issue.

girder/api/v1/file.py
manthey: I've added a note the `POST /file` and `POST /file/chunk` with some documentation.

girder/api/v1/file.py
```diff
@@ -67,6 +67,12 @@ def getFile(self, file, params):
     @access.user(scope=TokenScope.DATA_WRITE)
     @autoDescribeRoute(
         Description('Start a new upload or create an empty or link file.')
+        .notes('Use POST /file/chunk to send the contents of the file.  '
+               'The data for the first chunk of the file can be included with '
+               'this query by sending it as the body of the request using an '
+               'appropriate content-type and with the other parameters as '
+               'part of the query string.  If the entire file is uploaded as '
+               'a single via this call, the resulting file is returned.')
```
zachmullen: Probably remove "as a single" or change to "as a single chunk"

girder/api/v1/file.py
```diff
@@ -109,7 +115,22 @@ def initUpload(self, parentType, parentId, name, size, mimeType, linkUrl, refere
...
+            if size > 0 and cherrypy.request.headers.get('Content-Length'):
+                ct = cherrypy.request.body.content_type.value
+                if (ct not in cherrypy.request.body.processors and
+                        ct.split('/', 1)[0] not in cherrypy.request.body.processors):
+                    chunk = RequestBodyStream(cherrypy.request.body)
+            if chunk is not None and chunk.getSize() <= 0:
+                chunk = None
+
             try:
+                # This can be made more efficient by adding
```
zachmullen: If you add a TODO here, I will catch it via grep during release planning

zachmullen:
LGTM other than those two minor nits

zachmullen:
Are you planning to change our two clients to use this new mode?

manthey:
Yes, I'm planning on changing the clients.  I locally have change to the python client.  Since elsewhere we query the server version and then use that to determine the upload strategy, I figured on doing the same thing, which means that the girder version needs to change before I can have the clients use the new method.

PR #2112, opened 2017-06-15
Fix static analysis of plugin test files

brianhelba:
Static analysis tests on plugin test specs was previously broken. This fixes the issue, and removes some extra boilerplate from plugin test files.

* Fix an incorrect path, which prevented plugin test static analysis from running
* Use additional eslint rules for plugin test files
* Remove unnecessary eslint overrides from test files
* Fix style errors in tests
* Rename the "_prepareTestUpload" function to "girderTest._prepareTestUpload"

### DIFF
--- a/clients/web/test/spec/.eslintrc
+++ b/clients/web/test/spec/.eslintrc
@@ -1,5 +1,0 @@
-{
-  "globals": {
-    "_prepareTestUpload": true
-  }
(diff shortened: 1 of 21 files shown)


brianhelba:
Fixed the failure with a change to the final commit. I failed to stage 2 files before.

PR #2114, opened 2017-06-16
Test girder installation through cli 

sutartmelson:
This changes the install tests to install with command line arguments rather than calling the internal installation fictions directly. 

I think this is a better approach as it actually tests the cli arguments. This also increases the test coverage of the file significantly.

### DIFF
--- a/tests/cases/install_test.py
+++ b/tests/cases/install_test.py
@@ -18,6 +18,7 @@
 ###############################################################################
 
 import mock
+import sys
 import os
 import six
 import shutil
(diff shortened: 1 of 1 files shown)


sutartmelson:
I brought this up briefly on #2091. This PR address the low code coverage of #2091, though it doesn't seem to show up on codecov/project check. @zachmullen any idea why it says no report found?

zachmullen:
Not sure what causes this, I've seen it before from codecov though. We can merge without it passing.

manthey:
I think codecov doesn't produce the correct project report if the change is entirely on files that are untracked in coverage, even if the change does affect the project coverage.

PR #2118, opened 2017-06-19
Fixes #2117. Use requireScope method in item_tasks

### LINKED ISSUE #2117: Add a token model method to require token scope
Calling [`ensureTokenScope`](https://github.com/girder/girder/blob/637ab0e919bc355cbcb63a827eaabb8073c94408/girder/api/rest.py#L1026) always allows user auth tokens. Endpoint implementations that want to require a token scope need to do something like:
````
if not self.model('token').hasScope(token, scope):
    raise AccessException('Invalid token scopes: ...')
````
See discussion: https://github.com/girder/girder/pull/2098#discussion_r122524023

### PR DESCRIPTION
**Pull request by zachmullen**

### CODE REVIEW

File: plugins/item_tasks/plugin_tests/tasks_test.py [hunk 1]
```diff
@@ -103,12 +103,16 @@ def testAddItemTasksToFolderFromJson(self):
         with open(os.path.join(os.path.dirname(__file__), 'spec.json')) as f:
             spec = f.read()
         parsedSpec = json.loads(spec)
+
+        token = self.model('token').createToken(
+            user=self.admin, scope='item_task.set_task_spec.%s' % folder2['_id'])
```
msmolens: Extract the token here as well, and a couple place below.

zachmullen: I fixed it up in every place I could, but this particular one I left as an artificially created token since we don't run the job creation endpoint in advance of that one.

msmolens: Makes sense. Would it be good to negative test as well by, say, sending a request without a token?

zachmullen: I generally avoid negative testing specific endpoints, instead I'd rather rely on our standard security protocols and just have them well-tested in isolation. These cases are exceptions to normal authorization controls, so maybe we want to add more negative testing for them, but if so, let's do it on a separate topic.

PR #2119, opened 2017-06-19
Fix uploading into HDFS assetstore using new upload mode

zachmullen:
Our new upload mode was not being tested on HDFS assetstores, and it was broken because the RequestBodyStream object is not properly handled by requests as a valid `data` field. By wrapping it in an iterator, everything works.


mgrauer:
Is there a way to make testing the HDFS easier?  Could we package a Docker container with HDFS that would be easily mountable as an assetstore?  How about a container with HDFS that we use in Circle 2.0?

PR #2120, opened 2017-06-20
Cleanup Webpack watch mode

brianhelba:
This makes the configuration of Webpack's shared and core build settings more robust and maintainable. It also adds support for watch mode to be run via file polling, which is necessary when running inside most VMs (automatic toggling of this behavior will be added in a subsequent PR).

A full list of commits / changes:
* Remove unused webpack-dev-server
  * Even in watch mode, CherryPy is always used to serve Girder.
* Upgrade grunt-webpack and clean up the Grunt configuration of Webpack
* Ensure that watch mode always implies development mode too
* Add support for Webpack polling watch, via `WATCH_USEPOLLING` envvar
* Update Webpack config using a cleaner syntax
* Ensure that Webpack progress display is always correctly toggled
* Remove the unused Grunt-based watch task


brianhelba:
@girder/developers PTAL.

PR #2123, opened 2017-06-21
Add a package-lock file for core dependencies on npm@5

brianhelba:
This is probably the best sane way to force npm packages to the desired versions under npm@5.

This handles core dependencies (both production and dev) only. When production-only dependencies are installed by `npm install --production`, only a subset of this file will be used.

Plugins will not be covered by this, but will be directly installed with `npm install foo@3.2.1. ....` commands, after the core packages are installed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-/package-lock.json
 *.py[cod]
 node_modules
 docs/_build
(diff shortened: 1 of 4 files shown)


brianhelba:
This needs a little more work:
* It needs more testing to ensure dependency conflicts from plugins don't totally break things. However, I think we may need to accept a little bit of fragility, given that the current system never appropriately upgrades packages under normal use.
* It needs additional documentation about the process for updating `package-lock.json`.
* CircleCI setup should probably be tweaked.

brianhelba:
@girder/developers @jbeezley Ok, ready for review.

PR #2130, opened 2017-06-23
Fix issues with girder-install command

brianhelba:
* Ensure that "girder-install web" never modifies "package-lock.json"
* Clean up girder-install CLI documentation


brianhelba:
PTAL @girder/developers ; this fixes a bug with the command that's going to annoy developers until it's fixed.

PR #2133, opened 2017-06-23
Filter task by file input count

sutartmelson:
This allows the user to filter tasks by min and max number of file inputs.

@jeffbaumes


sutartmelson:
@zachmullen Can you take a look at this? I'm a little unsure about the tests I wrote for it. 

This allows you to specify a min and/or max number of file inputs when requesting tasks through the /item_task endpoint.

plugins/item_tasks/server/rest.py
```diff
@@ -34,17 +34,44 @@ def __init__(self):
     @autoDescribeRoute(
         Description('List all available tasks that can be executed.')
         .pagingParams(defaultSort='name')
+        .param('minFileInputs', 'Search tasks by minimum number of file inputs.', required=False,
+               dataType='int', default=0)
+        .param('maxFileInputs', 'Search tasks by minimum number of file inputs.', required=False,
```
zachmullen: maximum*

plugins/item_tasks/server/rest.py
```diff
@@ -34,17 +34,44 @@ def __init__(self):
...
+
+                if maxFileInputs:
+                    if fileCount >= minFileInputs and fileCount <= maxFileInputs:
+                        finalResults.append(item)
+                else:
+                    if fileCount >= minFileInputs:
+                        finalResults.append(item)
+            except KeyError:
+                continue
+        return finalResults
```
zachmullen: With this logic, the result set size could be smaller than the requested page size, even if there are more pages. We should encode this filtering in the model layer, then wrap the result in filterResultsByPermission().

sutartmelson: Is this done by changing the functionality of filterResultsByPermission or by changing the cursor argument?

zachmullen: We could provide these as arguments to a new model method that would be a generator function, iterating over the raw `find()` results and yielding only those that matched the filter parameters.

sutartmelson:
@zachmullen I believe I have made the necessary changes regarding filtering properly.

plugins/item_tasks/server/rest.py
```diff
@@ -34,17 +34,41 @@ def __init__(self):
...
+        if minFileInputs or maxFileInputs:
+            cursor = self._filterMinMaxFileInputs(cursor, minFileInputs, maxFileInputs)
+
         return list(self.model('item').filterResultsByPermission(
             cursor, self.getCurrentUser(), level=AccessType.READ, limit=limit, offset=offset,
             flags=constants.ACCESS_FLAG_EXECUTE_TASK))
 
+    def _filterMinMaxFileInputs(self, cursor, minFileInputs, maxFileInputs):
+        for item in cursor:
+            fileCount = 0
```
brianhelba: How about:
```
fileCount = sum(
    input['type'] == 'file'
    for input in
    item['meta']['itemTaskSpec'].get('inputs', [])
)
```
This would be simpler, and would remove the need for `try...except`.

plugins/item_tasks/server/rest.py
```diff
@@ -34,17 +34,41 @@ def __init__(self):
...
+            try:
+                for _input in item['meta']['itemTaskSpec']['inputs']:
+                    if _input['type'] == 'file':
+                        fileCount += 1
+
+                if maxFileInputs:
```
brianhelba: How about
```
if minFileInputs and fileCount < minFileInputs:
    continue
elif maxFileInputs and fileCount > maxFileInputs:
    continue
else:
   yield item
```
Maybe a little simpler?

plugins/item_tasks/server/rest.py
```diff
@@ -34,17 +34,41 @@ def __init__(self):
     @autoDescribeRoute(
         Description('List all available tasks that can be executed.')
         .pagingParams(defaultSort='name')
+        .param('minFileInputs', 'Search tasks by minimum number of file inputs.', required=False,
+               dataType='int', default=0)
```
brianhelba: While `default=0` and `default=None` (the implicit default) have the same truthiness and effect on the code, perhaps it'd be more consistent to make them both default to `None`? I don't feel really strongly though.

plugins/item_tasks/server/rest.py
```diff
@@ -34,17 +34,41 @@ def __init__(self):
     @autoDescribeRoute(
         Description('List all available tasks that can be executed.')
         .pagingParams(defaultSort='name')
+        .param('minFileInputs', 'Search tasks by minimum number of file inputs.', required=False,
```
brianhelba: I'd call this `Filter tasks by...`.

plugins/item_tasks/server/rest.py
```diff
@@ -34,17 +34,41 @@ def __init__(self):
...
+               dataType='int', default=0)
+        .param('maxFileInputs', 'Search tasks by maximum number of file inputs.', required=False,
```
brianhelba: I'd call this `Filter tasks by...`.

plugins/item_tasks/plugin_tests/tasks_test.py
```diff
@@ -126,6 +126,68 @@ def testAddItemTasksToFolderFromJson(self):
...
+        createTask('item1', 'Task 1')
+        createTask('item2', '1 File Task')
+        createTask('item3', '2 File Task')
+        createTask('item4', '3 File Task')
+        testMinMax(1, min=1, max=1)
+        testMinMax(4, max=3)
+        testMinMax(3, max=2)
+        testMinMax(2, min=2, max=3)
+        testMinMax(4)
+        testMinMax(1, min=3)
```
brianhelba: How about just adding a test where zero tasks are returned? e.g. `min=7`

plugins/item_tasks/plugin_tests/tasks_test.py
```diff
@@ -126,6 +126,76 @@ def testAddItemTasksToFolderFromJson(self):
...
+        createTask('item1', 'Task 1')
+        createTask('item2', '1 File Task')
+        createTask('item3', '2 File Task')
+        createTask('item4', '3 File Task')
+        testMinMax(1, min=1, max=1)
+        testMinMax(4, min=0, max=3)
+        testMinMax(3, min=0, max=2)
+        testMinMax(2, min=2, max=3)
+        testMinMax(4)
+        testMinMax(1, min=3)
```
brianhelba: One case should specify `max` but not `min`.

plugins/item_tasks/server/rest.py
```diff
@@ -34,17 +34,38 @@ def __init__(self):
     @autoDescribeRoute(
         Description('List all available tasks that can be executed.')
         .pagingParams(defaultSort='name')
+        .param('minFileInputs', 'Filter tasks by minimum number of file inputs.', required=False,
+               dataType='int',)
```
brianhelba: Trailing comma here.

brianhelba:
This now has merge conflicts. Please merge `origin/master` into this branch and push the result.

Issue #1302, opened 2016-04-08
girder.utility.server.configureServer is not idempotent


brianhelba:
It is possible to do:

```
from girder.utility.server import configureServer
configureServer()
```

to get an instance of Girder running without actually binding to any TCP ports. This is very useful for doing internal batch processing or scripting operations directly with the Girder models and database connection.

However, the `configureServer` function is not idempotent. If it is run a second time from the same Python context, it will fail.

PR #2134, opened 2017-06-23
Capture stdout/stderr to logs only when starting the server.

### LINKED ISSUE #2131: Redefining sys.stdout breaks readline
When spawning a repl using the `code` module, using arrow keys for history outputs control characters due to [this line](https://github.com/girder/girder/blob/656e421c8b11f1318ac7198403616e587c0563e7/girder/__init__.py#L206) in our logging setup.

To reproduce from a python shell:
```
import girder, code
code.interact()
# try using up/down arrows
```

Leave out `import girder` to see the arrow keys functioning correctly.

manthey:
Otherwise, importing girder by itself affects being able to use REPL.

Fixes #2131.

### DIFF
--- a/girder/__init__.py
+++ b/girder/__init__.py
@@ -108,7 +108,7 @@ class StreamToLogger:
     def write(self, buf):
         if not self.logger._girderLogHandlerOutput:
             self.logger._girderLogHandlerOutput = True
-            _originalStdOut.write(buf)
+            self.stream.write(buf)
             for line in buf.rstrip().splitlines():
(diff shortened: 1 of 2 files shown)


brianhelba:
Does this fix or relate to https://github.com/girder/girder/issues/1302 ?

danlamanna:
@brianhelba No, I think https://github.com/girder/girder/commit/ac7084e321e513324a89a37e038ba8e46c6815f7 does though I haven't stress tested it.

PR #2135, opened 2017-06-23
Static analysis improvements

brianhelba:
* Rename `.eslintrc` to `.eslintrc.json`
  * The old file name is deprecated by ESLint, in favor of one that expresses the file format.
  * This doesn't delete the old files yet, in case other plugins are still using them.
* Clean up `.eslintignore`
* Add static analysis tests to `girder-swagger.js`
* Add static analysis tests to the web-external client
* Add static analysis tests to the jQuery client
* Upgrade eslint-config-girder


brianhelba:
Filed https://github.com/girder/girder/issues/2136 , to remind us to delete the old files

.eslintrc
```diff
@@ -1,3 +1,3 @@
 {
-    "extends": "girder"
+  "extends": "girder"
```
manthey: Why are you changing the spacing on this?  Our js files we use 4 spaces; shouldn't we do the same on json files for consistency?

brianhelba: You just had to ask. :grimacing: 

So actually, I've been working on adding an EditorConfig file, to programmatically define standards on this. The bottom line is that a number of other the `.json` files in Girder are using 2 spaces:
* https://github.com/girder/girder/blob/master/package.json
* All those in https://github.com/girder/girder/tree/master/plugins/item_tasks/plugin_test

The main exception seems to be older test artifacts and `plugin.json` files. The test artifacts aren't developed on, and plugin definitions seem to be moving towards mostly YAML.

Thus, I picked 2 spaces as the standard for JSON. I agree that 4 would be more consistent with our JS and Python files (though not CMake, Pug, or Stylus).

I really don't care as long as it ends up consistant, and I'd be just as happy to go with 4, as JSON is super-easy to automatically reformat via tools.

PR #2137, opened 2017-06-23
Remove the unmaintained web-external client

brianhelba:
* Remove the unmaintained web-external client
* Update the External Web Clients docs for Girder 2


brianhelba:
@girder/developers PTAL.

PR #2138, opened 2017-06-24
Terms of Use plugin

brianhelba:
This adds a new "Terms of Use" plugin. From the documentation:
> This plugin allows collection admins to define a set of textual "Terms of Use", which other users must accept before browsing within the collection. The terms may be set with markdown-formatted text, and users will be required to re-accept the terms whenever the content changes. Logged-in users have their acceptances stored and remembered permanently, while anonymous users have their acceptances stored only on the local browser.

create
prompt
info

Remaining TODO:
* [ ] Finish client-side tests, increasing coverage

### DIFF
--- a/clients/web/src/templates/widgets/editCollectionWidget.pug
+++ b/clients/web/src/templates/widgets/editCollectionWidget.pug
@@ -4,10 +4,10 @@
       .modal-header
         button.close(data-dismiss="modal", aria-hidden="true", type="button") &times;
         h4.modal-title
-          if (collection)
-            | Edit collection
-          else
+          if (create)
             | Create collection
+          else
+            | Edit collection
       .modal-body
         .form-group
           label.control-label(for="g-name") Name
@@ -19,9 +19,9 @@
       .modal-footer
         a.btn.btn-small.btn-default(data-dismiss="modal") Cancel
         button.g-save-collection.btn.btn-small.btn-primary(type="submit")
-          if (collection)
-            i.icon-edit
-            | Save
-          else
+          if (create)
             i.icon-plus-squared
             | Create
+          else
+            i.icon-edit
+            | Save
--- a/clients/web/src/views/widgets/EditCollectionWidget.js
+++ b/clients/web/src/views/widgets/EditCollectionWidget.js
@@ -13,30 +13,37 @@ import 'girder/utilities/jquery/girderModal';
 /**
  * This widget is used to create a new collection or edit an existing one.
  */
-var EditCollectionWidget = View.extend({
+const EditCollectionWidget = View.extend({
     events: {
         'submit #g-collection-edit-form': function (e) {
             e.preventDefault();
 
-            var fields = {
+            const fields = {
                 name: this.$('#g-name').val(),
                 description: this.descriptionEditor.val()
             };
 
-            if (this.model) {
-                this.updateCollection(fields);
-            } else {
-                this.createCollection(fields);
-            }
-
             this.descriptionEditor.saveText();
             this.$('button.g-save-collection').girderEnable(false);
             this.$('.g-validation-failed-message').text('');
+
+            this._saveCollection(fields)
+                .always(() => {
+                    this.$('button.g-save-collection').girderEnable(true);
+                })
+                .done(() => {
+                    this.$el.modal('hide');
+                })
+                .fail((err) => {
+                    this.$('.g-validation-failed-message').text(err.responseJSON.message);
+                    this.$('#g-' + err.responseJSON.field).focus();
+                });
         }
     },
 
     initialize: function (settings) {
-        this.model = settings.model || null;
+        this.create = !settings.model;
+        this.model = settings.model || new CollectionModel();
         this.descriptionEditor = new MarkdownWidget({
             text: this.model ? this.model.get('description') : '',
             prefix: 'collection-description',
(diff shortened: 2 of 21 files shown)


clients/web/src/views/widgets/EditCollectionWidget.js
```diff
@@ -13,30 +13,37 @@ import 'girder/utilities/jquery/girderModal';
...
  * This widget is used to create a new collection or edit an existing one.
  */
-var EditCollectionWidget = View.extend({
+const EditCollectionWidget = View.extend({
     events: {
         'submit #g-collection-edit-form': function (e) {
             e.preventDefault();
 
-            var fields = {
+            let fields = {
```
zachmullen: should be `const`

clients/web/src/templates/widgets/editCollectionWidget.pug
```diff
@@ -4,10 +4,10 @@
       .modal-header
         button.close(data-dismiss="modal", aria-hidden="true", type="button") &times;
         h4.modal-title
-          if (collection)
-            | Edit collection
-          else
+          if (create)
```
zachmullen: To preserve compatibility of this template, let's just change the condition from `if collection` to `if collection && collection.id`.

clients/web/src/views/widgets/EditCollectionWidget.js
```diff
@@ -47,62 +54,49 @@ var EditCollectionWidget = View.extend({
...
-                view.create = false;
-            } else {
-                view.create = true;
-            }
-        });
+        this.$el.html(EditCollectionWidgetTemplate({
+            create: this.create,
+            collection: this.model
+        }));
+        let modal = this.$el
```
zachmullen: should be `const`

girder/api/rest.py
```diff
@@ -840,89 +840,90 @@ def handleRoute(self, method, path, params):
...
+            wildcards = {}
+            for routeComponent, pathComponent in six.moves.zip(route, path):
+                if routeComponent[0] == ':':  # Wildcard token
+                    wildcards[routeComponent[1:]] = pathComponent
+                elif routeComponent != pathComponent:  # Exact match token
+                    break
+            else:
+                return route, handler, wildcards
+
+        raise RestException('No matching route for "%s %s"' % (method.upper(), '/'.join(path)))
```
zachmullen: Are the changes to this file related / necessary to the ToU plugin? If not, could you cherry-pick them onto a separate topic and remove this commit from this one?

brianhelba: Did that last week: https://github.com/girder/girder/pull/2210

clients/web/src/views/widgets/EditCollectionWidget.js
brianhelba: Will fix.

clients/web/src/views/widgets/EditCollectionWidget.js
brianhelba: Will fix.

clients/web/src/templates/widgets/editCollectionWidget.pug
brianhelba: I'm not clear what you mean by compatibility. This is an implementation detail of a template that's already tightly bound to being used by `EditCollectionWidget`.

FYI, the `EditCollectionWidget.create` instance variable existed before, it was just not being injected into the template.

zachmullen: The template, prior to this change, could be used without passing the `create` variable, but it now requires it, so if someone else was using it, it would break for them. It's probably dubious that anybody was using it, but since there's an easy equivalent non-breaking change, we should go with that.

zachmullen:
Could you rebase this? @brianhelba

brianhelba:
I still need to make some changes here. Moving back to WIP for the very short term.

clients/web/src/templates/widgets/editCollectionWidget.pug
brianhelba: I've thought about this more, and I'm concerned that if anyone wants to subclass the `EditCollectionWidget`, the `create` instance variable might not remain strictly equivalent to `collection && collection.id`. Given that `EditCollectionWidget` exposes a public instance variable, `create`, the widget implementation should respect the variable's state as the determinant of whether the widget really is in "create" mode; alternatively, we should not have `create` as an instance variable at all, and always use the (IMO, non-obvious) test `collection && collection.id` to determine the mode of the widget.

I'd call this change a bugfix (to start respecting the `create` variable that always existed) of a part of the code that's not part of the API surface (the injected template variables), so I don't think compatibility is strictly affected. I'd like to keep it as-is.

brianhelba:
@girder/developers This is ready for review.

zachmullen:
@brianhelba the new test is failing in CI.

plugins/terms/server/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+def acceptCollectionTerms(self, collection, termsHash):
+    if not collection.get('terms'):
+        raise RestException('This collection currently has no terms.')
+
+    # termsHash should be encoded to a bytes object, but storing bytes into MongoDB behaves
+    # differently in Python 2 vs 3. Additionally, serializing a bytes to JSON behaves differently
+    # in Python 2 vs 3. So, just keep it as a unicode (or ordinary Python 2 str).
+    realTermsHash = base64.b64encode(
+        hashlib.sha256(collection['terms'].encode('utf-8')).digest()
+    ).decode('utf-8')
```
zachmullen: Would it be ok to just use the hex digest instead, since most people are used to seeing that as the text form of checksums?

plugins/terms/web_client/models/CollectionModel.js
```diff
@@ -0,0 +1,70 @@
...
+    // An empty string also means there are no terms.
+    return Boolean(this.get('terms'));
+};
+
+CollectionModel.prototype.currentUserHasAcceptedTerms = function () {
+    const termsHash = this._hashTerms();
+    const currentUser = getCurrentUser();
+    if (currentUser) {
+        const userAcceptedTerms = currentUser.get('terms');
+        // Lodash's _.get would be nice here
```
zachmullen: ES is really missing a `?.` operator.

brianhelba:
@zachmullen Updated.

zachmullen:
Did you figure out why the test was failing before? Was it a race condition or a logic problem?

brianhelba:
I accidentally included a development commit in this branch, so I just did some surgery to remove it. Since I was already rewriting history, I rebased this branch on top of the latest master.

Practically speaking, nothing in the diff should have changed (with the exception of my bad commit being gone).

@zachmullen I need your reapproval please.

brianhelba:
The test was failing before because the change in https://github.com/girder/girder/pull/2138/commits/4b887fda407be306a0ae0382c374cf5e849f9afb was not comprehensive enough, and failed to update a place where the accepted terms data structure was being built locally. I fixed this issue in https://github.com/girder/girder/pull/2138/commits/ef9e7f7e4790d1c1c6427137ad99ecade4a24a0b .

brianhelba:
The upshot is that the tests work, as they caught a defect that https://github.com/girder/girder/commit/4b887fda407be306a0ae0382c374cf5e849f9afb introduced.

mgrauer:
Way to go @brianhelba ! and thanks @zachmullen for your QA endurance!

🏃

plugins/terms/web_client/models/CollectionModel.js
```diff
@@ -0,0 +1,75 @@
...
+                // instantiating a new UserModel, and it would be too fragile to reproduce those
+                // here. We also don't want to trigger a brand-new login. So, just update the
+                // currentUser's "terms" attribute in-place, triggering a "change" event.
+                const userAcceptedTerms = currentUser.get('terms') || {};
+                // This code would be much cleaner with _.merge from Lodash.
+                userAcceptedTerms.collection = userAcceptedTerms.collection || {};
+                userAcceptedTerms.collection[this.id] = {
+                    hash: termsHash,
+                    // 'accepted' is from a server-set timestamp, so we don't know it here. However,
+                    // its value is irreverent, as it's for auditing purposes only.
```
msmolens: Is "irreverent" the right word?

brianhelba: No, but it doesn't care!!!

PR #2139, opened 2017-06-26
WIP: Add pytest compatibility

danlamanna:
Since pytest is capable of running unittest test cases (see https://docs.pytest.org/en/latest/unittest.html), this PR adds drop-in support for using pytest.

This means existing unittest tests can use regular asserts, and we can fully use pytest+fixtures for tests that don't require a CherryPy app (this is demonstrated by rewriting the token test module). As part of 3.0, we should visit converting our base test class into pytest-style fixtures and general test utility functions.

This is WIP since I still need to verify things are working 100%, also it's currently outputting coverage tables for every single test which should be changed to match the current behavior on master.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ __pycache__
 .eslintcache
 /npm-debug.log
 **/*.retry
+/.cache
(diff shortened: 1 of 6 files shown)


jbeezley:
I don't see coverage output for the plugins with this branch: https://2313-11854950-gh.circle-artifacts.com/0/tmp/circle-artifacts.sQEGmmp/coverage/python/index.html

zachmullen:
Status: not testing plugins at the moment, but otherwise almost ready.

danlamanna:
Superseded by #2412.

Issue #2129, opened 2017-06-22
girder-client functions should operate in cli and non cli contexts


danlamanna:
Based on the discussion @manthey and I had in #2085, we should ensure that methods added to girder-client should be callable from a non-cli context (girder client as a library). This is especially important in the context of remote jobs.

PR #2140, opened 2017-06-26
Allow uploadFileToFolder to take a file-like object

danlamanna:
Prior to this, `uploadFileToFolder` could only be called with files that already exist
on disk since it took a local filepath. This is
similar to how `uploadFileToItem` operates, however we should steer away
from this as we want the girder-cli library to be usable from remote
worker instances.

`uploadFileToItem` has been in a girder-client release and as such,
can't be fixed to conform to its folder counterpart without breaking
backwards compatibility.

This partially addresses #2129.

@manthey You are likely the only one using this at the moment, are you okay with breaking the function signature like this given it hasn't been in a release? (CLI usage remains unchanged).

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -883,6 +883,45 @@ class GirderClient(object):
         with open(filepath, 'rb') as f:
             return self._uploadContents(obj, f, filesize, progressCallback=progressCallback)
 
+    def uploadStreamToFolder(self, folderId, stream, filename, size, reference=None, mimeType=None,
+                             progressCallback=None):
+        """
+        Uploads a file-like object to a folder, creating a new item in the process. If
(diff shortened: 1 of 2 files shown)


manthey:
I'm fine with changing the function signature; I am only using it through the CLI.

manthey:
An alternate approach would have been to create two new functions: `uploadStreamToItem` and `uploadStreamToFolder`, where the existing methods would open a file, get its size, and pass the parameters to the new functions.  This would allow support for file-like objects in both cases, not break functionality, and still have the convenience of uploading from a path if that is desired.

danlamanna:
@manthey I selfishly implemented your suggested approach for folders and not items, I'm sure I'll need that soon enough but I think this PR can go in without it (unless you have strong feelings about it).

PR #2145, opened 2017-06-26
Add an EditorConfig file, so common editors can pick up indentation rules

brianhelba:
The [EditorConfig specification](http://editorconfig.org/) is recognized by a large number of editors, with plugins for nearly all other editors.

* Fix spelling mistakes in comments and docstrings
* Don't reference developer names in source code
* Fix a misspelled variable name
  * This appears to be unused anywhere in the source code.
* Remove trailing whitespace and add a final newline to source files
* Indent all shell / batch scripts with 4 spaces
* Indent Makefile with tabs
* Fix an indentation mistake in `taskRun.styl`
* Indent all developer-generated JSON files with 4 spaces
* Indent all CMake files with 2 spaces
* Indent all YAML files with 2 spaces
* Add an EditorConfig file, so common editors can pick up indentation rules


devops/ansible/roles/girder/library/test/run.sh
```diff
@@ -1,5 +1,5 @@
 #/bin/bash
 
 ansible-playbook -i .vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory \
-                 --private-key=.vagrant/machines/girder/virtualbox/private_key \
-                 -u vagrant $*
```
manthey: What's the reasoning behind using tabs in shell scripts?  And, why mix tabs and spaces?  This will make it so this is hard to read unless you're editor has tabstops at 4 spaces, which isn't a common setting.

I'm against tabs as white space except where absolutely required (Makefiles).

girder/utility/path.py
```diff
@@ -44,7 +44,7 @@ def encode(token):
 
 
 def decode(token):
-    """Unescape special characters in a token from a path representation.
+    """Un-escape special characters in a token from a path representation.
```
danlamanna: Do we not have a rule for docstrings? Almost all of Girder conforms to putting the subject on its own line, after the triple quotes.

brianhelba:
@danlamanna I'd really like to have an automated test for this. However, I tried both [eclint](https://github.com/jedmao/eclint) and [editorconfig-tools](https://www.npmjs.com/package/editorconfig-tools), and both had far too many false positives to be useful (e.g. complaining that a totally-empty `__init__.py` file had no ending newline).

I'll file a bug to eventually find and add automated linting of EditorConfig settings.

girder/utility/path.py
brianhelba: Agreed. I hope there's a Flake8 extension for this, which we could eventually add. For now, it's outside the scope of this PR.

devops/ansible/roles/girder/library/test/run.sh
brianhelba: I went with tabs only because lots of `.bat` / `.sh` already had them. Upon your request, I've switched the rules to require 4 spaces for those too, but kept tabs for `Makefile`, just because it's technically required.

danlamanna:
LGTM, though there were some changes my editor picked up on that weren't included here. I put all of them (except one which fixed the indentation in our sourced jasmine lib) in a branch so you can decide what makes sense to include, see https://github.com/girder/girder/compare/editorconfig...editorconfig-improvements.

brianhelba:
@danlamanna I included your commit and fixed a merge conflict. PTAL and approve again.

PR #2146, opened 2017-06-27
Use local fonts

manthey:
Download Google fonts and then use them locally.

### DIFF
--- a/grunt_tasks/webpack.config.js
+++ b/grunt_tasks/webpack.config.js
@@ -23,6 +23,7 @@ var path = require('path');
 
 var webpack = require('webpack');
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
+var GoogleFontsPlugin = require('google-fonts-webpack-plugin');
 
 var paths = require('./webpack.paths.js');
 var es2015Preset = require.resolve('babel-preset-es2015');
@@ -69,7 +70,14 @@ module.exports = {
             'window.jQuery': 'jquery'
         }),
         // Disable writing the output file if a build error occurs
-        new webpack.NoEmitOnErrorsPlugin()
+        new webpack.NoEmitOnErrorsPlugin(),
+        new GoogleFontsPlugin({
+            filename: 'googlefonts.css',
+            fonts: [{
+                family: 'Droid Sans',
+                variants: ['regular', '700']
+            }]
+        })
     ],
     module: {
         rules: [
--- a/grunt_tasks/webpack.config.js
+++ b/grunt_tasks/webpack.config.js
@@ -23,7 +23,6 @@ var path = require('path');
 
 var webpack = require('webpack');
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
-var GoogleFontsPlugin = require('google-fonts-webpack-plugin');
 
 var paths = require('./webpack.paths.js');
 var es2015Preset = require.resolve('babel-preset-es2015');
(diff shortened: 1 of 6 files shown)


grunt_tasks/webpack.config.js
```diff
@@ -69,7 +70,14 @@ module.exports = {
             'window.jQuery': 'jquery'
         }),
         // Disable writing the output file if a build error occurs
-        new webpack.NoEmitOnErrorsPlugin()
+        new webpack.NoEmitOnErrorsPlugin(),
+        new GoogleFontsPlugin({
```
zachmullen: How does this play with extractCSS plugin?

manthey: I know nothing about extractCSS, so I don't know.

zachmullen: Does the CSS file end up in the .js bundle, or the .css bundle?

manthey: Much like fontello, there is a passing reference to it in `clients/web/static/built/girder_lib.min.css`, but the main css file gets loaded by itself (neither is bundled), which is why it is explicitly referenced in the webroot.mako file.

I do see that the GoogleFontsPlugin is creating a css file for the main and per plugin, which is wasteful.

manthey: And, that passing reference is just the css defined in `clients/web/src/stylesheets/layout/global.styl`, not triggered by these changes.

jbeezley: I think it would be possible to import the google fonts css inside `global.styl` and remove the reference in the mako template.  As far as creating the file for each plugin, you could add the plugin for core only in [build.js](https://github.com/girder/girder/blob/master/grunt_tasks/build.js#L101).

manthey: I'm not sure how to refer to the built css file from `global.styl`.  I assume there is some `@import` stylus statement, but the path to reach the `googlefonts.css` file in `static/built` is eluding me.

package-lock.json
```diff
@@ -459,7 +459,7 @@
     "babylon": {
       "version": "6.17.4",
       "resolved": "https://registry.npmjs.org/babylon/-/babylon-6.17.4.tgz",
-      "integrity": "sha512-kChlV+0SXkjE0vUn9OZ7pBMWRFd8uq3mZe8x1K6jhuNcAFAtEnjchFAqB+dYEXKyd+JpT6eppRR78QAr5gTsUw=="
```
brianhelba: I'm not sure why your npm is touching and rewriting all these `integrity` fields. Which version of npm are you using?

manthey: npm 5.0.3

grunt_tasks/webpack.config.js
jbeezley: I'm not sure either.  We can probably defer that to a different PR.

manthey:
Ever since the package-lock.json went in, changing it fails on Circle unless we rebuild without cache.  See this conversation: https://discuss.circleci.com/t/destroy-npm-cache-node-modules-based-on-a-hash-of-package-json-or-if-present-npm-shrinkwrap-json/2985

jbeezley:
I'm not sure what's going on there.  @brianhelba any ideas?

brianhelba:
I may have a fix. I'll push it shortly.

manthey:
I'll note the issue with circle and npm are not isolated to this PR; any change to `package.json` / `package-lock.json` would have it.

brianhelba:
For reference, the fix is https://github.com/girder/girder/pull/2149

jeffbaumes:
@manthey thanks for picking up on what I had once attempted in this abandoned PR https://github.com/girder/girder/pull/798

PR #2149, opened 2017-06-27
Fix the caching of npm packages on CircleCI

brianhelba:
* Document the fact that npm@>=5 must be used to update `package-lock.json`
* Fix the caching of npm packages on CircleCI
* Regenerate the `package-lock.json` file, upgrading recursive dependencies

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,9 +17,12 @@ machine:
 dependencies:
   cache_directories:
     - "~/cache"
+    - "~/.npm"
   override:
     - sudo apt-get -q update; sudo apt-get install -q -y libgif-dev libsasl2-dev
     - ./scripts/install_cmake.sh
+    # The real npm cache is in ~/.npm, but CircleCI tries to cache node_modules, breaking upgrades
+    - rm -rf ./node_modules
     - npm install -g npm
     - npm config set progress false
     - set -o pipefail; pyenv exec pip install -U pip setuptools | cat
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -403,7 +403,13 @@ Web client libraries in Girder core are managed via `npm <https://www.npmjs.com/
 When a new npm package is required, or an existing package is upgraded, the following
 should be done:
 
-1. Update ``dependencies`` or ``devDependencies`` in ``package.json`` to add a new
+1. Ensure that version >=5 of npm is installed in your development environment:
+
+   .. code-block:: bash
+
+       npm install -g 'npm@>=5'
+
+2. Update ``dependencies`` or ``devDependencies`` in ``package.json`` to add a new
    *abstract* specifier for the package:
   * Packages that are bundled into the web client should generally use the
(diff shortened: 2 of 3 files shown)


brianhelba:
The final commit, which upgrades Webpack, is just for testing CircleCI. I'll remove it if this works.

docs/development.rst
```diff
@@ -403,21 +403,27 @@ Web client libraries in Girder core are managed via `npm <https://www.npmjs.com/
 When a new npm package is required, or an existing package is upgraded, the following
 should be done:
 
-1. Update ``dependencies`` or ``devDependencies`` in ``package.json`` to add a new
+1. Ensure that version >=5 of npm is installed in your development environment:
+
+   .. code-block:: bash
+
+       npm install -g  'npm@>=5'
```
manthey: Nit: extra space between `-g` and `'npm@>=5'`.

manthey:
This looks fine to me, but I'd feel better if the dummy commit added a package to both `package.json` and `package-lock.json` (unless you are completely certain that a change in version is sufficient to trigger the failure condition).

brianhelba:
Trying it with a new package now.

package.json
```diff
@@ -29,6 +29,7 @@
     "event-source": "0.1.1",
     "extract-text-webpack-plugin": "^2.1.0",
     "file-loader": "^0.10.0",
+    "geojs": "^0.12.2",
```
jbeezley: Why is geojs here?

circle.yml
```diff
@@ -17,9 +17,12 @@ machine:
 dependencies:
   cache_directories:
     - "~/cache"
+    - "~/.npm"
   override:
     - sudo apt-get -q update; sudo apt-get install -q -y libgif-dev libsasl2-dev
     - ./scripts/install_cmake.sh
+    # The real npm cache is in ~/.npm, but CircleCI tries to cache node_modules, breaking upgrades
+    - rm -rf ./node_modules
```
jbeezley: It's weird that this is necessary.  Wouldn't this also imply that deleting node_modules is needed for developers as wel

package.json
brianhelba: Just testing a new package.

brianhelba: This will be removed before merging.

circle.yml
brianhelba: One would think, but my development environment is always fine when I switch onto a branch with a different `package.json` / `package-lock.json` and rebuild in-place.

All this will go away in CircleCI 2, when we have much more direct control over what is cached.

brianhelba:
For the record, this just passed with a dummy commit, which upgraded one package and added another.

I'm dropping that commit now, to prepare for a real merge.

brianhelba:
Also, this may theoretically speed up builds on master, as npm packages that are required by plugins will now be properly cached. Previously, it's likely that those were having to be re-downloaded from npm, as they were removed by Girder core's invocation of `npm install` before being re-installed by the more explicit `npm install <packages>` Grunt step for plugins.

PR #2150, opened 2017-06-27
Cleanup and document client test startup procedure

brianhelba:
* Make girderTest.startApp chain onto girderTest.promise more robustly
* Use girderTest.startApp to start the app in testing
* Don't wrap test specs in "$(function..."
  * The test runner will only load and execute specs once the page is ready.
* Add additional documentation on plugin tests

### DIFF
--- a/plugins/item_tasks/plugin_tests/widgetsSpec.js
+++ b/plugins/item_tasks/plugin_tests/widgetsSpec.js
@@ -1,398 +1,670 @@
 girderTest.importPlugin('jobs', 'worker', 'item_tasks');
 // This is a unit test, so do not start the app
 
-
+var itemTasks;
 girderTest.promise.done(function () {
-    var itemTasks = girder.plugins.item_tasks;
+    itemTasks = girder.plugins.item_tasks;
+});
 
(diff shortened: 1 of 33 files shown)


brianhelba:
@girder/developers PTAL.

plugins/item_tasks/plugin_tests/widgetsSpec.js
```diff
@@ -1,777 +1,777 @@
 girderTest.importPlugin('jobs', 'worker', 'item_tasks');
 // This is a unit test, so do not start the app
 
-
+var itemTasks;
 girderTest.promise.done(function () {
-    var itemTasks = girder.plugins.item_tasks;
+    itemTasks = girder.plugins.item_tasks;
```
brianhelba: FYI, this is the only place in this commit that's not simply stripping a `$(function` and de-indenting. We need to preserve the previous behavior of waiting until the plugin is fully loaded (which is definitely true once `girderTest.promise` resolves) before being able to look it up.

The actual specs `describe` / `it` can of course still be defined right away, since they're not executed until the whole page is ready (which the the point of this whole commit).

PR #2151, opened 2017-06-27
Fix documentation typo for Beanstalk instructions

danlamanna:
@mgrauer


mgrauer:
This works, but I noticed when I turned on a plugin, the text saying restart the server showed up in addition to the text saying you can't restart the server.

My plugin got enabled this time after an EB application restart.


screen shot 2017-06-27 at 6 48 13 pm

danlamanna:
I addressed this issue in #2152

PR #2152, opened 2017-06-28
Stop showing rebuild/restart suggestion when cherrypy_server is false

danlamanna:


jeffbaumes:
+1 to not having the button appear at all in non-cherrypy mode (no note necessary either IMO).

danlamanna:
I agree with @brianhelba and @jeffbaumes, I'm going to open up a separate PR addressing everyone's concerns.

PR #2154, opened 2017-06-28
Allow read() on file-like objects.

manthey:
We expose Girder files as python file-like objects.  For the default
python file, the size of a read is optional.  We were requiring it,
which broke sending the file-like object to certain functions
(json.load, for instance).

### DIFF
--- a/girder/utility/abstract_assetstore_adapter.py
+++ b/girder/utility/abstract_assetstore_adapter.py
@@ -21,7 +21,7 @@ import six
 
 from girder.api.rest import setResponseHeader
 from girder.constants import SettingKey
-from girder.models.model_base import ValidationException
+from girder.models.model_base import GirderException, ValidationException
 from girder.utility import progress, RequestBodyStream
 from .model_importer import ModelImporter
 
@@ -47,6 +47,9 @@ class FileHandle(object):
         self._file = file
         self._adapter = adapter
         self._pos = None
+        # If a read is requested that is longer than the specified size, raise
+        # an exception.  This prevents unbounded memory use.
+        self._maximumReadSize = 16 * 1024 * 1024
 
         self.seek(0)
 
@@ -70,6 +73,8 @@ class FileHandle(object):
         """
         if size is None or size < 0:
             size = self._file['size'] - self._pos
+        if size > self._maximumReadSize:
+            raise GirderException('Read exceeds maximum allowed size.')
         data = six.BytesIO()
         length = 0
         for chunk in itertools.chain(self._prev, self._stream):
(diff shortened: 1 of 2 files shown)


manthey:
Python 3 says: ["If the argument is omitted, None, or negative, data is read and returned until EOF is reached."](https://docs.python.org/3/library/io.html?highlight=bytesio#io.BufferedIOBase).  The use of `None` isn't actually mentioned in Python 2 (only that omitting the size and a negative value read until EOF).

manthey:
Regardless of the amount of data asked for, `read` is allowed to return less than the specified number of bytes if in non-blocking mode.  We could modify the `FileHandle.read` method to emit no more than some maximum at a time so that if a process asks for a very large file it wouldn't crush the server's memory.  This would be as simple as adding a `size = min(size, MAXIMUM_FILE_READ_SIZE)` to the code.  That should be a different PR, though.

zachmullen:
I'd be in favor of such a maximum size enforcement.

manthey:
Any opinion on a sensible maximum size?

zachmullen:
16MB is the max size of documents in mongo, so that's probably a sensible value.

manthey:
Interestingly, in python's `json.load` method, it assumes that the file-like object is blocking (in both Python 2 and Python 3).  A quick grep of Python's packages shows that many methods assume file-like objects are blocking.  Based on this, always acting a non-blocking mode by having a maximum read size will probably cause more troubles than the danger of using too much memory.

zachmullen:
I always assume a read call is blocking, so that's fair. That's the reason I didn't allow sizeless reads in the original API. In what context are you trying to use them?

manthey:
I had code that looked like `data = json.loads(''.join(ModelImporter.model('file').download(file)()))`, and I thought it would be better as `data = json.load(ModelImporter.model('file').open(file))`, but that didn't work with Girder master because `json.load` does `fp.read()`.  I was hoping on a huge file that switching this would use less memory.

manthey:
I'd like it to work "in general", since in core Python, all file-like objects support `read` without parameters and it was surprising when that didn't work.  In current master, asking for a large size is dangerous as it stands.

I see several possibilities:
  - Functionally act like non-blocking I/O over some size (16Mb by default).  This would preserve the current behavior in *most* instances, allow the parameter-less `read`, and guard against memory issues with the existing behavior.  For small files and reasonable size requests, it would act as expected.
  - Throw an exception when asking for a parameter-less read and a large size.  This is odd in that we still allow a large specified size to eat up excessive amounts of memory.
  - Throw an exception when asking for any read over a large size.  This is more consistent than the previous  option, but is a change.

zachmullen:
I'm fine with either 2 or 3.

PR #2158, opened 2017-06-29
Move babel coverage config to .babelrc

jbeezley:
I was able to track down the problems with our javascript coverage reporting to the webpack helper in the dicom_viewer plugin.  For some reason, adding the babel loader there without including the coverage config breaks it's use elsewhere.  It probably shouldn't be doing that because the include/exclude rules are correct.  Maybe there is some caching going on.

I think we were originally using a `.babelrc` file, which we removed to fix problems with symlinking plugins into place.  This will probably break that use case again, but I can't think of a good way around it.  If anyone has any ideas, let me know.

See the difference: [before](https://2272-11854950-gh.circle-artifacts.com/2/tmp/circle-artifacts.UjI6cEz/coverage/js/plugins/item_tasks/web_client/models/WidgetModel.js.html) and [after](https://2289-11854950-gh.circle-artifacts.com/2/tmp/circle-artifacts.1grdGeC/coverage/js/plugins/item_tasks/web_client/models/WidgetModel.js.html)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,10 @@
+{
+  "env": {
+    "cover": {
+      "plugins": [[
+        "babel-plugin-istanbul",
+        {"include": ["**/clients/web/src/**/*.js", "**/plugins/*/web_client/**/*.js"]}
+      ]],
+    }
+  }
+}
--- a/grunt_tasks/webpack.config.js
+++ b/grunt_tasks/webpack.config.js
@@ -36,22 +36,6 @@ function fileLoader() {
     };
 }
 
-function _coverageConfig() {
-    try {
-        var istanbulPlugin = require.resolve('babel-plugin-istanbul');
-        return {
-            plugins: [[
-                istanbulPlugin, {
-                    exclude: ['**/*.pug', '**/*.jade', 'node_modules/**/*']
-                }
-            ]]
-        };
-    } catch (e) {
-        // We won't have the istanbul plugin installed in a prod env.
-        return {};
-    }
-}
-
 var loaderPaths = [path.resolve('clients', 'web', 'src')];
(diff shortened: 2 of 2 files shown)


zachmullen:
> This will probably break that use case again

It will break symlinking plugins altogether, or just break coverage reporting of them?

jbeezley:
I had forgotten the details of the problem.  I just tried symlinking all of the builtin plugins from outside the girder tree, and it worked fine building and running with and without coverage; however, the coverage **reporting** doesn't work for plugins that are installed this way.  I believe this is the only specific use case affected by this change.

manthey:
This might affect some external plugin's coverage reporting in their build process, since some of them use symlinking.

jbeezley:
Yes it will.  The other way to fix it is to add the coverage configuration [here](https://github.com/girder/girder/blob/master/plugins/dicom_viewer/webpack.helper.js#L19), but that seems... non-intuitive.

zachmullen:
Do we have any idea why/how the dicom_viewer loader spec is affecting the global babel loader spec? Shouldn't they be independent?

jbeezley:
They **should** be independent.  It's likely a bug somewhere in istanbul, babel, webpack, etc.  I tried to debug it a big and didn't get too far.

zachmullen:
I wonder if [this code](https://github.com/girder/girder/blob/master/grunt_tasks/plugin.js#L475-L484), particularly the grunt config merging logic, is doing weird stuff to our loader list.

jbeezley:
I see the issue, we are overriding the "include" option somewhere on the loaders so babel is being run twice.  First without coverage and then with.  So the coverage is reporting the transpiled code.  I'll have an alternative fix shortly.

PR #2160, opened 2017-06-29
Make the babel-loader test more explicit in dicom_viewer

jbeezley:
We append all the plugin paths to the `include` array.  Perhaps we should not do that, but in any case this prevents the babel-loader from running twice on other plugins' sources.

### DIFF
--- a/plugins/dicom_viewer/webpack.helper.js
+++ b/plugins/dicom_viewer/webpack.helper.js
@@ -1,7 +1,7 @@
 module.exports = function (config) {
     config.module.rules.push({
         resource: {
-            test: /\.glsl$/,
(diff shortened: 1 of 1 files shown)


zachmullen:
> We append all the plugin paths to the include array.

Is that just a result of grunt's wild config merging behavior?

jbeezley:
Nope, it's this code here: https://github.com/girder/girder/blob/master/grunt_tasks/plugin.js#L456-L469

zachmullen:
Gross. And of course I wrote it.

This change feels like a workaround, the real fix would be to somehow make that block work better, no?

jbeezley:
Probably, I would leave it for a different PR though.  There's really no harm in this change.  It's just a more specific regex.

Issue #2075, opened 2017-06-07
Issues with turning off EventStream


manthey:
Since PR #1687, we turn off the EventStream connection when a browser tab loses focus.  When you return to the tab, stale status can be displayed because recent status updates were missed.  Also, some plugins expect to get messages and miss them.

As an example, delete or move a group of items such that the action will take a while.  Change to a different tab.  Wait until the action completes plus the expiration time of the progress message.  Navigate back to the tab, and the status of the delete or move is still displayed and won't be cleared.  Refreshing the page fixes the issue.

Some job plugin widgets expect to observe status changes through the event stream, but if they occur when the stream is not being watched and more than 30 seconds before switching back to the tab, the widget misses them and is in an incorrect state.

Either we need to save a queue of events to replay, or clients need to know when a stream was disconnected and therefore could have missed information so that they can reload the appropriate information when they regain focus.

matthewma7:
I forget the reason we would like to close the connection. Is it the cost on the client or on the server?

Based on my experience with WebSocket, I simply had web connection open all the time but when there is a disconnection, all component would get a disconnect event and re-initialize its data. 
This pattern was easy to implement and worked great for me. (Probably attribute to socket.io implementation, it has heartbeat request etc.) Even on phones, switch out the browser app won't stop the connection, all messages are queued somewhere and will be fed to the app when user switch back to the browser app. (Not sure about the behavior if the socket.io fallback to the long poll)

zachmullen:
> I forget the reason we would like to close the connection. Is it the cost on the client or on the server?

It was actually because we had users opening many tabs to girder, and chrome counting the event sockets against a global max concurrent connections to a single domain. Not sure if that's still an issue even, but then we figured there simply wasn't a point receiving real time events on a backgrounded tab, though that could change in the future.

matthewma7:
Is this a time to consider switching to WebSocket? It has IE support and doesn't have the connection count limitation.

manthey:
We want to solve this for a stopped connection regardless of the underlying technology -- you can get the same behavior by putting a laptop to sleep and then waking it up again (or switching networks and reconnecting to a vpn, etc.).

zachmullen:
Right, websocket vs. eventsource doesn't matter here -- the main difference between those two technologies is that websockets are duplex and eventsource is one-way. Since we only need one-way, eventsource is the easier path.

matthewma7:
[...]

PR #2163, opened 2017-06-30
Emit events when the event stream starts and stops.

manthey:
The job list widget updates when the stream starts.  This keeps the list
accurate when the page has been hidden and reshown.

If a job is not in a completed state, job details are refreshed when the
stream starts.  This adds a `finished` method to the JobStatus object to
check if the job is in a finished state.

The progress list is cleared when the stream stops.  This prevents stale
progress from lingering.


manthey:
This is only a work-in-progress because it has no tests.  I'm not sure of the best way to add tests; we don't currently test the event stream heartbeat.  Suggestions welcome.

This addresses issue #2075.

Because a network connection might be intermittent, any solution to monitoring events must know how to recover after an interruption.  The heartbeat produces such an interruption, which actually makes a convenient manual test.

plugins/jobs/web_client/views/JobDetailsWidget.js
```diff
@@ -36,6 +36,14 @@ var JobDetailsWidget = View.extend({
             }
         });
 
+        this.listenTo(eventStream, 'g:eventStream.start', function (event) {
+            var status = this.job.get('status');
+            if (!JobStatus.finished(status)) {
+                this.job.once('g:fetched', this.render, this);
+                this.job.fetch();
```
matthewma7: I think since the fetch() will return a jQuery promise now, we could use .fetch().done() now.

manthey:
This is ready for review.

I don't understand codecov's complaint on the patch diff, as locally this has 100% coverage of the diff.

docs/plugin-development.rst
```diff
@@ -804,6 +804,27 @@ This demonstrates one simple use case for client plugins, but using these same
 techniques, you should be able to do almost anything to change the core
 application as you need.
 
+JavaScript events
+*****************
+
+The JavaScript client handles events from both the server and client-specific
+code.  The server events originate on the server and can be monitored by both
+the server's python code and the client's JavaScript code.  The client events
```
brianhelba: `python` -> `Python`

docs/plugin-development.rst
```diff
@@ -804,6 +804,27 @@ This demonstrates one simple use case for client plugins, but using these same
 techniques, you should be able to do almost anything to change the core
 application as you need.
 
+JavaScript events
+*****************
+
+The JavaScript client handles events from both the server and client-specific
+code.  The server events originate on the server and can be monitored by both
```
brianhelba: This is not quite accurate: server *events* are internal to the server-side code, whereas server *notifications* are made available to client.

docs/plugin-development.rst
brianhelba: Are we talking about client Backbone events? DOM events? Or just `EventStream` events? This should be more specific.

docs/plugin-development.rst
```diff
@@ -804,6 +804,27 @@ This demonstrates one simple use case for client plugins, but using these same
...
+
+If the connection to the server is interrupted, the client will not receive
+server events.  Periodically, the client will attempt to reconnect to the
+server to resume handling events.  Similarly, if client's browser tab is placed
+in the background for a long enough period of time, the connection that listens
+for server events will be stopped to prevent excessive resource use.  When the
+browser's tab regains focus, the client will once again receive server events.
+
+When the connection to the server's event stream is interrupted, a
+``g:eventStream.stop`` event is triggered.  When the server is once more
```
brianhelba: We should clarify that ``g:eventStream.stop`` is a Backbone event, triggered on the `EventStream` object.

clients/web/src/views/layout/ProgressListView.js
```diff
@@ -17,6 +17,9 @@ var ProgressListView = View.extend({
     initialize: function (settings) {
         this.eventStream = settings.eventStream;
         this.eventStream.on('g:event.progress', this._handleProgress, this);
+        // if the event stream disconnects, clear the progress so we don't have
+        // stale values lingering.
+        this.eventStream.on('g:eventStream.stop', this._clearProgress, this);
```
brianhelba: Use `this.listenTo` to prevent memory leaks, in case the ProgressListView gets destroyed before the `EventStream` object.

clients/web/src/views/layout/ProgressListView.js
```diff
@@ -51,6 +54,15 @@ var ProgressListView = View.extend({
         this._onUpdate();
     },
 
+    _clearProgress: function () {
+        _.each(Object.keys(this._map), (key) => {
```
brianhelba: This could be simplified (and use clearer variable names) as 
```
_.each(this._map, (progressId, progressWidget) => {
    progressWidget.destroy();
    ...
```

plugins/jobs/web_client/views/JobDetailsWidget.js
```diff
@@ -36,6 +36,13 @@ var JobDetailsWidget = View.extend({
             }
         });
 
+        this.listenTo(eventStream, 'g:eventStream.start', function (event) {
+            var status = this.job.get('status');
+            if (!JobStatus.finished(status)) {
+                this.job.fetch().done(() => { this.render(); });
```
brianhelba: This is a personal nit, but I think these sorts of constructs are more readable when formatted as:
```
this.job.fetch()
    .done(() => { this.render(); });
```
If you disagree, then please disregard.

plugins/jobs/web_client/JobStatus.js
```diff
@@ -79,19 +84,22 @@ JobStatus.registerStatus({
         value: 3,
         text: 'Success',
         icon: 'icon-ok',
-        color: '#53b653'
+        color: '#53b653',
+        finished: true
```
brianhelba: Rather than rely on the other statuses to have `undefined` -> `false` values of `finished`, I think we should explicitly define `finished: false` on them.

We can then more cleanly decide if we want custom / unknown status types to have a default `finished` value of `true` or `false`.

clients/web/src/views/layout/ProgressListView.js
```diff
@@ -51,6 +54,15 @@ var ProgressListView = View.extend({
...
+            this._map[key].destroy();
+            this._map[key].remove();
+            delete this._map[key];
+            this._onUpdate();
```
brianhelba: `this._onUpdate();` really only needs to be called once after the loop.

clients/web/src/views/layout/ProgressListView.js
```diff
@@ -51,6 +54,15 @@ var ProgressListView = View.extend({
...
+        });
+    },
+
     _onUpdate: function () {
         if (_.isEmpty(this._map)) {
```
brianhelba: This is outside the scope of your changes, but the whole `_onUpdate` function body could be simplified as:
```
this.$el.toggle( !_.isEmpty(this._map) );
```

docs/plugin-development.rst
manthey: Thanks.  I had realized this was inaccurate and could be clearer, but then forgot to rework it.

clients/web/src/views/layout/ProgressListView.js
manthey: I'll update it.

plugins/jobs/web_client/views/JobDetailsWidget.js
```diff
@@ -36,6 +36,13 @@ var JobDetailsWidget = View.extend({
             }
         });
 
+        this.listenTo(eventStream, 'g:eventStream.start', function (event) {
+            var status = this.job.get('status');
```
brianhelba: Use `let` or `const` here too.

clients/web/test/spec/utilitiesSpec.js
```diff
@@ -0,0 +1,44 @@
...
+        }, 'event stream heartbest to stop');
+        runs(function () {
+            expect(stop).toBeGreaterThan(0);
+            es.settings._heartbeatTimeout = 5000;
+        }, 'restart heartbeat');
+        waitsFor(function () {
+            return start;
+        }, 'restart to occur');
+        runs(function () {
+            expect(start).toBeGreaterThan(0);
```
manthey: The counters are reinitialized on line 20.  Since we are using a short timeout to trigger the stop, it is possible the restart will occur before we reset the timeout to `5000` (since the `runs` command actually executes in the next timeslice in jasmine then the last `waits` command), so we don't want to reset the start counter again.

manthey:
@brianhelba Thanks for the review -- I hope I've addressed all of this issues at this point.

PR #2164, opened 2017-06-30
Create thumbnail on upload

zachmullen:


zachmullen:
Fixed merge conflicts -- @girder/developers could someone take a pass at this?

plugins/thumbnails/plugin_tests/thumbnail_test.py
```diff
@@ -68,26 +71,25 @@ def setUp(self):
             else:
                 self.privateFolder = folder
 
-        events.unbind('thumbnails.create', 'test')
```
brianhelba: What was this doing before, and why was it removed?

brianhelba:
In general, what's the expected use case for this functionality? I think auto-creation on upload is a useful feature, but having to explicitly add the `reference` to every upload is burdensome. I see two ways to make this more accessible going forward:
* Add a system setting, to enable thumbnail creation on all uploads, with some pre-set width / height / crop options
* Extend the file upload widget, and add an option to request a thumbnail be created on the new file.

Obviously, these two cases are not mutually exclusive, and they affect different types of users. What are you thoughts?

plugins/thumbnails/plugin_tests/thumbnail_test.py
zachmullen: It wasn't removed, see line 77.

plugins/thumbnails/server/__init__.py
```diff
@@ -53,13 +54,46 @@ def removeThumbnailLink(event):
...
+
+    if isinstance(ref, dict) and isinstance(ref.get('thumbnail'), dict):
+        width = max(0, ref['thumbnail'].get('width', 0))
+        height = max(0, ref['thumbnail'].get('height', 0))
+
+        if not width and not height:
+            return
+
+        file = event.info['file']
+        item = ModelImporter.model('item').load(file['itemId'], force=True)
```
zachmullen: I believe everything that goes through the `data.process` does have an item, but I'll protect against that case to be more future-proof.

brianhelba: Not necessarily: https://github.com/girder/girder/blob/master/girder/models/upload.py#L204

plugins/thumbnails/plugin_tests/thumbnail_test.py
brianhelba: Oh nevermind, GitHub was displaying the diff in a weird way.

zachmullen:
I'm fine with more features in the future to support the cases you mentioned, but this is sufficient for what I wanted for now. Namely because I can pass `reference` via girder_client, including via girder-cli.

zachmullen:
(My actual use case was to support this through the worker, which this now does.)

plugins/thumbnails/server/__init__.py
```diff
@@ -53,13 +54,46 @@ def removeThumbnailLink(event):
...
+        utils.scheduleThumbnailJob(
+            file=file, attachToType='item', attachToId=item['_id'], user=event.info['currentUser'],
+            width=width, height=height, crop=ref['thumbnail'].get('crop', True))
```
zachmullen: I chose to extract the pythonic truthiness value of the "crop" field, which is not the same as validation but is probably good enough.

plugins/thumbnails/server/__init__.py
```diff
@@ -53,13 +54,50 @@ def removeThumbnailLink(event):
...
+        height = max(0, ref['thumbnail'].get('height', 0))
+
+        if not width and not height:
+            return
+        if not isinstance(width, int) or not isinstance(height, int):
+            return
+
+        file = event.info['file']
+        if 'itemId' in file:
+            item = ModelImporter.model('item').load(file['itemId'], force=True)
```
brianhelba: This is a minor nit, but this code could be de-indented by two levels if you made the validators on line 77 and 87 of the form:
```
if not x:
   return
```

plugins/thumbnails/server/__init__.py
```diff
@@ -53,13 +54,50 @@ def removeThumbnailLink(event):
...
+            crop = bool(ref['thumbnail'].get('crop', True))
```
brianhelba: This is fine.

If it weren't JSON, we should use [`toBool`](https://github.com/girder/girder/blob/master/girder/utility/__init__.py#L107), but for a JSON parameter, we can be less tolerant.

plugins/thumbnails/server/__init__.py
```diff
@@ -53,13 +54,50 @@ def removeThumbnailLink(event):
...
+    if isinstance(ref, dict) and isinstance(ref.get('thumbnail'), dict):
+        width = max(0, ref['thumbnail'].get('width', 0))
+        height = max(0, ref['thumbnail'].get('height', 0))
+
+        if not width and not height:
+            return
+        if not isinstance(width, int) or not isinstance(height, int):
+            return
+
+        file = event.info['file']
```
brianhelba: This is entirely theoretical, but I'd add the check of whether `'itemId'` is in the `file` to the top of this method. If it fails, we can skip the time spent parsing attached JSON, which could take relatively longer.

PR #2165, opened 2017-06-30
Add an optionalArgumentDecorator utility, fixing @boundHandler

brianhelba:
* Remove unnecessary arguments from use of the `@boundHandler` decorator
* Update the `rest_decorator_test` to use more sophisticated decorators
* Add an `optionalArgumentDecorator` utility
* Use the `optionalArgumentDecorator` where appropriate

With only the first two commits in this PR, usage of `@boundhandler` with other decorators and the `rest_decorator_test` will fail with:
```
ERROR: Failed to load plugin "test_plugin":
  File "girder/utility/plugin_utilities.py", line 103, in loadPlugins
    root, appconf, apiRoot = loadPlugin(plugin, root, appconf, apiRoot)
  File "girder/utility/plugin_utilities.py", line 228, in loadPlugin
    module = imp.load_module(moduleName, fp, pathname, description)
  File "tests/test_plugins/test_plugin/server.py", line 31, in <module>
    @describeRoute(None)
  File "girder/api/access.py", line 74, in dec
    @six.wraps(fun)
  File "/home/vagrant/.virtualenvs/girder/local/lib/python2.7/site-packages/six.py", line 792, in wrapper
    f = functools.wraps(wrapped, assigned, updated)(f)
  File "/usr/lib/python2.7/functools.py", line 33, in update_wrapper
    setattr(wrapper, attr, getattr(wrapped, attr))
AttributeError: 'boundHandler' object has no attribute '__name__'
```
due to the buggy implementation of the `@boundhandler` decorator as a class.

This PR corrects the issue more broadly, adding a new `optionalArgumentDecorator` utility, and simplifying the logic of decorators that can now use it.

### DIFF
--- a/tests/test_plugins/test_plugin/server.py
+++ b/tests/test_plugins/test_plugin/server.py
@@ -22,10 +22,11 @@ from girder.api import access
 from girder.api.describe import Description, describeRoute
 from girder.api.rest import boundHandler, rawResponse, Resource, setResponseHeader
 from girder.api.v1.collection import Collection
+from girder.constants import TokenScope
 from girder.utility.server import staticFile
 
 
-@access.public
+@access.user(scope=TokenScope.ANONYMOUS_SESSION)
 @boundHandler
 @describeRoute(None)
 def unboundHandlerDefaultNoArgs(self, params):
@@ -33,7 +34,7 @@ def unboundHandlerDefaultNoArgs(self, params):
     return not self.boolParam('val', params)
 
 
-@access.public
+@access.user(scope=TokenScope.ANONYMOUS_SESSION)
 @boundHandler()
 @describeRoute(None)
 def unboundHandlerDefault(self, params):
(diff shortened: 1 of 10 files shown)


brianhelba:
@girder/developers PTAL.

tests/test_plugins/test_plugin/server.py
```diff
@@ -22,31 +22,33 @@
 from girder.api.describe import Description, describeRoute
 from girder.api.rest import boundHandler, rawResponse, Resource, setResponseHeader
 from girder.api.v1.collection import Collection
+from girder.constants import TokenScope
 from girder.utility.server import staticFile
 
 
-@access.public
+@access.user(scope=TokenScope.ANONYMOUS_SESSION)
```
manthey: I don't follow the rationale of changing all of the public access points in this test (in the context of the bound handler decorator change) to require anonymous session scope.

brianhelba: The point of changing these tests is to present a more challenging test case, which fails before the subsequent commits in this PR are added.

Previously, when applied with the argument-calling syntax, `boundHandler` was not exposing some special properties (e.g. `__name__`) on the new function. When the `access.user` decorator subsequently re-wraps the `boundHandler`-wrapped function, the  `six.wraps` utility cannot find these special properties and fails.

The `access.public` decorator only attaches some additional properties to the target function, rather than re-wrapping it, so it does not trigger this sort of failure.

Finally, the addition of `TokenScope.ANONYMOUS_SESSION` is to ensure that the logic inside the applied decorators is actually being called.

PR #2166, opened 2017-07-03
Improve test database setup config

jbeezley:
This contains two changes to make the `setup_database` spec more robust:
1. `creator` is now optional for toplevel collections.  If the creator is not specified, then it will use the first user returned by a database lookup.
2. Support YAML aliases for using a common config object in multiple places in the spec.

### DIFF
--- a/tests/cases/setup_database_test.yml
+++ b/tests/cases/setup_database_test.yml
@@ -4,12 +4,12 @@ _common: &common
   folders:
     - name: 'Common'
       items:
-      - name: 'Item 1'
-      - name: 'Item 2'
+        - name: 'Item 1'
+        - name: 'Item 2'
       folders:
-      - name: 'Common subfolder'
-        items:
-        - name: 'Item 3'
+        - name: 'Common subfolder'
+          items:
+            - name: 'Item 3'
 
 users:
   # a typical admin user
--- a/tests/cases/setup_database_test.yml
+++ b/tests/cases/setup_database_test.yml
@@ -1,4 +1,16 @@
 ---
+# create a list of common folders that can be inserted in multiple places
+_common: &common
+  folders:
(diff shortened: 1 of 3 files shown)


tests/setup_database.py
```diff
@@ -98,17 +94,21 @@ def addCreator(spec, parent=None):
...
             # (users don't have creators)
             spec['creator'] = parent
-        else:
+        elif parent.get('creatorId'):
             # otherwise, use the creator of the parent
-            spec['creator'] = userIds[parent['creatorId']]
+            spec['creator'] = userModel.load(parent.get('creatorId'), force=True)
+    else:
+        # if all else fails, just use any user for the creator
+        spec['creator'] = userModel.findOne({}, force=True)
```
danlamanna: So is there anything stopping someone from having an item with no owner? What happens if this query returns nothing?

tests/cases/setup_database_test.yml
```diff
@@ -1,4 +1,16 @@
 ---
+# create a list of common folders that can be inserted in multiple places
+_common: &common
+  folders:
+    - name: 'Common'
+      items:
+      - name: 'Item 1'
```
danlamanna: Indentation

tests/setup_database.py
jbeezley: `creator` will be `None` when passed to the model creation method.  This works, but `creatorId` is `None` on the document.  This import scripts creates all users before creating anything else, so the only way this could happen is if there aren't any users defined at all.

danlamanna: Yeah, I guess that's what I'm worried about. I don't think Girder should have to support items/folders/collections having no creator unless that's something it already does. If it doesn't then I think this should raise an exception if there are no users found.

PR #2168, opened 2017-07-05
Expose the instantiated girder application as girder.app

jbeezley:
This is mostly useful for testing where we want to be able to access the view tree (`girder.bodyView`) inside the test without monkey patching.  It is also nice to have for interactive debugging of a running instance.

### DIFF
--- a/clients/web/test/testUtils.js
+++ b/clients/web/test/testUtils.js
@@ -1247,7 +1247,6 @@ $(function () {
  * application object.
  */
 girderTest.startApp = function () {
-    var app;
     girderTest.promise = girderTest.promise
         .then(function () {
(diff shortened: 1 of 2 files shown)


clients/web/test/testUtils.js
```diff
@@ -1278,10 +1278,11 @@ girderTest.startApp = function () {
                 parentView: null,
                 start: false
             });
+            girder.app = app;
             return app.start();
         })
         .then(function () {
-            girder.events.trigger('g:appload.after');
+            girder.events.trigger('g:appload.after', app);
```
brianhelba: It'd be simpler to just reference `girder.app` here, then remove the `var app` from the parent scope entirely.

PR #2169, opened 2017-07-06
Ensure that token scopes on an API key are valid, registered ones

zachmullen:


girder/models/api_key.py
```diff
@@ -50,8 +50,14 @@ def validate(self, doc):
...
                 raise ValidationException('Scope must be a list, or None.')
             if not doc['scope']:
-                raise ValidationException(
-                    'Custom scope list must not be empty.')
+                raise ValidationException('Custom scope list must not be empty.')
+
+            # Ensure only registered scopes are being set
+            admin = self.model('user').load(doc['userId'], force=True)['admin']
+            scopes = TokenScope.scopeIds(admin)
+            unknown = set(doc['scope']) - scopes
```
dorukozturk: @zachmullen Can we have a more descriptive name for unknown? Maybe unknownScope?

PR #2171, opened 2017-07-07
Add reuseExisting keyword to createLinkFile

Xarthisius:
This PR allows link files to be "reused" like all the other girder models.

### DIFF
--- a/girder/models/file.py
+++ b/girder/models/file.py
@@ -164,7 +164,8 @@ class File(acl_mixin.AccessControlMixin, Model):
 
         return doc
 
-    def createLinkFile(self, name, parent, parentType, url, creator, size=None, mimeType=None):
+    def createLinkFile(self, name, parent, parentType, url, creator, size=None,
+                       mimeType=None, reuseExisting=False):
         """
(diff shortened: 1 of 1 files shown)


girder/models/file.py
```diff
@@ -182,14 +183,26 @@ def createLinkFile(self, name, parent, parentType, url, creator, size=None, mime
...
         elif parentType == 'item':
             item = parent
 
+        if reuseExisting:
+            existing = self.findOne({
+                'itemId': item['_id'],
+                'name': name
+            })
+            if existing:
+                return existing
```
zachmullen: Should we also modify the link URL if it's changed?

Xarthisius: I'm not sure. How would it work for regular files? Would you update the content?

zachmullen: For a regular file, creation of the file document and updating its contents are independent, so there isn't a parallel, but I would imagine that if we had a `reuseExisting` flag on the upload process, it would update the contents and use the existing file document. In this case, since we have the url already, I think it makes the most sense to update it since the user is requesting that to be the URL.

PR #2172, opened 2017-07-09
Access list widget failed to render new entries

zachmullen:
1. Select a folder, open its access control dialog
2. Type the name of a user or group that does not have access, click a search result.

Expected behavior is that the user or group is appended to the list, but actually nothing appeared and an error message occurred:

```
girder_lib.min.js:4797 Uncaught TypeError: /Users/zach/dev/girder/clients/web/src/templates/widgets/accessEntry.pug:19
    17|         a.g-action-manage-flags(title="Access flags")
    18|           i.g-flag-icon.icon-flag
  > 19|             .g-flag-count-indicator(class=(entry.flags && entry.flags.length ? null : "hide"))= entry.flags.length
    20|         .g-flags-popover-container.hide(resourcetype=type, resourceid=entry.id, title="Access flags")
    21|           each flag, key in flagList
    22|             - var checked = _.contains(entry.flags, key) ? 'checked' : null;

Cannot read property 'length' of undefined
    at girder_lib.min.js:71403
    at template (girder_lib.min.js:71460)
    at child.<anonymous> (girder_lib.min.js:24152)
    at triggerEvents (girder_lib.min.js:18942)
    at triggerApi (girder_lib.min.js:18930)
    at eventsApi (girder_lib.min.js:18729)
    at child.Events.trigger (girder_lib.min.js:18920)
    at Object.<anonymous> (girder_lib.min.js:21135)
    at fire (girder_lib.min.js:8223)
    at Object.fireWith [as resolveWith] (girder_lib.min.js:8353)
```
This bug was introduced in 48a7ca8e448c554dddb9ea4f790c86588204deac

### DIFF
--- a/clients/web/src/templates/widgets/accessEntry.pug
+++ b/clients/web/src/templates/widgets/accessEntry.pug
@@ -16,7 +16,7 @@ div(class=`g-${type}-access-entry`, resourceid=entry.id)
       if !noAccessFlag && _.keys(flagList).length
         a.g-action-manage-flags(title="Access flags")
           i.g-flag-icon.icon-flag
-            .g-flag-count-indicator(class=(entry.flags && entry.flags.length ? null : "hide"))= entry.flags.length
+            .g-flag-count-indicator(class=(entry.flags && entry.flags.length ? null : "hide"))= (entry.flags || []).length
         .g-flags-popover-container.hide(resourcetype=type, resourceid=entry.id, title="Access flags")
           each flag, key in flagList
             - var checked = _.contains(entry.flags, key) ? 'checked' : null;


brianhelba:
@zachmullen The current code server code [always creates user and group access list entries with a `'flags': []` field](https://github.com/girder/girder/blob/master/girder/models/model_base.py#L978).

The problem here would seem to stem from old databases that were saved before 'flags' were added. To ensure that the API returns consistent results in the future, I think that [`getFullAccessList`](https://github.com/girder/girder/blob/master/girder/models/model_base.py#L1053) should be modified to add a default `'flags': []` field to all fetched entries that don't already have them. Alternatively, this field could be patched and saved back to the database as part of the `load` / `validate` lifestyle for relevant models.

This would also allow the relevant line in `accessEntry.pug` to be significantly simplified.

zachmullen:
@brianhelba this isn't happening on the stuff returned from the server, it's when we add a *new* entry to the list in the client, prior to saving it to the server. That said, you're right that we should probably fix this elsewhere, namely [here](https://github.com/girder/girder/blob/master/clients/web/src/views/widgets/AccessWidget.js#L293-L297) and [here](https://github.com/girder/girder/blob/master/clients/web/src/views/widgets/AccessWidget.js#L260-L265). I can't get to that fix right away, but can make it once I get back.

PR #2175, opened 2017-07-11
Don't symlink node_modules for packaging tests

jbeezley:
Using a symlink causes the following error:

```
npm ERR! /home/ubuntu/girder/node_modules/webpack/node_modules/fsevents
  is not a child of /home/ubuntu/build/env/lib/python3.4/site-packages/girder
```

This error is thrown by a postinstall script of some dependency.  We might get away with copying the directory instead, but with the new caching in npm 5, there isn't much of a speed up.  I think this will fix the recent build failures on master.


zachmullen:
Going to merge to try this out

jbeezley:
Now we're getting a different failure:
```
Error: Grunt targets were not built correctly
```
I'll try to debug it again.

PR #2179, opened 2017-07-12
Allow the HierarchyWidget to self-render after instantiation

### LINKED ISSUE #2178: web_client_routing test occasionally fails
The `web_client_routing` is occasionally timing out and failing.

For an example of this failure, see:
* https://circleci.com/gh/girder/girder/2371#tests/containers/2
* http://my.cdash.org/testDetails.php?test=32995108&build=1251490

The issue seems to be that routing at the start of a new test spec is occurring before a modal in a previous test spec has finished being rendered then being dismissed.

brianhelba:
The `HierarchyWidget` already self-renders after being instantiated, so there's no need for the creator to call the `.render()` method externally.

Additionally, in all places where a `HierarchyWidget` was being created, it was being instantiated within the same synchronous sequence as the rendering of its container template. Accordingly, there is minimal benefit to first instantiating a `HierarchyWidget` without an element, rendering the container template, then calling `setElement` to attach the `HierarchyWidget` within the container template. Instead, container templates are rendered, then the `HierarchyWidget` is instantiated with its element preset, which is significantly simpler.

This fixes #2178 , as the `HierarchyWidget` is no longer being rendered twice, which is particularly helpful in cases where the `HierarchyWidget` also displays a modal (and has to animate its show / hide behavior twice).

### DIFF
--- a/clients/web/src/views/body/FolderView.js
+++ b/clients/web/src/views/body/FolderView.js
@@ -19,7 +19,12 @@ var FolderView = View.extend({
         this.folderEdit = settings.folderEdit || false;
         this.itemCreate = settings.itemCreate || false;
 
+        this.render();
+    },
+
+    render: function () {
         this.hierarchyWidget = new HierarchyWidget({
+            el: this.$el,
             parentModel: this.folder,
             upload: this.upload,
             folderAccess: this.folderAccess,
@@ -29,11 +34,6 @@ var FolderView = View.extend({
             parentView: this
         });
 
-        this.render();
-    },
-
-    render: function () {
-        this.hierarchyWidget.setElement(this.$el).render();
         return this;
     }
 }, {
--- a/clients/web/src/views/body/CollectionView.js
+++ b/clients/web/src/views/body/CollectionView.js
@@ -69,15 +69,12 @@ var CollectionView = View.extend({
                 this.folder.set({
                     _id: settings.folderId
                 }).on('g:fetched', function () {
-                    this._createHierarchyWidget();
                     this.render();
                 }, this).on('g:error', function () {
                     this.folder = null;
-                    this._createHierarchyWidget();
                     this.render();
                 }, this).fetch();
             } else {
-                this._createHierarchyWidget();
                 this.render();
             }
         } else if (settings.id) {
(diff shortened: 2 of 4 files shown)


brianhelba:
This is still a WIP, as I need to do a little more manual testing.

brianhelba:
I'm pretty sure that I can also address #2181 with some additional work here. I'm going to try now.

brianhelba:
@girder/developers PTAL. This now fixes #2178 , but does not yet address #2181 (which will require separate changes).

clients/web/src/views/body/FolderView.js
```diff
@@ -19,7 +19,12 @@ var FolderView = View.extend({
         this.folderEdit = settings.folderEdit || false;
         this.itemCreate = settings.itemCreate || false;
 
+        this.render();
+    },
+
+    render: function () {
         this.hierarchyWidget = new HierarchyWidget({
```
zachmullen: Usually we try to avoid creating the child widget on each render call of the parent widget. Why this change?

brianhelba: In the case of `FolderView` here, it doesn't really matter at all. Since it's a body view, it's expected to self-render, so we could move the instantiation of `HierarchyWidget` up to the end of `FolderView.initialize` and remove the `FolderView.render` method entirely.

I put it here only for consistency with the `CollectionView` and `UserView`, where the location matters more, as described in this PR's description.

Honestly, I don't yet have a totally clean vision that I'd like to advocate for how and when child widgets get instantiated and rendered, but this is an improvement in efficiency over the old code, and fixes a broken test.

clients/web/src/views/body/FolderView.js
```diff
@@ -19,7 +19,12 @@ var FolderView = View.extend({
...
+            el: this.$el,
```
zachmullen: Heh, this is wonky. I did this long long ago out of pure laziness. I guess it persists to this day.

brianhelba: In theory, I see nothing wrong with it. To me, it should just naturally force us to consider whether we need the parent view, or whether the child view could be more directly instantiated via the router. In many cases, there is supplemental state, logic, or sub-sub-views that need to be added by the parent class. In this case, `FolderView` could probably be elided entirely, but that's obviously outside the scope of this PR.

PR #2182, opened 2017-07-13
Use Backbone's native mechanism to create DOM elements for Views

brianhelba:
Raw HTML snippets should not appear in Javascript code.


brianhelba:
This is still a WIP, as a few more complicated places also need to be fixed.

brianhelba:
@girder/developers PTAL.

PR #2184, opened 2017-07-14
Fix checking what needs updating in ansible.

manthey:
When we update a collection or other resource, we make a check to see if the values we want to set are already present.  This is done via `set(resource.items())`, which fails if any of the values in the resource dictionary are unhashable, such as the access flags list.  This only checks the values that could possibly be relevant, and, if the comparison still throws a `TypeError`, updates the resource anyway.

### DIFF
--- a/devops/ansible/roles/girder/library/girder.py
+++ b/devops/ansible/roles/girder/library/girder.py
@@ -330,6 +330,10 @@ options:
                     - A list of folder options
                     - Specified by the 'folder' option to the girder module
                     - (see 'folder:')
+            public:
+                required: false
+                description:
+                    - Set to true if the collection is public or false if private
             access:
                 required: false
                 description:
@@ -370,13 +374,17 @@ options:
             parentId:
                 required: true
                 description:
-                    - The ID of the parent collection
+                    - The ID of the parent collection/folder/user
             folders:
                 required: false
                 description:
                     - A list of folder options
                     - Specified by the 'folder' option to the girder module
                     - (see 'folder:')
+            public:
+                required: false
+                description:
+                    - Set to true if the folder is public or false if private
             access:
                 required: false
                 description:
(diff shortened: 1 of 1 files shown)


devops/ansible/roles/girder/library/girder.py
```diff
@@ -942,10 +950,15 @@ def update(self, _id, body, **kwargs):
         if _id in self.resources:
             current = self.resources[_id]
             # if body is a subset of current we don't actually need to update
-            if set(body.items()) <= set(current.items()):
-                return current
-            else:
-                return self.__apply(_id, self.client.put, body, **kwargs)
+            try:
+                if set(body.items()) <= set(
```
zachmullen: Where does this `body` come from? Is it the request body? Just trying to make sure it won't also contain any unhashable types.

manthey: It is the dictionary of values we are setting on the resource.  I don't know why it is called `body`, and I haven't checked that we can't pass non-string values in the ansible yml.  It wouldn't surprise me if we can.

The `try`/`except` will throw in this case and we may needlessly update the resource when it matches the desired state, but only wastes a Mongo call rather than breaks.

manthey: This should be sufficient for the general case.  We could instead do:
```
  if all([k in current and v == current.get(k) for k, v in body.items()]):
    return current
```

There are a few other spots in the library/girder.py file where we use the `set(foo.items())` motif (such as in the `setting` method).  These are probably brittle and should be tested.

Maybe @kotfic or @danlamanna can chime in.

kotfic: @manthey this seems fine to me.  The set approach was based on the assumption that we would be passing simple (hashable) key/value pairs.  I will note that i've run into some edge-cases where ansible does some weird mangling from YAML to Python dicts so your mileage may vary when trying to put/post complex values

PR #2185, opened 2017-07-14
RFC: Item tasks metadata outputs

zachmullen:
Given that I've successfully deployed this as part of the UofU Resonant course, it's probably good to go ahead and open it up to comments here.

In https://github.com/girder/girder_worker/pull/134, we added support to the worker to output metadata to girder in [two different ways](http://girder-worker.readthedocs.io/en/latest/plugins.html#metadata-outputs), but left it as a future effort to support these features in item_tasks. This PR fulfills that capability.

I hacked this together quickly for the course, so the names could use improvement, and perhaps the implementation as well, so please make suggestions here.

TODO

- [ ] Fix provenance links on job details view. They are current broken for the as_metadata output type.


zachmullen:
There is an example task for the course that uses each of these two features. Fixed metadata on output is shown [here](https://github.com/zachmullen/resonant_course/blob/master/instructions/Part4.md), and attaching metadata as output to a dynamic input item is shown [here](https://github.com/zachmullen/resonant_course/blob/master/instructions/Part5.md).

zachmullen:
Status: need someone to approve the general idea/design, once that's done I'll polish and add tests and any necessary docs.

jbeezley:
Is this something that is needed soon?  I think it is okay given the current infrastructure, but I think with where we are heading, this kind of code should moved over to girder-worker-utils.

zachmullen:
How do you envision the server side of item_tasks making use of girder-worker-utils to supersede what's written here?

jbeezley:
I talked with @kotfic just now about this just now.  Maybe girder-worker-utils isn't the correct place for it, but I think the direction we want to head is to have the "side effects" of tasks (uploading files, setting metadata, etc.) handled on the worker rather than by girder.  This also intersects with some of @cjh1's work on a girder storage backend.  It is still an open question (at least in my head) exactly how all of this will work.

Like I said, I think this change makes sense given the current architecture.  The question is do we want to put effort in here while we move away from this approach?  I'm okay with it if the answer is yes.

zachmullen:
> the direction we want to head is to have the "side effects" of tasks (uploading files, setting metadata, etc.) handled on the worker rather than by girder

I guess I don't understand the distinction -- isn't this already the case? The worker uploads files and sets metadata based on this spec.

jbeezley:
I may be misunderstanding the code.  Let me experiment with it some.

zachmullen:
Consider adding testing and documentation

PR #2170, opened 2017-07-06
Add a shortcut link to run an item_task directly on an item

sutartmelson:
This PR aims to a reduce the number of clicks required to run a task on a specific file. 

How it works:

- The user navigates to the item page
- Under the Actions menu, the user selects "Run task on item"
- A modal appears where the user selects a specific item task, then clicks "Configure"
- The web client takes the user to the specified tasks configuration page with the selected item pre-filled in the file input  

The modal that appears when the user selects "Run task on item" brings up a item task pagination widget that only lists tasks that take one file argument. 

optimised

@jeffbaumes  , @brianhelba Take a look and let me know if this is what you had in mind.

PR #2192, opened 2017-07-14
Create paginate task widget

sutartmelson:
This PR puts the pagination of item tasks (currently only done in the taskListView) into a widget view as to allow it to be used elsewhere. Specifically, this will be used in a modal to select a task for #2170. 

Essentially this just generalizes some of what the Tasks body view was doing so we can use it later. 

A filtering functionality will be added to this widget as part of #2170 

@brianhelba PTAL.


plugins/item_tasks/web_client/stylesheets/paginateTasksWidget.styl
```diff
@@ -0,0 +1,12 @@
+.g-task-list-header
+  margin-bottom 10px
+
+  .g-task-pagination
+    ul
+      margin 0px
```
zachmullen: This is more trivia than QA feedback, but when specifying a margin of 0, no unit is required 😄

plugins/item_tasks/web_client/templates/paginateTasksWidget.pug
```diff
@@ -3,6 +3,6 @@
 
 .list-group.g-task-list-container
   each task in tasks
-    a.list-group-item.g-execute-task-link(href=`#item_task/${task.id}/run`)
+    .list-group-item.g-execute-task-link(data-task-id=task.id)
```
zachmullen: In the past, people have stated their preference for using links to navigate, because it makes it easy to open them in a new tab or copy the URL from the browser. One way to achieve both behaviors is to conditionally make it either a link or an event-triggering element based on some variable that is passed into the widget & template.

plugins/item_tasks/web_client/stylesheets/paginateTasksWidget.styl
sutartmelson: Haha. While I still should have caught that, that's a copy and paste from [taskList.styl](https://github.com/girder/girder/blob/master/plugins/item_tasks/web_client/stylesheets/taskList.styl)

plugins/item_tasks/web_client/views/PaginateTasksWidget.js
```diff
@@ -0,0 +1,44 @@
...
+import PaginateWidget from 'girder/views/widgets/PaginateWidget';
+
+import ItemTaskCollection from '../collections/ItemTaskCollection';
+import template from '../templates/paginateTasksWidget.pug';
+import '../stylesheets/paginateTasksWidget.styl';
+
+var PaginateTasksWidget = View.extend({
+    events: {
+        'click .g-execute-task-link': function (event) {
+            let taskId = $(event.currentTarget).data('taskId');
```
zachmullen: Since this variable is never reassigned, `const` would be better than `let`.

plugins/item_tasks/web_client/views/TaskListView.js
```diff
@@ -1,31 +1,17 @@
...
     initialize: function () {
-        this.collection = new ItemTaskCollection();
-
-        this.paginateWidget = new PaginateWidget({
-            collection: this.collection,
+        this.paginateWidget = new PaginateTasksWidget({
+            el: this.$el,
             parentView: this
+        }).once('g:selected', function (params) {
+            let taskId = params.taskId;
```
zachmullen: Same here

plugins/item_tasks/web_client/views/TaskListView.js
```diff
@@ -1,31 +1,17 @@
...
+            router.navigate('item_task/' + taskId + '/run', {trigger: true});
```
zachmullen: This is a good case for using a template string

    `item_task/${taskId}/run`

plugins/item_tasks/web_client/templates/paginateTasksWidget.pug
brianhelba: Just to chime in, I think that having widgets like this only fire `g:selected` events, then letting the parent class hook up routing, provides a very clean abstraction mechanism and allows for reuse of these widgets with minimal dependencies.

Of course, at the end of the day, user experience should take precedence over developer convenience. So, perhaps the `PaginateTasksWidget` could take an optional parameter, which would be a callback function, which if provided, should take a single `task` argument (of an item_task model) and return a string (of a URL to be used as the task link `href`). The `PaginateTasksWidget` should still always continue to emit `g:selected` as before.

plugins/item_tasks/web_client/stylesheets/paginateTasksWidget.styl
```diff
@@ -0,0 +1,12 @@
...
+
+.g-execute-task-link
+    cursor pointer
+    
+.g-execute-task-link:hover
+    background-color #f0f0f0
```
msmolens: Add a newline at the end of the file.

plugins/item_tasks/web_client/templates/paginateTasksWidget.pug
msmolens: For reference or to lift ideas, https://github.com/girder/girder/pull/2004 also refactored the task list into a widget that's used from multiple views and changed the clickable list boxes to hyperlinks.

plugins/item_tasks/web_client/views/TaskListView.js
```diff
@@ -1,31 +1,20 @@
...
-        this.listenTo(this.collection, 'g:changed', () => {
-            this.render();
+        this.paginateWidget = new PaginateTasksWidget({
+            el: this.$el,
+            parentView: this,
+            hyperlinkCallback: function (task) {
+                return `#item_task/${task.id}/run`;
+            }
+        }).once('g:selected', function (params) {
+            const taskId = params.taskId;
```
brianhelba: Use an arrow function here, to keep it simple.

plugins/item_tasks/web_client/views/PaginateTasksWidget.js
```diff
@@ -0,0 +1,51 @@
...
+        this.listenTo(this.collection, 'g:changed', () => {
+            this.render();
+        });
+        this.collection.fetch(this.params);
+    },
+
+    render: function () {
+        this.$el.html(template({
+            tasks: this.collection.toArray(),
+            hyperlinkCallback: this.hyperlinkCallback
```
brianhelba: I don't really like the name of this parameter, but I can't think of anything better off the top of my head.

plugins/item_tasks/web_client/views/PaginateTasksWidget.js
```diff
@@ -0,0 +1,51 @@
...
+                taskId: taskId
+            });
+        }
+    },
+    /**
+     * @param [settings.hyperlinkCallback=null] A callback function, which if provided,
+     * takes a single task argument (of an item_task model) and returns a string
+     * of a URL to be used as the task link href. The g:selected event is triggered
+     * either way with the taskId passed as a parameter.
+     * @param params Any additional parameters to be passed with the fetch request.
```
brianhelba: This docstring needs to be cleaned up more.

plugins/item_tasks/web_client/templates/paginateTasksWidget.pug
```diff
@@ -3,6 +3,7 @@
 
 .list-group.g-task-list-container
   each task in tasks
-    a.list-group-item.g-execute-task-link(href=`#item_task/${task.id}/run`)
+    a.list-group-item.g-execute-task-link(
+      href=hyperlinkCallback && hyperlinkCallback(task), data-task-id=task.id)
```
brianhelba: Maybe put the other attribute on its own line, so it's more obvious.

sutartmelson: I couldn't get it on a new line without a "Invalid attribute separator found" error.

plugins/item_tasks/web_client/views/PaginateTasksWidget.js
sutartmelson: itemUrl? Is it the "callback" you don't like about it?

PR #2195, opened 2017-07-17
Ensure that all enumerators in copyFolder[public] are strings. Fixes #2194

### LINKED ISSUE #2194: Copying folders with public set to boolean fails
If you try to `copyFolder` with public set to True or False:

```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' --header 'Girder-Token: ...' 'http://localhost:8080/api/v1/folder/58dc09765de5256ac13bf222/copy?public=false&progress=false'
```
 you get:

```{
  "message": "TypeError: TypeError('sequence item 0: expected str instance, bool found',)",
  "trace": [
    "<FrameSummary file /home/xarth/codes/girder/girder/api/rest.py, line 567 in endpointDecorator>",
    "<FrameSummary file /home/xarth/codes/girder/girder/api/rest.py, line 1134 in POST>",
    "<FrameSummary file /home/xarth/codes/girder/girder/api/rest.py, line 890 in handleRoute>",
    "<FrameSummary file /home/xarth/codes/girder/girder/api/access.py, line 78 in wrapped>",
    "<FrameSummary file /home/xarth/codes/girder/girder/api/rest.py, line 417 in wrapped>",
    "<FrameSummary file /home/xarth/codes/girder/girder/api/describe.py, line 585 in wrapped>",
    "<FrameSummary file /home/xarth/codes/girder/girder/api/describe.py, line 728 in _validateParam>"
  ],
  "type": "internal"
}```

### PR DESCRIPTION
**Pull request by Xarthisius**
This fixes the issue, but IMO it's mighty confusing that `public` accepts strings, whereas everywhere else it's a boolean parameter.

### DIFF
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -735,8 +735,13 @@ class Folder(AccessControlledModel):
             name = srcFolder['name']
         if description is None:
             description = srcFolder['description']
-        if public == 'original':
-            public = srcFolder.get('public', None)
+        if public is not None and isinstance(public, six.string_types):
+            if public == 'original':
+                public = srcFolder.get('public', None)
+            elif public.lower() == 'true':
+                public = True
+            elif public.lower() == 'false':
+                public = False
         newFolder = self.createFolder(
             parentType=parentType, parent=parent, name=name,
             description=description, public=public, creator=creator,
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -738,10 +738,8 @@ class Folder(AccessControlledModel):
         if public is not None and isinstance(public, six.string_types):
             if public == 'original':
                 public = srcFolder.get('public', None)
-            elif public.lower() == 'true':
-                public = True
-            elif public.lower() == 'false':
-                public = False
+            else:
+                public = self.boolParam('public', params)
         newFolder = self.createFolder(
             parentType=parentType, parent=parent, name=name,
             description=description, public=public, creator=creator,
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -739,7 +739,7 @@ class Folder(AccessControlledModel):
             if public == 'original':
                 public = srcFolder.get('public', None)
             else:
-                public = self.boolParam('public', params)
+                public = public == 'true'
         newFolder = self.createFolder(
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: girder/models/folder.py [hunk 1]
```diff
@@ -735,8 +735,13 @@ def copyFolder(self, srcFolder, parent=None, name=None, description=None,
             name = srcFolder['name']
         if description is None:
             description = srcFolder['description']
-        if public == 'original':
-            public = srcFolder.get('public', None)
+        if public is not None and isinstance(public, six.string_types):
+            if public == 'original':
+                public = srcFolder.get('public', None)
+            elif public.lower() == 'true':
```
zachmullen:
```
else:
    public = self.boolParam('public', params)
```

In any case, the `lower()` is not necessary since `autoDescribeRoute` will validate that a valid value is passed from the enum.

zachmullen:
Thanks for the fix. Since we don't currently support multiple types, we'll have to stick with strings for now. Once you fix my small nit, I'll merge this.

zachmullen:
Note that there is an actual bug here that we should fix. The internal error you are seeing is occurring as a result of trying to report a request error [here](https://github.com/girder/girder/blob/master/girder/api/describe.py#L753).

    ', '.join(descParam['enum'])

Should be changed to

    ', '.join(str(v) for v in descParam['enum']))

Could you go ahead and also make that change on this topic so we can close #2194 ?

zachmullen:
Thanks @Xarthisius , will merge once circle finishes.

girder/models/folder.py
```diff
@@ -735,8 +735,11 @@ def copyFolder(self, srcFolder, parent=None, name=None, description=None,
             name = srcFolder['name']
         if description is None:
             description = srcFolder['description']
-        if public == 'original':
-            public = srcFolder.get('public', None)
+        if public is not None and isinstance(public, six.string_types):
+            if public == 'original':
+                public = srcFolder.get('public', None)
+            else:
+                public = self.boolParam('public', params)
```
zachmullen: Sorry, forgot this was an outdated method since autoDescribeRoute. We can change to

    public = public == 'true'

Issue #2198, opened 2017-07-17
Add additional Stylint rules


brianhelba:
Consider enabling the following additional Stylint rules, once issues with them are resolved:
* `globalDupe`: The file `apidocs.styl` has some duplicate properties; since this file is not built into the main `girder_lib.min.css` file, this is a false positive.
* `sortOrder`: This would require significant refactoring of Girder's code, but would yield a more deterministic code layout.
* `namingConvention`: This should be set to `camelCase`, but Stylus's regular expression for this is bugged: names like `$headerZ` are causing errors.

PR #2199, opened 2017-07-17
Add static analysis tests for Stylus files

### LINKED ISSUE #2173: Add linting for stylus files
We could use stylint.

brianhelba:
* Add static analysis tests for Stylus files
* Always import Nib when loading Stylus files
* Fix Stylus style issues
* Ensure `girder-version.json` ends with a newline

Fixes #2173

### DIFF
--- a/clients/web/src/stylesheets/layout/global.styl
+++ b/clients/web/src/stylesheets/layout/global.styl
@@ -1,17 +1,15 @@
 /**
  * Place global styles in this file.
  */
-@import '~nib/index.styl'
 
-*:not(i)
-  font-family 'Droid Sans',sans-serif
+body
+  // Override the default font from Bootstrap
+  font-family "Droid Sans", sans-serif !important
+  padding 0
+  word-wrap break-word
 
 i
   font-smoothing antialiased
-
-body
-  padding 0
-  word-wrap break-word
 
 a
   cursor pointer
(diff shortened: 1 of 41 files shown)


brianhelba:
@girder/developers PTAL.

brianhelba:
@danlamanna Can you take a look at the [changes to `metadataWidget.styl`](https://github.com/girder/girder/pull/2199/files#diff-00a98ee53528f3c70abf77e8f1aac493) and verify that they're acceptable?

clients/web/src/stylesheets/layout/global.styl
```diff
@@ -1,18 +1,16 @@
 /**
  * Place global styles in this file.
  */
-@import '~nib/index.styl'
-
-*:not(i)
```
brianhelba: FYI, this is one substantive change. Rather than apply `font-family` to `*:not(i)`, we apply it to `body`, and allow it to propagate down. This has the same effect on the document, but is much more efficient.

Interestingly, the `:not(i)` never had any real effect, since the style applied to the parent of any `i` element would simply propagate down to the `i`.

brianhelba:
Also FYI, I filed #2198 for any rules that seem like a good idea, but were too buggy or disruptive to implement here and now.

PR #2200, opened 2017-07-18
Add an ESLint style rule to require single quoted strings

brianhelba:
This was already the de facto standard in Girder, this just enforces the rule and tweaks some very old test files to comply.


brianhelba:
@girder/developers PTAL.

PR #2204, opened 2017-07-19
Adjust display of restart warning message and rebuild button

danlamanna:
The following now holds for enabling and disabling plugins on a
non-restartable server:
- The first time a plugin is enabled or disabled an informational
alert is displayed warning the user that plugins may not take effect
until a restart of Girder
- No additional messages are shown
- The rebuild/restart button is completely hidden

Supersedes #2152, cc @jeffbaumes @brianhelba

### DIFF
--- a/clients/web/src/templates/body/plugins.pug
+++ b/clients/web/src/templates/body/plugins.pug
@@ -1,14 +1,10 @@
 .g-body-title Plugins
-  .g-rebuild-container
+  .g-rebuild-container(class=cherrypyServer ? '' : 'hide')
     button.g-rebuild-and-restart.btn.btn-primary(title="Rebuild web code and restart the server",
         disabled=!cherrypyServer)
       i.icon-arrows-cw
       | Rebuild and restart
     .g-clear
-if !cherrypyServer
-  p
-    | Note that restarting the Girder server is disabled. As a result, plugins enabled or disabled here
-    | might not take effect until the system administrator has restarted Girder.
 .g-plugin-rebuild-restart-text
   | The selected plugins have changed.
   br
--- a/clients/web/src/views/body/PluginsView.js
+++ b/clients/web/src/views/body/PluginsView.js
@@ -127,10 +127,10 @@ var PluginsView = View.extend({
                         this.enabled.splice(idx, 1);
                     }
                 }
-                $('button.g-rebuild-and-restart').addClass('btn-danger');
+                this.$('button.g-rebuild-and-restart').addClass('btn-danger');
 
                 if (this.cherrypyServer) {
-                    $('.g-plugin-rebuild-restart-text').addClass('show');
+                    this.$('.g-plugin-rebuild-restart-text').addClass('show');
                 }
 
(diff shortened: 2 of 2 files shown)


jeffbaumes:
From the code it looks like a refresh or new tab would make the notification appear again, which seems reasonable to me. +1 to no rebuild/restart button at all, and having no other messages.

clients/web/src/views/body/PluginsView.js
```diff
@@ -57,6 +57,7 @@ var PluginsView = View.extend({
     initialize: function (settings) {
         cancelRestRequests('fetch');
         if (settings.all && settings.enabled) {
+            this.cherrypyServer = settings.cherrypyServer || true;
```
zachmullen: Won't this construct make it impossible to set `this.cherrypyServer` to false? I usually use a ternary and `_.has` for cases where the default is true.

clients/web/src/views/body/PluginsView.js
```diff
@@ -127,7 +128,22 @@ var PluginsView = View.extend({
                   }
               }
               $('button.g-rebuild-and-restart').addClass('btn-danger');
-              $('.g-plugin-rebuild-restart-text').addClass('show');
+
+              if (view.cherrypyServer) {
+                  $('.g-plugin-rebuild-restart-text').addClass('show');
+              }
+
+              if (!view.cherrypyServer && !_.has(view, 'displayedCherrypyNotification')) {
```
zachmullen: Probably worthwhile to convert this to an arrow function and ditch the `view` variable.

clients/web/src/views/body/PluginsView.js
danlamanna: Yikes, yes (side note, this code branch is entirely unused by Girder+plugins).

danlamanna:
@zachmullen The issues mentioned should be addressed. Better to use the commits as diffs since I re-indented a block.

clients/web/src/templates/body/plugins.pug
```diff
@@ -1,14 +1,10 @@
 .g-body-title Plugins
-  .g-rebuild-container
+  .g-rebuild-container(class=!cherrypyServer ? 'hide' : '')
```
zachmullen: Unless this `!` operator is intentionally used to cast to boolean, we should invert this condition for readability.

clients/web/src/views/body/PluginsView.js
```diff
@@ -113,23 +115,37 @@ var PluginsView = View.extend({
...
+                var plugin = $(event.currentTarget).attr('key');
+                if (state === true) {
+                    this.enabled.push(plugin);
+                } else {
+                    var idx;
+                    while ((idx = this.enabled.indexOf(plugin)) >= 0) {
+                        this.enabled.splice(idx, 1);
+                    }
+                }
+                $('button.g-rebuild-and-restart').addClass('btn-danger');
```
zachmullen: `this.$`

Issue #1066, opened 2015-11-04
Add a web client test to ensure that default setting are used


brianhelba:
Previously, some widgets on the system configuration page of the web client were not actually respecting the Settings `defaults` sent by the server. [This commit](https://github.com/girder/girder/commit/57a7258700d1fb8a4308adffe31062c3f956724f) fixes the bug for the particular widgets, but a web client test should be added to ensure that this doesn't break again.

manthey:
Outdated

PR #2205, opened 2017-07-19
Add new style rules for Pug templates

brianhelba:
This adds additional style checks to Pug templates, which codify rules that were previously being followed as conventions. This also fixes a number of issues that were found as a result of the new checks.

It's recommend that this be reviewed on a commit-by-commit basis, as each commit encompasses a single atomic change:
* Upgrade `eslint-config-girder`
* Re-indent `.pug-lintrc`
* Add new style rules for Pug templates
* Fix Pug style issues: indentation and id/class ordering
* Simplify logic in `userAccount` template
* Ensure that all HTML IDs are prefixed with `g-`
* Fix a bug, due to inconsistent naming in HDFS assetstore form IDs


brianhelba:
@girder/developers PTAL.

zachmullen:
Generally speaking, how disruptive are these changes to downstream projects using Girder's test infrastructure? There have been quite a few changes lately to style rules -- if this is breaking our downstream tests, I'd prefer to batch them together and do them all at once instead of releasing a handful at a time on a frequent basis.

brianhelba:
As mentioned in the PR message, this *should* just codify rules that were already conventions, but obviously a rule isn't guaranteed to be followed until we have a check for it. In summary, I don't know how downstream projects are being affected.

As [mentioned here](https://github.com/girder/girder/pull/2071#issuecomment-306539344), the intent of static analysis rules is to improve code quality. Accordingly, if the addition of a rule causes downstream plugin tests to fail, that should be a good thing, as this new feature has has revealed previously-undetected issues with the downstream code. If there is a particular rule that is not helping to improve code quality in some way, then it shouldn't be added to upstream Girder in the first place (and I'm happy to discuss and reconsider individual rules).

As for incremental vs. batch changes, I think it's much better to submit and review changes incrementally. Making many disparate style changes in the same commit increases the likelihood of a bug slipping through the PR review process. Plugins track master at their own risk: the build should never be broken, but some churn should always be expected. We have releases for those seeking stability.

Alternatively, if churn on master is really that problematic to downstream plugins, then I would say that we need to slow down our entire development cycle: PRs should have a mandatory waiting period to ensure all developers have adequate time to chime in with comments, new features should be assembled in integration branches, then merged to master only when they're totally complete, etc. I'm not advocating this, but that would reduce disruption to downstream projects.

zachmullen:
> Accordingly, if the addition of a rule causes downstream plugin tests to fail, that should be a good thing

In practice, it can cause headaches, though. You may be forcing someone to do tedious style fixes when they otherwise would want to be doing feature work or something else entirely.

I don't know if this is happening for sure, but I wanted to poll @girder/developers to get a sense of whether any of our users had complaints about this.

brianhelba:
I'm not clear on what the objection is:
* Is it that some developers see a certain degree of style strictness (or perhaps any style rules) as a burdensome and unproductive imposition on their time? Ok, well, they're free to override or define their own style rules in their local project config, or turn off static analysis entirely. Alternatively, if particular rules add more frustration than value, we can discuss not having them.
* Alternatively, is it that the changes are being merged incrementally as opposed to in one massive push? Again, in practice, I recognize that plugins track master, but they do so at their own peril. We could perhaps just start issuing releases more frequently, so plugins can track them without falling too far behind. I see moving towards more monolithic PRs as a significant risk to upstream code quality.

zachmullen:
I'm not voicing an objection, merely attempting to survey the community.

brianhelba:
@zachmullen Certainly, I'm happy to hear what others think.

Also, for full disclosure, there is one rule that I didn't include here: [`validateAttributeQuoteMarks`](https://github.com/pugjs/pug-lint/blob/master/docs/rules.md#validateattributequotemarks-----true). While our JS and Stylus files now include rules for quotation mark style and are internally consistent, our Pug files are all over the place. This is an issue that has real effects on development, as grep-ing for strings is much more tedious if if you have to specify `['"]` everywhere. However, since that change would touch so many more lines, I left it out of this PR for now, to prevent it from bogging down the review process, or in case it was controversial enough to be a poison pill for the rest of the valuable rules in this PR. If you'd like, I can submit that change in a separate PR, but we can ensure that it gets merged at the same time as this PR.

plugins/celery_jobs/web_client/templates/configView.pug
```diff
@@ -1,16 +1,16 @@
 .g-config-breadcrumb-container
 form#g-celery-jobs-config-form(role="form")
   .form-group
-    label.control-label(for="celery_jobs_broker") Broker URL
-    input.input-sm.form-control#celery_jobs_broker(
+    label.control-label(for="g-celery-jobs-broker") Broker URL
```
zachmullen: I might otherwise complain here, but this whole plugin is on the chopping block :)

brianhelba:
@zachmullen I was admittedly a little nervous about changing the element IDs, but:
* The plugin config pages have a lot of boilerplate, so the potential for bad practices to spread via copy-paste coding is high.
* They are just element IDs, not classes, so they will never be reused outside of these specific views (which as you pointed out, are unlikely to be wrapped).

danlamanna:
> The plugin config pages have a lot of boilerplate, so the potential for bad practices to spread via copy-paste coding is high.

This has actually been one of my biggest annoyances creating Girder plugins. I don't want to become too automagic, but I've often imagined some level of API that lets us `addConfiguration('text', 'Broker URL', 'myplugin.broker_url')` to avoid the inevitable copy-paste of ConfigView.js. In the event that someone has complex validation they can resort to that, but often plugins just want to store simple strings.

zachmullen:
@danlamanna don't let your dreams be dreams!

brianhelba:
@danlamanna Such improvements would also prevent bugs like https://github.com/girder/girder/issues/1066 and https://github.com/girder/girder/pull/2205/commits/3ace10c33b57c8e9f6c742804627c4869df15e8f from surfacing in the first place.

PR #2206, opened 2017-07-20
Enable tab-separated visualization with Candela.

### LINKED ISSUE #2190: Candela plugin: support tsv format
Not sure if this is an issue in the Girder plugin or Candela itself.

jeffbaumes:
Fixes #2190, along with a few other things to make the plugin more stable.

### DIFF
--- a/plugins/candela/package.json
+++ b/plugins/candela/package.json
@@ -2,7 +2,7 @@
     "name": "candela-plugin",
     "version": "0.1.0",
     "dependencies": {
-        "candela": "^0.12.1",
+        "candela": "^0.16.0",
         "datalib": "^1.7"
     }
 }
--- a/plugins/candela/package.json
+++ b/plugins/candela/package.json
@@ -2,7 +2,7 @@
     "name": "candela-plugin",
     "version": "0.1.0",
     "dependencies": {
-        "candela": "^0.16.0",
-        "datalib": "^1.7"
+        "candela": "~0.16.0",
+        "datalib": "~1.7"
     }
 }
--- a/plugins/candela/web_client/views/CandelaParametersView.js
+++ b/plugins/candela/web_client/views/CandelaParametersView.js
@@ -50,18 +50,16 @@ const CandelaParametersView = View.extend({
         this._inputWidgets.add(new WidgetModel({
             type: 'integer',
             title: 'Width',
-            description: 'The visualization width in pixels.',
             id: 'width',
             min: 0,
-            value: 800
+            value: 400
         }));
         this._inputWidgets.add(new WidgetModel({
             type: 'integer',
             title: 'Height',
-            description: 'The visualization height in pixels.',
             id: 'height',
             min: 0,
-            value: 800
+            value: 400
         }));
 
         // Build all the widget models from the vis spec
@@ -165,8 +163,9 @@ const CandelaParametersView = View.extend({
         });
         inputs.data = this._data;
 
-        this.$('.g-candela-vis')[0].style.width = inputs.width + 'px';
-        this.$('.g-candela-vis')[0].style.height = inputs.height + 'px';
+        if (this.vis && this.vis.destroy) {
+            this.vis.destroy();
+        }
         let vis = new this._component(this.$('.g-candela-vis')[0], inputs);
         vis.render();
     }
--- a/plugins/candela/web_client/views/CandelaParametersView.js
+++ b/plugins/candela/web_client/views/CandelaParametersView.js
@@ -63,7 +63,7 @@ const CandelaParametersView = View.extend({
         }));
 
         // Build all the widget models from the vis spec
-        this._component.options.forEach((input) => {
+        _.each(this._component.options, (input) => {
             if (input.type === 'number') {
                 this._inputWidgets.add(new WidgetModel({
                     type: 'number',
--- a/plugins/candela/web_client/views/CandelaParametersView.js
+++ b/plugins/candela/web_client/views/CandelaParametersView.js
@@ -166,6 +166,7 @@ const CandelaParametersView = View.extend({
         if (this.vis && this.vis.destroy) {
             this.vis.destroy();
         }
+        this.$('.g-candela-vis').empty();
         let vis = new this._component(this.$('.g-candela-vis')[0], inputs);
         vis.render();
     }
--- a/plugins/candela/web_client/views/CandelaParametersView.js
+++ b/plugins/candela/web_client/views/CandelaParametersView.js
@@ -24,10 +24,12 @@ const CandelaParametersView = View.extend({
 
     setData: function (data, columns) {
         this._data = data;
-        this._columns = ['(none)'].concat(columns);
-        this._multiColumns = this._columns.slice(1);
-        this._numericColumns = ['(none)'].concat(
-            this._columns.filter((d) => ['number', 'integer', 'date'].indexOf(this._data.__types__[d]) >= 0));
+        this._columns = ['(none)', ...columns];
+        this._multiColumns = columns;
+        this._numericColumns = ['(none)', ...columns.filter((column) => {
+            const columnType = this._data.__types__[column];
+            return _.contains(['number', 'integer', 'date'], columnType);
+        })];
         this._multiNumericColumns = this._numericColumns.slice(1);
         this.render();
     },
--- a/plugins/candela/web_client/views/CandelaParametersView.js
+++ b/plugins/candela/web_client/views/CandelaParametersView.js
@@ -27,7 +27,7 @@ const CandelaParametersView = View.extend({
         this._columns = ['(none)'].concat(columns);
         this._multiColumns = this._columns.slice(1);
         this._numericColumns = ['(none)'].concat(
-            this._columns.filter((d) => this._data.__types__[d] === 'number'));
+            this._columns.filter((d) => ['number', 'integer', 'date'].indexOf(this._data.__types__[d]) >= 0));
         this._multiNumericColumns = this._numericColumns.slice(1);
         this.render();
     },
@@ -87,9 +87,9 @@ const CandelaParametersView = View.extend({
                 if (_.isArray(input.domain)) {
                     values = input.domain;
                     if (input.type === 'string') {
-                        value = input.domain[0];
+                        value = input.default === undefined ? input.domain[0] : input.default;
                     } else {
-                        value = [];
+                        value = input.default === undefined ? [] : input.default;
                     }
                 } else {
                     let numeric = !input.domain.fieldTypes.includes('string');
--- a/plugins/candela/web_client/views/CandelaWidget.js
+++ b/plugins/candela/web_client/views/CandelaWidget.js
@@ -1,7 +1,6 @@
 import $ from 'jquery';
 
 import View from 'girder/views/View';
-import { AccessType } from 'girder/constants';
 import events from 'girder/events';
 import candela from 'girder_plugins/candela/node/candela/dist/candela';
 import datalib from 'girder_plugins/candela/node/datalib';
--- a/plugins/candela/web_client/views/CandelaWidget.js
+++ b/plugins/candela/web_client/views/CandelaWidget.js
@@ -3,8 +3,7 @@ import $ from 'jquery';
 import View from 'girder/views/View';
 import { AccessType } from 'girder/constants';
 import events from 'girder/events';
-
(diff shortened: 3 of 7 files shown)


jeffbaumes:
@zachmullen did you have any example files from users to verify this works as expected?

zachmullen:
I have a metabolomics TSV file I can try it out on.

plugins/candela/web_client/templates/configView.pug
```diff
@@ -6,5 +6,7 @@ p
   |  to render table files directly in Girder.
 
 p.
-  This plugin requires no configuration. Simply upload a CSV file as an item,
-  then set the Candela visualization type and options.
+  This plugin requires no configuration. Simply upload a CSV file as an item
```
brianhelba: I'm not sure why we expose a configuration option for this plugin at all. It seems like a lot of pointless code complexity just to document something.

I think that:
* all of the code for the Candella config page should be removed
* the text should be moved to a new section of `docs/plugins.rst`
* the line `url: "http://girder.readthedocs.io/en/latest/plugins.html#candella"` should be added to `plugins/candella/plugin.json`.

This whole process should also be done for the `vega` plugin.

If you find my suggestion to be outside the scope of this PR, I'd be happy to implement it myself in another PR.

zachmullen:
My test on a tsv file worked. The candela test had failed and the error message looked like it was eaten, I'm investigating now.

zachmullen:
58e1230 fixes the eating of the error. I'm not sure why or when that started happening, but it now shows the actual error that's occurring.

plugins/candela/web_client/views/CandelaWidget.js
```diff
@@ -44,61 +45,77 @@ var CandelaWidget = View.extend({
...
-                }
+        let cleanup = function () {
+            this.$('.g-item-candela')
+                .remove();
+        };
+
+        if (this.accessLevel < AccessType.READ) {
+            cleanup();
+            return;
+        }
```
zachmullen: Given that we pass in an item to the constructor of this widget, I don't think it's possible we'd ever attempt to use it on an item where we don't have read access, so we can remove the above block.

plugins/candela/web_client/views/CandelaWidget.js
```diff
@@ -44,61 +45,77 @@ var CandelaWidget = View.extend({
...
+        } else {
+            cleanup();
+            return;
+        }
+
+        this.$el.html(CandelaWidgetTemplate({
+            components: this._components
+        }));
+        this.parametersView.setElement($('.g-item-candela-parameters'));
+        parser(this.item.downloadUrl(), (error, data) => {
```
zachmullen: If candela has a promise-style calling mode, we should use that instead. If not, this is OK for now.

plugins/candela/web_client/views/CandelaWidget.js
```diff
@@ -44,61 +45,77 @@ var CandelaWidget = View.extend({
...
 
-                rename.forEach((d) => {
-                    data.__types__[d[1]] = data.__types__[d[0]];
-                    delete data.__types__[d[0]];
-                    data.forEach((row) => {
-                        row[d[1]] = row[d[0]];
-                        delete row[d[0]];
-                    });
+            rename.forEach((d) => {
+                data.__types__[d[1]] = data.__types__[d[0]];
```
zachmullen: I couldn't find any mention of `__types__` in the candela docs, is it a candela thing or a vega thing?

zachmullen: Also, the nested indexing here is hard for me to read. Could we assign some human readable names to `d[0]` and `d[1]`?

plugins/candela/web_client/templates/configView.pug
brianhelba: I'm fixing this in https://github.com/girder/girder/pull/2216 . If this gets merged first, I'll incorporate these documentation changes into that PR. Otherwise, this will need to be rebased.

plugins/candela/web_client/views/CandelaWidget.js
jeffbaumes: This `parser` function is part of the `datalib` library, not candela. I don't believe it has promise-style calling.

plugins/candela/web_client/views/CandelaWidget.js
jeffbaumes: The `d[0]` issue should be resolved.

`__types__` is added to the parsed array by the `datalib` parser methods, and includes inferred type information for each field. For some reason the library does this instead of returning another value.

plugins/candela/web_client/views/CandelaWidget.js
jeffbaumes: This code is now removed.

plugins/candela/web_client/templates/configView.pug
jeffbaumes: I agree with the proposed changes and will rebase if the other PR is merged first.

plugins/candela/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "candela-plugin",
     "version": "0.1.0",
     "dependencies": {
-        "candela": "^0.12.1",
+        "candela": "^0.16.0",
```
brianhelba: I don't think plugins should necessarily be bound by this rule, but [core Girder uses the tilde (`~`) range specifier for packages that are bundled into the web client](http://girder.readthedocs.io/en/latest/development.html#modifying-core-web-client-libraries).

plugins/candela/web_client/views/CandelaWidget.js
```diff
@@ -44,61 +44,72 @@ var CandelaWidget = View.extend({
...
-                for (let key in data.__types__) {
-                    if (data.__types__.hasOwnProperty(key)) {
-                        if (key === '') {
-                            rename.push(['', 'id']);
-                        } else if (key.indexOf('.') >= 0) {
-                            rename.push([key, key.replace(/\./g, '_')]);
-                        }
+            // Vega has issues with empty-string fields and fields with dots, so rename those.
+            let rename = [];
+            for (let key in data.__types__) {
```
brianhelba: This whole thing could be significantly simpler with [`_.map`](http://underscorejs.org/#map), since that does the `hasOwnProperty` filtering for free.

plugins/candela/web_client/views/CandelaWidget.js
```diff
@@ -44,61 +44,72 @@ var CandelaWidget = View.extend({
...
+                    delete row[d.from];
                 });
+            });
 
-                let columns = [];
-                for (let key in data.__types__) {
-                    if (data.__types__.hasOwnProperty(key)) {
-                        columns.push(key);
-                    }
+            let columns = [];
```
brianhelba: This could just use [`_.keys`](http://underscorejs.org/#keys).

plugins/candela/package.json
jeffbaumes: We do this even pre-1.0 packages? Caret treats 0.x differently. From the description of caret,

> Many authors treat a 0.x version as if the x were the major "breaking-change" indicator.

This has been true with Candela. If we changed Candela to tilde, it would break if we make more changes before 1.0.

Alternately we could pin Candela to a specific version, then use tilde when it reaches 1.0.

brianhelba: Normally, `~` is more strict that `^`, [with tilde allowing only patch-level changes](https://www.npmjs.com/package/semver#tilde-ranges-123-12-1).

However, in pre-1.0 packages, the two are effectively equivalent, so this would be more a style convention when applied to the version specifier for `candela` (though it would apply to `datalib`, keeping it on `1.7.*`, instead of `1.*` as it is now).

jeffbaumes:
I've moved to `_.each` and `_.keys`. I could not stand to see `_.each` and `forEach` in the same file, so changed even a few more cases than the ones @brianhelba pointed out.

plugins/candela/web_client/views/CandelaWidget.js
```diff
@@ -44,61 +45,65 @@ var CandelaWidget = View.extend({
...
-                              'parse the data file. Details have been logged in the console.',
-                        type: 'danger',
-                        timeout: 5000,
-                        icon: 'attention'
-                    };
-                    events.trigger('g:alert', info);
-                    console.error(error);
-                    return;
-                }
+        let cleanup = function () {
```
brianhelba: Though it doesn't matter in terms of behavior here, it's probably cleaner if lambdas are defined as arrow functions unless otherwise necessary.

plugins/candela/web_client/views/CandelaWidget.js
```diff
@@ -23,7 +22,9 @@ var CandelaWidget = View.extend({
         this._components = [];
```
brianhelba: This could also use `_.each`, or use a `_.pick` + `_.keys`.

plugins/candela/web_client/views/CandelaWidget.js
```diff
@@ -23,7 +22,9 @@ var CandelaWidget = View.extend({
...
         for (let component in candela.components) {
```
brianhelba: This could also use `_.each`, or use a `_.pick` + `_.keys`.

plugins/candela/web_client/views/CandelaWidget.js
jeffbaumes: This function is no longer needed actually. I've removed it.

plugins/candela/package.json
jeffbaumes: Ah, quite right. I've changed these to tildes.

plugins/candela/web_client/views/CandelaParametersView.js
```diff
@@ -27,7 +27,7 @@ const CandelaParametersView = View.extend({
         this._columns = ['(none)'].concat(columns);
         this._multiColumns = this._columns.slice(1);
         this._numericColumns = ['(none)'].concat(
```
brianhelba: How about
```
this._numericColumns = ['(none)', ...this._columns.filter((columnName) => {
    const columnType = this._data.__types__[columnName];
    return _.contains(['number', 'integer', 'date'], columnType);
})];
```
This took may way too long to comprehend as-is.

plugins/candela/web_client/views/CandelaWidget.js
```diff
@@ -44,61 +45,60 @@ var CandelaWidget = View.extend({
...
-                        }
-                    }
-                }
+        this.$el.html(CandelaWidgetTemplate({
+            components: this._components
+        }));
+        this.parametersView.setElement($('.g-item-candela-parameters'));
+        parser(this.item.downloadUrl(), (error, data) => {
+            if (error) {
+                let info = {
```
brianhelba: I see no reason for this to be a separate variable. Our style rules allow it to be at the same indentation level inlined into `events.trigger`.

plugins/candela/web_client/views/CandelaParametersView.js
```diff
@@ -27,7 +27,7 @@ const CandelaParametersView = View.extend({
         this._columns = ['(none)'].concat(columns);
```
brianhelba: This can now be `['(none)', ...columns]`.

jeffbaumes:
@zachmullen ready for another review. @brianhelba you can sanity-check that I've addressed your comments.

brianhelba:
LGTM.

zachmullen:
Looks like we've introduced a race condition in the candela test, it's failing on master:

http://my.cdash.org/testDetails.php?test=33585996&build=1263967

jeffbaumes:
Looks like the offending code would be in https://github.com/girder/girder/blob/master/plugins/candela/plugin_tests/candelaSpec.js#L16-L30, but I admit `girderTest.waitForLoad()` is like black magic to me. Should I add a `waitsFor` for there to be two folder links (`Public` and `Private`) before trying to click on one with `$('a.g-folder-list-link:last').click();`?

zachmullen:
Yep, that's exactly the right solution.

`waitForLoad` only pertains to when you're waiting for a dialog to close, I believe.

jbeezley:
I think `waitForLoad` just waits for all pending rest requests.  `waitForDialog` is the one that waits for a modal dialog to close.

zachmullen:
`waitForDialog` actually waits for a dialog to be *open*. In my experience, if you want to wait for a dialog to *close*, it's best to use `waitForLoad`, which does also wait for rest requests to be finished, but is quite prone to race conditions if used improperly.

PR #2210, opened 2017-07-20
Refactor "Resource._matchRoute" to cleanly return a route handler

brianhelba:
This makes the following changes:
* Refactor `Resource._matchRoute` to cleanly return a route handler
  * Since this method is private, we can safely change its function signature.
* Add a `Resource.getRouteHandler` method
* Deprecate the `handler` keyword argument of `Resource.removeRoute`

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -795,6 +795,22 @@ class Resource(ModelImporter):
                 resource=resource, route=route, method=method,
                 info=handler.description.asDict(), handler=handler)
 
+    def getRouteHandler(self, method, route):
+        """
+        Get the handler method for a given method and route.
+
+        :param method: The HTTP method, e.g. 'GET', 'POST', 'PUT'
+        :type method: str
+        :param route: The route, as a list of path params relative to the
+                      resource root, exactly as it was passed to the ``route`` method.
+        :type route: tuple[str]
+        :returns: The handler method for the route.
+        :rtype: Function
+        """
+        for registeredRoute, registeredHandler in self._routes[method.lower()][len(route)]:
+            if registeredRoute == route:
+                return registeredHandler
+
     def _shouldInsertRoute(self, a, b):
         """
         Return bool representing whether route a should go before b. Checks by
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -776,21 +776,25 @@ class Resource(ModelImporter):
         :type route: tuple[str]
         :param handler: The method called for the route; this is necessary to
                         remove the documentation.
-        :type handler: function
+        .. deprecated :: 2.3.0
+        :type handler: Function
         :param resource: the name of the resource at the root of this route.
         """
         self._ensureInit()
+
         nLengthRoutes = self._routes[method.lower()][len(route)]
-        for i in range(len(nLengthRoutes)):
-            if nLengthRoutes[i][0] == route:
+        for i, (registeredRoute, registeredHandler) in enumerate(nLengthRoutes):
+            if registeredRoute == route:
+                handler = registeredHandler
                 del nLengthRoutes[i]
                 break
+
         # Remove the api doc
-        if resource is None and hasattr(self, 'resourceName'):
-            resource = self.resourceName
-        elif resource is None:
-            resource = handler.__module__.rsplit('.', 1)[-1]
-        if handler and getattr(handler, 'description', None) is not None:
+        if resource is None:
+            resource = self.resourceName \
+                if hasattr(self, 'resourceName') \
+                else handler.__module__.rsplit('.', 1)[-1]
+        if getattr(handler, 'description', None) is not None:
             docs.removeRouteDocs(
                 resource=resource, route=route, method=method,
                 info=handler.description.asDict(), handler=handler)
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -838,91 +838,92 @@ class Resource(ModelImporter):
         :param method: The HTTP method of the current request.
         :type method: str
         :param path: The path params of the request.
-        :type path: `list`
+        :type path: tuple[str]
         """
-        if not self._routes:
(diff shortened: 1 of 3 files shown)


brianhelba:
@zachmullen @girder/developers PTAL.

zachmullen:
BTW, the way I've handled this same need in the past was [like this](https://github.com/girder/girder/blob/master/plugins/hdfs_assetstore/server/__init__.py#L79-L89), but your way is more robust, in case somebody has changed the assigned handler.

brianhelba:
I'll add the `getRouteHandler` as a new public method, which will call `_matchRoute` internally, but just return the handler.

`handleRoute` can continue to call `_matchRoute` directly, since it needs the extra return values.

brianhelba:
@zachmullen @girder/developers I've updated this PR, per the discussion. See the new initial commit message for a description of changes.

girder/api/rest.py
```diff
@@ -776,25 +776,45 @@ def removeRoute(self, method, route, handler=None, resource=None):
...
         # Remove the api doc
-        if resource is None and hasattr(self, 'resourceName'):
-            resource = self.resourceName
-        elif resource is None:
-            resource = handler.__module__.rsplit('.', 1)[-1]
-        if handler and getattr(handler, 'description', None) is not None:
+        if resource is None:
+            resource = self.resourceName \
+                if hasattr(self, 'resourceName') \
+                else handler.__module__.rsplit('.', 1)[-1]
```
zachmullen: Seems like we should just use `getattr` with the default argument to achieve this construct, no?

girder/api/rest.py
```diff
@@ -776,25 +776,45 @@ def removeRoute(self, method, route, handler=None, resource=None):
...
+        :type method: str
+        :param route: The route, as a list of path params relative to the
+                      resource root, exactly as it was passed to the ``route`` method.
+        :type route: tuple[str]
+        :returns: The handler method for the route.
+        :rtype: Function
+        """
+        for registeredRoute, registeredHandler in self._routes[method.lower()][len(route)]:
+            if registeredRoute == route:
+                return registeredHandler
```
zachmullen: I'm guessing the behavior of returning `None` is by design if nothing matches; is that what we want? Should it be an exception instead?

zachmullen:
I'll mark as approved once you fix the nitpick about `hasattr` -> `getattr`

girder/api/rest.py
brianhelba: Yes, it's intended to return `None` on failure. This fact should be added to the docstring.

I don't have a strong opinion about an exception or not. Normally, failing to find something wouldn't really semantically be an "error", but since the route table is something that should be relatively predictable, there's a good argument to say that callers should always call this with the expectation of finding what they're looking for (and raising an exception) if it doesn't it.

So, unless I hear more feedback, I'll update this to throw when there's no match.

girder/api/rest.py
brianhelba: Good point. Will do.

girder/api/rest.py
zachmullen: > I'll update this to throw when there's no match

That would be my preference, simply because I imagine not finding a route will almost always be an exceptional case rather than something that is allowable in the user's logic, and in those cases, having a reasonable error message is better than "Error: NoneType has no attribute 'foo'" or whatever it says.

brianhelba:
@zachmullen Updated.

girder/api/rest.py
```diff
@@ -776,25 +776,46 @@ def removeRoute(self, method, route, handler=None, resource=None):
...
+        :type route: tuple[str]
+        :returns: The handler method for the route.
+        :rtype: Function
+        :raises: `RestException`, when no route can be matched.
+        """
+        for registeredRoute, registeredHandler in self._routes[method.lower()][len(route)]:
+            if registeredRoute == route:
+                return registeredHandler
+        else:
+            raise RestException('Could not find route "%s %s"' % (method.upper(), '/'.join(route)))
```
zachmullen: We should probably make a specific exception type for this that inherits from `RestException` so that it can be caught in the case where someone does want to gracefully handle a missing route.

brianhelba: I picked this type for consistency with how `_matchRoute` (and hence `handleRoute`) works. If we want a different exception type, then `_matchRoute` should throw that type too (and `handleRoute` is free to catch and re-throw as `RestException`).

What specifically do you propose this should throw? I don't have strong opinions, since the rules for which exception to use are so poorly followed in practice.

zachmullen: Well, at the "top" level, RestException is used for client errors (400), AccessException becomes 401/403 depending on login state, and GirderException and any other uncaught exception types are reported as server errors (5xx).

The reason it makes sense that `handleRoute` throws a RestException is because it's called based on the user's requested path, hence a client error if there's no match.

This case probably does not represent a client error, however it is in the REST layer, so a RestException isn't a bad type... I'd propose

    class RouteNotFoundError(RestException):

brianhelba: I'm aware of the theoretical taxonomy of Girder's exceptions (though I take issue with the fact that they don't all inherit from a common base class), but as a practical matter, I was just pointing out that it's not followed consistently throughout the codebase.

I'd be happy if we had a family of more specific exceptions, and if existing code was updated to use them properly, but I'm concerned that adding a whole new exception type to serve a single expert-level method which nobody is using yet is a bit outside the scope of this PR.

Perhaps we could change this to throw an `Exception`, then add a more specific type in the future, when we have a chance to consider the taxonomy in a more comprehensive way. Compatibility will be maintained, since everything inherits from `Exception`.

zachmullen: That's fine.

brianhelba: Done.

brianhelba:
@zachmullen Pushed one more commit to bump up test coverage.

PR #2211, opened 2017-07-24
Follow the documented rules for npm package versions

brianhelba:
* Packages that are bundled into the web client should use the tilde range specifier.
* Packages that are part of the build or testing process should use the caret range specifier.

This effectively bumps versions for the following packages:
* `bootstrap-switch`
* `jsoneditor`
* `webpack`


zachmullen:
> Packages that are bundled into the web client should use the tilde range specifier.
  Packages that are part of the build or testing process should use the caret range specifier.

What is the rationale for this convention?

brianhelba:
This convention was documented back in https://github.com/girder/girder/pull/2123 . This PR just updates the `package.json` file to follow it.

The rationale is to be more permissive (allowing both patch and minor releases to upgrade) for build-system dependencies, since newly introduced bugs are more likely to manifest as build or test failures, where they can be immediately addressed. Bugs in runtime dependencies may be harder to detect, as tests may not necessarily be covering their functionality directly, so only patch-level releases are allowed to upgrade.

Note that in any case, due to the presence of `package-lock.json`, users building Girder with npm >= 5 will always get the exact same packages, until we periodically do an `npm upgrade` to automatically bump pinned dependencies.

I'm open to alternative policies, the current one just made sense to me after seeing the previous total lack of package specifier consistency, and some discussion with @jbeezley .

zachmullen:
Makes sense to me. I hadn't seen the #2123.

zachmullen:
BTW, it's worth testing the bootstrap switch after these changes. In my experience, that package has been finnicky. I haven't tested it myself.

brianhelba:
`bootstrap-switch` actually already looks broken on my local build of master. Can anyone else reproduce this issue on master, or on this branch?

ss

zachmullen:
I also saw that recently. Is it a result of a specific version of the library?

brianhelba:
`bootstrap-switch` has been `@3.3.2` for a very long time, since #/1146. Its (I assume recent) breakage must be the result of another code or package change.

brianhelba:
@zachmullen If you can confirm that `bootstrap-switch` is already broken on master, I'm inclined to just merge this now, since it doesn't make the situation any worse. We can fix `bootstrap-switch` in another PR.

However, if you think the [`3.3.2 -> 3.3.4` upgrade](https://github.com/Bttstrp/bootstrap-switch/releases) will make things harder to fix, we can wait on this change.

manthey:
I don't see this problem on my local build of Girder master, nor with this branch.

danlamanna:
I also don't see it on master.

brianhelba:
Ok, gonna merge this. I'll try to reproduce the slider weirdness on another machine, and file an issue accordingly.

PR #2213, opened 2017-07-25
Use JS arrow functions in place of "var that = this;"

brianhelba:


brianhelba:
@girder/developers PTAL.

jcfr:
:+1: 

Thanks for making this consistent across the code base.

clients/web/src/views/widgets/FolderListWidget.js
```diff
@@ -20,6 +20,19 @@ var FolderListWidget = View.extend({
         },
         'click a.g-show-more-folders': function () {
             this.collection.fetchNextPage();
+        },
+        'change .g-list-checkbox': function (event) {
```
zachmullen: 👍

clients/web/src/views/widgets/GroupAdminsWidget.js
```diff
@@ -21,46 +21,43 @@ var GroupAdminsWidget = View.extend({
...
                 text: 'Are you sure you want to demote <b>' +
                     _.escape(li.attr('username')) + '</b> to a moderator?',
                 escapedHtml: true,
-                confirmCallback: function () {
+                confirmCallback: () => {
                     var user = new UserModel({_id: userid});
-                    view.model.off('g:promoted').on('g:promoted', function () {
+                    this.model.off('g:promoted').on('g:promoted', () => {
                         this.trigger('g:moderatorAdded');
-                    }, view).promoteUser(user, AccessType.WRITE);
```
zachmullen: Removing the context binding passed to `on` actually changes the behavior, since unbinding is often done using that context reference. It may not matter here for lifecycle issues, but I'd be more comfortable leaving the context variable (though it can be changed to `this`).

brianhelba: That's a good point. I've revered the places that the `this` context was removed from an `.on` binding.

All of this `.off().on()` code should go away soon anyway, once the Promise infrastructure is improved.

PR #2214, opened 2017-07-25
Refactor Vagrantfile

### LINKED ISSUE #1912: Support poll option for 'girder-install web --watch'

brianhelba:
* Make "BIND_NODE_MODULES" false by default, unless on Windows.
* Make "BIND_NODE_MODULES" idempotent and toggleable
* Set "WATCH_USEPOLLING" in client, fixing watch mode

Fixes #1912

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,5 +1,18 @@
+# Vagrant > 1.8.1 is required due to
+# https://github.com/mitchellh/vagrant/issues/6793
+Vagrant.require_version ">= 1.8.3"
+
+def true?(obj)
+  obj = obj.to_s.downcase
+  obj != "false" && obj != "off" && obj != "0"
+end
+
 Vagrant.configure("2") do |config|
-  config.vm.box = ENV["VAGRANT_BOX"] || "ubuntu/trusty64"
+  vagrant_box = ENV.fetch("VAGRANT_BOX", "ubuntu/trusty64")
+  ansible_example_name = ENV.fetch("GIRDER_EXAMPLE", "girder-dev-environment")
+  is_testing = true?(ENV.fetch("ANSIBLE_TESTING", false))
+  is_client_testing = true?(ENV.fetch("ANSIBLE_CLIENT_TESTING", false))
+  bind_node_modules = true?(ENV.fetch("BIND_NODE_MODULES", Vagrant::Util::Platform.windows?))
 
   if Vagrant.has_plugin?("vagrant-cachier")
     config.cache.scope = :box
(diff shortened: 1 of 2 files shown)


brianhelba:
@girder/developers @danlamanna @manthey PTAL.

zachmullen:
> Just to clarify, is the switch to polling solely for performance?

IIRC, filesystem watches do not work across the VM boundaries. I think Mayeul ran into this before.

msmolens:
Issue related to watch polling: https://github.com/girder/girder/issues/1912

Vagrantfile
```diff
@@ -1,5 +1,18 @@
+# Vagrant > 1.8.1 is required due to
+# https://github.com/mitchellh/vagrant/issues/6793
+Vagrant.require_version ">= 1.8.3"
```
brianhelba: It's the rationale for why we require that minimum version on the line below. Otherwise, I'm not sure I understand your question.

Vagrantfile
```diff
@@ -8,11 +21,14 @@ Vagrant.configure("2") do |config|
...
+  config.vm.provider "virtualbox" do |virtualbox|
+    virtualbox.memory = 2048
+  end
 
-  # Disable port forwarding for ansible testing, since it makes running
-  # VMs in parallel harder.
-  unless ENV["ANSIBLE_TESTING"]
+  # Disable port forwarding for ansible testing, since it makes running VMs in parallel harder.
+  unless is_testing
     config.vm.network "forwarded_port", guest: 8080, host: 9080
```
brianhelba: Yes, I had the exact same experience a long time ago, and there wasn't a solution at the time. We should file an issue and revisit it though.

brianhelba:
To be clear, there is code in here to work around two bugs:
* #1912 , which until now was unresolved, but will now be seamlessly fixed
* #2027 , which seems to only be a problem on Windows, but was fixed in #2041 in an more broad way; this preserves the fix for #2027 , but makes it more selective and idempotent

If these things are not obvious from code comments, I can put in additional comments to link to these issues.

manthey:
The issue with symlinks exists on Windows in the vagrant filesystem doesn't support symlinks on Windows.  Can symlinks be an issue on linux if the underlying filesystem doesn't support symlinks?  For instance, suppose you try to create the vagrant box on a USB stick which uses FAT32, would it work?

manthey:
I confirm that these changes work on Windows.

Vagrantfile
manthey: I'd love to see automatic port collision handled, even if the post_up_message is sub-optimal.  We could add `auto_correct: true` and change the post_up_message to something like `Girder is running at http://localhost:9080 or another port if there was a conflict`.  When there is a collision, you see the message like `Fixed port collision for 8080 => 9080. Now on port 2200.` flit by, which is hard to pick out from the noise.

Vagrantfile
danlamanna: Never mind, the required version changed but the comment stayed the same (1.8.1) which threw me off.

brianhelba:
@manthey That's why we continue to provide the `BIND_NODE_MODULES` environment variable option, so users can force the re-mounting behavior. Those running FAT32 filesystems in Linux can just add that flag when their box is first provisioned. I'm not aware of a sane way to programmatically determine whether the host filesystem supports symlinks through sync'd filesystems (and this may not be the same as supporting symlinks in general).

For the vast majority of Linux / OSX developers, I'd call this an improvement, since they can now see `node_modules` on their host filesystem, which is quite useful for development.

PR #2217, opened 2017-07-26
Ignore packaging_install test files

sutartmelson:
The `packaging_install` test creates several folders that are not tracked or ignored. This PR ignores folders in `tests/packaging/entry_point_json_plugin/` and `tests/packaging/entry_point_plugin/` directories.

### DIFF
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -1,2 +1,4 @@
 assetstore
 cases/py_client/_testDownload
+packaging/entry_point_json_plugin/*/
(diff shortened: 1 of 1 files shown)


tests/.gitignore
```diff
@@ -1,2 +1,4 @@
 assetstore
 cases/py_client/_testDownload
+packaging/entry_point_json_plugin/*/
+packaging/entry_point_plugin/*/
```
danlamanna: I think the trailing */ are unnecessary.

PR #2218, opened 2017-07-26
Send token cookie for API key token generation

zachmullen:
@cjh1 this should fix the issue you ran into yesterday.


cjh1:
@zachmullen Awesome, thanks!

PR #2222, opened 2017-07-26
WIP: Update system prerequisite installation docs

brianhelba:
This brings the system prerequisite installation docs up to date with the actual requirements of Girder, and reorganizes them to be easier to follow.

It also drops the explicit installation docs for Windows, Debian, and Arch Linux because:
* Bare-metal installation on Windows is far too unstable. Better alternatives (Docker and Virtualbox) exist for that platform.
* The Arch Linux docs were incomplete.
* Debian and Arch Linux are untested and used by a smaller minority of real-world platforms. Also, since those distributions require more advanced Linux skills to install and maintain, users would *tend* to be more sophisticated and able to follow upstream installation docs.

As @danlamanna pointed out, it's an inherent liability to maintain documentation on how to install third-party software, as it must be kept up to date. However, I think there's enough value in lowering the barriers to entry (particularly for non-software developer users) for installing Girder on the bare-metal of common platforms that the docs should stay, but be no more complex or verbose than necessary.

Finally, note that a future PR will add documentation for installation via Docker and Vagrant, which should probably be the preferred solutions for new users.


danlamanna:
Relevant Vagrant docs: http://girder.readthedocs.io/en/latest/development.html#vagrant

brianhelba:
I think Vagrant and Docker should be separate pages, then this page can be renamed to something like "native / bare-metal installation" which would include both the system prerequisites and the Python install info (maybe still as separate sub-pages).

brianhelba:
If anyone would like to work on Vagrant and/or Docker documentation, please feel free. I won't be able to get to it until next week at the earliest.

zachmullen:
Conclusion: We still need MacOS instructions. @zachmullen will fill these out.

zachmullen:
@brianhelba I only found one thing that was obviously missing from the macOS instructions, and I addressed that one to a level I feel is appropriate. The remaining TODOs are under python for linux, which I assume you're planning to handle.

brianhelba:
@danlamanna Per our discussion, I've split the prerequisites into dependencies and installation quickstart pages.

Also, see my comment in #2286 regarding a path forward on Python 3.

This is now ready for review.

brianhelba:
@danlamanna Also, I haven't yet tested the updated Beanstalk config.

docs/installation-quickstart.rst
```diff
@@ -0,0 +1,219 @@
...
+
+   .. group-tab:: macOS
+
+      Install `Homebrew <https://brew.sh/>`_.
+
+      To install all the prerequisites at once just use:
+
+      .. code-block:: bash
+
+         brew install python
```
danlamanna: Curl, git, ldap?

docs/installation-quickstart.rst
```diff
@@ -0,0 +1,219 @@
...
+
+Python Virtual Environment (optional)
+-------------------------------------
+
+To create and enable a Python virtual environment, run the commands:
+
+.. code-block:: bash
+
+   virtualenv girder_env
+   source girder-env/bin/activate
```
danlamanna: Underscore

docs/installation-quickstart.rst
```diff
@@ -0,0 +1,219 @@
...
+         echo "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/3.6 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-3.6.list
+         sudo apt-get update
+         sudo apt-get install -y mongodb-org-server mongodb-org-shell
+
+      MongoDB server will register itself as a systemd service (called ``mongod``). To start it immediately and on every
+      reboot, run the commands:
+
+      .. code-block:: bash
+
+         sudo systemctl start mongod
```
danlamanna: Doesn't Ubuntu always autostart, regardless of systemd?

docs/dependencies.rst
```diff
@@ -0,0 +1,134 @@
...
+* `Python <https://www.python.org>`_ v2.7 or v3.5+
+
+  * This is necessary to run most of the server software.
+
+  .. warning:: Some Girder plugins do not support Python v3 at this time due to third party library dependencies.
+               Namely, the ``hdfs_assetstore`` plugin and the ``metadata_extractor`` plugin will only be available in a
+               Python v2.7 environment.
+
+  .. note:: Girder performs continuous integration testing using Python v2.7 and Python v3.5. Girder *should* work on
+            newer Python v3.6+ versions as well, but that support is not verified by Girder's auftomated testing at this
```
danlamanna: Automated

docs/dependencies.rst
```diff
@@ -0,0 +1,134 @@
...
+  * v21+ is required to parse Girder's ``setup.py`` file, but the latest version can and should be installed via pip.
+
+* A Python virtual environment (optional)
+
+  * This is necessary to prevent conflicts with Girder's Python dependencies and allow installation as a non-root user.
+
+  * Virtual environments are managed by the `virtualenv package <https://virtualenv.pypa.io/>`_ in Python v2.x and by
+    the built-in `venv module <https://docs.python.org/3/library/venv.html>`_ in Python v3.5+.
+
+    .. note:: Developers needing to work on several virtual environments should consider using other packages that help
```
danlamanna: How necessary is this section?

docs/dependencies.rst
```diff
@@ -0,0 +1,134 @@
...
+    * `Amazon SES <https://aws.amazon.com/ses/>`_ may be simpler to manage for deployments already in AWS.
+
+    * Any other SMTP service should be compatible, though some may only accept mail from a limited set of "From"
+      addresses (which may be configured via a setting within Girder).
+
+  * Advanced system administrators may wish to run their own SMTP server, like
+    `Postfix <http://www.postfix.org/documentation.html>`_, as a relay.
+
+    * Beware that self-hosted SMTP relays may need additional configuration with IP whitelisting, SPF, DKIM, and other
+      measures before their mail can be reliability delivered to some email providers.
```
danlamanna: Reliably

mgrauer:
@brianhelba This is close, please refresh and merge soon.

docs/installation-quickstart.rst
brianhelba: @zachmullen added this section. I don't personally have a local OSX environment to test what actually needs to be installed (and isn't provided by the base environment). If anyone else can actually confirm that other things needed to be added here, I'd suggest they add it in the future.

docs/dependencies.rst
brianhelba: I didn't add it, and I don't feel strongly about it either way. I'd be fine with removing it in a follow-up PR.

docs/installation-quickstart.rst
brianhelba: Nope, not on systemd.

brianhelba:
@danlamanna Updated.

manthey:
Rats.  There are downstream systems that use the install_mongo.sh and install_cmake.sh scripts in CI.

brianhelba:
I'd suggest copying the scripts to downstreams, if it's required that they install that software outside of a package manager. Girder's supported deployment methods (Ansible roles, Docker images, documented install environments) don't include systems that would need to use such scripts, so I don't think Girder should be responsible for maintaining them, particularly as we transition away from cloning the Git repo for deployments.

PR #2227, opened 2017-07-27
Add Unique ID item to item view

### LINKED ISSUE #2188: Show item ID on item view
This should be a field in the Info section.

Pierre-Assemat:
Fixes #2188

### DIFF
--- a/clients/web/src/templates/body/itemPage.pug
+++ b/clients/web/src/templates/body/itemPage.pug
@@ -48,6 +48,10 @@
   .g-item-updated.g-info-list-entry
     i.icon-clock
     | Updated on #{formatDate(item.get('updated'), DATE_SECOND)}
+  .g-item-info-line.g-info-list-entry(property="id")
+    i.icon-share
+    span Unique ID:
+      span.g-bold-part #{item.id}
 
 .g-item-metadata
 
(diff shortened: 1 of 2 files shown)


clients/web/src/templates/body/itemPage.pug
```diff
@@ -48,6 +48,10 @@
   .g-item-updated.g-info-list-entry
     i.icon-clock
     | Updated on #{formatDate(item.get('updated'), DATE_SECOND)}
+  .g-item-info-line.g-info-list-entry(property="id")
+    i.icon-share
+    span Unique ID:
+      span.g-bold-part #{item.id}
```
zachmullen: This looks like this when rendered:

screen shot 2017-07-27 at 5 26 49 pm

We should make sure there's a space between the elements, and also remove the unnecessary stuff that was copy/pasted from the folder info dialog.

```
.g-item-id.g-info-list-entry
   i.icon-share
   | Unique ID: #[b= item.id]
```

Or if we think it looks better without the bold text, just


```
.g-item-id.g-info-list-entry
   i.icon-share
   | Unique ID: #{item.id}
```

Pierre-Assemat: Yes I will do that tomorrow !
Thanks for the feedback

clients/web/test/spec/itemSpec.js
```diff
@@ -107,6 +107,7 @@ function _addItemToFolder(folder) {
     runs(function () {
         expect($('.g-item-name').text()).toBe('Test Item Name');
         expect($('.g-item-description').text().trim()).toBe('Test Item Description');
+        expect($('.g-item-id').text()).toBeDefined();
```
zachmullen: Might be better to make sure this matches a regular expression.

PR #2230, opened 2017-07-28
New Sorting widget

Xarthisius:
This adds a simple widget that allows to sort a collection using pre-populated fields list. I added it to userList view as an example (and mainly because I needed it there). Since I'm CSS-impaired I'd appreciate all hints how to properly format it.


clients/web/src/templates/widgets/sortCollectionWidget.pug
```diff
@@ -0,0 +1,14 @@
...
+    i.icon-sort-name-down
+  button.g-collection-sort-actions.btn.btn-sm.btn-default.dropdown-toggle(
+      data-toggle="dropdown", title="Sort by Field")
+    | Sort by
+    i.icon-down-dir
+  ul.g-collection-sort-menu.dropdown-menu(role="menu")
+    each fieldName, field in fields
+      li(role="presentation")
+        a.g-collection-sort-link(role="menuitem", g-sort=field)
+          | #{fieldName}
```
zachmullen: `= fieldName`

clients/web/test/spec/sortWidgetSpec.js
```diff
@@ -0,0 +1,72 @@
...
+        runs(function () {
+            expect($('a.g-collection-sort-link').text()).toBe('Last NameCreation DateUsed Space');
+            expect($('a.g-collection-sort-link:contains("Creation Date")').length).toBe(1);
+            $('a.g-collection-sort-link:contains("Creation Date")').click();
+        });
+
+        waitsFor(function () {
+            // FIXME: the condition here should be: wait for re-render of collection,
+            //   but I don't know how to express that...
+            return $('.g-user-link').text() === 'AFirstName ALastNameCFirstName CLastNameBFirstName BLastName';
```
Xarthisius: @zachmullen any suggestion how to make this more robust?

zachmullen: For something like this, I might waitFor the condition of the text not being empty, then afterward use `expect` with `contains` on the text to assert that it contains a certain substring.

Xarthisius: It won't work that way, text() is already non empty at this point, but still holds previous values. `spyOn` looks promising (I sneak-peeked at PR 2237) but I don't know how to use it properly.

zachmullen: Ah I see, you're asserting the order of them. I'd say the best way would be to assert that the text matches a regex, where you add wildcards around them in case any additional data is added in the future.

clients/web/src/views/body/UsersView.js
```diff
@@ -63,6 +74,7 @@ var UsersView = View.extend({
         }));
 
         this.paginateWidget.setElement(this.$('.g-user-pagination')).render();
+        this.sortCollectionWidget.setElement(this.$('.g-user-sort')).render();
```
zachmullen: We should add a little right-margin on this so it isn't right up against the search widget.

clients/web/src/views/widgets/SortCollectionWidget.js
```diff
@@ -0,0 +1,51 @@
...
+import 'bootstrap/js/dropdown';
+
+/**
+ * This widget is used to provide a consistent widget for sorting
+ * pages of a Collection by a choosen field.
+ */
+var SortCollectionWidget = View.extend({
+    events: {
+        'click a.g-collection-sort-link': function (event) {
+            var cid = $(event.currentTarget).attr('g-sort');
```
zachmullen: This variable name is misleading, we should call it sortField or something.

clients/web/src/views/body/UsersView.js
```diff
@@ -42,6 +43,16 @@ var UsersView = View.extend({
             parentView: this
         });
 
+        this.sortCollectionWidget = new SortCollectionWidget({
+            collection: this.collection,
+            parentView: this,
+            fields: {
+                lastName: 'Last Name',
+                created: 'Creation Date',
+                size: 'Used Space'
```
zachmullen: `size` and `created` are not indexed fields at the moment, we should add indexes to them in the user model.

clients/web/src/stylesheets/body/userList.styl
```diff
@@ -48,6 +48,7 @@
 
   .g-user-sort
     display inline-block
+    margin-right: 12px
```
zachmullen: No : required

clients/web/test/spec/sortWidgetSpec.js
```diff
@@ -33,7 +33,7 @@ describe('Sort user list', function () {
         girderTest.goToUsersPage()();
         runs(function () {
             expect($('.g-user-list-entry').length).toBe(3);
-            expect($('a.g-user-link').text()).toBe('AFirstName ALastNameBFirstName BLastNameCFirstName CLastName');
+            expect($('a.g-user-link').text()).toMatch(/.*AFirstName ALastNameBFirstName BLastNameCFirstName CLastName.*/);
```
zachmullen: Probably most robust to add a `.*` in between them in case we add more data that happens to fall between the values.

zachmullen:
Thanks @Xarthisius , this sort of widget was a long time coming 👍

PR #2234, opened 2017-07-31
Fixes #2233. Install npm deps of staticWebDependencies

### LINKED ISSUE #2233: Must install npm deps of staticWebDependencies
We fail to do this currently, the bug can be reproduced by:

    girder-install web --plugins candela

And a workaround is

    girder-install web --plugins candela,item_tasks

zachmullen:
This just removes the conditional block around plugin npm install logic.

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -511,21 +511,19 @@ module.exports = function (grunt) {
         }
 
         // process all external dependencies when building
-        if (buildPlugin) {
-            processPluginDependencies(plugin, dir, config);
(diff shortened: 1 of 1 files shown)


jbeezley:
Won't this always npm install all plugin dependencies?  One issue is that dependencies like vega require extra system libraries to compile node bindings.  With this change, I don't think you can avoid requiring them.

zachmullen:
It will only npm install plugin dependencies for the plugins that are requested to be built. If you don't depend on vega, it won't be configured or built at all.

jbeezley:
Right, let me test it out then.

PR #2237, opened 2017-07-31
Use Jasmine spies to mock functions in testing

brianhelba:


brianhelba:
@girder/developers PTAL.

@zachmullen @manthey Even if you don't review this, please take note, since it applies to code that you've written.

zachmullen:
Thanks, I gave it a quick once-over, but will mark final review tomorrow.

clients/web/test/specRunner.js
```diff
@@ -176,6 +176,16 @@ page.onLoadFinished = function (status) {
         page.evaluate(function () {
             if (window.girderTest) {
                 girderTest.promise.done(function () {
+                    // Allow Jasmine to compare RegExp using toEqual, toHaveBeenCalledWith, etc.
+                    jasmine.getEnv().addEqualityTester(function (a, b) {
```
zachmullen: Hm, I see the value here, but this could cause some crazy headache in the future if someone is trying to actually test that a method was called with a certain regexp using jasmine mocks.

Is there a way to just add a new `toHaveBeenCalledWithRegex` method so that we don't have to change comparisons globally?

zachmullen: Or alternatively, to pass a custom comparator to a single invocation of `toHaveBeenCalledWith`?

zachmullen:
General question: how compatible is the spy API between this version of jasmine and the next major version?

clients/web/test/specRunner.js
brianhelba: `toHaveBeenCalledWith` is not a general matcher, it's specific to asserting that a spied method was called at some point in time with the given argument(s) (in any parameter position). To do so, it delegates to [`jasmine.Env.prototype.equals_`](https://github.com/jasmine/jasmine/blob/v1.3.1/src/core/Env.js#L228), which already supports the comparison of two actual regexes. This method also has lots of "equality" testers that perform matching in all sorts of non-strict ways (this is the same method used by the `toEqual` matcher). Since `toHaveBeenCalledWith` is already doing loose `toEqual`-style comparison, instead of strict (using `===`) `toBe`-style comparison, I'd call it a feature that you can now test spied arguments against regexes. If a user ever wanted really strict comparison, they shouldn't use `toHaveBeenCalledWith` in the first place, regardless of whether this additional equality tester is present.

However, this does highlight a small bug in how this custom equality tester was added: it should also ensure that the other argument is a string, or else the [`RegExp.prototype.test`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) could be called with an invalid argument. Once that bug is fixed, comparison of two regexes will continue to work properly in all cases.

zachmullen: > However, this does highlight a small bug in how this custom equality tester was added: it should also ensure that the other argument is a string

Yes, let's make that change

brianhelba:
For the purposes of Girder, the next major version should actually be the [Sinon library](http://sinonjs.org/), rather than Jasmine 2.0. The biggest difference is that Sinon spies do not mock behavior by default, which you explicitly have to ask for in Jasmine 1.3 with `andCallThrough`.

A transition out of Jasmine 1.3 is absolutely going to require human review of all code that uses the Jasmine API. Given that, I'd say that the transition to use Jasmine 1.3-style spies now is still an improvement, as it makes the intended semantics of the mocking much more obvious to other developers. I'm less concerned about migrations that require renaming of some methods, and more concerned about a future migration that require the developer to reverse-engineer the intent of non-standard DIY mocks to implement them for the first time using a spy library.

PR #2241, opened 2017-08-01
Upgrade Babel loader and presets

brianhelba:
* Use the latest `babel-loader` for Webpack
* Cache Babel outputs
* Add `babel-preset-es2016`, to support the exponentiation (`**`) operator

### DIFF
--- a/grunt_tasks/webpack.config.js
+++ b/grunt_tasks/webpack.config.js
@@ -25,7 +25,10 @@ var webpack = require('webpack');
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
 
 var paths = require('./webpack.paths.js');
-var es2015Preset = require.resolve('babel-preset-es2015');
+// Resolving the Babel presets here is required to support symlinking plugin directories from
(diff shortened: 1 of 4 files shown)


grunt_tasks/webpack.config.js
```diff
@@ -25,7 +25,6 @@ var webpack = require('webpack');
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
 
 var paths = require('./webpack.paths.js');
-var es2015Preset = require.resolve('babel-preset-es2015');
```
brianhelba: Thanks, I remember seeing some discussion about that previously, but never read the details myself. I'll retest with the current config, and switch back to `require.resolve` if it's still broken.

I think we really need to ensure that there are code comments when these sorts of non-obvious workarounds are implemented.

brianhelba: Rather than bog down this PR with any of the symlink issues, I've reverted that change and added a code comment.

We can test and revisit in the future.

PR #2242, opened 2017-08-02
npm packaging

zachmullen:


jbeezley:
I might suggest an alternative that would serve to separate girder-the-application from girder-the-js-library.  It isn't totally fleshed out, but it would look something like this:

1. Create a `package.json` file in `/clients/web/src` that contains only runtime dependencies (i.e. no webpack/grunt libraries).  I would name this package something like `@girder/core` or `@girder/girder`.
2. Modify the root level `package.json` to depend on this to pull in the run time dependencies (to avoid duplicating them).  It would look something like this:
```js
dependencies: {
  "@girder/core": "file:./clients/web/src",
  // rest of the build dependencies
}
```
3.  When we move to a more conventional packaging for plugins, we could do something similar, adding plugin packages to `optionalDependencies`.

This would avoid pulling in all kinds of dependencies that aren't necessary for using girder as a client library and would put the sources at the root level of the package.  It would even allow us to use the toplevel npm package as a girder "application" package containing all of the built sources.

zachmullen:
That means we can move the build-time requirements, but I'm not sure where they should go...

* Move them to @girder/core under devDependencies?
* Keep them as dependencies of the outer package and remove them altogether from @girder/core?

clients/web/src/package.json
```diff
@@ -0,0 +1,29 @@
...
+        "url": "https://github.com/girder/girder/issues"
+    },
+    "license": "Apache-2.0",
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/girder/girder.git"
+    },
+    "dependencies": {
+        "as-jqplot": "~1.0.8",
+        "babel-preset-es2015": "^6.24.1",
```
brianhelba: This package should not be here.

package.json
```diff
@@ -84,5 +75,12 @@
         "lint": "eslint --cache clients/web Gruntfile.js grunt_tasks scripts clients/web/test && pug-lint clients/web/src/templates",
         "docs": "grunt docs",
         "esdoc": "esdoc"
+    },
+    "files": [
+        "clients/web/src/"
+    ],
+    "main": "clients/web/src/index.js",
```
brianhelba: I'm thinking that this entry point should be in the `@girder/core` package.

This file should be essentially a super-build for the Girder "app", whereas `@girder/core` is the Girder library. So, this file's entry point should be `clients/web/src/main.js`.

brianhelba:
Keep the build-time dependencies in the outer packags. As it was before, it won't matter whether they're under `dependencies` or `devDependencies`, so I'd keep them in-place under `dependencies`, so we can continue to abuse the mechanism for build-only vs development installs.

If you put them in the inner `@girder/core` package, then the only way to get them will be to install  `@girder/core` package directly as the top-level package, which would require `@girder/core` to depend directly on plugins (so they can all be built), which we don't want. If you keep the build deps in the top-level package, we can eventually have that package be the one to import and build all the `@girder/core` and plugin code.

package.json
zachmullen: I can go either way on this. [Here](https://webpack.github.io/docs/resolving.html#resolving-an-absolute-path) are the docs of how the `main` field of package.json is used by webpack (which is the only way we are using this field, since neither of these makes sense as a node entry point). Obviously our index.js should be the `main` for girder/core, but I'm not sure how someone might even use `main` in the outer app package -- maybe it makes sense to simply omit it for now. What do you think?

jbeezley: I think we can remove it.  The outer package shouldn't really be dependency of anything else anyway.  The only reason I can see to publishing it is to provide the built sources, in which case the user will just copy them to a local static resources path.  In which case, "resolving" the main script doesn't make much sense.

zachmullen:
@girder/developers I'd really like to write some docs on this, but the best place for it in the current docs would be the "External Web Clients" section, which is slated for destruction/rewrite. Is it OK to put it there and we can salvage it later, or should I put it into a different section?

zachmullen:
I'm going to open this up for review. One thing I want to consider is the naming -- we have the `girder` top-level npm package, and now we have the `@girder` org. One change I'd like to consider is perhaps making the core src distribution be `girder`, and changing the outer package to be `@girder/app` or something. Since most of the people using the external packaging will be using the core sdist rather than the outer app, I think the name `girder` looks more official.

zachmullen:
Once we are happy with the changes, I'll make the versions in the package.json be `2.3.0`; once it passes packaging tests on master after being merged, then I will do `npm publish`.

jbeezley:
I'm happy with this as is.  I don't have much a preference about naming except that we should decide now, because changing it later will be much more of a problem.  I would mildly favor moving everything into the `@girder` org and posting a deprecation warning the to current `girder` package pointing people to the correct place.

zachmullen:
My preference would be that what I have as `@girder/core` in this repo should become `girder`, the reason being that a lot of our existing plugin code and examples have import statements that look like 

    import ... from 'girder/...';

And if we make the core src package be called simply "girder", all that will still work even if they're using a version installed from npm package.

jbeezley:
Yeah, that makes sense.  It would probably be best to move core to girder then.

package.json
```diff
@@ -44,19 +39,14 @@
...
         "pug": "^2.0.0-rc.2",
         "pug-loader": "^2.3.0",
-        "remarkable": "~1.7.1",
-        "sprintf-js": "~1.0.3",
         "style-loader": "^0.13.2",
         "stylus": "^0.54.5",
         "stylus-loader": "^3.0.1",
         "swagger-ui": "~2.2.10",
         "toposort": "^1.0.3",
-        "underscore": "~1.8.3",
```
brianhelba: Underscore is used in Gruntfiles too. It should stay here.

brianhelba: To be clear, it should be in both files now. For consistency with the version convention, this one should now use a `^` version specifier.

brianhelba:
Just so I'm clear, we're going to skip anything with `files` or `.npmignore` because:
* The `clients/web/src` package simply includes everything inside that directory
* The top-level (`app`) package will not be published for now, but just exists as a build tool

Correct?

zachmullen:
As written, the top-level `@girder/app` package uses `files` to whitelist only relevant files.

brianhelba:
If we're going to publish `@girder/app`, then including `clients/web/src` in it would be an unwanted duplication. The right way would be to include those files via a dependency on the `girder` package, but without forcing every developer to use `npm link` to do development, I'm not yet sure how make `@girder/app` work in both development and when installed.

I see no reason to publish `@girder/app` right now, and it might be best to just hold off until we have a more complete solution that incorporates plugins too.

zachmullen:
> I see no reason to publish @girder/app right now, and it might be best to just hold off until we have a more complete solution that incorporates plugins too.

That's fine by me. I'll remove that from the release instructions, and remove it from the whitelist. I'll leave `clients/web/static` in place in the outer app package whitelist; those files are not pulled in at build time, but are referenced at runtime by the main app. If we ever feel like we need, for example, the girder icon PNGs to be bundled in the core package, my vote would be to move them under `src` and pull them in at build time using file-loader.

zachmullen:
In fact, it's probably a bug that `Girder_Mark.png` isn't already pulled in via file-loader into the widget that uses it.

brianhelba:
Agreed. `Girder_Mark.png` should probably be moved to `clients/web/src/assets`.

brianhelba:
For what it's worth, I've just validated that the `package-lock.json` file is correct, as it produces the same* output when rebuilt from scratch on my machine.

\* It doesn't actually produce the same output, due to `fsevents` being optional. but that's just npm being npm.

brianhelba:
`Girder_Mark.png` will be handled in #2244 , since it requires non-trivial bugfixes.

zachmullen:
Thanks @brianhelba 

Since this is potentially risky as a concept, I'd like to have at least one other person mark approval before I merge and publish the package.

brianhelba:
Agreed, we can wait for 2x approvals on this. I assume @jbeezley can provide the other review.

PR #2244, opened 2017-08-02
Support runtime loading of static file assets

brianhelba:
Changes in this PR are:
* Add `getApiRoot` and `getStaticRoot` methods to web client
* Deprecate `apiRoot` and `staticRoot` in web client's "girder/rest"
* Set [Webpack's `publicPath`](https://webpack.js.org/configuration/output/#output-publicpath), so static content can be fetched at runtime
* Include `Girder_Mark.png` in the web client and load it with Webpack
  * The use of [`require` is supported by `pug-loader`](https://github.com/pugjs/pug-loader#embedded-resources), but would also work with any import resolution mechanism that has `require` globally defined.
* Upgrade file-loader, and use more robust config options for it

This fixes a long-standing bug where any file assets that were imported by Webpack were never actually referenced with the correct external URL path. This was previously not a problem in practice upstream, as file asset importing was never actually used, despite having [an extensive Webpack configuration to support it](https://github.com/girder/girder/blob/5fd1ef844b3afdea58b541e2e0f634534e868098/grunt_tasks/webpack.config.js#L151). Other downstream users of Girder's build system [had to work around this bug](https://github.com/ImageMarkup/isic-archive/blob/2db881a838835d7c38970447b46b1e6324c6a8f1/Gruntfile.js#L110) by explicitly defining `pluginPath`.

Note that this does not affect file assets loaded via `url()` in Stylus files ([like `jsoneditor-icons.png`](https://github.com/girder/girder/blob/5fd1ef844b3afdea58b541e2e0f634534e868098/clients/web/src/stylesheets/widgets/metadataWidget.styl#L65), which was previously the only file asset loaded by core Girder). Those are [resolved by `stylus-loader`](https://github.com/girder/girder/blob/5fd1ef844b3afdea58b541e2e0f634534e868098/grunt_tasks/webpack.config.js#L113), and emit a relative path to the outputted file's name (e.g. `url("~jsoneditor/dist/img/jsoneditor-icons.png")` transforms to `url(assets/jsoneditor-icons-bf23347e.png)` in the built code). The browser resolves this relative path with respect to the path of the referencing file (in this example, `girder_lib.min.css`, which is typically served at `static/built/girder_lib.min.css`), which not-entirely-coincidentally is correct. It would probably be more robust to resolve everything directly through Webpack, instead of ever using `stylus-loader`, in case CSS  files and file assets are ever in a different place, but that can be attempted in a future PR.


brianhelba:
This does not yet address file assets loaded for plugins. That is probably going to require one additional fix, which I'll make after some additional testing.

Until then, this is technically a WIP, but is still a strict improvement over the current configuration (which is broken for both core and plugins).

brianhelba:
It looks like setting `__webpack_public_path__` only affects the chunk that it's called from. Since plugins are defined in their own chunk, each plugin will have to set `__webpack_public_path__` locally. I'm hoping we can find a way around this requirement, but for now, I'd like to start review as-is, since none of the built-in plugins actually load file assets (outside of Stylus).

brianhelba:
@girder/developers PTAL.

brianhelba:
Added 1 more commit, upgrading `file-loader`. I tried out the new `useRelativePath` option, and it works for imports in Javascript files, but it breaks imports in Pug files.

clients/web/src/rest.js
```diff
@@ -5,28 +5,75 @@ import Backbone from 'backbone';
...
- * @param root The full root path for the static content.
+ * @param {string} root The root path for the static content.
  */
 function setStaticRoot(root) {
-    staticRoot = root;
+    // Strip trailing slash
+    staticRoot = root.replace(/\/$/, '');
+    // publicPath would normally be set in the Webpack config file, but is not known at compile
+    // time.
+    __webpack_public_path__ = `${getStaticRoot()}/built/`; // eslint-disable-line no-undef, camelcase
```
zachmullen: This feels wonky, like we shouldn't be having to do this. We use assets from CSS files in the item_tasks plugin, is the limitation just that we can't use them from pug files? Would it make more sense to set the `src` attribute programmatically from javascript at render time?

brianhelba: This is the correct way to do things. The CSS is being loaded by the Browser's native mechanism for handling `url()` in CSS, [which allows relative paths in the spec](https://developer.mozilla.org/en-US/docs/Web/CSS/url).

In Javascript (whether its in native JS or a Pug file), there is no native concept of a relative path: JS files are run the exact same way, regardless of how they were sourced or loaded by the browser. So, if you want to load static assets via Webpack in JS, you need to have `publicPath` set, because the path that the files were emitted to on disk is not necessarily connected to the URL that the files can be fetched from.

It's definitely a bug that `publicPath` was never set before. The extensive caviet comment is a result of the fact that ES6 imports are always hoisted to the top of a module (unlike `require`), in combination with the fact that Girder's architecture decided to allow `staticRoot` to be set at any point at runtime.

PR #2245, opened 2017-08-03
Clean up and fix minor bugs in Grunt code

brianhelba:
* Upgrade grunt-shell and cleanup its usage for building docs
* Reorganize the Grunt Fontello building, ensuring correct task ordering
* Roll the "init" Grunt task into the "default" Grunt task
* Rename the "setup" task to "local-config", and clean it up
* Upgrade grunt-contrib-uglify

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -61,9 +61,7 @@ module.exports = function (grunt) {
         pluginDir: 'plugins',
         staticDir: 'clients/web/static',
         isSourceBuild: isSourceBuild,
-        default: {
-            setup: {}
-        }
+        default: {}
     });
 
     if (isSourceBuild) {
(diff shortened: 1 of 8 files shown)


Gruntfile.js
```diff
@@ -98,33 +95,19 @@ module.exports = function (grunt) {
...
      *     'uglify:a': {
      *       'dependencies': ['copy:a']
      *     }
      *   }
      * }
      *
-     * The init task will run `copy:a` followed by `uglify:a`.
+     * The 'default' task will run `copy:a` followed by `uglify:a`.
      */
-    grunt.registerTask('init', sortTasks(grunt.config.get('init')));
```
jbeezley: I don't know of any examples of this, but it's possible there might be plugins that rely on the init task.  What do you think about replacing this task with a deprecation warning?  I think you could also merge the `init` object with the `default` object before running the next line.

brianhelba: Sounds reasonable, will do.

brianhelba: @jbeezley Done.

grunt_tasks/localConfig.js
```diff
@@ -0,0 +1,45 @@
...
+module.exports = function (grunt) {
+    const confDir = `${grunt.config.get('girderDir')}/conf`;
+    const localConfigPath = `${confDir}/girder.local.cfg`;
+
+    grunt.config.merge({
+        copy: {
+            'local-config': {
+                src: `${confDir}/girder.dist.cfg`,
+                dest: localConfigPath
+            }
```
jbeezley: I don't know about doing this as part of the client build.  It is really a server issue.  I think a more appropriate place to create this file automatically would be as part of the python install.

zachmullen: Agreed, this should not be part of the web client build.

brianhelba: Yeah, I was really surprised to see it here too, and agree that it should live in the Python code.

However, this PR is just a refactor / bugfix of some of existing functionality, so moving the functionality is a bit out of scope. I've created an issue here: https://github.com/girder/girder/issues/2263 , which may be good for @sutartmelson or @Pierre-Assemat to handle.

jbeezley: That works for me.

PR #2247, opened 2017-08-03
Allow celery tasks to perform girder_worker.run transitions

cjh1:


cjh1:
@kotfic This should fix up the issue we discussed.

PR #2255, opened 2017-08-04
Resolve a race condition in the candela test

jeffbaumes:
See https://github.com/girder/girder/pull/2206#issuecomment-320256826.

This passed 20 times in a row on my machine. We'll see how Circle likes it.

### DIFF
--- a/plugins/candela/plugin_tests/candelaSpec.js
+++ b/plugins/candela/plugin_tests/candelaSpec.js
@@ -10,14 +10,9 @@ describe('Test the candela UI.', function () {
         runs(function () {
             expect($('#g-user-action-menu.open').length).toBe(0);
             $('.g-user-text>a:first').click();
-        });
-        girderTest.waitForLoad();
-
-        runs(function () {
             expect($('#g-user-action-menu.open').length).toBe(1);
             $('a.g-my-folders').click();
         });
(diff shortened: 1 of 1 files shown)


plugins/candela/plugin_tests/candelaSpec.js
```diff
@@ -19,6 +19,10 @@ describe('Test the candela UI.', function () {
         });
         girderTest.waitForLoad();
```
zachmullen: We should remove the `waitForLoad` here as well as the one above, and we can combine the previous two `runs` blocks into one since we don't have any asynchronous action to wait for between them.

jeffbaumes: To be clear, it would look like this, correct?

```
        runs(function () {
            expect($('#g-user-action-menu.open').length).toBe(0);
            $('.g-user-text>a:first').click();
            expect($('#g-user-action-menu.open').length).toBe(1);
            $('a.g-my-folders').click();
        });
```

zachmullen: Yep!

jeffbaumes: Ok this change is in. The first build of this PR failed on installing requirements-dev, hopefully this works.

PR #2256, opened 2017-08-04
Guard against exceptions in entry point plugins.

manthey:
If an entry point plugin's load method is called and throws an exception, Girder will fail to start.

I found that this would occur in Python 3.5 but not in Python 2.7 when the large_image plugin was installed in devel (-e) mode; it worked fine in install (no -e) mode.

### DIFF
--- a/girder/utility/plugin_utilities.py
+++ b/girder/utility/plugin_utilities.py
@@ -68,9 +68,9 @@ def _recordPluginFailureInfo(plugin, traceback):
 def _clearPluginFailureInfo(plugin):
     """
     If a plugin loaded, clear any failure information that may have been set by
-    an earlier failed attempy.
+    an earlier failed attempt.
 
-    :param plugin: The name of the plugin that laoded.
+    :param plugin: The name of the plugin that loaded.
     :type plugin: str
     """
     _pluginFailureInfo.pop(plugin, None)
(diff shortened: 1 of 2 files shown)


girder/utility/plugin_utilities.py
```diff
@@ -65,6 +65,17 @@ def _recordPluginFailureInfo(plugin, traceback):
     }
 
 
+def _clearPluginFailureInfo(plugin):
+    """
+    If a plugin loaded, clear any failure information that may have been set by
+    an earlier failed attempy.
```
jbeezley: *attempt

girder/utility/plugin_utilities.py
```diff
@@ -65,6 +65,17 @@ def _recordPluginFailureInfo(plugin, traceback):
...
+
+    :param plugin: The name of the plugin that laoded.
```
jbeezley: *loaded

girder/utility/plugin_utilities.py
```diff
@@ -295,10 +307,23 @@ def findEntryPointPlugins(allPlugins):
                         logprint.exception(
                             'ERROR: Plugin "%s": plugin.yml is not valid '
                             'YAML.' % entry_point.name)
-        except ImportError:
+        except (ImportError, SystemError):
+            # Fall throw and just try to load the entry point below.  If there
```
jbeezley: Fall through?

PR #2257, opened 2017-08-04
RFC: Documentation overhaul

zachmullen:
Fire up that loud, another round of docs!

Please take a look at the proposed front-page changes, this simply establishes a straw man TOC structure, and in particular, overhauls the organization and content that appears on the front page of the docs.

To build locally, check out this branch and run `npm run docs`, which requires you to have installed the stuff in `requirements-dev.txt`.


jeffbaumes:
@zackgalbreath I recently discovered that from the RTD site you can (temporarily) turn on any branch to be built and hosted. Does it makes sense to do that for this branch so people can easily see the updated docs?

zachmullen:
Good idea, I've triggered a build. When it's done, they'll appear [here](http://girder.readthedocs.io/en/doc-reorg/).

jeffbaumes:
Thanks, the changes so far on this branch look great to me.

jcfr:
Release notes could be added at the end. Given the current doc, it would like this. See https://github.com/girder/girder/pull/2276#issuecomment-321375574

ronichoudhury:
This looks great!
- I like the new introduction, including the focus on extensibility.
- I *love* the new tutorials section :)
- The "decomposition by audience" text reads very well and provides a rather concrete sense of context.

docs/tutorials.rst
```diff
@@ -0,0 +1,38 @@
...
+behavior to support new authentication services or protocols, add a new backend
+storage engine for file storage, or even interact with a completely different DBMS
+to persist system records -- the extent to which plugins are allowed to modify and
+extend the core system behavior is nearly limitless.
+
+Plugins are self-contained in their own directory within the Girder source tree.
+Therefore they can reside in their own separate source repository, and are installed
+by simply copying the plugin source tree under an existing Girder installation's
+`plugins` directory. The Girder repository contains several generally
+useful plugins out of the box, which are described in the :doc:`plugins` section.
```
mgrauer: do you want to make the distinction about 'built-in' vs 'external/3rd party' plugins here?  basically do you want to canonize the terms in the docs?

docs/tutorials.rst
```diff
@@ -0,0 +1,38 @@
+Tutorials
+=========
+
+
+# TODO(zach) move the following section somewhere useful
```
mgrauer: I agree, maybe a developer guide, or just a general architecture and background section would be good here.  It will probably be more clear after there is at least one tutorial.

docs/developer-install.rst
```diff
@@ -0,0 +1,210 @@
+Development Installation Guide
+==============================
+
+# TODO break this up into dev vs. prod
```
mgrauer: I agree, this is a lot of content on one page.

docs/deploy.rst
```diff
@@ -1,7 +1,7 @@
 .. _deploy:
 
-Deploy
-======
+Deployment Guide
+================
```
mgrauer: I know it wasn't added as a part of this PR, but there is a lot of detail about Google Container Engine.  Do we want that level of detail here, and do we want to support all these details for the future?  When is the last time anyone tried this?

docs/index.rst
```diff
@@ -26,11 +26,18 @@ Girder: a data management platform
 What is Girder?
 ---------------
 
-Girder is a free and open source web-based **data management platform** developed by
+Girder is a libre open source web-based **data management platform** developed by
```
zachmullen: "free" is ambiguous: free as in beer vs. free as in freedom. Libre was the word created to express the second form precisely. I don't have strong feelings about this wording, I'm open to suggestions. If we keep libre, I'm certainly OK with a link to its meaning.

docs/index.rst
```diff
@@ -46,51 +53,38 @@ that have some or all of the following requirements:
...
+    to third-party authentication services such as OAuth or LDAP/Active Directory.
 
 * **Authorization management**
     Girder supports a simple access control scheme that allows both user-based
     and role-based access control on resources managed in the system. The project
     has undergone rigorous security audits and has extensive automated testing
     to exercise authorization behavior and ensure correctness.
 
-For an overview of the concepts present in Girder, we recommend checking out the :doc:`user-guide`.
+* **Offline analytics / data processing**
```
zachmullen: I always struggle to find the right word for what the worker does: offline? batch? distributed?

What I'm trying to express is what you said, that it does not occur on the web server and that it is asynchronous.

docs/index.rst
```diff
@@ -46,51 +53,38 @@ that have some or all of the following requirements:
...
+    Many users of Girders also deploy its companion offline processing tool,
+    `Girder Worker <http://girder-worker.readthedocs.io/en/latest/>`_. Using the worker,
+    you can run batch analytics on data hosted in Girder using python scripts
+    or arbitrary Docker images, with real-time job monitoring, logging, and statistics
```
zachmullen: I think you're just personally traumatized by that word

docs/tutorials.rst
zachmullen: I'd be OK with that.

docs/deploy.rst
zachmullen: If we are going to enumerate every platform we can deploy on, it definitely shouldn't include instructions for each one. I'd rather just say that the system can be deployed on a wide variety of systems, from bare metal to IaaS to managed services like EBeanStalk.

docs/index.rst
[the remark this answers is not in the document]
mgrauer: I'm fine with libre, as long as there is a link or an explanation, and if you want to be an advocate for this word.  We could also say "free as in beer and free as in speech", I'm fine either way.

docs/index.rst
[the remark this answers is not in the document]
mgrauer: Howsabout "asynchronous remote execution"?

the title section could be "Asynchronous remote analytics / data processing" and the paragraph text could "also deploy its companion offline processing tool, GW, which provides asynchronous remote execution."

docs/index.rst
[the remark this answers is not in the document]
mgrauer: you may be on to something

docs/deploy.rst
mgrauer: agreed.

I don't want to drop the instructions on EB because we are actively using that (though maybe this thinking doesn't scale, because everyone could have their set of favored deployments...), but I'd be fine splitting those out somewhere else.  If this specific page could have just the most important or canonical deployment types, that would make it easier to read through.

For the GCE stuff specifically, I say we remove it.

docs/index.rst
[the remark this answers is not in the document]
ronichoudhury: "Offline" almost certainly is the wrong word. I'd go with "offloaded" or "delegated" to mean that the computation gets done on a dedicated process separate from Girder itself. "Asynchronous" is close but because we're in web-world, the term feels overloaded.

docs/deploy.rst
ronichoudhury: My (perhaps unpopular) opinion on this is that the detailed instructions for a given deployment platform could be made into a tutorial that's targeted to "Deploying Girder on ElasticBeanstalk" or whatever. Having a multitude of those is not a problem (or at least, not as much of a problem as stuffing all of them into the `Deployment Guide` section).

And keep in mind, my use of the word "tutorial" is pretty fluid - when I say "tutorial", the thing I'm thinking of might be called a "recipe" or "developer cookbook entry" by someone else.

docs/index.rst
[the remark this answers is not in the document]
jcfr: See https://en.wikipedia.org/wiki/Gratis_versus_libre

docs/index.rst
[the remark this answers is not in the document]
jcfr: What about "Distributed analytics/execution/data processing " ?

docs/index.rst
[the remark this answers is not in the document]
jcfr: `live job monitoring` ?

docs/deploy.rst
jcfr: Each tutorial would have the following metadata:
* known to work with Gider X.Y.Z
* last updated on YYYY-MM-DD
* last tested on YYYY-MM-DD
* authors: John Doe (Foo Inc), ...

zachmullen: Honestly, that's the very thing that worries me about the idea of having a whole bunch of tutorials with little curation -- for a newcomer developer, they are just presented with some large, overwhelming list of links, and then they have to try to sort through which ones are actually up to date. I'd rather have a smaller list where everything is known (or at least expected) to work with the version that the docs correspond to. We have a limited team of maintainers, and doc maintenance is not something we historically have a ton of resources to allocate toward.

jcfr: > I'd rather have a smaller list where everything is known (or at least expected) to work with the version that the docs correspond to.

Of course.

> doc maintenance is not something we historically have a ton of resources to allocate toward.

Which is sad ... without explaining and documenting how to use a (open source) project, its adoption and use internal or external will fade ... 

The good news is that Girder has been a great example regarding documentation

I like to say *If it is not tested and documented, it is broken*



All of that said, I was just sharing my experience acquired while tackling similar issue with 3D Slicer. 

It is still nice to keep track of half backed tutorials, experiments, etc .. these simply need to be linked of a place labeled as such and not prominently visible from the documentation landing page. As you told, we do not want to confuse or overwhelm new comers. 

To help move the link around, understanding when an experiment/tutorial/etc .. has been created, last tested is quite important. With the resources at hand, we can't always manually test everything or write test for it. 

I think that this rather easy to add metadata will help make inform (and unbiased) decisions. 

Idea: These "half baked" experiments/tutorials/... could be linked of a "Girder Labs" Github wiki page ?

kotfic: I'm actually a huge fan of the external 'tutorial' idea for more exotic deployment scenarios. This documentation should cover whatever we consider to be the 'canonical' deployment with a section that essentially advertises that other deployments are possible (with links to tutorials).  

This also provides an opportunity to say something like "If you're interested in more complex deployments please send us an email,  or drop by our gitter channel blah blah blah"  I want to be *very* careful to not come across as doing a direct sell,  but i don't think its bad to communicate "We've done this in the past and we'd be happy to help you do it now"

docs/tutorials.rst
```diff
@@ -0,0 +1,38 @@
...
+
+Girder's server-side architecture is focused around the construction of RESTful
+web APIs to afford minimal coupling between the backend services and the
+frontend clients. This decoupling allows multiple clients all to use the same
+server-side interface. While Girder does contain its own single-page javascript
+web application, the system can be used by any HTTP-capable client, either inside
+or outside of the web browser environment. Girder can even be run without its
+front-end application present at all, only serving the web API routes.
+
+The web API is mostly used to interact with resources that are represented by **models**
```
kotfic: Would you consider adding something here that identifies models as [Data access objects](https://en.wikipedia.org/wiki/Data_access_object)?  This was confusing for me initially when working with girder coming from other web based systems where ```model``` referred to objects that were instantiated with the data from individual records.

zachmullen:
@zachmullen @ronichoudhury @mgrauer will meet to move this forward.

zachmullen:
@mgrauer @ronichoudhury following up from our meeting, let's merge this into the integration branch and fix it up there. Remaining open issues from this PR should be added into the TODO list on https://github.com/girder/girder/milestone/6

brianhelba:
FYI, #2336 is now merged.

docs/dev-installation.rst
```diff
@@ -0,0 +1,206 @@
+
+Developer Installation Guide
+============================
+
+If you wish to develop Girder or Girder plugins, you can either
+:ref:`install Girder natively <native_install>` on your machine or
```
brianhelba: I'd like to push Vagrant as the default way for new developers to be exposed to Girder:
* It's fewer commands, and platform-agnostic, making it way less effort and less intimidating, which lowers barriers to entry.
* It's far safer for them, as their host machine won't be cluttered with things like MongoDB, etc. that they may not want otherwise.
* It's safer for us, as we don't have to account for the machine's prior state, and get support requests / negative PR from people that were unable to install easily.

So, if @girder/developers agree, can we reorder these docs to place Vagrant first, and perhaps emphasize that this is preferred (including in our trainings and other communications)? Power users are always free to install it on bare metal, but we should let them opt into that set of risk / rewards.

To support this, I hope to port Vagrant to Ubuntu 16.04 soon.

zachmullen:
* Rebase
* Add stubs for todo sections
* Merge?

jcfr:
> Rebase

👍  

With 413 modified files, it not in a reviewable state

zachmullen:
Yeah... forgot to merge master into doc-reorg-integration

docs/dev-installation.rst
```diff
@@ -1,9 +1,161 @@
...
+Creating a virtual environment
+------------------------------
+
+While not strictly required, it is recommended to install Girder within
+its own `virtual environment <http://docs.python-guide.org/en/latest/dev/virtualenvs/>`_
+to isolate its dependencies from other python packages.  To generate a new
+virtual environment, first install/update the ``virtualenv`` and ``pip``
+packages ::
+
+   sudo pip install -U virtualenv pip
```
jcfr: Ideally, these instructions should be tested and extracted from a real shell script using `.. literalinclude::`. For example, see https://github.com/NeurodataWithoutBorders/pynwb/blob/b715e5523c3e2b2b1bd219c23a8b06cbe6cb9445/docs/source/example.rst#examples

docs/dev-installation.rst
```diff
@@ -1,9 +1,161 @@
...
+Obtain the Girder source code by cloning the Git repository on
+`GitHub <https://github.com>`_: ::
+
+    git clone https://github.com/girder/girder.git
+    cd girder
+
+To run the server, you must install some external Python package
+dependencies: ::
+
+    pip install -e .
```
jcfr: Same thing here, all of these should be tested

docs/dev-installation.rst
zachmullen: Cool feature, but out of scope for this PR.

jcfr: Fair enough

jcfr:
👍  For convenience, here is rebased topic that would allow to keep history clean: https://github.com/girder/girder/compare/doc-reorg-rebased

docs/index.rst
```diff
@@ -104,14 +98,6 @@ Table of contents
 .. toctree::
    :maxdepth: 2
 
-   admin-docs
-   user-docs
+   installation
+   user-guide
    developer-docs
```
jcfr: Could we keep the changelog ?

zachmullen: Yes, but I'm not sure whether it should be a top level heading or go under another section. @girder/developers  thoughts?

mgrauer: I don't have a strong opinion about where, but I would like it to be included. How about as a top level heading so we don't have to argue about which other heading it should nest under?

zachmullen:
I'm assuming this was closed on accident, but for some reason I am not allowed to reopen it. @brianhelba are you able to do so?

brianhelba:
Yes, this was an accident. I'll attempt to fix.

To fully explain, I deleted `doc-reorg-integration` from `origin`, because `doc-reorg-integration` was fully merged into `origin/master`, so I assumed it was old and just needed to be pruned. However, it looks like GitHub now considers this PR closed because the target branch (`doc-reorg-integration`) is gone, even though the topic branch (`doc-reorg`) still remains on `origin`. Sorry. :fearful:

brianhelba:
@zachmullen Fixed.

PR #2259, opened 2017-08-04
Accept metadata object during creation of item or folder

Xarthisius:
This allows to create item/folder with metadata with just one call. Right now, for reasons I don't understand all tests using *POST /item* and *POST /folder* fail with `Invalid JSON passed in request body.` even though metadata parameter is not required (and given default in one case)

### DIFF
--- a/girder/api/v1/folder.py
+++ b/girder/api/v1/folder.py
@@ -151,10 +151,12 @@ class Folder(Resource):
         .param('parentType', "Type of the folder's parent", required=False,
                enum=['folder', 'user', 'collection'], strip=True)
         .param('parentId', 'Parent ID for the new parent of this folder.', required=False)
+        .jsonParam('metadata', 'A JSON object containing the metadata keys to add',
+                   paramType='form', requireObject=True, required=False)
         .errorResponse('ID was invalid.')
         .errorResponse('Write access was denied for the folder or its new parent object.', 403)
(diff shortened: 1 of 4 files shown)


zachmullen:
> all tests using POST /item and POST /folder fail with Invalid JSON passed in request body

That's because you specified your `jsonParam` with `type='body'`, in fact you want it to be set to `'form'`.

jbeezley:
Would it be reasonable to add something similar to the `PUT` request as well?

zachmullen:
That would be reasonable, albeit duplicative of the `.../metadata` endpoints. I've never been fond of how that API was structured from the get-go, so I'm fine with adding a different, easier way to do it.

danlamanna:
I've wanted this to be the case for quite a while! Though I wonder if it
makes sense for this logic to be applied to updating items/folders in
this PR.

One benefit of that would be to eventually allow our client-side
item/folder saves to automatically handle updating metadata instead of
having to make multiple requests and function calls via the
MetadataMixin.js.

zachmullen:
We don't need to make that change on this PR; I told @Xarthisius that I would handle the JS side of this after this goes in, and I can do it on that PR.

jbeezley:
Yeah, I don't mean to pile more work on this PR.  We can always make an issue for it to do later.

PR #2261, opened 2017-08-05
Fixes #2253. Include version.js inside girder npm package

### LINKED ISSUE #2253: version.js not included in npm package
By virtue of being .gitignored, the file doesn't make it into the package, and we import it inside index.js, so our recent npm girder package does not yet work.

zachmullen:
Our code in index.js requires version.js to exist, but version.js
is dynamically generated at build time and is not checked into
source control. To work around this, we generate a sensible
version.js file as a prepublish step on the source package.


zachmullen:
I published a test package from this branch to make sure everything was in order. Try:

```
npm install girder@2.3.1-alpha.1
ls -la node_modules/girder
cat node_modules/girder/version.js 
```

To test out the auto-generated version.js, I'd recommend starting a `girder-install web --watch`, then `cd clients/web/src ; node _prepublish.js`. After the rebuild, your Girder front page should show the new info, which should not contain a git SHA.

Issue #2251, opened 2017-08-03
Display number of users in a Girder instance


mgrauer:
I have a need to display the number of users in a girder instance, to admin users only, in the web client.  This could be just a number in parenthesis next to the users nav link, e.g.  

    Users (42)

I'd be fine with a generalization of this, e.g. maybe it could be enabled/disabled by a config setting, or available to all or only admin users as a config setting.

brianhelba:
For site admin users, who can see all users, this is actually a very cheap and easy operation to perform: we just count the number of documents in the `user` collection.

If we want to support non-admin users seeing this, we'd have to do standard access filtering, since users are access controlled and could be private in theory. To do this on every user in the DB is likely expensive enough that we should probably at least do it in an aggregation pipeline, if we allow it at all.

The only open question is where we should expose this info. To be consistent with the other endpoints that list `nFolders`, `nItems`, etc., I'd propose that we make a `/user/details` endpoint that will return an object like:
```
{
    'nUsers': 42
}
```

@girder/developers Any thoughts about this?

jeffbaumes:
I've always thought it a bit odd that normal users can see the full user list (that's the current default behavior, right?). I don't know why it is needed, since users can be discovered via the search menu. It seems like the user list is mostly for admins, and even seems like a slight security issue being able to enumerate all users.

So, I'd suggest only admins to see/access the Users sidebar item, and the (full) number of users can be displayed in parentheses.

brianhelba:
@jeffbaumes That would be a major change, which I'd also have some thoughts about. Can you file that (`Hide user list from normal users`) as a separate issue, so we can cleanly debate the scope and implications there?

jeffbaumes:
To scope down my thoughts, I think @brianhelba's `nUsers` field should only be admin-accessible. Why would a normal user want to know how many users they have access to see? And, as @brianhelba notes, this case would be more complex to implement.

brianhelba:
Until #2252 goes somewhere, I see nothing wrong with exposing the count functionality to all users if we can do so with reasonable performance. The alternative (which you can do right now), is just to page through the users list, manually counting the number of users until you reach the end.

Pierre-Assemat:
New pull request in working progress #2262

Pierre-Assemat:
Once the server side is done, what are the expectation on girder client side ?

brianhelba:
I think putting something at the top of the `#users` page (`UsersView`) would be the simplest and best.

[...]

PR #2262, opened 2017-08-07
Add a new route : user/details to get the number of users

Pierre-Assemat:
This PR display the number of users (only available with an admin user). 
The `user-icon` is annotated with the count.

This is the view of the count of user : 
screenshot from 2017-08-15 17-34-50

### DIFF
--- a/clients/web/src/collections/UserCollection.js
+++ b/clients/web/src/collections/UserCollection.js
@@ -1,5 +1,6 @@
 import Collection from 'girder/collections/Collection';
 import UserModel from 'girder/models/UserModel';
+import { restRequest } from 'girder/rest';
 
 var UserCollection = Collection.extend({
     resourceName: 'user',
@@ -8,6 +9,15 @@ var UserCollection = Collection.extend({
     // Override default sort field
     sortField: 'lastName',
     secondarySortField: 'firstName'
+}, {
+    getUsersCount: function () {
+        return restRequest({
+            path: 'user/details'
+        })
+            .then((resp) => {
+                return resp['nUsers'];
+            });
+    }
 });
 
 export default UserCollection;
--- a/clients/web/src/collections/UserCollection.js
+++ b/clients/web/src/collections/UserCollection.js
@@ -10,7 +10,7 @@ var UserCollection = Collection.extend({
     sortField: 'lastName',
     secondarySortField: 'firstName'
 }, {
-    getUsersCount: function () {
+    getTotalCount: function () {
         return restRequest({
             path: 'user/details'
         })
--- a/clients/web/src/collections/UserCollection.js
+++ b/clients/web/src/collections/UserCollection.js
@@ -12,7 +12,8 @@ var UserCollection = Collection.extend({
 }, {
     getTotalCount: function () {
         return restRequest({
-            path: 'user/details'
+            url: 'user/details',
+            method: 'GET'
         })
             .then((resp) => {
                 return resp['nUsers'];
--- a/clients/web/src/stylesheets/body/userList.styl
+++ b/clients/web/src/stylesheets/body/userList.styl
@@ -58,3 +58,22 @@
     input[type="text"]
       width 170px
       margin-right 13px
+
+  .g-user-count-container
+    display inline-block
+    color #888
+    font-size 16px
+
+    i.icon-user
+      position relative
+      top 2px
+
+    .g-user-count
+      position relative
+      background-color #eee
+      border-radius 3px
+      color rgba(0, 0, 0, 0.75)
+      bottom -7px
+      right 7px
+      padding 0 0.2em
+      font-size 9px
--- a/clients/web/src/stylesheets/body/userList.styl
+++ b/clients/web/src/stylesheets/body/userList.styl
@@ -64,7 +64,7 @@
     color #888
     font-size 16px
 
-    i.icon-user
+    i
       position relative
       top 2px
 
--- a/clients/web/src/stylesheets/body/userList.styl
+++ b/clients/web/src/stylesheets/body/userList.styl
@@ -61,19 +61,18 @@
 
   .g-user-count-container
     display inline-block
-    color #888
-    font-size 16px
+    position relative
 
     i
-      position relative
-      top 2px
+      color #888
+      font-size 16px
+      vertical-align middle
 
     .g-user-count
-      position relative
       background-color #eee
       border-radius 3px
-      color rgba(0, 0, 0, 0.75)
-      bottom -7px
-      right 7px
       padding 0 0.2em
       font-size 9px
+      position absolute
+      left 72%
+      top 55%
--- a/clients/web/src/templates/body/userList.pug
+++ b/clients/web/src/templates/body/userList.pug
@@ -7,6 +7,9 @@
     button.g-user-create-button.btn.btn-sm.btn-default
       i.icon-plus-squared
       |  Create user
+    i.icon-male
+      span.badge
+        | #{getUsersCount}
 
 each user in users
   .g-user-list-entry
--- a/clients/web/src/templates/body/userList.pug
+++ b/clients/web/src/templates/body/userList.pug
@@ -1,5 +1,10 @@
 .g-user-list-header
   .g-user-pagination
+  if getCurrentUser() && getCurrentUser().get('admin')
+    .g-user-count-container
+      i.icon-user
+      span.g-user-count
+        = getUsersCount
   .g-user-sort
   form.g-user-search-form.form-inline(role="form")
     .form-group.g-users-search-container
@@ -7,9 +12,6 @@
     button.g-user-create-button.btn.btn-sm.btn-default
       i.icon-plus-squared
       |  Create user
-    i.icon-user
-      span.badge
-        = getUsersCount
 
 each user in users
   .g-user-list-entry
--- a/clients/web/src/templates/body/userList.pug
+++ b/clients/web/src/templates/body/userList.pug
@@ -7,9 +7,9 @@
     button.g-user-create-button.btn.btn-sm.btn-default
       i.icon-plus-squared
       |  Create user
-    i.icon-male
+    i.icon-user
       span.badge
-        | #{getUsersCount}
+        = getUsersCount
 
 each user in users
   .g-user-list-entry
--- a/clients/web/src/templates/body/userList.pug
+++ b/clients/web/src/templates/body/userList.pug
@@ -1,6 +1,7 @@
+- const isAdmin = currentUser && currentUser.get('admin');
 .g-user-list-header
   .g-user-pagination
-  if getCurrentUser() && getCurrentUser().get('admin')
+  if isAdmin
     .g-user-count-container
       i.icon-user
       span.g-user-count
@@ -8,7 +9,7 @@
   .g-user-sort
   form.g-user-search-form.form-inline(role="form")
     .form-group.g-users-search-container
-  if getCurrentUser() && getCurrentUser().get('admin')
+  if isAdmin
     button.g-user-create-button.btn.btn-sm.btn-default
       i.icon-plus-squared
       |  Create user
--- a/clients/web/src/templates/body/userList.pug
+++ b/clients/web/src/templates/body/userList.pug
@@ -4,8 +4,7 @@
   if isAdmin
     .g-user-count-container
       i.icon-user
-      span.g-user-count
-        = getUsersCount
+      span.g-user-count= usersCount
   .g-user-sort
   form.g-user-search-form.form-inline(role="form")
     .form-group.g-users-search-container
--- a/clients/web/src/views/body/UsersView.js
+++ b/clients/web/src/views/body/UsersView.js
@@ -34,7 +34,7 @@ var UsersView = View.extend({
     initialize: function (settings) {
         cancelRestRequests('fetch');
         this.collection = new UserCollection();
-        var promiseArray = [];
+        const promiseArray = [];
         promiseArray.push(this.collection.fetch());
 
         this.paginateWidget = new PaginateWidget({
@@ -60,11 +60,11 @@ var UsersView = View.extend({
         }).on('g:resultClicked', this._gotoUser, this);
 
         if (getCurrentUser() && getCurrentUser().get('admin')) {
-            var userCountpromise = UserCollection.getUsersCount()
-                .done((resp) => {
-                    this.usersCount = resp;
+            const userCountPromise = UserCollection.getTotalCount()
+                .done((count) => {
+                    this.usersCount = count;
                 });
-            promiseArray.push(userCountpromise);
+            promiseArray.push(userCountPromise);
         }
         this.register = settings.dialog === 'register' && getCurrentUser() &&
                         getCurrentUser().get('admin');
--- a/clients/web/src/views/body/UsersView.js
+++ b/clients/web/src/views/body/UsersView.js
@@ -34,9 +34,8 @@ var UsersView = View.extend({
     initialize: function (settings) {
         cancelRestRequests('fetch');
         this.collection = new UserCollection();
-        this.collection.on('g:changed', function () {
-            this.render();
-        }, this).fetch();
+        var promiseArray = [];
+        promiseArray.push(this.collection.fetch());
 
         this.paginateWidget = new PaginateWidget({
             collection: this.collection,
@@ -60,14 +59,27 @@ var UsersView = View.extend({
             parentView: this
         }).on('g:resultClicked', this._gotoUser, this);
 
+        if (getCurrentUser() && getCurrentUser().get('admin')) {
+            var userCountpromise = UserCollection.getUsersCount()
+                .done((resp) => {
+                    this.usersCount = resp;
+                });
+            promiseArray.push(userCountpromise);
+        }
         this.register = settings.dialog === 'register' && getCurrentUser() &&
                         getCurrentUser().get('admin');
+
+        $.when(...promiseArray)
+            .done(() => {
+                this.render();
+            });
     },
 
     render: function () {
         this.$el.html(UserListTemplate({
             users: this.collection.toArray(),
             getCurrentUser: getCurrentUser,
+            getUsersCount: this.usersCount,
             formatDate: formatDate,
             formatSize: formatSize,
             DATE_DAY: DATE_DAY
--- a/clients/web/src/views/body/UsersView.js
+++ b/clients/web/src/views/body/UsersView.js
@@ -72,9 +72,7 @@ var UsersView = View.extend({
 
         $.when(...promiseArray)
             .done(() => {
-                this.listenTo(this.collection, 'g:changed', () => {
-                    this.render();
-                });
+                this.listenTo(this.collection, 'g:changed', this.render);
                 this.render();
             });
     },
@@ -83,7 +81,7 @@ var UsersView = View.extend({
         this.$el.html(UserListTemplate({
             users: this.collection.toArray(),
             currentUser: getCurrentUser(),
-            getUsersCount: this.usersCount,
+            usersCount: this.usersCount,
             formatDate: formatDate,
             formatSize: formatSize,
             DATE_DAY: DATE_DAY
--- a/clients/web/src/views/body/UsersView.js
+++ b/clients/web/src/views/body/UsersView.js
@@ -34,6 +34,7 @@ var UsersView = View.extend({
     initialize: function (settings) {
         cancelRestRequests('fetch');
         this.collection = new UserCollection();
+
         const promiseArray = [];
         promiseArray.push(this.collection.fetch());
 
@@ -71,6 +72,9 @@ var UsersView = View.extend({
 
         $.when(...promiseArray)
             .done(() => {
+                this.listenTo(this.collection, 'g:changed', () => {
+                    this.render();
+                });
                 this.render();
             });
     },
@@ -78,7 +82,7 @@ var UsersView = View.extend({
     render: function () {
         this.$el.html(UserListTemplate({
             users: this.collection.toArray(),
-            getCurrentUser: getCurrentUser,
+            currentUser: getCurrentUser(),
             getUsersCount: this.usersCount,
             formatDate: formatDate,
             formatSize: formatSize,
--- a/clients/web/test/spec/userSpec.js
+++ b/clients/web/test/spec/userSpec.js
@@ -87,9 +87,15 @@ describe('Create an admin and non-admin user', function () {
         }, 'front page to display');
     });
 
-    it('check redirect to front page after logout from users list page', function () {
+    it('check the user count with an admin user', function () {
         girderTest.login('admin', 'Admin', 'Admin', 'adminpassword!')();
         girderTest.goToUsersPage()();
+        runs(function () {
+            expect($('span.g-user-count').text()).toBe('3');
+        });
+    });
+
+    it('check redirect to front page after logout from users list page', function () {
         girderTest.logout()();
         waitsFor(function () {
             return $('.g-frontpage-title:visible').length > 0;
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -202,8 +202,7 @@ class User(Resource):
         Description('Get the number of users.')
     )
     def getUsersDetails(self):
-        userModelCollection = self.model('user').collection
-        nUsers = userModelCollection.count()
+        nUsers = self.model('user').find().count()
         return {'nUsers': nUsers}
 
     @access.user
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -43,7 +43,7 @@ class User(Resource):
         self.route('GET', ('authentication',), self.login)
         self.route('GET', (':id',), self.getUser)
         self.route('GET', (':id', 'details'), self.getUserDetails)
-        self.route('GET', ('details',), self.getUserNumber)
+        self.route('GET', ('details',), self.getUsersDetails)
         self.route('POST', (), self.createUser)
         self.route('PUT', (':id',), self.updateUser)
         self.route('PUT', ('password',), self.changePassword)
@@ -197,14 +197,14 @@ class User(Resource):
         self.model('user').remove(user)
         return {'message': 'Deleted user %s.' % user['login']}
 
+    @access.admin
     @autoDescribeRoute(
         Description('Get the number of users.')
     )
-    def getUserNumber(self):
+    def getUsersDetails(self):
         userModelCollection = self.model('user').collection
         nUsers = userModelCollection.count()
         return {'nUsers': nUsers}
-
 
     @access.user
     @filtermodel(model='user')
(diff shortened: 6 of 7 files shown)


girder/api/v1/user.py
```diff
@@ -43,6 +43,7 @@ def __init__(self):
         self.route('GET', ('authentication',), self.login)
         self.route('GET', (':id',), self.getUser)
         self.route('GET', (':id', 'details'), self.getUserDetails)
+        self.route('GET', ('details',), self.getUserNumber)
```
zachmullen: Since this is just the count of users, let's rename it to `/user/count`.

girder/api/v1/user.py
```diff
@@ -196,6 +197,15 @@ def deleteUser(self, user):
         self.model('user').remove(user)
         return {'message': 'Deleted user %s.' % user['login']}
 
+    @autoDescribeRoute(
```
zachmullen: We should decorate this method with `access.admin` to make it explicit that it is only usable by admins.

tests/cases/user_test.py
```diff
@@ -856,3 +856,21 @@ def testPrivateUser(self):
...
+            email='admin@admin.com', password='adminadmin')
+        # Create an user1 and user2
+        self.model('user').createUser('user1', 'passwd1', 'tst1', 'usr1',
+                                      'user1@user.com')
+        self.model('user').createUser('user2', 'passwd2', 'tst2', 'usr2',
+                                      'user2@user.com')
+        # Make sure we get a 200 when trying to count users 
+        resp = self.request(path='/user/details', method='GET')
+        self.assertStatus(resp, 200)
+        self.assertEqual(resp.json['nUsers'], "3")
```
zachmullen: This should fail as written, it should be equal to `3`, not `"3"`.

tests/cases/user_test.py
```diff
@@ -856,3 +856,21 @@ def testPrivateUser(self):
...
+            firstName='Admin', lastName='Admin', login='admin',
+            email='admin@admin.com', password='adminadmin')
+        # Create an user1 and user2
+        self.model('user').createUser('user1', 'passwd1', 'tst1', 'usr1',
+                                      'user1@user.com')
+        self.model('user').createUser('user2', 'passwd2', 'tst2', 'usr2',
+                                      'user2@user.com')
+        # Make sure we get a 200 when trying to count users 
+        resp = self.request(path='/user/details', method='GET')
+        self.assertStatus(resp, 200)
```
zachmullen: The `self.assertStatusOk(resp)` should be used to assert a 200-level response. That will still work if we change to a 201 or 202 or some other 200-level code in the future.

tests/cases/user_test.py
```diff
@@ -856,3 +856,21 @@ def testPrivateUser(self):
...
+        """
+        Make sure display number of user is correct
+        """
+        # Create an admin user
+        self.model('user').createUser(
+            firstName='Admin', lastName='Admin', login='admin',
+            email='admin@admin.com', password='adminadmin')
+        # Create an user1 and user2
+        self.model('user').createUser('user1', 'passwd1', 'tst1', 'usr1',
+                                      'user1@user.com')
```
zachmullen: This will fit all on one line since we allow 100 chars.

girder/api/v1/user.py
```diff
@@ -42,7 +42,8 @@ def __init__(self):
         self.route('GET', ('me',), self.getMe)
         self.route('GET', ('authentication',), self.login)
         self.route('GET', (':id',), self.getUser)
-        self.route('GET', (':id', 'details'), self.getUserDetails)
+        self.route('GET', (':id', 'count'), self.getUserDetails)
```
zachmullen: I missed this before -- there should not be an `:id` token in the path, since this does not refer to a specific user but the whole collection of users in the system.

zachmullen: Actually what happened was I think you changed the old endpoint on accident, rather than your new one below.

girder/api/v1/user.py
```diff
@@ -196,6 +197,15 @@ def deleteUser(self, user):
         self.model('user').remove(user)
         return {'message': 'Deleted user %s.' % user['login']}
 
+    @access.admin
+    @autoDescribeRoute(
+        Description('Get the number of users.')
+    )
+    def getUserNumber(self):
```
mgrauer: This method name is a bit unclear, when I read it, I interpret it to be get a number from a specific user (which doesn't exist, but in many data models this would exist).

Please change it to `getUserCount`

tests/cases/user_test.py
```diff
@@ -856,3 +856,19 @@ def testPrivateUser(self):
             path='/file/chunk', user=pvt, fields=fields, files=files)
         self.assertStatusOk(resp)
         self.assertEqual(resp.json['itemId'], itemId)
+
+    def testNumberUser(self):
```
mgrauer: also change this one to `testUserCount`

tests/cases/user_test.py
```diff
@@ -856,3 +856,19 @@ def testPrivateUser(self):
...
+        """
+        Make sure display number of user is correct
```
mgrauer: Change this text to `Test that the user count is correct.`

tests/cases/user_test.py
```diff
@@ -856,3 +856,19 @@ def testPrivateUser(self):
...
+        """
+        # Create an admin user
+        self.model('user').createUser(
+            firstName='Admin', lastName='Admin', login='admin',
+            email='admin@admin.com', password='adminadmin')
+        # Create an user1 and user2
+        self.model('user').createUser('user1', 'passwd1', 'tst1', 'usr1', 'user1@user.com')
+        self.model('user').createUser('user2', 'passwd2', 'tst2', 'usr2', 'user2@user.com')
+        # Make sure we get a 200-level when trying to count users
```
mgrauer: You can probably remove this comment line because it is implied by the name `assertStatusOk`, and also it is implied from the fact that this is a test for that specific endpoint.

Generally try to avoid adding comments for explaining exactly what the code is doing and is obvious from the code.  Instead the comments should communicate intent, or else explain assumptions or what the code is doing that may be complex or non-obvious.  Of course what is "obvious" to one person isn't to another, and it is always a bit a judgement call as to whether it is useful to add in a comment like this.

brianhelba:
@zachmullen I think this should be called `/user/details`, to be consistent with the endpoints:
* `/collection/:id/details`
* `/folder/:id/details`

both of which return only return counts of sub-entries (and not any other details).

In addition to consistency, I could see that we might want to expose additional statistics about all users at some point, which would fit nicely into an existing route at `/user/details`.

zachmullen:
That's fair, I'm fine with leaving it as details then.

girder/api/v1/user.py
```diff
@@ -196,6 +197,15 @@ def deleteUser(self, user):
         self.model('user').remove(user)
         return {'message': 'Deleted user %s.' % user['login']}
 
+    @access.admin
+    @autoDescribeRoute(
+        Description('Get the number of users.')
+    )
+    def getUsersDetails(self):
+        userModelCollection = self.model('user').collection
```
brianhelba: Girder's `Model` exposes its own version of `find()` (which returns a PyMongo cursor), so we should use that, instead of the lower-level `collection.find()`. So, this code can just be:
```
nUsers = self.model('user').find().count()
```

girder/api/v1/user.py
```diff
@@ -196,6 +197,14 @@ def deleteUser(self, user):
         self.model('user').remove(user)
         return {'message': 'Deleted user %s.' % user['login']}
 
+    @access.admin
+    @autoDescribeRoute(
+        Description('Get the number of users.')
+    )
+    def getUsersDetails(self):
+        nUsers = self.model('user').collection.find().count()
```
brianhelba: This should be:
```
self.model('user').find().count()
```
The [Model itself has a `find` method](https://github.com/girder/girder/blob/master/girder/models/model_base.py#L237), which returns a PyMongo cursor, so we don't need to go to the underlying PyMongo collection to call the method.

tests/cases/user_test.py
```diff
@@ -856,3 +856,27 @@ def testPrivateUser(self):
...
+        users = [self.model('user').createUser(
+            'usr%s' % num, 'passwd', 'tst', 'usr', 'u%s@u.com' % num)
+            for num in [0, 1]]
+        resp = self.request(path='/user/details', user=admin, method='GET')
+        self.assertStatusOk(resp)
+        self.assertEqual(resp.json['nUsers'], 3)
+        # test for a no-admin user
+        resp = self.request(path='/user/details', user=users[0], method='GET')
+        self.assertStatus(resp, 403)
+        resp = self.request(path='/user/details', user=users[1], method='GET')
```
brianhelba: I see no reason to test this with two different non-admin users. Just try it with one non-admin user.

clients/web/src/templates/body/userList.pug
```diff
@@ -7,6 +7,9 @@
     button.g-user-create-button.btn.btn-sm.btn-default
       i.icon-plus-squared
       |  Create user
+    i.icon-male
+      span.badge
+        | #{getUsersCount}
```
brianhelba: You should add some text here, to make it clear what this value is.

brianhelba: You should add some text here, to make it clear what this value is.

clients/web/src/templates/body/userList.pug
```diff
@@ -7,6 +7,9 @@
     button.g-user-create-button.btn.btn-sm.btn-default
       i.icon-plus-squared
       |  Create user
+    i.icon-male
```
brianhelba: I'm not sure if this is the best icon, but maybe others can comment.

clients/web/src/views/body/UsersView.js
```diff
@@ -34,9 +34,8 @@ var UsersView = View.extend({
     initialize: function (settings) {
         cancelRestRequests('fetch');
         this.collection = new UserCollection();
-        this.collection.on('g:changed', function () {
-            this.render();
-        }, this).fetch();
+        var promiseArray = [];
```
brianhelba: In new JS code, use `let` for things that will be reassigned, and `const` otherwise. This should be `const`.

clients/web/src/views/body/UsersView.js
```diff
@@ -60,14 +59,27 @@ var UsersView = View.extend({
             parentView: this
         }).on('g:resultClicked', this._gotoUser, this);
 
+        if (getCurrentUser() && getCurrentUser().get('admin')) {
+            var userCountpromise = UserCollection.getUsersCount()
```
brianhelba: Use `const`, and name this variable with proper camelCase (as `userCountPromise`).

clients/web/src/views/body/UsersView.js
```diff
@@ -60,14 +59,27 @@ var UsersView = View.extend({
...
+                .done((resp) => {
```
brianhelba: I'd name this parameter `count`, rather than `resp`, since it's just an integer count, not a response body.

clients/web/src/collections/UserCollection.js
```diff
@@ -8,6 +9,15 @@ var UserCollection = Collection.extend({
     // Override default sort field
     sortField: 'lastName',
     secondarySortField: 'firstName'
+}, {
+    getUsersCount: function () {
```
brianhelba: How about naming this `getTotalCount` instead? We already know it's for users, since it's attached to the `UserCollection`.

mgrauer:
@Pierre-Assemat Could you add a screenshot to the PR of how the client side looks?

clients/web/src/templates/body/userList.pug
zachmullen: handy pug notation:

    = getUsersCount

clients/web/src/templates/body/userList.pug
zachmullen: We typically use `icon-user`.

Pierre-Assemat:
This is with the icon-user and without word to announce it.
I thought put some border but without to much things it could be lighter
Maybe if it was on the right it should be better, what do you think ? 

screenshot from 2017-08-08 08-54-32

mgrauer:
You can look at how [permission flag counts](https://github.com/girder/girder/pull/1984) were displayed, this might give you some other ideas, and would help us have a consistent look and feel for displaying counts next to icons.

I think the count would make more sense displayed next to (perhaps even to the left of) the user paging buttons, since both of these are related to the count of users.  Other people may have better or more specific thoughts, I'm fine with anything sensible.

Pierre-Assemat:
So is it better like this ? Where the number of users is always visible by an admin user 

screenshot from 2017-08-08 10-11-33

clients/web/src/templates/body/userList.pug
Pierre-Assemat: I think if we do that, it will be to heavy on the page, with the sort by, amd the create user button it's already to much,
What do you think about that ? @zachmullen  @mgrauer

mgrauer:
@brianhelba didn't want to display the user count in the GlobalNavView for reasons explained in this [comment](https://github.com/girder/girder/issues/2251#issuecomment-320755155).

Sorry @Pierre-Assemat , I wasn't clear when I said "to the left".  I meant to the left of the paging widgets, but still within the `g-user-list-header` , not within the GlobalNavView, but thank you for trying to implement that.

Also, I was just on a video call and saw an example of how to display user counts, this might be another interesting example.

screen shot 2017-08-08 at 10 43 17 am

girder/api/v1/user.py
brianhelba: @Pierre-Assemat As a reminder, this issue should also be fixed at some point.

Pierre-Assemat:
You mean like this ? Need more space between the user count and the page preview but I just try to know what your expectation

screenshot from 2017-08-08 12-45-52

jeffbaumes:
Is it just my eyes or is the "2" in italics (maybe incorrectly inside the icon's `i` tag)? I don't feel it should be.

clients/web/src/templates/body/userList.pug
```diff
@@ -7,6 +7,9 @@
     button.g-user-create-button.btn.btn-sm.btn-default
       i.icon-plus-squared
       |  Create user
+    i.icon-user
+      span.badge
```
zachmullen: @jeffbaumes noticed that this is in italics -- that's because the text is inside the `i` tag, it should be a sibling instead.

mgrauer:
Now that I see it, I'd prefer it on the right side of the paging buttons.  The paging buttons look better IMO lined up along the left edge with the users list.

Pierre-Assemat:
I like this one, what do you think ?

screenshot from 2017-08-08 13-34-44

mgrauer:
I like the "bubble" display of the count.  Can you

 * move the count/user icon to the right of the paging buttons
 * remove the "italic" font from the count (see Zach's comment above) for details
 * see what the bubble looks like when you have 50 or 500 as the number?

Pierre-Assemat:
We have the good one !! ;)

screenshot from 2017-08-08 14-01-20

Pierre-Assemat:
What about this one ? 

screenshot from 2017-08-08 14-30-40

zachmullen:
I hate to belabor a minor style issue, but we do have some prior art that we may want to use in this case for the sake of consistency:

screen shot 2017-08-08 at 2 48 54 pm

These are found in the hierarchy widget in the top bar.

mgrauer:
I agree @zachmullen, we should have consistency in style.

I propose that the location of the current user icon + count bubble stays where it is (to the immediate right of the paging widgets), but the style is copied from the existing folder icons, including the gray color of the icons (currently the user icon is black).

Sorry for sending you in the wrong direction @Pierre-Assemat .

Pierre-Assemat:
So I try to respect the same style
What do you think about that :

screenshot from 2017-08-08 21-28-25

It's the same blue than the other uses on the page

Pierre-Assemat:
Or maybe more discreet : 

screenshot from 2017-08-09 08-32-47

mgrauer:
Look here for specific examples, it would be good to use the same styles and structures.  It is better for consistency in the look and feel of the UI, and also from a maintenance perspective in that we are doing things the same way in the code in multiple places, rather than doing something different in every situation.  Ideally we would be able to reuse code rather than copying it, but I think copying is fine in this case so as to avoid any refactoring and to keep this PR to a tight scope.

https://github.com/girder/girder/blob/f51a00c3c9afd45fb4b11e4a7471fbfd52d6cb0e/clients/web/src/templates/widgets/hierarchyBreadcrumb.pug#L24

https://github.com/girder/girder/blob/6b185d945432f4b3000359cca4fa1ba5d56f6051/clients/web/src/stylesheets/widgets/hierarchyWidget.styl#L47

Pierre-Assemat:
We can't do exactly the same because the background color is white this time.
So we have 2 options : 
 - First use a colored background for the count-content like this : 

screenshot from 2017-08-09 11-22-16


- Or use a white background but add a border like this : 

screenshot from 2017-08-09 11-24-28

I think the first is better and fit more in the context, what do you think ?

clients/web/test/spec/userSpec.js
```diff
@@ -96,6 +96,16 @@ describe('Create an admin and non-admin user', function () {
         }, 'front page to display');
     });
 
+    it('check the user count with an admin user', function () {
+        girderTest.login('admin', 'Admin', 'Admin', 'adminpassword!')();
+        girderTest.goToUsersPage()();
+        runs(function () {
+            expect($('span.g-user-count')).toBeDefined();
```
brianhelba: This assertion will always pass, because even empty jQuery selectors are still truthy objects.

The next line will fail if the element doesn't exist, so you can remove this line entirely.

clients/web/test/spec/userSpec.js
```diff
@@ -96,6 +96,16 @@ describe('Create an admin and non-admin user', function () {
         }, 'front page to display');
     });
 
+    it('check the user count with an admin user', function () {
+        girderTest.login('admin', 'Admin', 'Admin', 'adminpassword!')();
+        girderTest.goToUsersPage()();
+        runs(function () {
```
brianhelba: You can move this entire block (really, just the assertion on line 104) up to line 93. There's no reason for another whole test spec.

clients/web/src/views/body/UsersView.js
```diff
@@ -60,14 +59,27 @@ var UsersView = View.extend({
...
+        $.when(...promiseArray)
+            .done(() => {
+                this.render();
+            });
     },
 
     render: function () {
         this.$el.html(UserListTemplate({
             users: this.collection.toArray(),
             getCurrentUser: getCurrentUser,
```
brianhelba: It'd be cleaner if we made this template just take the `currentUser`, instead of a function that has to be evaluated:
```
    currentUser: getCurrentUser(),
```

clients/web/test/spec/userSpec.js
Pierre-Assemat: I wondered if it's better to have separate tests, than put this in the line 93 because it's two different tests

clients/web/src/templates/body/userList.pug
```diff
@@ -1,5 +1,10 @@
 .g-user-list-header
   .g-user-pagination
+  if getCurrentUser() && getCurrentUser().get('admin')
```
brianhelba: Assuming that the template takes a `currentUser` parameter, you can [define a variable](https://pugjs.org/language/interpolation.html) at the top of the template:
```
- const isAdmin = currentUser && currentUser.get('admin');
```
Then all your `if` blocks can use it:
```
  if isAdmin
```

clients/web/src/stylesheets/body/userList.styl
```diff
@@ -58,3 +58,22 @@
     input[type="text"]
       width 170px
       margin-right 13px
+
+  .g-user-count-container
+    display inline-block
+    color #888
+    font-size 16px
+
+    i.icon-user
```
brianhelba: I wouldn't apply the `.icon-user` selector here, so the position doesn't break if we ever change the specific icon that's used. The fact that it's an `i` element inside a `.g-user-count-container` is specific enough.

clients/web/test/spec/userSpec.js
brianhelba: That's reasonable. You can still reduce some redundancy though, if you let this spec run first and:
* login
* go to users page
* check that the count is correct

Then have the `'check the user count with an admin user'` spec, which will:
* logout
* wait for the front page to display

You do have some state carrying over between specs (specifically, the fact that there's a logged in user, navigated to the users page), which isn't ideal, but is more than good enough for Girder front-end tests.

clients/web/src/views/body/UsersView.js
```diff
@@ -60,14 +59,27 @@ var UsersView = View.extend({
...
+            .done(() => {
+                this.render();
+            });
     },
 
     render: function () {
         this.$el.html(UserListTemplate({
             users: this.collection.toArray(),
-            getCurrentUser: getCurrentUser,
+            currentUser: getCurrentUser,
```
brianhelba: This is a function call.

Pierre-Assemat: The Ci test failed because of that : 

screenshot from 2017-08-10 18-08-05

Delete this line break the event of the sort widget button, It's working if I let them, but I think it's not clean at all so I wait you advice on this point.

brianhelba: The issue appears to be that future collection `g:changed` events triggered by fetches initiated by the `SortCollectionWidget` are no longer triggering `UsersView.render` (since we use promises, rather than event binding to trigger the initial render).

I see three possible ways to fix this:
1. Bind collection `g:changed` events to `UsersView.render` at the very end of `UsersView.render`, so additional events will cause additional renders, but the first event will not cause a render.
2. Make `SortCollectionWidget` trigger an event on itself when it's clicked. It may be cleaner to have the `SortCollectionWidget` not actually initiate the `Collection.fetch` at all, to prevent it mutating the state of the collection that it's passed. Alternatively, it could initiate the `fetch`, but pass the `fetch`'s promise as an event parameter.
3. Refactor `UsersView` to render its header and its list independently. This is probably the "best" change, as list updates shouldn't need to re-render the header, but it's also the most invasive, as it will require creating several additional templates and possibly additional views.

Regardless of the option chosen, most of option 2 should be implemented in a separate PR.

@girder/developers Thoughts?

Pierre-Assemat: I tried to implement the first one, it's working but if we are using the sort button a little bit too much : there is a bug, and we feel like it take lot of time to respond...

I think I will end this PR with the first solution and we could open a new PR to change that with the best solution : (2) or (3).

clients/web/src/collections/UserCollection.js
```diff
@@ -8,6 +9,15 @@ var UserCollection = Collection.extend({
     // Override default sort field
     sortField: 'lastName',
     secondarySortField: 'firstName'
+}, {
+    getTotalCount: function () {
+        return restRequest({
+            path: 'user/details'
```
brianhelba: In light of https://github.com/girder/girder/pull/2273 , can you make this parameter be `url: ` instead of `path: `?

Pierre-Assemat: Of course !!

clients/web/src/views/body/UsersView.js
```diff
@@ -81,6 +94,10 @@ var UsersView = View.extend({
             this.createUserDialog();
         }
 
+        this.collection.bind('g:changed', () => {
```
brianhelba: This should be:
```
this.listenTo(this.collection, 'g:changed', () => {
```
and moved up to line 76.

clients/web/src/collections/UserCollection.js
Pierre-Assemat: But it cause 4 tests failure, and the user page doesn't load anymore

brianhelba: Ah, then either rebase this branch onto `origin/master` or merge `origin/master` into this branch.

mgrauer:
@Pierre-Assemat Is this ready to be reviewed again or do you have more work to do here?

Pierre-Assemat:
I think it's ready ! :tada:

brianhelba:
Reviewing this again now.

brianhelba:
@Pierre-Assemat Can you edit the description at the top of this PR to give more details about what it does, and to include a screenshot of the final UI?

clients/web/src/stylesheets/body/userList.styl
```diff
@@ -58,3 +58,22 @@
...
+    i
+      position relative
+      top 2px
+
+    .g-user-count
+      position relative
+      background-color #eee
+      border-radius 3px
+      color rgba(0, 0, 0, 0.75)
+      bottom -7px
```
brianhelba: Negative positioning always makes me a bit nervous. What you could do instead is:
* Make the parent element (`.g-user-count-container`) use `position relative`. That doesn't change anything directly, but [makes it considered to be a "positioned element"](https://developer.mozilla.org/en-US/docs/Web/CSS/position#Types_of_positioning).
* Remove the `position relative`,  `bottom -7px` and `right 7px` rules from `.g-user-count`.
* Add `position absolute`, `left 65%`, and `top 50%` rules to `.g-user-count`. Tweak the percentages to your liking.

This will result in `.g-user-count-container` having a more tightly constrained size, which will be exactly the same as the size of the `i` element within. The `.g-user-count` within will not longer affect the size of the `.g-user-count-container` in any way, but instead will just be positioned as a free element within that.

clients/web/src/templates/body/userList.pug
```diff
@@ -1,9 +1,15 @@
+- const isAdmin = currentUser && currentUser.get('admin');
 .g-user-list-header
   .g-user-pagination
+  if isAdmin
+    .g-user-count-container
+      i.icon-user
+      span.g-user-count
```
brianhelba: This should be renamed to `usersCount` too.

clients/web/src/views/body/UsersView.js
```diff
@@ -60,14 +60,30 @@ var UsersView = View.extend({
...
+                    this.usersCount = count;
+                });
+            promiseArray.push(userCountPromise);
+        }
         this.register = settings.dialog === 'register' && getCurrentUser() &&
                         getCurrentUser().get('admin');
+
+        $.when(...promiseArray)
+            .done(() => {
+                this.listenTo(this.collection, 'g:changed', () => {
```
brianhelba: Also, this could be simplified as:
```
this.listenTo(this.collection, 'g:changed', this.render);
```

clients/web/src/stylesheets/body/userList.styl
Pierre-Assemat: It's working perfectly, thanks !

clients/web/src/views/body/UsersView.js
```diff
@@ -72,9 +72,7 @@ var UsersView = View.extend({
 
         $.when(...promiseArray)
             .done(() => {
-                this.listenTo(this.collection, 'g:changed', () => {
-                    this.render();
-                });
+                this.listenTo(this.collection, 'g:changed', this.render());
```
brianhelba: You don't want to call `render` here, just pass a reference to the function.

brianhelba:
@Pierre-Assemat  I see one minor bug with the `listenTo`, then I think this should be finished.

mgrauer:
Nice work @Pierre-Assemat !  👷 

Thanks @brianhelba and @zachmullen for your excellent QA.

Issue #2258, opened 2017-08-04
Make item_tasks to support output directory parameter in Slicer CLIs


cdeepakroy:
I was trying to ingest a Slicer CLI called [SegmentVesselsCNNSeeds](https://github.com/KitwareMedical/ITKTubeTK/blob/master/Applications/SegmentVesselsCNNSeeds/SegmentVesselsCNNSeeds.xml) from TubeTK that does vessel segmentation using deep learning and outputs a bunch of files to a directory and it throws the following error in the jobs page:

```
Using TensorFlow backend.
No handlers could be found for logger "ctk_cli.module"
HTTP push failed (http://deep-precision-7910:8080/api/v1/folder/5984ab742205464a552cd4b9/item_task_slicer_cli_xml). Response: {"message": "Output parameter type directory is currently not supported.", "type": "validation"}
HTTPError: 400 Client Error: Bad Request for url: http://deep-precision-7910:8080/api/v1/folder/5984ab742205464a552cd4b9/item_task_slicer_cli_xml?image=kitwaremedical%2Ftubetk-applications&args=%5B%22SegmentVesselsCNNSeeds%22%5D&pullImage=False
  File "/home/girder/.virtualenvs/girder_env/local/lib/python2.7/site-packages/celery/app/trace.py", line 374, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/girder/.virtualenvs/girder_env/local/lib/python2.7/site-packages/celery/app/trace.py", line 629, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/girder/Libs/girder_worker/girder_worker/tasks.py", line 19, in run
    return core.run(*pargs, **kwargs)
  File "/home/girder/Libs/girder_worker/girder_worker/core/utils.py", line 123, in wrapped
    return fn(*args, **kwargs)
  File "/home/girder/Libs/girder_worker/girder_worker/core/__init__.py", line 203, in run
    io.push(data, outputs[name], **dict({'task_output': task_output}, **kwargs))
  File "/home/girder/Libs/girder_worker/girder_worker/core/io/__init__.py", line 150, in push
    return _push_map[mode](data, spec, **kwargs)
  File "/home/girder/Libs/girder_worker/girder_worker/core/io/http.py", line 200, in push
    request.raise_for_status()
  File "/home/girder/.virtualenvs/girder_env/local/lib/python2.7/site-packages/requests/models.py", line 935, in raise_for_status
[...]

PR #2265, opened 2017-08-07
Add 'new-directory' output specification to item_tasks

sutartmelson:
Client-side fix to #2258

### DIFF
--- a/plugins/item_tasks/server/rest.py
+++ b/plugins/item_tasks/server/rest.py
@@ -174,23 +174,18 @@ class ItemTask(Resource):
         """
 
         # Find the corresponding output specification for the given outputID
-        output = None
-        for out in outputSpec:
-            if outputId == out['id']:
-                output = out
-                break
-
-        # If a corresponding output is found, check if its parent type is valid
-        if output:
-            if output['type'] == 'new-file' and parentType not in {'item', 'folder'}:
-                return False
-            elif output['type'] == 'new-folder' and\
-                    parentType not in {'folder', 'user', 'collection'}:
-                return False
-            else:
-                return True
+        for output in outputSpec:
+            if outputId == output['id']:
+                # If a corresponding output is found, check if its parent type is valid
+                if output['type'] == 'new-file' and parentType not in {'item', 'folder'}:
+                    return False
+                elif output['type'] == 'new-folder' and\
+                        parentType not in {'folder', 'user', 'collection'}:
+                    return False
+                else:
+                    return True
         else:
-            raise ValidationException('Invlaid outpud id: %s.' % outputId)
+            raise ValidationException('Invalid output id: %s.' % outputId)
 
     @access.user(scope=constants.TOKEN_SCOPE_EXECUTE_TASK)
     @filtermodel(model='job', plugin='jobs')
--- a/plugins/item_tasks/server/rest.py
+++ b/plugins/item_tasks/server/rest.py
@@ -139,7 +139,7 @@ class ItemTask(Resource):
 
         return transformed
 
-    def _transformOutputs(self, outputs, token, job):
+    def _transformOutputs(self, outputs, token, job, task):
         """
         Validates and sanitizes the output bindings. If they are Girder outputs, adds
         the necessary token info. If the token does not allow DATA_WRITE, or if the user
@@ -150,7 +150,7 @@ class ItemTask(Resource):
             if v['mode'] == 'girder':
                 ensureTokenScopes(token, TokenScope.DATA_WRITE)
                 ptype = v.get('parent_type', 'folder')
-                if ptype not in {'item', 'folder'}:
+                if not self._validateOutputParentType(k, ptype, task['outputs']):
                     raise ValidationException('Invalid output parent type: %s.' % ptype)
 
                 parent = self.model(ptype).load(
@@ -167,6 +167,29 @@ class ItemTask(Resource):
                 raise ValidationException('Invalid output mode: %s.' % v['mode'])
 
         return transformed
+
+    def _validateOutputParentType(self, outputId, parentType, outputSpec):
+        """
+        Checks if the output parent type is compatible with the output type.
+        """
+
+        # Find the corresponding output specification for the given outputID
+        output = None
+        for out in outputSpec:
+            if outputId == out['id']:
+                output = out
+
+        # If a corresponding output is found, check if its parent type is valid
+        if output:
+            if output['type'] == 'new-file' and parentType not in {'item', 'folder'}:
+                return False
+            elif output['type'] == 'new-folder' and\
+                    parentType not in {'folder', 'user', 'collection'}:
+                return False
+            else:
+                return True
+        else:
+            return False
 
     @access.user(scope=constants.TOKEN_SCOPE_EXECUTE_TASK)
     @filtermodel(model='job', plugin='jobs')
@@ -211,7 +234,7 @@ class ItemTask(Resource):
             'kwargs': {
                 'task': task,
                 'inputs': self._transformInputs(inputs, token),
-                'outputs': self._transformOutputs(outputs, token, job),
+                'outputs': self._transformOutputs(outputs, token, job, task),
                 'validate': False,
                 'auto_convert': False,
                 'cleanup': True
--- a/plugins/item_tasks/server/rest.py
+++ b/plugins/item_tasks/server/rest.py
@@ -178,6 +178,7 @@ class ItemTask(Resource):
         for out in outputSpec:
             if outputId == out['id']:
                 output = out
+                break
 
         # If a corresponding output is found, check if its parent type is valid
         if output:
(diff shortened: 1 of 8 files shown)


plugins/item_tasks/web_client/models/WidgetModel.js
```diff
@@ -302,6 +305,8 @@ var WidgetModel = Backbone.Model.extend({
             return 'Value must be a folder';
         } else if (this.get('type') === 'new-file' && type !== 'folder') {
             return 'Value must be a folder';
+        } else if (this.get('type') === 'new-directory' && type !== 'folder') {
```
brianhelba: What if we want to create the new folder at the top level, directly inside a collection (or a user's home directory).

plugins/item_tasks/web_client/models/WidgetModel.js
```diff
@@ -42,6 +42,9 @@ import tinycolor from 'tinycolor2';
  *   * new-file:
  *      an output file (contains an existing folder id and a
  *      name that will be used for the new item.
+ *   * new-directory:
```
brianhelba: Do we really want to call this "directory"? Everything else in Girder uses the terminology of "folders".

zachmullen: +1 for "folder" instead.

plugins/item_tasks/web_client/models/WidgetModel.js
```diff
@@ -302,6 +305,8 @@ var WidgetModel = Backbone.Model.extend({
             return 'Value must be a folder';
         } else if (this.get('type') === 'new-file' && type !== 'folder') {
             return 'Value must be a folder';
+        } else if (this.get('type') === 'new-folder' && !_.contains(['folder', 'collection', 'user'], type)) {
+            return 'Value must be a folder';
```
brianhelba: This error message is not quite accurate any more.

plugins/item_tasks/web_client/views/ControlWidget.js
```diff
@@ -191,14 +194,38 @@ var ControlWidget = View.extend({
...
+        if (type === 'new-folder') {
+            title = 'Create a new folder';
+            help = 'Browse to a path, enter a name, then click "Save"';
+            input = {
+                label: 'Name',
+                placeholder: 'Choose a name for the new folder',
+                validate: (val) => {
+                    // validation on the "new item name"
+                    if (!val) {
+                        return 'Please provide an folder name.';
```
brianhelba: "provide **a** folder name"

plugins/item_tasks/web_client/views/ControlWidget.js
```diff
@@ -191,14 +194,38 @@ var ControlWidget = View.extend({
...
+                    }
+                },
+                default: this.model.get('fileName')
+            };
+            // validation on the parent model
+            validate = (model) => {
+                var type = model.get('_modelType');
+                if (!_.contains(['folder', 'collection', 'user'], type)) {
+                    return 'Invalid parent type, please choose a folder.';
```
brianhelba: Not quite accurate.

plugins/item_tasks/web_client/views/ControlWidget.js
```diff
@@ -191,14 +194,38 @@ var ControlWidget = View.extend({
...
+            input = {
+                label: 'Name',
+                placeholder: 'Choose a name for the new folder',
+                validate: (val) => {
+                    // validation on the "new item name"
+                    if (!val) {
+                        return 'Please provide an folder name.';
+                    }
+                },
+                default: this.model.get('fileName')
```
brianhelba: I'm not sure: do we expect this value to be set from the task spec?

sutartmelson:
@brianhelba My initial thought was to wait for the server side to add tests.. but I now realize I need not wait for that. I'll add tests today.

plugins/item_tasks/web_client/views/ControlWidget.js
sutartmelson: I'm not sure. Is this task spec outlined anywhere? I tried looking at girder_worker but didn't find comprehensive format specifications.

jbeezley: The spec has largely grown organically.  The most comprehensive documentation is here: https://github.com/girder/girder/blob/master/plugins/item_tasks/web_client/models/WidgetModel.js#L5

zachmullen:
I just tested this out, and it's working well. One thing that isn't working yet is populating the widget based on a previous job, which is done [here](https://github.com/girder/girder/blob/master/plugins/item_tasks/web_client/views/TaskRunView.js#L61-L80).

plugins/item_tasks/web_client/collections/WidgetCollection.js
```diff
@@ -22,6 +22,10 @@ var WidgetCollection = Backbone.Collection.extend({
                     params[m.id + '_girderFolderId'] = m.value().get('folderId');
                     params[m.id + '_name'] = m.value().get('name');
                     break;
+                case 'new-folder':
```
sutartmelson: I more or less blindly edited this file, so I'm not sure if this is correct or where this is actually used.

zachmullen:
Hm, that last commit isn't working for me, the name is not populated from the previous job in the directory output field.

sutartmelson:
@zachmullen  Apparently it's `parent_type` and not `parentType`. I don't really understand how I got it to work earlier then if that's wrong. Regardless, it should work now.

zachmullen:
🎉  this is now working on the client side. One last thing I'd like to fix on this branch: the server side still complains with `Error: Invalid output parent type: user.`. Could we fix the validation to allow parent_type to be user or collection in the case of a new-folder output? This would be in `_transformOutputs` I think.

plugins/item_tasks/server/rest.py
```diff
@@ -168,6 +168,29 @@ def _transformOutputs(self, outputs, token, job):
...
+    def _validateOutputParentType(self, outputId, parentType, outputSpec):
+        """
+        Checks if the output parent type is compatible with the output type.
+        """
+
+        # Find the corresponding output specification for the given outputID
+        output = None
+        for out in outputSpec:
+            if outputId == out['id']:
+                output = out
```
zachmullen: and `break`

plugins/item_tasks/server/rest.py
```diff
@@ -168,6 +168,29 @@ def _transformOutputs(self, outputs, token, job):
...
+        if output:
+            if output['type'] == 'new-file' and parentType not in {'item', 'folder'}:
+                return False
+            elif output['type'] == 'new-folder' and\
+                    parentType not in {'folder', 'user', 'collection'}:
+                return False
+            else:
+                return True
+        else:
+            return False
```
zachmullen: This should probably raise its own `ValidationException` since it's not really an invalid parent type, but actually an invalid output ID.

zachmullen:
This works for me now, thanks!

@brianhelba could you give a final review on this?

plugins/item_tasks/server/rest.py
```diff
@@ -168,6 +168,30 @@ def _transformOutputs(self, outputs, token, job):
 
         return transformed
 
+    def _validateOutputParentType(self, outputId, parentType, outputSpec):
+        """
+        Checks if the output parent type is compatible with the output type.
+        """
+
+        # Find the corresponding output specification for the given outputID
+        output = None
```
brianhelba: Python loop iteration variables persist in the scope outside, so you could use `out` in place of `output` below (or vice versa)

plugins/item_tasks/server/rest.py
```diff
@@ -168,6 +168,30 @@ def _transformOutputs(self, outputs, token, job):
...
+        for out in outputSpec:
+            if outputId == out['id']:
+                output = out
+                break
+
+        # If a corresponding output is found, check if its parent type is valid
+        if output:
```
brianhelba: Python `for` loops take an optional `else:` clause, which only runs if the loop iterates to completion without any `break` being triggered. You could use this to eliminate both sides of the `if output:` conditional, moving the contained logic up one level.

mgrauer: The for/else in Python is frequently misunderstood.  I'd avoid using it if possible.

https://aroberge.blogspot.com/2015/04/loopy-else-in-python.html

zachmullen: I actually really like the for/else language feature, but agree that it is confusing because it's unfamiliar since it's not present in other programming languages. That said, it's a language construct that exists precisely for these sorts of cases, and it's decidedly more elegant than the more C-like construct that is currently written.

I agree with the article that the main source of confusion is the term `else` as opposed to a more explicit one like `nobreak`.

TLDR: While I personally would have written this using for/else, I do not feel that it needs to be changed since it works as written since there are tradeoffs.

plugins/item_tasks/server/rest.py
```diff
@@ -168,6 +168,30 @@ def _transformOutputs(self, outputs, token, job):
...
+            if output['type'] == 'new-file' and parentType not in {'item', 'folder'}:
+                return False
+            elif output['type'] == 'new-folder' and\
+                    parentType not in {'folder', 'user', 'collection'}:
+                return False
+            else:
+                return True
+        else:
+            raise ValidationException('Invlaid outpud id: %s.' % outputId)
```
mgrauer: Invalid output

PR #2267, opened 2017-08-07
Add Terraform assetstore module

danlamanna:
This is a reusable Terraform component for creating an S3 bucket
which is compliant as a Girder assetstore.

### DIFF
--- a/devops/terraform/README.rst
+++ b/devops/terraform/README.rst
@@ -1,0 +1,7 @@
+Terraform
+-----------
+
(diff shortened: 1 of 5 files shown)


danlamanna:
To summarize a conversation @kotfic and I had:

It turns out only one role can be attached to an instance profile, and many policies can be attached to a role. Since users may want to have their own role, this module will only provide the S3 bucket and a single policy object which is the set of actions that Girder needs to utilize an S3 assetstore. Callers will be responsible for creating their own role/instance profile and attaching things correctly.

danlamanna:
@kotfic This should be functioning how we discussed now.

PR #2268, opened 2017-08-07
Expose "verify" parameter of requests

zachmullen:
This is the easiest way to allow control of SSL verification, including
pointing to alternative CA bundles

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -187,7 +187,7 @@ class GirderClient(object):
             return "https"
 
     def __init__(self, host=None, port=None, apiRoot=None, scheme=None, apiUrl=None,
-                 cacheSettings=None, progressReporterCls=None):
+                 cacheSettings=None, progressReporterCls=None, verify=None):
         """
         Construct a new GirderClient object, given a host name and port number,
         as well as a username and password which will be used in all requests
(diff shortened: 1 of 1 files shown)


danlamanna:
I needed this the other day and ended up sidestepping the issue by using
GC through the CherryPy Girder instead of the NGINX proxy which was
using a self signed cert.

As a result I started thinking about the use of the requests session
object and I think it makes sense to automatically create one and use it
for all HTTP(S) interactions and then leave it public for the caller to
manipulate. (This would probably be best as a BaseUrlSession:
http://toolbelt.readthedocs.io/en/latest/sessions.html#requests_toolbelt.sessions.BaseUrlSession).

The inevitable issue with the current approach is that every requests
flag (http://docs.python-requests.org/en/master/api/#requests.Session)
we want to add requires a change be made to Girder client (either to
__init__, or sendRestRequest, or get/put/post etc). If we expose the
session object the user can just use the raw requests API.

#whatdowewant #girder3 :P

zachmullen:
@danlamanna does that mean you object to this change?

danlamanna:
No, sorry. I didn't start with the "uhh this is kinda related so I'll
just brain dump some thoughts here".

zachmullen:
I'm considering closing this and just using the session to accomplish what I need to do with the client.

PR #2270, opened 2017-08-08
Item tasks from girder_worker describe decorators

jbeezley:
This implements the girder side of https://github.com/girder/girder_worker/pull/162.  There a few things that still need to happen before this can go in:

* [x] Add client side tests
* [x] Merge https://github.com/girder/girder_worker/pull/172
* [x] Finish and merge https://github.com/girder/girder_worker/pull/119 (otherwise item_tasks will be a python 2 only plugin)
* [x] Cut release for girder_worker so we can add a dependency on a released version of girder_worker for item_tasks

In addition, I'm not a big fan of how some of the functions/modules/endpoints etc. are named in this PR.  For one, what should these tasks be called?  There are json tasks and slicer tasks now.  After talking with @kotfic, I named these "celery tasks".  There are also "girder worker extension", which for the moment are entrypoint names bound to `girder_worker_plugins`, but that probably isn't obvious.  This will all certainly require a documentation overhaul.  I would appreciate any feedback on how to rename things or move them around to be more clear.


zachmullen:
This is part of the larger effort @kotfic @jbeezley and I are discussing / working on. Will wait for now.

jbeezley:
I have rebased this branch and ported all the changes coming from girder_worker.  The one remaining issue here has to do with the fact that there is no pypi release of girder_worker.  I solved the issue via the last commit on this branch for all but the py3 integration and packaging tests by installing from github before installing the requirements file.  I could do the same for the girder-dev-environment ansible config and hacking the packaging tests, but we should probably solve this issue in general by making a release of girder-worker.

jbeezley:
I believe I finally got the tests working on this branch, so I've taken off the WIP.  It would be a good time to have others attempt to look at it.  As I've noted before, this doesn't yet support Girder I/O types.  I'll start looking into how to support them here or in a future PR, whichever is most appropriate.

zachmullen:
There are some unresolved issues on the g_w side of this. @kotfic and @jbeezley will take a look and get back on this PR.

Issue #2030, opened 2017-05-25
Use promises more consistently and thoroughly


brianhelba:
This is a proposal to standardize around jQuery 3 Promises as the preferred way to handle asynchronous programming in Girder's web code.

At the moment, there are several issues:
* Girder contains mostly jQuery 2 Promises, with a few ES 6 Promises.
* Many asynchronous methods do not return Promises.
* Even where promises are available, many places use transient event binding instead.
* The jQuery promise API is used inconsistently and with deprecated methods.

To resolve this, it is proposed that we:
* [ ] Eliminate any ES 6 Promises
  * They are not [compatible with IE 11](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Browser_compatibility), which comprises [about 14% of desktop browsers in the US](http://gs.statcounter.com/browser-market-share/desktop/united-states-of-america) and [8.5% in Europe](http://gs.statcounter.com/browser-market-share/desktop/europe). More importantly, some of Girder's users in institutional environments (e.g. hospitals) are forced to use IE. While Girder doesn't explicitly support IE, we don't have to go out of our way to break it either.
  * Their performance (under V8) is currently not superior to third-party promise implementations, and [is often significantly worse](https://softwareengineering.stackexchange.com/a/279003).
  * Most of Girder, and web code in general, uses jQuery for Promises, so there would be a large overhead to consistently switch to ES6 Promises.
  * The alternative [Bluebird promise library](http://bluebirdjs.com/docs/features.html) is probably overkill for now.
    * Its killer `.promisifyAll` feature is directly intended for Node.js.
    * Its map / reduce functions are also probably most useful in a server environment, where lots of similar asynchronous calls may be made in parallel.
    * Its error catching and stack saving features are less important for a front-end app, where uncaught errors are usually not as devastating (though perhaps somewhat important to log).
    * Its `.delay` function is effectively a nice promisified wrapper around `window.setTimeout`. Perhaps we should replicate this as a Girder utility method.

* [ ] Eliminate and stop using `.error` to handle jqXHR Promises.
  * This method has been deprecated for 6 years, [since jQuery 1.8](https://bugs.jquery.com/ticket/9399), and is removed in jQuery 3.
  * `.fail` should be used instead
  * [ ] Since Girder still uses `.error` so prevalently, we should patch it to print a warning when used (which will also allow it to work under jQuery 3).

* [ ] Upgrade to jQuery 3
[...]

PR #2273, opened 2017-08-09
Unify restRequest with Backbone.ajax and $.ajax

brianhelba:
Now, vanilla Backbone Models can use `Backbone.sync`-calling methods with the Girder API. The call chain will typically be:
`Backbone.sync` -> `Backbone.ajax` -> `restRequest` -> `Backbone.$.ajax` -> `$.ajax`

`restRequest` will now expect a strict superset (just adding `girderToken`) of the parameters that `$.ajax` expects:
* Existing calls to `restRequest` have been updated to use the `url` and `method` arguments, as these are preferred by `$.ajax`, in place of `path` and `type` respectively. Use of the `url` argument is now deprecated, and will print a warning. Use of the `type` argument is also deprecated, but will not print anything, because it may still be used by Backbone.
* The `dataType` argument to `restRequest` will no longer be set to `'json'` by default . Instead, the default behavior of `$.ajax` is allowed to occur, autodetecting the response type from MIME headers,
which Girder's API will set to `'application/json'` when appropriate.

The `mockRestRequest` and `unmockRestRequest` methods are now deprecated, as mocking code for tests should not contaminate the functionality within source files. The tests using these functions have been updated to use Jasmine spies.

This PR's commits are structured to have one logical change per commit, for ease of review.

### DIFF
--- a/clients/web/src/rest.js
+++ b/clients/web/src/rest.js
@@ -75,25 +75,31 @@ setStaticRoot(
 );
 
 /**
- * Make a request to the REST API. Bind a "done" handler to the return
- * value that will be called when the response is successful. To bind a
- * custom error handler, bind a "fail" handler to the return promise,
- * which will be executed in addition to the normal behavior of logging
- * the error to the console. To override the default error handling
- * behavior, pass an "error" key in your opts object; this should be done
- * any time the server might throw an exception from validating user input,
- * e.g. logging in, registering, or generally filling out forms.
+ * Make a request to the REST API.
  *
- * @param {Object} opts Options for the request, most of which will be passed through to `$.ajax`.
- * @param {string} opts.url The resource path, e.g. "user/login"
+ * This is a wrapper around {@link http://api.jquery.com/jQuery.ajax/ $.ajax}, which also handles
+ * authentication and routing to the Girder API, and provides a default for error notification.
+ *
+ * Most users of this method should attach a "done" handler to the return value. In cases where the
+ * server is ordinarily expected to return a non-200 status (e.g. validating user input), the
+ * "error: null" argument should probably be provided, and errors should be processed via an
+ * attached "fail" handler.
+ *
+ * Before this function is called, the API root must be set (which typically happens automatically).
+ *
+ * @param {Object} opts Options for the request, most of which will be passed through to $.ajax.
+ * @param {string} opts.url The resource path, relative to the API root, without leading or trailing
+ *        slashes. e.g. "user/login"
  * @param {string} [opts.method='GET'] The HTTP method to invoke.
- * @param [opts.data] The query string or form parameter object.
+ * @param {string|Object} [opts.data] The query string or form parameter object.
  * @param {?Function} [opts.error] An error callback, as documented in
- *        `$.ajax <http://api.jquery.com/jQuery.ajax/>`_, or null. If not provided, this will have a
- *        default behavior of triggering a `'g:alert'` global event, with details of the error.
+ *        {@link http://api.jquery.com/jQuery.ajax/ $.ajax}, or null. If not provided, this will
+ *        have a default behavior of triggering a 'g:alert' global event, with details of the
+ *        error, and logging the error to the console. It is recommended that you do not ever pass
+ *        a non-null callback function, and handle errors via promise rejection handlers instead.
  * @param {string} [opts.girderToken] An alternative auth token to use for this request.
  * @returns {$.Promise} A jqXHR promise, which resolves and rejects
- *          `as documented by $.ajax<http://api.jquery.com/jQuery.ajax/#jqXHR>`_.
+ *          {@link http://api.jquery.com/jQuery.ajax/#jqXHR as documented by $.ajax}.
  */
 let restRequest = function (opts) {
     opts = opts || {};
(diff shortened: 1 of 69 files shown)


danlamanna:
+1

It's nice to have an interface closer to $.ajax and to configure
Backbone in a way that sync will correctly work within the context of
Girder. Do you have any thoughts about what follows this PR,
particularly with regards to Model.js and what the path forward is. For
example being able to rely on Backbone.Model.save instead of our own.

brianhelba:
@danlamanna Yes, my plan is to continue to move Girder towards upstream Backbone.

Models will probably be less controversial to unify on the base class, as we just need to ensure that all methods are triggering the correct native Backbone events. Relatedly, more work will also need to be done to move nearly all usages of `restRequest` out of Views and into Models or Collections, and ensure that these methods return Promises (testing this is going to require some updates to our test infrastructure, to better support actual unit tests). Finally, Views should be updated to rely on Promises for single async events, and native Backbone events on models otherwise.

Collections will likely require fewer lines of code to be refactored, but is going to require a very disruptive change to the `fetch` method. I don't have a well-formed plan on how to do this yet, but the least-painful way may be to make an alias to the existing `fetch`, called something like `load` (better names are welcome), find/replace all usages, and deprecate `fetch` until v3. Upon v3, we can remove Girder's implementation of `fetch`, and allow the native Backbone implementation to be exposed. Of course, we also need to do a thorough review of the native helper methods, to ensure they're doing what the Girder API expects. Again, real unit testing will be critical to this process.

Beyond https://github.com/girder/girder/issues/2030 , I haven't laid much of this out in a GH issue yet, partly because I haven't fully planned it out myself yet. If anyone wants to continue discussing this though, I'll make an issue, so this PR doesn't go too far off track.

clients/web/src/rest.js
```diff
@@ -84,19 +84,25 @@ setStaticRoot(
...
+ * @param {Object} opts Options for the request, most of which will be passed through to `$.ajax`.
+ * @param {string} opts.url The resource path, e.g. "user/login"
+ * @param {string} [opts.method='GET'] The HTTP method to invoke.
+ * @param [opts.data] The query string or form parameter object.
+ * @param {?Function} [opts.error] An error callback, as documented in
+ *        `$.ajax <http://api.jquery.com/jQuery.ajax/>`_, or null. If not provided, this will have a
+ *        default behavior of triggering a `'g:alert'` global event, with details of the error.
+ * @param {string} [opts.girderToken] An alternative auth token to use for this request.
+ * @returns {$.Promise} A jqXHR promise, which resolves and rejects
+ *          `as documented by $.ajax<http://api.jquery.com/jQuery.ajax/#jqXHR>`_.
```
jbeezley: I don't know of any way to parse restructured text with esdoc (or jsdoc); however, there are plugins that support markdown.  In this case, you can just use the [`@url`](http://usejsdoc.org/tags-inline-link.html) tag.

clients/web/src/rest.js
```diff
@@ -84,19 +84,25 @@ setStaticRoot(
...
  * e.g. logging in, registering, or generally filling out forms.
  *
- * @param path The resource path, e.g. "user/login"
- * @param data The form parameter object.
- * @param [type='GET'] The HTTP method to invoke.
- * @param [girderToken] An alternative auth token to use for this request.
+ * @param {Object} opts Options for the request, most of which will be passed through to `$.ajax`.
+ * @param {string} opts.url The resource path, e.g. "user/login"
+ * @param {string} [opts.method='GET'] The HTTP method to invoke.
+ * @param [opts.data] The query string or form parameter object.
```
jbeezley: Could add `{string|Object}` here?

brianhelba:
Improved docs, per feedback from @jbeezley .

brianhelba:
@girder/developers I'm reconsidering the rename from `path` -> `url` in this PR. While `url` is the option that `$.ajax` uses, that does not transform the URLs in any way. `restRequest` instead demands a "relative-ish" URL component, and unconditionally prepends the API root. The function documentation now clearly states the expected format of this option, but I'm not sure what is more "obvious" to a user transitioning from `$.ajax` to `restRequest`. I don't want to support both (unless it's via deprecation), as we should have *"one right way"*. Thoughts?

jbeezley:
I was going to bring that up, but after thinking about it, I decided what you've done is okay at least for now.  `$.ajax` supports both `path` and `url`, so I don't think that distinction is relevant.  The only reason I can see the custom behavior as being a problem is with an external application that combines different API servers.  Even then, you could easily get around it by wrapping our custom `Backbone.ajax`.

A plausible alternative would be to prefix path in the base `Model.url` and `Collection.url` methods, but I would want to do that as a breaking change.

brianhelba:
In response to some inquiries, let me elaborate a bit more on the rationale for migrating the parameter `path` -> `url`:

Lots of client code was abusing the fact that `restRequest` passed any unhandled / undocumented arguments through to `$.ajax`. It made sense to make it official that `restRequest` be a wrapper around `$.ajax` with a strict superset of parameters (to just handle Girder-specific authentication), because:
* `$.ajax` is already quite well known in the web development community, with lots of official and unofficial (StackOverflow, etc.) documentation. Adopting its API for `restRequest` (which is one of the most fundamental Girder-specific functions) should reduce barriers to entry for new developers.
* Backbone can use `restRequest` as the handler for `Backbone.ajax` without any compatibility concerns.
* Per the principle of least surprise, it did not make sense to expose another brand-new parameter (`path`) which serves essentially the same function as `url`. While there is a subtle difference in the way this parameter is handled by `restRequest` (as the API root is prepended), this is now well documented.
* Passing contradictory or redundant parameters (e.g. previously passing both `url` and `path`) can now more often be handled by `$.ajax`, leaving less custom logic in Girder.

PR #2274, opened 2017-08-09
Add support for using dogpile.cache within Girder

danlamanna:
First, a few things to note for [dogpile.cache](https://bitbucket.org/zzzeek/dogpile.cache):
- It's a small, simple library which is actively maintained
- It's created/maintained by the author of SQLAlchemy and Mako (among others).
- It aims to be glue code for caching backends, and nothing more
    - The author has explicitly stated that he does not want to maintain a large number of backends, and has instead designed the library in a way that can be very easily extended.

This PR does the following:
- Creates a custom backend for using `cherrypy.request` which is local to each individual request thread
- Creates 2 cache "regions", one for per-request caching and one for global caching
- Replaces some custom caching done in Girder with the provided decorators from dogpile
- Adds caching to the setting model as a proof of concept

This PR is WIP for a couple of reasons:
1) @manthey is planning on being the first to utilize this in order to port a caching mechanism from large image. I'd like to wait on merging this until the first users can see what's wrong with it.
2) Tests

### DIFF
--- a/girder/__init__.py
+++ b/girder/__init__.py
@@ -28,6 +28,7 @@ import traceback
 
 from girder.constants import LOG_ROOT, MAX_LOG_SIZE, LOG_BACKUP_COUNT, TerminalColor, VERSION
 from girder.utility import config, mkdir
+from girder.utility.cache import cache, requestCache
 
 __version__ = '2.3.0'
 __license__ = 'Apache 2.0'
@@ -258,6 +259,26 @@ def logprint(*args, **kwargs):
         _originalStdOut.write('%s\n' % data)
         _originalStdOut.flush()
 
+
+def _setupCache():
+    """
+    Setup caching based on configuration file.
+
+    If the cache is disabled, both the global and per-request cache are set
+    to the null backend, effectively making them a noop. This allows for
+    decorating functions with the cache region, despite it being unused.
+    """
+    curConfig = config.getConfig()
+
+    if not curConfig['cache']['enabled']:
+        cache.configure(backend='dogpile.cache.null')
+        requestCache.configure(backend='dogpile.cache.null')
+    else:
+        cache.configure_from_config(curConfig['cache'], 'cache.global.')
+        requestCache.configure_from_config(curConfig['cache'], 'cache.request.')
+
+
+_setupCache()
 
 # Expose common logging levels and colors as methods of logprint.
 logprint.info = functools.partial(logprint, level=logging.INFO, color='info')
--- a/girder/models/setting.py
+++ b/girder/models/setting.py
@@ -26,6 +26,7 @@ from ..constants import GIRDER_ROUTE_ID, GIRDER_STATIC_ROUTE_ID, SettingDefault,
 from .model_base import Model, ValidationException
 from girder import logprint
 from girder.utility import config, plugin_utilities, setting_utilities
+from girder.utility.cache import cache
 from girder.utility.model_importer import ModelImporter
 from bson.objectid import ObjectId
 
@@ -97,6 +98,10 @@ class Setting(Model):
 
         return doc
 
+    @cache.cache_on_arguments()
+    def _get(self, key):
+        return self.findOne({'key': key})
+
     def get(self, key, default='__default__'):
         """
         Retrieve a setting by its key.
@@ -106,7 +111,8 @@ class Setting(Model):
         :param default: If no such setting exists, returns this value instead.
         :returns: The value, or the default value if the key is not found.
         """
-        setting = self.findOne({'key': key})
+        setting = self._get(key)
+
         if setting is None:
             if default is '__default__':
                 default = self.getDefault(key)
@@ -133,7 +139,11 @@ class Setting(Model):
         else:
             setting['value'] = value
 
-        return self.save(setting)
+        setting = self.save(setting)
+
+        self._get.invalidate(self, key)
+
+        return setting
 
     def unset(self, key):
         """
@@ -143,6 +153,7 @@ class Setting(Model):
         :param key: The key identifying the setting to be removed.
         :type key: str
         """
+        self._get.invalidate(self, key)
         for setting in self.find({'key': key}):
             self.remove(setting)
 
--- a/girder/utility/cache.py
+++ b/girder/utility/cache.py
@@ -1,0 +1,28 @@
+import cherrypy
+from dogpile.cache import make_region, register_backend
+from dogpile.cache.backends.memory import MemoryBackend
+
+
+class CherrypyRequestBackend(MemoryBackend):
+    """
+    A memory backed cache for individual CherryPy requests.
+
+    This provides a cache backend for dogpile.cache which is designed
+    to work in a thread-safe manner using cherrypy.request, a thread local
+    storage that only lasts for the duration of a request.
+    """
+    def __init__(self, arguments):
+        pass
+
+    @property
+    def _cache(self):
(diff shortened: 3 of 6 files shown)


girder/__init__.py
```diff
@@ -259,6 +260,26 @@ def logprint(*args, **kwargs):
...
+
+    if not curConfig['cache']['enabled']:
+        cache.configure(backend='dogpile.cache.null')
+        requestCache.configure(backend='dogpile.cache.null')
+    else:
+        cache.configure_from_config(curConfig['cache'], 'cache.global.')
+        requestCache.configure_from_config(curConfig['cache'], 'cache.request.')
+
+
+_setupCache()
```
zachmullen: Is `girder.__init__` the right place to run this? Or would it be better to have it run inside the server configuration or something? I'm trying to think of whether users of the girder python library for things other than girder-server might want to use it without the need for caching.

girder/__init__.py
```diff
@@ -259,6 +260,26 @@ def logprint(*args, **kwargs):
...
+    to the null backend, effectively making them a noop. This allows for
+    decorating functions with the cache region, despite it being unused.
+    """
+    curConfig = config.getConfig()
+
+    if not curConfig['cache']['enabled']:
+        cache.configure(backend='dogpile.cache.null')
+        requestCache.configure(backend='dogpile.cache.null')
+    else:
+        cache.configure_from_config(curConfig['cache'], 'cache.global.')
```
zachmullen: Is there reason to have the word `cache` in the setting name? Seems redundant with it being in the `cache` section.

zachmullen: On second thought, I actually think it is a good idea, I withdraw my question.

girder/models/setting.py
```diff
@@ -133,7 +139,11 @@ def set(self, key, value):
         else:
             setting['value'] = value
 
-        return self.save(setting)
+        setting = self.save(setting)
+
+        self._get.invalidate(self, key)
```
zachmullen: Is there a way to modify an existing cache value, or is `invalidate` the only option?

girder/utility/cache.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    def __init__(self, arguments):
+        pass
+
+    @property
+    def _cache(self):
+        if not hasattr(cherrypy.request, '_girderCache'):
+            setattr(cherrypy.request, '_girderCache', {})
+
+        return cherrypy.request._girderCache
```
zachmullen: I'm guessing this doesn't enforce any eviction policy?

girder/models/setting.py
danlamanna: Assuming we want to proactively cache these, `set` is probably what we want.

manthey:
@danlamanna Are there any packages that add an LRU memory cache to dogpile.cache?  I see a pull request that was made on dogpile.cache (https://bitbucket.org/zzzeek/dogpile.cache/pull-requests/32/add-a-cachetools-lru-lfu-in-memory-backend/diff), but it was rejected.

Otherwise, we have to do something similar to what was done in the rejected pull request using cachetools, or write our own LRU backend.

danlamanna:
> Are there any packages that add an LRU memory cache to dogpile.cache?
Not that I'm aware of.

girder/__init__.py
danlamanna: > Is girder.__init__ the right place to run this?

I don't think it is. The more I work with Python the more I find myself having a strong dislike for import side effects.

girder/utility/cache.py
```diff
@@ -0,0 +1,28 @@
...
+
+    This provides a cache backend for dogpile.cache which is designed
+    to work in a thread-safe manner using cherrypy.request, a thread local
+    storage that only lasts for the duration of a request.
+    """
+    def __init__(self, arguments):
+        pass
+
+    @property
+    def _cache(self):
```
danlamanna: This method should probably check that `cherrypy.request` exists and raise an exception if the developer is trying to use the request cache outside of a request context.

setup.py
```diff
@@ -63,6 +63,7 @@ def run(self, *arg, **kw):
     'boto3',
     'CherryPy',
     'click',
+    'dogpile.cache',
```
manthey: Should this also be added to `requirements.txt`?  If so, then `requirements.txt` looks like it is missing `click`, `numpy`, and `ctk-cli` from older changes, too.

danlamanna:
@zachmullen @brianhelba This is rebased and ready for another pass.

zachmullen:
@danlamanna I think it would be a great exercise to port these tests to pytest and develop any required cache-related fixtures. If you have time, doing it on this PR would be great, but if not I am happy to take a crack at it on a follow-up branch.

danlamanna:
> I think it would be a great exercise to port these tests to pytest and develop any required cache-related fixtures.

I updated the test to use pytest, but didn't need any custom fixtures aside from extending the server fixture. I'm not entirely sure if it was the best approach, ptal.

girder/test/test_cache.py
```diff
@@ -0,0 +1,75 @@
...
+    _setupCache()
+
+    yield server
+
+
+def testCachesAreAlwaysConfigured():
+    from girder.utility.cache import cache, requestCache
+
+    assert cache.is_configured
+    assert requestCache.is_configured
```
zachmullen: I'd feel more comfortable if this was `assert cache.is_configured is True`; just using the truthiness value could allow for errors, for instance if `is_configured` were actually a function that is supposed to be called. (Even if it's not a function, it's best to maintain this precedent of explicitness.)

girder/test/test_cache.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def testSettingsCache(server):
```
zachmullen: This test does not appear to use the request thread cache or other server components, but relies on the server fixture. Does the caching behavior itself necessarily rely on the server?

danlamanna: No, it should be using the `db` fixture.

girder/test/test_cache.py
danlamanna: Done.

girder/__init__.py
```diff
@@ -262,6 +263,29 @@ def logprint(*args, **kwargs):
         _originalStdOut.flush()
 
 
+def _setupCache():
```
zachmullen: Is this function idempotent? It would be great if we could guarantee that it was, since it's called during server setup. See #2530

danlamanna: The global variables (cache and requestCache) are simple interfaces to the cache backends, the backends are completely replaced on each server startup.

danlamanna:
rebased and resolved merge conflicts.

brianhelba:
Can we name the new object `_cache` for now? I'd rather have more opportunity to see how this performs in the wild, before we commit to exposing it as a public symbol on our API surface.

danlamanna:
> Can we name the new object _cache for now? I'd rather have more opportunity to see how this performs in the wild, before we commit to exposing it as a public symbol on our API surface.

Did you mean name the file `_cache.py` or prefix the `cache` and `requestCache` objects with an underscore?

danlamanna:
> Can we name the new object _cache for now? I'd rather have more opportunity to see how this performs in the wild, before we commit to exposing it as a public symbol on our API surface.

@brianhelba done, PTAL.

mgrauer:
Nice @danlamanna, thanks for your patience.

📆

PR #2275, opened 2017-08-09
Make item tasks modal focus the first input field

danlamanna:
This commit makes all tabs of the "Auto-configure task items" modal
put focus on their first input, this happens both on render and on
toggling of different tabs.


danlamanna:
> Is this something we should explicitly test?

I don't think that level of granularity is necessary here.

PR #2276, opened 2017-08-09
docs: Include release notes in generate docs

jcfr:
See #2232


jcfr:
screenshot from 2017-08-09 16-37-32

zachmullen:
Thanks for showing the technique; we have a doc reorg happening over in #2257 -- any debates about where this should go in the doc tree should happen in the context of the new tree.

jcfr:
👍

PR #2277, opened 2017-08-10
Single POST uploads from the python client.

manthey:
Change the girder python client to use a single POST for uploads when possible.  This only applies to small files being sent to a sufficient Girder version.

This takes advantage of the change in PR #2111.  Our tests go through both the old and new code paths.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -913,6 +913,14 @@ class GirderClient(object):
         if reference:
             params['reference'] = reference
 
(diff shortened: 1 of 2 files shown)


zachmullen:
Has this been tested with the S3 assetstore?

manthey:
I haven't done so.  It looks like our py_client/cli_test only tests with the default test assetstore (which is a filesystem assetstore).  A test of S3 should be added to the server_upload test for the one-post method added in PR #2111, since that would exercise the upload method in a more general way.

manthey:
Test added.

zachmullen:
Do we cover both branches of your new block?

manthey:
Yes.

PR #2278, opened 2017-08-10
Fix JSONeditor spacing

### LINKED ISSUE #2189: JSON editor truncates output in "code" view
The last few lines do not show up if the document is fairly long.

sutartmelson:
Fixes #2189

A css rule made the editor extend 35px bellow the parent metadata element as it did not account for the height of the header.

I think :)


zachmullen:
I verified that this fixes the issue for me.

How would you feel about also upgrading to the latest version of this package? I'm told there are several improvements, it is probably worth spending 30-60 minutes to see if it's an easy upgrade.

sutartmelson:
Sure, I'll look into it.

PR #2279, opened 2017-08-10
Add new settings to set the contact email address

### LINKED ISSUE #2266: Allow the "contact" link to point to a different email
The Contact [link](https://github.com/girder/girder/blob/f445dca9b4f95c77374b335c04e717d102d9dbc6/clients/web/src/templates/layout/layoutFooter.pug#L3) on the web client currently goes to "kitware@kitware.com".  I need to make this configurable, so a particular deployment can point elsewhere.

Pierre-Assemat:
Fixes #2266

### DIFF
--- a/clients/web/src/templates/body/systemConfiguration.pug
+++ b/clients/web/src/templates/body/systemConfiguration.pug
@@ -7,7 +7,7 @@ form.g-settings-form(role="form")
       input#g-core-contact-email-address.form-control.input-sm(
           type="text", value=settings['core.contact_email_address'] || '',
           placeholder=`Default: ${defaults['core.contact_email_address'] || 'none'}`,
-          title="The contact email address for support.")
+          title="The contact email address for users to obtain support with this instance.")
   .form-group
     label(for="g-core-cookie-lifetime") Cookie duration (days)
     input#g-core-cookie-lifetime.form-control.input-sm(
--- a/clients/web/src/templates/body/systemConfiguration.pug
+++ b/clients/web/src/templates/body/systemConfiguration.pug
@@ -1,5 +1,13 @@
 .g-body-title System configuration
 form.g-settings-form(role="form")
+  .g-settings-form-container
+    h4 Instance Branding
+    .form-group
+      label(for="g-core-contact-email-address") Contact email address
+      input#g-core-contact-email-address.form-control.input-sm(
+          type="text", value=settings['core.contact_email_address'] || '',
+          placeholder=`Default: ${defaults['core.contact_email_address'] || 'none'}`,
+          title="The contact email address for support.")
   .form-group
     label(for="g-core-cookie-lifetime") Cookie duration (days)
     input#g-core-cookie-lifetime.form-control.input-sm(
--- a/clients/web/src/templates/layout/layoutFooter.pug
+++ b/clients/web/src/templates/layout/layoutFooter.pug
@@ -1,6 +1,6 @@
 .g-footer-links
   a#g-about-link(href="http://girder.readthedocs.org/en/latest/user-guide.html") About
-  a#g-contact-link(href="mailto:kitware@kitware.com") Contact
+  a#g-contact-link(href=contactHref) Contact
   a(href=apiRoot) Web API
   a#g-bug-link(href="https://github.com/girder/girder/issues/new") Report a bug
 
--- a/clients/web/src/templates/layout/layoutFooter.pug
+++ b/clients/web/src/templates/layout/layoutFooter.pug
@@ -1,6 +1,7 @@
 .g-footer-links
   a#g-about-link(href="http://girder.readthedocs.org/en/latest/user-guide.html") About
-  a#g-contact-link(href=contactHref) Contact
+  if (contactHref !== '')
+    a#g-contact-link(href=contactHref) Contact
   a(href=apiRoot) Web API
   a#g-bug-link(href="https://github.com/girder/girder/issues/new") Report a bug
 
--- a/clients/web/src/templates/layout/layoutFooter.pug
+++ b/clients/web/src/templates/layout/layoutFooter.pug
@@ -1,6 +1,6 @@
 .g-footer-links
   a#g-about-link(href="http://girder.readthedocs.org/en/latest/user-guide.html") About
-  if (contactHref !== '')
+  if contactHref
     a#g-contact-link(href=contactHref) Contact
   a(href=apiRoot) Web API
   a#g-bug-link(href="https://github.com/girder/girder/issues/new") Report a bug
--- a/clients/web/src/views/App.js
+++ b/clients/web/src/views/App.js
@@ -39,7 +39,7 @@ var App = View.extend({
     initialize: function (settings) {
         this._started = false;
         settings = settings || {};
-        this.contactEmail = settings.contactEmail || '';
+        this.contactEmail = settings.contactEmail || null;
         if (settings.start === undefined || settings.start) {
             this.start();
         }
--- a/clients/web/src/views/App.js
+++ b/clients/web/src/views/App.js
@@ -39,6 +39,7 @@ var App = View.extend({
     initialize: function (settings) {
         this._started = false;
         settings = settings || {};
+        this.contactEmail = settings.contactEmail || '';
         if (settings.start === undefined || settings.start) {
             this.start();
         }
@@ -131,7 +132,8 @@ var App = View.extend({
         });
 
         this.footerView = new LayoutFooterView({
-            parentView: this
+            parentView: this,
+            contactEmail: this.contactEmail
         });
 
         this.progressListView = new ProgressListView({
--- a/clients/web/src/views/layout/FooterView.js
+++ b/clients/web/src/views/layout/FooterView.js
@@ -12,13 +12,15 @@ var LayoutFooterView = View.extend({
 
     initialize: function (settings) {
         const mail = settings.contactEmail || '';
-        this.emailHref = 'mailto:' + mail;
+        if (mail !== '') {
+            this.emailHref = 'mailto:' + mail;
+        }
     },
 
     render: function () {
         this.$el.html(LayoutFooterTemplate({
             apiRoot: getApiRoot(),
-            contactHref: this.emailHref
+            contactHref: this.emailHref || ''
         }));
         return this;
     }
--- a/clients/web/src/views/layout/FooterView.js
+++ b/clients/web/src/views/layout/FooterView.js
@@ -11,16 +11,14 @@ import 'girder/stylesheets/layout/footer.styl';
 var LayoutFooterView = View.extend({
 
     initialize: function (settings) {
-        const mail = settings.contactEmail || null;
-        if (mail !== null) {
-            this.emailHref = `mailto: ${mail}`;
-        }
+        const contactEmail = settings.contactEmail || null;
+        this.contactHref = contactEmail !== null ? `mailto:${contactEmail}` : null;
     },
 
     render: function () {
         this.$el.html(LayoutFooterTemplate({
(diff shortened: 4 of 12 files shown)


girder/models/setting.py
```diff
@@ -279,6 +279,12 @@ def validateCoreEmailFromAddress(doc):
             raise ValidationException('Email from address must not be blank.', 'value')
 
     @staticmethod
+    @setting_utilities.validator(SettingKey.CONTACT_EMAIL_ADDRESS)
+    def validateCoreContactEmailAddress(doc):
+        if not doc['value']:
+            raise ValidationException('Email from address must not be blank.', 'value')
```
danlamanna: There is a configuration option for specifying a regex email addresses must conform to, that can probably be used here as well.

girder/utility/webroot.py
```diff
@@ -88,9 +90,18 @@ def __init__(self, templatePath=None):
...
             'staticRoot': '',
-            'title': 'Girder'
+            'title': 'Girder',
+            'contactEmail': ModelImporter.model('setting').get(SettingKey.CONTACT_EMAIL_ADDRESS)
         }
 
+        events.bind('model.setting.save.after', CoreEventHandler.WEBROOT_SETTING_CHANGE,
+                    self._onSettingSave)
+
+    def _onSettingSave(self, event):
```
danlamanna: This sort of thing (which exists in other places in Girder thanks to yours truly) will not  work in a distributed installation. 

For the purposes of this PR this might be fine, but I think that #2274 might try to remove these patterns. For more detail about it there's a thread on girder-developers with the subject "Signalling between multiple Girder servers".

brianhelba: Indeed, as part of fully implementing #2274, the `WebrootBase.indexHtml` should be switched to use a proper cache. I see nothing more that can be done on this PR.

girder/models/setting.py
brianhelba: Per [the discussion here](https://stackoverflow.com/questions/8147805/the-mailto-setting-a-proper-name-on-the-recipient) we probably want to support more than a simple email address regex (especially since the default `EMAIL_FROM_ADDRESS` default already uses a proper name component).

The space and URL encoding described by the RFC also should be done, but probably not in the validator (as that will make the value hard to edit). The best place is probably in the Pug template.

clients/web/src/templates/layout/layoutFooter.pug
```diff
@@ -1,6 +1,6 @@
 .g-footer-links
   a#g-about-link(href="http://girder.readthedocs.org/en/latest/user-guide.html") About
-  a#g-contact-link(href="mailto:kitware@kitware.com") Contact
+  a#g-contact-link(href=href) Contact
```
brianhelba: Can this be something like `contactHref`?

clients/web/src/views/App.js
```diff
@@ -40,6 +40,7 @@ var App = View.extend({
         this._started = false;
         settings = settings || {};
         if (settings.start === undefined || settings.start) {
+            this.contactEmail = settings.contactEmail;
```
brianhelba: We should always store this setting in an instance variable, even if we're not starting the app.

clients/web/src/views/layout/FooterView.js
```diff
@@ -9,9 +9,16 @@ import 'girder/stylesheets/layout/footer.styl';
...
+    initialize: function (settings) {
+        var mail = settings.contactEmail; // Has to change
+        this.emailHref = 'mailto:' + mail;
+    },
+
     render: function () {
         this.$el.html(LayoutFooterTemplate({
-            apiRoot: getApiRoot()
+            apiRoot: getApiRoot(),
+            href: this.emailHref
```
brianhelba: Can this be something like `contactHref`?

clients/web/src/views/layout/FooterView.js
```diff
@@ -9,9 +9,16 @@ import 'girder/stylesheets/layout/footer.styl';
  * This view shows the footer in the layout.
  */
 var LayoutFooterView = View.extend({
+
+    initialize: function (settings) {
+        var mail = settings.contactEmail; // Has to change
```
brianhelba: This should have a fallback value, in case `settings.contactEmail` is not provided.

brianhelba: Also, use `let` or `const`.

Pierre-Assemat: the fallback value can be the default email ?

brianhelba: I'm not sure, but maybe not. If a default is needed for this widget, it means that it's being used in another application that didn't pass enough parameters at all.

Maybe if no user is passed at all, then don't use an `href`? What do you think?

Pierre-Assemat: I used a blank ('') instead, but do you mean test if `settings.contactEmail` is provided and if not don't use href and hide the contact link ?

brianhelba: I think that's a good idea.

This case will only apply when a some external code imports and reuses this view, but it should be handled gracefully.

clients/web/src/views/layout/FooterView.js
```diff
@@ -9,9 +9,18 @@ import 'girder/stylesheets/layout/footer.styl';
  * This view shows the footer in the layout.
  */
 var LayoutFooterView = View.extend({
+
+    initialize: function (settings) {
+        const mail = settings.contactEmail || '';
```
brianhelba: In `initialize`, I think we should always set `this.contactEmail` to be equal to `settings.contactEmail`, with a default of `null` if it's not provided.

Then, in `render`, we should conditionally apply the concatenation with `mailto:` (encoding it if necessary), then pass that result (or `null`) into the template.

This will also `LayoutFooterView` to be subclassed such that it takes the same parameters, but renders them in a different way.

clients/web/src/views/App.js
```diff
@@ -39,6 +39,7 @@ var App = View.extend({
     initialize: function (settings) {
         this._started = false;
         settings = settings || {};
+        this.contactEmail = settings.contactEmail || '';
```
brianhelba: Let's use `null` for the value when no setting is provided, since that will never be confused with a case where a server sends a setting value that's purposely set to empty string.

tests/cases/webroot_test.py
```diff
@@ -42,6 +42,9 @@ def testAccessWebRoot(self):
         self.assertTrue('girder_app.min.js' in body)
         self.assertTrue('girder_lib.min.js' in body)
 
+        emailAddress = self.model('setting').get(SettingKey.CONTACT_EMAIL_ADDRESS)
```
brianhelba: This is just going to get the default.

To ensure that changes to this setting are immediately respected, we should next set this setting to something totally different, then make another request to get the webroot, and ensure that the value is updated.

clients/web/src/templates/layout/layoutFooter.pug
```diff
@@ -1,6 +1,7 @@
 .g-footer-links
   a#g-about-link(href="http://girder.readthedocs.org/en/latest/user-guide.html") About
-  a#g-contact-link(href="mailto:kitware@kitware.com") Contact
+  if (contactHref !== '')
```
brianhelba: [Pug conditionals](https://pugjs.org/language/conditionals.html) don't need parentheses.

I think you can check whether this is truthy at all (`if contactHref`), which will fail if it's `null` (intentionally not set) or an empty string (intentionally set, but empty).

clients/web/src/templates/body/systemConfiguration.pug
```diff
@@ -1,5 +1,13 @@
 .g-body-title System configuration
 form.g-settings-form(role="form")
+  .g-settings-form-container
+    h4 Instance Branding
+    .form-group
+      label(for="g-core-contact-email-address") Contact email address
+      input#g-core-contact-email-address.form-control.input-sm(
+          type="text", value=settings['core.contact_email_address'] || '',
+          placeholder=`Default: ${defaults['core.contact_email_address'] || 'none'}`,
+          title="The contact email address for support.")
```
brianhelba: How about `title="The contact email address for users to obtain support with this instance."`? I'm open to even better phrasing, but this is a little more descriptive.

girder/models/setting.py
brianhelba: We still need to address this.

clients/web/src/views/layout/FooterView.js
```diff
@@ -9,9 +9,18 @@ import 'girder/stylesheets/layout/footer.styl';
...
+        if (mail !== '') {
+            this.emailHref = 'mailto:' + mail;
```
brianhelba: Use [an ES6 template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for this concatenation.

brianhelba:
After reading more of [RFC6068](https://tools.ietf.org/html/rfc6068) myself, it's clear that we cannot use a `name-addr` (like `'Kitware Support <kitware@kitware.com>'` or `'Foo Bar <foo@bar.com>'`.

Although that spec also requires us to percent-encode many other characters that may appear in a valid email address (e.g. `+`), I think we can add the percent-encoding as part of a later enhancement. It will need to be done in the Javascript code, rather than the Python validation, as it's specific to the `mailto:` encoding, not the value of the email address itself. Some basic validation of the email address in the setting would also be nice, but apparently [that's not trivial either](https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address).

So, for now, I think you need to just replace the default and test email addresses with just the core `addr-spec` components (e.g.  `'kitware@kitware.com'` or `'foo@bar.com'`).

Pierre-Assemat:
Do I need to test if it's a good email address with some regular expression too ?

brianhelba:
`EMAIL_FROM_ADDRESS` is not used in a `mailto: ` link, and so does not need to follow the stricter rules of [RFC6068](https://tools.ietf.org/html/rfc6068), rather, it can probably use [RFC5322](https://tools.ietf.org/html/rfc5322).

Because:
* `EMAIL_FROM_ADDRESS` already does not validate using an email regex
* There is the potential for duplicate functionality with the `email_regex` setting from Girder's config file
* A robust [regex for email-validation](https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address) looks to be quite complicated

I think it's justified to merge this without extensive validation, then fix things more globally in a separate PR. I don't want to derail this entire PR for a corner case, as long as we eventually fix the corner case. I'll file an issue for it now.

brianhelba:
https://github.com/girder/girder/issues/2282

clients/web/src/views/layout/FooterView.js
```diff
@@ -9,9 +9,18 @@ import 'girder/stylesheets/layout/footer.styl';
  * This view shows the footer in the layout.
  */
 var LayoutFooterView = View.extend({
+
+    initialize: function (settings) {
+        const mail = settings.contactEmail || null;
+        if (mail !== null) {
+            this.emailHref = `mailto: ${mail}`;
```
brianhelba: [RFC6068](https://tools.ietf.org/html/rfc6068) says that there's no space between `mailto:` and the email address,

Also, in my opinion, it'd be cleaner overall to name this local variable `const contactEmail`, and the instance variable `this.contactHref`.

Pierre-Assemat: When you propose to convert into JSON you mean in the webroot.py ? Just like that : 
`contactEmail = json.loads(ModelImporter.model('setting').get(SettingKey.CONTACT_EMAIL_ADDRESS))`

And I just insert this into the template webroot.mako ?

mgrauer:
Two in one day!  🥂

Issue #1093, opened 2015-11-18
Add option for instance brand name


brianhelba:
Currently, emails automatically sent by Girder core have the brand name `"Girder"` hardcoded in their subject line:
- [`"Girder: You've been invited to a group"`](https://github.com/girder/girder/blob/3d869245e8650cc71ab5b56975fe45f57c6f78b8/girder/api/v1/group.py#L277)
- [`'Girder: Password reset'`](https://github.com/girder/girder/blob/3d869245e8650cc71ab5b56975fe45f57c6f78b8/girder/api/v1/user.py#L308)
- [`'Girder: Temporary access'`](https://github.com/girder/girder/blob/3d869245e8650cc71ab5b56975fe45f57c6f78b8/girder/api/v1/user.py#L337)

Girder should support some sort of "brand name" setting, (which would default to `"Girder"`) that would be templated into these email subject lines.

Additionally, this "brand name" should be automatically templated into / used in:
- [email body headers](https://github.com/girder/girder/blob/3d869245e8650cc71ab5b56975fe45f57c6f78b8/girder/mail_templates/_header.mako#L3)
- [email body footers](https://github.com/girder/girder/blob/3d869245e8650cc71ab5b56975fe45f57c6f78b8/girder/mail_templates/_footer.mako)
- [the Webroot `title` property](https://github.com/girder/girder/blob/3d869245e8650cc71ab5b56975fe45f57c6f78b8/girder/utility/webroot.py#L89), for use in the [HTML title](https://github.com/girder/girder/blob/3d869245e8650cc71ab5b56975fe45f57c6f78b8/girder/utility/webroot.mako#L5)
- [the main layout header](https://github.com/girder/girder/blob/3d869245e8650cc71ab5b56975fe45f57c6f78b8/clients/web/src/templates/layout/layoutHeader.jade#L2)
- possibly [the front page](https://github.com/girder/girder/blob/3d869245e8650cc71ab5b56975fe45f57c6f78b8/clients/web/src/templates/body/frontPage.jade), but most non-trivial Girder deployments will want to override that with more custom text

jcfr:
:+1:

brianhelba:
@girder/developers Does anyone have time to work on this?

zachmullen:
I can take a look at it.

Pierre-Assemat:
Link where the brand name is used : 
[...]

PR #2283, opened 2017-08-15
Add 'brandName' setting in server side

Pierre-Assemat:
Fixes : #1093 

screenshot from 2017-08-17 09-59-16

Default configuration with 'Girder'  as `brandName` :
screenshot from 2017-08-17 09-58-49

Then an other brandName is set :
screenshot from 2017-08-17 10-00-16
screenshot from 2017-08-17 10-00-34


girder/api/v1/group.py
```diff
@@ -247,7 +247,7 @@ def inviteToGroup(self, group, userToInvite, level, quiet, force):
                 })
                 mail_utils.sendEmail(
                     to=userToInvite['email'], text=html,
-                    subject="Girder: You've been invited to a group")
+                    subject="%s: You've been invited to a group" % SettingKey.BRAND_NAME)
```
brianhelba: `SettingKey.BRAND_NAME` is equal to `'core.brand_name'`. I think you need to use the setting model API to `get` the setting from the database here.

girder/api/v1/user.py
```diff
@@ -298,7 +298,7 @@ def resetPassword(self, email):
         html = mail_utils.renderTemplate('resetPassword.mako', {
             'password': randomPass
         })
-        mail_utils.sendEmail(to=email, subject='Girder: Password reset', text=html)
+        mail_utils.sendEmail(to=email, subject='%s: Password reset' % SettingKey.BRAND_NAME, text=html)
```
brianhelba: Again, this needs to be a query.

girder/api/v1/user.py
```diff
@@ -323,9 +323,10 @@ def generateTemporaryPassword(self, email):
 
         html = mail_utils.renderTemplate('temporaryAccess.mako', {
             'url': url,
-            'token': str(token['_id'])
+            'token': str(token['_id']),
+            'brandName': SettingKey.BRAND_NAME
```
brianhelba: Again, this needs to be a query.

girder/api/v1/user.py
```diff
@@ -323,9 +323,10 @@ def generateTemporaryPassword(self, email):
...
         })
-        mail_utils.sendEmail(to=email, subject='Girder: Temporary access', text=html)
+        mail_utils.sendEmail(to=email, subject='%s: Temporary access' % SettingKey.BRAND_NAME, text=html)
```
brianhelba: Again, this needs to be a query.

girder/constants.py
```diff
@@ -340,3 +342,6 @@ class CoreEventHandler(object):
 
     # For adding a user into its own ACL.
     USER_SELF_ACCESS = 'core.grantSelfAccess'
+
+    # For updating the cached webroot HTML when settings change.
+    WEBROOT_SETTING_CHANGE = 'core.updateWebrootSettings'
```
brianhelba: All files need to end with a newline. This is missing one.

girder/models/setting.py
```diff
@@ -165,6 +165,13 @@ def getDefault(self, key):
         return None
 
     @staticmethod
+    @setting_utilities.validator(SettingKey.BRAND_NAME)
+    def validateCoreBrandName(doc):
+        if not doc['value']:
+            raise ValidationException('The brand name is not correct', 'value')
```
brianhelba: A slightly more descriptive error, like `'The brand name may not be empty'`, would be better.

Pierre-Assemat: Ok thank you !!

girder/api/v1/group.py
Pierre-Assemat: I was'n sure, but each time I call `ImportModel` it do a get request to the server ?
That's not slowing the app ?

brianhelba: To be clear, this is already Python code, which *is* the server, so there's no HTTP requests made at all.

The `ModelImporter.model('setting')` bit just does a dynamic lookup of the correct Python class to use; this is just causing some global Python variables to be accessed, which is fast.

The `.get()` function call internally runs a PyMongo `find` query to complete, so it will take a bit more time. Eventually, we may want to cache more of these database queries (see @danlamanna 's work in #2274 ), but [premature optimization is the root of all evil](http://wiki.c2.com/?PrematureOptimization), so I wouldn't worry about it right now.

girder/api/v1/user.py
```diff
@@ -323,9 +328,14 @@ def generateTemporaryPassword(self, email):
 
         html = mail_utils.renderTemplate('temporaryAccess.mako', {
             'url': url,
-            'token': str(token['_id'])
+            'token': str(token['_id']),
+            'brandName': ModelImporter.model('setting').get(SettingKey.BRAND_NAME)
```
brianhelba: Don't pass this here, assume it's injected by `renderTemplate`.

tests/cases/mail_test.py
```diff
@@ -92,7 +92,8 @@ def testPluginTemplates(self):
 
         # Make sure it also works from in-mako import statements
         content = mail_utils.renderTemplate('temporaryAccess.mako', {
-            'url': 'x'
+            'url': 'x',
+            'brandName': 'FooBar'
```
brianhelba: Don't pass this here, assume it's injected by `renderTemplate`.

girder/api/describe.py
```diff
@@ -436,7 +436,7 @@ def __init__(self, templatePath=None):
         self.vars = {
             'apiRoot': '',
             'staticRoot': '',
-            'title': 'Girder - REST API Documentation',
+            'brandName': ModelImporter.model('setting').get(constants.SettingKey.BRAND_NAME),
```
brianhelba: This is not getting updated until we restart the server. It probably needs an event handler too.

tests/cases/webroot_test.py
```diff
@@ -45,11 +45,17 @@ def testAccessWebRoot(self):
         emailAddress = self.model('setting').get(SettingKey.CONTACT_EMAIL_ADDRESS)
         self.assertTrue('contactEmail: \'%s\'' % emailAddress in body)
         self.model('setting').set(SettingKey.CONTACT_EMAIL_ADDRESS, 'foo@bar.com')
+
+        brandName = self.model('setting').get(SettingKey.BRAND_NAME)
+        self.assertTrue('<title>%s</title>' % brandName in body)
+        self.model('setting').set(SettingKey.BRAND_NAME, 'FooBar')
         # A new request to update changes
```
brianhelba: How about one more newline here, for clarity?

clients/web/src/views/body/FrontPageView.js
```diff
@@ -37,16 +37,18 @@ var FrontPageView = View.extend({
         }
     },
 
-    initialize: function () {
+    initialize: function (settings) {
         cancelRestRequests('fetch');
+        this.brandName = settings.brandName;
```
brianhelba: This should have a default, if `settings.brandName` is not provided. `'Girder'` would be fine.

clients/web/src/views/layout/HeaderView.js
```diff
@@ -17,7 +17,9 @@ var LayoutHeaderView = View.extend({
         }
     },
 
-    initialize: function () {
+    initialize: function (settings) {
+        this.brandName = settings.brandName || null;
```
brianhelba: This default should be `'Girder'`, as `null` will not render cleanly, and not showing the dependent elements isn't an option in this case.

clients/web/src/views/App.js
```diff
@@ -40,6 +40,7 @@ var App = View.extend({
         this._started = false;
         settings = settings || {};
         this.contactEmail = settings.contactEmail || null;
+        this.brandName = settings.brandName || null;
```
brianhelba: This actually can stay as `null`, as long as dependent views have a default fallback of `'Girder'`. Leaving it as `null` preserves the information that an instance name was not set on App instantiation.

tests/cases/webroot_test.py
```diff
@@ -45,11 +45,17 @@ def testAccessWebRoot(self):
         emailAddress = self.model('setting').get(SettingKey.CONTACT_EMAIL_ADDRESS)
         self.assertTrue('contactEmail: \'%s\'' % emailAddress in body)
         self.model('setting').set(SettingKey.CONTACT_EMAIL_ADDRESS, 'foo@bar.com')
+
+        brandName = self.model('setting').get(SettingKey.BRAND_NAME)
+        self.assertTrue('<title>%s</title>' % brandName in body)
```
brianhelba: Checking the `<title>` is good here, but let's also check the Javascript value.

Pierre-Assemat: What do you mean by checking the Javascript value ? 
I think we just need to add some client side tests, is that you're talking about ?

zachmullen: I haven't looked at the specs for python's order-of-operations to know whether this is safe, but it would make me more comfortable to be explicit here using parens:

    self.assertTrue(('<title>%s</title>' % brandName) in body)

brianhelba: I just mean the same test that you do for `contactEmail`, to ensure it's passed as a parameter to `App`.

girder/api/describe.py
Pierre-Assemat: `WEBROOT_SETTING_CHANGE = 'core.updateWebrootSettings'` wasn't do the job ?

tests/cases/webroot_test.py
Pierre-Assemat: It's better to add the parens ?

girder/api/describe.py
brianhelba: The bind right now will update the webroot at `/`, but not the API description page. You should be able to test and see that it's not updated until a restart.

Pierre-Assemat: Ok I just see that !
I will try to figure out which is the best way to prevent this issue

Pierre-Assemat: I added a event Handler on the Api page but it's not working when I try to set the brandname with a blank (set by default)

Pierre-Assemat:
Maybe we should add a test to check if the brand name is correctly set on the api/v1 front page too, what do you think ?

clients/web/src/views/App.js
```diff
@@ -221,7 +223,8 @@ var App = View.extend({
 
             settings = _.extend(settings, {
                 el: this.$('#g-app-body-container'),
-                parentView: this
+                parentView: this,
+                brandName: this.brandName
```
brianhelba: This will cause the `brandName` to be passed to every single body view. While this solves the issue at hand, I think it might be a good idea to introduce an application settings registry to handle this and other data (some of which may not be view parameters).

One thing we could do for now is to pass this parameter now as `brandName`, and note that it's unstable and subject to being moved.

@girder/developers What do you all think?

Pierre-Assemat: I think I will discuss this with @ronichoudhury because he wants to add some other configurations to this PR

ronichoudhury: I like the idea of a settings registry because then we can flexibly add things to it as needed in future PRs without affecting every body view, etc.

clients/web/src/templates/body/frontPage.pug
```diff
@@ -1,20 +1,20 @@
...
+    b Welcome to #{brandName}!
     |  If this is your first time here, you might want to check out the
     a(target="_blank", href='http://girder.readthedocs.org/en/latest/user-guide.html')  User Guide
     |  to learn the basics of how to use the application.
     |  To browse the data hosted on this server, start on the
     a.g-collections-link  Collections
     |  page. If you want to search for specific data on this server, use the
     a.g-quicksearch-link  Quick Search
-    |  box at the top of the screen. Developers who want to use the Girder REST API
+    |  box at the top of the screen. Developers who want to use the #{brandName} REST API
```
ronichoudhury: I am not sure this instance of the word "Girder" should be replaced by the brand name. The software is still Girder, after all, we just want to brand the *instance* differently. Right?

ronichoudhury:
One thing I'd like to request: I think the "banner name" and the "title text" should be independently configurable, perhaps with a fallback where they match each other.

In particular, I am going to have this problem with the DKC facelift, where we will want the banner to read something like "Kitware Data", while the title will have a slightly longer, different message ("Welcome to data.kitware.com" or similar).

clients/web/src/views/App.js
Pierre-Assemat: Ok, I think it's a good idea ! But I'm not sure to know how we can do that

ronichoudhury: @brianhelba, what do you think of merging the PR without a full-on settings registry right now, then working on that next? @mgrauer may have an opinion on this as well.

clients/web/src/templates/body/frontPage.pug
Pierre-Assemat: It's true, I think we have to keep some track of 'Girder' even if an admin user can configure lot of other things...

Pierre-Assemat:
Yes I can do that : 2 differents settings for the "banner name" and the "title text",
but to be sure :

This is the banner :
screenshot from 2017-08-24 11-41-56

And this is the title text : 
screenshot from 2017-08-24 11-42-13

am I right ?

ronichoudhury:
Yep! So the word "Girder" is currently the same in both places; I think it makes sense to let it vary. The one in the banner can be the `brandName` and the other one should be called maybe `titleName` or `titleText` or something like that.

Pierre-Assemat:
Ok sounds good, I do that right now !

clients/web/src/views/App.js
mgrauer: Let's do the settings registry in a later PR to keep the scope specific here.

girder/api/describe.py
Pierre-Assemat: @ronichoudhury do you think the titleName has to be set in the API too ?

clients/web/src/views/App.js
brianhelba: Yes, a settings registry is way out of scope here, but if we agree it's a good idea, we should get it in before the next release, as the current implementation of passing things to every body view isn't very clean.

clients/web/src/templates/body/frontPage.pug
brianhelba: Personally, I'd err more on the side of letting instance admins take total brand ownership of the instance. If the instance contains their data, then they may want to brand the API to fetch data as theirs.

I think a "Powered by Girder" string in the footer would be an appropriate place to still inject "our" Girder branding by default.

girder/utility/server.py
```diff
@@ -76,7 +75,8 @@ def _configureStaticRoutes(webroot, plugins, event=None):
     webroot.updateHtmlVars({
         'apiRoot': config.getConfig()['server']['api_root'],
         'staticRoot': staticRoot,
-        'plugins': plugins
+        'plugins': plugins,
+        'title': model_importer.ModelImporter.model('setting').get(constants.SettingKey.BRAND_NAME)
```
brianhelba: I don't think this needs to be here. The Webroot will be instantiated with the correct value.

girder/utility/webroot.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, templatePath=None):
             'plugins': [],
             'apiRoot': '',
             'staticRoot': '',
-            'title': 'Girder',
+            'title': ModelImporter.model('setting').get(SettingKey.BRAND_NAME),
```
brianhelba: For consistency, let's pass this as brandName too.

brianhelba: And for compatibility, continue to pass `title` as a hardcoded value, with a code comment that it's depreciated. The `title` should no longer be used in the `webroot.mako` template though.

ronichoudhury:
Why? If the additional changes are of the same type as the ones already in the pipeline, why not just take care of them now?

Pierre-Assemat:
Maybe we can just merge like that, with a little differentiation on the front page and the next PR will implement the settings registry ?

brianhelba:
This is fine, I didn't realize @Pierre-Assemat has already done the work.

It looks like `titleName` is used in only one place on the front page right now. Can someone provide a high level overview of the difference in semantics between these two strings, where they should be used generally, and how they should be document to explain this?

brianhelba:
The settings in this last commit look like they should go into the `homepage` plugin. I think this is getting too far out of scope for this PR, which should touch branding across the entire Girder stack. Maybe this should be added in a follow-up PR?

ronichoudhury:
The homepage plugin will require a lot of modification to support this sort of customization. Maybe you're right that it belongs there, but the other changes are of the same kind and size as setting the brand name, plus I have an immediate need for the customization to finish my "DKC facelift" work.

Since I have a need, and the changes are small, I'd say let's get this merged in now and then have a separate discussion about whether things belong in the homepage plugin.

Pierre-Assemat:
Do you need the banner color setting too ?

Pierre-Assemat:
@brianhelba @mgrauer what do you think about that : 

screenshot from 2017-08-25 17-53-32

I can't remove the space below because of the button for set the default color which isn't an input in the form :
The input (color picker) is linked to the label and so I can't put the button and the color picker inline with an other method that use `position: relative`.. And it causes the space below

clients/web/src/stylesheets/body/systemConfig.styl
```diff
@@ -5,14 +5,18 @@
...
+
+  #g-core-banner-color
+    display inline
+    max-width 15%
+
+  #g-core-banner-default-color
+    width 60px
+    display inline
+    position relative
+    bottom 5px
```
Pierre-Assemat: @ronichoudhury  the bug comes from the `bottom 5px`

brianhelba:
I've merged in the changes from #2300 . Here's what the updated front page looks like with the default brand name (Girder):
frontpage
Here's the updated front page with a different brand name set:
frontpage1

@ronichoudhury What do you think of this?

brianhelba:
@Pierre-Assemat I noticed that clearing the value of webroot-passed settings does not actually cause the webroot to update. I've added some new test code to demonstrate this failure.

The call stack in this case is:
* [`System.setSetting`](https://github.com/girder/girder/blob/master/girder/api/v1/system.py#L95)
* [`Setting.unset`](https://github.com/girder/girder/blob/master/girder/models/setting.py#L147)
* [`Model.remove`](https://github.com/girder/girder/blob/master/girder/models/model_base.py#L473)

Since the model is not technically being "saved" in this case, the `'model.setting.save.after'` event is not being triggered.

You'll need to make a change to fix this bug, please.

Pierre-Assemat:
Adding the `'brandName': model_importer.ModelImporter.model(
                           'setting').get(constants.SettingKey.BRAND_NAME)`
In the `webroot.updateHtmlVars` in server.py fix the bug like I said to you before ! 
But you seemed find that depreciated, can I just put it again ?
Or do I need to find an other way to fix that using the `model.setting.save` ?

girder/utility/webroot.py
```diff
@@ -90,17 +90,32 @@ def __init__(self, templatePath=None):
...
+        elif settingDoc['key'] == SettingKey.BRAND_NAME:
+            self.updateHtmlVars({'brandName': settingDoc['value']})
+
+    def _onSettingRemove(self, event):
+        settingDoc = event.info
+        if settingDoc['key'] == SettingKey.CONTACT_EMAIL_ADDRESS:
+            self.updateHtmlVars({'contactEmail': ModelImporter.model(
+                'setting').getDefault(SettingKey.CONTACT_EMAIL_ADDRESS)})
+        elif settingDoc['key'] == SettingKey.BRAND_NAME:
+            self.updateHtmlVars({'brandName': ModelImporter.model(
```
brianhelba: Putting the line break here is a little awkard for me to read. How about:
```
self.updateHtmlVars({
    'contactEmail': ModelImporter.model('setting').getDefault(
        SettingKey.CONTACT_EMAIL_ADDRESS)})
```
That is admittedly 1 line longer, but fits inside 100 columns and is easier to read (for me).

girder/utility/webroot.py
```diff
@@ -90,17 +90,32 @@ def __init__(self, templatePath=None):
...
         settingDoc = event.info
         if settingDoc['key'] == SettingKey.CONTACT_EMAIL_ADDRESS:
             self.updateHtmlVars({'contactEmail': settingDoc['value']})
+        elif settingDoc['key'] == SettingKey.BRAND_NAME:
+            self.updateHtmlVars({'brandName': settingDoc['value']})
+
+    def _onSettingRemove(self, event):
+        settingDoc = event.info
+        if settingDoc['key'] == SettingKey.CONTACT_EMAIL_ADDRESS:
+            self.updateHtmlVars({'contactEmail': ModelImporter.model(
```
brianhelba: Same line break issue here.

girder/api/describe.py
```diff
@@ -447,10 +448,26 @@ def __init__(self, templatePath=None):
...
+
+    def _onSettingSave(self, event):
+        settingDoc = event.info
+        if settingDoc['key'] == SettingKey.BRAND_NAME:
+            self.updateHtmlVars({'brandName': settingDoc['value']})
+
+    def _onSettingRemove(self, event):
+        settingDoc = event.info
+        if settingDoc['key'] == SettingKey.BRAND_NAME:
+            self.updateHtmlVars({'brandName': ModelImporter.model(
```
brianhelba: Same line break issue here.

girder/utility/webroot.py
Pierre-Assemat: I'm ok with this change

mgrauer:
I will speak for @ronichoudhury because he is out right now and I talked over with him what he wants--I hope I am getting it right, be we can always tweak this work again later after merging.

Generally this looks good, the only choice I would question is `This instance of SlideAtlas was built from git version X`.  I think this makes more sense as an instance of `Girder` rather than the `brandName`, but I am willing to be overruled here.

brianhelba:
@mgrauer @ronichoudhury I agree that it's debatable as to whether things like `This instance of X was built from` and `To use the X REST API` should refer to `"Girder"` or the `brandName`. I generally come down to prefer `brandName` wherever possible, to allow instance administrators to convey a sense of unique ownership to their end users. Once an instance admin spins up a site with *their users* and *their data*, I would argue that the site, API, etc. should describe itself in terms of *their branding*, even if they have not yet written their own custom plugins.

I think it would be good to add a few more discrete references to `Powered by Girder` on the site (e.g. see [this footer](https://isic-archive.com/)). However, I think we should respect that it's only "Powered by" Girder, and not try to force branded instances to *be* Girder. This is open source software, and I'd much rather err on the side of end users being slightly unclear about what the capabilities of "stock upstream Girder" vs a "branded instance" are, than come across as overly pushy of our brand. Indeed, one could even see the lack of clarity between "stock" and "branded" instances as a positive thing: it's incredibly easy to make Girder your own.

Finally, I think these design choices should primarily reflect the needs of new and less sophisticated (at Girder development) users, and users looking to rapidly spin up new branded instance for marketing purposes. Experts (including the maintainers of [data.kitware.com](https://data.kitware.com/)) should find it relatively easy to author plugins to provided unlimited customization of any parts of the UI, in ways that  single-line settings can never fulfill.

brianhelba:
I'm going to ahead and merge this. The only point of uncertainty concerns text on the front page, which is definitely going to be modified further by https://github.com/girder/girder/pull/2318 . I suggest that we use that PR as an opportunity to iterate further.

PR #2284, opened 2017-08-15
Json editor update

sutartmelson:
Update the JSONEditor dependency to its current (5.9.3) version.

### DIFF
--- a/clients/web/src/stylesheets/widgets/metadataWidget.styl
+++ b/clients/web/src/stylesheets/widgets/metadataWidget.styl
@@ -62,11 +62,11 @@
     border-bottom 1px solid #f0f0f0
 
     button
-      background #fbfbfb url("~jsoneditor/dist/img/jsoneditor-icons.png")
+      background #fbfbfb url("~jsoneditor/dist/img/jsoneditor-icons.svg")
       border 1px solid #f0f0f0
 
       :disabled
-        background #fbfbfb url("~jsoneditor/dist/img/jsoneditor-icons.png")
+        background #fbfbfb url("~jsoneditor/dist/img/jsoneditor-icons.svg")
         border 1px solid #f0f0f0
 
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -39,6 +39,7 @@
         "grunt-shell": "^2.1.0",
         "grunt-webpack": "^3.0.2",
         "grunt-zip": "^0.17.1",
+        "jsoneditor": "^5.9.3",
```
zachmullen: I think this only needs to go in the clients/web/src one, not here.

brianhelba: This package actually belongs in `clients/web/src/package.json`, since it's a dependency of the web client sources. You can remove it from this file entirely, and add it to that one using the `~` version specifier (instead of `^`).

brianhelba: Oh, I see it's already there. Then this line can be removed entirely.

sutartmelson:
I can't figure out how to get the proper `package-lock.json` file pushed. There always seems to be countless changes to the file or none at all.

clients/web/src/stylesheets/widgets/metadataWidget.styl
```diff
@@ -62,11 +62,11 @@
     border-bottom 1px solid #f0f0f0
 
     button
```
brianhelba: None of these styles have any effect, since the classes are now named `.jsoneditor-menu`, `.jsoneditor-search`, etc. Furthermore, if you correct the class names, these styles actually cause the appearance to break.

Also, the `border` setting of `.jsoneditor` is being overridden by the more-specific upstream style, which operates on `div.jsoneditor`. If you want this border style to have any effect, you'll need to have an `!important`.

sutartmelson: Supplemental PR addressing these concerns: #2359

PR #2285, opened 2017-08-15
Use grunt-contrib-pug to build test HTML

brianhelba:
* Rename "grunt_tasks/dev.js" to "grunt_tasks/test.js"
* Use grunt-contrib-pug to build test HTML


brianhelba:
I'm not very happy with the churn of `package-lock.json` due to `fsevents`, which is an optional package that only gets installed on macOS. I'm unable to preserve it when I update the `package-lock.json` file on my Linux machine, and even doing a line-by-line stage of the changes is inadequate, since some of the other packages get marked as `optional` or don't get flattened due to the presence of `fsevents`.

The build should work either way with `fsevents` included or excluded from `package-lock.json`, since ensuring that package is at a certain version should never affect development.

brianhelba:
@girder/developers PTAL.

brianhelba:
@girder/developers Conflict fixed.

Issue #2271, opened 2017-08-09
Add ability to install plugins from Ansible role


danlamanna:
Ideally the user could specify a list of paths or git paths and have `girder-install plugin` be run against them.

mgrauer:
There would also need to be a way to specify system dependencies tied to the plugin.

zachmullen:
@mgrauer what do you mean?

mgrauer:
This issue was migrated from a different repo which was intended for a very specific purpose, but doing this operation there was out of scope.  

The intention is to specify to Ansible to install a 3rd party Girder plugin, one that doesn't live in the Girder repo.  That 3rd party plugin might require some additional system package for building or execution.

For this to work you'd need to specify where to find the plugin to install it (e.g. a git path) and also any system-deps needed by that plugin, and then tell the Girder Ansible role to install/activate that plugin.

danlamanna:
The role would probably require callers to install the system packages
themselves, and just deal with git clone/girder-install.

mgrauer:
To try to add some clarity to waters I had previously muddled:

 * This should **not** install system deps, but should document that the user is responsible for getting them installed outside of the role, e.g. in their site playbook
 * This should probably not clone the plugin from git or copy it from some other location, to allow more flexibility to the user, this should also be documented that it is the user's responsibility
 * This should install the plugin, allowing for args to be passed to the install command `girder-install plugin`
 * This should **not** enable the plugin and rebuild the web client for the plugin, that is the responsibility of the user, this should also be documented

🌊

PR #2288, opened 2017-08-16
add girder_plugins argument to ansible role

opadron:
Here's my initial attempt to address #2271.

I'm not too familiar with the role, or how to go about testing my changes, so I probably need a little guidance to put the finishing touches on this.

I hope folks find this helpful.

### DIFF
--- a/devops/ansible/roles/girder/tasks/main.yml
+++ b/devops/ansible/roles/girder/tasks/main.yml
@@ -60,7 +60,7 @@
   when: girder_daemonize
 
 - name: Install Girder Plugins
-  command: "{{ girder_install_executable|default('girder-install') }} plugin {{ item.args|default('') }} {{ item.path }}"
+  command: "{{ girder_install_executable|default('girder-install') }} plugin {{ item.args|default('') }} {{ item.path|default(item) }}"
   with_items: "{{ girder_plugins }}"
 
 - name: Build Girder (web)
--- a/devops/ansible/roles/girder/tasks/main.yml
+++ b/devops/ansible/roles/girder/tasks/main.yml
@@ -60,10 +60,7 @@
   when: girder_daemonize
 
 - name: Install Girder Plugins
-  command: "{{ girder_install_executable|default('girder-install') }} plugin {{ item }}"
-  args:
-    chdir: "{{ girder_path }}"
-  when: girder_web
+  command: "{{ girder_install_executable|default('girder-install') }} plugin {{ item.args|default('') }} {{ item.path }}"
   with_items: "{{ girder_plugins }}"
 
 - name: Build Girder (web)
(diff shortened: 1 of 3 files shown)


danlamanna:
> I'm not too familiar with the role, or how to go about testing my
> changes, so I probably need a little guidance to put the finishing
> touches on this.
The easiest way is to probably fiddle with
devops/ansible/examples/girder-dev-environment/site.yml and use vagrant
up from Girder root.

> args:
>   chdir: "{{ girder_path }}"
I think this can be removed.

> when: girder_web
This can be removed too, server-side plugins should be able to be installed
as well.

> girder_plugins: []
I'm not sure if this should be a list of strings or a list of
dictionaries, so the user can specify "path" and "flags". For example a
user might want to girder-install plugin --symlink path (or any of the
other flags it takes).

opadron:
@danlamanna What do you think about making the `girder_plugins` option a list of `string|dict` objects.

i.e.:

```YAML
girder_plugins:
  - path: "/path/to/girder/plugin"
    symlink: no # the default

  - "/path/to/girder/plugin"  # same as above

  - path: "/other/path"
    symlink: yes
    other_options????
```

danlamanna:
> @danlamanna What do you think about making the `girder_plugins` option
> a list of `string|dict` objects.

I think it would be nice assuming the YAML logic for handling
strings/dicts interchangeably isn't too complicated.

I'm torn on exposing attributes like symlink as opposed to an args/flags
string where the caller passes something like:
- { "path": "/some/path", "args": "--symlink" }

The rationale is that with this level of interop we don't have the
maintenance burden of altering the ansible role every time a flag gets
added. The flipside is that the girder-install API changes pretty
rarely, so maybe it's not that big a deal.

devops/ansible/roles/girder/tasks/main.yml
```diff
@@ -59,6 +59,10 @@
 - include: daemon.yml
   when: girder_daemonize
 
+- name: Install Girder Plugins
+  command: "{{ girder_install_executable|default('girder-install') }} plugin {{ item.args|default('') }} {{ item.path }}"
```
danlamanna: After I changed `item.path` to `item.path|default(item)`, this example usage worked for me:
```
    girder_plugins:
      - path: "{{ ansible_user_dir }}/other_plugin"
        args: "--symlink"
      - "{{ ansible_user_dir }}/myplugin"
```

opadron: ooh, slick!  With that change, you can also preface the string with your args.

```YAML
girder_plugins:
  - "--symlink {{ ansible_user_dir }}/myplugin"
```
I'll leave it to others to decide if that's desirable or not! 😝

opadron:
@danlamanna LGTY?

PR #2289, opened 2017-08-17
Add a tooltip on 'users-count-icon'

Pierre-Assemat:
There is 3 new tooltips on : 
- users-count-icon
- folder-count-icon
- item-count-icon

### DIFF
--- a/clients/web/src/templates/body/userList.pug
+++ b/clients/web/src/templates/body/userList.pug
@@ -1,10 +1,12 @@
 - const isAdmin = currentUser && currentUser.get('admin');
+- const userTooltip = usersCount === 1 ? `${usersCount} total user` : `${usersCount} total users`;
 .g-user-list-header
   .g-user-pagination
   if isAdmin
-    .g-user-count-container
+    .g-user-count-container(title=userTooltip)
       i.icon-user
       span.g-user-count= usersCount
+
   .g-user-sort
   form.g-user-search-form.form-inline(role="form")
     .form-group.g-users-search-container
--- a/clients/web/src/views/widgets/HierarchyWidget.js
+++ b/clients/web/src/views/widgets/HierarchyWidget.js
@@ -476,12 +476,16 @@ var HierarchyWidget = View.extend({
         this.$('.g-child-count-container').addClass('hide');
 
         var showCounts = _.bind(function () {
+            const folderCount = formatCount(this.parentModel.get('nFolders'));
             this.$('.g-child-count-container').removeClass('hide');
-            this.$('.g-subfolder-count').text(
-                formatCount(this.parentModel.get('nFolders')));
+            this.$('.g-subfolder-count').text(folderCount);
+            const folderTooltip = folderCount === 1 ? `${folderCount} total folder` : `${folderCount} total folders`;
+            this.$('.g-subfolder-count-container').attr('title', folderTooltip);
             if (this.parentModel.has('nItems')) {
-                this.$('.g-item-count').text(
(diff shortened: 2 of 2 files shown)


clients/web/src/templates/body/userList.pug
```diff
@@ -2,7 +2,7 @@
 .g-user-list-header
   .g-user-pagination
   if isAdmin
-    .g-user-count-container
+    .g-user-count-container(title='Users count')
```
brianhelba: To make it even more obvious, how about using
```
`${usersCount} total users`
```
as the tooltip string here?

brianhelba:
@Pierre-Assemat We have similar icons on the hierarchy widget, to display folder and item counts. Can you also add similar tooltips to those? This PR would be a good place to do so.

brianhelba:
Also, to be safe, please add 
```
import 'bootstrap/js/tooltip';
```
to the top of the view file.

clients/web/src/templates/body/userList.pug
Pierre-Assemat: Yes I like that !

brianhelba:
I was actually mistaken, you don't need the `import 'bootstrap/js/tooltip';` for this style of tooltip. See https://github.com/girder/girder/pull/2295 for an attempt to resolve this confusion generally.

For now, you can remove the `import` statement.

brianhelba:
If there is one user/item/folder, the tooltip reads out in a way that's grammatically incorrect. Can you add some special case logic (a ternary `if` statement should work well) so that a count of `1` is not pluralized.

clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -476,12 +476,16 @@ var HierarchyWidget = View.extend({
         this.$('.g-child-count-container').addClass('hide');
 
         var showCounts = _.bind(function () {
+            const folderCount = formatCount(this.parentModel.get('nFolders'));
             this.$('.g-child-count-container').removeClass('hide');
-            this.$('.g-subfolder-count').text(
-                formatCount(this.parentModel.get('nFolders')));
+            this.$('.g-subfolder-count').text(folderCount);
+            var tooltip = folderCount === 1 ? `${folderCount} total folder` : `${folderCount} total folders`;
```
brianhelba: This would be cleaner as `const folderTooltip` (since we want to use `let` and `const`, not `var`).

clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -476,12 +476,16 @@ var HierarchyWidget = View.extend({
...
+            this.$('.g-subfolder-count-container').attr('title', tooltip);
             if (this.parentModel.has('nItems')) {
-                this.$('.g-item-count').text(
-                    formatCount(this.parentModel.get('nItems')));
+                const itemCount = formatCount(this.parentModel.get('nItems'));
+                this.$('.g-item-count').text(itemCount);
+                tooltip = itemCount === 1 ? `${itemCount} total item` : `${itemCount} total items`;
```
brianhelba: This would be cleaner as `const itemTooltip`, so we're not needlessly sharing a variable with the separate code above.

clients/web/src/templates/body/userList.pug
```diff
@@ -1,10 +1,12 @@
 - const isAdmin = currentUser && currentUser.get('admin');
+- const userTooltip = usersCount === 1 ? `${usersCount} total user` : `${usersCount} total users`
```
brianhelba: Since this is technically a statement, it should end with a `;`.

PR #2291, opened 2017-08-17
Ensure package-lock.json contains just Linux-only dependencies

brianhelba:
Most of the problems here are related to the fact that an optional dependency, `fsevents`, is installable on macOS and not on Linux.

By requiring the `package-lock.json` file to be generated only on Linux platforms, we:
* Prevent some installation errors on Linux, where it tries to install `fsevents` and fails
* Reduce churn of `package-lock.json`, when it's updated on different platforms
* Allow macOS users to continue using `fsevents`, albeit with a non-deterministic version


brianhelba:
Updated PR description with rationale.

PR #2292, opened 2017-08-17
Fix order of module resolution for plugins

zachmullen:
Prior to this change, each plugin was overwriting the module
resolution list and hard-coding the girder/node_modules dir
as the first element. This order was incorrect; we actually
want it to look inside its own node_modules first. After this change,
the order that will be searched for both modules and loaders is:

1. Inside the plugin's own `node_modules` dir.
~~2. Inside dependent plugins' `node_modules` dirs, in reverse toposorted
   order.~~ (removed -- this is no longer the case in my final changeset because it's not necessary)
3. Inside girder's `node_modules` dir.


zachmullen:
@jeffbaumes @ronichoudhury This is the more robust fix I came up with for the issue Jeff ran into earlier. 

Jeff, could you try out this change with your recent candela changes?

jeffbaumes:
It works on my candela changes, thanks!

grunt_tasks/plugin.js
```diff
@@ -369,16 +370,16 @@ module.exports = function (grunt) {
                                 [`girder_plugins/${plugin}/node`]: pluginNodeDir,
                                 [`girder_plugins/${plugin}`]: webClient
```
jeffbaumes: With this change, it would seem that the current aliasing of the node_modules for plugins here may no longer be necessary, since any plugins the plugin depends on should have been added to the search path. Could these aliases be removed?

jeffbaumes: Thinking another minute on this, perhaps the first `node` alias could be removed, but maybe not the second, which is for importing code from the plugin itself.

ronichoudhury:
This looks great @zachmullen. I'm withholding an actual approval since @jeffbaumes is testing in the field.

grunt_tasks/plugin.js
zachmullen: I do think we should remove the first one, since other plugins shouldn't be referencing npm deps of other plugins since that's probably undefined behavior.

However, this might be considered a breaking change, so may involve more conversation.

jeffbaumes: It went in Dec 9, 2016 in #1664, and only the latter alias is documented in https://github.com/girder/girder/blob/master/docs/plugin-development.rst#extending-the-client-side-application, but I am also ok with leaving it.

zachmullen:
I ended up changing this fix because the previous fix was incorrect -- we actually don't need to modify the global webpack options now that we have proper dynamically linked modules.

zachmullen:
This code now works, there is just one problem -- because plugins now look in their own local node_modules first instead of the core ones, candela is now finding its local version of jquery rather than dynamically linking to the same one used by core, which is required for things like the bootstrap slider / colorpicker to work. Item_tasks depends on this, and because candela imports item_tasks code statically, it breaks.

zachmullen:
I'm investigating some solutions to that dilemma. If anybody has experience with this problem, please let me know.

zachmullen:
That last commit fixes the issue I mentioned. This is now ready for another pass @jeffbaumes

grunt_tasks/plugin.js
```diff
@@ -401,13 +391,23 @@ module.exports = function (grunt) {
...
+                            // Make sure we use the core jquery rather than pulling one
+                            // from a plugin's local node modules. Multiple jquery versions
+                            // breaks due to use of side effects for jquery plugins.
+                            'jquery': path.resolve(paths.node_modules, 'jquery')
+                        },
                         modules: [
-                            path.resolve(dir, 'node_modules')
+                            path.resolve(dir, 'node_modules'),
+                            pluginNodeDir,
+                            paths.node_modules
```
jeffbaumes: So it looks like the search order is:

1. `plugins\<plugin_name>\node_modules` which exists if it is a plugin that builds independently with its own `node_modules`.
2. `node_modules\girder_plugin_<plugin_name>\node_modules` which exists if it is a plugin with a `package.json` that builds with the Girder build.
3. Girder's `node_modules` if it is a standard plugin without a `package.json`.

Is this right? Could we put a bit of documentation up to this effect?

zachmullen: > 1. plugins\<plugin_name>\node_modules which exists if it is a plugin that builds independently with its own node_modules.

This is correct.

> 2. node_modules\girder_plugin_<plugin_name>\node_modules which exists if it is a plugin with a package.json that builds with the Girder build.

My understanding is that this is the location used if a plugin sets `localNodeModules: true`, and the location itself is a bit of an internal detail, the rationale of which I don't fully understand at this time.

> Girder's node_modules if it is a standard plugin without a package.json.

This fallback path will be used regardless of whether the plugin has its own local node_modules, in case it needs to pull in something from the global node_modules. (That's probably what you meant, just wanted to clarify.)

Thinking about it a little more, I think I will actually switch the order of 1 and 2 -- I think in the case where a plugin has `localNodeModules: true` and also has its own `node_modules` directory that was installed for whatever reason, during the girder plugin build, it should pull from the `localNodeModules` stuff first, since that is definitely a girder plugin directive, and we don't know the reason `node_modules` is there.

I will add documentation about what we are guaranteeing, and how to alter it using webpack.helper.js. I'll leave off the specifics of the path used in the case of `localNodeModules: true`, since that is an implementation detail that may even change in the future.

docs/plugin-development.rst
```diff
@@ -618,66 +621,35 @@ Each type needs to be installed differently due to how node manages external pac
...
 
-      {
-          "name": "MY_PLUGIN",
-          "grunt": {
-              "dependencies": {
-                  "grunt-contrib-concat": ">=1.0.1"
-              }
-          }
-      }
+   1. The plugin's local modules directory that is created if ``npm.localNodeModules`` has
```
jeffbaumes: This documentation appears to be under the javascript code block, and renders incorrectly if you go to View in GitHub.

docs/plugin-development.rst
```diff
@@ -612,72 +609,41 @@ Each type needs to be installed differently due to how node manages external pac
...
+When you use the `import` directive within your plugin code, for example:
+
+.. code-block:: javascript
+
+    import foobar from 'foobar';
+
+The build process will search for the ``'foobar'`` module in the following locations, in order:
+
+1. The plugin's local modules directory that is created if ``npm.localNodeModules`` has
+  been set to ``true`` in the plugin configuration file.
```
jeffbaumes: This still appears just a tad off, at least on GitHub. Note the quote bar on the continuation lines:

image

zachmullen:
Fixed it up again, here's what it looks like in the built version:

screen shot 2017-09-05 at 2 30 18 pm

PR #2294, opened 2017-08-18
Smaller candela

jeffbaumes:
Depends on #2292, which should be merged first.

Creates a smaller candela plugin by only including portions of the library (vega, treeheatmap plugins) instead of the entire bundle. Reduces the minified plugin bundle from 4.8 MB to 1 MB.

Since candela uses ES6 features, we need to add a webpack rule to use babel on the candela sources.


zachmullen:
We should probably rebase this on latest master just to sanity check that it still works. Otherwise LGTM

jeffbaumes:
Thanks @ronichoudhury for your modularization work for candela.

PR #2295, opened 2017-08-18
Delegate tooltip styling to the App

brianhelba:
This delegates all styled tooltip behavior to the top-level App. This removes many lines of code, while ensuring that all tooltips are styled consistently. Previously, many tooltips were not styled with the Bootstrap effect.


brianhelba:
@girder/developers PTAL.

zachmullen:
@brianhelba will change this to remove even more code.

brianhelba:
@girder/developers PTAL. I've already manually tested this, but I'd appreciate someone else to try it out too. Let me know if any tooltips are getting stuck (which the Bootstrap ones are prone to do).

PR #2297, opened 2017-08-21
Fix rest requests honoring `type`.

manthey:
An unintentional breaking change was introduced in PR #2273.  Before, our ajax defaults (rightly or wrongly) preferred `type` over `method`, and set a default `type`.  jQuery's ajax call prefers `method`, so if a caller used either `type` *or* `method`, we would honor the caller's value.  Now, since we set a default `method`, if a caller uses `method`, it is ignored.

Further, we had previous set the dataType default to 'json'.  PR #2273 removed this, which then allows jQuery to use "Intelligent Guess".  Since this is also a breaking change, I've reverted it.

### DIFF
--- a/clients/web/src/rest.js
+++ b/clients/web/src/rest.js
@@ -104,6 +104,7 @@ setStaticRoot(
 let restRequest = function (opts) {
     opts = opts || {};
     const defaults = {
+        dataType: 'json',
         method: 'GET',
         girderToken: getCurrentToken() || cookie.find('girderToken'),
 
@@ -171,6 +172,10 @@ let restRequest = function (opts) {
         throw new Error('restRequest requires a "url" argument');
     }
     args.url = `${getApiRoot()}${args.url.substring(0, 1) === '/' ? '' : '/'}${args.url}`;
+    if (args.type) {
+        args.method = args.type;
+        delete args.method;
+    }
 
     if (args.girderToken) {
         args.headers = args.headers || {};
--- a/clients/web/src/rest.js
+++ b/clients/web/src/rest.js
@@ -172,9 +172,11 @@ let restRequest = function (opts) {
         throw new Error('restRequest requires a "url" argument');
     }
     args.url = `${getApiRoot()}${args.url.substring(0, 1) === '/' ? '' : '/'}${args.url}`;
+    // jQuery's ajax method prefers `method` over `type`.  Since we specify a
+    // default `method`, if the caller supplies a `type`, use it instead.
     if (args.type) {
         args.method = args.type;
-        delete args.method;
+        delete args.type;
     }
 
     if (args.girderToken) {
--- a/clients/web/src/rest.js
+++ b/clients/web/src/rest.js
@@ -104,7 +104,6 @@ setStaticRoot(
 let restRequest = function (opts) {
     opts = opts || {};
     const defaults = {
-        dataType: 'json',
         // the default `method` (aliased as `type`) is 'GET'
         girderToken: getCurrentToken() || cookie.find('girderToken'),
 
--- a/clients/web/src/rest.js
+++ b/clients/web/src/rest.js
@@ -105,7 +105,7 @@ let restRequest = function (opts) {
     opts = opts || {};
     const defaults = {
         dataType: 'json',
-        method: 'GET',
+        // the default `method` (aliased as `type`) is 'GET'
         girderToken: getCurrentToken() || cookie.find('girderToken'),
 
         error: (error, status) => {
@@ -172,12 +172,6 @@ let restRequest = function (opts) {
         throw new Error('restRequest requires a "url" argument');
     }
     args.url = `${getApiRoot()}${args.url.substring(0, 1) === '/' ? '' : '/'}${args.url}`;
-    // jQuery's ajax method prefers `method` over `type`.  Since we specify a
-    // default `method`, if the caller supplies a `type`, use it instead.
-    if (args.type) {
-        args.method = args.type;
-        delete args.type;
-    }
 
     if (args.girderToken) {
         args.headers = args.headers || {};


manthey:
Ping @brianhelba

clients/web/src/rest.js
```diff
@@ -171,6 +172,10 @@ let restRequest = function (opts) {
         throw new Error('restRequest requires a "url" argument');
     }
     args.url = `${getApiRoot()}${args.url.substring(0, 1) === '/' ? '' : '/'}${args.url}`;
+    if (args.type) {
+        args.method = args.type;
+        delete args.method;
```
zachmullen: Might be worth a short comment here, either briefly explaining this block or just linking to this PR

brianhelba:
I'm curious, is there a place where the "intelligent guess" logic for `dataType` is failing? All of Girder's non-download endpoints that would be used with `restRequest` *should* be returning `Content-Type: application/json`?

clients/web/src/rest.js
```diff
@@ -171,6 +172,12 @@ let restRequest = function (opts) {
         throw new Error('restRequest requires a "url" argument');
     }
     args.url = `${getApiRoot()}${args.url.substring(0, 1) === '/' ? '' : '/'}${args.url}`;
+    // jQuery's ajax method prefers `method` over `type`.  Since we specify a
+    // default `method`, if the caller supplies a `type`, use it instead.
+    if (args.type) {
```
brianhelba: `$.ajax` also has a default of `method: 'GET'`. Instead of this block, could we just remove line 108, then allow `$.ajax` to overwrite the default with either `type` or `method`, if they're supplied?

manthey:
I have no idea if "Intelligent Guess" ever chooses anything different than json for our purposes -- but changing this is a breaking change and shouldn't be done until Girder 3 (if ever).  Some plugin may be relying on the `dataType` being json.

clients/web/src/rest.js
manthey: We can.  Or we could remove line 108 and add one line that says `args.method = args.method || args.type || 'GET'` to be explicit about *our* choices.  If we want are trying to use `$.ajax` defaults, then perhaps we shouldn't deprecate the use of `path`, since `$.ajax` hasn't done so.

brianhelba: The original intent of #2273 was to bring the behavior of `restRequest` closer to `$.ajax`, and make it documented and explicit that `restRequest` is a thin wrapper around `$.ajax` (previously, some code was abusing this behavior, even though it was never documented). For libraries like `jQuery` that are well documented, broadly understood, and "leaked" to effectively be part of Girder's API, I think it's more sensible to delegate to their defaults whenever possible, which both decreases the LOC (and associated maintenance burden) in Girder and makes it easier for new developers to learn Girder, as they're more likely to understand `jQuery` or be able to effectively search StackOverflow, etc. for help.

@manthey What do you think of this?

manthey: The intent is fine -- we can remove the default of `method`/`type` being `GET`, since it is the default of `$.ajax` (but I'll add a comment to that effect).  Though we still need to leave the default `dataType` as `json`, as removing that is a breaking change.

brianhelba:
Given that `restRequest` is intended to be used strictly with the Girder API, I think switching to "Intelligent Guest" should be compatible, as effectively we're just shifting logic from the client to the server. All existing versions of Girder should already be setting `Content-Type: application/json` on all non-binary-returning REST endpoints; if they're not, then this change has exposed a bug in upstream Girder, but not broken any comparability guarantees.

The advantage of the change is that the client actually becomes more permissive and flexible, as it can potentially start to fetch endpoints of other `Content-Types` (e.g. for a plugin's XML or binary endpoint) without having to explicitly worry about setting `dataType`, making it easier to use.

clients/web/src/rest.js
brianhelba: A comment sounds good, thanks.

manthey:
Changing this is a breaking change -- there may be a plugin endpoint that doesn't emit json (and calls `setResponseHeader`) which now throws an error when called elsewhere, but changing this to "Intelligent Guess" would return valid xml or text.  The current released behavior is for the error to occur (as the caller is required to override the `dataType`).  This would make it succeed where it failed before.  It seems a fundamental part of our current API.

It may be a *good* change, but that doesn't make it non-breaking.

brianhelba:
Can you point me to an example of a plugin that's impacted by this change. I'm fine with doing the "right" thing in isolation without such an example, but it would help to see how downstream developers are actually impacted by this.

I see this change as a backwards-compatible new feature, perhaps to be included in the CHANGELOG as:
> `restRequest` is now more permissive about the endpoints it can be called on, as `dataType` is now an optional parameter for non-JSON endpoints

Would you see this release note as requiring a major version bump?

As I see it, changes to existing functions are not breaking *per se*, rather, a expectation (which ideally was already documented) of downstream developers needs to be violated in a way that negatively impacts their code. I don't see how making a function more permissive about how it's able to be called qualifies as a breakage.

manthey:
I don't have a use case which would break because of this.  But I also don't know of all of the plugins that exist for Girder.  The slicer_cli_web plugin can return xml from some dynamically created endpoints (the cli xmlspec endpoints, specifically).  When it consumes them, it explicitly specifies a `dataType` of xml.  If the xmlspec is consumed by another plugin, we could be changing its behavior.

We've been cautious about breaking changes in other spots.

I'm happy to go with a majority consensus, if anyone else would case to weigh in on this.

brianhelba:
Just to be clear, as I understand it, the changed behavior (including the example of anyone consuming an xmlspec) is that `restRequest` calls which did not set `dataType` will have previously failed with [a `'parsererror'`](https://github.com/girder/girder/blob/7c3a9cfae7bea515c243b7cbaa3a60e7ea7f5453/clients/web/src/rest.js#L138), but will now work correctly.

I'm also happy to get some others' opinions on this.

manthey:
@zachmullen Would you care to weigh in on `dataType`?

zachmullen:
I don't understand every nuance of jquery's behavior w.r.t. `dataType`, but from reading their docs, it defaults to using the content-type response header, which to me is the most sensible default for `restRequest` as well.

manthey:
Okay.  `dataType` is now unspecified (as done in PR #2273).  Hopefully nothing was depending on the explicit value of `json`.

PR #2298, opened 2017-08-23
Render item task descriptions as markdown

zachmullen:
Since item descriptions are rendered as markdown, and the auto-generated slicer CLI descriptions contain markdown, we should render these as markdown as well. After, looks like:

screen shot 2017-08-23 at 1 56 11 pm


zachmullen:
In case anyone was wondering why I moved the elements around: the browser does not allow `<a>` tags to contain `<p>` tags, so we can no longer nest the description inside the link.

PR #2300, opened 2017-08-24
Update front page text

ronichoudhury:
This is part of the "DKC facelift".

This PR is going to Girder because it's a general improvement on the front page content that every instance of Girder should have when built or installed out of the box. All the information currently shown is here, just restructured a little bit to ease reading.

There are also some links to various media we have published about Resonant and Girder, as well as a mention of Kitware and our consulting services (while of course an instance can override this text by using the homepage plugin).

To complete the "facelift" will require support from #2283 and related, upcoming PRs in the same vein as that one:
- [ ] Need to be able to independently set the banner "brand name" and the "title welcome text"
- [ ] Need to be able to set the "subtitle" text
- [ ] Need a way to customize the opening paragraph of the content itself (the part that currently reads "**Welcome to Girder!**")
- [ ] Need a way to set the banner background color

### DIFF
--- a/clients/web/src/templates/body/frontPage.pug
+++ b/clients/web/src/templates/body/frontPage.pug
@@ -67,7 +67,7 @@
         | visualization. To learn more, you can read our series of
         a(target="_blank", href="https://blog.kitware.com/tag/resonant/")  blog posts
         |  about Resonant, or
-        a(target="_blank", href="http://www.kitware.com/company/contact_kitware.php")  contact us
+        a(target="_blank", href="http://www.kitware.com/contact-us")  contact us
         |  to learn how we can help you solve your data problems.
 
       if versionInfo.git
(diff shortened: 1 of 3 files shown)


ronichoudhury:
I need some help diagnosing the failing tests. Locally, both tests (`web_client_version` and `web_client_routing`) fail with `AttributeError: 'module' object has no attribute 'web_client_test'`.

clients/web/src/templates/body/frontPage.pug
```diff
@@ -7,43 +7,75 @@
...
         | .
+
+      li
+        | Girder is a part of
+        a(target="_blank", href="http://resonant.kitware.com")  Resonant
+        | , Kitware's open-source platform for data management, analytics, and
+        | visualization. To learn more, you can read our series of
+        a(target="_blank", href="https://blog.kitware.com/tag/resonant/")  blog posts
+        |  about Resonant, or
+        a(target="_blank", href="http://www.kitware.com/company/contact_kitware.php")  contact us
```
danlamanna: Should this link to https://www.kitware.com/contact-us/ instead? (Your link is a 301 for this).

ronichoudhury: Good catch, thanks @danlamanna

Issue #2299, opened 2017-08-24
Should be able to provide metadata when creating an item


cjh1:
and add parameter to girder_client ...

zachmullen:
Thanks to @Xarthisius this is now possible: https://github.com/girder/girder/commit/c0925ba59516e5201b7553d7a1e79c77e375186c

We still need to add it to the girder_client.

cjh1:
Excellent, i was looking at my branch which is a little out of date. I can add it to the client.

manthey:
This was done ages ago.

PR #2301, opened 2017-08-24
Add metadata parameter to client creation functions

cjh1:


cjh1:
Fixes #2299

clients/python/girder_client/__init__.py
```diff
@@ -618,15 +618,24 @@ def listFile(self, itemId, limit=None, offset=None):
...
+                   metadata=None):
         """
         Creates and returns an item.
+
+        :param parentFolderId: the folder this item should be created in.
+        :param name: the item name.
+        :param description: a description of the item.
+        :param reuseExisting: whether to return an existing item if one with
+            same name already exists.
+        :param metadata: JSON metadata string to set on item.
```
jbeezley: What do you think about making this take a dict rather than serialized json?

cjh1: I don't have a problem with that, just trying to be consistent with existing functions, such as [setFolderAccess](https://github.com/girder/girder/blob/master/clients/python/girder_client/__init__.py#L779)

jbeezley: I see.  I don't have much experience using girder_client.  Consistency is important, but it would be good to migrate to a more pythonic API going forward.  Perhaps we can make an issue to jsonify these kinds arguments if the value passed is not a string.  For now, this works for me.

cjh1: I agree it would be much more intuitive if these were dicts.

zachmullen: I'd be in favor of testing the type, and if it's an instance of `six.string_types` we can pass it as is, otherwise we can call `json.dumps` on it. We could do the same for `setFolderAccess`.

cjh1: I can do that

cjh1: Done

PR #2304, opened 2017-08-25
Ansible - Source virtualenv in login shell

sutartmelson:
This relocates the virtualenv folder to be more visible and sources the virtual environment upon login.

### DIFF
--- a/devops/ansible/examples/girder-dev-environment/site.yml
+++ b/devops/ansible/examples/girder-dev-environment/site.yml
@@ -4,7 +4,7 @@
   vars:
     girder_update: no
     girder_force: no
-    girder_virtualenv: "{{ ansible_user_dir }}/virtualenvs/girder"
+    girder_virtualenv: "{{ ansible_user_dir }}/.virtualenvs/girder"
     girder_always_build_assets: yes
     girder_web_extra_args: "--dev --all-plugins"
   pre_tasks:
--- a/devops/ansible/examples/girder-dev-environment/site.yml
+++ b/devops/ansible/examples/girder-dev-environment/site.yml
@@ -4,7 +4,7 @@
   vars:
     girder_update: no
     girder_force: no
-    girder_virtualenv: "{{ ansible_user_dir }}/.virtualenvs/girder"
+    girder_virtualenv: "{{ ansible_user_dir }}/virtualenvs/girder"
     girder_always_build_assets: yes
     girder_web_extra_args: "--dev --all-plugins"
   pre_tasks:
@@ -70,3 +70,11 @@
       become: yes
       become_user: root
       when: ansible_os_family == "RedHat"
+
+    - name: Activate virtualenv on login
+      lineinfile:
+        dest: ~/.profile
+        line: 'source {{girder_virtualenv}}/bin/activate'
+        insertafter: 'EOF'
+        regexp: 'source {{girder_virtualenv}}/bin/activate'
+        state: present


devops/ansible/examples/girder-dev-environment/site.yml
```diff
@@ -4,7 +4,7 @@
   vars:
     girder_update: no
     girder_force: no
-    girder_virtualenv: "{{ ansible_user_dir }}/.virtualenvs/girder"
+    girder_virtualenv: "{{ ansible_user_dir }}/virtualenvs/girder"
```
zachmullen: I believe the `.virtualenvs` directory is a convention of [virtualenvwrapper](https://bitbucket.org/dhellmann/virtualenvwrapper), however I don't think the wrapper is installed by our provisioning scripts.

sutartmelson: I suppose the sourcing of the virtual environment is the essential part of this PR. I just found it confusing when I would `vagrant ssh` in. I didn't know girder was installed into a virtualenv, nor is the virtualenv visible right away. 

Automatically sourcing the virtualenv makes it simpler. Would you ever want to ssh into it and not be in the virtualenv?

zachmullen: I'm all for auto-sourcing the virtualenv. I'd just probably leave the directory as `.virtualenvs` since that's a common convention.

devops/ansible/examples/girder-dev-environment/site.yml
```diff
@@ -70,3 +70,11 @@
       become: yes
       become_user: root
       when: ansible_os_family == "RedHat"
+
+    - name: Activate virtualenv on login
+      lineinfile:
+        dest: ~/.profile
+        line: 'source {{girder_virtualenv}}/bin/activate'
+        insertafter: 'EOF'
+        regexp: 'source {{girder_virtualenv}}/bin/activate'
```
brianhelba: This `regexp` parameter is not necessary, as [an omitted `regexp` will cause the module to match the literal `line`](https://github.com/ansible/ansible/blob/v2.3.2.0-1/lib/ansible/modules/files/lineinfile.py#L385). Futhermore, explicitly passing the `line` value as `regexp` will cause the value to be interpreted as a regex, which we definitely are not expecting. In this case, it will just match the `.` as a wildcard, but any other regex-special characters in the path would cause it to break. Furthermore, explicitly passing this argument conveys the semantics that both `line` and `regexp` are required, even if they are the same value, which may entice other developers to copy-paste this pattern and encounter into the special-regex-character issue.

devops/ansible/examples/girder-dev-environment/site.yml
```diff
@@ -70,3 +70,11 @@
       become: yes
       become_user: root
       when: ansible_os_family == "RedHat"
+
+    - name: Activate virtualenv on login
+      lineinfile:
+        dest: ~/.profile
+        line: 'source {{girder_virtualenv}}/bin/activate'
+        insertafter: 'EOF'
```
brianhelba: We don't actually care where the line is inserted, so it may be best to omit this line, to allow `lineinfile` to apply its own best judgement (which defaults to 'EOF' anyway).

devops/ansible/examples/girder-dev-environment/site.yml
```diff
@@ -70,3 +70,11 @@
...
+        regexp: 'source {{girder_virtualenv}}/bin/activate'
+        state: present
```
brianhelba: Note, this line is also unnecessary (since the default is `present`), but since we do really care that it's this value, I have no problem being explicit.

devops/ansible/examples/girder-dev-environment/site.yml
```diff
@@ -70,3 +70,11 @@
       become: yes
       become_user: root
       when: ansible_os_family == "RedHat"
+
+    - name: Activate virtualenv on login
+      lineinfile:
+        dest: ~/.profile
```
brianhelba: We should not use `~` here, since that's a Bash construct. Ansible provides the fact `ansible_user_dir`, which represents the location more robustly.

devops/ansible/examples/girder-dev-environment/site.yml
```diff
@@ -70,3 +70,11 @@
...
+        line: 'source {{girder_virtualenv}}/bin/activate'
```
brianhelba: Ansible style dictates that we pad the variable with spaces like: `{{ girder_virtualenv }}`.

PR #2305, opened 2017-08-25
Webpack helpers for plugins should only affect that plugin

zachmullen:
When implementing the geojs plugin I discovered a bug related to the webpack helper files. To build the geojs plugin, I need to use one containing some `externals` declarations. However, I noticed that after doing that, other plugins that contained those symbols were broken at runtime (namely the jobs plugin which also references d3, but *not* as an external symbol).

After looking at plugin.js, I saw that the contents of `webpack.helper.js` for plugins was being merged into the *global* webpack options, rather than the webpack options for building that specific plugin. In fact, there was no way via the helper to modify anything other than the global options, which was inadvertently clobbering other plugins.

This PR contains a working solution, but may be controversial -- it now unconditionally merges the helper configuration with the plugin-specific config, and disallows modifying the global options. I would contend, as a straw man, that it shouldn't be necessary for plugins to modify the global stuff, and should only be modifying how they themselves are built. This is because we now link plugins dynamically, so their build processes are no longer intertwined.

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -389,6 +389,12 @@ module.exports = function (grunt) {
                     entry: {
                         [helperConfig.pluginEntry]: [main]
                     },
+                    module: {
+                        // We set these since webpack.helper.js functions may
+                        // expect these properties to be defined.
+                        loaders: [],
+                        rules: []
+                    },
                     output: {
                         path: path.join(paths.web_built, 'plugins', plugin),
                         filename: `${output}.min.js`,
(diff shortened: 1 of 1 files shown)


brianhelba:
This is absolutely a breaking change, but I totally agree with what it's doing, and was hoping to see it happen in the 3.0 release. The only question is whether we consider the current behavior to be so egregious as to make the change now, which I'm not sure how I feel about yet.

zachmullen:
I personally found the existing behavior to be so counterintuitive as to consider it a bug, and I don't know of any practical way that downstreams would need or want to modify the global options, and furthermore I actually don't think they should even be allowed to. So, while this is breaking, I think it's only breaking buggy behavior, and I'd want it to go in now.

brianhelba:
I'm fine with the change.

jbeezley:
I am also fine with this.

zachmullen:
Thanks for taking a look all, that seems like a good quorum.

PR #2309, opened 2017-08-27
Fixes #2307. Support local uploads into S3 assetstores

### LINKED ISSUE #2307: Creating thumbnails into S3 assetstore fails
```
{"message": "AttributeError: AttributeError(\"StringIO instance has no attribute 'getSize'\",)", "trace": [["girder/api/rest.py", 567, "endpointDecorator", "val = fun(self, args, kwargs)"], ["girder/api/rest.py", 1159, "POST", "return self.handleRoute(method, path, params)"], ["girder/api/rest.py", 905, "handleRoute", "val = handler(**kwargs)"], ["girder/api/access.py", 63, "wrapped", "return fun(*args, **kwargs)"], ["girder/api/rest.py", 417, "wrapped", "val = fun(*args, **kwargs)"], ["girder/api/describe.py", 637, "wrapped", "return fun(*args, **kwargs)"], ["/home/ubuntu/girder/plugins/thumbnails/server/rest.py", 65, "createThumbnail", "return utils.scheduleThumbnailJob(file, attachToType, attachToId, user, width, height, crop)"], ["/home/ubuntu/girder/plugins/thumbnails/server/utils.py", 19, "scheduleThumbnailJob", "jm.scheduleJob(job)"], ["/home/ubuntu/girder/plugins/jobs/server/models/job.py", 311, "scheduleJob", "events.trigger('jobs.schedule', info=job)"], ["girder/events.py", 291, "trigger", "handler['handler'](e)"], ["/home/ubuntu/girder/plugins/jobs/server/__init__.py", 43, "scheduleLocal", "fn(job)"], ["/home/ubuntu/girder/plugins/thumbnails/server/worker.py", 39, "run", "newFile = createThumbnail(**job['kwargs'])"], ["/home/ubuntu/girder/plugins/thumbnails/server/worker.py", 121, "createThumbnail", "attachParent=True)"], ["girder/models/upload.py", 106, "uploadFromFile", "upload = self.handleChunk(upload, six.BytesIO(data))"], ["girder/models/upload.py", 144, "handleChunk", "upload = adapter.uploadChunk(upload, chunk)"], ["girder/utility/s3_assetstore_adapter.py", 177, "uploadChunk", "return self._proxiedUploadChunk(upload, chunk)"], ["girder/utility/s3_assetstore_adapter.py", 260, "_proxiedUploadChunk", "size = chunk.getSize()"]], "type": "internal"}
```

zachmullen:
The S3 assetstore's proxied upload code expected the stream object to be a `RequestBodyStream`. In the local upload case, however, we were not passing one.

### DIFF
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -26,6 +26,7 @@ from girder.api import rest
 from girder.constants import SettingKey
 from girder.utility import assetstore_utilities
 from .model_base import Model, GirderException, ValidationException
+from girder.utility import RequestBodyStream
 from girder.utility.progress import noProgress
 
 
@@ -103,7 +104,7 @@ class Upload(Model):
             if not data:
                 break
 
-            upload = self.handleChunk(upload, six.BytesIO(data))
+            upload = self.handleChunk(upload, RequestBodyStream(six.BytesIO(data), size))
 
         return upload
 
@@ -417,12 +418,12 @@ class Upload(Model):
             else:
                 chunk = data
             if len(chunk) >= chunkSize:
-                upload = self.handleChunk(upload, six.BytesIO(chunk))
+                upload = self.handleChunk(upload, RequestBodyStream(six.BytesIO(chunk), len(chunk)))
                 progress.update(increment=len(chunk))
                 chunk = None
 
         if chunk is not None:
-            upload = self.handleChunk(upload, six.BytesIO(chunk))
+            upload = self.handleChunk(upload, RequestBodyStream(six.BytesIO(chunk), len(chunk)))
             progress.update(increment=len(chunk))
 
         return upload


brianhelba:
Just from inspection, it seems like one previous problem was that `S3AssetstoreAdapter.uploadChunk`'s internal helpers were calling `.getSize()` on the chunk. While this approach solves the immediate issue at hand, I'm concerned that there are more general problems lurking:
* [`AbstractAssetstoreAdapter.uploadChunk`](https://github.com/girder/girder/blob/master/girder/utility/abstract_assetstore_adapter.py#L175) claims to accept a file-like object for its `chunk` parameter.
* [`FilesystemAssetstoreAdapter.uploadChunk`](https://github.com/girder/girder/blob/master/girder/utility/filesystem_assetstore_adapter.py#L150) seems to be willing to take a `chunk` of text types too.
  * This code checks the size of its `chunk` inputs with the [`AbstractAssetstoreAdapter.getChunkSize`](https://github.com/girder/girder/blob/master/girder/utility/abstract_assetstore_adapter.py#L308) method, which claims to work on chunks of many different types. However, `AbstractAssetstoreAdapter.getChunkSize` actually *fails* (it gives up and returns `None`) on `RequestBodyStream` chunk types...though this doesn't matter, since the failure just causes sanity checks to be skipped.
*  `S3AssetstoreAdapter.uploadChunk` does not accept general file-file objects (breaking its polymorphic contract) for a `chunk`, but implicitly (with no documentation) requires it to specifically be a  `RequestBodyStream`....except it also allows `chunk` to be a text string that doesn't actually contain chunk data, in which case the method's behavior is totally altered.

I think a truly general fix is too ambitious here, but I'm concerned that this fix just further entrenches the bug in `S3AssetstoreAdapter.uploadChunk`, where it does not actually allow file like objects. As a slightly more maintainable alternative fix, can I suggest:
* Fix `AbstractAssetstoreAdapter.getChunkSize` to actually return the size of `RequestBodyStream` chunk types.
* Use `AbstractAssetstoreAdapter.getChunkSize` instead of `.getSize` in the internal helpers of `S3AssetstoreAdapter.uploadChunk`.
* Remove this PR's re-wrapping of `BytesIO` objects -> `RequestBodyStream` from the `Upload` model, since according to the contract of `AbstractAssetstoreAdapter.uploadChunk`, there's nothing wrong with passing a `BytesIO` object.
* Add some code to the beginning of `S3AssetstoreAdapter.uploadChunk`, whereby if it's not already getting a `RequestBodyStream` `chunk`, it will wrap `chunk` into `RequestBodyStream` (setting the size as it does so).

brianhelba:
@zachmullen What do you think?

zachmullen:
The main issue at the heart of this is that file-like objects do not have a general way of reporting their size. There's `seek` and `tell` in some cases, but those methods don't exist for all file-like objects, and in particular, the ones we use are *not* seekable.

Even the term "file-like object" isn't well-defined. In our case we only provide a `read`able file object, for instance. I don't believe there's any general mechanism baked into python to determine whether something is a file-like object; the closest thing for our purposes would probably be to check if it has a callable attribute `read`.

All that said, I'd be ok with multiple different options.

1.

> Fix AbstractAssetstoreAdapter.getChunkSize to actually return the size of RequestBodyStream chunk types.

I'm happy to do this, it seems like a generally good idea. I can do it on this PR or a different one.

2. Fix the documentation surrounding the assetstore adapter abstraction to disambiguate the type we are actually passed, since `file-like` is ambiguous. In the abstract case, we should guarantee that a `RequesautBodyStream` is the type that will be passed. The S3 case is a special one where we allow a differently-structured request because it's only receiving a small JSON object, and that still fits with the abstraction since we allow different behaviors for specific assetstore types.

brianhelba:
Indeed, Python has historically been very vague about "file-like objects", and the definition has often ended up as a matter of duck-typing with the `read` method.

However, they've made efforts to fix this with the [`io` module](https://docs.python.org/2/library/io.html), which provides actual abstract implementation guarantees. To take full advantage of this module, we should:
* Replace all uses of `open(...)` with [`io.open`](https://docs.python.org/2/library/io.html#io.open), which takes the same signature but returns an object with a `io`-module-compatible type in Python2.
* Change [`FileHandle`](https://github.com/girder/girder/blob/master/girder/utility/abstract_assetstore_adapter.py#L29) and [`RequestBodyStream`](https://github.com/girder/girder/blob/master/girder/utility/__init__.py#L149) to inherit from [`io.IOBase`](https://docs.python.org/2/library/io.html#io.IOBase) (or some subclass), and ensure that their methods conform to the abstract specification.

Obviously this is way out of scope of this PR, but it is worthwhile future work.

---

As for the type that `AbstractAssetstoreAdapter.uploadChunk` should expect, I think it makes no sense that this should expect a `RequestBodyStream` in general. From the name, `RequestBodyStream` would seem to represent *request* bodies, and it's legitimate to want to upload far more general things than just request bodies into assetstores.

I see the fundamental difficulty as the fact that `S3AssetstoreAdapter` needs to know the size of its `chunk`. This is at odds with the fact that file-like-objects / `io.IOBase` objects are streams, and deliberately do not make guarantees about their size by design. Accordingly, we need to determine whether `AbstractAssetstoreAdapter.uploadChunk` is *actually* indented to operate on *real streams*. If the behavior of `AbstractAssetstoreAdapter.uploadChunk` implementations need to know the size of their `chunk`, then `chunk` should not be a stream, it should be (or stay as) a `bytes` string. The act of converting a `bytes` string into a `BytesIO` is essentially agreeing to give up random access capabilities in exchange for the ability to be used as a more general stream.

---

At this point, I've actually come full circle to want to merge the PR as-is. While the current expectations are totally inconsistant, at least they are also undocumented, so we have more flexibility to "bugfix" them away in the future. I'm concerned that adding more documentation (such as an explicit requirement for `chunk` to be `RequesautBodyStream`) is technically still going in the wrong direction, but will also make things harder to fix outside of a major version release.

zachmullen:
> At this point, I've actually come full circle to want to merge the PR as-is. While the current expectations are totally inconsistant, at least they are also undocumented, so we have more flexibility to "bugfix" them away in the future.

+1. We should make some stronger type guarantees, but that can be done on a future PR.

PR #2310, opened 2017-08-28
improve argument handling in girder ansible module

opadron:
Specifically, fixes two issues in how the girder ansible module handles parameters given to its `setting` mode.

First, the module fails to handle the case of None for `value` (default value) and of None for `existing_value` (returned from girder server).

Second, the module tries to compare the new and existing values of a setting for the sake of idempotency, but seems to use a few simple heuristics that assume certain hashability properties of the two values.  Specifically, it seems to require that the values be limited to simple, or first-order compoundness.  For example, it accepts a compound object of scalars, but not a compound object of such compound objects.  This causes trouble when combined with Ansible's insistance on deserializing anything that looks like JSON at a configuration document's leaves.

To address both these issues, this commit makes two major changes.  First, it adds a function for canonicalizing the values.  This canonicalization is performed by deserializing valid JSON strings into Python objects, and/or traversing such objects looking for valid JSON fragments to deserialize.  Second, the comparison heuristics are replaced with a full deep comparison to most accurately determine if a system setting had, in fact, been changed.

### DIFF
--- a/devops/ansible/roles/girder/library/girder.py
+++ b/devops/ansible/roles/girder/library/girder.py
@@ -877,7 +877,7 @@ EXAMPLES = '''
 
 
 def unjsonify(a):
-    '''Convert json parts to python objects.
+    """Convert json parts to python objects.
 
     Tries to convert a json string or a compund object consisting of partial
     json strings to a full python object.  Returns either a json scalar (string,
@@ -887,7 +887,7 @@ def unjsonify(a):
     Note that this function does not detect cycles in an object graph and, if
     provided an object with one, will run until out of memory, at the stack
     limit, or until at some other run-time limit.
-    '''
+    """
 
     # if string, try to loads() it
     if isinstance(a, basestring):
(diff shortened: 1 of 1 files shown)


kotfic:
@opadron I'm happy to review this,  Is it blocking anything critical?  It may take me a day or two to get to it.

kotfic:
@opadron Mostly this makes sense to me,  except I'm not clear on why we can't use just ``` unjsonify(existing_value) == unjsonify(value) ```? Is ```deep_equality``` really necessary?

devops/ansible/roles/girder/library/girder.py
```diff
@@ -876,6 +876,84 @@
...
+            a = json.loads(a)
+            # pass-through to below
+        except ValueError:
+            return a
+
+    if isinstance(a, list):
+        return [unjsonify(x) for x in a]
+
+    if isinstance(a, dict):
+        return {str(k):unjsonify(v) for k,v in a.items()}
```
kotfic: can we also just return a if none of those three is true

opadron:
@kotfic thanks for looking into this.  It's not a hard block, but getting this merged would make some things that I am working on much easier.

> @opadron Mostly this makes sense to me, except I'm not clear on why we can't use just unjsonify(existing_value) == unjsonify(value)? Is deep_equality really necessary?

well, you see...

    ... **experiments in an interactive Python prompt** ...

...wait, is `==` a deep comparison!?  Since when!?  Awesome! :+1:

I'll go ahead and throw a few more cleanup commits that should spruce this right up.

opadron:
@kotfic LGTY?

opadron:
Hooray! 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #2311, opened 2017-08-28
Fix folder creator value.

manthey:
When creating folders specified from user entries, all folders from the preceding user are assigned the next user as the creator, not the current user.


manthey:
@jbeezley This doesn't affect PR #2166.

PR #2312, opened 2017-08-28
When checking if a resource path is accessible, check the access on each level of the path, not just the top and bottom.

manthey:


jbeezley:
This is probably a worthwhile change, but it does elude to the ambiguity of what should happen when a public folder is contained in a private folder.  We have discussed it before, but I don't know that we ever came to a conclusion.  Currently, you can request the document itself via the API, but you get a REST error when trying view it in the web client.

manthey:
I think in the context of a resource path, that you shouldn't be able to reach a public folder in a private folder you don't have access to -- it would confirm the existence of the private folder, which we otherwise don't want to do (I think).

Perhaps we need a resource path that has some sort or shortened form that allows access to items and folders bypassing the parents (e.g., `/folder/<id>`, `/item/<id>`, `/file/<id>`).  This would then validate access on just that folder (or on the folder containing the specified item or file).  For folders and items, their child folders, items, and files (as appropriate) could be reachable below the `/<resource>/<id>` (e.g., `/folder/<id>/<folder name>/<item name>/<file name>`).  Have such id-based references would at least provide one way to handle that file/item/folder names might not be unique.

PR #2314, opened 2017-08-29
Make getmembers() call python 2/3 compatible

kotfic:
See: https://stackoverflow.com/questions/17019949/why-is-there-a-difference-between-inspect-ismethod-and-inspect-isfunction-from-p

### DIFF
--- a/devops/ansible/roles/girder/library/girder.py
+++ b/devops/ansible/roles/girder/library/girder.py
@@ -20,7 +20,7 @@
 import json
 import os
 import requests
-from inspect import getmembers, ismethod, getargspec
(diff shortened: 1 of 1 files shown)


kotfic:
Sorry,  more context -

As of ansible 2.2 it is possible to run ansible with python3.  We do some magic to inject various methods from the GirderClientModule into the argument_spec.  this previously relied on ismethod() which is no longer a valid concept in python 3.

@danlamanna can you ping me on the status of testing this library module?  It would be nice to ensure it works in a python 3 environment if that isn't too burdensome

PR #2315, opened 2017-08-30
Fix a bug in `girder-install --watch-plugin`

jbeezley:
This fixes a bug in `--watch-plugin` caused by https://github.com/girder/girder/pull/2305

```
Loading "plugin.js" tasks...ERROR
>> TypeError: Cannot read property 'module' of undefined
```

raised at `grunt_tasks/plugin.js:481` because when a plugin is configured but not built, the configuration object is not created.  This moves the generation of the configuration outside of the `buildPlugin` block.  Because it still isn't added to the `default` object, the plugin will not be built.

### DIFF
--- a/grunt_tasks/plugin.js
+++ b/grunt_tasks/plugin.js
@@ -449,8 +449,19 @@ module.exports = function (grunt) {
                         dependencies: ['build'] // plugin builds must run after core build
                     }
                 };
+
+                var baseConfig = configOpts.webpack[`${output}_${plugin}`];
+                var newConfig = webpackHelper(baseConfig, helperConfig);
+                if (_.has(newConfig.module, 'loaders')) {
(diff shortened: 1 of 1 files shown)


zachmullen:
I haven't tested this, but it looks like it should work to fix the issue. However, is it the right fix? Now that webpack.helper.js is only allowed to modify the plugin's own webpack build config, does it make sense to read the webpack helper file at all if `buildPlugin` is false?

jbeezley:
There is probably a better fix, but this was the most obvious to me.  The only side effect of that portion of code is adding a new section to the grunt config at `config.webpack.${plugin}_${output}`.  Because nothing is added to run that webpack task, it doesn't have any effect on the actual build.

jbeezley:
@zachmullen PTAL

PR #2317, opened 2017-08-31
Allow The Browser Widget Deferred Validation

sutartmelson:
While maintaining its original functionality of accepting a validation function that takes input and returns a string, the BrowserWidget can now accept a validation function that takes input and returns a promise as to allow deferred validation.

### DIFF
--- a/clients/web/src/views/widgets/BrowserWidget.js
+++ b/clients/web/src/views/widgets/BrowserWidget.js
@@ -49,7 +49,7 @@ var BrowserWidget = View.extend({
         // store options
         settings = settings || {};
         this.titleText = settings.titleText || 'Select an item';
-        this.validate = settings.validate || function () {};
+        this.validate = settings.validate || _.constant($.Deferred().resolve().promise());
         this.helpText = settings.helpText;
         this.showItems = settings.showItems;
         this.showPreview = _.isUndefined(settings.showPreview) ? true : !!settings.showPreview;
--- a/clients/web/src/views/widgets/BrowserWidget.js
+++ b/clients/web/src/views/widgets/BrowserWidget.js
@@ -152,16 +152,7 @@ var BrowserWidget = View.extend({
     },
 
     /**
-     * _validate independently validates the input-element and the selected-model.
-     * The 4 possible outcomes of this validation are as follows...
-     *
-     * Case |  input-element  |  selected-model
-     *  1.  |      Valid      |      Valid
-     *  2.  |      Valid      |     Invalid
-     *  3.  |     Invalid     |      Valid
-     *  4.  |     Invalid     |     Invalid
-     *
-     * The code path for each case is commented inline
+     * Independently validate the input-element and the selected-model.
      */
     _validate: function () {
         // Validate input element
@@ -193,8 +184,8 @@ var BrowserWidget = View.extend({
             }
         }
 
-        let invalidInputElement;
-        let invalidSelectedModel;
+        let invalidInputElement = null;
+        let invalidSelectedModel = null;
         // We want to wait until both promises to run to completion, which only happens if they're
         // accepted. So, chain an extra handler on to them, which transforms a rejection into an
         // acceptance.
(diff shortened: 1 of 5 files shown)


jbeezley:
I understand the need for asynchronous validation in the current implementation, but I don't think we should overload the conventions of the `validate` function defined by backbone.  Would it be possible to move it to a new method that allows asynchronous validation and always returns a promise?

sutartmelson:
@jbeezley I don't see what backbone defined convention `validate` should be following. This isn't `validate` as in ensure the backbone model is valid, it's `validate` to ensure the selected girder item fits some specific criteria, more or less independent of backbone.

I'm also happy to change it. Could you be more specific about moving it to a new method?

jbeezley:
Sorry, I was a little confused.  I though you were calling the `WidgetModel`'s validate method here.

I'm still not terribly happy with the idea of accepting a function that can either return a value or a promise.  What about a new option like `validateAsync` for this purpose?  It does add bloat, so we might consider deprecating the `validate` options as part of this.

Perhaps there is a possibility to do some refactoring so that asynchronous validation is not necessary.  I don't know. @brianhelba do you have any opinions here?

brianhelba:
Even if we're able to refactor and eliminate the need for async operations to validate in this case, I could easily see other validations in the future needing to do things asynchronously, so I think we should fully consider the implications now.

I agree that it's a bit ugly to have `validate` methods sometimes return a static value and sometimes return a promise. To be fair, I also think with its current behavior it's ugly and non-obvious to have `undefined` semantically represent acceptance and a string represent rejection. Promises easily encapsulate the orthogonal information of both acceptance / rejection and the reason for the resolution. Indeed promises can be used to contain static values, though the syntax is a bit more clunky (`return 'Failed'` vs `$.Deferred().reject('Failed').promise()`.

So, assuming we want to support async validation, we could:
1. Allow `validate` to return either static or promise values.
2. Expect `validate` to always return a promise value, and updating existing methods (which shouldn't be that much work). If it's possible for other plugins to define their own item_task types, we'd have to deprecate static returned values, and put in some comparability code and a warning message for comparability until the next major release.
3. Add another `validateAsync` method for promise-returning validation. I don't see the point to deprecating `validate` in this case (as we could just pick option 2), unless we really like the name of `validateAsync` much better.

After discussing this with @sutartmelson , I shied away from option 2 (which I actually think is the cleanest in principal), specifically because I didn't want to frustrate @jbeezley , et.al. with a more disruptive change than is absolutely necessary, so I initially supported option 1 as a compromise. However, if everyone agrees that having different return types is confusing, I'd fully support using option 2. I think option 3 adds too much complexity to the API surface (unless you want to treat it as a `validate` -> `validateAsync` rename), and seems like it's being a bit overly discriminatory towards async operations, even though promises allow them to be represented inside a synchronously-returnable value.

I don't have a really strong option though.

jbeezley:
The validate function was originally intended to be compatible with Backbone models' validate methods, which was the source of my original confusion.  I'm okay with changing it.  My motivation for suggesting a new method for async validation was to maintain backward compatibility.

I think the way to go is option 2.  My only request is to move the block at https://github.com/girder/girder/pull/2317/files#diff-099bf83c2490042b01e040b7ef42eb87R167 to a new private function to reduce the complexity inside the `_submitButton` function.

sutartmelson:
@brianhelba @jbeezley  I went ahead with option 2. It still allows static validation but warns of deprecation.  PTAL and let me know if there's anything I overlooked.

clients/web/src/views/widgets/BrowserWidget.js
```diff
@@ -148,27 +148,100 @@ var BrowserWidget = View.extend({
...
+     *
+     * The code path for each case is commented inline
+     */
+    _validate: function () {
+        // Validate input element
+        let inputValidation;
         if (this.input && this.input.validate) {
-            inputMessage = this.input.validate(this.$('#g-input-element').val());
+            inputValidation = this.input.validate(this.$('#g-input-element').val());
+            if (!_.isFunction(inputValidation.then)) {
```
jbeezley: This throws and error if `inputValidation` is undefined.

clients/web/src/views/widgets/BrowserWidget.js
```diff
@@ -148,27 +148,100 @@ var BrowserWidget = View.extend({
...
-            this.$('.g-selected-model').addClass('has-error');
-            this.$('.g-validation-failed-message').removeClass('hidden').text(message);
-        } else {
-            this.root = this._hierarchyView.parentModel;
-            this.$el.modal('hide');
-            this.trigger('g:saved', model, this.$('#g-input-element').val());
+        // Validate selected element
+        const selectedModel = this.selectedModel();
+        let selectedValidation = this.validate(selectedModel);
+        if (!_.isFunction(selectedValidation.then)) {
```
jbeezley: Same here.

plugins/item_tasks/web_client/views/ControlWidget.js
```diff
@@ -202,18 +213,16 @@ var ControlWidget = View.extend({
...
                 },
                 default: this.model.get('fileName')
             };
             // validation on the parent model
             validate = (model) => {
                 var type = model.get('_modelType');
-                if (!_.contains(['folder', 'collection', 'user'], type)) {
-                    return 'Invalid parent type, please choose a collection, folder, or user.';
-                }
+                var condition = _.contains(['folder', 'collection', 'user'], type);
```
brianhelba: `let` or `const`

plugins/item_tasks/web_client/views/ControlWidget.js
```diff
@@ -226,18 +235,16 @@ var ControlWidget = View.extend({
...
                 },
                 default: this.model.get('fileName')
             };
             // validation on the parent model
             validate = (model) => {
                 var type = model.get('_modelType');
-                if (type !== 'folder') {
-                    return 'Invalid parent type, please choose a folder.';
-                }
+                var condition = type === 'folder';
```
brianhelba: `let` or `const`

PR #2318, opened 2017-09-01
WIP: Add settings front page

Pierre-Assemat:
This PR will put together all the different settings on the front page : 
- Header
- SubHeading Text
- Welcome Text
- Tab name
- Banner Color 
...


Pierre-Assemat:
I notice a bug between 'Mozilla Browser' (first image) and 'Chrome Browser' (second image) : 

mozilla

chrome

Working on Mozilla I had to put an attribute `bottom 5px` to allign the `defaultButton` with the color picker but in chrome this attribute is embarrassing.. ;) 
The issue is simple to fix but I wanted to notice that because I think it's important to check on multiple browser the render of some girder views and be sure to avoid that kind of bugs..

What do you think ? @girder/developers

brianhelba:
With the exception of banner color, all of these settings affect only the rendering `frontPage.pug`. We already have a `homepage` plugin to do this, so I would strongly suggest that these be `homepage` plugin settings instead of core settings. This will significantly simplify the implementation too, since all the settings do not need to be propagated through the webroot, `App`, etc. The `homepage` plugin could just operate in a mode where it either:
* Takes these settings for some basic amount of customization
* Takes markdown for total customization

Pierre-Assemat:
Yes it's true, some of this settings should be on the Homepage plugin in particular : 
- title name
- subheading text
- welcome text

But I don't think the color banner and the tab name could be part of this plugin, I think we will need to pass them in the same way of the BrandName 

What do you think ?

ronichoudhury:
I lean towards saying that the banner color *should* be part of the homepage plugin - it is a prominent part of the homepage, after all :). I am willing to be convinced that it should go somewhere else, but I'm not sure where.

Tab name could reasonably be part of the branding settings. It is one of the names presented by the site to visitors, and as such seems to be squarely in the realm of "branding". As with the above, I'm willing to be contradicted on this.

jeffbaumes:
I'd recommend the following setting name changes:

* Tab name -> Page title (since it goes in the `<title>` element in the web page)
* Title name -> Header name (since it appears in the header bar)

A web page is not always rendered in the context of being a "tab" so I found that confusing, I at first thought it meant would appear in a tab within the Girder UI.

ronichoudhury:
I'd go a bit further and say `Header name` should be `Header text` (and, in parallel, `Subheading` should be `Subheader`).

Pierre-Assemat:
I totally agree to change the name but the 'Header' will not be part of this PR
Take a look at https://github.com/girder/girder/pull/2328

For now I separate this two PR because it was simpler to start a different way (using the Homepage plugin instead of using the webroot.mako and app.js to pass these settings) to make these changes. 
So there are two PR for the same goal but I will separate this two different ways and we have to discuss to know which settings has to be part of girder_core changes or of the homepage plugin.

Then we would be able to do an complete and cleaner PR if we want all these changes in only one PR.

ronichoudhury:
Let's go ahead and close this pull request (I'll leave the actual closing to you, @Pierre-Assemat), since the effort is proceeding on #2328. We can leave the branch here so we can borrow ideas from it as we move forward.

Issue #2313, opened 2017-08-28
Allow searching DICOM metadata


Pierre-Assemat:
For now, there was only two different search mode:
- `textSearch`
- `prefixSearch`

And we can't search inside metadata of file or item. For the DicomViewer plugin it will be interesting to be able to search inside tags/values using the search bar.

The question is : 

Does these changes are used enough to be a girder_core changes and allowed more generally new search mode or a way to search in all metadata ?
Or does these changes have to stay inside the Dicom plugin ? 

More details about the expectations : 
The search will start with any tag (key) that match to a value (or prefix) and return the item not the file.

mgrauer:
The requirements I have are

 * search for metadata that is the same across the files of an item, and return an item, i.e. search for the study instead of the series
 * search by prefix (at least have this as a mode) of a value, in addition to the entire value
 * search for any tag matching a given value or prefix

I'd like to this via an extension to the standard search UI (mainly because I think it will be the lowest effort and because users won't have to learn a new search UI), potentially with some new search modes added in, e.g. a dropdown option for "Dicom metadata" and "Dicom metadata prefix".  I'd be ok with a custom search syntax, e.g. `my_key=my_value`, or for value only queries `value:my_value`.

PR #2320, opened 2017-09-03
Store Dicom files metadata into the Dicom item

### LINKED ISSUE #2319: Separate common/unique DICOM metadata from file and store them into the item
We need to change the way DICOM metadata are stored into each file to move the common metadata under a special key like :
`item['dicomMeta'] = {
  'commonKey': 'commonValue',
  ...
}`

And the same with unique metadata, which could be a list of each data :
`item['dicomFiles'] = [
  {
    'fileId': ObjectId(...),
    'meta': {
      'uniqueKey: 'uniqueValue',
      ...
    }
  },
  ...
]`

Pierre-Assemat:
The goal of this PR is to change de way DICOM metadata are stored to store all the common metadata into the DICOM Item : 

`item['dicom']['meta'] = {
  'commonKey': 'commonValue',
  ...
}`
`item['dicom']['files'] = [
  'file1',
  'file2',
  ...
]`

In order to prepare futures PR which fix : 
- [Allow searching DICOM metadata #2313](https://github.com/girder/girder/issues/2313)
- [Allow new resource search modes to be added at runtime #2338](https://github.com/girder/girder/issues/2338)
- [Add a new view for search results #2339](https://github.com/girder/girder/issues/2339)

This PR clarify the dicom_viewer behavior by changing the client side to adapt the client to the new endpoint on server side

Fixes #2319


zachmullen:
What's the motivation here? I'm worried about the quantity of data that could be persisted on the item document -- often these images are hundreds of files, each with perhaps hundreds of tags. We could easily run up against the 16MB document limit imposed by mongo.

Pierre-Assemat:
The goal of this PR is to prepare dicom item to contain the metadata of his child files in order to be able to search inside by maybe adding a new search mode in a futur PR (https://github.com/girder/girder/issues/2313)

It's why we need to be able to access easily and efficiently to these datas. But you're right, we will need to see if theses changes will fit in mongo's restrictions

plugins/dicom_viewer/server/__init__.py
```diff
@@ -46,6 +46,65 @@ def getDicom(self, item, filters, force):
...
+        dicomFileMeta = parse_file(file[0])
+        item['dicomMeta'] = dicomFileMeta
+        for additionalFile in files[1:]:
+            additionalFileMetadata = parse_file(additionalFile)
+            item['dicomMeta'] = removeUniqueMetadata(dicomFileMeta, additionalFileMetadata)
+        return ModelImporter.model('item').save(item)
+
+
+def removeUniqueMetadata(dicomMeta, additionalMeta):
+    dicomValues = []
```
brianhelba: Python `dict`s have random access, so I think your implementation here can be significantly simplified. Let's discuss this directly.

Pierre-Assemat:
I don't really know which kind of icon should be there to give some sens to the dicom endpoint :

screenshot from 2017-09-28 09-55-16

I thought about this four, do you have any thought @girder/developers ?

I like the 'recycle' and the 'stethoscope' ones.

plugins/dicom_viewer/plugin_tests/data/000000.dcm.sha512
```diff
@@ -0,0 +1 @@
+9a3ee0830fd1b4da2babe5cdb30d4fa0021898c8ba3167b08bffb719e89894087f3cf27d2307cec440c71f13589e39b8fa2c5d094ce90ae5240c522870a52068
```
brianhelba: Can you add a newline to the end of each of these 4 files? The old `hashsum_download` logic was not doing so.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,59 +1,117 @@
...
+
+
+def _extractFileData(file, dicom):
+    """Extract the usefull data to be stored in the `item['dicom']['files']`.
+    In this way it become simpler to sort them and store them.
+    """
+    return {
+        '_id': file.get('_id'),
+        'name': file.get('name'),
+        'SeriesNumber': dicom.get('SeriesNumber'),
```
brianhelba: Can we put the 3 DICOM fields under sub-a field called `dicom`? So,
```
{
    '_id': ...,
    'name': ...,
    'dicom: {
        'SeriesNumber': ...,
        ...
    }
}
```

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,59 +1,117 @@
...
+
+
+def _sort_key(f):
+    """These properties are used to sort the files into the item."""
     return (
-        dicom.get('SeriesNumber'),
-        dicom.get('InstanceNumber'),
-        dicom.get('SliceLocation'),
-        f['name'],
+        f.get('name'),
```
brianhelba: I think `name` should be a final fallback, if the first 3 fields fail to disambiguate slices.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,59 +1,117 @@
...
+            # Save the item
+            Item().save(item)
+
+
+def _extractFileData(file, dicom):
+    """Extract the usefull data to be stored in the `item['dicom']['files']`.
+    In this way it become simpler to sort them and store them.
+    """
+    return {
+        '_id': file.get('_id'),
```
brianhelba: `_id` and `name` are guaranteed to be present, so there's no need to use `.get`

plugins/dicom_viewer/server/__init__.py
```diff
@@ -98,21 +156,36 @@ def parse_file(f):
...
+        return
+    item = Item().load(file['itemId'], force=True)
+    if 'dicom' in item:
+        item['dicom']['meta'] = removeUniqueMetadata(item['dicom']['meta'], fileMetadata)
+    else:
+        # In this case the uploaded file is the first of the item
+        item['dicom'] = {}
+        item['dicom']['meta'] = fileMetadata
+        item['dicom']['files'] = []
+    item['dicom']['files'].append(_extractFileData(file, fileMetadata))
```
brianhelba: This new file needs to be added in a sorted position within the existing `files` list. A test should verify that it's doing so (e.g. by uploading files one at a time, out of order).

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,59 +1,117 @@
...
+    return {
+        '_id': file.get('_id'),
+        'name': file.get('name'),
+        'SeriesNumber': dicom.get('SeriesNumber'),
+        'InstanceNumber': dicom.get('InstanceNumber'),
+        'SliceLocation': dicom.get('SliceLocation')
+    }
+
+
+def _sort_key(f):
```
brianhelba: Can we call this this `_getDicomFileSortKey`?

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,59 +1,117 @@
...
-    dicom = f.get('dicom', {})
+    def makeDicomItem(self, item):
+        """Try to convert an existing item into a "DICOM item", which contains a
+        "dicomMeta" field with DICOM metadata that is common to all DICOM files.
+        """
+        metadataReference = None
+        dicomFiles = []
+
+        for file in Item().childFiles(item):
+            dicomMeta = parse_file(file)
```
brianhelba: Please rename `parse_file` to `_parseFile`.

plugins/dicom_viewer/server/event_helper.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class EventHelper(object):
+    """
+    Helper class to wait for plugin's data.process event handler to complete.
+    Usage:
+        with EventHelper('event.name') as helper:
+            self.model('upload').uploadFile(...)
+            handled = helper.wait()
+    """
```
msmolens: Since this file is copied directly from my repository, can you:
1. Split out adding this class into its own commit
1. Include the following in the commit message:
```
From: Max Smolens <max.smolens@kitware.com>
```

Pierre-Assemat: Of course, maybe this class should be integrate into Girder Core, it's really a great tools to manage event.
I will discuss about that with @brianhelba

brianhelba: Honestly, this functionality should be upstreamed into core Girder. @msmolens if you have time to do so soon, we can just merge the new feature into this branch. Otherwise, we can move forward with a temporary implementation in this plugin  (crediting the source), and move it upstream later.

brianhelba: @msmolens As much as I agree with making this functionality into its own commit in principle, it might be too difficult to do so, since this is a collaborative branch, and rewriting history would be disruptive. @Pierre-Assemat and I will see what we can do, however.

plugins/dicom_viewer/server/__init__.py
Pierre-Assemat: If they're uploaded in the reverse order, and the test check if they are well sorted, do you think it's sufficient?

brianhelba: The sort here should use `_getDicomFileSortKey`, which will produce the same output every time. It shouldn't matter what order a user uploads single files, the resulting `files` list should always be the same.

Pierre-Assemat: Yes, this and the test are updated in the two last commit!

plugins/dicom_viewer/plugin_tests/dicom_viewer_test.py
```diff
@@ -22,69 +28,151 @@ class DicomViewerTest(base.TestCase):
...
+        # Check the endpoint 'parseDicom' for an non admin user
+        dicomItem = Item().load(item['_id'], force=True)
+        dicomItem = self._purgeDicomItem(dicomItem)
+        path = '/item/%s/parseDicom' % dicomItem.get('_id')
+        resp = self.request(path=path, method='POST', user=user)
         self.assertStatus(resp, 403)
+
+    def _uploadDicomFiles(self, item, user):
+        # Upload the files in the reverse order to check if they're well sorted
+        for i in range(0,4):
```
brianhelba: To test out-of-order uploads, let's make this `reversed(range(0, 4))`.

zachmullen: `range(4, 0, -1)`

plugins/dicom_viewer/plugin_tests/dicom_viewer_test.py
```diff
@@ -22,69 +28,151 @@ class DicomViewerTest(base.TestCase):
...
+            file = os.path.join(self.dataDir, '00000%i.dcm' % i)
+            with open(file, 'rb') as fp:
+                dcmData = fp.read()
+            self.assertTrue(dcmData)
+
+            with _EventHelper('dicom.handler.success') as helper:
+                dcmFile = Upload().uploadFromFile(
+                    obj=six.BytesIO(dcmData),
+                    size=os.path.getsize(file),
+                    name='00000%i.dcm' % (3 - i),
```
brianhelba: Use the real name here, and use the loop control for the ordering.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,67 +1,129 @@
...
-from girder.models.file import File
+from girder.constants import AccessType
 from girder.models.item import Item
+from girder.models.file import File
 
 
 class DicomItem(Resource):
 
-    @access.public(scope=TokenScope.DATA_READ)
+    @access.user
```
brianhelba: We should re-add the `scope=TokenScope.DATA_READ` argument to this decorator.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,67 +1,129 @@
...
+        """
+        metadataReference = None
+        dicomFiles = []
+
+        for file in Item().childFiles(item):
+            dicomMeta = _parseFile(file)
+            if dicomMeta is None:
+                continue
+            dicomFiles.append(_extractFileData(file, dicomMeta))
+            dicomFiles.sort(key=_getDicomFileSortKey)
```
brianhelba: Don't need to sort this after every insertion, just once at the end.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,67 +1,129 @@
...
+            if metadataReference is None:
```
brianhelba: You *could* write this as a ternary statement:
```
metadataReference = dicomMeta if metadataReference is None else _removeUniqueMetadata(metadataReference, dicomMeta)
```
or formatted as:
```
metadataReference = dicomMeta \
    if metadataReference is None else \
    _removeUniqueMetadata(metadataReference, dicomMeta)
```

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,67 +1,129 @@
...
+                metadataReference = dicomMeta
+            else:
+                metadataReference = _removeUniqueMetadata(metadataReference, dicomMeta)
+
+        if len(dicomFiles) is not 0:
+            # Sort the dicom files
+            dicomFiles = sorted(dicomFiles, key=_getDicomFileSortKey)
```
brianhelba: This will create a new list, rather than sort the existing one in place. Use `dicomFiles.sort` instead.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,67 +1,129 @@
...
+            if dicomMeta is None:
+                continue
+            dicomFiles.append(_extractFileData(file, dicomMeta))
+            dicomFiles.sort(key=_getDicomFileSortKey)
+            if metadataReference is None:
+                metadataReference = dicomMeta
+            else:
+                metadataReference = _removeUniqueMetadata(metadataReference, dicomMeta)
+
+        if len(dicomFiles) is not 0:
```
brianhelba: You can just say `if dicomFiles:`. Empty lists are falsey.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,67 +1,129 @@
...
+            # Sort the dicom files
+            dicomFiles = sorted(dicomFiles, key=_getDicomFileSortKey)
+            # Store in the item
+            item['dicom'] = {}
```
brianhelba: Can do this in one statement:
```
item['dicom'] = {
   'meta': ...
}
```

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,67 +1,129 @@
...
-        f['name'],
+        f.get('name'),
+        meta.get('SeriesNumber'),
+        meta.get('InstanceNumber'),
+        meta.get('SliceLocation')
+    )
+
+
+def _removeUniqueMetadata(dicomMeta, additionalMeta):
+    """Return only the common data between the two inputs.
```
brianhelba: This docstring should have a `"""` only on the first line. Multi-line docstrings should always have their `"""` on separate lines. I think it's ok to do single-line docstrings on the same line as the `"""` though.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,67 +1,129 @@
...
+            # Store in the item
+            item['dicom'] = {}
+            item['dicom']['meta'] = metadataReference
+            item['dicom']['files'] = dicomFiles
+            # Save the item
+            Item().save(item)
+
+
+def _extractFileData(file, dicom):
+    """Extract the usefull data to be stored in the `item['dicom']['files']`.
```
brianhelba: Fix this docstring too.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,67 +1,129 @@
...
-                dicom = f['dicom']
-                dicom = dict((k, dicom[k]) for k in filters if k in dicom)
-                f['dicom'] = dicom
-        return files
-
-
-def sort_key(f):
-    dicom = f.get('dicom', {})
+    def makeDicomItem(self, item):
+        """Try to convert an existing item into a "DICOM item", which contains a
```
brianhelba: Docstring.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -1,67 +1,129 @@
...
+        meta.get('SliceLocation')
+    )
+
+
+def _removeUniqueMetadata(dicomMeta, additionalMeta):
+    """Return only the common data between the two inputs.
+    Only work if all the data element are hashable,
+    this means not have any dict or list as properties
+    """
+    return dict(
```
brianhelba: We should eventually re-evaluate this, and see if we can do it recursively, since it only works for single-level lists here. For right now, we can leave it alone though.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -89,30 +151,46 @@ def parse_file(f):
...
 
 
-def process_file(f):
-    f['dicom'] = parse_file(f)
-    return File().save(f)
-
-
-def handler(event):
-    process_file(event.info['file'])
+def _handler(event):
```
brianhelba: Rename this to `_uploadHandler`, to be a little more descriptive.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -89,30 +151,46 @@ def parse_file(f):
...
+    file = event.info['file']
+    fileMetadata = _parseFile(file)
+    if fileMetadata is None:
+        return
+    item = Item().load(file['itemId'], force=True)
+    if 'dicom' in item:
+        item['dicom']['meta'] = _removeUniqueMetadata(item['dicom']['meta'], fileMetadata)
+    else:
+        # In this case the uploaded file is the first of the item
+        item['dicom'] = {}
```
brianhelba: Combine these 3 lines into one statement.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -89,30 +151,46 @@ def parse_file(f):
...
+        item['dicom']['meta'] = fileMetadata
+        item['dicom']['files'] = []
+    item['dicom']['files'].append(_extractFileData(file, fileMetadata))
+    item['dicom']['files'].sort(key=_getDicomFileSortKey)
+    Item().save(item)
+    events.trigger('dicom.handler.success')
```
brianhelba: How about calling this event `dicom_viewer.upload.success`?

plugins/dicom_viewer/plugin_tests/dicom_viewer_test.py
```diff
@@ -22,69 +28,151 @@ class DicomViewerTest(base.TestCase):
...
+        dicomItem = self._purgeDicomItem(dicomItem)
+        path = '/item/%s/parseDicom' % dicomItem.get('_id')
+        resp = self.request(path=path, method='POST', user=user)
         self.assertStatus(resp, 403)
+
+    def _uploadDicomFiles(self, item, user):
+        # Upload the files in the reverse order to check if they're well sorted
+        for i in range(0,4):
+            file = os.path.join(self.dataDir, '00000%i.dcm' % i)
+            with open(file, 'rb') as fp:
```
brianhelba: There's no need to `read` the whole thing into `dcmData` and then wrap it in a `six.BytesIO`. The actual `fp` object, is a file-like object that's accepted by `uploadFromFile`. You will just need to indent the whole `uploadFromFile` to keep it inside the `with open(file, 'rb') as fp:` context manager.

brianhelba: FYI, you can [combine `with` statements](https://docs.python.org/2/reference/compound_stmts.html) like:
```
with open(file, 'rb') as fp, _EventHelper('dicom.handler.success') as helper:
```

plugins/dicom_viewer/plugin_tests/dicom_viewer_test.py
```diff
@@ -22,69 +28,151 @@ class DicomViewerTest(base.TestCase):
...
+                dcmData = fp.read()
+            self.assertTrue(dcmData)
+
+            with _EventHelper('dicom.handler.success') as helper:
+                dcmFile = Upload().uploadFromFile(
+                    obj=six.BytesIO(dcmData),
```
brianhelba: If you want to test non-DICOM files, you can create a "fake" file:
```
nonDicomContent = 'hello world\n'
Upload().uploadFromFile(
    obj=six.BytesIO(nonDicomContent),
    size=len(nonDicomContent)
...
```

plugins/dicom_viewer/server/__init__.py
Pierre-Assemat: What about like that ? 
```
metadataReference = (
    dicomMeta
    if metadataReference is None else
    _removeUniqueMetadata(metadataReference, dicomMeta)
)
```

brianhelba:
@Pierre-Assemat I've split my changes out into https://github.com/girder/girder/pull/2443 . Let's discuss rebasing this PR to reduce the number of commits, and to have a much smaller number of frontend changes.

PR #2325, opened 2017-09-04
Fixes #2322. Do not HTML-escape values inlined into javascript

### LINKED ISSUE #2322: Girder branding renders escaped text for special characters
image

zachmullen:


brianhelba:
What is the motivation behind this change?

Also, I don't think this will work, as dumping to JSON is not sufficient to protect against an injection attack with `<\script>`. In this context, I think the tokens `'`, `"`, and  `<\script>` need to be escaped. Does [the string from here](https://github.com/girder/girder/pull/2279#pullrequestreview-56490060) still work as an attack?

brianhelba:
Oh, I see the issue. My comments still apply.

In light of how tricky this is, I think we should have our tests for these variables try injection attacks.

zachmullen:
Are we really worried about injection attacks from a site admin?

brianhelba:
Even if we totally trust the site admin person, the threat comes from "person with the site admin's credentials" or "person with the ability to set a site setting, via some other vulnerability (e.g. unsecured MongoDB)".

zachmullen:
Ok, I will propose an alternative solution.

zachmullen:
@brianhelba PTAnotherL.

screen shot 2017-09-04 at 1 23 48 pm

brianhelba:
@zachmullen This revision no longer solves the original issue:
escape

[Mako's `h` filter](http://docs.makotemplates.org/en/latest/filtering.html) uses [`markupsafe.escape`, which escapes the 5 characters `'`, `"`, `<`, `>`, and `&`](https://github.com/pallets/markupsafe/blob/master/markupsafe/_native.py#L22). Unfortunately, different escaped characters [for `'` and `"` are used by Underscore](http://underscorejs.org/docs/underscore.html#section-160), so a two way mapping does not work properly.

I believe that I've got a solution, which will more strictly follow [the OWASP recommendations for this case:
https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet#RULE_.233_-_JavaScript_Escape_Before_Inserting_Untrusted_Data_into_JavaScript_Data_Values
I'll post it as soon as possible.

brianhelba:
@zachmullen @Pierre-Assemat PTAL at the updated solution in my last commit.

Ping @danlamanna , as an extra pair of eyes, since this is a security issue.

jeffbaumes:
Thanks for looking into this. It's odd that the task "safely rendering a user-entered string in HTML" is such a nuanced problem. Seems like it should be a solved problem. But I understand that there are issues around balancing security with fidelity with the user's intent.

zachmullen:
I approve this, though I cannot actually approve since I opened the PR.

brianhelba:
@danlamanna Do you have any thoughts as to where else it would possibly be used? My gut is that if we have to escape some Python string into Javascript in places beyond where it is now, we're probably doing something wrong, but maybe I'm not being imaginative enough.

PR #2328, opened 2017-09-05
Add custom settings in Homepage plugin for the Front Page

Pierre-Assemat:
These settings are :
- header
- subheading --> subheader
- welcome text

This PR has the same goal of https://github.com/girder/girder/pull/2318 but without using the webroot to pass the settings, in this changes only the homepage plugin is touch

### DIFF
--- a/plugins/homepage/server/__init__.py
+++ b/plugins/homepage/server/__init__.py
@@ -27,6 +27,7 @@ MARKDOWN = 'homepage.markdown'
 HEADER = 'homepage.header'
 SUBHEADER = 'homepage.subheader'
 WELCOME_TEXT = 'homepage.welcome_text'
+LOGO = 'homepage.logo'
 
 NAME = 'Homepage Assets'
 
@@ -59,6 +60,7 @@ class Homepage(Resource):
             HEADER: self.model('setting').get(HEADER),
             SUBHEADER: self.model('setting').get(SUBHEADER),
             WELCOME_TEXT: self.model('setting').get(WELCOME_TEXT),
+            LOGO: self.model('setting').get(LOGO),
             'folderId': folder['_id']
         }
 
@@ -79,7 +81,11 @@ def validateHomepageSubHeadingText(event):
 
 
 @setting_utilities.validator(WELCOME_TEXT)
-def validateHomepagewelcomeText(event):
+def validateHomepageWelcomeText(event):
+    pass
+
+@setting_utilities.validator(LOGO)
+def validateHomepageLogo(event):
     pass
 
 
--- a/plugins/homepage/server/__init__.py
+++ b/plugins/homepage/server/__init__.py
@@ -25,7 +25,7 @@ from girder.utility.model_importer import ModelImporter
 
 MARKDOWN = 'homepage.markdown'
 HEADER = 'homepage.header'
-SUBHEADING_TEXT = 'homepage.subheading_text'
+SUBHEADER = 'homepage.subheader'
 WELCOME_TEXT = 'homepage.welcome_text'
 
 NAME = 'Homepage Assets'
@@ -36,7 +36,7 @@ class Homepage(Resource):
         super(Homepage, self).__init__()
         self.resourceName = 'homepage'
         self.route('GET', ('markdown',), self.getMarkdown)
-        self.route('GET', ('settings',), self.getSettingFrontPage)
+        self.route('GET', ('settings',), self.getSettings)
 
     @access.public
     @describeRoute(
@@ -53,11 +53,11 @@ class Homepage(Resource):
     @describeRoute(
         Description('Public url for getting the homepage settings.')
     )
-    def getSettingFrontPage(self, params):
+    def getSettings(self, params):
         folder = getOrCreateAssetsFolder()
         return {
             HEADER: self.model('setting').get(HEADER),
-            SUBHEADING_TEXT: self.model('setting').get(SUBHEADING_TEXT),
+            SUBHEADER: self.model('setting').get(SUBHEADER),
             WELCOME_TEXT: self.model('setting').get(WELCOME_TEXT),
             'folderId': folder['_id']
         }
@@ -73,7 +73,7 @@ def validateHomepageTitle(event):
     pass
 
 
-@setting_utilities.validator(SUBHEADING_TEXT)
+@setting_utilities.validator(SUBHEADER)
 def validateHomepageSubHeadingText(event):
     pass
 
--- a/plugins/homepage/server/__init__.py
+++ b/plugins/homepage/server/__init__.py
@@ -20,13 +20,9 @@
 from girder.api import access
 from girder.api.describe import Description, describeRoute
 from girder.api.rest import Resource
-from girder.constants import SettingDefault
-from girder.utility import setting_utilities
 from girder.utility.model_importer import ModelImporter
 
 from . import constants
-
-import six
 
 
 class Homepage(Resource):
@@ -49,42 +45,6 @@ class Homepage(Resource):
             constants.PluginSettings.LOGO: self.model('setting').get(constants.PluginSettings.LOGO),
             'folderId': folder['_id']
         }
-
-@setting_utilities.validator({
-    constants.PluginSettings.MARKDOWN,
-    constants.PluginSettings.HEADER,
-    constants.PluginSettings.SUBHEADER,
-    constants.PluginSettings.WELCOME_TEXT
-})
-def validateHomepageMarkdown(doc):
-    if not isinstance(doc['value'], six.string_types):
-        raise ValidationException('The setting is not a string', 'value')
-
-
-@setting_utilities.validator(constants.PluginSettings.LOGO)
-def validateHomepageLogo(doc):
-    if not isinstance(doc['value'], six.string_types):
-        pass
-
-
-@setting_utilities.default(constants.PluginSettings.MARKDOWN)
-def defaultHomepageMarkdown():
-    return ''
-
-
-@setting_utilities.default(constants.PluginSettings.HEADER)
-def defaultHomepageHeader():
-    return 'Girder'
-
-
-@setting_utilities.default(constants.PluginSettings.SUBHEADER)
-def defaultHomepageSubheader():
-    return 'Data management platform'
-
-
-@setting_utilities.default(constants.PluginSettings.WELCOME_TEXT)
-def defaultHomepageWelcomeText():
-    return 'Welcome to Girder!'
 
 
 def getOrCreateAssetsFolder():
--- a/plugins/homepage/server/__init__.py
+++ b/plugins/homepage/server/__init__.py
@@ -37,26 +37,15 @@ class Homepage(Resource):
         super(Homepage, self).__init__()
         self.resourceName = 'homepage'
         self.route('GET', ('markdown',), self.getMarkdown)
-        self.route('GET', ('settings',), self.getSettings)
 
     @access.public
     @describeRoute(
-        Description('Public url for getting the homepage markdown.')
+        Description('Public url for getting the homepage markdown and settings.')
     )
     def getMarkdown(self, params):
         folder = getOrCreateAssetsFolder()
         return {
             MARKDOWN: self.model('setting').get(MARKDOWN),
-            'folderId': folder['_id']
-        }
-
-    @access.public
-    @describeRoute(
-        Description('Public url for getting the homepage settings.')
-    )
-    def getSettings(self, params):
-        folder = getOrCreateAssetsFolder()
-        return {
             HEADER: self.model('setting').get(HEADER),
             SUBHEADER: self.model('setting').get(SUBHEADER),
             WELCOME_TEXT: self.model('setting').get(WELCOME_TEXT),
--- a/plugins/homepage/server/__init__.py
+++ b/plugins/homepage/server/__init__.py
@@ -17,45 +17,10 @@
 #  limitations under the License.
 ###############################################################################
 
-from girder.api import access
-from girder.api.describe import Description, describeRoute
-from girder.api.rest import Resource
-from girder.utility.model_importer import ModelImporter
-
+# Import constants for the side effect of registering settings
 from . import constants
-
-
-class Homepage(Resource):
-    def __init__(self):
-        super(Homepage, self).__init__()
-        self.resourceName = 'homepage'
-        self.route('GET', ('markdown',), self.getMarkdown)
-
-    @access.public
-    @describeRoute(
-        Description('Public url for getting the homepage properties including markdown content.')
-    )
-    def getMarkdown(self, params):
-        folder = getOrCreateAssetsFolder()
-        return {
-            constants.PluginSettings.MARKDOWN: self.model('setting').get(constants.PluginSettings.MARKDOWN),
-            constants.PluginSettings.HEADER: self.model('setting').get(constants.PluginSettings.HEADER),
-            constants.PluginSettings.SUBHEADER: self.model('setting').get(constants.PluginSettings.SUBHEADER),
-            constants.PluginSettings.WELCOME_TEXT: self.model('setting').get(constants.PluginSettings.WELCOME_TEXT),
-            constants.PluginSettings.LOGO: self.model('setting').get(constants.PluginSettings.LOGO),
-            'folderId': folder['_id']
-        }
-
-
-def getOrCreateAssetsFolder():
-    collection = ModelImporter.model('collection').createCollection(
-        constants.COLLECTION_NAME, public=False, reuseExisting=True)
-    folder = ModelImporter.model('folder').createFolder(
-        collection, constants.COLLECTION_NAME, parentType='collection', public=True, reuseExisting=True)
-    return folder
+from . import rest
 
 
 def load(info):
-    info['apiRoot'].homepage = Homepage()
-
-
+    info['apiRoot'].homepage = rest.Homepage()
--- a/plugins/homepage/server/__init__.py
+++ b/plugins/homepage/server/__init__.py
@@ -20,16 +20,13 @@
 from girder.api import access
 from girder.api.describe import Description, describeRoute
 from girder.api.rest import Resource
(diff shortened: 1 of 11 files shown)


plugins/homepage/web_client/views/FrontPageView.js
```diff
@@ -10,7 +10,27 @@ wrap(FrontPageView, 'render', function (render) {
...
+                    this.$('.g-frontpage-subtitle').text(res['homepage.subheading_text']);
+                }
+                if (res['homepage.welcome_text'] !== '') {
+                    this.$('.g-frontpage-welcome-text').text(res['homepage.welcome_text']);
+                }
+            }, this));
+        } else {
+            this.$el.html(renderMarkdown(resp['homepage.markdown']));
+        }
     }, this));
```
ronichoudhury: This is good for now, but we probably need a way to explicitly select whether the user wishes to completely replace the content via the markdown box, or change a few items via the other settings. Forcing the user to delete information (from the markdown box) just to activate a different config path is a UI failing in my opinion.

Pierre-Assemat: Doing that, for now, force the markdown to be priority on the other settings, but I'm agree it shouldn't be the correct method to do that

Pierre-Assemat:
I don't think we can put the banner color setting into this PR, like you said, this don't touch the front page, but I think we can use the configuration of https://github.com/girder/girder/pull/2318 if we agree on the fact that being able to change the color banner from the settings page of girder_core make sens.

For the logo I think it's possible and not very tough to implement that in the homepage plugin, I can try to start a first approach and let you see.

Enable to enter a markdown into the welcome text setting is a good idea, and I don't think it will be tough either.

And then I'm not sure to understand the 'special directive to represent that bullet point content' but do you mean a little preview box of the front page with the different bloc inside (header, logo, subheader, welcome text, ...) where the user can move each blocs to render a even more custom homepage ?

plugins/homepage/web_client/views/FrontPageView.js
ronichoudhury: Ok, we can file an issue so we remember to address this later.

ronichoudhury:
On Wed, Sep 6, 2017 at 11:39 AM Pierre Assemat wrote:

I don't think we can put the banner color setting into this PR, like you
> said, this don't touch the front page, but I think we can use the
> configuration of #2318 <https://github.com/girder/girder/pull/2318> if we
> agree on the fact that being able to change the color banner from the
> settings page of girder_core make sens.
>
I like this approach. As I said in a separate thread, I think the banner
color can simply be considered as one aspect of instance branding.

For the logo I think it's possible and not very tough to implement that in
> the homepage plugin, I can try to start a first approach and let you see.
>
Fantastic - go ahead and work on this and I’ll watch for the commit to come
across :)

Enable to enter a markdown into the welcome text setting is a good idea,
> and I don't think it will be tough either.
>
Same as above :-D

And then I'm not sure to understand the 'special directive to represent
> that bullet point content' but do you mean a little preview box of the
> front page with the different bloc inside (header, logo, subheader, welcome
> text, ...) where the user can move each blocs to render a even more custom
> homepage ?
>
Sorry, I didn’t express this very well.

I just meant, maybe we could have a special string in the markdown content
box that represents the block of bullet points that can’t be recreated
manually. So you could do something like:

Welcome to Foobar!

This website blah blah blah.

{{bulletpoints}}

Please contact soandso@foobaz.com for help!

This way we could place extra content *after* the “bullet points” if we
wanted to.

But this is just a pie-in-the-sky idea at this point. I wouldn’t worry
about it right now. If enough people think it’s a good idea (I’m not really
convinced of it myself at this point) we can file an issue to handle it in
the future.
​

Pierre-Assemat:
There is still some weird bug when I try to test it, some time it's working very well and some other time not at all... There is some phantom screenshot : 

phantom-2017-09-06t23 09 55 328z
phantom-2017-09-06t23 13 53 707z

The view should be like this : 

screenshot from 2017-09-06 19-19-00

I don't know yet if it's because of the test itself (miss a wait for...) or if there is an real issue in the implementation of the change settings

plugins/homepage/server/__init__.py
```diff
@@ -32,6 +37,7 @@ def __init__(self):
         super(Homepage, self).__init__()
         self.resourceName = 'homepage'
         self.route('GET', ('markdown',), self.getMarkdown)
+        self.route('GET', ('settings',), self.getSettings)
```
brianhelba: There should be no need to define another new route. The new settings can be returned as additional fields in the response of the existing route.

Pierre-Assemat:
I implement the logo setting using `input[file]` the configuration page is like this for now : 

screenshot from 2017-09-08 14-24-31

I have few questions : @ronichoudhury  @brianhelba 
- I wonder if using this type of input could bring some security issue ?
- Where do you think it will be better to display the preview logo ?
- Do we need to allow the admin user to reuse the girder logo by default after changed it ?

And because of security reason, it's impossible to set the file value of an `input[file]`, so even if a logo is already save in the DB, the input value will always be 'no file selected' value after reload the page
(it's the case for the screenshot above, the Girder logo is already save into the DB)

Pierre-Assemat:
screenshot from 2017-09-08 16-12-13

Pierre-Assemat:
I think this PR is ready to be reviewed @girder/developers

plugins/homepage/server/__init__.py
```diff
@@ -35,21 +40,45 @@ def __init__(self):
 
     @access.public
     @describeRoute(
-        Description('Public url for getting the homepage markdown.')
+        Description('Public url for getting the homepage markdown and settings.')
```
jcfr: What about:

```
Public url for getting the homepage properties including markdown content.
```

plugins/homepage/server/__init__.py
```diff
@@ -23,7 +23,12 @@
 from girder.utility import setting_utilities
 from girder.utility.model_importer import ModelImporter
 
-KEY = 'homepage.markdown'
+MARKDOWN = 'homepage.markdown'
+HEADER = 'homepage.header'
+SUBHEADER = 'homepage.subheader'
+WELCOME_TEXT = 'homepage.welcome_text'
```
jcfr: Should this be `welcome_template` ? Look like the final text is rendered differently based on the username, etc ...

ronichoudhury: The text at the bottom (with bullet points) is independent from the "welcome text" (which appears directly above the bullet point text). Some documentation could probably help with explaining this.

ronichoudhury:
@jcfr The welcome_text is arbitrary markdown that will be placed at the top of the gray box (replacing the **Welcome to <brandname>!**) that serves as the default.

Using the "markdown" box will replace the entire front page content. We do need a way to explicitly say whether the user wants that markdown box, or the individual settings, but we deferred it for now.

ronichoudhury:
I believe it's important to be able to "clear" the logo dialog (and restore the default Girder logo).

It's also important to be consistent when displaying it. If it disappears from the input element on refresh, I'd say we need to have a separate `img` element next to the dialog that displays the current, if any, data url associated with the logo property. Does that make sense or did I miss anything fundamental?

jcfr:
> The welcome_text is arbitrary markdown  ... (replacing the Welcome to !) that serves as the default.

In that case, I suggest to show the default markdown in the box and give opportunity for the admin to tweak it.

Field like username, ... could be supported by using syntax like `{girder.firstname} {girder.lastname}` (or similar)


> logo

Is the logo specific to he home page ? Or for the whole instance ?

Pierre-Assemat:
the logo is specific to the homepage only

jcfr:
> the logo is specific to the homepage only

I see. Is there already a location to configure the instance "favicon" (it is the icon apprearing in the url bar and tab of your browser) ?  May the logo could go there instead of the homepage plugin ? 

Or should we have a "branding" plugin ?

Pierre-Assemat:
For now all the Branding settings are in girder_core, like the color banner setting : https://github.com/girder/girder/pull/2334
I don't know if a plugin for this is preferable rather than a girder_core feature. 
I think this PR is limited to the Homepage and if we consider this kind of feature are very usefull then we could open a new PR at this moment.

brianhelba:
I can't believe we've been missing out on using
beam-clipart-ibeam12
as Girder's logo for all this time!

jcfr:
> I think this PR is limited to the Homepage and if we consider this kind of feature are very usefull then we could open a new PR at this moment.

Makes sense.

That said, I still think the welcome text and markdown should be consolidated.

ronichoudhury:
@jcfr I'm not sure what you mean by "consolidated". Could you elaborate a bit?

jcfr:
> I'm not sure what you mean by "consolidated". Could you elaborate a bit?

Earlier you mentionned

> > The welcome_text is arbitrary markdown  ... (replacing the Welcome to !) that serves as the default.

In that case, I suggest to show the default markdown in the box and give opportunity for the admin to tweak it.

Field like username, ... could be supported by using syntax like `{girder.firstname} {girder.lastname}` (or similar)

This basically means the only field of the API call would be: 
* header
* subheader
* markdown
* default_markdown  (to easily revert back)

When the site admin wants to custmize the welcome text, he would then have the current welcome text as a starting point.

ronichoudhury:
Ah I think I understand.

Since this would use some templating logic that doesn't exist yet, I'd prefer if we could QA/merge the current PR as is, and then take a look at performing the upgrade you just described. I like that approach but it will take a bit of design and a bit of bikeshedding to get it right.

Is that ok?

brianhelba:
I have some thoughts about how far these additional settings should go:

> I see. Is there already a location to configure the instance "favicon" (it is the icon apprearing in the url bar and tab of your browser) ? May the logo could go there instead of the homepage plugin ?

While the homepage logo and favicon may often have the same appearance, they are necessarily going to be stored in different file formats. Unless we implement a favicon generator via image processing, there'd have to be at least 2 settings for both.

> Or should we have a "branding" plugin ? (instead or in addition of this one)

If users want to do truly extensive appearance customization of Girder, the best way right now is via a custom plugin. A single `brandName` string does provide a huge amount of customization power with respect to its implementation and usage complexity, and it's something that a user may immediately want to change within the first few minutes of installing Girder, so I support its inclusion. Alternatively, given that Girder's current system for adding settings does not scale well, something like favicon settings in core would start (in my opinion) to go down the scary path of bloat; additionally, if a user is fabricating their own logo and favicon, then we can assume that they're probably more invested in using Girder, so it's less unreasonable to expect them to enable it via a custom plugin.

I'm not sure what a `branding` plugin would contain, but for it to be truly general and useful, I think it would need to have the ability to apply custom styles to all views, which a is terrifyingly difficult system to do well (judging from how other projects have handled it). I'd personally much rather invest the time into further modularization of Girder widgets, making it easier to just build up custom UIs from scratch, instead of allowing components to be heavily restyled in-place.

---

As a final takeaway, I think we should remember that Girder is not Wordpress. That means styling will probably always be more difficult, but it also means that deep customization should always be much easier.

jcfr:
Hi Brian,

Thanks for taking the time to reply. Agreed with you, I thought it was
worth discussing it. May be your comment could be part of the documentation
on "what is Girder , what is not Girder"

In the mean time, have a good weekend.
Jc

jcfr:
> Is that ok?

Sounds good.

We can definitively keep the improvement for later.



On Sep 8, 2017 5:06 PM, "Roni Choudhury" <notifications@github.com> wrote:

Ah I think I understand.

Since this would use some templating logic that doesn't exist yet, I'd
prefer if we could QA/merge the current PR as is, and then take a look at
performing the upgrade you just described. I like that approach but it will
take a bit of design and a bit of bikeshedding to get it right.

Is that ok?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/girder/girder/pull/2328#issuecomment-328224882>, or mute
the thread

ronichoudhury:
Thanks @jcfr.

ronichoudhury:
Thanks @manthey for your help in diagnosing the transient test failure!

Pierre-Assemat:
@brianhelba can you do a quick review when you will have time please ? I think this PR is ready to be merged

plugins/homepage/server/__init__.py
```diff
@@ -35,21 +40,45 @@ def __init__(self):
 
     @access.public
     @describeRoute(
-        Description('Public url for getting the homepage markdown.')
+        Description('Public url for getting the homepage properties including markdown content.')
```
brianhelba: Since this provides much more than just markdown now, I'd be ok if we renamed this endpoint. It would be an incompatible change, but I'm not sure if this plugin needs to have a stable external API. What do you think @girder/developers ?

plugins/homepage/web_client/views/FrontPageView.js
```diff
@@ -10,7 +10,23 @@ wrap(FrontPageView, 'render', function (render) {
         method: 'GET',
         url: 'homepage/markdown'
     }).done(_.bind(function (resp) {
```
brianhelba: Change this whole callback to be an arrow (`=>`) function.

plugins/homepage/web_client/views/FrontPageView.js
```diff
@@ -10,7 +10,23 @@ wrap(FrontPageView, 'render', function (render) {
...
-        this.$el.html(renderMarkdown(resp['homepage.markdown']));
+        if (resp['homepage.markdown'] === '') {
+            _.bind(render, this)();
```
brianhelba: I think you want `render.call(this)` here. See [these docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) for more info.

plugins/homepage/web_client/views/FrontPageView.js
```diff
@@ -10,7 +10,23 @@ wrap(FrontPageView, 'render', function (render) {
         method: 'GET',
         url: 'homepage/markdown'
     }).done(_.bind(function (resp) {
-        this.$el.html(renderMarkdown(resp['homepage.markdown']));
+        if (resp['homepage.markdown'] === '') {
```
brianhelba: I think it's fine to just check truthiness in all of these cases. If for some reason the value is `null` or `undefined`, we want to ensure the behavior stays the same.

plugins/homepage/web_client/views/ConfigView.js
```diff
@@ -17,7 +17,54 @@ var ConfigView = View.extend({
...
+                value: this.welcomeText.val()
+            },
+            {
+                key: 'homepage.logo',
+                value: this.logoUrl
             }]);
+        },
+
+        'change #g-homepage-logo': function (event) {
+            var reader = new FileReader();
```
brianhelba: Encoding this as a data URI is more complicated, will require the entire image to be stored in the database as a setting, and will require the entire image to be re-downloaded inside the REST endpoint every time the homepage is loaded. Instead, we should make this be an ordinary Girder file upload, which will upload its content into the existing `Homepage Assets` collection. The setting can then store the `fileId` of the uploaded image, and use it to construct a local URL to embed the image (which the browser can cache).

plugins/homepage/web_client/views/ConfigView.js
```diff
@@ -17,7 +17,54 @@ var ConfigView = View.extend({
...
+            this.$('.g-preview-logo').removeClass('hidden');
+            /* Clear the value of the hidden file input to avoid 'Chrome'
+            to not triger the change event when the same image is set,
+            this line fix a bug */
+            this.$('#g-homepage-logo').val(null);
+        },
+
+        'click #g-homepage-default-logo-btn': function (event) {
+            this.$('.g-preview-logo').attr({
+                'src': require('girder/assets/Girder_Mark.png'),
```
brianhelba: Let's not call `require` twice like this. It's harmless, but slow.

plugins/homepage/server/__init__.py
```diff
@@ -23,7 +23,12 @@
 from girder.utility import setting_utilities
 from girder.utility.model_importer import ModelImporter
 
-KEY = 'homepage.markdown'
+MARKDOWN = 'homepage.markdown'
```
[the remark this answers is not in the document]
brianhelba: Without default settings, the plugin breaks the homepage when it's first enabled.

ronichoudhury:
Thanks @brianhelba for the comprehensive review.

I just spoke to Brian in meatspace and he will work with @Pierre-Assemat to resolve the issues he highlighted.

plugins/homepage/web_client/views/ConfigView.js
opadron: I don't know if this is relevant, but for QIDW (just before work on this PR began), I did something very similar where the logo was an item uploaded to girder, and then the fileId was the setting.  But I also took URI construction out of the equation for clients by moving that logic over to a dedicated server-side route that would just look up the setting, and return a redirect to the right file.  This way, clients can point to a specific URI and know that they'll end up fetching the right file, even after changes to the settings.

brianhelba:
@Pierre-Assemat I've pushed some updates to reorganize the plugin (making it more maintainable) and to store uploaded logos as Girder files.

Please try out the updates, and make sure everything works for you. Also, I didn't spend any real time styling the updated logo setting on the config page, so if you want to spend some time on the CSS, feel free.

Pierre-Assemat:
I notice a particular behavior with the `uploadWidget` :
- I can upload only one file, and after it's impossible to change it except with the `use default` button, I think is the behavior of the widget and normally an admin user won't to change the logo many times without seeing the render of the front page, what do you think ? @brianhelba

brianhelba:
@Pierre-Assemat That's a good point. I think in general, the `UploadWidget` should have an option to "redo this upload". Feel free to [file an issue](https://github.com/girder/girder/issues/new) for this idea, so we can talk about implementing it later.

For the purposes of this PR, since it's only admins who are uploading, it's fine if we expect users to just refresh the page to upload something different. If `UploadWidget` had that feature, we'd enable it for this page, but I don't think it's worth it to implement here alone.

PR #2330, opened 2017-09-06
Properly encode filenames in Content-Disposition headers.

manthey:
Prior to this, names that included quotes or unicode codepoints might produce broken Content-Dispositions.

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -281,7 +281,7 @@ def setContentDisposition(filename, disposition='attachment', setHeader=True):
     safeFilename = unicodedata.normalize('NFKD', filename).encode('iso8859-1', 'ignore')
     utf8Filename = filename.encode('utf8', 'ignore')
     value = disposition + b'; filename="' + safeFilename.replace(
-        '\\', '\\\\').replace(b'"', b'\\"') + b'"'
+        b'\\', b'\\\\').replace(b'"', b'\\"') + b'"'
     if safeFilename != utf8Filename:
         quotedFilename = six.moves.urllib.parse.quote(utf8Filename)
         if not isinstance(quotedFilename, six.binary_type):
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -269,9 +269,11 @@ def setContentDisposition(filename, disposition='attachment', setHeader=True):
         Content-Disposition header, but do not set it.
     :returns: the content-disposition header value.
     """
-    if ((disposition not in ('inline', 'attachment') and
-         disposition and not disposition.startswith('form-data')) or not filename):
-        return
+    if (not disposition or (disposition not in ('inline', 'attachment') and
+                            not disposition.startswith('form-data'))):
+        raise RestException('Error: Content-Disposition is not a recognized value.')
+    if not filename:
+        raise RestException('Error: Content-Disposition filename is empty.')
     if not isinstance(disposition, six.binary_type):
         disposition = disposition.encode('iso8859-1', 'ignore')
     if not isinstance(filename, six.text_type):
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -270,7 +270,7 @@ def setContentDisposition(filename, disposition='attachment', setHeader=True):
     :returns: the content-disposition header value.
     """
     if ((disposition not in ('inline', 'attachment') and
-         not disposition.startswith('form-data')) or not filename):
+         disposition and not disposition.startswith('form-data')) or not filename):
         return
     if not isinstance(disposition, six.binary_type):
         disposition = disposition.encode('iso8859-1', 'ignore')
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -258,8 +258,8 @@ def setContentDisposition(filename, disposition='attachment', setHeader=True):
     Set the content disposition header to either inline or attachment, and
     specify a filename that is properly escaped.  See
     developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition,
-    tools.ietf.org/html/rfc2183, and tools.ietf.org/html/rfc6266 for
-    specificatioons and details.
+    tools.ietf.org/html/rfc2183, tools.ietf.org/html/rfc6266, and
+    tools.ietf.org/html/rfc5987 for specifications and details.
 
     :param filename: the filename to add to the content disposition header.
     :param disposition: either 'inline' or 'attachment'.  None is the same as
@@ -271,7 +271,8 @@ def setContentDisposition(filename, disposition='attachment', setHeader=True):
     """
     if (not disposition or (disposition not in ('inline', 'attachment') and
                             not disposition.startswith('form-data'))):
-        raise RestException('Error: Content-Disposition is not a recognized value.')
+        raise RestException(
+            'Error: Content-Disposition (%r) is not a recognized value.' % disposition)
     if not filename:
         raise RestException('Error: Content-Disposition filename is empty.')
     if not isinstance(disposition, six.binary_type):
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -26,6 +26,7 @@ import posixpath
 import six
 import sys
 import traceback
+import unicodedata
 
 from . import docs
 from girder import events, logger, logprint
@@ -255,7 +256,10 @@ def setCurrentUser(user):
 def setContentDisposition(filename, disposition='attachment', setHeader=True):
     """
     Set the content disposition header to either inline or attachment, and
-    specify a filename that is properly escaped.
+    specify a filename that is properly escaped.  See
+    developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition,
+    tools.ietf.org/html/rfc2183, and tools.ietf.org/html/rfc6266 for
+    specificatioons and details.
 
     :param filename: the filename to add to the content disposition header.
     :param disposition: either 'inline' or 'attachment'.  None is the same as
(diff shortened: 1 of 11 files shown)


girder/api/rest.py
```diff
@@ -252,6 +252,45 @@ def setCurrentUser(user):
...
+    # should work on all browsers/
+    safeFilename = filename.encode('ascii', 'ignore').replace(b'"', b'')
+    encodedFilename = filename.encode('utf8', 'ignore')
+    value = disposition + b'; filename="' + safeFilename + b'"'
+    if safeFilename != encodedFilename:
+        quotedFilename = six.moves.urllib.parse.quote(encodedFilename)
+        if not isinstance(quotedFilename, six.binary_type):
+            quotedFilename = quotedFilename.encode('ascii', 'ignore')
+        value += b'; filename*=UTF-8\'\'' + quotedFilename
+    if not isinstance(value, six.text_type):
```
brianhelba: Why would `value` not be a `six.text_type` at this point?

girder/api/rest.py
```diff
@@ -252,6 +252,45 @@ def setCurrentUser(user):
...
+
+    :param filename: the filename to add to the content disposition header.
+    :param disposition: either 'inline' or 'attachment'.  None is the same as
+        'attachment'.  Any other value skips setting the content disposition
+        header.
+    :param setHeader: if False, return the value that would be set to the
+        Content-Disposition header, but do not set it.
+    :returns: the content-disposition header value.
+    """
+    if disposition is None:
```
brianhelba: I think this is unnecessary complexity. The default can simply be `'attachment'`, and `None` should be an error (triggered implicitly when `.encode`) is called on it. The only time we need to use a default parameter value of `None` is when the value would be a mutable object.

girder/api/rest.py
```diff
@@ -252,6 +252,45 @@ def setCurrentUser(user):
     cherrypy.request.girderUser = user
 
 
+def setContentDisposition(filename, disposition='attachment', setHeader=True):
+    """
+    Set the content disposition header to either inline or attachment, and
+    specify a filename that is properly escaped.
```
brianhelba: Can you include a reference to the RFC that this should confirm to?

Also, did you review the relevant RFC, and how well does this implementation conform to it?

girder/api/rest.py
```diff
@@ -252,6 +252,45 @@ def setCurrentUser(user):
...
+    if not isinstance(disposition, six.binary_type):
+        disposition = disposition.encode('ascii', 'ignore')
+    if not isinstance(filename, six.text_type):
+        filename = filename.decode('utf8', 'ignore')
+    # Using ASCII for the safeFilename is more restrictive than necessary, but
+    # should work on all browsers/
+    safeFilename = filename.encode('ascii', 'ignore').replace(b'"', b'')
+    encodedFilename = filename.encode('utf8', 'ignore')
+    value = disposition + b'; filename="' + safeFilename + b'"'
+    if safeFilename != encodedFilename:
```
brianhelba: [Flask's implementation of this](https://github.com/pallets/flask/blob/master/flask/helpers.py#L546) seems a lot cleaner on the surface. Do you think it would be worthwhile to crib more of what they have (potentially utilizing the `unicodedata` library)?

manthey: Flask defers a bunch of the work to werkzeug.  In werkzeug's Header() class sending a list of values is supported, which is then encoded by werkzeug, handling unicode and string/bytes appropriately.  Multiple parameters are added in an arbitrary order (so in Flask, the `filename*` parameter might be before the `filename` parameter), which doesn't seem to affect any browser (see http://greenbytes.de/tech/tc2231/).

Flask uses the `unicodedata` library to decompose the unicode before making the fallback `filename` value, which is actually a clever way to de-accent characters where possible while converting to ISO-8859-1.

manthey: We'd have to modify setResponseHeader to accept a werkzeug-like dictionary of parameters to make our implementation look as 'clean' as Flask's.

girder/api/rest.py
manthey: As often the MDN page has a good summary of the issue and links to the relevant RFCs (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition), and see https://tools.ietf.org/html/rfc2183 and https://tools.ietf.org/html/rfc6266.

I'll add a comment.

manthey:
Adding a special case to `setResponseHeader` adds complexity unless we add a werkzeug-like ability to add a dictionary of values to the header.  What would `setResponseHeader` do if sent 'Content-Disposition' and 'inline; filename="<some unicode value>"`?  Would it try to extract out the improperly formed filename and reencode it?  Something else?  If we accepted `setResponseHeader('Content-Disposition', 'inline', {'filename': <some name>})`, then reencoding `filename` and possibly adding `filename*` could make sense.

girder/api/rest.py
manthey: I'll remove the check.

tests/cases/file_test.py
```diff
@@ -781,6 +784,18 @@ def testFilesystemAssetstore(self):
         copyTestFile = self._testUploadFile('helloWorld1.txt')
         self._testCopyFile(copyTestFile)
 
+        # Test unicode filenames for content disposition.  The test name has
+        # quotes, a Linear-B codepoint, Cyrllic, Arabic, Chinese, and an emoji.
+        filename = u'Unicode "sample" \U00010088 ' + \
```
brianhelba: I think it'd be a good idea to put this test string somewhere general, and start using it as test data for any place that should be expected to handle general strings (which includes Unicode, quotes, etc.) gracefully.

There is some risk of "overfitting" to this one test string, but if we think of new corner cases that challenge string handling, we can just add them to the master string.

tests/cases/file_test.py
```diff
@@ -781,6 +784,18 @@ def testFilesystemAssetstore(self):
...
+                   u'\u043e\u0431\u0440\u0430\u0437\u0435\u0446 ' + \
+                   u'\u0639\u064a\u0646\u0629 \u6a23\u54c1 \U0001f603'
+        file = self._testUploadFile(filename)
+        file = self.model('file').load(file['_id'], force=True)
+        testval = 'filename="%s"; filename*=UTF-8\'\'%s' % (
```
brianhelba: Since this probably most useful as a regression test, maybe it's actually better to hardcode the result that we're expecting to be in `filename*`, so we implicitly test the full stack that `setContentDisposition` relies on. If you still want the more pure unit test, then you can still continue to dynamically compute the result, but compare it to the hardcoded string.

@manthey What do you think?

manthey: I'll change this to a hardcoded value (not today, though).

tests/cases/file_test.py
manthey: We probably should have a set of test strings and one master complex string like this one.  I could see a pure right-to-left string giving different results than a mixed direction string, for instance.

tests/cases/file_test.py
brianhelba: My own trial (with just `print(repr(...))`) indicates that the values may be respectively:
`'Unicode \\"sample\\"     '`
`'Unicode%20%22sample%22%20%F0%90%82%88%20%D0%BE%D0%B1%D1%80%D0%B0%D0%B7%D0%B5%D1%86%20%D8%B9%D9%8A%D9%86%D8%A9%20%E6%A8%A3%E5%93%81%20%F0%9F%98%83'`
but I'll trust however you want to encode them in this file.

girder/api/rest.py
```diff
@@ -252,6 +253,48 @@ def setCurrentUser(user):
...
+    :param disposition: either 'inline' or 'attachment'.  None is the same as
+        'attachment'.  Any other value skips setting the content disposition
+        header.
+    :param setHeader: if False, return the value that would be set to the
+        Content-Disposition header, but do not set it.
+    :returns: the content-disposition header value.
+    """
+    if (not disposition or (disposition not in ('inline', 'attachment') and
+                            not disposition.startswith('form-data'))):
+        raise RestException('Error: Content-Disposition is not a recognized value.')
```
zachmullen: It's always nice in error messages like this to include the offending value in the message. :)

girder/api/rest.py
```diff
@@ -252,6 +253,48 @@ def setCurrentUser(user):
...
+    if not filename:
+        raise RestException('Error: Content-Disposition filename is empty.')
+    if not isinstance(disposition, six.binary_type):
+        disposition = disposition.encode('iso8859-1', 'ignore')
```
zachmullen: Why ISO 8859-1? I couldn't find mention of it in the standard mentioned in the docstring, is it mentioned somewhere further up the RFC chain?

zachmullen: Nevermind, I found it. [For posterity](https://tools.ietf.org/html/rfc5987)

girder/api/rest.py
manthey: Done.

girder/api/rest.py
manthey: Added RFC 5987 link into the comments.

brianhelba: https://tools.ietf.org/html/rfc5987#section-1

girder/api/rest.py
```diff
@@ -271,7 +271,8 @@ def setContentDisposition(filename, disposition='attachment', setHeader=True):
     """
     if (not disposition or (disposition not in ('inline', 'attachment') and
                             not disposition.startswith('form-data'))):
-        raise RestException('Error: Content-Disposition is not a recognized value.')
+        raise RestException(
+            'Error: Content-Disposition (%r) is not a recognized value.' % disposition)
```
brianhelba: The use of `%r` is definitely more appropriate here. We should remember to use this in place of `%s` when we display erroneous values throughout the code.

PR #2332, opened 2017-09-07
Add a girder-client console script

danlamanna:
> It was somewhat confusing that the package name on PyPI is
"girder-client" while the actual executable is girder-cli. This adds a
duplicate console script named girder-client, effectively aliasing
girder-client to girder-cli.

### DIFF
--- a/clients/python/setup.py
+++ b/clients/python/setup.py
@@ -60,7 +60,8 @@ setup(
     zip_safe=False,
     entry_points={
         'console_scripts': [
-            'girder-cli = girder_client.cli:main'
+            'girder-cli = girder_client.cli:main',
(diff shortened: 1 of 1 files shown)


danlamanna:
Unrelated, it seems like we may want to remove/replace the "Development Status :: 3 - Alpha" classifier with "Development Status :: 5 - Production/Stable".

danlamanna:
> Should we change the docs to use the new one? Should we deprecate the
> old one, or are we OK with just keeping both?

In an ideal world I would say deprecate girder-cli and keep
girder-client, but I'm not sure how to go about deprecating an entire
console script. Perhaps a warning on each command usage to stderr? Could
that potentially break any existing automation built with girder-cli?

PR #2334, opened 2017-09-08
Color banner setting in System Configuration

Pierre-Assemat:


brianhelba:
@Pierre-Assemat I'd like to review this before it's merged. Is it ready now, or still a WIP?

ronichoudhury:
@brianhelba I believe this is ready for review, so please go ahead.

I'll leave it to Pierre to remove the WIP tag.

ronichoudhury:
Actually there is one problem with CSS positioning for a button, but Pierre and I discussed the fix and I think he's going to apply that. It's a small fix, so you can wait until that shows up, or just review now.

girder/utility/webroot.mako
```diff
@@ -24,7 +24,8 @@
                 el: 'body',
                 parentView: null,
                 contactEmail: '${contactEmail | h}',
-                brandName: '${brandName | h}'
+                brandName: '${brandName | h}',
+                bannerColor: '${bannerColor}'
```
brianhelba: Changes in #2325 are causing merge conflicts now. Also, all edits to `webroot.mako` should now use the new conventions established by that PR.

girder/models/setting.py
```diff
@@ -171,6 +171,12 @@ def validateCoreBrandName(doc):
             raise ValidationException('The brand name may not be empty', 'value')
 
     @staticmethod
+    @setting_utilities.validator(SettingKey.BANNER_COLOR)
+    def validateCoreBannerColor(doc):
+        if not doc['value']:
```
brianhelba: We can probably do a better validation than just this. A regex would work well here.

clients/web/src/views/layout/HeaderView.js
```diff
@@ -40,7 +40,7 @@ var LayoutHeaderView = View.extend({
         this.$el.html(LayoutHeaderTemplate({
             brandName: this.brandName
         }));
-
+        this.$('.g-header-wrapper').css('background-color', this.bannerColor);
```
brianhelba: In my opinion, in order of priority, I think CSS information ought to be set:
1. In a stylesheeet
2. On an explicit `style="..."` attribute in the HTML
3. At runtime in Javascript

While 1 is not possible to do dynamically, we could do 2 by passing the value into the Pug template.

This isn't ideal, but IMO, it's a little better at maintaining the (content / style / behavior) distinction of (HTML / CSS / Javascript), and may make it slightly easier for a naive developer to notice where the header bar color is set dynamically. However, maybe other @girder/developers feel differently?

clients/web/src/views/body/SystemConfigurationView.js
```diff
@@ -65,7 +65,10 @@ var SystemConfigurationView = View.extend({
                 this.$('#g-settings-error-message').text(resp.responseJSON.message);
             }, this));
         },
-        'click #g-restart-server': restartServerPrompt
+        'click #g-restart-server': restartServerPrompt,
+        'click #g-core-banner-default-color': function (event) {
```
brianhelba: Since we're not using `event`, you can omit it. Javascript is fine with passing too many parameters (they are always available in the magic `arguments` variable) and not enough parameters (arguments default to `undefined` in the callee) into a function.

brianhelba: To be more specific, only traditional `function` functions have a magic `arguments` variable. Arrow `=>` functions do not. Both types of functions can still take too many or too few parameters.

girder/models/setting.py
Pierre-Assemat: Do you have an idea why this kind of verification doesn't work ? 

screenshot from 2017-09-11 11-01-44

jbeezley: You're throwing an error if the regex matches.  Try inverting the condition.

zachmullen: That regex should work for hex colors. Is there a reason we want to limit this to just hex? I could see an admin wanting to just type one of the named colors like "white".

Pierre-Assemat: Oh right ! Yes it's true, I will change that

brianhelba: @zachmullen When this widget is clicked, a color picker appears, which automatically fills in in hex values, and does not seem to allow the manual entry of non-hex color values. While it's technically possible to submit a color alias via an API call, I think it would simplify a lot of assumptions (maybe with security implications) if we just require hex colors to be used here.

girder/utility/webroot.mako
Pierre-Assemat: Yes of course, do you think the `bannerColor` need to be escape in the same way ?

girder/models/setting.py
Pierre-Assemat: screenshot from 2017-09-11 13-22-05

If we enter a string like "blue", "red", etc.. The color change and the string is converted to an hexadecimal value (Mozilla color picker)
So it doesn't raise an error even if we only check for hexadecimal value
Moreover I think the check of different color name will be such a pain and not very fundamental

clients/web/src/templates/layout/layoutHeader.pug
```diff
@@ -1,4 +1,4 @@
-.g-header-wrapper
+.g-header-wrapper(style='background-color:' + bannerColor)
```
brianhelba: Since you're setting `background-color` here, please remove it from [the Stylus file](https://github.com/girder/girder/blob/master/clients/web/src/stylesheets/layout/header.styl#L4).

clients/web/src/views/body/SystemConfigurationView.js
```diff
@@ -65,7 +65,10 @@ var SystemConfigurationView = View.extend({
                 this.$('#g-settings-error-message').text(resp.responseJSON.message);
             }, this));
         },
-        'click #g-restart-server': restartServerPrompt
+        'click #g-restart-server': restartServerPrompt,
+        'click #g-core-banner-default-color': function () {
+            this.$('#g-core-banner-color').attr('value', this.defaults['core.banner_color']);
```
brianhelba: Please check out [this helpful post](http://lucybain.com/blog/2014/attribute-vs-property/) on the distinction between an attribute and a property. If you needed to modify one of these directly, you'd want to use the [jQuery `.prop` method](http://api.jquery.com/prop/) method, since you'd be modifying a property of the DOM element, as opposed to an attribute of the HTML.

However, for `input` elements, jQuery helpfully provides [a `.val` mutator method](http://api.jquery.com/val/#val2). You should use that.

clients/web/src/stylesheets/body/systemConfig.styl
```diff
@@ -5,6 +5,18 @@
...
+  display block
+
+  #g-core-banner-color
+    display inline
+    max-width 15%
+
+  #g-core-banner-default-color
+    width 65px
+    display inline
+    vertical-align top
```
brianhelba: I don't see what this `vertical-align` is changing. Why is it here?

clients/web/src/stylesheets/body/systemConfig.styl
```diff
@@ -5,6 +5,18 @@
   .bootstrap-switch-handle-off
       word-wrap normal
 
+#g-core-banner-color-container
+  display block
```
brianhelba: `div` is a block element by definition. This doesn't need to be set explicitly.

clients/web/src/stylesheets/body/systemConfig.styl
```diff
@@ -5,6 +5,18 @@
...
+
+  #g-core-banner-color
+    display inline
+    max-width 15%
```
brianhelba: On larger displays, making this element a percentage of the overall viewport width looks strange:
image

Can you change it to use a fixed number of pixels instead? I'm fine if it continues to use either `max-width` or `width` (depending on whether you want to allow it to shrink or not).

brianhelba:
I'm also a little concerned that Girder's branding text will no longer be legible if a lighter color is used for the banner:
image

[This helpful post](https://stackoverflow.com/a/3943023) provides a compact formula for determining whether to set text (in this case, [the `color` property](https://github.com/girder/girder/blob/master/clients/web/src/stylesheets/layout/header.styl#L5)) to `#000000` or `#ffffff`. I'd suggest:
* dropping that logic (preferably, the latter W3C-compliant version) into the `LayoutHeaderView.render` method
* directly passing the ideal text color as another variable into `layoutHeader.pug`
* set `color` inline in `layoutHeader.pug` (in addition to `background-color`)
* remove the hardcoded `color` from the Stylus file

Please let me know if you need any clarification or help, or if you think this will take too long.

jcfr:
> directly passing the ideal text color as another variable into

:+1: 

Color pick could propose related pantone colors ... (look like there are few libraries on npm related to pantone.)

jeffbaumes:
@brianhelba @jcfr I feel those other color features could potentially be pushed to another issue/PR - does not seem critical to this PR.

If we want to support lighter banner colors, I would recommend putting both colors (foreground, background) in direct control of the user (nothing automatic). If anything, I would perhaps show a warning showing if the contrast is known to be too low ([this code](https://stackoverflow.com/a/9733420) could aid with that - contrast less than 3 could show a warning).

brianhelba:
@jeffbaumes If implementing the contrast feature delays this PR, I intend to defer it. The feature *should* just be a few lines of copy-pasted code, but I'll sync with @Pierre-Assemat on what the actual complexity looks like.

I would actually oppose the addition of a second "foreground color" setting. As currently architected, this would have to be another core setting (which currently require a significant amount of code to maintain), and it would have even more limited oppertunities for generalizability than the setting added by this PR.

brianhelba:
This LGTM!

@Pierre-Assemat I refactored `LayoutHeaderView._getTextColor` a bit to make it clearer.

PR #2335, opened 2017-09-09
Rest jobinfospec

dorukozturk:
This pr is for attaching jobInfoSpec to the remote jobs and for specifying a parent job on the girder end, if there is a parent specified on the celery end.


plugins/jobs/server/utils.py
```diff
@@ -0,0 +1,28 @@
...
+def jobInfoSpec(job, token=None, logPrint=True):
+    """
+    Build the jobInfo specification for a task to write status and log output
+    back to a Girder job.
+
+    :param job: The job document representing the worker task.
+    :type job: dict
+    :param token: The token to use. Creates a job token if not passed.
+    :type token: str or dict
+    :param logPrint: Whether standard output from the job should be
```
cjh1: show be printed?

cjh1: What about stderr?

cjh1: Sorry didn't see it was WIP

dorukozturk: It is totally fine. Actually after talking to @kotfic , we decided to go with an event based system. I will do some refactoring.

plugins/jobs/server/job_rest.py
```diff
@@ -81,8 +82,8 @@ def listJobs(self, userId, parentJob, types, statuses, limit, offset, sort):
         .param('title', '', required=True)
         .param('type', '', required=True)
         .modelParam('parentId', 'ID of the parent job.', model='job',
-                    plugin='jobs', level=AccessType.ADMIN,
-                    destName='parentJob', paramType='query', required=False)
+                    plugin='jobs', destName='parentJob',
+                    paramType='query', required=False, force=True)
```
cjh1: What does this force parameter do here?

plugins/worker/server/__init__.py
```diff
@@ -212,11 +212,33 @@ def validTransitions(event):
         event.preventDefault().addResponse(states)
 
 
+def attachParentJob(event):
+    """Attach parentJob before a model is saved."""
+    jobModel = ModelImporter.model('job', 'jobs')
+    job = event.info
+    try:
+        celeryParentTaskId = job['celeryParentTaskId']
+        parentJob = jobModel.findOne({'celeryTaskId': celeryParentTaskId})
```
cjh1: We should ensure a index on celeryTaskId

dorukozturk:
@cjh1 I changed the things you mentioned. Also added a test. PTALOneMoreTime?

PR #2352, opened 2017-09-20
Fully implement and test projection filtering

### LINKED ISSUE #2343: Fully implement _isInclusionProjection
[`Model._isInclusionProjection`](https://github.com/girder/girder/blob/master/girder/models/model_base.py#L572) needs to support exclusion projections.

sutartmelson:
Fixes #2343

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -569,32 +569,103 @@ class Model(ModelImporter):
         """
         return 1
 
-    def _isInclusionProjection(self, fields):
-        """
-        Test whether a projection filter is an inclusion filter (whitelist) or exclusion
-        projection (blacklist) of fields, as defined by MongoDB find() method `projection` param.
 
-        :param fields: A mask for filtering result documents by key, or None to return the full
-            document, passed to MongoDB find() as the `projection` param.
-        :type fields: `str, list of strings or tuple of strings for fields to be included from the
-            document, or dict for an inclusion or exclusion projection`.
-        """
-        if fields is None:
-            return False
+def _isInclusionProjection(fields):
+    """
+    Test whether a projection filter is an inclusion filter (whitelist) or exclusion
+    projection (blacklist) of fields, as defined by MongoDB find() method `projection` param.
 
+    :param fields: A mask for filtering result documents by key, or None to return the full
+        document, passed to MongoDB find() as the `projection` param.
+    :type fields: `str, list of strings or tuple of strings for fields to be included from the
+        document, or dict for an inclusion or exclusion projection`.
+    """
+    if fields is None:
+        return False
+
+    if not isinstance(fields, dict):
+        # If this is a list/tuple/set, that means inclusion
+        return True
+
+    for k, v in six.viewitems(fields):
+        if k != '_id':
+            # We are only allowed either inclusion or exclusion keys in a dict, there can be no
+            # mixing of these, with the only exception being that the `_id` key can be set as
+            # an exclusion field in a dict that otherwise holds inclusion fields.
+            return v
+
+    # Empty dict or just _id field
+    return fields.get('_id', True)
+
+
+def _overwriteFields(fields, overwrite):
+    """
+    Overwrite the projection filter to either include (in the case of an inclusion filter) or
+    not exclude (in the case of an exclusion filter) the contents of overwrite.
+
+    :param fields: A mask for filtering result documents by key, or None to return the full
+        document, passed to MongoDB find() as the `projection` param.
+    :type fields: `str, list of strings, or tuple of strings for fields to be included from the
+        document, or dict for an inclusion or exclusion projection`.
+    :param overwrite: Additional document key(s) to be included or not excluded in fields.
+    :type overwrite: `str, list of strings, or tuple of strings for fields to be included from the
+        document.`
+    """
+    if fields is None:
+        return fields
+
+    overwrite = list(overwrite)
+    if _isInclusionProjection(fields):
         if not isinstance(fields, dict):
-            # If this is a list/tuple/set, that means inclusion
-            return True
+            # Inclusion projection (str, list, or tuple)
+            copy = list(fields)
+            for entry in overwrite:
+                if entry not in copy:
+                    copy.append(entry)
+        else:
+            # Inclusion projection (dict)
+            copy = dict(fields)
+            for entry in overwrite:
+                copy[entry] = True
+    else:
+        # Exclusion projection (dict)
+        copy = dict(fields)
+        for entry in overwrite:
+            if entry in copy.keys():
+                del copy[entry]
+    return copy
 
+
+def _removeOverwrittenFields(doc, fields):
+    """
+    Edit the document to be consistent with what the user originally requested, undoing what may
+        have been overwritten by _overwriteFields().
+
+    :param doc: A document returned by MongoDB find()
+    :type doc: dict
+    :param fields: The origian mask for filtering result documents by key, as specified by the user.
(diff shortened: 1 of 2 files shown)


girder/models/model_base.py
```diff
@@ -569,32 +569,103 @@ def subtreeCount(self, doc):
...
+    overwrite = list(overwrite)
+    if _isInclusionProjection(fields):
         if not isinstance(fields, dict):
-            # If this is a list/tuple/set, that means inclusion
-            return True
+            # Inclusion projection (str, list, or tuple)
+            copy = list(fields)
+            for entry in overwrite:
+                if entry not in copy:
+                    copy.append(entry)
```
brianhelba: This would be simpler (and theoretically more efficient) as:
```
copy = list( set(fields) | set(overwrite) )
```

girder/models/model_base.py
```diff
@@ -569,32 +569,103 @@ def subtreeCount(self, doc):
...
+        else:
+            # Inclusion projection (dict)
+            copy = dict(fields)
```
brianhelba: FYI, there's also:
```
copy = dict(fields)
copy.update( dict.fromkeys(overwrite, True) )
```
but it's your call which you prefer.

girder/models/model_base.py
```diff
@@ -569,32 +569,103 @@ def subtreeCount(self, doc):
...
+            for entry in overwrite:
+                copy[entry] = True
+    else:
+        # Exclusion projection (dict)
+        copy = dict(fields)
+        for entry in overwrite:
+            if entry in copy.keys():
```
brianhelba: If you're going to check if something is in a `dict`, you should definitely just do:
```
if entry in copy:
```
otherwise, you're creating an entire list / view of the keys every time through the loop, then doing a list search, which itself is linear-time.

Alternatively, you could just omit the check entirely, and do:
```
for entry in overwrite:
    copy.pop(entry, None)
```

girder/models/model_base.py
```diff
@@ -569,32 +569,103 @@ def subtreeCount(self, doc):
...
+            if not v:
+                doc.pop(k, None)
+            else:
+                whitelist.append(k)
+        if whitelist:
+            for k in list(doc.keys()):
+                if k not in whitelist:
+                    del doc[k]
+    else:
+        fields = list(fields)
```
brianhelba: No need to make a copy of `fields` here (iterating through a `tuple` works the same, anyway). If it's a string, we would need to cast it to a list.

Also, let's add a test for the case when `fields` is a string.

girder/models/model_base.py
```diff
@@ -569,32 +569,103 @@ def subtreeCount(self, doc):
...
+        for k in list(doc.keys()):
```
brianhelba: If you want to iterate over keys, it's most memory-efficient to do:
```
for k in six.viewkeys(doc):
```
 in Girder, or:
```
for k in doc.viewkeys():
```
in Python2, or
```
for k in doc.keys()
```
in Python3.

brianhelba: As this is now, it's got time complexity `O(doc * fields)`. It's possible to get it down to `O(fields * log(doc))`, which for `fields << doc` would actually matter.

girder/models/model_base.py
```diff
@@ -1267,24 +1338,16 @@ def load(self, id, level=AccessType.ADMIN, user=None, objectId=True,
 
         # Ensure we include access and public, they are needed by requireAccess
         loadFields = copy.copy(fields)
```
brianhelba: No need to copy here, if we're doing it inside `_overwriteFields`.

tests/cases/model_test.py
```diff
@@ -64,6 +65,96 @@ def setUp(self):
...
+        overrideFields = ['access', 'public']
+
+        retval = _overwriteFields(inclusionProjDict, overrideFields)
+        self.assertEqual(retval, inclusionProjDict)
+        retval = _overwriteFields(inclusionProjList, overrideFields)
+        self.assertEqual(retval, inclusionProjList)
+        retval = _overwriteFields(exclusionProjDict, ['newValue'])
+        self.assertEqual(retval, exclusionProjDict)
+        retval = _overwriteFields(inclusionProjDict, ['newValue'])
+        self.assertEqual(retval, {
```
brianhelba: Let's test the fact that `retval` should not be mutated by `_overwriteFields`.

girder/models/model_base.py
sutartmelson: The problem I was having is that the dictionary is changing since I'm deleting keys. So copying doc.keys() to a list solved that problem.

girder/models/model_base.py
zachmullen: This would be a great behavior for `dict.__sub__` IMO, but sadly that operator is not implemented.

tests/cases/model_test.py
sutartmelson: `retval` is never passed to `_overwriteFields`? I'm not really sure what that would be testing against?

Or are you saying the first argument, in this case `inclusionProjDict` should not be mutated by `_overwriteFields`?

brianhelba:
@sutartmelson Actually, there are [2 uncovered lines in `_isInclusionProjection`](https://codecov.io/gh/girder/girder/src/cbbf8411be944ab1759fda73145b150b6911fb75/girder/models/model_base.py#L573). Would it be easy to add another test statement to cover these?

sutartmelson:
@brianhelba yup, I'm on it.

brianhelba:
@manthey Can you take a look?

manthey:
It looks fine.

One petty comment: at least in early versions of python `newdict = original.copy()` was mildly faster than copying a dict via `newdict = dict(original)`, so to my eyes seeing the constructor is less efficient (and less obvious) than using `original.copy()` or `dict.copy(original)`.  I haven't checked, so it is distinctly possible that they are internally the same at this point.

girder/models/model_base.py
```diff
@@ -1251,40 +1313,36 @@ def load(self, id, level=AccessType.ADMIN, user=None, objectId=True,
...
-        :type fields: `str, list of strings or tuple of strings for fields to be included from the
-            document, or dict for an inclusion or exclusion projection`.
-        :param exc: If not found, throw a ValidationException instead of
-            returning None.
+        :type fields: list or dict
+        :param exc: If not found, throw a ValidationException instead of returning None.
         :type exc: bool
         :raises ValidationException: If an invalid ObjectId is passed.
         :returns: The matching document, or None if no match exists.
         """
```
brianhelba: No need to for an extra newline here.

girder/models/model_base.py
```diff
@@ -1251,40 +1313,36 @@ def load(self, id, level=AccessType.ADMIN, user=None, objectId=True,
...
         # Ensure we include access and public, they are needed by requireAccess
-        loadFields = copy.copy(fields)
-        if not force and self._isInclusionProjection(fields):
-            if isinstance(loadFields, dict):
-                loadFields['access'] = True
-                loadFields['public'] = True
-            else:
-                loadFields = list(set(loadFields) | {'access', 'public'})
+        loadFields = fields
+        overwriteFields = {'access', 'public'}
```
brianhelba: You can put this definition inside the `if not force:` block.

brianhelba:
@sutartmelson This still LGTM. Feel free to fix the nits raised by myself and @manthey in an additional commit if you'd like (I'll reapprove), or just merge this yourself if you're happy as-is.

brianhelba:
@sutartmelson Nevermind, let's merge as-is. I'll implement my suggestions in a forthcoming PR.

PR #2353, opened 2017-09-20
Create a "2.x-maintenance" branch, which is always synced with "master"

brianhelba:
This creates a `2.x-maintenance` branch, to correspond to the existing [`1.x-maintenance` branch](https://github.com/girder/girder/tree/1.x-maintenance). The semantics of `2.x-maintenance` are: "always points to the latest version (released or otherwise) in the 2.x.x series".

This provides a semantically stable ref for downstream plugins / projects to track after 3.0 has been released, before they are able to undertake the effort of a 2->3 migration. This concept (as `1.x-maintenance`) has already proven useful for some downstream users.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -107,8 +107,13 @@ Install from Git repository
 Obtain the Girder source code by cloning the Git repository on
 `GitHub <https://github.com>`_: ::
 
-    git clone https://github.com/girder/girder.git
+    git clone -b 2.x-maintenance https://github.com/girder/girder.git
     cd girder
+
+.. note:: Note, it is strongly recommended that downstream (i.e. for production or to
+   support plugin development) users installing from Git track the ``2.x-maintenance`` branch, as
+   this branch will always point to the latest version (which is typically pre-release) in the 2.x.x
+   series.
 
 To run the server, you must install some external Python package
 dependencies: ::
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -107,7 +107,7 @@ Install from Git repository
 Obtain the Girder source code by cloning the Git repository on
 `GitHub <https://github.com>`_: ::
 
-    git clone -b 2.x-maintenance https://github.com/girder/girder.git
+    git clone --branch 2.x-maintenance https://github.com/girder/girder.git
     cd girder
 
 .. note:: Note, it is strongly recommended that downstream (i.e. for production or to
--- a/circle.yml
+++ b/circle.yml
@@ -42,3 +42,4 @@ deployment:
     commands:
       - curl 'https://doc.esdoc.org/api/create' -X POST --data-urlencode "gitUrl=git@github.com:girder/girder.git"
       - ./scripts/galaxy_deploy.sh
+      - git push git@github.com:girder/girder.git master:2.x-maintenance


brianhelba:
I'd like to document this (along with the existing `1.x-maintenance` branch) somewhere, so we can make the semantic guarantees explicit. @girder/developers any suggestions for where to put this info?

brianhelba:
@girder/developers PTAL.

mgrauer:
I thought when you explained this to me you said 2.x-maintenance would track the latest 2.x release branch, (rather than as you said in the top comment here "released or otherwise")? Did I get that wrong?

What I like about the idea of the 2.x-maintenance and 1.x-maintenance branches tracking the releases is that there is a symmetry with the 3.x-maintenance branch--we can introduce that one to track the latest release in 3.x series, which gives downstreams an option to track released versions rather than tracking master, and also will allow downstreams not to have to change when at some point master points to 4.x code.

I do generally agree with the notion of a 1.x- and 2.x-maintenance branch (and even 3.x-) but want to get this detail nailed down.

brianhelba:
@mgrauer Thanks for clarifying this. It's possible that I said something different in my conversation with you, but I'm confident that this PR reflects the best way forward.

Right now, `1.x-maintenance` and `v1.7.1` do point to the same commit, but that it only incidental. When a new issue needs to be backported, it is applied to `1.x-maintenance`. If we want to do a release at that point, we can, but we are not obligated to.

This PR is creating a `2.x-maintenance` branch with the same semantics: always pointing to the *latest* 2.x commit, regardless of whether that's a release. Right now, this serves to provide a stable reference for downstreams to always track the bleeding edge of 2.x development. After the 3 release, it will also serve as the `HEAD` for new commits to the 2.x series (because tags can't be committed / merged to without detaching the `HEAD`). For both functions, it is necessary for `2.x-maintenance` to point to the latest commit, but not *necessarily* the latest release.

If a downstream wants the latest *release*, there has always been an easy way: install from `pip` using an abstract version specifier (i.e. `pip install girder == 2.*`).

brianhelba:
@girder/developers I've just added documentation, so this is ready for review. PTAL.

docs/installation.rst
```diff
@@ -107,9 +107,14 @@ Install from Git repository
 Obtain the Girder source code by cloning the Git repository on
 `GitHub <https://github.com>`_: ::
 
-    git clone https://github.com/girder/girder.git
+    git clone -b 2.x-maintenance https://github.com/girder/girder.git
```
manthey: This line bothers me, even though it is justified below.  It is unusual to show a `git clone` with the `-b` as part of standard installation instructions.  I think it might be better to have this without the branch, and have the code block showing its use inside the following note.

brianhelba: We have a separate page for a [Developer Installation](https://github.com/girder/girder/blob/master/docs/dev-installation.rst) of Girder, where we definitely should encourage people to check out `master`.

However, the line at hand is inside documentation referring to the installation process for general usage, where long-term stability could matter. Under those circumstances, it's far safer to put people on `2.x-maintenance` by default, rather than `master`.

I imagine that those users who are sophisticated enough to appreciate the difference between the two branches will easily be able to modify the `git` command to suit their needs (and honestly would probably be happier if we made the documentation more concise by eliding a lot of these commands). Less sophisticated or less attentive users may also not read the whole note or appreciate its implications, so I'd rather do the "best thing" by default, and place the caveats or alternatives deeper down where intrepid power users can still access them.

brianhelba:
@manthey Switched to the long-form command argument.

PR #2359, opened 2017-09-25
Remove unused and outdated stylus code

sutartmelson:
Supplemental PR to #2284 removing unused and outdated stylus code for the JSONEditor package

### DIFF
--- a/clients/web/src/stylesheets/widgets/metadataWidget.styl
+++ b/clients/web/src/stylesheets/widgets/metadataWidget.styl
@@ -51,39 +51,6 @@
   .ace_editor, textarea
     height 365px !important
 
-  .ace_content * // @stylint ignore
-    font-family monospace !important
-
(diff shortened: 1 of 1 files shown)


Pierre-Assemat:
I'm not sure but I think these changes will fix an error I have since few weeks when I try to compile the web client

screenshot from 2017-09-12 21-57-16

sutartmelson:
@Pierre-Assemat I think this will make that error go away, but not because the code being removed is erroneous. That error is more likely caused by not having the correct version of the JSONEditor package. 

The old package used a png, while the updated package is using an svg, so if you have the old package installed, it fails to find it. 

Try this inside of your girder dir:

```bash
rm -r node_modules
npm install
```

Pierre-Assemat:
The way I found to fix this during a short time is to delete the node_modules folder and reinstall the web client

PR #2360, opened 2017-09-25
Fix a race condition in homepage plugin test

brianhelba:
This test was failing non-deterministically. This fix uses a more reliable method to test whether the plugin settings page was saved. `girder.rest.numberOutstandingRestRequests()` is very unreliable and should be avoided whenever possible.

### DIFF
--- a/plugins/homepage/plugin_tests/homepageSpec.js
+++ b/plugins/homepage/plugin_tests/homepageSpec.js
@@ -48,7 +48,7 @@ function _verifyHomepageSettings(elem) {
     _verifyMarkdownContent(elem);
 }
 
-describe('homepage plugin test ', function () {
+describe('homepage plugin test', function () {
     it('registers an admin user', girderTest.createUser(
         'admin', 'admin@example.com', 'Mark', 'Down', 'password'
     ));
@@ -66,12 +66,13 @@ describe('homepage plugin test ', function () {
 
             _verifyMarkdownContent($('.g-markdown-preview'));
 
+            $('#g-alerts-container').empty();
             $('#g-homepage-form').submit();
         });
 
         waitsFor(function () {
-            return girder.rest.numberOutstandingRestRequests() === 0;
-        }, 'rest requests to finish');
+            return $('.alert:contains("Settings saved.")').length > 0;
+        }, 'settings to save');
     });
 
     it('verifies homepage content as admin user', function () {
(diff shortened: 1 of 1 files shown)


brianhelba:
@zachmullen That's why I added a new check after the first save. If every individual save is gated by this check, then there will never be multiple saves pending at once.

zachmullen:
Yeah. As long as whoever comes along and copypastes this in the future realizes that that part is necessary, we'll be fine. :)

brianhelba:
Indeed, this is a step backwards in my battle against [cargo cult programming](https://en.wikipedia.org/wiki/Cargo_cult_programming), but it may require more substantive changes to the test infrastructure to fix properly.

manthey:
@brianhelba When you say that "`girder.rest.numberOutstandingRestRequests()` is very unreliable", do you know an instance when it is inaccurate?  If so, is it over reporting or under reporting?  If that is the case, we should fix it.  Or do you mean that when tests are written that use it directly, that those tests are not waiting for the right thing?

brianhelba:
I can't strictly say that it's inaccurate, just that it's unreliable. Specifically, the `waitsFor` of `girder.rest.numberOutstandingRestRequests() === 0` allows code to proceed when we would intuitively think that there ought to be requests that are still pending full resolution. I don't know the root cause of this, but it may not necessarily be due to a simple "bug" in the code, but rather a set of less-tractable issues around when a request is "outstanding" vs "fully handled".

An example of this problem would be the case being addressed in this PR.

manthey:
I'm skeptical that `girder.rest.numberOutstandingRestRequests()` is unreliable -- I think it just isn't what you want.  There are many things that happen asynchronously, so the number of outstanding rest requests is almost never a sufficient condition.  The `waitsForLoad` and `waitsForDialog` functions are intended to wait for all asynchronous actions to resolve (but they could easily be insufficient).  Specifically, there are times when a request has returned, but due to animation, callbacks, and other delays, the expected data and associated DOM has not shown up yet.

Waiting for the alert toast to be visible is probably not what you really want either -- it only shows up for a finite duration.  The test will *most likely* poll it during that time, but if the test is run in parallel and the thread is resource starved, it *could* miss it.  This is a case where `waitsForLoad` would have been appropriate.

brianhelba:
@manthey I think I actually agree with your first point, so let me rephrase and clarify my earlier statement to say: "using `girder.rest.numberOutstandingRestRequests() === 0` is often an unreliable way to determine if an asynchronous operation has been fully resolved", which is often what we are really testing for.

Yes, the toast polling is not ideal either, it's just the best fix I could think of to get the tests working again. Further improvements are still welcome.

I don't believe `waitsForLoad` will help here, since the page is not actually loading or reloading: it's just saving in the background.

PR #2361, opened 2017-09-25
Fix a tooltip bug with the JSONEditor

### LINKED ISSUE #2358: JSONEditor popup bug
When editing metdata, if you switch the jsoneditor to `Code` the text pop up remains. If you swtch back to `Tree` another pop up remains. I'm not yet sure if this is an issue with the JSONEditor package or girder web. 
screenshot

brianhelba:
Fixes #2358

### DIFF
--- a/clients/web/src/views/widgets/MetadataWidget.js
+++ b/clients/web/src/views/widgets/MetadataWidget.js
@@ -151,6 +151,8 @@ var MetadatumEditWidget = View.extend({
         'click .g-widget-metadata-save-button': 'save',
         'click .g-widget-metadata-delete-button': 'deleteMetadatum',
         'click .g-widget-metadata-toggle-button': function (event) {
+            const target = $(event.currentTarget);
+            target.tooltip('destroy');
             var editorType;
             // @todo modal
             // in the future this event will have the new editorType (assuming a dropdown)
@@ -189,7 +191,9 @@ var MetadatumEditWidget = View.extend({
 
     deleteMetadatum: function (event) {
         event.stopImmediatePropagation();
-        var metadataList = $(event.currentTarget.parentElement).parent();
+        const target = $(event.currentTarget);
+        target.tooltip('destroy');
+        var metadataList = target.parent().parent();
         var params = {
             text: 'Are you sure you want to delete the metadatum <b>' +
                 _.escape(this.key) + '</b>?',
@@ -209,7 +213,9 @@ var MetadatumEditWidget = View.extend({
 
     cancelEdit: function (event) {
         event.stopImmediatePropagation();
-        var curRow = $(event.currentTarget.parentElement).parent();
+        const target = $(event.currentTarget);
+        target.tooltip('destroy');
+        var curRow = target.parent().parent();
         if (this.newDatum) {
             curRow.remove();
         } else {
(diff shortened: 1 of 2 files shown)


clients/web/src/views/widgets/MetadataWidget.js
```diff
@@ -317,7 +317,16 @@ var JsonMetadatumEditWidget = MetadatumEditWidget.extend({
...
             modes: ['code', 'tree'],
-            onError: function () {
+            onModeChange: () => {
+                // JSONEditor removes tooltip-containing elements without destroying their
+                // Bootstrap-styled tooltips first. Since this callback is not called until after
+                // the original tooltip-containing element is destroyed, the only way to clean up is
+                // to remove the tooltip from DOM.
+                // When used in an external app with no tooltip styling enabled, this call should
+                // have no effect (assuming there's no other ".tooltip" elements.
+                $('.tooltip').remove();
```
zachmullen: Should this be `this.$`?

brianhelba: Unfortunately, no. The tooltip is attached directly below `#g-app-body-container`, which is much higher in the hierarchy than this widget's element. I was considering using the selector `'#g-app-body-container>.tooltip'), but I wanted to keep it general enough to also correct tooltip problems if this was reused in another external app.

sutartmelson:
The popup still has weird behavior insofar as it doesn't disappear on its own, you must click somewhere first. But its definitely an improvement. 

As long as we're okay with that and you were already aware, looks good to me. I can approve it.

brianhelba:
@girder/developers  @sutartmelson The fact that button tooltips remain until clicking somewhere else is a natural consequence of accepting the [Bootstrap default of `{trigger: 'hover focus'}`](https://getbootstrap.com/docs/3.3/javascript/#tooltips-options) when initializing our tooltips. JSONEditor makes its buttons focusable, so the click causes the focus to stay on the button (and the tooltip to pop up) until another click clears the focus.

If we want, we could easily initialize tooltips with `{trigger: 'hover'}`, but it would make them inaccessible to users who are not using a traditional pointing device (e.g. mobile, keyboard selection, etc.). I don't have any personal feelings on that tradeoff.

brianhelba:
@sutartmelson @zachmullen Please test and try it now.

brianhelba:
@zachmullen Agreed. I'm not yet sure whether to totally blame this on Bootstrap (not giving us a cleaner way to veto tooltip initialization and not cleaning up old tooltips) or partially blame the code at hand (in both the JSONEditor library, and the Backbone architecture of the `MetadataWidget`-related classes).

I've scoped the workaround into a private method, and I plan to revist this as part of a larger overhaul of the way that our Views are initialized / destroyed.

zachmullen:
BTW, the candela test is failing with the same race condition -- is your earlier fix incorporated into this branch?

brianhelba:
@zachmullen No, #2369 isn't merged yet. I wanted to give it a little time to cool off, in case @jeffbaumes or other @girder/developers want to review it.

brianhelba:
@sutartmelson Does this fix all the issues you've found?

sutartmelson:
@brianhelba Yup!

PR #2362, opened 2017-09-26
[WIP] Plugin refactor

jbeezley:
In our meeting, there was a desire expressed to share my PoC refactor and to annotate the changes.  I think the easiest place to do that is through a PR.  There is much left unimplemented, but it should at least provide information about where I am headed.  These changes are not meant to be final in any way, and I may very well just abandon this branch and start over.

I'll comment on specific code blocks, but this is a general overview.
* `plugin_utilities` was rewritten from scratch with all of the module hacking removed.  Some of the functions that were there are not yet reimplemented.  Most of the other changes in core are to support this and work around pieces that are not yet implemented.
* A few small changes were made to support mounting a static directory outside of girder's python package.  Eventually we'll want to make the default path derived from `pkg_resources` and allow mounting a custom path via commandline options or configuration options.
* The jobs plugin was updated to work with the new conventions.   This is done in the third commit and is the most important part to take note of because similar changes will be necessary for all other plugins.


girder/constants.py
```diff
@@ -55,10 +55,10 @@
...
 
 #: The local directory containing the static content.
-#: Should contain ``clients/web/static``.
-STATIC_ROOT_DIR = ROOT_DIR
-if not os.path.exists(os.path.join(STATIC_ROOT_DIR, 'clients')):
-    STATIC_ROOT_DIR = PACKAGE_DIR
+STATIC_ROOT_DIR = os.environ.get(
+    'GIRDER_STATIC_ROOT',
+    os.path.join(ROOT_DIR, 'clients', 'web', 'static')
+)
```
jbeezley: This was done to validate that I can mount an external path for the static assets.  Eventually this will be built into configuration/cli options for both girder's server and `girder-install web`.

girder/utility/install.py
```diff
@@ -78,9 +78,10 @@ def _getPluginBuildArgs(buildAll, plugins):
     plugins = list(plugin_utilities.getToposortedPlugins(plugins, ignoreMissing=True))
 
     # include static-only dependencies that are not in the runtime load set
-    staticPlugins = plugin_utilities.getToposortedPlugins(
-        plugins, ignoreMissing=True, keys=('dependencies', 'staticWebDependencies'))
-    staticPlugins = [p for p in staticPlugins if p not in plugins]
+    # staticPlugins = plugin_utilities.getToposortedPlugins(
+    #     plugins, ignoreMissing=True, keys=('dependencies', 'staticWebDependencies'))
+    # staticPlugins = [p for p in staticPlugins if p not in plugins]
+    staticPlugins = []
```
jbeezley: I still haven't figured out how to handle this kind of dependency.  It is really only for building the client.  At the moment, the `plugin.json` file is still used by the build system, but it duplicates information added to the entry_point.  It will be nice to consolidate them, but I haven't yet decided how to propagate that information down to grunt.

girder/utility/model_importer.py
```diff
@@ -27,6 +27,14 @@
 # cache of instantiated models that have been lazy-loaded.
 _modelInstances = {}
 
+_pluginModelClasses = {}
+
+
+def girder_model(cls, name, plugin=None):
+    plugin_registry = _pluginModelClasses.setdefault(plugin, {})
+    plugin_registry[name] = cls
+    logger.info('Registered model %s in plugin %s' % (name, plugin))
```
jbeezley: This is just a temporary hack to register plugin models for `ModelImporter`.  Eventually, I want to create a metaclass or a class decorator to register models and make it consistent between core and plugin models.

girder/utility/plugin_utilities.py
```diff
@@ -17,462 +14,94 @@
...
-            pluginDir = os.path.dirname(module.__file__)
-            module.PLUGIN_ROOT_DIR = pluginDir
-            girder.plugins.__dict__[name] = module
-            isPluginDir = True
+    def __call__(self, func):
+        @six.wraps(func)
+        def wrapped(*arg, **kw):
+            return func(*arg, **kw)
+        wrapped.config = self.config
+        return wrapped
```
jbeezley: This config decorator was defined in https://github.com/girder/girder/pull/1255.  It is used to decorate the entry_point function and is intended to replace the `plugin.json` files.  I have left the original implementation for now.

girder/utility/plugin_utilities.py
```diff
@@ -17,462 +14,94 @@
...
+    config = getattr(mgr[plugin].plugin, 'config', {})
+    defaultConfig = {
+        'name': plugin,
+        'description': '',
+        'url': '',
+        'version': '',
+        'dependencies': [],
+        'staticWebDependencies': set()
+    }
+    return dict(defaultConfig, **config)
```
jbeezley: This just gets the configuration from a specific plugin using `stevedore`.  There are hooks you can add to add error handling and logging.

girder/utility/plugin_utilities.py
```diff
@@ -17,462 +14,94 @@
...
-            allPlugins[entry_point.name]['dependencies'])
+    mgr = NamedExtensionManager(
+        namespace=NAMESPACE,
+        names=plugins,
+        invoke_on_load=True,
+        invoke_args=(info,),
+        warn_on_missing_entrypoint=True,
+        verify_requirements=True
+    )
+    return mgr.names()
```
jbeezley: This uses `stevedore` to initialize all plugins from a list.  All that it does is loop through the names, find the associated entrypoint, and calls it with the given arguments.  Again, there are hooks we can add for error handling.

girder/utility/server.py
```diff
@@ -190,7 +187,7 @@ def loadRouteTable(reconcileRoutes=False):
     :returns: The non empty routes (as a dict of name -> route) to be mounted by CherryPy
               during Girder's setup phase.
     """
-    pluginWebroots = plugin_utilities.getPluginWebroots()
+    pluginWebroots = {}  # plugin_utilities.getPluginWebroots()
```
jbeezley: Plugin webroots is a feature not implemented yet.  It should only be a matter of adding it to the entrypoint decorator.

girder/utility/server.py
```diff
@@ -238,8 +235,7 @@ def setup(test=False, plugins=None, curConfig=None):
     cherrypy.tree.mount(None, routeTable[constants.GIRDER_STATIC_ROUTE_ID],
                         {'/':
                          {'tools.staticdir.on': True,
-                          'tools.staticdir.dir': os.path.join(constants.STATIC_ROOT_DIR,
-                                                              'clients/web/static'),
+                          'tools.staticdir.dir': constants.STATIC_ROOT_DIR,
```
jbeezley: This constant is now the *actual* static root so we don't have to append `clients/web/static` everywhere.  Because this is configurable, I think we will eventually want to move it to `config`.

plugins/jobs/girder_plugin_jobs/__init__.py
```diff
@@ -43,6 +44,12 @@ def scheduleLocal(event):
         fn(job)
 
 
+@plugin.config(
+    name='jobs',
+    description='A general purpose plugin for managing offline jobs.',
+    url='http://girder.readthedocs.io/en/latest/plugins.html#jobs',
+    version='2.0.0'
+)
```
jbeezley: This is the entry point decorator providing the information that used to come from `plugin.json`.

plugins/jobs/girder_plugin_jobs/__init__.py
```diff
@@ -19,8 +19,9 @@
 
 import importlib
 
-from girder import events
+from girder import events, plugin
 from . import constants, job_rest
+from .models import job as job_model  # noqa
```
jbeezley: Here we are importing the `job` model to "register" it with `ModelImporter`.  Alternatively, this could be done inside the `load` function below.

plugins/jobs/girder_plugin_jobs/models/job.py
```diff
@@ -24,7 +24,9 @@
 from girder import events
 from girder.constants import AccessType, SortDir
 from girder.models.model_base import AccessControlledModel, ValidationException
-from girder.plugins.jobs.constants import JobStatus, JOB_HANDLER_LOCAL
+from girder.utility.model_importer import girder_model
+
+from ..constants import JobStatus, JOB_HANDLER_LOCAL
```
jbeezley: `girder.plugins` was removed so we use a relative import, or alternatively an absolute import from the actual package name `girder_plugin_jobs.constants`.

plugins/jobs/girder_plugin_jobs/models/job.py
```diff
@@ -585,3 +587,6 @@ def listChildJobs(self, job):
                                                 user=user,
                                                 level=AccessType.READ):
             yield r
+
+
+girder_model(Job, 'job', 'jobs')
```
jbeezley: This is a hacky method to register a model.  This will be replaced with a cleaner implementation.

plugins/jobs/setup.py
```diff
@@ -0,0 +1,10 @@
+from setuptools import setup
+
+
+setup(
+    name='girder-plugin-jobs',
+    py_modules=['girder_plugin_jobs'],
+    entry_points={
+        'girder.plugin': 'jobs = girder_plugin_jobs:load'
+    }
+)
```
jbeezley: This is a minimal `setup.py` (actually the `py_modules` might not be necessary).  The key point that makes it a girder plugin is the `entry_points` line.  The entry point name is `girder.plugin` and the value string means the plugin name is `jobs` and the entrypoint function is in the module `girder_plugin_jobs` called `load`.  The naming is arbitrary, plugins can point the entry point to any function.

zachmullen:
Superseded by new work

Issue #2338, opened 2017-09-11
Allow new resource search modes to be added at runtime


brianhelba:
In addition to the existing `prefix` and `fulltext` search modes for the `resource/search` endpoint, plugins should be able to register their own search modes. The registration method should accept at least:
* A new mode identifier (to be used as the `mode` parameter to `resource/search`)
* A function reference, to a function which will be called with search parameters when a search occurs, and should return search results

PR #2363, opened 2017-09-26
Add a mode registry and improve search behavior 

Pierre-Assemat:
This PR is changing the way search mode are registered in order to allow external plugins
to add new mode of search by passing some parameters : 
- mode name
- set of allowed type to search on it
- a function to handle the new search  

Before like this:
`allowedSearchTypes = {'collection', 'folder', 'group', 'item', 'user'}`
The registry is now a dictionary stored in `constance.py` in the `class ResourceRegistry`.

The dict of the allowed search mode is like:
```
ALLOWED_SEARCH_MODE = {
        'mode': {
            'types': {'collection', 'folder', ...},
            'method': function
        }, ...
}
```

Fixes : #2338


girder/utility/search_mode_utilities.py
```diff
@@ -0,0 +1,102 @@
...
+def addSearchMode(mode, handler):
+    """This function is able to modify an existing search mode."""
+    _allowedSearchMode.update({
+        mode: handler
+    })
+
+
+def removeSearchMode(mode):
+    """Return False if the mode wasn't in the search mode registry."""
+    return _allowedSearchMode.pop(mode, False)
```
brianhelba: It'd be cleaner if this function always returned a bool (i.e. `True` if the handler was removed).

girder/utility/search_mode_utilities.py
```diff
@@ -0,0 +1,102 @@
...
+
+def getSearchModeHandler(mode):
+    if mode in _allowedSearchMode:
+        return _allowedSearchMode[mode]
+    return None
+
+
+def addSearchMode(mode, handler):
+    """This function is able to modify an existing search mode."""
+    _allowedSearchMode.update({
```
brianhelba: This can be a simple dict assignment.

girder/utility/search_mode_utilities.py
```diff
@@ -0,0 +1,102 @@
...
+    elif name == 'item':
+        return Item()
+    elif name == 'user':
+        return User()
+    else:
+        return None
+
+
+# Add dynamically the default search mode
+addSearchMode('text', 'textSearch')
```
brianhelba: This should pass a function as the handler, not a string.

girder/api/v1/resource.py
```diff
@@ -69,28 +68,17 @@ def search(self, q, mode, types, level, limit, offset):
         level = AccessType.validate(level)
         user = self.getCurrentUser()
 
-        if mode == 'text':
-            method = 'textSearch'
+        if self._isDefaultSearchMode(mode):
```
brianhelba: Up here, the goal is to have this just do:
```
searchHandler = search_mode_utilities.getSearchModeHandler(mode)
searchResults = searchHandler(
    query=q,
    types=types,
    user=user,
    level=level,
    limit=limit,
    offset=offset,
)
return searchResults
```

girder/utility/search_mode_utilities.py
```diff
@@ -0,0 +1,102 @@
...
+Plugins can modify this set to allow other search mode.
+By Default only two modes are allowed :
+  - text: to search in plain text
+  - prefix: to search with a prefix
+Their handlers are directly define in the base model.
+"""
+_allowedSearchMode = {}
+
+
+def defaultSearchModeHandler(q, mode, types, user, level, limit, offset):
```
brianhelba: If you want to keep some of this logic without having to duplicate it, you could always use [partial function application](https://docs.python.org/2/library/functools.html#functools.partial).

girder/utility/search_mode_utilities.py
```diff
@@ -0,0 +1,91 @@
...
+        if model is not None:
+            results[modelName] = [
+                model.filter(d, user) for d in getattr(model, method)(
+                    query=query, user=user, limit=limit, offset=offset, level=level)
+            ]
+    return results
+
+
+def getSearchModeHandler(mode):
+    if mode in _allowedSearchMode:
```
brianhelba: You could just use [the `.get` method](https://docs.python.org/2/library/stdtypes.html#dict.get), to make this a one-line implementation.

girder/utility/search_mode_utilities.py
```diff
@@ -0,0 +1,91 @@
...
+        return _allowedSearchMode[mode]
+    return None
+
+
+def addSearchMode(mode, handler):
+    """This function is able to modify an existing search mode."""
+    _allowedSearchMode[mode] = handler
```
brianhelba: Let's have this raise a `GirderException` if you try to overwrite a search mode that's already registered. This will be a nice safeguard against programming bugs.

girder/utility/search_mode_utilities.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def removeSearchMode(mode):
+    """Return a boolean to know if the mode was removed from the search mode registry or not."""
+    if _allowedSearchMode.pop(mode, None) is not None:
```
brianhelba: If you want to be really fancy, you *could* say:
```
return _allowedSearchMode.pop(mode, None) is not None
```

girder/utility/search_mode_utilities.py
```diff
@@ -0,0 +1,91 @@
...
+Plugins can modify this set to allow other search mode.
+By Default only two modes are allowed :
+  - text: to search in plain text
+  - prefix: to search with a prefix
+Their handlers are directly define in the base model.
+"""
+_allowedSearchMode = {}
+
+
+def defaultSearchModeHandler(query, mode, types, user, level, limit, offset):
```
brianhelba: Can you just move this function down the very bottom of this file, so it sits above the two calls to `addSearchMode`?

girder/utility/search_mode_utilities.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def removeSearchMode(mode):
+    """Return a boolean to know if the mode was removed from the search mode registry or not."""
+    if _allowedSearchMode.pop(mode, None) is not None:
+        return True
+    return False
+
+
+def _getModel(name):
```
brianhelba: Per feedback from @zachmullen , we're still allowed to use `ModelImporter` in these cases, so this whole function can go away.

girder/api/v1/resource.py
```diff
@@ -66,30 +65,23 @@ def __init__(self):
...
-                model = self.model(name, plugin)
-            else:
-                model = self.model(modelName)
-
-            results[modelName] = [
-                model.filter(d, user) for d in getattr(model, method)(
-                    query=q, user=user, limit=limit, offset=offset, level=level)
-            ]
-
+        if handler is not None:
```
brianhelba: If `handler` is None, I think we should raise a `RestException` (since the user-provided `mode` was invalid), rather than silently returning an empty dict.

girder/utility/search_mode_utilities.py
```diff
@@ -0,0 +1,91 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+###############################################################################
+from functools import partial
+
+from girder.models.assetstore import Assetstore
```
brianhelba: All these imports can go away, just to use `ModelImporter`.

girder/utility/search_mode_utilities.py
zachmullen: Correct. To elaborate, `ModelImporter` should be used if there is a need to dynamically map strings to model types. If the model type needed is known statically, pythonic imports should be used.

brianhelba:
@Pierre-Assemat Feel free to press the merge button yourself.

PR #2364, opened 2017-09-27
Fix the systemConfiguration template to fit in the style of the page

Pierre-Assemat:
The view missed some space between each label : 
screenshot from 2017-09-27 12-59-22

Adding the class `.form-goup` fix that : 
screenshot from 2017-09-27 12-59-47


Pierre-Assemat:
Thanks Zach ! I will merge that as soon as the tests pass

PR #2366, opened 2017-09-28
Ensure that the "fields" parameter to "Model.load" always works

brianhelba:
This implements the new utilities from #2352 across all `Model.load` methods.

Commits are separated logically:
* Ensure that the `fields` parameter to `Model.load` always works
* Logically reorder some statements in `item_test`
* Rename `_overwriteFields` and `_removeOverwrittenFields`

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -1324,7 +1324,6 @@ class AccessControlledModel(Model):
         :raises ValidationException: If an invalid ObjectId is passed.
         :returns: The matching document, or None if no match exists.
         """
-
         # Warn of str type deprecation for `fields` param
         if isinstance(fields, six.string_types):
             logger.warning('String data type for fields param is deprecated, \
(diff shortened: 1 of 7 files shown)


girder/models/model_base.py
```diff
@@ -628,10 +628,10 @@ def _overwriteFields(fields, overwrite):
         return copy
 
     @staticmethod
-    def _removeOverwrittenFields(doc, fields):
+    def _removeSupplementalFields(doc, fields):
         """
         Edit the document to be consistent with what the user originally requested, undoing what may
-            have been overwritten by _overwriteFields().
+            have been overwritten by _supplementFields().
```
zachmullen: dedent

brianhelba: @zachmullen Fixed.

zachmullen:
Couple of test failures on this merge: https://circleci.com/gh/girder/girder/3306

Any ideas?

brianhelba:
@zachmullen Checking it out now.

brianhelba:
@zachmullen I don't see these failures in http://my.cdash.org/index.php?project=girder

zachmullen:
Me neither, sometimes that cdash just fails to parse the results, or maybe CTest failed to upload them.

brianhelba:
@zachmullen I can't figure out why the terms test failed, but I'm honestly having trouble reproducing the exact series of logged REST requests that are in the failure log.

I'm a lot less familiar with the jobs test code, so I'd have to look into it in more detail to track down any issues. However, the jobs tests are rather tedious to reproduce by hand, since there's so much database state that gets set with the test. Let me know if we see this again, and I'll give it more attention.

PR #2369, opened 2017-09-28
Attempt to fix a race condition in the candela test

brianhelba:
Example of the failure: http://my.cdash.org/testDetails.php?test=35155877&build=1296108

### DIFF
--- a/plugins/candela/plugin_tests/candelaSpec.js
+++ b/plugins/candela/plugin_tests/candelaSpec.js
@@ -19,7 +19,7 @@ describe('Test the candela UI.', function () {
         }, 'Public and Private folders to appear');
 
         runs(function () {
-            $('a.g-folder-list-link:last').click();
+            $('a.g-folder-list-link:contains("Public")').click();
         });
         girderTest.waitForLoad();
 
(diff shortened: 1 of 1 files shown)


plugins/candela/plugin_tests/candelaSpec.js
```diff
@@ -104,19 +104,21 @@ describe('Test the candela UI.', function () {
...
+        }, 'menu to open');
 
         runs(function () {
-            expect($('#g-user-action-menu.open').length).toBe(1);
             $('a.g-my-folders').click();
         });
-        girderTest.waitForLoad();
+        waitsFor(function () {
+            return $('g-folder-list-link').length > 0;
+        }, 'user folder list to load');
```
zachmullen: Is this new `waitsFor` sufficient to fix the race condition, or are these other changes necessary?

brianhelba: I'm not sure what is causing the failure exactly, as I only have logs to go by. I'm just trying to make things more robust in general.

brianhelba:
@zachmullen Forgot a single character, so had to re-push.

PR #2371, opened 2017-09-28
Don't run CI on the "2.x-maintenance" branch, as it's a mirror of "master"

brianhelba:


brianhelba:
@girder/developers Can someone please approve this? Docs are here: https://circleci.com/docs/1.0/configuration/#branches

Otherwise, we're running two builds for every push to `master`.

PR #2375, opened 2017-10-02
Fixes #2374. Enable tools.proxy.on in beanstalk

### LINKED ISSUE #2374: Add tools.proxy.on = True to Beanstalk girder.cfg
This should be the default in a Beanstalk deploy

zachmullen:


dorukozturk:
@zachmullen Can you inform me a little bit about what problem this PR is solving and how do I reproduce the bug? Do I just deploy this and see if this is working?

zachmullen:
The OAuth module was not working in deployment because it was behind a proxy, and the callback URL was the local URL rather than the proxy URL.

zachmullen:
To test, go to the config page for the oauth plugin, and look at the **Authorized redirect URI** values; they should be the proxy URL, rather than the local URL.

PR #2376, opened 2017-10-03
Idiomatic model instantiation

zachmullen:
We discussed this as something on our wish list. This change supports that with minimal effort and is non-breaking. All the old ModelImporter semantics work just as before, and its behavior will remain since it does the job of converting strings to models (which is occasionally useful in REST endpoints). The main change is that that is no longer the *only* prescribed way of accessing the model layer, and in fact the new way should be preferred in every case except when there is a need to convert user-supplied strings to the appropriate model class.

**Guide for Reviewers**

* The first commit is the entirety of the core change. It's very small and is sufficient to enable the new semantics.
* The second commit just adds a test to verify the expected singleton behavior of model classes.
* All subsequent commits are refactoring of the codebase to use the new semantics rather than `ModelImporter`.


zachmullen:
WIP for now, if we approve of the general concept, I am happy to go through and convert core and plugin code to use the new semantics wherever possible.

jbeezley:
This works for the Girder 3 changes I'm making.  Specifically, I will have to modify the model importer logic to get the correct plugin module from the entry point rather than using `girder.plugins` import path.

This is a nice incremental change that I can support.  In the future, I would like to move away from the use of `importlib`.  This could be accomplished by creating a model registry to store the class instances.  I had in mind something like this:

```python
@register('job', plugin='jobs')
class Job(Model):
    pass
```

We could possibly sneak this into the metaclass transparently, but trying to detect the correct plugin name from inside the metaclass automatically seems ugly and brittle.  Something explicit like this might be a possibility:

```python
class Job(RegisteredModel('job', plugin='jobs')):
    pass
```
or
```python
class Job(Model):
    modelName = 'job'
    modelPlugin = 'jobs'
```

zachmullen:
> I would like to move away from the use of importlib

I also want to do that, and my intent was to do it on a second PR after this one. We can do a half measure that is just an implementation change of `ModelImporter` that only uses the guessing/importlib stuff as a fallback if there is no explicitly registered model for the given strings. Then for Girder 3, we can remove the fallback and require explicit registration if people want to use `ModelImporter` behavior.

jbeezley:
That works for me.

mgrauer:
@zachmullen Can you add that additional work to a new issue targeting the [Girder 3 milestone](https://github.com/girder/girder/issues?q=is%3Aopen+is%3Aissue+milestone%3A%223.0+release%220)?

zachmullen:
@mgrauer good call: https://github.com/girder/girder/issues/2377

zachmullen:
I have buy-in from @danlamanna and @jbeezley ... I'll move forward with the wide refactor unless anyone objects soon. @girder/developers

danlamanna:
Mind adding this to the changelog?

zachmullen:
This is now ready. Everything other than the first 2 commits and a4f0824 is just refactoring everything.

jbeezley:
I still see a few places where `ModelImporter` is being used with the model type known statically.
```
girder/utility/plugin_utilities.py:        setting = model_importer.ModelImporter().model('setting')
girder/utility/server.py:        settings = model_importer.ModelImporter().model('setting')
girder/utility/server.py:    setting = model_importer.ModelImporter().model('setting')
plugins/celery_jobs/server/__init__.py:    ModelImporter.model('user').load(doc['value'], force=True, exc=True)
plugins/celery_jobs/server/__init__.py:    userId = ModelImporter.model('setting').get(PluginSettings.CELERY_USER_ID)
plugins/celery_jobs/server/__init__.py:    user = ModelImporter.model('user').load(userId, force=True)
plugins/celery_jobs/server/__init__.py:    settingModel = ModelImporter.model('setting')
plugins/celery_jobs/server/__init__.py:        job = ModelImporter.model('job', 'jobs').save(job)

```

zachmullen:
Good catch, I was grepping for `ModelImporter.model('`, not `ModelImporter().model('`. As for the celery_jobs ones: I omitted those on purpose, mostly because that plugin is slated for summary destruction.

jbeezley:
Yeah, I wouldn't bother with celery_jobs.

manthey:
It isn't strictly related to this PR, but is there a reason we have the lines
```
global _modelInstances
```
in `girder/utility/model_importer.py`?  `_modelInstances` refers to a dictionary, which is accessible and modifiable to those routines without the `global` keyword.

Would using the `_modelInstances` dictionary be sufficient for the tests that use `_modelSingletons`?

zachmullen:
> Would using the _modelInstances dictionary be sufficient for the tests that use _modelSingletons?

`_modelInstances` is kept around for mapping strings to models. `_modelSingletons` is now the authoritative record of all models that have been instantiated; they may exist in `modelSingletons` but not in `_modelInstances`. They're both internal symbols and neither strictly needs to exist other than for testing where we drop the database between each test routine, so we can easily rework them in the future.

danlamanna:
It seems like this introduced an issue where importing resources now attempts to communicate with Mongo. For instance `from girder.api.v1.api_key import ApiKey` outputs "Connecting to mongodb..." when it didn't before.

zachmullen:
That's actually not from the model layer, but from the rest layer. Notice that if you do


    from girder.models.api_key import ApiKey

no database connection gets made.

danlamanna:
I see. It seems like it's coming from the changes to `filtermodel` and `modelParam`. Specifically, it looks like we're now instantiating models (connecting to the db) at declaration time rather than run-time.

zachmullen:
Good catch. We can defer the model instantiation to `__call__` rather than `__init__`.

girder/api/describe.py
```diff
@@ -259,15 +261,16 @@ def modelParam(self, name, description=None, model=None, destName=None, paramTyp
...
         item model parameter named ``item``, and ensure that the calling user
         has at least ``READ`` access on that item. For parameters passed in
         the query string or form data, for example a request like
         ``POST /item?folderId=...``, you must specify the ``paramType``.
 
-        >>> modelParam('folderId', 'The ID of the parent folder.',
+        >>> modelParam('folderId', 'The ID of the parent folder.', model=Folder,
         ...            level=AccessType.WRITE, paramType='query')
 
         Note that in the above example, ``model`` is omitted; in this case, the
```
msmolens: This PR added a `model` parameter to the above example, so this statement is confusing.

Issue #2378, opened 2017-10-04
Possible racing condition with createLinkFile


Xarthisius:
I get a weird errors when trying to create a lot of link files:

``` 
...
    register_path(endpoint, gc_folder, "%s/%s" % (path, obj['name']))
  File "/girder/plugins/wholetale/server/rest/harvester.py", line 104, in register_path
    register_path(endpoint, gc_folder, "%s/%s" % (path, obj['name']))
  File "/girder/plugins/wholetale/server/rest/harvester.py", line 102, in register_path
    mimeType='application/octet-stream', reuseExisting=True)
  File "/girder/girder/models/file.py", line 227, in createLinkFile
    file = self.save(file)
  File "/girder/girder/models/model_base.py", line 418, in save
    events.trigger('model.%s.save.created' % self.name, document)
  File "/girder/girder/events.py", line 291, in trigger
    handler['handler'](e)
  File "/girder/girder/models/file.py", line 326, in _propagateSizeToItem
    self.propagateSizeChange(item, fileDoc['size'])
  File "/girder/girder/models/file.py", line 253, in propagateSizeChange
    '_id': item['_id']
TypeError: 'NoneType' object is not subscriptable
```

code that triggers it looks roughly like that:

```
    def register_path(endpoint, parent, path, root=False):
        gc_folder = ... # some folder
        external_resource = ...  # some magical external file/folder structure
        for obj in external_resource:
           if obj['type'] == 'file':
                gc_item = itemModel.createItem(
                    obj['name'], user, gc_folder, reuseExisting=True)
                url = ...
                fileModel.createLinkFile(obj['name'], gc_item, 'item', url,
                    user, size=int(obj['size']),
                    mimeType='application/octet-stream', reuseExisting=True)
            else:
                register_path(endpoint, gc_folder, "%s/%s" % (path, obj['name']))
```

I believe it's a racing condition, because sometimes it's triggered, and sometimes not. Also happens at different stages of parsing `external_resource`. I'm using mongo replica set as a backend. Could that be possible source of non-atomic db writes?

zachmullen:
> I'm using mongo replica set as a backend. Could that be possible source of non-atomic db writes?

It could be... can you reproduce it on a local instance that is not using replicasets? If you change the write concern in the model layer, does it go away?

zachmullen:
In any case I think it's a bug that the `load` in `_propagateSizeToItem` does not have `exc=True`.

Xarthisius:
[...]

PR #2380, opened 2017-10-05
Write concern support for mongo client

zachmullen:
@Xarthisius here is a proposed fix for #2378 , could you test this and see if it works for you?


zachmullen:
Note that this also sets read preference as a config file option -- prior to this change, we were not setting the correct option for read preference.

danlamanna:
> Note that this also sets read preference as a config file option -- prior to this change, we were not setting the correct option for read preference.

I think maintaining a mapping between our own version of the mongo option name and the actual option name is prone to errors like these, perhaps we should just let configurations be a direct pass through to `pymongo.MongoClient` (with the exception of replica_set since it's already being supported). This would also allow support for all of the other Mongo options with no additional work (like the write concern timeout), though it looks like this may be supported via the URI already.

manthey:
All of the Mongo options CAN be encoded in the URI.  I think the main reason for exposing them in our settings is to make it obvious that it is something that should be considered (or, perhaps, we just need more documentation).

zachmullen:
> All of the Mongo options CAN be encoded in the URI.

This is true, but part of this is about setting sensible default values.

danlamanna:
> I think the main reason for exposing them in our settings is to make it obvious that it is something that should be considered (or, perhaps, we just need more documentation).

I think this is still accomplishable without changing the names of the options.

> This is true, but part of this is about setting sensible default values.

Maybe this can be solved without extensive documentation by putting them in `girder.dist.cfg` to begin with, in convention with the rest of Girder's default configuration values.

zachmullen:
@danlamanna had to resolve conflicts, could you approve again?

PR #2393, opened 2017-10-09
Ignore D106 and D107 for flake8 tests.

manthey:
Two new error codes were added in pydocstyle as of version 2.1.0.  These D106 and D107, and are variations on codes D101 and D102, which we are ignoring.  Because these are variations, we should also ignore D106 and D107.


manthey:
The PEP257 checker (pydocstring) released version 2.1.0 yesterday, which breaks our style tests.  See http://www.pydocstyle.org/en/latest/error_codes.html.

PR #2395, opened 2017-10-10
Allow endpoints to be mounted at a prefix

cjh1:
@zachmullen This is what I came up with. This is not a breaking change, I believe. 

Basically it allows a Resource instance to be passed into route(...). We then generate a mapping of Resource instance to mounted paths. We use this mounted path when generating the swagger doc. So you can mount a resource at a prefix like this:

When defining your routes you need to pass in self as the resource:
```python
self.route('POST', (), self.create_tomo, resource=self)
```

Then the resource can be mounted at a prefix like this:
```python
info['apiRoot'].test = Prefix()
info['apiRoot'].test.myresource =  Resource()
```
We could get more magical and remove the need for explicit creation of the prefix, but I'm not sure that is a good idea.

As a follow up to this I think we should add a breaking change in Girder 3 to make the Resource instance the default for the resource parameter [here](https://github.com/girder/girder/blob/master/girder/api/rest.py#L780). Then we can depreciate the use of resourceName ...

If this looks OK, will add a test or two.


zachmullen:
> When defining your routes you need to pass in self as the resource:

Can't we make `self` be the default for `resource`, since it's the first argument anyway?

cjh1:
@zachmullen We could but that would be a change in behavior/breaking change? Was planning on delaying that until Girder 3.

girder/api/rest.py
```diff
@@ -1193,3 +1193,9 @@ def wrapped(*args, **kwargs):
         return fun(ctx, *args, **kwargs)
 
     return wrapped
+
+class Prefix():
```
zachmullen: use new-style class

cjh1: Sure, I copied ApiDocs which was old school, wan't sure why?

cjh1:
@zachmullen Thanks for taking a look,  I will address those issues.

girder/api/rest.py
zachmullen: oversight :) feel free to fix that too

cjh1:
@zachmullen I think I have addressed your review comments. Here is am example of what it looks like in the swagger world:

screen

I am currently including the resource as well as the prefix in the tag, but there is an argument to used just the first part, in this can mdb. Thoughts?

docs/plugin-development.rst
```diff
@@ -239,6 +239,37 @@ classes, and we can add it to the API in the ``load()`` method.
     def load(info):
         info['apiRoot'].cat = Cat()
 
+Adding a prefix to an API
+*************************
+
+It is possible to provide a prefix to your API, allowing associate endpoints to
```
zachmullen: associated*

docs/plugin-development.rst
```diff
@@ -239,6 +239,37 @@ classes, and we can add it to the API in the ``load()`` method.
...
+
+.. code-block:: python
+
+    from girder.api.rest import Resource, Prefix
+
+    class Cat(Resource):
+        def __init__(self):
+            super(Cat, self).__init__()
+
+            self.route('GET', (), self.findCat, resource=self)
```
zachmullen: I don't fully understand this yet -- how does passing `resource=self` actually change the behavior of the endpoint, relative to passing `resource=None`?

cjh1: If None is passed as resource then a default is set [here](https://github.com/girder/girder/blob/master/girder/api/rest.py#L784-L785), the default is a string rather than the resource instance. We need the resource instance so we can find out where it has been mounted. Does that make sense?

zachmullen: I see. So making the default be `self` would *technically* be a breaking change, but practically, would it break anything?

cjh1: yes, probably not, in any case I think most plugins provide ```resourceName```, just being cautious.

zachmullen: Aren't we past the 2.x-maintenance / master deviation? Can we start making breaking changes on master now?

cjh1: I was hoping you could answer that :smile:

cjh1: Part of me was hoping I could also get this a release relatively soon, so I could use it.

zachmullen: That makes sense, but I don't want to propagate this usage that will be deprecated quite quickly in favor of omitting the argument... I'm tempted to change the default on this topic so that you can get this for free without having to pass an extra argument to every route. Am I correct in understanding that that would only break if you were already passing a `resource` parameter that was somehow different from where your resource was mounted?

cjh1: Correct. It is my understand that resource/resourceName is just used for the swagger doc, so currently we are relying on developer to mount things in the same place. So I think the only think we would break is people who have this mismatch in which can we are fixing them!?!?

cjh1:
Here is what it looks like with just the first part of the prefix as the tag:
screen

docs/plugin-development.rst
zachmullen: Right. So let's go ahead and make `self` the default value. This means we can also remove the `resourceName` attribute, right?

zachmullen:
I'm not sure where I stand on whether to display the prefix in the top level or not, what do you prefer? Swagger UI really only gives us two levels of nesting, AFAIK.

cjh1:
I prefer the second, I think its most useful, allowing your to group all endpoints associated with a plugin in a single place.

docs/plugin-development.rst
cjh1: @zachmullen and I spoke on hangout and we agreed to default to self ( this resource instance ) if ```resourceName``` is not provided and ```resource``` is ```None```. I will push this change.

zachmullen:
Works for me. It's just a presentation thing, it doesn't affect the path, so we can always change it later if we don't like it.

docs/plugin-development.rst
```diff
@@ -239,6 +239,38 @@ classes, and we can add it to the API in the ``load()`` method.
...
+            self.route('PUT', (':id',), self.updateCat)
+            self.route('DELETE', (':id',), self.deleteCat)
+
+        def getCat(self, id, params):
+            ...
+
+    def load(info):
+        info['apiRoot'].meow.cat = Cat()
+
+The endpoint are now mounted at meow/cat/
```
zachmullen: The endpoints* are now mounted at `meow/cat/`.

girder/api/rest.py
```diff
@@ -767,6 +767,9 @@ def route(self, method, route, handler, nodoc=False, resource=None):
 
         :type nodoc: bool
         :param resource: The name of the resource at the root of this route.
+            The resource instance (self) can also be passed. This allows the
+            mount path to be lookup. This allows a resource to be mounted at a
```
zachmullen: looked up*

docs/plugin-development.rst
```diff
@@ -239,6 +239,38 @@ classes, and we can add it to the API in the ``load()`` method.
...
+            self.route('GET', (':id',), self.getCat)
+            self.route('POST', (), self.createCat)
+            self.route('PUT', (':id',), self.updateCat)
+            self.route('DELETE', (':id',), self.deleteCat)
+
+        def getCat(self, id, params):
+            ...
+
+    def load(info):
+        info['apiRoot'].meow.cat = Cat()
```
zachmullen: Shouldn't the line `info['apiRoot'].meow = Prefix()` appear here?

cjh1: Yes, not sure what happened to it!

cjh1:
@zachmullen Are you still in favor of depreciating the use of ```resourceName```, I looked at making this change but it looked a little more involved than I thought.

zachmullen:
I don't feel strongly either way.

cjh1:
Ok, in which case I will leave it.

Issue #2396, opened 2017-10-10
Using the REST API to get the list of Folders and Items failed if we don't pass an Id or an text for searching


Pierre-Assemat:
Trying to get the list of folders and items by the REST API respond as bad request : 

![screenshot from 2017-10-10 13-11-03](https://user-images.githubusercontent.com/29961430/31400159-99f9a884-adbc-11e7-9072-ecdc750503fc.png)
 
The response is the same for the route : `http://localhost:8079/api/v1/item?limit=50&sort=lowerName&sortdir=1`

Do you have any thoughts  @girder/developers ?

zachmullen:
That endpoint requires you to pass either a folderId or text field. If you don't pass either, it doesn't know what to do. If you want, we can add a `notes` to the description that says as much.

Pierre-Assemat:
Yeah I just saw that in the function `find`

Pierre-Assemat:
I think it's clear in the code, maybe not enough in the REST API, I don't know

zachmullen:
Let's go ahead and add the `notes` field to the description.

PR #2398, opened 2017-10-10
Add new description notes to both item and folder in the REST API

Pierre-Assemat:
screenshot from 2017-10-10 13-44-12
---
screenshot from 2017-10-10 13-46-35

Fixes: #2396

### DIFF
--- a/girder/api/v1/folder.py
+++ b/girder/api/v1/folder.py
@@ -50,6 +50,10 @@ class Folder(Resource):
     @filtermodel(model='folder')
     @autoDescribeRoute(
         Description('Search for folders by certain properties.')
+        .notes('You must pass either a "folderId" or "text" field'
+               'to specify how you are searching for folders.  '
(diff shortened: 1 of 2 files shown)


girder/api/v1/folder.py
```diff
@@ -50,6 +50,9 @@ def __init__(self):
     @filtermodel(model='folder')
     @autoDescribeRoute(
         Description('Search for folders by certain properties.')
+        .notes('To have a valid request you need to pass either a folderId or a text field.  '
+               'If you omit one of these parameters the request will fail and respond : '
+               '"Invalid search mode."')
```
zachmullen: Let's make the sentence:

> You must pass either a "folderId" or "text" field to specify how you are searching for folders.

Pierre-Assemat: Only this sentence or can I keep the 2 last sentence too ?

PR #2399, opened 2017-10-10
Support simple async function calls

zachmullen:
Prior to this change, all async events went through the global event registry as a broadcast. This new syntax makes it easy to trigger anonymous async function calls, without the need for a broadcast. The call looks like:

    girder.events.daemon.trigger(info=..., callback=_myFn)

If you do not need to pass any info, the `info` kwarg can be omitted.

@opadron PTAL

### DIFF
--- a/girder/events.py
+++ b/girder/events.py
@@ -124,8 +124,12 @@ class ForegroundEventsDaemon(object):
     def stop(self):
         pass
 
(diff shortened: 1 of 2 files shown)


zachmullen:
This has the nice bonus of removing another of our side effects on module import.

PR #2296, opened 2017-08-18
Prevent Moment.js locale files from being included in the Webpack build

brianhelba:
Also, add Moment.js as an explicit dependency, as it's used directly.

This reduces the production build size by about 10%.

Note, while this excludes locale files from the built, it does not remove any localization functionality from Girder, as the non-default locales were never actually being enabled or used at runtime. If we wanted to support alternate date localizations in the future, we'd need to start including some files into the build (un-ignoring them), but we'd also need to globally configure Moment at run time to use them, according to some set of conditions.

PR #2404, opened 2017-10-11
Change how we exclude moment locales.

manthey:
Include *one* moment.js locale (English).  This is done in the webpack-approved way listed at https://webpack.js.org/plugins/context-replacement-plugin/.  It should cause less trouble for plugins that list moment as a requirement.

### DIFF
--- a/grunt_tasks/webpack.config.js
+++ b/grunt_tasks/webpack.config.js
@@ -68,9 +68,9 @@ module.exports = {
         // '__webpack_public_path__', since it's not always known at build-time.
     },
     plugins: [
-        // Exclude all of Moment.js's extra locale files (from where they're imported in either the
-        // ES6-source or UMD-built imports), to reduce build size.
-        new webpack.IgnorePlugin(/^\.\/locale$/, /moment(?:\/src\/lib\/locale)?$/),
+        // Exclude all of Moment.js's extra locale files except English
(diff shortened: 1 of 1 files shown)


manthey:
@brianhelba I think this is a preferred way to accomplish what was done in PR #2296, and it does so without breaking plugins that specify moment.js in their package.json files.

brianhelba:
I considered this approach, and there was a reason why I chose the `IgnorePlugin` over the `ContextReplacementPlugin`, but I'll have to review my work to remember what it was.

Can you provide an example of a plugin that directly depends on `moment.js`?

I'll review this PR regardless, as I agree it's important to be tolerant if plugins want to do reasonable things with `moment.js` on their own.

manthey:
There are two that I know of: https://github.com/Kitware/minerva and https://github.com/Kitware/vaui, both of which can be changed.  There may be others.

With the method that is currently in master, doing a `girder-install web --dev --all-plugins` will throw a console error for a missing locale (even if the plugin that is trying to use it isn't enabled).  With this change (and a clean build), this doesn't happen.

If there is a reason to `IgnorePlugin`, please document it.

brianhelba:
I'll have a full response within the next day.

zachmullen:
Any reason we shouldn't merge this?

manthey:
@brianhelba If you have something to comment on, please do so, otherwise, I'll merge this tomorrow.

PR #2406, opened 2017-10-13
Remove Bootstrap tooltip styling

brianhelba:
All tooltips-containing elements (i.e. those with a `title` attribute) will now use native browser tooltips.


brianhelba:
I've fixed the style errors, and actually fixed a few other bugs where widgets depended on Bootstrap popover but did not explicitly include it. @zachmullen PTAL again.

PR #2408, opened 2017-10-13
[WIP] CSS refresh

jeffbaumes:
Feedback welcome/appreciated. I've spent a little time the last few days doing a refresh of the Girder CSS. Some high-level changes:

* Using Open Sans instead of Droid Sans to update the look and feel. I am able to use the same build path as before to get the font from Google Fonts.
* Softer colors. The colors are Stylus variables so we could easily bump up the contrast if it's too soft.
* Use of flex styling. It could be used in many more places, but for now I've updated the title bar so that it does not need float left. The search box has moved from the left to the right side.
* Less busy interface. I've removed most instances of box shadows.
* Consistent top-level lists. Collections, Users, and Groups pages were not uniform in their styling - now they are much closer.
* The left panel is now fixed in position (unless in mobile responsive layout mode), so scrolling down will not make the menu items get lost.

image

Other than some minor things (font update in build, update Python/JS to support the custom title bar styles), all changes are in Stylus files only. The DOM looks exactly as it did before, so there are no incompatible things apart from if code/plugins assumed certain styling.

If folks think it is ok, I believe this could be merged without doing much more (it's already touching several files). There is plenty of opportunity for additional incremental updates to improve style and consistency.

This should likely wait to go in until we disconnect 2.x maintenance from master. This (or whatever this evolves into) could be billed as a modernized interface for 3.0.

### DIFF
--- a/clients/web/src/stylesheets/body/adminConsole.styl
+++ b/clients/web/src/stylesheets/body/adminConsole.styl
@@ -10,7 +10,8 @@ ul.g-admin-options
     text-transform uppercase
     font-weight 600
     display block
-    line-height 2.2
+    line-height 52px
+    vertical-align center
     font-size 14px
     i
       margin-right 16px
--- a/clients/web/src/stylesheets/body/adminConsole.styl
+++ b/clients/web/src/stylesheets/body/adminConsole.styl
@@ -1,20 +1,23 @@
+@import "../layout/layoutVars"
+
 ul.g-admin-options
-  list-style-type none
-  font-size 18px
   padding-left 0
   margin 0
 
   >li
+    list-style-type none
+    text-decoration none
+    text-transform uppercase
+    font-weight 600
     display block
-    padding 12px 10px
-    font-weight bold
-    color #666
-
-    >a
+    line-height 2.2
+    font-size 14px
+    i
+      margin-right 16px
+      font-size 24px
+    a
       color inherit
-      display block
 
   >li:hover
-    background-color #f4f4f4
+    color $baseColor
     cursor pointer
-    color #222
--- a/clients/web/src/stylesheets/body/collectionList.styl
+++ b/clients/web/src/stylesheets/body/collectionList.styl
@@ -1,22 +1,26 @@
+@import "../layout/layoutVars"
+
 .g-collection-list-entry
 
   .g-collection-title
-    font-size 17px
+    font-size $resourceTitleSize
 
   .g-collection-right-column
-    border-left 1px solid #ddd
     padding-left 8px
     display inline
-    width 290px
+    width 320px
     font-size 14px
     float right
-    color #909090
+    color $lightColor
 
   .g-collection-description
     background-color white
     padding 8px 12px
     border-radius 3px
-    border 1px solid #dedede
+    border $border
+
+  .g-collection-created-date, .g-user-space-used
+    color $normalColor
 
 .g-collection-list-header
   padding-bottom 10px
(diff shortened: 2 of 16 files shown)


Xarthisius:
I like the color/theme changes. One nitpick though: I use tiling wm and often work in two column setup. Looks like a lot space is wasted for left menu bar in the new version:
girder

Can it be slightly more thin?

manthey:
I would like to advocate higher-contrast.  At least to my perception, the top-level links look like they are intended to be disabled.

There are web-standards for contrast: https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html

jeffbaumes:
Round 2:

image

The sidebar was reduced from 310px to 230px.

Contrast was increased to have a minimum of 4.5 contrast as per the standard:
* Light text: [4.7](http://leaverou.github.io/contrast-ratio/#rgba%2851%2C51%2C51%2C.7%29-on-%23f5f5f5) 
* Normal text: [7.4](http://leaverou.github.io/contrast-ratio/#rgba%2851%2C51%2C51%2C0.85%29-on-%23f5f5f5)
* Emphasized text: [11.6](http://leaverou.github.io/contrast-ratio/#rgba%2851%2C51%2C51%2C1%29-on-%23f5f5f5)

Since there is no longer a big differential on emphasized and normal/light text, I changed the side menu to have the background go to white when selected.

Xarthisius:
Looks great to me!

Pierre-Assemat:
I like the idea of being more consistent with each page (collections, folder, users...). Really nice work!
The only things I wonder are : 

- the position of the search bar : now on the right, the tooltip for choose the search mode is hiding the current user and is really stuck on the right side of the page, maybe we could think if there are any other better disposition, I don't know... 
screenshot from 2017-10-16 14-18-27

- I found the global impression very clear, maybe too much, I don't really know if it's because I'm not use to it yet or if it's cause by the predominant white. I don't think the header bar should be white by default. 
Plus, when we change the color of the header bar, the text inside should adapt its color automatically, this was working before, but now the color became blue when the background is darker : 
screenshot from 2017-10-16 14-26-39
I think we need to change something to avoid this kind of case where the text become illegible because of few settings.

What do you think @jeffbaumes ?

mwoehlke-kitware:
More of an observation than a criticism, but has the metric difference between Droid Sans and Open Sans been considered? Open Sans is extremely similar to Droid Sans *except* metrically; it is non-trivially wider and has greater line height for the same point size, which will cause the same text to take up more space with the change. OTOH, Roboto (which is the successor to Droid Sans) is less aesthetically identical but has similar metrics.

Here's a comparison:

font-test
[(as PDF)](https://github.com/girder/girder/files/1431417/font-test.pdf)

jeffbaumes:
Thanks for the input @mwoehlke-kitware. Typefaces can spawn lots of discussion, and I don't claim Open Sans is the very best. I will say that to me, having something more different for a refresh is a plus for the reason that "it looks new", hence I'd prefer Open Sans to Roboto. It has seemed to not cause many layout issues so far in my experience.

mwoehlke-kitware:
Fair enough. I classified it as an "observation", since I'm not worried about making *a* change, but making an *unintended* change. So long as we're *aware* this change will result in different text metrics, that's okay.

(Of course, due to Droid Sans disappearing, we've gone and [pushed](https://github.com/girder/girder/pull/2451) Open Sans to 2.x, which somewhat undermines the "new look" argument :smile:.)

jeffbaumes:
@Pierre-Assemat I've pushed some changes to address the coloring of the user name (now is the same as the chosen contrasting color for the title) and now position the search box popovers below to avoid crowding.

Ready for another review.

zachmullen:
@jeffbaumes we marked this as WIP; if you are ready for a final review pass, please remove that tag.

jeffbaumes:
This will be made obsolete by a larger UI refresh.

PR #2412, opened 2017-10-16
pytest implementation

danlamanna:
This is a (buggy) proof of concept for how pytest/tox could look like in Girder 3. This works in a backwards compatible manner in that the user now has to run `ctest` as well as `pytest` to run all tests, but all of the existing test infrastructure remains in place.

A few notes:
- This adds the `girder._test` module, for various reasons I think we want the test module as part of Girder, particularly it would be nice for plugins to be able to import our utility functions without altering their `PYTHONPATH` at test time. What this means is that plugins will be able to run their tests entirely without CMake if they want. The entire module is underscored so we can avoid providing backwards compatibility of our fixtures/utilities until we figure out the correct level of abstraction/weed out any bad ideas.
- This adds a `pytest_girder` package (which is a pytest plugin), this will allow downstream to depend on the package and automatically be provided with Girder fixtures and CLI options for running tests.
- test_token.py is a good example of seeing how the current fixtures work, with some tests using none and others using many.
- Tests were converted quickly and don't take full advantage of pytest functionality (parametrization, marking, etc).


zachmullen:
Let's add documentation, including what fixtures are available from core.

zachmullen:
Let's also sort out how to load plugins for a test function.

danlamanna:
In trying to get as close to apples to apples as I can, the 5 or so
test modules I've copied over come in a few seconds quicker than CTest
(16s vs 20s). I expect however that pytest will end up taking longer
than CTest when all tests are converted simply because there will be
more of them (each function is a test rather than each class).

There are a few things to keep in mind:

- Tests that don't need every fixture will help combat this
  There are tests that don't need CherryPy or Mongo, they will be
  faster to start.
  
- Ensure Girder tests can be run back to back in the same process
  It takes 300-500ms just to import/bootstrap CherryPy/Girder, since
  each test is sandboxed in its own subprocess this adds considerable
  amounts of time. pytest allows us to run in the same process if we
  want, which revealed some flaws (#2379) that have been fixed/worked 
  around.

- Our hashing takes a long time
  For every user we create we're adding roughly 300ms to our
  test. While slow hashing is obviously intentional, our base test
  fixtures may want to hardcode the 'password' hash to shave a decent
  amount of time off our tests.

For example, running tests in the same process brought the pytest
version from 16 to 12 seconds, then precomputing hashes for the default
user accounts create brought it down to 8.

All of this being said, there's nothing noticeably slower about using
pytest. Any additional time it takes would be a result of creating
more tests which are further isolated.

pytest_girder/setup.py
```diff
@@ -0,0 +1,20 @@
...
+    ],
+    install_requires=[
+        'girder',
+        'pytest'
+    ],
+    entry_points={
+        'pytest11': [
+            'girder = pytest_girder.plugin'
+        ]
+    })
```
jbeezley: We'll almost certainly need to push this to pypi for external plugins to use.  It can wait until we work out some of the plugin refactoring, but we need to figure out a release workflow... Are the versions synchronized with girder versions?  If not, when do we push these releases?

cmake/run_circleci.sh
```diff
@@ -19,8 +19,15 @@ case $CIRCLE_NODE_INDEX in
...
-ctest -VV -S $HOME/girder/cmake/circle_continuous.cmake
+
+if [ "$TEST_GROUP" == "python" ]; then
+    if ! pytest --tb=long --junit-xml=$CIRCLE_TEST_REPORTS/pytest-${CIRCLE_NODE_INDEX}.xml; then
+        touch $HOME/build/test_failed
+    fi
+fi
+
+# pytest also generates coverage, so don't reset coverage at the start of the run
+ctest -E '^py_coverage_reset$' -VV -S $HOME/girder/cmake/circle_continuous.cmake
```
jbeezley: There is a part of me that just wants to declare testing bankruptcy, and only include coverage from the pytest framework.  I suppose we can wait until we have more coverage before doing it.

danlamanna:
> This is looking really nice.  I want to try to create some tests
> myself before approving.  Maybe I'll add them as a commit on this PR.

Yes, please do this. It would be very helpful for someone else to take a
stab at using this package and revealing any obvious missteps. (Use
unittest2pytest when doing this, to catch the low hanging fruit).

> My main concern with what is here is that the new tests are longer and
> more complex than I would prefer.  Was your intent to just create a
> one-to-one copy of the existing tests or were there test restart
> timing concerns going on?

The tests are mostly copy pasted, making only the necessary tweaks. We
can clean up tests as we go, this is more of matter of getting the base
usage of pytest within Girder worked out.

> This doesn't need to be a part of this PR, but I think it would make
> sense to collect many of these tooling configurations inside
> `setup.cfg`.  Most tools (except tox, notably) support this natively.

I would like this if it reduced the amount of configuration files we
have, rather than just renaming pytest.ini. Though, I think having
pytest.ini in the root of a project makes it obvious that running
`pytest` should "just work".

> We'll almost certainly need to push this to pypi for external plugins
> to use.  It can wait until we work out some of the plugin refactoring,
> but we need to figure out a release workflow... Are the versions
> synchronized with girder versions?  If not, when do we push these
> releases?

I'm not sure what the longstanding policy for it should be, but at least
for now it's pre-1.0 and has no compatibility guarantees. In the long
run, I can see this package getting very little churn after the initial
phase of downstream users (and plugins especially) start exercising it.

I addressed the rest of your code concerns with a new push.

jbeezley:
I passed on a few minor changes and ported over the base model tests.  It is all pretty mechanical.  I tried a more substantive test, and I ran into issues with needing more fixtures.  I think this is okay to be merged as is, but before we make a real push to port over all the tests and require people use the new framework, we need to provide more of these fixtures.  I'm not totally clear on how to move forward with what would have to be parametrized fixtures such as "file".  Have you put any thought into this?

jbeezley:
@girder/developers  I think we are nearing merging of this branch.  If anyone wants to chime in about it, now would be a good time.

mgrauer:
Can you update the changelog? 

Also, can you update the docs for any relevant changes? At least look through these sections

http://girder.readthedocs.io/en/latest/development.html#server-side-testing
http://girder.readthedocs.io/en/latest/development.html#client-side-testing
http://girder.readthedocs.io/en/latest/development.html#initializing-the-database-for-a-test
http://girder.readthedocs.io/en/latest/development.html#ansible-testing
http://girder.readthedocs.io/en/latest/developer-cookbook.html#adding-automated-tests
http://girder.readthedocs.io/en/latest/plugin-development.html#automated-testing-for-plugins

girder/test/test_access.py
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#  Copyright 2014 Kitware Inc.
```
jcfr: 2017

danlamanna: Are we allowed to just remove the date? If so I would do it across Girder so we don't have to have sporadic commits changing these values each year.

ronichoudhury:
According to my 30 seconds of Googling, the important year to include in a
copyright notice is the year of first publication. Adding a date range that
includes the current year is allowed but doesn't seem to add any actual
information. I'd say leave it at 2014 and don't worry about changing it
each year (or if we think that somehow dates the software, then make it say
`Copyright 2017-2017`).

http://www.copyrightlaws.com/copyright-notice-year/

manthey:
The actual U.S. law (from https://www.copyright.gov/title17/92chap4.html):

(b) Form of Notice.—If a notice appears on the copies, it shall consist of the following three elements:
(1) the symbol © (the letter C in a circle), or the word “Copyright”, or the abbreviation “Copr.”; and
(2) the year of first publication of the work; in the case of compilations or derivative works incorporating previously published material, the year date of first publication of the compilation or derivative work is sufficient. The year date may be omitted where a pictorial, graphic, or sculptural work, with accompanying text matter, if any, is reproduced in or on greeting cards, postcards, stationery, jewelry, dolls, toys, or any useful articles; and
(3) the name of the owner of copyright in the work, or an abbreviation by which the name can be recognized, or a generally known alternative designation of the owner.

jeffbaumes:
VTK [removed](https://github.com/Kitware/VTK/commit/af4841f566dc595051a00f23d6be2f44f51bbe15) the year from the copyright in their source code in 2003, opting to put the year in one place, (which does not seem to have been bumped since 2015):

https://github.com/Kitware/VTK/blob/master/Copyright.txt

Given the pain of updating the year, I would recommend following this and placing the full copyright with year in only one place in the repo.

pytest_girder/pytest_girder/utils.py
```diff
@@ -0,0 +1,143 @@
...
+
+    if not exception and response.output_status.startswith(b'500'):
+        raise AssertionError("Internal server error: %s" %
+                             getResponseBody(response))
+
+    return response
+
+
+def buildHeaders(headers, cookie, user, token, basicAuth, authHeader):
+    from girder.models.token import Token
```
danlamanna: @jbeezley I understood your commit moving imports into the fixture functions since fixtures.py is automatically imported, but shouldn't we be able to import this at the top level since test modules have to import it explicitly?

requirements-dev.txt
```diff
@@ -15,6 +15,13 @@ mccabe==0.6.1
 pep8==1.7.0
 pyflakes==1.5.0
 
+pytest==3.1.1
+pytest-xdist==1.18.1
+pytest-cov==2.5.1
+pytest-mongodb==2.1.1
```
danlamanna: I think items like `pytest-mongodb` need to be put into the `setup.py` of `pytest_girder`.

pytest_girder/pytest_girder/fixtures.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+    Mock out bcrypt password hashing to avoid unecessary testing bottlenecks.
+    """
+    with mock.patch('bcrypt.hashpw') as hashpw:
+        hashpw.side_effect = lambda x, y: x
+        yield hashpw
+
+
+@pytest.fixture
+def db(request, mongodb):
```
danlamanna: @jbeezley 
I think we should remove `pytest-mongodb` and "roll it ourselves" for the following reasons:
- We won't be using their fixtures since we have our own fixture system (which is higher level, folders, items, etc)
- It always drops databases before running, breaking the flag we added: `--drop-db={pre,post,never}`.
- It provides its own option for specifying the mongo URI, conflicting with our own

The logic I see it bringing to the table for us can be boiled down to this [8 line function](https://github.com/mdomke/pytest-mongodb/blob/develop/pytest_mongodb/plugin.py#L69).

pytest_girder/pytest_girder/utils.py
jbeezley: No, it is still necessary because the utils module is imported by the fixtures module.

pytest_girder/pytest_girder/fixtures.py
jbeezley: I rewrote the mongo mocking part to use mongomock directly.  You're right, it's much simpler now.

zachmullen:
Woohoo!

PR #2433, opened 2017-10-17
girder_client: Add support for --ssl-verify/--no-ssl-verify

jcfr:


zachmullen:
@danlamanna I believe you had objections to passing all these low level options through the constructor. Do you think this should just be something we do in the CLI layer instead?

jcfr:
My 0.02, I think re-implementing your own client just for the sake of disabling SSL verification would be counter productive for most user in the situation described by Matt ... 

And if there is strong objection, similarly to HTTPS_PROXY .. we could instead support an environment variable ?

docs/python-client.rst
```diff
@@ -95,6 +95,13 @@ The client also supports ``username`` and ``password`` args. If only the
 ``username`` is specified, the client will prompt the user to interactively
 input their password.
 
+Disabling SSL verification
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Specifying ``--no-ssl-verify`` allows to ignore SSL verification. This is
+usually required when using client behind a proxy that is not configured to
```
mgrauer: when using the client

mgrauer:
After learning recently that Bazel doesn't support ENV Vars, I'd like to have an alternative path, i.e. not just an ENV Var only path.

zachmullen:
I don't have an objection to the command line argument, but I might object to having it be a parameter to the GirderClient constructor, and I think @danlamanna would object more strongly to that. Rather, we could just set the verify flag to false on the GirderClient that is instantiated in cli.py.

zachmullen:
I like this solution better, however the `verify` option to the request session can take values besides just true and false, such as a string pointing to a cert bundle to use. Would it be possible to support all of its forms through the command line argument?

danlamanna:
> I don't have an objection to the command line argument, but I might object to having it be a parameter to the GirderClient constructor, and I think @danlamanna would object more strongly to that.

Agreed, however we're extending `GirderCli` and not `GirderClient` here so I'm slightly less concerned. In the ideal world `GirderClient` would always have a session object present that we could manipulate, that way this flag would just require setting `ctx.obj.session.verify = False`. For now this solution LGTM, though I think the previously stated solution should be a goal of Girder 3.

Speaking further about our API surface, I don't think `cli.py` should be a part of it. Our compatibility should be in terms of the actual commands users of `girder-cli ...` type and in the library API presented in `girder_client/__init__.py` but not in the functions/decorators detailing the intricacies of how we make our CLI so.

jcfr:
> parameter to the GirderClient constructor,

To be accurate, it is a parameter of the `GirderCli` ctor and not the `GirderClient`. This is required because there is a call to authenticate in the constructor itself and we need to configure the session before it happens.

> Would it be possible to support all of its forms through the command line argument?

Yes, we should be able to implement such an interface

>   API surface ... not in the functions/decorators detailing the intricacies of how we make our CLI

👍

zachmullen:
> Speaking further about our API surface, I don't think cli.py should be a part of it. 

Agreed. Sorry for the noise, I guess I misread this at first as being in the GirderClient constructor.

jcfr:
Updated PR address the last comment of Mike.

> re: test 

Should we add a test for the `--no-ssl-verify` case ?  If yes, I do not have the bandwidth this afternoon to work on this ... later this week or the next one.

mgrauer:
@danlamanna you said this LGTY, is that an approval? Do you require a test or can that be postponed to a later issue?

zachmullen:
Are we going to address the non-boolean forms of `verify` with separate CLI arguments, or try to refactor this one to support them?

clients/python/girder_client/cli.py
```diff
@@ -154,16 +155,21 @@ def format_options(self, ctx, formatter):
                    '[default: %s]' % GirderClient.DEFAULT_API_ROOT,
               show_default=True,
               cls=_AdvancedOption)
-@click.option('--ssl-verify/--no-ssl-verify', default=True,
-              help='Enable or disable SSL Verification',
+@click.option('--no-ssl-verify', is_flag=True, default=False,
```
jcfr: It is usually better to avoid double negation. That is why having the variable named `ssl_verify` instead of `no_ssl_verify`

clients/python/girder_client/cli.py
```diff
@@ -187,9 +193,14 @@ def main(ctx, username, password,
                 'Option "--api-url" and option "--%s" are mutually exclusive.' %
                 name.replace("_", "-"))
 
+    if certfile:
+        verify = certfile
+    else:
+        verify = not no_ssl_verify
```
jcfr: Here is an example of double negation to avoid ..

clients/python/girder_client/cli.py
danlamanna: Would it be simpler to just have a flag named `--insecure`, similar to curl?

jcfr: > the positive is the default and only the negative needs to be specified.

This is why the following was working:

```
@click.option('--ssl-verify/--no-ssl-verify', default=True,
              help='Enable or disable SSL Verification',
```

Default: ssl-verify is True

Then, if --no-ssl-verify is passed .. it is False.

I suggest you keep the previous flag, and simply add the certfile support.

docs/python-client.rst
```diff
@@ -95,6 +95,13 @@ The client also supports ``username`` and ``password`` args. If only the
 ``username`` is specified, the client will prompt the user to interactively
 input their password.
 
+Disabling SSL verification
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Specifying ``--no-ssl-verify`` allows to ignore SSL verification. This is
+usually required when using the client behind a proxy that is not configured to
+accept the certificate of the given host.
+
```
jcfr: Very nice. Could you also add an entry like:

```
Specifying a custom SSL certificate
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Specifying ``--certfile /path/to/custom_bundle.pem`` allows to use a custom "bundle" of Certificate Authority (CA)
public keys (CA certs) for performing the SSL verification applied when the ``https`` scheme is associated with
the API url.

By default, the carefully curated collection of Root Certificates from Mozilla is used. 
See https://pypi.python.org/pypi/certifi
```

jcfr:
Well done 👍 

One last nitpick associated with the documentation and we should be all set.

jcfr:
👍  LGTM

Being the original author of the PR, the user interface doesn't let me approve the PR.

jcfr:
I also just tested this new case:

````
girder-cli --insecure --certfile /home/jcfr/Projects/Slicer/Base/QTCore/Resources/Certs/Slicer.crt --host data.kitware.com download 59e6294b8d777f31ac647633
Usage: girder-cli [OPTIONS] COMMAND [ARGS]...

Error: Option "--insecure" and option "--certfile" are mutually exclusive.
```

sutartmelson:
@danlamanna or @zachmullen  Could you give this a review?

zachmullen:
I like the name `--no-ssl-verify` better than `--insecure`, since the latter doesn't really give an indication of what it is doing.

sutartmelson:
As @danlamanna  suggested, `--insecure` follows how [curl](https://curl.haxx.se/docs/manpage.html#-k) does it so maybe it's intuitive to some users.

zachmullen:
wget uses `--no-check-certificate` and `--certificate`

sutartmelson:
`--no-ssl-verify` it is

clients/python/girder_client/cli.py
```diff
@@ -146,9 +155,21 @@ def format_options(self, ctx, formatter):
                    '[default: %s]' % GirderClient.DEFAULT_API_ROOT,
               show_default=True,
               cls=_AdvancedOption)
+@click.option('--no-ssl-verify', is_flag=True, default=False,
+              help='Disable SSL Verification',
+              show_default=True,
+              cls=_AdvancedOption
+              )
+@click.option('--certfile', default=None,
```
zachmullen: One other nit, could we change this to `--certificate`? The underlying API supports passing a file *or* a directory containing certificates.

PR #2435, opened 2017-10-17
[WIP] Add methods to serialize python client instances

jbeezley:
@kotfic and I were speaking this morning about the need for a "serialization" of a girder client instances for transmission to celery tasks.  This is done currently in a number of places by just storing the API url and authentication token.  This PR is an RFC of sorts that introduces a concrete API to do this instead.  Along with the serialization methods on the class there is a json encoder class and object_hook for integration with the standard json module.

For the moment, this only captures the essential attributes of the girder client object as its state.  There might be other attributes we want to collect depending on how this is used.  `incomingMetadata` and `localMetadata` seem like an in memory cache.  There is also the [incoming](https://github.com/girder/girder/pull/2433) `sslVerify` attribute.  I could also be convinced that none of this is a good idea and it should be up applications to know what information is needed.


kotfic:
I think i am a big fan of this idea as it allows for much more explicit management of girder client instantiation in the worker and hides details of the api_url/token implementation behind the girder_client object.

This seems to rely on a 'magic key'  (e.g. ```__girder_client__```) being present in the dict for de-serialization.  Given what's taking place in [jbeezley/girder_worker_utils](https://github.com/jbeezley/girder_worker_utils) with transformation objects,  does it make sense to standardize this magic key behavior?

jbeezley:
I'm tagging this a WIP for now.  To clarify @kotfic's comment, I will add some sort of class serialization registry mechanism in girder_worker_utils that this will use so that a single object hook function can handle the deserialization generically.

clients/python/girder_client/__init__.py
```diff
@@ -1716,3 +1719,61 @@ def _checkResourcePath(self, objId):
...
+
+    >>> serialized = json.dumps(GirderClient(), cls=JSONEncoder)
+    """
+    def default(self, o):
+        if isinstance(o, GirderClient):
+            return o.toDict()
+        return super(JSONEncoder, self).default(o)
+
+
+def object_hook(data):
```
zachmullen: camelCase

jbeezley:
There is a general mechanism for class deserialization in girder_worker_utils.  I'll revisit this once we depend on it in girder.

PR #2437, opened 2017-10-19
Mention discourse site in our README

zachmullen:


mgrauer:
@brianhelba Once you add the announcements topic in discourse, could you update this to point there and give a mention of it?

PR #2438, opened 2017-10-20
API key control setting

zachmullen:
This is a new system setting that can be used to disable API key functionality on an instance. This is a security requirement for some of our users.

I'd like to disable the UI at some point too, but that will be easier if we can have settings with a public flag, which is likely a 3.0 change.


manthey:
This disable the restful endpoint for generating api keys.  Do we need to disable them in the model layer, too?

zachmullen:
> Do we need to disable them in the model layer, too?

Don't think so, this is just to disallow end users to use API keys.

brianhelba:
@zachmullen Can you elaborate on how disabling API keys improves security?

zachmullen:
> Can you elaborate on how disabling API keys improves security?

This will be used on an instance where users are only allowed to login via a specific OpenID provider that enforces a two-factor authentication policy. The sysadmins for this instance would not approve of API-key style authentication since it does not use two-factor authentication.

girder/api/v1/api_key.py
```diff
@@ -18,10 +18,11 @@
 ###############################################################################
 
 from ..describe import Description, autoDescribeRoute
-from ..rest import Resource, filtermodel
+from ..rest import Resource, RestException, filtermodel
```
manthey: As of PR #2498, we're importing exceptions from `girder.exceptions`.

Issue #2409, opened 2017-10-13
Cache config options for the girder-cli


mgrauer:
@jamiesnape has need of caching config options for the girder-cli without using ENV vars, as those aren't usable from within Bazel, config options such as the API key and URL.

Perhaps saying `girder-cli configure` and then it prompts to enter a URL and key and then stores that somewhere for use by future calls.

There was some movement towards this using config files, that then stalled and was going to be done by click, but I don't think there has been any progress since then.

See #1850

PR #2442, opened 2017-10-25
[WNIP] Cache cli config options in config file

sutartmelson:
Addressees #2409 

The config file looks something like this
```
[CLI]
api_url = http://localhost:9080/api/v1
api_key =
username = girder
password = girder
host =
scheme =
port =
api_root =
```

### DIFF
--- a/clients/python/girder_client/cli.py
+++ b/clients/python/girder_client/cli.py
@@ -126,30 +126,46 @@ class _Group(click.Group):
 
 def _parseConfigFile(path):
     config = six.moves.configparser.ConfigParser(allow_no_value=True)
-    params = ['username', 'password', 'api_key', 'api_url', 'scheme', 'host', 'port', 'api_root']
+    params = ['username', 'password', 'api_key', 'api_url', 'scheme', 'host', 'port', 'api_root',
+              'no_ssl_verify', 'certificate']
     data = {}
-    with open(path, 'r') as configFile:
-        config.readfp(configFile)
-        for key in params:
-            value = config.get('CLI', key)
-            if value:
-                data[key] = value
-    return data
+    try:
+        with open(path, 'r') as configFile:
+            config.readfp(configFile)
+            for key in params:
+                try:
+                    value = config.get('CLI', key)
+                    if value:
+                        data[key] = value
+                except six.moves.configparser.NoOptionError:
+                    continue
+        return data
+    except six.moves.configparser.NoSectionError:
+        return []
 
 
-def _loadConfigPaths():
-    paths = []
-    if 'GIRDER_CLIENT_CONFIG' in os.environ:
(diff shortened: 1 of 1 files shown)


clients/python/girder_client/cli.py
```diff
@@ -163,6 +177,16 @@ def main(ctx, username, password, api_key, api_url, scheme, host, port, api_root
...
+    if config:
+        ctx.params.update(_parseConfigFile(config))
+        username = ctx.params.get('username')
+        password = ctx.params.get('password')
+        api_key = ctx.params.get('api_key')
+        api_url = ctx.params.get('api_url')
+        scheme = ctx.params.get('scheme')
+        host = ctx.params.get('host')
+        port = ctx.params.get('port')
+        api_root = ctx.params.get('api_root')
```
zachmullen: If someone passes `--config` but also passes one of these options explicitly, shouldn't the explicit option override the config file one?

sutartmelson: That sounds like useful behavior for the user. I'll do that.

clients/python/girder_client/cli.py
```diff
@@ -113,6 +114,15 @@ def format_options(self, ctx, formatter):
         self.format_commands(ctx, formatter)
 
 
+def _parseConfigFile(path):
```
danlamanna: Our existing configuration is ini-style files so we should probably remain consistent with that.

sutartmelson: Should this be added to an existing config file somewhere or pass in the path of a cli specific config file?

danlamanna: @zachmullen can chime in here, but I think we should probably copy the configuration path precendence set here: https://github.com/girder/girder/blob/61aa5833e91b43d1ea03fa9506175089/girder/utility/config.py#L43-L50 (with a differently named configuration file).

zachmullen: 👍  

1. Value of `--config` option if passed.
2. Value of `GIRDER_CLIENT_CONFIG` env var if set.
3. ~/.girder/girder_client.cfg
4. /etc/girder_client.cfg

clients/python/girder_client/cli.py
```diff
@@ -177,21 +197,29 @@ def main(ctx, username, password, api_key, api_url, scheme, host, port, api_root
     input his/her password.
     """
     # --api-url and URL by part arguments are mutually exclusive
+    if not config:
+        paths = _loadConfigPaths()
+        if paths:
+            config = paths[0]
```
zachmullen: I'm not sure whether the behavior here should be to only look at one config file, or just list them all in order of precedence and have ConfigParser attempt to read all of them. I'm leaning toward the latter, which would eliminate a lot of this custom logic.

jcfr:
I suggest to also add tests.

PR #2443, opened 2017-10-25
Refactor DICOM viewer web client

brianhelba:


brianhelba:
@danlamanna Updated to fix the small issues we discussed in-person.

PR #2444, opened 2017-10-25
Fix issue with having two endpoint mount under prefix

cjh1:


girder/utility/resource.py
```diff
@@ -11,7 +11,6 @@ def _walkTree(node, path=[]):
             full_path = list(path)
             full_path.append(k)
             routeMap[v] = full_path
-            path = []
```
zachmullen: Just noticed this is a mutable default var... maybe we should change it from `[]` to `()`?

cjh1: Good plan, done

PR #2449, opened 2017-10-29
Allow the Ansible module to run in Python3

brianhelba:


brianhelba:
I don't know whether this is the only change necessary, but this should at least allow the static analysis tests to pass in Python 3.

PR #2450, opened 2017-10-30
Search in dicom metadata

Pierre-Assemat:
The `dicom_viewer` plugin add a new search modes into the search mode registry : 

- Substring search in both keys and values

Fixes: #2313

### DIFF
--- a/clients/web/src/templates/widgets/searchModeSelect.pug
+++ b/clients/web/src/templates/widgets/searchModeSelect.pug
@@ -6,4 +6,4 @@ for mode in modes
       input.g-search-mode-radio(
           type="radio", name="search-mode", value=mode,
           checked=(mode === currentMode ? 'checked' : undefined))
-      = descriptionModes(mode)
+      = getModeDescription(mode)
--- a/clients/web/src/templates/widgets/searchModeSelect.pug
+++ b/clients/web/src/templates/widgets/searchModeSelect.pug
@@ -6,7 +6,4 @@ for mode in modes
       input.g-search-mode-radio(
           type="radio", name="search-mode", value=mode,
           checked=(mode === currentMode ? 'checked' : undefined))
-      if mode === 'prefix'
-        |  Search by prefix
-      else if mode === 'text'
-        |  Full text search
+      = descriptionModes(mode)
--- a/clients/web/src/views/widgets/SearchFieldWidget.js
+++ b/clients/web/src/views/widgets/SearchFieldWidget.js
@@ -84,7 +84,7 @@ var SearchFieldWidget = View.extend({
         this.placeholder = settings.placeholder || 'Search...';
         this.getInfoCallback = settings.getInfoCallback || null;
         this.types = settings.types || [];
-        this.modes = settings.modes || ['text', 'prefix'];
+        this.modes = settings.modes || SearchFieldWidget.getListMode();
 
         if (!_.isArray(this.modes)) {
             this.modes = [this.modes];
@@ -135,7 +135,8 @@ var SearchFieldWidget = View.extend({
             },
             content: _.bind(function () {
                 return SearchHelpTemplate({
-                    mode: this.currentMode
+                    mode: this.currentMode,
+                    descriptionHelp: SearchFieldWidget.getHelpDescription(this.currentMode)
                 });
             }, this)
         }).click(function () {
@@ -152,7 +153,8 @@ var SearchFieldWidget = View.extend({
             content: _.bind(function () {
                 return SearchModeSelectTemplate({
                     modes: this.modes,
-                    currentMode: this.currentMode
+                    currentMode: this.currentMode,
+                    descriptionModes: SearchFieldWidget.getModeDescription
                 });
             }, this)
         }).click(function () {
@@ -195,7 +197,7 @@ var SearchFieldWidget = View.extend({
             data: {
                 q: q,
                 mode: this.currentMode,
-                types: JSON.stringify(this.types)
+                types: JSON.stringify(SearchFieldWidget.getListType(this.types, this.currentMode))
             }
         }).done(_.bind(function (results) {
             this.ajaxLock = false;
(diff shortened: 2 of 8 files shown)


jcfr:
@fedorov  Once integrated, this will add DICOM metadata search. 

Cc: @mgrauer

plugins/dicom_viewer/server/__init__.py
```diff
@@ -195,9 +196,44 @@ def _uploadHandler(event):
...
+    # The insensitive Case is not perfect
+    # We will need to search into object value as improvment.
+    jsQuery = """
+        return Object.keys(obj.dicom.meta).some(
+            function(key) {
+                return (key.toLowerCase().indexOf('%(key)s'.toLowerCase()) !== -1)  ||
+                    obj.dicom.meta[key].toString().toLowerCase()
+                        .indexOf('%(value)s'.toLowerCase()) !== -1;
+            })
+        """ % {'key': query, 'value': query}
```
brianhelba: Right now, there's no need for the `return`, since `$where` can accept a simple JavaScript expression.

However, I think you should wrap this whole `jsQuery` into a JS function, so we can define some intermediate variables for performance. Specifically, I think you should define variables for at least:
* `'%(key)s'.toLowerCase()`
* `'%(value)s'.toLowerCase()`
* `obj.dicom.meta`

brianhelba:
Looks really good so far. Comments inline, mostly changes to naming and user-facing English strings.

clients/web/src/templates/widgets/searchModeSelect.pug
```diff
@@ -6,7 +6,4 @@ for mode in modes
       input.g-search-mode-radio(
           type="radio", name="search-mode", value=mode,
           checked=(mode === currentMode ? 'checked' : undefined))
-      if mode === 'prefix'
-        |  Search by prefix
-      else if mode === 'text'
-        |  Full text search
+      = descriptionModes(mode)
```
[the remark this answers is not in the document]
Pierre-Assemat: I'm not sure it's possible, take a look at [this issue](https://stackoverflow.com/questions/38295332/pug-jade-input-is-a-self-closing-element-input-but-contains-nested-conten)

plugins/dicom_viewer/server/__init__.py
```diff
@@ -195,9 +197,46 @@ def _uploadHandler(event):
...
+    Provide a substring search on both keys and values.
+    """
+    if types != ['item']:
+        raise RestException('The dicom search is only able to search in Item.')
+
+    # The insensitive Case is not perfect
+    # We will need to search into object value as improvment.
+    jsQuery = """
+        function() {
+            const key = '%(key)s'.toLowerCase();
```
brianhelba: `const` is ES6 syntax. For compatibility with older versions of MongoDB, we should only use ES5 syntax, so this should be `var`.

brianhelba: Also, can you name this `queryKey` and `queryValue`? In addition to being clearer, this will fix a bug in the current implementation.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -195,9 +197,46 @@ def _uploadHandler(event):
...
+            const value = '%(value)s'.toLowerCase();
+            const meta = obj.dicom.meta;
```
brianhelba: Can you name this `dicomMeta`?

plugins/dicom_viewer/server/__init__.py
```diff
@@ -195,9 +197,46 @@ def _uploadHandler(event):
...
+                    return (key.toLowerCase().indexOf(key) !== -1)  ||
+                        meta[key].toString().toLowerCase().indexOf(value) !== -1;
+                })
+            }
+        """ % {'key': query, 'value': query}
+
+    # Sort the documents inside MongoDB
+    cursor = Item().find({'dicom': {'$exists': True}, '$where': jsQuery})
+    # Filter the result
+    result = Item().filterResultsByPermission(cursor, user, level, limit, offset)
```
brianhelba: You can inline this `filterResultsByPermission` into the list comprehension below.

clients/web/src/views/widgets/SearchFieldWidget.js
```diff
@@ -253,6 +258,57 @@ var SearchFieldWidget = View.extend({
             }
         }, this));
     }
+}, {
+    _allowedSearchMode: {},
+
+    addMode: function (mode, types, description, help) {
+        if (!SearchFieldWidget._allowedSearchMode[mode]) {
```
brianhelba: This is minor, but in my opinion, it's probably clearer style to check whether the `mode` already exists [using `_.has`](http://underscorejs.org/#has), and if so, throw the error within the body of the `if`. Then, do the normal work of registering the mode in the body of the function (not inside an `if` or `else`).

I think this is preferable because it tests any preconditions at the start of the function call, then moves into the function logic with the assumption that they're true. This allows all the logic to be at the top (non-indented) level of the function, and keeps the error handling (throw an exception) close to the error check (the condition).

clients/web/src/templates/widgets/searchModeSelect.pug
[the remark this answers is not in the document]
brianhelba: Ah, I see, I thought that the content was within the `input` as text, and the style was misleading, but I forgot `input` was self-closing. So, this is just plain text that's being injected.

Can you refactor this code so that the text is directly inside a normal `label`, and the parent element is a `div`? It doesn't have to be exactly this, but the current hierarchy doesn't make much sense when compiled.

plugins/dicom_viewer/server/__init__.py
Pierre-Assemat: What do you mean by inline this ?

brianhelba: Combine this line and the next.

Pierre-Assemat: But `filterResultsByPermission` take a cursor, and `filter` take only one element, how can I compatc them into one line ?
Something like: 
```
result = {
        'item': Item().filterResultsByPermission([Item().filter(d, user) for d in result], user, level ...)
}
```
is ok ?

zachmullen: `'item': [Item().filter(i, user) for i in Item().filterResultsByPermission(...)]`

Pierre-Assemat: mmh yes it make sens! Thanks!

Pierre-Assemat: In order to fit in the 100 characters limitation what is the best : 
```
result = {
        'item': [Item().filter(d, user) 
            for d in Item().filterResultsByPermission(cursor, user, level, limit, offset)]
    }
``` 
or 
```
    result = {
        'item': [Item().filter(d, user) for d in Item().filterResultsByPermission(
            cursor, user, level, limit, offset)]
    }
```
or again
```
    result = {
        'item': [
            Item().filter(d, user)
            for d in Item().filterResultsByPermission(cursor, user, level, limit, offset)
        ]
    }

```

zachmullen: I personally like the third one, but the second one is OK too.

Pierre-Assemat: Me too, but is this easily readable ?

zachmullen: I think it is, it's a fairly common way to break up a list comprehension. If you don't think it's readable enough, you could always make a variable called `cursor` and use that in the comprehension instead of inlining the `filterResultsByPermission` call.

Pierre-Assemat: That's good to me, thank you Zach!

mgrauer:
Great stuff, way to go guys.

clients/web/src/views/widgets/SearchFieldWidget.js
```diff
@@ -195,7 +197,10 @@ var SearchFieldWidget = View.extend({
             data: {
                 q: q,
                 mode: this.currentMode,
-                types: JSON.stringify(this.types)
+                types: JSON.stringify(_.intersection(
+                    this.types,
+                    SearchFieldWidget.getModeTypes(this.currentMode))
+                )
```
msmolens: Performing an intersection here breaks search results when `SearchFieldWidget` is initialized with a custom type, because `types` will be empty.

PR #2451, opened 2017-10-31
Switch to Open Sans.

manthey:
On 10/31/2017, Droid Sans is no longer available from Google.  Open Sans is very similar.  Open Sans has a true italic, so the italic variations have been added.

It is possible that Droid Sans will come back, but currently Google lists it as only available from Monotype.


jeffbaumes:
Thanks! Wanted to do this anyway - https://github.com/girder/girder/pull/2408

PR #2457, opened 2017-11-03
Always JSON encode setting values

cjh1:
Otherwise type information is lost and validation of settings may fail. For example setting a boolean value in a playbook gets converted to a python boolean (True or False). This is sent in the the request as the string 'True' or 'False'. When the setting value it received a json.loads(...) is performed and the value is still a str rather then an boolean. By encode as JSON, True or False would become 'true' or 'false', which is decode correctly.

### DIFF
--- a/devops/ansible/roles/girder/library/girder.py
+++ b/devops/ansible/roles/girder/library/girder.py
@@ -1849,10 +1849,7 @@ class GirderClientModule(GirderClient):
 
             params = {
                 'key': key,
-                'value': (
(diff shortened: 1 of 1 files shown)


danlamanna:
Is there an easy built-in setting we can use for a test of this in
test_setting.yml? (The tests aren't currently being run, but that's soon
to change).

cjh1:
@danlamanna Not that I know of, came across it with the oauth plugin.

PR #2464, opened 2017-11-07
Fix regression with model name

cjh1:


cjh1:
Done :-)

cjh1:
@zachmullen I made it static, it doesn't need self.

PR #2467, opened 2017-11-08
When importing S3, show a browser widget button.

manthey:
We have this on the filesystem assetstore import.


manthey:
We should probably also do this on the HDFS assetstore.

We don't seem to have any explicit tests for this on the filessystem assetstore variant, and I haven't added any here (but tested it locally).

PR #2468, opened 2017-11-08
Support range requests on S3 access.

manthey:
Without this, seeks don't work in a file context.


manthey:
@zachmullen perhaps you'd be the best person to review this.

manthey:
Yes, it has been tested with an actual S3 assetstore.

Issue #2469, opened 2017-11-09
Make it harder to delete collections


manthey:
We've had a request to make it harder to delete collections.  A user thought they were deleting a file and accidentally deleted a large collection when clicked the confirmation dialog.  They recommend something like Github uses where you must type the name of the collection to confirm deleting it -- something like "Please type 'DELETE <collection name>' and click the Delete button to confirm deleting the collection".

Perhaps this should also apply to deleting users.

zachmullen:
+1

mgrauer:
I'm quite sympathetic to this because I once deleted a folder with 200 G in subfolders--so this case (folder) still wouldn't be covered, but Collection seems like the right level to avoid annoying the user with confirmations for every delete request.

Perhaps as part of the deletion confirmation, we could say "you want to delete a User/Collection with X folders and Y bytes" so they can see the consequences--though we may have that info easily at hand.

manthey:
At the peril of making things complicated, we could have a number-of-things and size-of-data threshold when deleting folders to engage the harder-to-delete method.

kotfic:
+1 for thresholds

Pierre-Assemat:
Just to have some clarification, the idea is to make sure the user know what is doing?
- This is limited only to Collection? or others models like folder, or user too?
- Do you think only a message with size-of-data or number of element is sufficient? 
- Or do we need an user interaction like enter a message ('delete', or 'yes'...)?

danlamanna:
> At the peril of making things complicated, we could have a
> number-of-things and size-of-data threshold when deleting folders to
> engage the harder-to-delete method.

Length of time the collection has been around could be another useful
metric.

manthey:
I think that initially this just applies to collections, but should be made so it can be reused when we decide where else we want it.  I *think* we'll also want it for users.  I imagine that the delete confirmation would become a dialog, sort of like this:


![image](https://user-images.githubusercontent.com/8781639/32629676-ffc434f0-c567-11e7-9de1-4d483badd0dd.png)

Pierre-Assemat:
Ok so let start on a dialog box reusable, maybe by specifying the type (folder, user, collection...)
I will make a PR soon with the first prototype.

mgrauer:
This sounds like a good approach to me. The only concern I'd have is if the collection has a long name or maybe a name that isn't easy to type on a certain keyboard, that could be a pain. Minor quibble.

How about requiring Users to type "DELETE". 

[...]

PR #2473, opened 2017-11-10
Add security when deleting Collection

Pierre-Assemat:
This PR will add a new security on the confirm dialog box. This will ask the user to enter 'DELETE name_of_collection' and press the "Delete" button.

The dialog box confirm is reusable for any other models by adding these 3 parameters : 
- `name`: name of the element to delete
- `additionalText`: some additional text with information (like size of data, number of items or folders)
- `msgConfirmation`: true for activate the confirmation, false is the value by default

Fixes: #2469

### DIFF
--- a/clients/web/src/dialog.js
+++ b/clients/web/src/dialog.js
@@ -93,7 +93,6 @@ function confirm(params) {
         el.text(params.text);
     }
     if (params['msgConfirmation']) {
-        $('#g-confirm-text').css({'max-width': '50%'});
         if (params.escapedHtml) {
             $('.g-additional-text').html(params.additionalText);
         } else {
--- a/clients/web/src/dialog.js
+++ b/clients/web/src/dialog.js
@@ -62,6 +62,11 @@ function handleOpen(name, options, nameId) {
  * @param {Boolean} [params.escapedHtml] If you want to render the text as HTML rather than
  *        plain text, set this to true to acknowledge that you have escaped any
  *        user-created data within the text to prevent XSS exploits.
+ * @param {Boolean} [params.msgConfirmation] If you want to add a new security before
+ *        perform an action. This will ask to enter a specific string "params.yesText params.name"
+ * @param {String} [params.additionalText] Additional text to display before the confirmation
+ *        input.
+ * @param {String} [params.name] The name to enter in order to confirm an action.
  * @param {Function} [params.confirmCallback]Callback function when the user confirms the action.
  */
 function confirm(params) {
@@ -70,7 +75,10 @@ function confirm(params) {
         yesText: 'Yes',
         yesClass: 'btn-danger',
         noText: 'Cancel',
-        escapedHtml: false
+        escapedHtml: false,
+        msgConfirmation: false,
+        additionalText: '',
+        name: ''
     }, params);
     $('#g-dialog-container').html(ConfirmDialogTemplate({
         params: params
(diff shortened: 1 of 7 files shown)


Pierre-Assemat:
The dialog box looks like that : 

screenshot from 2017-11-14 17-44-17

There is no specific style on it, do you think this is sufficient or do we need something a little bit more attractive ?

clients/web/src/dialog.js
```diff
@@ -70,24 +75,50 @@ function confirm(params) {
...
-        $('#g-dialog-container').modal('hide');
-        params.confirmCallback();
+        if (params['msgConfirmation']) {
+            const key = `${params.yesText.toUpperCase()} ${params.name}`;
+            let msg = $('#g-confirm-text').val();
+            if (msg.toUpperCase() === key.toUpperCase()) {
+                $('#g-dialog-container').modal('hide');
+                params.confirmCallback();
+            } else if (msg.toUpperCase() === '') {
+                $('.g-msg-error').html(`Error: Tape <b>'${key}'</b>, then click on the button`);
```
manthey: `Tape` should be `Type`.  It might be better to change this message to `Error: You need to enter <b>'${key}'</b>.`.

clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -435,6 +435,16 @@ var HierarchyWidget = View.extend({
...
             }, this)
         };
+        if (type === 'collection') {
+            params = _.extend({
+                name: this.parentModel.escape('name'),
+                additionalText: '<b>' + this.parentModel.escape('name') + '</b>' +
+                                ' contains <b>' + this.parentModel.escape('nFolders') +
+                                ' folders</b> taking up <b>' +
+                                formatSize(this.parentModel.escape('size')) + '</b>',
+                msgConfirmation: true
```
manthey: If the size and `nFolders` are both 0, I think we should skip the confirmation check.

clients/web/test/spec/collectionSpec.js
```diff
@@ -304,6 +304,24 @@ describe('Test collection actions', function () {
             return $('#g-confirm-button:visible').length > 0;
         }, 'delete confirmation to appear');
 
+        waitsFor(function () {
```
manthey: Can you also add a test for deleting with no value entered, since we have a code path for that error?

Pierre-Assemat: This test has to try to delete a collection without enter any value, but it cannot succeed right ?

manthey: Yes.

clients/web/src/dialog.js
```diff
@@ -70,24 +75,50 @@ function confirm(params) {
...
 
-    var el = $('#g-dialog-container').find('.modal-body>p');
+    const el = $('#g-dialog-container').find('.modal-body>p:first-child');
     if (params.escapedHtml) {
         el.html(params.text);
     } else {
         el.text(params.text);
     }
+    if (params['msgConfirmation']) {
+        $('#g-confirm-text').css({'max-width': '50%'});
```
manthey: It would be better to have this in a `styl` definition file.  We have other dialog style css in `clients/web/src/stylesheets/layout/global.styl`.

clients/web/src/views/body/CollectionView.js
```diff
@@ -156,6 +138,37 @@ var CollectionView = View.extend({
...
+                    router.navigate('collections', {trigger: true});
+                }).destroy();
+            }, this)
+        };
+        if (this.model.get('nFolders') !== 0 || this.model.get('size') !== 0) {
+            params = _.extend({
+                additionalText: '<b>' + this.model.escape('name') + '</b>' +
+                                ' contains <b>' + this.model.escape('nFolders') +
+                                ' folders</b> taking up <b>' +
+                                formatSize(this.model.escape('size')) + '</b>',
```
manthey: `size` should be an integer, so this should be `formatSize(parseInt(this.model.get('size'), 10))`, using `parseInt` to guard against bad values rather than `escape`.

clients/web/src/dialog.js
Pierre-Assemat: Yes you're right!
Do you think this should use a class instead and put it within global.styl ?
Even if it applies only on this special case

manthey: I would leave it id specific, leaving it to be generalized if we need to at a later time.

Pierre-Assemat: Ok sounds great!

clients/web/src/views/widgets/HierarchyWidget.js
```diff
@@ -435,6 +435,17 @@ var HierarchyWidget = View.extend({
...
             }, this)
         };
+        if (type === 'collection' &&
+           (this.parentModel.get('nFolders') !== 0 || this.parentModel.get('size') !== 0)) {
+            params = _.extend({
+                name: this.parentModel.escape('name'),
+                additionalText: '<b>' + this.parentModel.escape('name') + '</b>' +
+                                ' contains <b>' + this.parentModel.escape('nFolders') +
+                                ' folders</b> taking up <b>' +
+                                formatSize(this.parentModel.escape('size')) + '</b>',
```
manthey: This should use `parseInt`, too.

Pierre-Assemat: Yes indeed, my bad.

PR #2480, opened 2017-11-14
Adding plugin to display contents of CSV, TSV in a table

jeffbaumes:
A simple viewer that loads the contents of items ending with .csv or .tsv into the browser and makes a paginated table, akin to the logic of the Candela plugin.

image

Still needs a bit of documentation.


manthey:
Should this check the mime type as well, so that files with a mime type `text/csv` would also work regardless of their name?  Should this be based on the file name not the item item (or ignore items with a number of files that isn't 1)?

danlamanna:
Mind adding an entry to `CHANGELOG.rst`?

jeffbaumes:
This is ready for review. @manthey PTAL when you get a chance.

danlamanna:
So this downloads the entire file and stores it in browser memory? Should there be some sort of threshold for this? Even though it requires clicking the arrow to render, there is certainly a point at which it will just crash the browser.

jeffbaumes:
Yeah. The candela plugin currently does the same thing without asking (that should be changed). I haven't looked into whether the CSV parser can handle partial files - ideally we'd request the beginning bytes of the file and show the first X rows. The easiest way I can think of to guard against crashing the browser is simply to not even try to show a table at all when the size is above a certain threshold. Any ideas for what that threshold would be? The in-memory representation currently is fairly bloated, with name-value pairs in objects for each row, so in-memory size is likely significantly higher than on-disk size.

danlamanna:
From
https://help.github.com/articles/rendering-csv-and-tsv-data/#handling-errors:
"Our rendering only works for files up to 512KB. Anything bigger than
that slows down the browser."

512KB seems very conservative to me, I would think on the order of 3-5MB
would seem more reasonable?

jeffbaumes:
Here are some scalability numbers:

| Rows | Columns | File size | Time to load table view | Chrome memory footprint after load |
| --- | --- | --- | --- | --- |
| 1K | 20 | 74KB | 0.5s | 19MB |
| 10K | 20 | 750KB | 2s | 22MB |
| 100K | 20 | 7MB | 15s | 66MB |
| 1M | 20 | 72MB | 142s | 425MB |

None of these crashed my browser and you can always page through the table quickly once loaded, but the time to load gets prohibitive. My thinking is to go with a 10MB limit, add a "working" indicator while it's loading because it could take 20 seconds or so, and notify the user when the file is too large.

manthey:
I think it would be useful to test a tsv file (and the next page / previous page buttons).

I tested this with a large 27 Mb file, and the performance was acceptable.

I also tested this with a very wide csv file.  The formatting could be nicer:
image
It is tedious to have to scroll down to scroll horizontally and then scroll up to go to the next 10 rows.

plugins/table_view/web_client/stylesheets/tableWidget.styl
```diff
@@ -0,0 +1,18 @@
...
+  display flex
+
+  .g-item-table-view-title
+    flex 1
+
+.g-item-table-view-container
+
+  table
+    display block
+    overflow scroll
```
manthey: I'd recommend adding something like
```
li.g-table-view-page-next,li.g-table-view-page-prev
  user-select none
```
so that one doesn't select (and highlight) the text of the buttons.  I found that clicking repeatedly would end up blink the highlight on the button text otherwise.  This gets added to `.btn` classes automatically, so the other way would be to change the css class.

jeffbaumes:
Ready for review. Addressed the following:

* There are now explicit states for loading, data too large (30MB limit), parsing error, in addition to collapsed and expanded. Appropriate messages and interactions are available for each state.
* Added test for TSV file.
* Made Prev/Next button text nonselectable.
* Ensure table cell text does not wrap so the view remains at a fixed height.

PR #2481, opened 2017-11-15
Fix the terms client test.

manthey:
When we use `window.location.assign`, make sure to wait for something that we expect to change.  Just doing `girderTest.waitForLoad()` is insufficient, since window.assign doesn't cause the change to start until the next time slice, and `waitForLoad` will return immediately that things are loaded (probably a `waits(1)` would have worked instead of the `waitsFor` that was added, but that seems more brittle).

Once we've properly waited for the `window.location.assign`, fix waiting for creating an item.  A comment was added that waiting caused a test to fail, which was a good indication that we *weren't* waiting for the correct thing.

Also, fix throwing an error in the test utilities to use the error constructor.

### DIFF
--- a/plugins/terms/plugin_tests/termsSpec.js
+++ b/plugins/terms/plugin_tests/termsSpec.js
@@ -1,3 +1,5 @@
+/* global girderTest, describe, it, expect, runs, waitsFor, girder, beforeEach */
+
 girderTest.importPlugin('terms');
 girderTest.startApp();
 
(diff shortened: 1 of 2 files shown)


plugins/terms/plugin_tests/termsSpec.js
```diff
@@ -79,10 +89,10 @@ describe('Navigate to a non-collection folder and item', function () {
 
     it('create an item', function () {
         runs(function () {
-            return $('.g-create-item').click();
+            $('.g-create-item').click();
         });
-        // This causes the test to fail
-        // girderTest.waitForDialog();
+        girderTest.waitForDialog();
```
brianhelba: @manthey To be clear, the reason you believe this line was failing before was that the page state was typically synchronized from the test by this point?

plugins/terms/plugin_tests/termsSpec.js
```diff
@@ -65,12 +67,20 @@ describe('Navigate to a non-collection folder and item', function () {
...
             expect($('.g-hierarchy-breadcrumb-bar>.breadcrumb>.active').text()).toBe('Public');
             var folderId = window.location.hash.split('/')[3];
             expect(folderId).toMatch(/[0-9a-f]{24}/);
             window.location.assign('#folder/' + folderId);
         });
+        // after setting a window location, waitForLoad is insufficient, as the
+        // page hasn't yet started making requests and it looks similar to
+        // before the location change.  Wait for the user header to be hidden,
+        // and then wait for load.
+        waitsFor(function () {
```
brianhelba: I suppose it's theoretically possible to have another race condition here, whereby the navigation happens so fast that the `waitsFor` never gets to poll it during the incomplete state? However, I admit that under the current test conditions, that's highly unlikely.

manthey: It can't happen, since we were at `#user/id/folder/id` and go to `#folder/id`, and the .g-user-header element will no longer be present after that (it is the element that lists the user when `#user` is part of the route).

plugins/terms/plugin_tests/termsSpec.js
manthey: Before, we failed to wait for the page transition to `#folder/id` before invoking the dialog.  If we waited for the dialog, that transition eventually could trigger and remove the dialog, causing this to fail.  Now, we properly wait for `#folder/id` and then open a dialog so we can properly wait for the dialog.

Issue #2482, opened 2017-11-15
item_tasks client test is locale dependent


manthey:
The item_tasks client tasks test is locale dependent, and fails when the tests are run in the C locale instead of the en.UTF-8 locale.

We expect names to be sorted alphabetically, which means that, in e.UTF-8, they are in the order 'item-tasks-demo', 'me/my_image', 'PET phantom...', but in the C locale this changes to 'PET phantom...', 'item-tasks-demo', 'me/my_image'.  To make the test more robust, we either need to (a) make sure that they sort correctly regardless of locale (e.g., change "PET phantom..." to "pet phantom..."), (b) change the client test to find the correct item in the list and not insist on a strict position, (c) force mongo to use a specific locale (but that requires Mongo >= 3.4).

danlamanna:
Is running the test suite in a different locale as simple as specifying
LANG=C ctest ... ? Was this the only test that failed out of our test
suite?

manthey:
Yes, that is all that is required to run in another locale, and, yes, this was the only test that failed.

I encountered this on @brianhelba's provisional circle-ci version 2 branch in their environment (for Python 3, though I don't think that is relevant).  There, I *can't* just change the locale via LANG=en.UTF-8, as they don't have the English locale installed.

manthey:
I've fixed the test in PR #2483 via method (b).

PR #2483, opened 2017-11-15
Make the task client test work in different locales.

manthey:
Because we sort tasks by name, the sort order varies by locale.  For instance, in the C locale, capital letters sort before lowercase, whereas in the en.UTF-8 locale, the alphabetically position is more import than case.  Get the position of data explicitly for testing to avoid locale dependency.

### DIFF
--- a/plugins/item_tasks/plugin_tests/tasksSpec.js
+++ b/plugins/item_tasks/plugin_tests/tasksSpec.js
@@ -277,12 +277,9 @@ describe('Navigate to the new JSON task', function () {
 
         runs(function () {
             expect($('.g-execute-task-link').length).toBe(3);
-            // get the position of the alphabetized entry; depending on locale
-            // it may be in different positions
-            var position = $('.g-execute-task-link').index($(':contains("me/my_image:latest 0"):last') );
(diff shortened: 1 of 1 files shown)


brianhelba:
I think this would be much simpler to just get rid of position variables entirely, and look up DOM elements directly each time using the `:contains` pseudo-selector. IMO, that's what the test should have done all along, so there's no reason to pile on even more complex logic to keep the test logic exactly the same.

brianhelba:
Also, even if this addresses the issue manifested in #2482 (which is good), I'm more concerned about Girder's dependency on `LOCALE` in general. We should eventually address this issue more directly.

manthey:
Of course Girder is dependent on locale, since we use Mongo and text sorting and Mongo is dependent on locale.  As of Mongo 3.4 you can specify Mongo's collation locale on a per-query basis.  Ideally, we'd use the client's locale to advise the end points so that text would be sorted according to that locale.

brianhelba:
Thanks for putting the rationale in your commit messages. Since some developers don't structure their commits at all, I'd suggest making a note in the PR description that reviewers should read them.

PR #2485, opened 2017-11-16
[api/docs] Add tag describing origin of each method

Xarthisius:
This PR allows to filter swagger api docs by the origin (plugin or core) of an endpoint, by adding a custom tag.

Long story short, I've been recently asked to present "how, and to what extent, did we expanded existing Girder's capabilities?". Having that filter would have been extremely helpful... ;-)

### DIFF
--- a/girder/api/docs.py
+++ b/girder/api/docs.py
@@ -20,6 +20,7 @@
 import collections
 import functools
 import six
+import inspect
 from girder import logprint
 
 models = collections.defaultdict(dict)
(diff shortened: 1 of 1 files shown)


girder/api/docs.py
```diff
@@ -53,8 +54,13 @@ def _toOperation(info, resource, handler):
     """
     Augment route info, returning a Swagger-compatible operation description.
     """
+    module_name = inspect.getmodule(handler).__name__
+    if module_name.startswith('girder.plugins'):
+        module_tag = 'plugin:' + module_name.split('.')[2]
```
cjh1: @Xarthisius  Sorry for the delay in reviewing this PR. This check looks a little fragile, in fact it will be broken by Girder 3.0 as plugins are moving to being installed using pip. It seems like your use case is to tags plugins that you have added to a Girder instance. So instead of automatically tagging all plugins ( based on module name) could we instead add API to explicitly add tags to a plugins endpoints, for example add a parameter to the describe route decorator?

zachmullen:
@Xarthisius is this still a desired feature? The pip installable plugin branch from @jbeezley is nearly ready, so your feature could be developed against that branch. I'll close this PR since it won't go in under this form.

PR #2488, opened 2017-11-20
WIP: Support dynamic content sizes for multi-part upload

danlamanna:
Warning: this was written off the cuff only with the FS assetstore.

> This should allow the uploader to stream data into Girder without
knowing the final size ahead of time, useful for dynamic outputs from
analyses or other data sources which may be larger than normal.

One of the advantages of multipart upload is the ability to stream data to the server as it's being created (without knowing the exact total byte count). This PR alters the upload model to allow the size to be `None` - in which case the bytes are done being sent when `finalizeUpload` is called and the resulting `size` is `upload['received']`. Applications that were passing `size` should continue to function as they did before.

AFAICT, this doesn't seem to cause any conceptual problems with our existing assetstores since both S3 and GridFS allow for this behavior.

This needs:
- To be tweaked to work with other assetstores
- More tests
- More sanity checking

### DIFF
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -78,12 +78,15 @@ class File(Resource):
                'this query by sending it as the body of the request using an '
                'appropriate content-type and with the other parameters as '
                'part of the query string.  If the entire file is uploaded via '
-               'this call, the resulting file is returned.')
+               'this call, the resulting file is returned, to ensure that Girder '
+               'knows the entire file was uploaded, the size parameter is required. '
+               'If the size parameter is omitted, Girder will not finalize the file '
+               'upload until an empty chunk is POSTed.')
         .responseClass('Upload')
         .param('parentType', 'Type being uploaded into.', enum=['folder', 'item'])
         .param('parentId', 'The ID of the parent.')
         .param('name', 'Name of the file being created.')
-        .param('size', 'Size in bytes of the file.', dataType='integer', required=False)
+        .param('size', 'Size in bytes of the file.', required=False)
         .param('mimeType', 'The MIME type of the file.', required=False)
         .param('linkUrl', 'If this is a link file, pass its URL instead '
                'of size and mimeType using this parameter.', required=False)
@@ -115,7 +118,9 @@ class File(Resource):
                     url=linkUrl, parent=parent, name=name, parentType=parentType, creator=user,
                     size=size, mimeType=mimeType), user)
         else:
-            self.requireParams({'size': size})
+            if size is not None:
+                size = int(size)
+
             assetstore = None
             if assetstoreId:
                 self.requireAdmin(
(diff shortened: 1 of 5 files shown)


zachmullen:
I'd like to review this before it gets merged, will take a look next week

cjh1:
Is this using Chunked transfer encoding?

danlamanna:
> I'd like to review this before it gets merged, will take a look next
> week

Nod. It's nowhere near getting merged at the moment, just wanted to put
something up in raw form in case there was any immediate feedback.

> Is this using Chunked transfer encoding?

I don't believe so, no. I hadn't actually seen that we had support for
this until grepping just now, though it doesn't seem to be actively
being used.


Ultimately, I want to do whatever is the most idiomatic in the web world
that still allows us to stream data into Girder without knowing how many
bytes there will be. The only hard limitation I can see is that chunks
must be uploaded sequentially due to Girder's on-the-fly checksumming.

zachmullen:
The idiomatic way would be with chunked transfer encoding header, but this also involves girder policy issues about pre-approval of upload sizes for core file uploads.

cjh1:
I would have thought 'Chunked transfer encoding' would be the most idiomatic, but I may well be wrong, oops @zachmullen beat me to it ... Also the sequential upload requirement is kinda at odds with chunked transfer encoding as the chunk are supposed to be uploaded independently.

cjh1:
My interest in this is it would be nice to be able stream file contents out of a docker container ...

zachmullen:
@cjh1 that was precisely the use case I was aiming for when I originally tested this end-to-end, but I never added any support to the core file upload for it. The initial prototype assumed a custom endpoint would handle the stream so we could punt on the core policy decisions.

I'd be amazed if it still works from the worker side, but that's definitely the prime use case for me too.

danlamanna:
> The idiomatic way would be with chunked transfer encoding header, but
> this also involves girder policy issues about pre-approval of upload
> sizes for core file uploads.

Could you expand on this? Does this have something to do with the quota
functionality?

zachmullen:
Quota enforcement is the most obvious application of Girder's core policy of requiring the upload initialization step to require the final size of the file to be declared. It's a guarantee that Girder has made with a relatively high degree of explicitness, so breaking it could have implications we haven't considered, beyond just the quota plugin.

cjh1:
@zachmullen I am giving the "transfer encoding adapter" a new lease on life as part of dock_run ... So if we end up allow transfer encoding on the upload endpoint we will be able to use it ...

zachmullen:
Also, we should make it explicit that the way to do this mode of upload is to use `Transfer-Encoding: chunked`, and our testing should do that. @cjh1 should probably help design that support since he's implemented the worker side of it.

zachmullen:
This concept should persist, but the code for it will look different, based on the comments on this branch.

Issue #709, opened 2015-02-20
Prefer href rather than event binding where possible


zachmullen:
We have places in the web client where we have an element where we bind a click action that basically just calls navigate on the router to another page. We should instead just set the fragment as the href of those links where it makes sense to do so, which will allow users to see the URL appear in their browser when they hover, and do things like open link in new tab, copy link address, etc.

brianhelba:
+1

cpatrick:
+1000

-Patrick

jcfr:
:+1:

manthey:
This was largely done.  I'm closing the issue.

PR #2489, opened 2017-11-21
Add href properties to front page links that were missing them

ronichoudhury:
It was bothering me that the front page links to the collections page and the logged-in user's home folder and user account info were javascript-only - this adds appropriate href properties to those links.

TODO:
- [x] remove leading slashes from href properties
- [x] remove stray `a` from pug property list
- [x] apply change to links in /#admin
- [x] remove event handlers for appropriate clicks
- [x] add content to CHANGELOG


danlamanna:
This solves one of my pet peeves of Girder breaking middle/control click
opening links in new tabs. However, these problems still exist in the
user menu (top right dropdown), is it possible to fix these instances as
well?

ronichoudhury:
I'll give that a go, @danlamanna.

While we're on the subject, I was wondering about the login/logout/register links - those don't have actual concrete URLs do they?

jbeezley:
Not for logout to my knowledge, but there are these which work when you aren't logged in.

https://data.kitware.com/#?dialog=login
https://data.kitware.com/#?dialog=register

ronichoudhury:
@danlamanna: I did the menu links.

@jbeezley: Thanks! I added hrefs for the login and register links though I'm less sure those should be there. They seem to work perfectly though (even allowing for middle-click). Let me know what you think.

clients/web/src/templates/body/frontPage.pug
```diff
@@ -16,17 +16,17 @@
         ul
           li
             | To explore the data hosted on this server, start with the
-            a.g-collections-link  Collections
+            a.g-collections-link(href="/#collections")  Collections
```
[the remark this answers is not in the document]
ronichoudhury: It's just the leading slash that would cause that breakage?

mgrauer:
Can you add an entry for this to the CHANGELOG file? I could see it going in changes or bugfixes, your call. It should go in a `Web Client` subsection.

clients/web/src/templates/layout/layoutHeaderUser.pug
```diff
@@ -6,11 +6,11 @@
     #g-user-action-menu.dropdown
       ul.dropdown-menu(role="menu")
         li(role="presentation")
-          a.g-my-folders
+          a.g-my-folders(a href=`/#user/${user.id}`)
```
brianhelba: There is an `a` here as an attribute.

clients/web/src/templates/body/frontPage.pug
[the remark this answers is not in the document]
brianhelba: Yes, we should start all these with the fragment `#`.

clients/web/src/templates/layout/layoutHeaderUser.pug
```diff
@@ -6,11 +6,11 @@
...
             i.icon-folder
             | My folders
         li(role="presentation")
-          a.g-my-settings
+          a.g-my-settings(a href=`/#useraccount/${user.id}/info`)
```
brianhelba: Here too.

danlamanna:
Also the links on #admin.

mgrauer:
related to #709

brianhelba:
Just for the record, the JS `click` event handlers for all these should be removed before this is merged, but I assume you'll do that later.

ronichoudhury:
I'm workin' on it :wink:

ronichoudhury:
@girder/developers: I believe I've addressed all raised concerns. PTAL and let me know if I missed anything. Thanks!

clients/web/src/views/layout/LoginView.js
```diff
@@ -54,10 +54,6 @@ var LoginView = View.extend({
             }, this));
         },
 
-        'click a.g-register-link': function () {
```
brianhelba: I don't think you want to be removing this, since you're not updating the template for *this* particular view.

clients/web/src/views/layout/RegisterView.js
```diff
@@ -75,10 +75,6 @@ var RegisterView = View.extend({
 
             this.$('#g-register-button').girderEnable(false);
             this.$('.g-validation-failed-message').text('');
-        },
-
-        'click a.g-login-link': function () {
```
brianhelba: I don't think you want to be removing this, since you're not updating the template for *this* particular view.

clients/web/src/views/layout/ResetPasswordView.js
```diff
@@ -38,14 +38,6 @@ var ResetPasswordView = View.extend({
 
             this.$('#g-reset-password-button').girderEnable(false);
             this.$('.g-validation-failed-message').text('');
-        },
-
-        'click a.g-register-link': function () {
```
brianhelba: I don't think you want to be removing these, since you're not updating the template for *this* particular view.

ronichoudhury:
@brianhelba I fixed up the appropriate templates and merged master in. PTAL and let me know if there are still problems (or if I introduced new problems).

clients/web/src/views/layout/LoginView.js
```diff
@@ -54,10 +54,6 @@ var LoginView = View.extend({
...
-            events.trigger('g:registerUi');
```
brianhelba: I'm concerned about what the loss if these event triggers will do to the reusability of the `LoginView`, etc. While triggering events on the global event bus (the current behavior) isn't ideal, it does not impose any particular routing setup on the owner of the view. Changing this will force any user of the `LoginView` to adapt the same user-facing routing scheme as Girder's web client app.

Accordingly, I think we should skip adding `href` for any behavior that is implemented by an `events.trigger` (in all the places in this PR).

Getting rid of `router.navigate` is a worthwhile change, and doesn't have any broader implications for reusability.

jbeezley: Indeed, this would break how we use the login dialog in HistomicsTK.  I think you could get away with leaving in the href if you added a `preventDefault` call in these handlers.

ronichoudhury: I put back the events for the login/register/forgot-password dialogs. Let me know if it looks ok.

clients/web/src/templates/body/frontPage.pug
```diff
@@ -37,14 +37,14 @@
           ul
             li
               | To explore the data hosted on this server, start with the
-              a.g-collections-link  Collections
+              a.g-collections-link(href="/#collections")  Collections
```
manthey: Leading slash.

ronichoudhury: Good catch, thanks.

clients/web/src/views/layout/LoginView.js
```diff
@@ -55,7 +55,7 @@ var LoginView = View.extend({
         },
 
         'click a.g-register-link': function () {
-            events.trigger('g:registerUi');
+            events.trigger('g:registerUI');
```
brianhelba: You changed the spelling here.

clients/web/src/views/layout/HeaderUserView.js
```diff
@@ -15,23 +14,7 @@ import 'bootstrap/js/dropdown';
  */
 var LayoutHeaderUserView = View.extend({
     events: {
-        'click a.g-login': function () {
-            events.trigger('g:loginUi');
```
brianhelba: The `events.trigger` here need to stay.

clients/web/src/templates/layout/resetPasswordDialog.pug
```diff
@@ -15,9 +15,9 @@
         .g-validation-failed-message
         .g-bottom-message
           if registrationPolicy !== 'closed'
-            a.g-register-link  Register
```
brianhelba: Remove here.

clients/web/src/templates/layout/resetPasswordDialog.pug
```diff
@@ -15,9 +15,9 @@
...
+            a.g-register-link(href="#?dialog=register")  Register
             |  |
-          a.g-login-link  Login
```
brianhelba: And remove here.

clients/web/src/templates/layout/registerDialog.pug
```diff
@@ -27,7 +27,7 @@
         if !currentUser
           .g-bottom-message
             | Already have an account?
-            a.g-login-link  Log in here.
```
brianhelba: And remove here.

clients/web/src/templates/layout/loginDialog.pug
```diff
@@ -16,7 +16,7 @@
           .g-bottom-message
             if registrationPolicy !== 'closed'
               | Don't have an account yet?
-              a.g-register-link  Register here.
```
brianhelba: And remove here.

brianhelba:
@zachmullen @ronichoudhury I mostly agree. Since dialog routing is such a minefield, I'd just prefer to drop anything that says `#?dialog=login` (including on the front page), because I don't see any significantly increased usability from it, and it could eventually become a liability that's more work to remove and/or support in the future.

For the record, with respect to future plans, I definitely support rethinking routing. My own goals revolve more around the encapsulation of widgets (i.e. having them not trigger routing directly, nor trigger global events directly, but rather having them trigger events on themselves, which the instantiating class can hook up to routing or other actions). This is totally compatible with other routing changes too.

brianhelba:
> We should keep the changes that trivially change from event handlers that call router.navigate to hrefs.

Also, to be clear, I think this is the real value in this PR, and it seems totally uncontroversial. I'd support merging this bit ASAP.

ronichoudhury:
@zachmullen PTAAL

zachmullen:
I am fine with this, but you may have to remove the dialog routes from the front page in order to appease @brianhelba

ronichoudhury:
The question is, does anyone currently link directly into the login/registration/forgot-password dialogs from outside? Such links will break if we remove that stuff from the front page. I'm not sure there's much value in changing this right now, as opposed to waiting for a routing overhaul effort.

brianhelba:
Yeah, I'm still concerned about those, and I don't see how they add any real value.

brianhelba:
Also, the more we use these links, the harder it will be to remove them (in terms of people and versions, not technology). I'd rather not go further down the path of dialog routing. We can come up with better ways of making a login / logout link, if that's needed (even without a total overhaul)

ronichoudhury:
@brianhelba, I got rid of the dialog routing on the front page.

zachmullen:
@ronichoudhury the web client tests are failing because `.click()` on an href element is for some reason not equivalent to following its href attribute. In the past I've worked around this by substituting the click events with e.g.:

    window.location.assign($('a.g-file-list-link').attr('href'));

However we may want to solve this generally by augmenting jquery's click() function within testUtils.js.

brianhelba:
https://stackoverflow.com/a/21762745 ?

zachmullen:
@brianhelba didn't know that... seems like a bug in jquery. I'm fine with changing our testing code to do that instead.

clients/web/test/testUtils.js
```diff
@@ -11,6 +11,17 @@ window.alert = function (msg) {
     expect('Alert was used').toBe('Alerts should not be present');
 };
 
+// Augment jQuery's click() function to follow href properties in links if they exist.
+$.fn.click = _.wrap($.fn.click, function (old) {
+    var args = Array.prototype.slice.call(arguments, 1);
```
brianhelba: Why can't you do `arguments.slice(1)` here?

clients/web/test/testUtils.js
```diff
@@ -11,6 +11,17 @@ window.alert = function (msg) {
...
+    var href = $(this).attr('href');
```
brianhelba: You only really need to bother building and evaluating this selector if `args.length === 0`, so you should probably have a nested `if` (rather than a single level with combined conditions) and create this selector only if necessary.

brianhelba: Do you need to re-wrap `this`? We're already calling `.click()` on a jQuery selector, right?

brianhelba: While uncommon, it is possible for non-`<a>` elements to have an `href` attribute, sometimes with different semantics. I think we should also check that `this` is an `<a>` element.

clients/web/test/testUtils.js
```diff
@@ -11,6 +11,17 @@ window.alert = function (msg) {
...
+    if (args.length === 0 && href) {
+        window.location = href;
```
brianhelba: I'm having trouble consistently reproducing the `click` failure on an actual Girder site. Is it possible that this is a browser-based issue (i.e. PhantomJS)? Can you provide more details in code comments on when this workaround is necessary, and when it could be removed in the future?

clients/web/test/testUtils.js
zachmullen: The silliness of javascript: `arguments` is not an `Array`.

clients/web/test/testUtils.js
zachmullen: Probably a good idea, although I'm curious what other elements support href attributes?

clients/web/test/testUtils.js
zachmullen: Are you actually clicking, or calling jquery `click()` to try and reproduce it?

brianhelba: I was slightly mistaken initially, I was calling jQuery's `.click` on something that coincidentally also had Backbone behavior on it. Now I'm able to reproduce the bug.

However, per https://stackoverflow.com/a/21762745 I am able to make things work properly by `.click()`ing the underlying DOM element. I think it'd be safer for this method to just unwrap the jQuery selector and call [`click()` on the native element](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click). That comes back around to my other question: what is the type of `this` here?

clients/web/test/testUtils.js
brianhelba: From a quick search of MDN: several different things under `svg`, `area`, etc.

When we're monkey-patching such a core part of jQuery in our test environment, it's probably best to not make assumptions about what's being clicked.

clients/web/test/testUtils.js
brianhelba: Ah, I see now.

However, it appears that this usage of `arguments` is a known optimization killer:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments
* https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments
I think it's fine in testing, but should definitely never make it to production. Could you add a code comment stating so?

zachmullen: This whole patch might be going away, @ronichoudhury and I were running into spooky behavior with the last failing test and we may need a different workaround entirely. It's premature to review any of the commits on this branch that are fixing up the tests.

brianhelba: Got it, no problem.

ronichoudhury:
Well, it's a Christmas miracle!

Somehow, removing the "static" `router.navigate()` event handlers was making the tests fail. My recent commits (manually) reverts the removal of these event handlers, and of the experimental change the `$.fn.click()`.

This is clearly a problem, as removing those handlers *ought not* to have broken the tests. **However**, I am of the opinion that this represents a bug I've serendipitously revealed in this PR, and that fixing that bug is not in scope for this PR.

I'd love to have a new round of reviews (from @brianhelba and @zachmullen) on the current state of this branch (including a weigh-in from @zachmullen on whether the other changes to the testing infrastructure on this branch should also be reverted). If we want to open a new issue regarding the odd behavior that comes with removing those event handlers to launch follow-on work from this PR, that makes sense.

In the end, I just think a change as small as mine should not blossom into a bug-fixing expedition if the change itself can be layered onto the codebase without breaking anything.

Let me know what you think, and thanks for your patience.

clients/web/src/views/layout/HeaderUserView.js
```diff
@@ -26,11 +26,12 @@ var LayoutHeaderUserView = View.extend({
         'click a.g-logout': logout,
 
         'click a.g-my-folders': function () {
-            router.navigate(`user/${getCurrentUser().id}`, {trigger: true});
```
brianhelba: Why is this being changed?

clients/web/src/views/layout/HeaderUserView.js
```diff
@@ -26,11 +26,12 @@ var LayoutHeaderUserView = View.extend({
...
+            router.navigate('user/' + getCurrentUser().get('_id'), {trigger: true});
         },
 
         'click a.g-my-settings': function () {
-            router.navigate(`useraccount/${getCurrentUser().id}/info`, {trigger: true});
```
brianhelba: Why is this being changed?

clients/web/src/views/layout/HeaderUserView.js
ronichoudhury: It shouldn't be. I will revert this change and the one a few lines below.

mgrauer:
You have finally found the secret chord!

The fourth, the fifth
The minor fall and the major lift.

🎼

zachmullen:
> Note, the new behavior works such that both the Javascript handler and the browser's href navigation will fire, essentially routing a user twice, every time these are clicked.

If this is true, I consider that a bug. Why can't we remove the javascript handler?

ronichoudhury:
> 5. I'm not sure this topic branch needs to be 28 commits long, so you could consider a squash merge or rebasing.

I'm not comfortable doing an interactive rebase because I don't know what should be reordered/squashed. What is the benefit of doing it? Or, put another way, why does it matter that there were 28 (or more, or fewer) commits on this branch?

I've addressed most of the other concerns (minus the comment about bugginess, which Zach has taken up in a different thread -- honestly, not sure what to do about that one right now).

manthey:
I think rebasing a branch is fine *if* no one has explicitly commented on specific lines of code in the branch.  However, once someone has commented on existing code, rebasing it loses the context of that comment, which makes determining why the comment was made more difficult.

CHANGELOG.rst
```diff
@@ -44,6 +44,11 @@ Changes
 * Exceptions are now all accessible in the ``exceptions`` module and are descended from the ``GirderBaseException`` class.
 * Require npm 5.2+ (with npm 5.6+ strongly recommended) to build the web client
 
+Web Client
+^^^^^^^^^^
+* Use ``href`` properties for navigation links in addition to JavaScript onclick events. (`#2489 <https://github.com/girder/girder/pull/2489>`_)
+* Change instances of ``.g-server-config`` to ``a.g-server-config`` to enable adding of ``href`` properties to those links
```
brianhelba: This selector won't actually work, since it's not an element that is both an `a` and a `.g-server-config`. Still, I think this provides more than adequate notice for anyone that may care about this particular change in the admin console page (which I doubt anyone would).

ronichoudhury: Oops... I pushed one last doc fix. Once the tests pass I'll merge this. Thanks Brian!

ronichoudhury: Oh, @brianhelba, I guess you'll have to re-review after the tests pass.

zachmullen:
Just to clarify -- does this cause the navigation to happen twice when links are clicked? If that's the case I don't want to merge yet since I consider that an introduction of a new bug...

ronichoudhury:
@zachmullen, @brianhelba: what would be the best way to tell if that's what's happening?

brianhelba:
@zachmullen By changing the URL destinations (of the HREF and JavaScript) to different values and adding a debug statement to the JS, you can see that both trigger (first the JS, then the HREF). I'd be happy for someone to try to reproduce.

We could start calling `event.preventDefault`, but at that point, what would the benefits of the HREF still be? (That's a real question)

ronichoudhury:
The benefit of the `href` remains what it was, namely that it revives the generally expected behavior of links.

However, that seems like a very serious bug. It seems like the correct thing to do is address #2558 first, then add these `href` properties back in.

A less careful approach is to merge this now and then address #2558.

brianhelba:
Sorry, to clarify, my point was just that *if* we use `event.preventDefault` (and I don't outright object, though I think it's a hack), then links would not have the generally expected behavior, even if they seem to work for most users. We ought to reconsider the justification of this PR at that point (though perhaps it still has one).

manthey:
I'll look into why the tests fail (issue #2558).  I'd like to see this tested properly and merged in, as hrefs are valuable to the user.

manthey:
I'll assert that telling a user where they will be taken when clicking a link is important and valuable.

When we remove the explicit backbone events with routing the tests fail because there is no event that jQuery fires on `click()`.  As noted before, this can be done by calling the HTMLElement's `click()` function instead.  Adding the hrefs and removing the backbone events has several benefits, so I think it is worth doing (benefits: the user knows where the link will take them; they can right-click; we reduce needless routing code).

I think the correct course of action is removing the backbone events that do the routing and modifying the tests to use `$(elem)[0].click()` instead of `$(elem).click()`.  The jQuery general click handler that was introduced briefly in this PR works, but obfuscates what is happening.  Switching how we generate clicks means we've changed the javascript needed in testing, which could be seen as an API change, but, I argue, that this should only change tests and not real-world code (and we've broken all sorts of tests lately with refactoring linting and other things, so I don't think we should let such a change stop us).

brianhelba:
@manthey I agree with all of that.

I kind of assumed that calling `click` on the raw HTMLElement has already been tried and was failing for some reason, but I was never clear on what experiments @ronichoudhury and @zachmullen had tried before.

manthey:
@brianhelba The issue of why raw `click` events didn't work was that we were misusing the `li` and `a` elements in the admin console, so all of the references to that would have had to change.  This probably would have broken the tests on external plugins that have settings pages.  The latest commit fixes issues without requiring any test changes (see the commit description).

manthey:
Add, with the most recent commit, it will be easy to add hrefs to item and folder lists without harming the hierarchy widget (you'll be able to right click to open the item in a new window AND click on the item to select it in the hierarchy widget).  Such a change would be just adding the appropriate href in the pug template and removing the `click a.g-item-list-link` handler.

ronichoudhury:
@brianhelba could you take another look at this so we can sweep it up? :smiley_cat:  Thanks!

clients/web/src/views/App.js
```diff
@@ -31,6 +31,19 @@ import 'girder/stylesheets/layout/global.styl';
...
 
 var App = View.extend({
+    events: {
+        'click a': (evt) => {
+            if (!evt.isDefaultPrevented()) {
+                var elem = $(evt.target),
+                    href = elem.attr('href');
+                if (elem.is('a') && href && href.substr(0, 1) === '#') {
+                    router.navigate(href.substr(1), {trigger: true});
+                    evt.preventDefault();
```
zachmullen: Does this code make other people uncomfortable, or is it just me?

During a meeting someone (perhaps jokingly) insinuated that part of the reason this problem was so hellish was because of either Backbone or Girder's idiosyncrasies. Since we're seriously entertaining the idea of a rewrite in a new framework, I think it would be a great exercise to try and figure out how this problem would be addressed in a different candidate framework (Vue or React).

manthey: We could get rid of this code; it allows `$(<a element with href>).click()` to work, which prevents breaking tests AND is nice when poking around manually.  It could also be moved to just the testing app, rather than in the real app (but any differences between the testing app and the real app could hide bugs because the behavior is different).

If we had used hrefs as the preferred method of routing wherever possible, this wouldn't have been neceesary -- all of our tests would have used `$(<a element with href>)[0].click()` if they were using jQuery (and that now works, too).

We have a tension between wanting to make things convenient for the general case, wanting code purity, wanting to make testing easy, wanting to be able to customize everything, wanting to have API stability, and wanting many features.  I proposed this change because I think it gets us a decent balance of these goals; there are other ways to do it -- we could break existing tests, we could make the test app and real app different, we could refuse to have the utility of hrefs, etc.  There may be a better way to accomplish the same thing (or a better balance of the goals).

If we do a major rewrite, I think we should use hrefs as routing wherever possible; it surprises the user less and builds in functionality of being able to restore state.

zachmullen: Thanks for the explanation. If this is mostly just to get our links to work as expected in test code, I would be more comfortable putting it just in testing code, though the risk of hiding bugs as a result is noted.

> If we do a major rewrite, I think we should use hrefs as routing wherever possible

Definitely agreed, I think all of us are on board with that. I'm particularly interested in how other frameworks handle the pattern of widgets that sometimes need to do routing and other times need to emit events when elements are clicked.

One bit of good news is that a rewrite would probably allow us to jettison jquery altogether, which would make this testing headache go away.

matthewma7: I briefly thought about this myself. I think new View framework will be able to hide complex logic inside a component so it's clean to use and looks clean like [this](https://reacttraining.com/react-router/web/api/Link), but internally it's doing something like [this](https://github.com/STRML/react-router-component/commit/9984d765d01d4e8e67d3da7eb39b035253e62925).

brianhelba: Indeed, both `react-router` and `vue-router`, do the "right thing", in that even when you use their native syntax for adding rotatable elements, they end up creating `<a href...`> elements in the DOM/

brianhelba:
I just discussed this with @manthey . It's my preference that this go into testing, since:
* it's essentially correcting a deficiency the way that we use jQuery to simulate user input for testing, rather than an actual issue that needs to be corrected in production
* it won't bloat the `App` class, which we reuse downstream in less predictable ways

In an ideal situation that may never fully happen, we'd be running two different versions of jQuery in testing: one internally imported by Girder's modules, and one that lives globally on `$` that's sourced by the test harness. In that case, it would be possible to actually modify the global `$`'s `.click` method to do whatever we want for compatibility and developer convenience, without worrying about breaking production code.

For now, @manthey will move the extra logic to testing, and I'll be happy to approve.

PR #2493, opened 2017-11-27
Improve the wait condition for the swagger test.

manthey:
Instead of waiting an arbitrary time, wait until the condition we need is satisfied.

There have been some intermittent failures of the swagger test on CI.  This may fix them if they are time dependent.

### DIFF
--- a/clients/web/test/spec/swaggerSpec.js
+++ b/clients/web/test/spec/swaggerSpec.js
@@ -6,10 +6,12 @@ describe('Test the swagger pages', function () {
         runs(function () {
             expect($('li#resource_system.resource .heading h2 a').text()).toBe('system');
         });
-        // There seems to be some delay between the link showing and when swaggerUi actually
-        // binds the event handler. We don't have a good hook into that binding, so we hack
-        // it with a 0.1s delay instead.
-        waits(100);
+        // There seems to be some delay between the link showing and when
+        // swaggerUi actually binds the event handler.  Wait until jquery
+        // reports that the event is bound
+        waitsFor(function () {
+            return $._data($('li#resource_system.resource .heading h2 a')[0], 'events') !== undefined;
+        }, 'events to be bound');
         runs(function () {
             $('li#resource_system.resource .heading h2 a').click();
         });


clients/web/test/spec/swaggerSpec.js
```diff
@@ -6,10 +6,12 @@ describe('Test the swagger pages', function () {
...
         });
-        // There seems to be some delay between the link showing and when swaggerUi actually
-        // binds the event handler. We don't have a good hook into that binding, so we hack
-        // it with a 0.1s delay instead.
-        waits(100);
+        // There seems to be some delay between the link showing and when
+        // swaggerUi actually binds the event handler.  Wait until jquery
+        // reports that the event is bound
+        waitsFor(function () {
+            return $._data($('li#resource_system.resource .heading h2 a')[0], 'events') !== undefined;
```
brianhelba: In general, I like to use `.first()` instead of `[0]` to get the first element of a set of selectors, as the former returns another jQuery selector while the latter returns a DOMElement (which often needs to be coerced back to a selector in order to be useful). The exception is obviously when you're using raw DOM APIs that can't operate on selectors.

In this case, it doesn't really matter, since jQuery will just do the appropriate thing with the first argument to `_data`, regardless of its type.

brianhelba: Per the [jQuery release notes](http://blog.jquery.com/2012/08/09/jquery-1-8-released/):
> $(element).data(“events”): In version 1.6, jQuery separated its internal data from the user’s data to prevent name collisions. However, some people were using the internal undocumented “events” data structure so we made it possible to still retrieve that via .data(). This is now removed in 1.8, but you can still get to the events data for debugging purposes via $._data(element, "events"). Note that this is not a supported public interface; the actual data structures may change incompatibly from version to version.

This isn't an ideal solution, since we're using a private unstable API. However, I'm fine with this for now, as if we ever upgrade jQuery and this breaks, it should fail totally and consistently, which is easy to detect and fix. This is certainly preferable to the status quo of nondeterministic failures.

brianhelba: As an interesting fact, it appears as though you *could* use `$.hasData` here too, as [that searches the private data object too](https://github.com/jquery/jquery/blob/3.2.1/src/data.js#L73). However, I would not trust that behavior to be stable either, and your solution is less magical and more likely to fail cleanly if private data changes in the future.

manthey: Oddly, jQuery requires the DOM element in the `_data` call, so `.first()` doesn't work here.

PR #2496, opened 2017-11-28
Fix complex plugin dependencies parsing.

manthey:
If plugins A and B are enabled, and plugin A depends on plugin C, and B depends on C and D, and D depends on E, the dependencies of plugin D are never analyzed if plugin A is examined before B (and, hence, plugin E won't be added).

Also, when the plugins are toggled via settings, log why other plugins are enabled.

### DIFF
--- a/girder/utility/plugin_utilities.py
+++ b/girder/utility/plugin_utilities.py
@@ -41,7 +41,7 @@ import traceback
 import pkg_resources
 from pkg_resources import iter_entry_points
 
-from girder import logprint
+from girder import logprint, logger
 from girder.constants import GIRDER_ROUTE_ID, GIRDER_STATIC_ROUTE_ID, PACKAGE_DIR, ROOT_DIR, \
     ROOT_PLUGINS_PACKAGE, SettingKey
 from girder.models.model_base import ValidationException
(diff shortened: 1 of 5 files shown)


manthey:
The important change is just converting a `return` to a `continue` in a routine that was last touched by @zachmullen two years ago.

girder/utility/plugin_utilities.py
```diff
@@ -122,7 +122,7 @@ def loadPlugins(plugins, root, appconf, apiRoot=None, buildDag=True):
     return root, appconf, apiRoot
 
 
-def getToposortedPlugins(plugins=None, ignoreMissing=False, keys=('dependencies',)):
```
danlamanna: Is there a reason `logNew` has to exist at all? Can we just always add the info log message?

girder/utility/plugin_utilities.py
```diff
@@ -122,7 +122,7 @@ def loadPlugins(plugins, root, appconf, apiRoot=None, buildDag=True):
...
+def getToposortedPlugins(plugins=None, ignoreMissing=False, keys=('dependencies',), logNew=False):
```
danlamanna: Is there a reason `logNew` has to exist at all? Can we just always add the info log message?

danlamanna:
Can you configure your editor to [wrap](https://chris.beams.io/posts/git-commit/#wrap-72) your commit messages?

girder/utility/plugin_utilities.py
brianhelba: Agreed, I think we should not clutter the business logic API with logging options. I think the best practice is to log everything (sending it to `DEBUG` level if you're concerned about spam under the default configuration) then control the output via filters in handlers.

brianhelba:
@manthey LGTM

PR #2498, opened 2017-11-29
Gracefully fallback when an assetstore type no longer exists.

manthey:
When a plugin supplying an assetstore fails to load, existing assetstores can't load the appropriate adapter.  We can't add capacity information, and that failure made it impossible to list the assetstore (including in the web client).  This falls back to adding unknown capacity information.

### DIFF
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -23,9 +23,9 @@ import datetime
 
 from ..describe import Description, autoDescribeRoute
 from girder.api import access
-from girder.api.rest import Resource, AccessException, filtermodel, setCurrentUser
+from girder.api.rest import Resource, filtermodel, setCurrentUser
 from girder.constants import AccessType, SettingKey, TokenScope
-from girder.exceptions import RestException
+from girder.exceptions import RestException, AccessException
 from girder.models.password import Password
 from girder.models.setting import Setting
 from girder.models.token import genToken, Token
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -23,8 +23,9 @@ import datetime
 
 from ..describe import Description, autoDescribeRoute
 from girder.api import access
-from girder.api.rest import Resource, RestException, AccessException, filtermodel, setCurrentUser
+from girder.api.rest import Resource, AccessException, filtermodel, setCurrentUser
 from girder.constants import AccessType, SettingKey, TokenScope
+from girder.exceptions import RestException
 from girder.models.password import Password
 from girder.models.setting import Setting
 from girder.models.token import genToken, Token
--- a/girder/models/assetstore.py
+++ b/girder/models/assetstore.py
@@ -21,6 +21,8 @@ import datetime
 
 from .model_base import Model, ValidationException, GirderException
 from girder.constants import AssetstoreType, SortDir
+from girder.utility import assetstore_utilities
+from girder.utility.abstract_assetstore_adapter import AbstractAssetstoreAdapter
 
 
 class Assetstore(Model):
@@ -31,8 +33,6 @@ class Assetstore(Model):
         self.name = 'assetstore'
 
     def validate(self, doc):
-        from girder.utility import assetstore_utilities
-
         # Ensure no duplicate names
         q = {'name': doc['name']}
         if '_id' in doc:
@@ -73,7 +73,6 @@ class Assetstore(Model):
         :type assetstore: dict
         """
         from .file import File
-        from girder.utility import assetstore_utilities
 
         files = File().findOne({'assetstoreId': assetstore['_id']})
         if files is not None:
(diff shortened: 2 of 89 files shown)


girder/models/assetstore.py
```diff
@@ -120,7 +120,16 @@ def addComputedInfo(self, assetstore):
         from .file import File
         from girder.utility import assetstore_utilities
 
-        adapter = assetstore_utilities.getAssetstoreAdapter(assetstore)
+        try:
+            adapter = assetstore_utilities.getAssetstoreAdapter(assetstore)
+        except GirderException as e:
```
brianhelba: I think the right thing to do here is to subclass `GirderException`, and only catch instances of that subclass. That won't break the existing behavior, since the subclass still is an instance of `GirderException`.

girder/models/assetstore.py
```diff
@@ -120,7 +120,16 @@ def addComputedInfo(self, assetstore):
...
+            if e.identifier != assetstore_utilities.EXCEPTION_ID_NO_ADAPTER:
+                raise
+            # If the adapter doesn't exist, use the abstract adapter, since
+            # this will just give the default capacity information
+            from girder.utility.abstract_assetstore_adapter import AbstractAssetstoreAdapter
```
brianhelba: Why is it necessary to do the import at runtime here, instead of at import-time of this module?

girder/models/assetstore.py
manthey: No where else do we subclass `GirderException`, which surprised me.  Oddly, we have this concept of an identifier in that exception (which, clearly, the subclass could use).

girder/models/assetstore.py
manthey: The existing code imported at runtime.  It looks like @zachmullen did that for the File import (is that necessary, since we don't invoke it until called?).  We shouldn't need to do that for the assetstore utilities or the abstract assetstore; I'll change those.

girder/models/assetstore.py
brianhelba: That sounds like a good idea. The subclass should be able define its identifier at the class-level, rather than in every constructor call.

IMO, we should drop the identifiers entirely, but we can always do that globally in a future pass, with more developer consensus.

girder/utility/assetstore_utilities.py
```diff
@@ -30,6 +31,16 @@
 }
 
 
+class GirderNoAssetstoreAdapterException(GirderException):
```
brianhelba: This is an honest question: would `NoAssetstoreAdapterGirderException` make more sense as a name?

danlamanna:
> This is an honest question: would `NoAssetstoreAdapterGirderException`
> make more sense as a name?

girder.exceptions.NoAssetstoreAdapter?

zachmullen:
exceptions.no.girder.Assetstore.Exception?

manthey:
I like @danlamanna's suggestion; we could move the `GirderException`, `AccessException`, and `ValidationException` to an `exceptions` module (and we'd need to expose them in `models.model_base` so as not to break existing code).  But I'm lazy, so I won't change things unless there is some consensus.

manthey:
@danlamanna Which do you like better of the two names you used (`girder.exceptions.NoAssetstoreAdapter` or `girder.exceptions.AssetstoreAdapterMissing`)?

I'll have a commit with the new `exceptions` module shortly.

danlamanna:
> @danlamanna Which do you like better of the two names you used
> (`girder.exceptions.NoAssetstoreAdapter` or
> `girder.exceptions.AssetstoreAdapterMissing`)?

I think girder.exceptions.NoAssetstoreAdapter.

manthey:
This makes the exceptions `girder.exceptions.GirderException`, `girder.exceptions.AccessException`, `girder.exceptions.ValidationException`, and `girder.exceptions.NoAssetstoreAdapter`.

girder/utility/path.py
```diff
@@ -21,20 +21,13 @@
 
 import re
 from ..constants import AccessType
-from ..models.model_base import AccessException, GirderException, ValidationException
+from ..exceptions import AccessException, GirderException, ValidationException
+from ..exceptions import ResourcePathNotFound as NotFoundException
```
brianhelba: There are still a number of statements that `raise NotFoundException` in this file. These should all be changed to `raise ResourcePathNotFound`, so that viewers of the code aren't confused about what the canonical name of the exception is.

I expect that you should then have a `from ..exceptions import ResourcePathNotFound` statement and a global `NotFoundException = ResourcePathNotFound` statement at the top, with a Flake8 ignore directive on the latter, since it shouldn't be used, but will just make the symbol available in this module's namespace for compatibility.

To be clear, I think using the `from ... import ... as ...` would be fine if it's only used for symbol definition compatibility. If the symbols are actually being referenced within the code, I think we need to consistently use the new name,

manthey: Agreed.

manthey:
As I had mentioned in my previous commit message, I did not change all of the references to importing `RestException` to import from the `exceptions` modules, as it is still accessible from the `api.rest` module and this would be a fairly extensive change.  I wanted to make sure that others could object if they didn't like that.  I'll change it (since @brianhelba just recommended doing so).

girder/api/v1/user.py
```diff
@@ -23,8 +23,9 @@
 
 from ..describe import Description, autoDescribeRoute
 from girder.api import access
-from girder.api.rest import Resource, RestException, AccessException, filtermodel, setCurrentUser
+from girder.api.rest import Resource, AccessException, filtermodel, setCurrentUser
```
brianhelba: `AccessException` here should be imported from `girder.exceptions`.

PR #2502, opened 2017-12-01
Add the pytest tests to the ctest test list.

manthey:
This runs the pytest tests as one of the ctest tests so that the entire project can still be tested with a single command and that coverage is collected together.

I recognize that we want to move away from CMake eventually, but until we do, I think that we need to maintain it.

This is marked as a work-in-progress because it will currently run the pytest tests twice in Circle.  Either we can remove the special pytest code there, or we can leave what is there and change how we invoke ctest.  Opinions?

### DIFF
--- a/cmake/run_circleci.sh
+++ b/cmake/run_circleci.sh
@@ -20,6 +20,8 @@ esac
 
 mkdir $HOME/build
 
+ctest -R '^py_coverage_reset$' -VV -S $HOME/girder/cmake/circle_continuous.cmake
+
 if [ "$TEST_GROUP" == "python" ]; then
     if ! pytest --tb=long --junit-xml=$CIRCLE_TEST_REPORTS/pytest-${CIRCLE_NODE_INDEX}.xml; then
         touch $HOME/build/test_failed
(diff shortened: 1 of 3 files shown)


kotfic:
Sorry,  i'm having a hard time following - what exactly is the purpose of running pytest through ctest? If it's to collect coverage,  isn't that an artifact of using ctest to run coverage? Shouldn't we not be running side-effect based non-tests like coverage combined/report in ctest?

As-is, this obfuscates all of pytest's options making it essentially unusable.  Do you see people running tests separately from ctest locally? Is this intended to be run just on CI?

manthey:
When examining a branch (or my own code), I often want to run all the tests locally -- not just the tests that are specifically related to the branch.  I want to see if any tests fail and what coverage exists for the branch.  And, ideally, it should be easy to run this in a canonical way.

Prior to the pytest addition, I could run "ctest --output-on-failure" and when it was done look at the coverage report.  After the pytest addition, I can't.  This makes it so I can again.

I don't much care what method we have to get a simple way to run all the tests and get all the coverage., but it needs to exist.  Running `ctest -R '^py_coverage$'; pytest; ctest -E '^py_coverage$'` probably works, but seems obtuse.

manthey:
Per an offline discussion, @danlamanna asked if we could keep the CI running pytest explicitly rather than through ctest to ensure that it is easy to run by itself (rather than just within the environment set up by cmake).

With the last commit, this is being done.

PR #2504, opened 2017-12-05
System setting to disable logging in with a password

zachmullen:
@manthey PTAL

### DIFF
--- a/clients/web/src/templates/body/systemConfiguration.pug
+++ b/clients/web/src/templates/body/systemConfiguration.pug
@@ -102,6 +102,13 @@ form.g-settings-form(role="form")
               type="checkbox", checked=(
                 settings['core.api_keys'] === null ? defaults['core.api_keys'] : settings['core.api_keys']))
           | Enable API keys
+    .form-group
+      .checkbox
+        label(for="g-core-enable-password-login")
+          input#g-core-enable-password-login(
+              type="checkbox", checked=(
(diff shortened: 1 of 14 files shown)


manthey:
If you have no oauth providers enabled and you disable password login, there are no login methods.  Should we guard against that in any way?

jeffbaumes:
@manthey It would seem difficult to know whether another login mechanism (could be oauth or something else in the future) was enabled through a plugin, so I'd suggest leaving it to the admin to do the right thing.

manthey:
Then, I'll assume that if an admin turns off all login methods for all admins, that someone has access to the database so they can turn them back on the hard way.

brianhelba:
@jeffbaumes @manthey Apropos to administrators managing to lock themselves out of their own Girder instance, note that we currently have [a block of code which verifies and enables a user if they're the only user in the instance](https://github.com/girder/girder/blob/master/girder/models/user.py#L124). Since the default setting is to not require verification / enablement, this is another case of not trusting / expecting admins to "to the right thing".

I'm generally uncomfortable with putting in settings that allow admins to break their own instance. Can we add some additional guards to prevent this case, or warn users?

Also, I'd like to look further into how we're passing public settings. Please hold on merging until I have a chance to review further.

danlamanna:
Can we add a changelog entry under added features?

PR #2507, opened 2017-12-06
Screenshots in circle

zachmullen:
@girder/developers PTAL

This PR is organized into two distinct parts:

* 13705e6 makes it so that our phantom screenshots get submitted to CircleCI as artifacts, as can be seen [here](https://circleci.com/gh/girder/girder/3992#artifacts/containers/2).
* db5ff87 removes CDash submission from our CI process, meaning we no longer use CDash at all during CI. (Screenshots were really the last thing we were using it for since everything else can be found on the CircleCI dashboard). If this is too controversial, we can revert this commit.


brianhelba:
I support this change, but I'd prefer if we implement this on top of https://github.com/girder/girder/pull/2500 , which is close to being completed. The artifact submission process is different in CircleCI v2, and that PR also moves the screenshots to be output in their own directory. Both of these factors will cause merge conflicts between this PR and https://github.com/girder/girder/pull/2500 .

zachmullen:
@brianhelba you're right.

@manthey is #2500 going to land in the next few days? Would you like me to try and push this change onto that topic, or should I just wait?

manthey:
I think #2500 is nearly ready.  I'd recommend waiting until it is merged (or basing a PR off of it) rather than adding to it.

zachmullen:
Sounds good, I'll wait for merge. It looks like you may have already done something on that topic for getting screenshots as artifacts, is that correct?

manthey:
@brianhelba changed the code to output the screenshots into a build/test/artifacts directory.  We store these artifacts from the workflow steps that run any web tests.  I suppose we should force a failure to make sure we get the screenshot.

brianhelba:
I've rebased this on CircleCI2, and dropped the first commit to store screenshots as artifacts, as that's already been done in https://github.com/girder/girder/pull/2500/commits/3acb38d8d1caa4abfb96fffca32851eee0ba0be3 .

zachmullen:
Need a toBe(true) in there

zachmullen:
We're running the web client test in 3 different builds on Circle, is that intentional?

brianhelba:
Yes, anything that's an "integrationTest" is something special that blurs the lines of well-behaved client or server tests, and that includes a Javascript-based upload test.

brianhelba:
@zachmullen I'll leave it to you to explore the way artifacts are uploaded. You may want to adjust the documentation based on how you experience it as a "new" user.

Feel free to pop off my final commit when you're satisfied.

zachmullen:
Fair enough, but then why do they also run in webBuild_webTest?

zachmullen:
Not a big deal, but it takes 7 clicks to get down to the screenshot listing from a failed build due to the deeply nested directory structure:

screen shot 2017-12-12 at 5 16 52 pm

Is there a way to make them appear in a higher-level artifact directory? If not it's fine.

zachmullen:
Also where does the web_client_data_s3 test run?

brianhelba:
Agreed, the 7 clicks is annoying, but I'd consider it a UI bug within CircleCI. We could submit a ticket and ask them to address it. I'd oppose making hask-ish changes to our directory layout right now, since we still have near-term plans to add a lot more content to the `build/` directory.

brianhelba:
If it doesn't have `LABELS girder_integration` and it's a web client test, then it'd run in the `py3_webBuild_webTest` job.

zachmullen:
I see it now, sorry for the noise.

zachmullen:
I approve, but cannot technically mark as approved.

PR #2508, opened 2017-12-07
Port several more tests to pytest

zachmullen:
@danlamanna @jbeezley let me know what I'm doing wrong, I'll try and port a few here


mgrauer:
+1 to aggressively promoting the pytest way, others can watch these PRs to learn and discuss best practices, so do not be shy.

Question: is our coverage in CI set up to show the coverage diffs accurately on these test migration PRs?

pytest_girder/pytest_girder/fixtures.py
```diff
@@ -17,7 +19,7 @@ def bcrypt():
         yield hashpw
 
 
-@pytest.fixture
+@pytest.fixture(scope='module')
```
jbeezley: By sharing this database fixture across modules, you are effectively sharing state between test functions.  This is something we are actively trying to avoid not just for increased parallelism, but to enhance the quality and debug-ability of our tests.

I'm actually somewhat surprised this even works with the current tests.  It makes me think that I might not actually understand what's going on.  :confused:

zachmullen: I'm going to be doing my lightning talk tomorrow about this very subject, hoping it leads to some design decisions. The current implementation of the `db` fixtures is already a little problematic, but there are engineering tradeoffs to consider either way.

Function-scope fixtures may depend on module-scope fixtures, but not the other way around.

The reason I think this works is because our other fixtures clean up their documents on teardown.

jbeezley: Fixtures clean up their documents, but tests do not.  I'm on the side of explicitly making all test functions independent even at the cost of execution time.  It sounds like something we can discuss after your lightning talk tomorrow.

zachmullen:
Let's have this religious war tomorrow where others can weigh in. I'm going to remove 7bdaa3f from this branch, I mostly pushed it to demonstrate a point, namely that as we add more granularity, we significantly increase the overall runtime of the system, and that overhead bottleneck is due almost entirely to the `db` fixture.

I'm going to issue a straw man proposal tomorrow that for CI, we simply use `--mock-db`, which alleviates this bottleneck, and which would make us feel much more free to go down the rabbit hole of high testing granularity. I'm hoping that idea is amenable to those who share @jbeezley 's religious preference for unit testing over larger integration-style testing. @girder/developers please consider this question in advance of tomorrow's talk if you are interested in this subject.

jbeezley:
Yeah, I'm in favor of `--mock-db`.  I originally had it enabled by default, but @danlamanna thought that it would be a more controversial topic that we could address later.

zachmullen:
Removed WIP tag, we can cut this branch off here. I ported 4 tests over, and in the process added two new core fixtures that should be scrutinized in review:

* `fsAssetstore`: a filesystem assetstore fixture that creates an assetstore in a uniquely named temp dir. This assetstore does not clean itself up yet, since I didn't know if it would be worthwhile to try and remove all files within it.
* `smtp`: the SMTP server used in testing mail functionality. I don't like how this is implemented due to its not-strictly-necessary dependence on `server`, but I am comfortable with its interface. This mock SMTP server is also the source of those "bad file descriptor" errors, but after a good bit of tinkering I still couldn't get those to go away, I still don't know if we are misusing asyncore + threading or if this is a bug internal to asyncore or smtpd.

pytest_girder/pytest_girder/utils.py
```diff
@@ -1,10 +1,115 @@
...
+
 from six import BytesIO
-from six.moves import urllib
+from six.moves import queue, range, urllib
+
+_startPort = 31000
+_maxTries = 100
+
+
+class MockSmtpServer(smtpd.SMTPServer):
```
danlamanna: I had a random stash or branch somewhere (that I can't find now) that was almost able to remove this in favor of [pytest-localserver](https://pypi.python.org/pypi/pytest-localserver/). It's probably out of scope for this, but we should investigate it at some point.

girder/test/test_size.py
```diff
@@ -0,0 +1,136 @@
...
+    File().save(file2, validate=False)
+
+    yield Hierarchy(collections=(c1, c2), folders=(f1, f2), items=(i1, i2), files=(file1, file2))
+
+    Collection().remove(c1)
+    Collection().remove(c2)
+
+
+def assertNodeSize(resource, model, size):
+    model = model().load(resource['_id'], force=True)
```
jbeezley: Add `__tracebackhide__ = True` to fix the traceback for assertion functions.

https://docs.pytest.org/en/latest/example/simple.html#writing-well-integrated-assertion-helpers

girder/test/test_size.py
```diff
@@ -0,0 +1,136 @@
...
+    assertStatusOk(resp)
+    assertNodeSize(admin, User, 10)
+    assertNodeSize(hierarchy.collections[1], Collection, 1)
+
+    # Delete the sub folder
+    resp = server.request(
+        path='/folder/%s' % hierarchy.folders[1]['_id'], method='DELETE', user=admin)
+    assertStatusOk(resp)
+    assertNodeSize(admin, User, 0)
+    assertNodeSize(hierarchy.collections[1], Collection, 1)
```
jbeezley: I would prefer these functions be split into multiple tests that make assertions on a single rest call.

girder/test/test_rest_utils.py
```diff
@@ -0,0 +1,148 @@
...
+def testRequireParamsDictMode():
+    resource = rest.Resource()
+    resource.requireParams('hello', {'hello': 'world'})
+    resource.requireParams('hello', {'hello': None})
+
+    with pytest.raises(rest.RestException, match='Parameter "hello" is required.$'):
+        resource.requireParams(['hello'], {'foo': 'bar'})
+
+    with pytest.raises(rest.RestException, match='Parameter "hello" is required.$'):
+        resource.requireParams(['hello'], None)
```
jbeezley: Maybe parametrize this?

girder/test/test_size.py
zachmullen: Is there a rationale for this?

pytest_girder/pytest_girder/utils.py
zachmullen: That would be awesome, it does claim:

> provides a threaded instance of smtpd.SMTPServer runnning on localhost.

I can look into that in a future PR if you can't find your code.

girder/test/test_size.py
jbeezley: I think the idea of testing multiple state transitions in a single test is counter to what a unit test should be.  While what is done here is not terrible, it sanctifies the idea that you could add another rest request in here to cover another behavior.  My fear is that this will lead to the several hundred line test cases that is common in the old tests.

Perhaps our expectations of what migrating to this new infrastructure means.  I fully expect that many of the test cases will have to be completely rewritten.  I don't think we can expect to retain the exact assertions that existed before.

zachmullen: Makes sense.

pytest_girder/pytest_girder/fixtures.py
```diff
@@ -129,3 +155,47 @@ def user(db, admin):
...
+    from girder.constants import ROOT_DIR
+    from girder.utility import plugin_utilities
+
+    path = os.path.join(ROOT_DIR, 'tests', 'test_plugins')
+    old = plugin_utilities.getPluginDir
+    plugin_utilities.getPluginDir = mock.Mock(return_value=path)
+
+    yield path
+
+    plugin_utilities.getPluginDir = old
```
jbeezley: We haven't solved the problem of testing plugins yet.  I hesitate to bake in the assumption that plugin tests exist inside girder's package.  For girder 3, plugin tests will be include inside each plugin's package.  What we do before then is up in the air.  Because this isn't used yet, could you remove it from this PR?

zachmullen: This isn't for testing plugins, it's for loading the fake plugins that we use in core testing. Its only contractual guarantee is that it will cause plugins to be loaded from the fake directory instead of the real one.

zachmullen: (Also it's used currently in the mail test)

jbeezley: Okay, I see.  We'll probably need to revisit this for the pip installable plugins changes, but for now I think it's okay.

zachmullen: Yeah, since it's just for testing core and not plugins, I don't care if we need to break it later since it'll just be us having to deal with that.

zachmullen:
PTAL, I broke the size_test up to be 1 state change per test.

girder/test/test_size.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#  Copyright 2013 Kitware Inc.
```
jbeezley: What is our policy on updating this dates?

zachmullen: No idea... I'd love to remove this header from all the files

manthey: I thought we decided the date should be in just one file and we could remove it elsewhere.

jbeezley: Okay, we can take care of that elsewhere.

PR #2510, opened 2017-12-11
[RFC] Mock mongo database in CircleCI

zachmullen:
This follows from our discussion on Friday. @girder/developers PTAL

### DIFF
--- a/cmake/run_circleci.sh
+++ b/cmake/run_circleci.sh
@@ -23,7 +23,7 @@ mkdir $HOME/build
 ctest -R '^py_coverage_reset$' -VV -S $HOME/girder/cmake/circle_continuous.cmake
 
 if [ "$TEST_GROUP" == "python" ]; then
-    if ! pytest --tb=long --junit-xml=$CIRCLE_TEST_REPORTS/pytest-${CIRCLE_NODE_INDEX}.xml; then
+    if ! pytest --mock-db --tb=long --junit-xml=$CIRCLE_TEST_REPORTS/pytest-${CIRCLE_NODE_INDEX}.xml; then
         touch $HOME/build/test_failed
     fi
 fi


zachmullen:
I agree this may be premature. I'm going to mark it an RFC and leave it up for discussion.

brianhelba:
@zachmullen If you change this to run `pytest` twice, then I'd be comfortable with merging this right now, so we can start to gather empirical validation of MongoMock, which may be useful in a future decision to switch completely.

zachmullen:
I'm trying to make it run faster, not slower ;)

I think in the end we'll want to do mongomock on topic branches, and non-mocked on master.

brianhelba:
Per @manthey 's profiling, it looks like the current runtimes are about:
pytest in master: 116.64 seconds
pytest --mock-db: 6.99 seconds

I'd be fine with adding another 7 second test right now, particularly if it gives us more empirically-derived confidence in MongoMock. In the CircleCI 2 branch #2500 , it won't even decrease the overall turnaround time of the test suite.

zachmullen:
116.64 seconds? Not sure where that number came from, but [the latest master build](https://circleci.com/gh/girder/girder/4053) pytest only took 19.63 seconds against real mongodb. Are you talking about on CircleCI 2?

brianhelba:
It looks like Mongomock is already broken, due to the changes in https://github.com/girder/girder/pull/2512/commits/19f75242143439de0f8026e315c4fab171b00489 , which handles [a deprecation in a newer Pymongo](http://api.mongodb.com/python/current/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient.get_default_database).

If we want to move forward with mocking, I think we'll have to assume maintainership of [the Mongomock repo](https://github.com/mongomock/mongomock). They are actively looking for maintainers, and it appears as though they've already got a backlog of legitimate PRs from other users. @girder/developers  is this something you think we should try to do?

jeffbaumes:
I would be ok with this since Girder maintainers collectively already have a reasonable understanding of mongo (and need to maintain that understanding in the reasonable future), and this repo is only in the low-hundreds of Python LOC.

Keeping mongo mocking working potentially saves us several minutes of waiting per PR (and in local test runs) so we are overall saving time as long as we don't need to invest tons of time into mongomock.

It also requires someone to champion the maintenance. Is someone willing to take this on and reach out to the current maintainers?

zachmullen:
I went in and looked through the mongomock repository, it didn't scare me off. It's not a ton of code, it's well structured and easy to understand. That makes me comfortable using it in CI, but I would only want to become a nominal maintainer if that granted us the power to merge PRs (in accordance with their rules) and push updated packages to pypi. Do we have a sense if that would be possible?

zachmullen:
I'd also be comfortable leaving mongomock disabled during our CI and just pushing the ramdisk pathway since that still tests real mongo and alleviates a lot of the time cost.

zachmullen:
> The ramdisk stuff is already enabled now on CircleCI 2

I see that for our mongo 3 containers we are using the ramdisk version of the image, but is that the case for the py2/mongo2 steps?

brianhelba:
MongoDB 2.6 is too old to have a [ramdisk image that's provided by CircleCI](https://hub.docker.com/r/circleci/mongo/tags/), so we just use the upstream image.

However, I'd like to move the test with MongoDB 2.6 to use MognoDB 3.2 (which does have a ramdisk image) instead, per #2516. Feedback on this would be welcome, via that issue.

zachmullen:
Circling back to mongomock, I've fixed the issue and opened an upstream PR https://github.com/mongomock/mongomock/pull/365.

Hopefully that will be released in short order, if not we can always consider forking...

brianhelba:
@zachmullen I see a number of other easy-fix PRs open at Mongomock, so I suspect that the project really is unmaintained. However, I'm not sure if we need to fork it quite yet. Have you contacted the lead developer directly and discussed assuming maintainership of that repo / project?

zachmullen:
I pinged the person that it says to ping on the front page. No response yet.

Would it be better to change our mongomock package dependency to a git reference, or push it to pypi under a different name?

zachmullen:
I'm going to close this and move the discussion over to #2516 since I think the ramdisk is the way to go for CI.

zachmullen:
@girder-developers Just to follow up on this, https://github.com/mongomock/mongomock/pull/365 did get merged, but it's unclear when it will be released (last release was March 2017). Would it be useful / OK to point to a git URI rather than a pypi release for mongomock?

PR #2511, opened 2017-12-11
Improve the interface for loading test plugins in core tests

zachmullen:
@jbeezley PTAL, I've changed how we load the fake plugins we use in core testing. Since plugins won't load without the server being setup, I've moved the logic into the `server` fixture and made it so that `pytest.mark.testPlugins(['my', 'plugins'])` is used to indicate what test plugins should be loaded. I think this could also be a good syntax for loading real plugins in the future.


jbeezley:
I can't think of any specific problems with this approach.  I would want @danlamanna to take a look before merging.

danlamanna:
> I can't think of any specific problems with this approach.

I can't either, though I think there are still some unknowns with how Girder 3 will deal with plugin testing.

Whole modules can be marked too by setting a module-level `pytestmark` variable.

PR #2512, opened 2017-12-12
Update CircleCI primary Docker image

brianhelba:
This updates CircleCI to use an auto-built base image, to add deployment of master, and to test against the latest Python 3.5. This also fixes several Girder bugs with Python 3.5.

* Use the auto-built Docker ["girder/girder_test" image](https://hub.docker.com/r/girder/girder_test/) for CI
* Add a workflow step to deploy master
* Test Python 3.5 as the latest version
* Make MockSmtpReceiver compatible with Python 3.5
* Use PyMongo >3.5 to ensure support for latest MongoDB servers
* `json.loads` returns a `JSONDecodeError` in Python 3.5


brianhelba:
@manthey @girder/developers PTAL.

devops/ansible/roles/girder/library/girder.py
```diff
@@ -896,7 +897,7 @@ def unjsonify(a):
         try:
             a = json.loads(a)
             # pass-through to below
-        except ValueError:
+        except json.JSONDecodeError if sys.version_info >= (3, 5) else ValueError:
```
manthey: This isn't necessary, as a `json.JSONDecodeError` is a subclass of `ValueError`.

girder/models/model_base.py
```diff
@@ -90,7 +90,7 @@ def reconnect(self):
         typically not have to call this method.
         """
         db_connection = getDbConnection()
-        self.database = db_connection.get_default_database()
+        self.database = db_connection.get_database()
```
manthey: I see you've increased the pymongo version to allow for this (since it won't work with pymongo < 3.4).  Is there any other benefit to a newer pymongo?

plugins/item_tasks/server/__init__.py
```diff
@@ -40,8 +41,8 @@ def _onUpload(event):
     they are used to link item task outputs back to a job document.
     """
     try:
-        ref = json.loads(event.info.get('reference'))
-    except (ValueError, TypeError):
+        ref = json.loads(event.info.get('reference', ''))
+    except json.JSONDecodeError if sys.version_info >= (3, 5) else ValueError:
```
manthey: What happened to the `TypeError` catching?

tests/cases/find_entry_point_plugins_test.py
```diff
@@ -112,7 +113,9 @@ def resource_stream_json_value():
         failures = getPluginFailureInfo()
         self.assertIn('entry_point_plugin_bad_json', failures)
         self.assertIn('traceback', failures['entry_point_plugin_bad_json'])
-        self.assertIn('ValueError', failures['entry_point_plugin_bad_json']['traceback'])
+        self.assertIn(
+            'JSONDecodeError' if sys.version_info >= (3, 5) else 'ValueError',
```
manthey: This is probably the only spot where you need to differentiate `JSONDecodeError` from `ValueError`.

.circleci/config.yml
```diff
@@ -4,7 +4,7 @@ jobs:
   py2_serverInstall_serverTest:
     docker:
       # This image uses the oldest version of many dependencies
-      - image: girder/girder_test_py2:latest
+      - image: girder/girder_test:latest-py2
```
manthey: Could we tag the images consistently?  If one is `latest-3.5`, this one should be `latest-2.7` (or the other should be `latest-py3`).

.circleci/config.yml
```diff
@@ -313,9 +311,8 @@ jobs:
       - run:
           name: Set up Python 3.5
           command: |
-             # 3.5.2 is already installed in circle, so use it
-             pyenv install 3.5.2 || true
-             pyenv global 3.5.2
+             pyenv install 3.5.4 || true
```
manthey: Circle has 3.5.2 installed by default.  Using 3.5.4 will cause it to download the version of python, which will slow down the build.  Is there a need to test with 3.5.4 instead of 3.5.2?

.circleci/config.yml
brianhelba: They will be consistent when this branch is merged, but there is a final "WIP" commit on this branch, which is necessary to QA it (since otherwise this would run against the old Docker image fromt master).

The actual image names will be `girder/girder_test:latest-py2` and `girder/girder_test:latest-py3`.

devops/ansible/roles/girder/library/girder.py
jcfr: While true, I see no arm in catching that specific exception. It would also be consistent with how other calls to json.loads are handled.

.circleci/config.yml
brianhelba: I thought about this, and it's to be consistent with the Docker jobs where we test the "latest" Python 3.5, which is 3.5.4. Given that we've actually seen Girder code break across patch versions of Python (e.g. the hashsum serialization logic), I thought it was significant enough to care about being consistent between our two (Docker vs machine) infrastructures. Once we move to testing Python 3.6, we'll have to install a different version here anyway.

devops/ansible/roles/girder/library/girder.py
brianhelba: Yeah, my intent was to capture the most specific exception possible, but perhaps this is overkill. I do anticipate dropping support for Python 3.4 at some point in the medium-term, at which point `sys.version_info >= (3, 5)` would simplify to `six.PY3`.

I'm on the fence, so if you think simplicity is more important, I'd be fine with dropping most of this change.

girder/models/model_base.py
brianhelba: First-class compatibility with `mongod` 3.6.

plugins/item_tasks/server/__init__.py
brianhelba: If it exists, `'reference'` will always be a string (since [it came from the HTTP request](https://github.com/girder/girder/blob/master/girder/models/upload.py#L254)). The only case a `TypeError` should be emitted would be if the key was not found, and `None` was returned by default. By returning `''` by default instead, a `TypeError` *should* no longer occur.

Indeed, if a `TypeError` does occur in the future, it means that one of our preconditions around the `'reference'` data type was violated, so we don't want to blindly suppress the `TypeError`.

tests/cases/find_entry_point_plugins_test.py
brianhelba: Maybe, see above.

devops/ansible/roles/girder/library/girder.py
manthey: It is ugly and hard to maintain to have each spot where we try to catch JSON decoding errors have this sort of conditional.  If we want the specificity, then, much like six, we should have an abstracted JSONDecodeError which we can catch instead (where we'd define it to be this expression and then individual sections of code can just catch the abstracted error).

brianhelba:
@manthey PTAL again. Once you approve, I'll pop the final "WIP" commit (then you'll need to re-approve), then I'll merge.

brianhelba:
I've only modified the "json.loads returns a JSONDecodeError in Python 3.5" commit. Everything else is the same.

brianhelba:
@girder/developers Merging this now.

The tests are technically failing on this branch, as it references version of the Docker image that doesn't yet exist. Once this is in master, the image will be auto-built, and everything should work going forward.

PR #2514, opened 2017-12-13
Run pytest separately from ctest in ci.

manthey:


.circleci/config.yml
```diff
@@ -38,13 +38,19 @@ jobs:
...
             TEST_GROUP: python
             PYTHON_VERSION: 2.7
             PYTHON_EXECUTABLE: /home/circleci/project/girder_env/bin/python
             VIRTUALENV_EXECUTABLE: /home/circleci/project/girder_env/bin/virtualenv
           working_directory: girder_build
+      - run:
+          name: Run pytest
+          command: pytest --cov-append --tb=long --junit-xml="$CIRCLE_WORKING_DIRECTORY/girder/build/test/artifacts/pytest-2.7.xml"
+          working_directory: girder
+      - store_artifacts:
```
brianhelba: Why are you treating this XML output as an artifact? Shouldn't it be stored in `build/test/coverage`, and included in the meta-report that's submitted to `codecov`?

If we want to have a coverage "artifact", I think it should be an HTML summary of coverage.

danlamanna:
> Why are you treating this XML output as an artifact? Shouldn't it be
> stored in `build/test/coverage`, and included in the meta-report
> that's submitted to `codecov`?

I asked for it to be treated as an artifact for its (eventual) use in
Circle, consistent with the usage in Circle 1.

> If we want to have a coverage "artifact", I think it should be an HTML
> summary of coverage.

Is there a reason not to have both?

PR #2517, opened 2017-12-13
Python coverage cleanup in CircleCI

brianhelba:
* Attempt to submit pytest results to CircleCI for analysis
* Remove the PYTHON_COVERAGE CMake option
  * Python coverage will now always run. It was already always being run for Pytest tests.
* Remove the PYTHON_BRANCH_COVERAGE CMake option
  * This option should be set directly the Coverage config file (which is the only way to set it for Pytest).
* Remove the PYTHON_COVERAGE_CONFIG CMake option and unify Coverage configs
  * Both the legacy CMake+unittest and Pytest runners will now use the same .coveragerc file.
* Always write coverage reports to "build/test/coverage"
  * This ensures that Pytest-generated coverage reports are written to the same place as unittest / CMake-generated reports.
* Write intermediate Python coverage files inside "build/test/coverage"
* Reorganize the root .gitignore file                                                                                                                                                      
  * This also stops ignoring the "htmlcov" directory, which should not longer be generated by Javascript coverage.


manthey:
This LGTM, if @danlamanna is content with not having the junit xml as an artifact.

.coveragerc
```diff
@@ -1,5 +1,12 @@
 [run]
+data_file = build/test/coverage/server_temp/.coverage
 omit =
     girder/external/*
     girder/test/*
 parallel = True
+
+[html]
+directory = build/test/coverage/server_html
```
danlamanna: `server.html`

.gitignore
```diff
@@ -1,23 +1,27 @@
...
+.vagrant
+*.py[cod]
+**/*.retry
+
+# IDE / OS generated
+.DS_Store
+.idea/
+.PyCharm50/
+.*.swo
+.*.swp
```
zachmullen: Not really related to this PR, but we should remove all the platform specific stuff from this file.

danlamanna:
> Not really related to this PR, but we should remove all the platform
> specific stuff from this file.

Are you talking about the editor files?

zachmullen:
Talking about

```
# IDE / OS generated
.DS_Store
.idea/
.PyCharm50/
.*.swo
.*.swp
```

brianhelba:
@girder/developers PTAL. This is now ready for review (and I've updated the PR description with the commit messages).

manthey:
There are downstream plugins that were using the `PYTHON_COVERAGE_CONFIG` and `PYTHON_BRANCH_COVERAGE` options.  The ones I know of can do without these.  Do we need a migration guide?

danlamanna:
@brianhelba I pushed a few commits after what we discussed.

pytest.ini
```diff
@@ -1,3 +1,4 @@
 [pytest]
 addopts = --verbose --showlocals --cov=girder --cov-append --cov-config=.coveragerc --cov-report=""
+cache_dir = build/test/cache/pytest
```
brianhelba: @danlamanna Is there a reason this needs to be nested, and can't be `pytest_cache`? Are there other test caches we're planning on supporting?

brianhelba:
@manthey Can you point to the downstream usages `PYTHON_COVERAGE_CONFIG` and `PYTHON_BRANCH_COVERAGE`?

As a short note (which we can add to the changelog): if you want a custom coverage config file or want to enable branch coverage, just define your own `coverage combine` test, and pass the config file as an option.

Of course, in the future, with Tox, this will likely be even easier, but at the moment, the presence of these variables was blocking the unification of the Pytest and CMake/unittest coverage configs. Also in the future (once these changes make it easier to quickly generate and view details on coverage reports), I plan to re-evaluate the default setting for Python branch coverage and see if enabling it would suit Girder.

danlamanna:
> @danlamanna Is there a reason this needs to be nested, and can't be
> `pytest_cache`? Are there other test caches we're planning on
> supporting?

No, fixed.

brianhelba:
@danlamanna PTAL @ one final commit from me (rationale in the commit message).

Otherwise, I'd consider this PR to be complete. I want to wait for @manthey 's response before a merge though.

tests/PythonTests.cmake
```diff
@@ -114,21 +101,16 @@ function(add_python_test case)
...
-      WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
-      COMMAND "${PYTHON_EXECUTABLE}" -m unittest -v ${module}
-    )
-  endif()
+  file(MAKE_DIRECTORY "${PROJECT_SOURCE_DIR}/build/test/coverage/python_temp")
+
+  add_test(
+    NAME ${name}
+    WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
+    COMMAND "${PYTHON_COVERAGE_EXECUTABLE}" run
```
manthey: Do we need the `--parallel-mode` flag to make distinctly named coverage file?

manthey:
@brianhelba Here is one: https://github.com/girder/slicer_cli_web/blob/master/.travis.yml (just in its travis build) -- I am certain that this was cargo-cult copied from somewhere else, and adds no value, but my fear is that someone is using these features in a more serious way.

How would you pass a custom rc file to the python coverage command (or how does it pick up a custom config file)?  I guess I have to read the internals of the coverage command to know.

tests/PythonTests.cmake
brianhelba: This option is now hardcoded into the `.coveragerc` file.

.coveragerc
```diff
@@ -1,5 +1,12 @@
...
 omit =
     girder/external/*
     girder/test/*
 parallel = True
+
+[html]
+directory = build/test/artifacts/python_coverage
+
+[xml]
+output = build/test/coverage/python.xml
```
manthey: What's with the name drift?  The standard name for this file is `coverage.xml`.  We had changed that to `server.xml`, and now to `python.xml`?

brianhelba: This was after discussion with @danlamanna . `coverage.xml` is an ambiguous name, since we also have JS coverge that may be output in an XML format. The previous name of `server.xml` is also not properly descriptive, since this also includes coverage from the Python client.

At this point, I'm satisfied that this ought to be the final name. I'd chalk up the churn to normal development uncertainty and iteration.

manthey: The think the ambiguity is debatable.  Nothing in the name `coverage.xml` implies python, and nothing in `python.xml` implies coverage.  Less ambiguous would be something like `py_coverage.xml` to match the `js_coverage.xml` file name from javascript.

brianhelba:
@manthey Looking at slicer_cli_web, it appears that they use `PYTHON_COVERAGE_CONFIG` to point to a local coverage file that's functionally identical to the one that's was / will be used by upstream Girder. They do use `PYTHON_BRANCH_COVERAGE` to turn on Python branch coverage (which Girder currently has off); however, I'd like to re-address Python branch coverage in a follow-up standalone branch and likely turn it on for all of Girder.

In the future, if plugins want to run coverage with custom options then they can just write a custom coverage config file and point Pytest towards it with the appropriate CLI flag; for legacy unittest tests, they can write their own `coverage run ...` CMake test.

brianhelba:
@girder/developers @danlamanna I've just added a CHANGELOG entry and resolved a merge conflict. This is ready for final approval and merge.

danlamanna:
Can you make the changelog entry link to this PR?

.coveragerc
brianhelba: @manthey Good point. Done.

brianhelba:
@danlamanna Done.

danlamanna:
@brianhelba Did you ever file upstream issue(s) about [creating the coverage directories](https://github.com/girder/girder/pull/2517/files#diff-4f8aea714ed1849ac5947f40376926f8R7)? If so could we add code comments and/or issues?

PR #2518, opened 2017-12-14
Fix the deploy step of CI

brianhelba:
This will fix the CI tests on `master`.


brianhelba:
I've tested this by excluding the branch filter from this job.

PR #2519, opened 2017-12-14
Make all Python requirements abstract

brianhelba:
This:
* Moves all core (abstract) dependencies into `setup.py`, which is where they'll stay. The only correct way to install Girder's dependencies from source is now: `pip install ./girder` (with an optional `-e` flag), as the `requirements.txt` file is gone.
* Makes `requirements-dev.txt` abstract.
  * Eventually, we may want to have CI test against a concrete set of package and development dependencies, but Girder doesn't yet have the infrastructure to support that. Until then, all developers and CI should be using the same set of abstract dependencies.
* Moves all plugin dependencies into per-plugin `requirements.txt` files. When pip-installable plugins are added in the future, plugin dependencies will be expressed directly within each plugin's own `setup.py` file.


manthey:
Almost all of downstream testing does something like `pip install .[sftp,plugins] -r requirements-dev.txt -r clients/python` or some subset of that.  Removing (or renaming) `requirements-dev.txt` breaks all of those.  Recently, we added `-e pytest_girder` to `requirements-dev.txt` which broke a lot of downstream testing because, previously, we could do `pip install $girder_directory -r "$girder_directory/requirements-dev.txt"` (I'm also not sure why the pytest portion must get installed in editable mode, which is an unrelated issue).

How do other projects install development requirements?  Could we make it an extra_requires (e.g., `pip install $girder_directory[sftp,plugins,dev,python_client]`, or, better, `pip install $girder_directory[full-dev]`)?  Tracking these changes on downstream projects is tedious at best.

We need to lower the churn on how to deploy girder, not increase it.

brianhelba:
@manthey Agreed:
> If we do this, we'd probably copy/rename `requirements-dev.in` back to `requirements-dev.txt`, since a number of downstream repos are likely installing that file in their own testing.

However, I'm also open to the idea of moving the canonical way to install these out of some `requirements-dev` file and into another mechanism. An extras requirement would work, but I have the intuition that we're sort of abusing what that mechanism is supposed to do, since extras are published with the package (and yet development is impossible after installing from pip).

Another possibility would be to add them as requirements of `pytest_girder`, which we assume everyone who's doing development will be installing. Then, the `install for development` step becomes `pip install pytest_girder`.

Finally, note that setuptools does contain a `tests_require` option, however it's my understanding that this is generally unsupported and unused by the Python ecosystem, and is only usable if you buy into the setuptools method of launching tests.

brianhelba:
The original PR here was:
> This creates a series of `requirements*.in` files, to be the single source of truth for Girder (and plugins) abstract Python package requirements.
>
> It was intended to be used with [pip-tools](https://github.com/jazzband/pip-tools) to generate a `requirements.txt` and `requirements-dev.txt` "lockfile". While this is still possible with the command `pip-compile --output-file requirements.txt requirements.in plugins/*/requirements.in`, it's not possible to generate a lockfile that can actually be used with `pip-sync` or `pip install -r` on all platforms, as Girder's requirements vary across versions of Python (with different requirements for 2.7, 3.4, 3.5, etc.).
>
>Additionally, it should be noted that under the status quo, Girder is being installed in CI with `pip install -e .`, and as such is using an abstract set of dependencies, instead of a lockfile.
>
>So, what do @girder/developers think of just moving forwards with only abstract requirements, and declining to create a Python package lockfile? A lockfile would only be useful during testing, and even if we had it, we'd still want to run another series of tests that tested abstract requirements, since that's what is actually being used in production.
>
> If we do this, we'd probably copy/rename `requirements-dev.in` back to `requirements-dev.txt`, since a number of downstream repos are likely installing that file in their own testing.

But after discussion with @manthey , I'm updating it to be a simpler solution that doesn't address the question of concrete dependencies yet.

brianhelba:
@girder/developers PTAL again. This is ready for review.

setup.py
```diff
@@ -59,57 +58,41 @@ def run(self, *arg, **kw):
...
-            'snakebite',
-            'hachoir-core',
-            'hachoir-metadata',
-            'hachoir-parser'
-        ]
-    })
-if sys.version_info[0:2] < (3, 5):
-    install_reqs.append('funcsigs')
+extrasReqs = {}
+for pluginName in os.listdir('plugins'):
```
manthey: This may not do the right thing when there are external plugins.  Currently, if an external plugin has a `setup.py` file *and* a `requirements.txt` file, when you ask girder to install the plugin, it runs `pip install -e {pluginPath}`.  If there isn't a `setup.py` file, then it installs `pip install -r requirements.txt`.  Now, if you run `pip install .[plugins]`, this will install the `requirements.txt` file, whereas before it would have not installed the external plugin.  Worse, if those external plugins have conflicting pinned requirements, this will fail.

brianhelba: I think we should exclude external plugins entirely. We already have the established `girder-install plugin` command, and external plugins (which could disappear at any time) don't really belong as an extras option of core.

manthey: That would be fine (but, obviously, `listdir` is an insufficient enumeration technique).

brianhelba: Yeah, I think I'll use `.gitignore`. I just need to verify that it still works once it gets packaged.

brianhelba:
@manthey Updated. PTAL.

manthey:
Can you add a comment in the `.gitignore` file that it is being used by setup.py?

setup.py
```diff
@@ -59,57 +58,54 @@ def run(self, *arg, **kw):
...
+extrasReqs = {}
+# To avoid conflict with the `girder-install plugin' command, this only adds built-in plugins with
+# extras requirements.
+with open(os.path.join('plugins', '.gitignore')) as builtinPluginsIgnoreStream:
+    builtinPlugins = set()
+    for line in builtinPluginsIgnoreStream:
+        # Plugin .gitignore entries should end with a /, but we will tolerate those that don't;
+        # (accordingly, note the non-greedy qualifier for the match group)
+        builtinPluginNameRe = re.match(r'^!(.+?)/?$', line)
+        if builtinPluginNameRe:
```
jcfr: nitpick: append to the list only if directory exists

brianhelba: `builtinPlugins` is only an intermediate set, to be used in the following loop, so it will indeed contain false positive entries like `.gitignore`. However, we only query `builtinPlugins` for real directories that are returned by `os.listdir('plugins')`.

If someone else is planning on using `builtinPlugins` for another purpose in `setup.py`, then we should indeed perform additional validation and filtering. However, as the most recent comment indicates, this code is only a Girder-specific temporary measure to hold us over until pip-installable plugins force each plugin to have its own `setup.py` (instead of its own `requirements.txt`).

@jcfr With this context, do you think it's safe to keep the code simple as it is now, or should we add further guards against future people abusing this variable for other purposes?

jcfr: Thanks for the clarification. 

> do you think it's safe to keep the code simple as it is now

👍 

May be could add 

```
del builtinPlugins
```

of declare a an internal function named `def _get_directories_from_gitignore()`

and set:

```
builtinPlugins =  _get_directory_names_from_gitignore()
```

after the loop.

brianhelba: @jcfr I've pushed another commit to ensure `builtinPlugins` is accurate. Do you think this is adequate?

jcfr:
Thanks for working on this @brianhelba , the requirements gonna be much clearer now 👍

PR #2521, opened 2017-12-14
Fuse plugin

manthey:
Assuming that the OS has fuse support, if this plugin is enabled AND there is a config entry in the `[server_fuse]` section of `"path = <directory>"`, a FUSE will be mounted at that directory giving access to all files in Girder.

Additional mounts can be made with specific user access.

Note that if Girder is `SIGKILL`ed with open file handles on the FUSE, it may not be possible to fully clean up the open file system, and defunct processes may linger.  This is a defect in FUSE.  Use `SIGTERM`.

This has been tested on linux (Ubuntu) and MacOS.


plugins/fuse/server/server_fuse.py
```diff
@@ -0,0 +1,492 @@
...
+    :param path: the location where this mount will be in the local filesystem.
+        This should be an empty directory.
+    :param level: access level used when checking which resources are available
+        within the FUSE.  This is ignored currently, but could be used if
+        non-readonly access is ever implemented.
+    :param user: the user used for authorizing resource access.
+    :param force: if True, all resources are available without checking the
+        user or level.
+    :returns: True if successful.  'present' if the mount is already present.
+        None on failure.
```
kotfic: This return API seems a little confusing. Are there conditions when you would want to call mountServerFuse() and know that it was already present rather than just knowing that the path is mounted?

plugins/fuse/server/server_fuse.py
```diff
@@ -0,0 +1,492 @@
...
+    Unmount all mounted FUSE mounts.
+    """
+    for name in list(fuseMounts.keys()):
+        unmountServerFuse(name)
+
+
+def unmountServerFuse(name):
+    """
+    Unmount a mounted FUSE mount.  This may fail if there are open files on the
+    mount.
```
kotfic: So does this mean I can call unmountServerFuse(...) and it can fail silently without any way of knowing that?

plugins/fuse/server/server_fuse.py
```diff
@@ -0,0 +1,492 @@
...
+        if flags & (os.O_APPEND | os.O_ASYNC | os.O_CREAT | os.O_DIRECTORY |
+                    os.O_EXCL | os.O_RDWR | os.O_TRUNC | os.O_WRONLY):
+            raise fuse.FuseOSError(errno.EROFS)
+        info = {
+            'path': path,
+            'handle': File().open(resource['document']),
+            'lock': threading.Lock(),
+        }
+        with self.openFilesLock:
+            fh = self.nextFH
```
kotfic: Sorry,  i don't think i'm following the need for the nextFH file indexing,  why not just use info['handle'].fileno()?

plugins/fuse/server/server_fuse.py
```diff
@@ -0,0 +1,492 @@
...
+
+
+# There can be multiple FUSE mounts, each with different user permissions.  If
+# the mount needs to be exposed in some manner to something other than the
+# main server, this could be used to limit which resources are visible.  Since
+# the server can already access all resources, this is not of utility from
+# within the server itself.
+fuseMounts = {}
+fuseMountsLock = threading.RLock()
+
```
kotfic: Should fuseMounts and fuseMountsLock be public symbols here?  Are they intended to be used outside the context of this module/plugin?

plugins/fuse/server/server_fuse.py
manthey: The handle may not refer to a physical file -- it could be an S3 assetstore file, or a database file, etc.  It may not have its own fileno.

plugins/fuse/server/server_fuse.py
manthey: No.  I'll prefix them with _ to mark them private.

jbeezley:
@kotfic There is currently no mechanism for writing pytest-style tests in plugins.

plugins/fuse/server/server_fuse.py
manthey: It won't be silent; it will report an error.  I'll have to refresh my memory on exactly what is reported, though.

plugins/fuse/server/server_fuse.py
manthey: One of my ideas is that you could make a per-token or per-user mount and only expose files that are readable to that user.  In this case, you might have code that always mounts that in a predictable  spot per user, in which case knowing the mount is already present would be useful.

So far, I've only found actual utility it have the full-access mount.  Since this is accessible to the server code, and the server code already has full access to all files, making a user-limited mount would only be useful if the consuming process can't reach other local filesystem locations.

kotfic:
@jbeezley womp womp

zachmullen:
> womp womp

We're getting closer to plugins in pytest, but it's possible that it might need to wait until girder 3

plugins/fuse/server/server_fuse.py
kotfic: would it make sense to have a public function that explicitly makes the presence check? I'm just concerned about code that uses this function needing to check for ```True``` and ```"present"``` and ```None```

opadron: After taking a look.  I'm inclined to agree with @kotfic.  Making the presence check its own call and simplifying the return value of this function would seem to make for a cleaner API.

manthey: Yes -- I added an `isServerFuseMounted` method in one of the later commits.

plugins/fuse/server/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+    return server_fuse.getServerFusePath(MAIN_FUSE_KEY, 'file', file)
+
+
+# Add two methods to the file model.
+#  getFilePath returns a local file path, but not necessarily in the FUSE.
+#  getFuseFilePath returns a local file path within the main FUSE.
+File.getFilePath = getFilePath
+File.getFuseFilePath = getFuseFilePath
+
+startFromConfig()
```
kotfic: @manthey can you help me understand why this is being called as an import side-effect?

manthey: It could easily be moved to the `load` function for the plugin.  I'm not sure why I did it as an import side effect.

plugins/fuse/server/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+    if path:
+        try:
+            mkdir(path)
+            return server_fuse.mountServerFuse(MAIN_FUSE_KEY, path, force=True)
+        except Exception:
+            logger.exception('Can\'t mount resource fuse: %s' % path)
+            return False
+
+
+def getFilePath(self, file):
```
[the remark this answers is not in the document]
manthey: We do it throughout the Girder file model.  We certainly can change it, but I was trying to be consistent with that.

plugins/fuse/server/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+    :param file: file resource document.
+    :returns: a path on the local file system.
+    """
+    return server_fuse.getServerFusePath(MAIN_FUSE_KEY, 'file', file)
+
+
+# Add two methods to the file model.
+#  getFilePath returns a local file path, but not necessarily in the FUSE.
+#  getFuseFilePath returns a local file path within the main FUSE.
+File.getFilePath = getFilePath
```
kotfic: It looks like these are only being used in the tests?  is there another reason these are being added directly to the File model?

manthey: Yes -- I expect that most Girder plugins that will rely on the Fuse plugin will not want to care about the implementation details, and will just want a file path when necessary.  This extends the File model with the two functions.  Then, in a plugin like large_image, we could write:

```
# We don't need this conditional, if we know the fuse plugin is enabled
if hasattr(fileModel, 'getFilePath'):
    return fileModel.getFilePath(fileDoc)
adapter = fileModel.getAssetstoreAdapter(fileDoc)
if hasattr(adapter, 'fullPath'):
    return adapter.fullPath(fileDoc)
```

And, if we needed access to a file that might have a companion file (for instance, an item that has `foo` and `foo.xml`, and we are using an external program to parse `foo` that expects the xml file), then you would use `getFuseFilePath` to ensure that the two files appear side-by-each.

manthey:
All fuse mount errors return `RuntimeError: 1`, which is what libfuse returns (https://github.com/libfuse/libfuse/blob/master/lib/helper.c#L274).  We could wrap this and return a message like:

`"Failed to mount FUSE.  Does the mountpoint (%s) exist and is it empty?  Does the user have permission to create FUSE mounts?  It could be another FUSE mount issue, too."`

There are several other possible error conditions.  Moreover, if you have an older version of libfuse, mount directories didn't have to be empty, so checking for ourselves isn't really appropriate.

If you think such a message is clearer (or suggest a better message), then I can wrap the FUSE initialization from fusepy and log the error.  Since the traceback probably isn't helpful, I would just use `logger.error`, not `logger.debug`.

kotfic:
Its really frustrating that fuse is implemented that way.  All the same yes,  I think it would be better to offer something other than ```RuntimeError: 1```.  I think that is a good error message

plugins/fuse/server/__init__.py
kotfic: To summarize from in person conversation,  I believe the plan is to:
1. Remove getFilePath from here and introduce it as an API on the File model in Girder (a separate PR)
2. When/if that is implemented duck-punch FIle.getFilePath to include Fuse functionality (in a separate PR)
3. Use getFuseFilePath as a utility function, don't attach it to File Model here.

Is that your recollection @manthey ?

kotfic:
This is potentially controversial and or optimizing from a problem that doesn't exist yet,  what do you (others) think about disabling cherrypy auto-restart when the fuse plugin is enabled?  This would keep changing a file in girder from ripping the file-system out from under something (other than girder) that might be using it?

kotfic:
We also mentioned adding some user-facing documentation that captures the info in the description of this PR.

plugins/fuse/plugin_tests/server_fuse_test.py
```diff
@@ -0,0 +1,484 @@
...
+            self.assertTrue(os.path.exists(os.path.join(mountpath, fullpath)))
+            self.assertFalse(os.path.exists(os.path.join(mountpath, fullpath + '.other')))
+            size = os.path.getsize(os.path.join(mountpath, fullpath))
+            if self.knownPaths[fullpath]:
+                with open(os.path.join(mountpath, fullpath)) as file1:
+                    self.assertEqual(file1.read().strip(), self.knownPaths[fullpath])
+                self.assertGreater(size, 0)
+            stat = os.stat(os.path.join(mountpath, fullpath))
+            # The mtime should be recent
+            self.assertGreater(stat.st_mtime, time.time() - 1e5)
```
kotfic: separate test

plugins/fuse/plugin_tests/server_fuse_test.py
```diff
@@ -0,0 +1,484 @@
...
+            path = fullpath
+            # All parents should be folders and have zero size.
+            while '/' in path:
```
kotfic: separate test

plugins/fuse/plugin_tests/server_fuse_test.py
```diff
@@ -0,0 +1,484 @@
...
+        self.extraMount = 'test'
+        self.assertTrue(server_fuse.mountServerFuse(
+            self.extraMount, mountpath, level=AccessType.READ, user=self.user))
+        # The OS can cache stat, so wait 1 second before using the mount.
+        time.sleep(1)
+        self.assertEqual(sorted(os.listdir(mountpath)), sorted(['user', 'collection']))
+        for fullpath in self.knownPaths:
+            canSee = not fullpath.startswith('user/admin')
+            self.assertEqual(os.path.exists(os.path.join(mountpath, fullpath)), canSee)
+            if canSee and self.knownPaths[fullpath]:
```
kotfic: can we pull this into separate tests so we're testing the user can see what they're supposed to see,  and can't see what they're not supposed to see?

plugins/fuse/plugin_tests/server_fuse_test.py
```diff
@@ -0,0 +1,484 @@
...
+        self.user2 = User().findOne({'login': 'second'})
+        curConfig = config.getConfig()
+        self.mainMountPath = curConfig['server_fuse']['path']
+        self.extraMountPath = tempfile.mkdtemp()
+        self.extraMount = None
+        self.knownPaths = {
+            'user/admin/Private/Item 1/File 1A': 'File 1A',
+            'user/admin/Private/Item 1/File 1B': 'File 1B',
+            'user/admin/Private/Item 2/File 2': 'File 2',
+            'user/admin/Private/Item Without File/': None,
```
kotfic: Some of your logic could be simplified if you split this out into ```self.knownPaths``` and ```self.knownBadPaths```

plugins/fuse/plugin_tests/server_fuse_test.py
```diff
@@ -0,0 +1,484 @@
...
+            self.assertEqual(os.path.exists(os.path.join(mountpath, fullpath)), canSee)
+            if canSee and self.knownPaths[fullpath]:
+                with open(os.path.join(mountpath, fullpath)) as file1:
+                    self.assertEqual(file1.read().strip(), self.knownPaths[fullpath])
+            elif self.knownPaths[fullpath]:
+                with self.assertRaises(IOError):
+                    with open(os.path.join(mountpath, fullpath)) as file1:
+                        pass
+            path = fullpath
+            # All parents should be folders and have zero size.
```
kotfic: separate test

plugins/fuse/plugin_tests/server_fuse_test.py
```diff
@@ -0,0 +1,484 @@
...
+            while '/' in path:
+                path = path.rsplit('/')[0]
+                self.assertEqual(os.path.isdir(os.path.join(mountpath, path)),
+                                 'Private' not in path)
+
+    def testFilePath(self):
```
kotfic: not sure if this is still necessary if we are dropping getFilePath in this PR?

plugins/fuse/plugin_tests/server_fuse_test.py
```diff
@@ -0,0 +1,484 @@
...
+            self.assertEqual(fusepath[:len(self.mainMountPath)], self.mainMountPath)
+            with open(filepath) as file1:
+                with open(fusepath) as file2:
+                    self.assertEqual(file1.read(), file2.read())
+            subpath = fusepath[len(self.mainMountPath):].lstrip('/')
+            if self.knownPaths.get(subpath):
+                with open(fusepath) as file1:
+                    self.assertEqual(file1.read().strip(), self.knownPaths[subpath])
+
+    def testFilePathNoFullPath(self):
```
kotfic: not sure if this is still necessary if we are dropping getFilePath in this PR?

plugins/fuse/plugin_tests/server_fuse_test.py
```diff
@@ -0,0 +1,484 @@
...
+        # remount with user
+        self.assertFalse(server_fuse.isServerFuseMounted(
+            self.extraMount, level=AccessType.READ, user=self.user))
+        self.assertTrue(server_fuse.mountServerFuse(
+            self.extraMount, mountpath, level=AccessType.READ, user=self.user))
+        # The OS can cache stat, so wait 1 second before using the mount.
+        time.sleep(1)
+        self.assertTrue(os.path.exists(publicFile))
+        self.assertTrue(os.path.exists(privateFile))
+        # remount without any changes should work, too.
```
kotfic: separate test

manthey:
> This is potentially controversial and or optimizing from a problem that doesn't exist yet, what do you (others) think about disabling cherrypy auto-restart when the fuse plugin is enabled? This would keep changing a file in girder from ripping the file-system out from under something (other than girder) that might be using it?

I'm against doing this.  In development and testing, I frequently let Girder auto-restart.  Yes, if something has the file handle open, Girder may not shutdown until that handle is released, but once it is released, Girder stops and then restarts.  This has helped show where something has cached a file handle in an unexpected way.

plugins/fuse/plugin_tests/server_fuse_test.py
manthey: Here, the entry with a None isn't bad in any way.  The parent path exists, but there is no file in this item.  In several tests, we check if the parent filesystem folders (Girder users, collections, folders, or items) exist and have certain properties.  If there is a Girder file, we also check if the filesystem has a file.  Splitting this into two objects will result in duplicating many of the test loops, since we want to check that the parent folders have those properties regardless.

If you think splitting this up and making numerous more tests aids clarity, then I can do that.  I fear it will reduce clarity, since it will no longer be obvious that we are testing for the same behavior of the parent folders.

kotfic:
Yeah,  obviously my viewpoint is more informed by girder-worker,  and I was thinking it would be annoying if my task died because all of a sudden the file-system it was working on disappeared,  but i think you're right,  we can just leave it as is and address it as/if issues come up

plugins/fuse/plugin_tests/server_fuse_test.py
```diff
@@ -0,0 +1,484 @@
...
+        Test the default mount point has access to all of the expected files.
+        """
+        mountpath = self.mainMountPath
+        self.assertEqual(sorted(os.listdir(mountpath)), sorted(['user', 'collection']))
+        # Check that all known paths exist and that arbitrary other paths don't
+        for fullpath in self.knownPaths:
+            self.assertTrue(os.path.exists(os.path.join(mountpath, fullpath)))
+            self.assertFalse(os.path.exists(os.path.join(mountpath, fullpath + '.other')))
+            size = os.path.getsize(os.path.join(mountpath, fullpath))
+            if self.knownPaths[fullpath]:
```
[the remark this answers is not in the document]
manthey: Creating separate tests ends up duplicating some setup for each.  Instead, this test could be made clearer, like so:
```
    def testMainMount(self):
        """
        Test the default mount point has access to all of the expected files.
        """
        mountpath = self.mainMountPath
        self.assertEqual(sorted(os.listdir(mountpath)), sorted(['user', 'collection']))
        # Check that all known paths exist and that arbitrary other paths don't
        for testpath, contents in six.iteritems(self.knownPaths):
            localpath = os.path.join(mountpath, testpath)
            # The path must exist
            self.assertTrue(os.path.exists(localpath))
            # The path plus an arbitrary string must not exist
            self.assertFalse(os.path.exists(localpath + '.other'))
            # If the path is a file, check that it equals the expected value
            # and reports a non-zero size
            if contents:
                size = os.path.getsize(localpath)
                with open(localpath) as file1:
                    self.assertEqual(file1.read().strip(), contents)
                self.assertGreater(size, 0)
            # The mtime should be recent
            stat = os.stat(localpath)
            self.assertGreater(stat.st_mtime, time.time() - 1e5)
            # All parents should be folders and have zero size.
            subpath = testpath
            while '/' in subpath:
                subpath = subpath.rsplit('/')[0]
                localpath = os.path.join(mountpath, subpath)
                self.assertTrue(os.path.isdir(localpath))
                self.assertEqual(os.path.getsize(localpath), 0)
                # An arbitrary alternate file shjould not exist
                self.assertFalse(os.path.exists(localpath + '.other'))
```
Further, in the user and second user mount tests, the `canSee` variable uses a simple string to determine if the path should be visible -- these could all be abstracted to a single test function that would use an optional substring match and include the visibility tests.

If this has to be broken into multiple tests, then each test function has the same two lines:
```
        for testpath, contents in six.iteritems(self.knownPaths):
            localpath = os.path.join(mountpath, testpath)
```

kotfic:
Your test suggestions make sense to me.  The winning feature of me is comments that describe the behavior each assertion is testing.  I believe the only open issues are:
+ https://github.com/girder/girder/pull/2521#discussion_r169428986
+ Cleanup tests + add comments
+ User documentation

manthey:
@kotfic I think I've addressed all of your comments.

docs/plugins.rst
```diff
@@ -676,3 +676,34 @@ The Vega plugin uses the `Vega library <http://trifacta.github.io/vega>`_ to ren
 JSON objects directly in the Girder application. To use it, simply upload the JSON file as an item,
 and then set a "vega: true" metadata field on the item. The visualization will then be rendered
 directly in the item view.
+
+Server FUSE
+----------- 
+
+When this plugin is enabled *and* the appropriate configuration option is set,
+it mounts a user-space filesystem that allows reading any file in Girder as if
```
kotfic: maybe "mounts a read-only user-space filesystem" ?  I think its important to use the word 'read-only'

plugins/fuse/plugin_tests/server_fuse_test.py
```diff
@@ -75,27 +77,34 @@ def testMainMount(self):
...
-                self.assertFalse(os.path.exists(os.path.join(mountpath, path + '.other')))
-                if '/' in path:
-                    self.assertEqual(os.path.getsize(os.path.join(mountpath, path)), 0)
+            subpath = testpath
+            while '/' in subpath:
+                subpath = subpath.rsplit('/')[0]
+                localpath = os.path.join(mountpath, subpath)
+                self.assertTrue(os.path.isdir(localpath))
+                self.assertEqual(os.path.getsize(localpath), 0)
+                # An arbitrary alternate file shjould not exist
```
kotfic: should

manthey:
@kotfic I also added an entry to the changelog.  Can you reapprove?

docs/plugins.rst
```diff
@@ -681,11 +681,11 @@ Server FUSE
 ----------- 
 
 When this plugin is enabled *and* the appropriate configuration option is set,
-it mounts a user-space filesystem that allows reading any file in Girder as if
-it were a physical file.  This allows external libraries that require file
-access to read Girder files, regardless of which assetstore they are stored on.
-It also uses the underlying operating system's caching to improve reading these
-files.
+it mounts a real-only user-space filesystem that allows reading any file in
```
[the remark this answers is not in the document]
manthey: Darn it.  I'll make a PR to fix it.

PR #2523, opened 2017-12-14
Set a maximum height for modal dialogs.

manthey:
Currently, if a modal dialog is tall, the webpage as a whole gets a scrollbar and the dialog extends past the bottom.  I think it looks better to have the dialog have a scroll bar (since the page may already be scrolled).  The bootstrap css has different margins depending on window size; this makes one policy that is consistent.

This has the virtue that scrolling within the dialog doesn't affect the scroll position on the underlying main page.


jbeezley:
I've done this in other girder-based applications and found it to work well.  I'll let it sit for a while before approving.

PR #2529, opened 2017-12-14
Port custom root test to pytest

zachmullen:
This involved breaking out a random off-the-cuff assertion about
plugin import semantics into its own file since it didn't really
belong in the original test.


danlamanna:
> The main purpose of this commit is to put band-aids over several
problems that all result from a single bad assumption, which
is treating the girder server as a single global python object.
In the old testing mode, we only ever had one cherrypy server per
test process, so we didn't run into these issues where we treated
the python global namespace as equivalent to the context of the
cherrypy server.   
In our new pytest framework, we need to spin the entire server
context up and down repeatedly in the same python global context,
meaning we either need the server fixture cleanup process to
tear everything down to a good state, or to make the server startup
process idempotent, or (ideally) both. This commit fixes the issues I ran
into, and does a little bit of both.

I ran into these types of issues a lot. Part of the issue is in Girder and part of it is in CherryPy since it uses module level variables for storing everything. Flask handles this by designing your application around a single "app" object which can then be instantiated multiple times in the same process. Items like the database connection, event daemon, etc all get attached to the instantiated app object rather than being scattered around in globals. See:
http://flask.pocoo.org/docs/0.12/patterns/appfactories/
http://flask.pocoo.org/docs/0.12/appcontext/#purpose-of-the-application-context

zachmullen:
https://github.com/girder/girder/issues/2530

zachmullen:
Depends on #2527

zachmullen:
This is now ready for review

girder/utility/server.py
```diff
@@ -248,7 +250,7 @@ def setup(test=False, plugins=None, curConfig=None):
     cherrypy.tree.mount(girderWebroot.api, '/api', appconf)
 
     # Mount everything else in the routeTable
-    for (name, route) in six.viewitems(routeTable):
+    for name, route in six.viewitems(routeTable):
```
danlamanna: Is this actually a failing linting test?

girder/api/describe.py
```diff
@@ -462,8 +462,10 @@ def __init__(self, templatePath=None):
             'mode': mode
         }
 
+        events.unbind('model.setting.save.after', CoreEventHandler.WEBROOT_SETTING_CHANGE)
```
danlamanna: Just an FYI, these handlers will be removed once #2274 is merged.

girder/utility/server.py
```diff
@@ -170,6 +170,8 @@ def configureServer(test=False, plugins=None, curConfig=None):
 
     _configureStaticRoutes(root, plugins)
 
+    # Must unbind this handler so that this method is idempotent
+    girder.events.unbind('model.setting.save.after', '_updateStaticRoutesIfModified')
```
danlamanna: Same with these handlers.

girder/test/test_plugin_import.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+import types
+
+
+@pytest.mark.testPlugins(['test_plugin'])
+def testPluginImportSemantics(server):
```
danlamanna: Could this be thought of as a test of the `testPlugins` marker itself? I have a local branch where I've been trying to work on testing `pytest-girder`.

danlamanna:
I think master needs to be remerged due to some recent changes with Circle.

girder/test/test_plugin_import.py
zachmullen: Yes, it does end up asserting that that decorator is behaving correctly.

girder/utility/server.py
zachmullen: It wasn't, I am just not a fan of unnecessary parens unless they add some readability.

PR #2531, opened 2017-12-15
Change the default for pytest to drop databases.

manthey:
Otherwise, when running tests locally, one ends up with lots of databases in the local mongo instance.  This change matches the behavior we have on the non-pytest tests.

### DIFF
--- a/pytest_girder/pytest_girder/plugin.py
+++ b/pytest_girder/pytest_girder/plugin.py
@@ -8,5 +8,7 @@ def pytest_addoption(parser):
     group.addoption('--mongo-uri', action='store', default='mongodb://localhost:27017',
                     help=('The base URI to the MongoDB instance to use for database connections, '
                           'default is mongodb://localhost:27017'))
-    group.addoption('--drop-db', action='store', default='post', choices=('pre', 'post', 'never'),
-                    help='When to destroy testing databases, default is post (after running tests)')
+    group.addoption('--drop-db', action='store', default='both',
+                    choices=('both', 'pre', 'post', 'never'),
+                    help='When to destroy testing databases, default is both '
+                    '(before and after running tests)')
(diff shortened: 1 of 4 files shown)


danlamanna:
Should this change be the default for `pytest-girder`, changed [here](https://github.com/girder/girder/blob/c687b73/pytest_girder/pytest_girder/plugin.py#L11) then?

manthey:
Yes, I think I should have changed this in pytest-girder.

brianhelba:
I agree with @danlamanna. This currently the wrong place to set this default.

brianhelba:
Also, what's the point of the `pre` option. That is, why don't tests always drop a database before starting to run on it?

I understand the debugging utility of `post`, of course.

danlamanna:
> Also, what's the point of the `pre` option. That is, why don't tests
> always drop a database before starting to run on it?

For some reason I really thought this was the default behavior of our
ctest harness, so I added it (and made it the default) in pytest.

manthey:
The default is to drop *both* before and *after*.  There is a cryptic option to fail if there is an existing database (if dropping *before* is actually necessary), and an option to not drop when finished.  Here, we don't have the *both* option; perhaps we should and it should be the default.

pytest_girder/pytest_girder/plugin.py
```diff
@@ -8,5 +8,7 @@ def pytest_addoption(parser):
     group.addoption('--mongo-uri', action='store', default='mongodb://localhost:27017',
                     help=('The base URI to the MongoDB instance to use for database connections, '
                           'default is mongodb://localhost:27017'))
-    group.addoption('--drop-db', action='store', default='pre', choices=('pre', 'post', 'never'),
-                    help='When to destroy testing databases, default is pre (before running tests)')
+    group.addoption('--drop-db', action='store', default='both',
```
zachmullen: I like this as the default for local dev testing, but could we make CI use `never`? It has the advantage of being faster, and preserving the databases for debugging in CI if we need to go poke around with SSH enabled.

manthey: Would `pre` be better for CI?  If you are sshing in and rerunning the tests, then it would at least clear the old data?

zachmullen: I think if you're SSHing in and re-running the tests, you'd be doing so manually, i.e. using the normal default of `both`.

PR #2528, opened 2017-12-14
Tox integration

danlamanna:

PR #2532, opened 2017-12-15
Add documentation for using tox

danlamanna:
@manthey PTAL. I tried to document the most common scenarios - unfortunately the tox documentation focuses mainly on writing `tox.ini` files and doesn't have many usage examples I could find.

Follow on of #2528


manthey:
Thanks.

PR #2533, opened 2017-12-15
RFC: Developer ToC

danlamanna:
This isn't intended to be merged, but I wanted to lay out some thoughts for what could be a table of contents for Girder development.

The ToC is stubbed out with fake links, in this PoC the actual location where the links go to isn't relevant. What's important here is if people think a high level overview ToC is a good idea for Girder, and what the high level items should be. It's also important that this list be strictly curated, for a link (and especially a new section) to be added it has to be justified to avoid the ToC becoming a sort of dumping ground that's very difficult to sort through.

Visual representation:
40833fe0-3535-49a0-bd55-d4ca54922f50

### DIFF
--- a/docs/developer-docs.rst
+++ b/docs/developer-docs.rst
@@ -1,9 +1,39 @@
 Developer Documentation
 =======================
 
-.. toctree::
-   :maxdepth: 2
+Models
+$$$$$$
 
-   dev-installation
-   tutorials
-   api-docs
+Python models: `Overview <#>`_ | `Built in models <#>`_ | `Creating custom models <#>`_
+
+JavaScript models: `Overview <#>`_ | `Built in models <#>`_ | `Creating custom models <#>`_
+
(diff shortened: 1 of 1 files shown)


mgrauer:
This is a great start. I'd like you to add stubs for CI, Plugin development (though this has cross cutting concerns), maybe JS routing, maybe Python client.

danlamanna:
@girder/developers I wanted to add a bit more to this and maybe stub out one more level of depth, but I'm curious to know how people feel about this in general.

ronichoudhury:
This structure is great, and I like the admonition to justify the addition of a link or level to this ToC.

That being said, I'm still a big fan of the top-level breakdown from the [Django docs](https://docs.djangoproject.com/en/2.0/) and I think a very similar approach would go a long way towards making the Girder docs easier to digest for both newcomers and old hands.

Thanks @danlamanna!

PR #2537, opened 2017-12-19
Change pytest_girder's "drop-db" argument to "keep-db"

brianhelba:
Tests should always be run on a clean database (there are / should be other pytest-compatible mechanisms to pre-populate test data). Accordingly, the "drop-db" option can be simplified to a boolean "keep-db" option. The default of "keep-db" is false (to destroy the database after testing), to prevent naive execution of "pytest" from leaving unwanted database clutter.

CI will run with "keep-db" enabled, to allow inspection of the database state after SSHing into a CI machine.

### DIFF
--- a/pytest_girder/pytest_girder/fixtures.py
+++ b/pytest_girder/pytest_girder/fixtures.py
@@ -26,7 +26,7 @@ def db(request):
 
     Provides a Mongo test database named after the requesting test function. Mongo databases are
     created/destroyed based on the URI provided with the --mongo-uri option and tear-down
-    semantics are handled by the --drop-db option.
+    behavior is modified by the --keep-db option.
     """
     from girder.models import _dbClients, getDbConnection, pymongo
     from girder.models.model_base import _modelSingletons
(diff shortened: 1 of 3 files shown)


brianhelba:
This addresses issues first raised in #2531.

manthey:
As long as we have defaulted the pytest databases to being kept in CI, should we do the same for the non-pytest databases?

zachmullen:
Probably a good idea.

manthey:
Should there be a way to turn off `--keep-db`?  If you ssh into CI and want to run this via tox, is there a way to override the `--keep-db` option?  I would expecting that running `tox -e circleci-py35 -- --no-keepdb` or something similar ought to work.

brianhelba:
@manthey I think at the moment, the best way is to just run `pytest` directly, if you're already SSH'd into a CI machine and want to run things in a different way.

I'm very hesitant to go back down the path of re-wrapping every one of our tool options at every layer of abstraction. Pytest is still very usable as a direct command, but I'm concerned it will start to be unusable if we start tacking on more complicated arguments, then hide them under Tox.

brianhelba:
@manthey I'd like to merge, but I'll wait for your acknowledgement before doing so.

manthey:
I see the virtue of tox in a development environment to run against multiple versions of python.  I don't see its virtue in our CI context; it is slow to create a new virtual environment, and it looks like we aren't advocating its use when debugging the CI.  This concern doesn't affect this PR, though.

brianhelba:
@manthey I agree that Tox is rather redundant under our current CI system. However, I hope that once we have all tests running through Tox, we can dramatically simplify the CI environment and configuration (e.g. have only one primary Docker image, don't create a virtualenv explicitly, etc.), and allow Tox to take responsibility for the matrix of multiple Python versions.

PR #2541, opened 2017-12-20
Run pytest in strict mode, to flag unregistered markers

brianhelba:
* Modularize the work done by pytest_girder's "pytest_configure" hook
* Register "testPlugins" as a custom pytest_girder marker
* Run pytest in strict mode, to flag unregistered markers


brianhelba:
@danlamanna More docs sounds good, but I don't have a strong vision for what they should be, or any intuition about the naming convention. Anything you'd like to add would be welcome though.

danlamanna:
I'm fine with merging this as is, we can address docs as we add future markers. For now the only ones I can think of are test plugins (which might change in Girder3?) and external data (which doesn't yet exist).

PR #2543, opened 2017-12-21
Move Python tool configuration to "setup.cfg"

brianhelba:
This is a broadly accepted location for Python-ecosystem tooling to store
configuration info. An equally-supported location (by all 3 of the tools at
hand) would be "tox.ini".


brianhelba:
For the record:
* Flake8 itself puts [its own `[flake8]` configuration into `tox.ini`](https://github.com/PyCQA/flake8/blob/master/tox.ini#L140)
* Pytest itself puts [its own `[pytest]` and `[flake8]` configuration into `tox.ini`](https://github.com/pytest-dev/pytest/blob/master/tox.ini#L197)

brianhelba:
Rebased to account for #2544 

@danlamanna PTAL again

brianhelba:
@girder/developers This should have included the changelog entry:
> Flake8 settings are now able to be automatically detected by many editors and IDEs. The `flake8` tool may now be invoked directly from the command line, without necessarily using a CMake test.

This is a solid step towards respecting the native use of development tools, and not re-wrapping them with magic options inside CMake or Python `console_scripts`.

I'll add this to the changelog in another PR.

manthey:
Is there a way for external plugins to combine their own setup.cfg files with Girder's setup.cfg file?  An external plugin could use tools that Girder doesn't and not want to override any of Girder's settings.

kotfic:
I was under the impression that once plugins were python packages (and thus freed from being installed under girder's source tree) they would be free to set up their own tool configuration however they saw fit.

manthey:
Plugins aren't python packages yet.

Using cmake to build a project, if the plugin has a setup.cfg file, it is being used in preference to Girder's setup.cfg file for flake8 at least (and thus I see a D107 linting error).

kotfic:
I guess it is not surprising that some of the tooling is breaking considering we're nesting python packages inside each-other.  Is it possible to migrate the settings in the plugin setup.cfg?

Also,  meta:  @brianhelba if i'm not mistaken @manthey would not have approved these changes if he had reviewed this PR -  What is the preferred way to indicate that after merge?

manthey:
It turns out that the setup.cfg of the plugin was unnecessary, so I can solve the immediate problem by removing the plugin's file.  I'm not opposed to this change (or even sure that it would have warranted delaying merging it).  However, when we break a tooling path, we should tell developers how they are supposed to use the new tooling.

kotfic:
if it's alright i'd like to move discussion about this over to the mailing list otherwise its going to muck up the comment history of the PR

brianhelba:
To clarify for all, there was never a publicly-settable CMake variable for defining the path to the Flake8 config file. Previously, it was hardcoded (via an internal variable) to always point to `${PROJECT_SOURCE_DIR}/tests/flake8.cfg`, where `${PROJECT_SOURCE_DIR}` is always the Girder repository root.

This change just allows Flake8 to use its internal autodetection logic to find the config file, but doesn't *remove* the ability for downstreams to set it, since they never could.

---

As for an issue with multiple `setup.cfg` files, I don't fully understand what the issue is. Is it that a plugin's own `setup.cfg` (which I assume has no Flake8 config within it?) is causing Girder's `setup.cfg` to be missed by Flake8? If so, I'm not sure how that happens, since Flake8 (when run by CMake) is still being run with a CWD of Girder's root.

manthey:
Yes, the plugin's setup.cfg file, which has NO flake8 section within it is causing flake8 to use it instead of Grider's setup.cfg.  I think I'd have to look through flake8's code to find out how to decides which setup.cfg to use -- all of the other tooling uses Girder's setup.cfg.

manthey:
This is definitely an oddness in flake8.  Their code (in flake8's `options/config.py`) walks back from the source file directory, walking up the file tree until it finds one or more configuration files at that level.  For example, with a `setup.cfg` file in the plugin's root, flake8 looks for:
```
'/home/ubuntu/girder/plugins/HistomicsTK/server/setup.cfg'
'/home/ubuntu/girder/plugins/HistomicsTK/server/tox.ini'
'/home/ubuntu/girder/plugins/HistomicsTK/server/.flake8'
'/home/ubuntu/girder/plugins/HistomicsTK/setup.cfg'
'/home/ubuntu/girder/plugins/HistomicsTK/tox.ini'
'/home/ubuntu/girder/plugins/HistomicsTK/.flake8'
```
Without the setup.cfg in the root, it instead looks for:
```
'/home/ubuntu/girder/plugins/HistomicsTK/server/setup.cfg'
'/home/ubuntu/girder/plugins/HistomicsTK/server/tox.ini'
'/home/ubuntu/girder/plugins/HistomicsTK/server/.flake8'
'/home/ubuntu/girder/plugins/HistomicsTK/setup.cfg'
'/home/ubuntu/girder/plugins/HistomicsTK/tox.ini'
'/home/ubuntu/girder/plugins/HistomicsTK/.flake8'
'/home/ubuntu/girder/plugins/setup.cfg'
'/home/ubuntu/girder/plugins/tox.ini'
'/home/ubuntu/girder/plugins/.flake8'
'/home/ubuntu/girder/setup.cfg'
'/home/ubuntu/girder/tox.ini'
'/home/ubuntu/girder/.flake8'
```
Existence of any config file at one level (even if it has nothing to do with flake8) is enough to stop it from searching any further toward the root.

See flake8's issue here: https://gitlab.com/pycqa/flake8/issues/267

PR #2544, opened 2017-12-21
Enable Python branch coverage

brianhelba:


brianhelba:
@girder/developers FYI, this drops our coverage percentage by 3.3%, down to 88% overall. This is not a bug, but just a (hopefully) more accurate reflection of which Python branch conditions are actually getting executed during test.

I'll be merging soon.

mgrauer:
@brianhelba Could you drop in a concrete example of what was considered covered before and is no longer?

brianhelba:
@mgrauer The example [at the top of the Coverage.py branch coverage docs](http://coverage.readthedocs.io/en/coverage-4.4.2/branch.html) should hopefully be illustrative. The idea is that that even if you're executing every line of code on its own, you may not be traversing every conditional path through the code, and so your coverage should be decreased in the form of "partial" line hits.

The actual examples of changes in Girder are simply those reported in [this PR's Codecov report](https://codecov.io/gh/girder/girder/compare/7e3f351db4e382dd433841b8f1eebf8ab7d57cb4...396de29e162246bdf2da00969a0328e2508eea69).

PR #2545, opened 2017-12-21
Enable JS branch coverage

### LINKED ISSUE #2539: Decide on one format for reporting JS coverage to Codecov
Depending on the file format that coverage information is serialized as, Codecov reports subtly different results. We should settle on using one format, that expresses coverage in a way that's most useful to us as developers.

Istanbul can support [the following coverage formats](https://github.com/istanbuljs/istanbuljs/tree/master/packages/istanbul-reports/lib). Codecov can support at least [the following coverage formats](https://docs.codecov.io/v4.3.0/reference#section-acceptable-report-formats).

For reference, here are Codecov outputs for:
* [The status quo (Lcov and Cobertura)](https://codecov.io/gh/girder/girder/tree/accf0a2497b074f6edd36123b2e58f29820a00ec/clients/web/src)
* [Lcov](https://codecov.io/gh/girder/girder/tree/f3ee9d8c4e9139edbf8fee6b1d30fb474fe7418a/clients/web/src)
* [Cobertura](https://codecov.io/gh/girder/girder/tree/3c4bed8c9c250a61407ec8983d72d5e2c2485749/clients/web/src)
* [Istanbul JSON](https://codecov.io/gh/girder/girder/tree/528357af82a0e71a46a3a8c3c73faa87189166cb/clients/web/src)
* [Clover XML](https://codecov.io/gh/girder/girder/tree/5ae2736a1310eeccf497a731a18b246aa596144f/clients/web/src)

brianhelba:
This makes two significant changes:
* Javascript branch coverage is enabled, showing lines that may only be partially covered.
  * This causes the overall coverage percentage to decrease, but it is a (hopefully) more accurate reflection of which Javascript branch conditions are actually getting executed during test.
* Codecov will now report separate metrics for server, python-client, and web-client coverage.
  * These are fundamentally separate components, so conflating them under one coverage metric is misleading.

The full list of commits are:
* Use nyc instead of istanbul to report coverage
* Report Javascript branch coverage
* Capture Javascript HTML coverage report as a CircleCI artifact
* Disable Codecov patch status
* Split Codecov reports into separate values for different components
* Don't cause test failures when coverage is too low
  * Coverage being too low should not trigger test-time failures, as these cause CircleCIs workflow steps to fail and subsequent operations to not run. Coverage threshold failures will be reported to the PR via Codecov, which will still block merging.
* Add PR #2543 to the changelog
* Move a Python coverage setting to config file

Fixes #2539


brianhelba:
@girder/developers PTAL

.codecov.yml
```diff
@@ -1,6 +1,23 @@
 comment: false
 coverage:
   status:
+    # Patch status gives meaningless results whenever a test is modified without changing the
+    # correspondingly tested code, and may be difficult to interpret without context. Disable
+    # posting it to GitHub.
+    patch: false
     project:
-      default:
-        threshold: 1
```
jbeezley: You might keep the threshold option.  Our javascript coverage in particular tends to vary from run to run due to async timing.

brianhelba: This is obsoleted by the `target` option. The check now passes if the coverage is above the threshold (which means we can insist on it always passing).

danlamanna:
Introducing `nyc` as a dependency has caused sporadic issues with the base `vagrant up` command for Girder. There appear to be some issues with how it does file locking during build, see https://github.com/Hashnode/mern-cli/issues/16#issuecomment-232751302.

brianhelba:
@danlamanna Can you create an issue for this? It sounds like an npm issue with installing packages under Vagrant that happens to affect NYC more often than others?

PR #2546, opened 2017-12-21
Move Javascript tool configuration to "package.json"

brianhelba:
It is broadly accepted for Javascript-ecosystem tooling to store its configuration info in "package.json".


brianhelba:
This should wait for #2545 to be merged first, then also incorporate `.nycrc.json`.

brianhelba:
@girder/developers PTAL.

PR #2548, opened 2018-01-03
Run npm binaries via "npx"

brianhelba:
* Run npm binaries via "npx"
* Document the required version of npm
  * 5.2 is needed for the "npx" command
  * 5.6 is needed to generate platform-independent package-lock.json files
* Don't install Grunt globally, since npx can be used instead

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -42,6 +42,7 @@ Security Fixes
 Changes
 -------
 * Exceptions are now all accessible in the ``exceptions`` module and are descended from the ``GirderBaseException`` class.
+* Require npm 5.2+ (with npm 5.6+ strongly recommended) to build the web client
 
 Deprecations
 ------------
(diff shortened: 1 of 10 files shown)


brianhelba:
This removes more CMake code. It also makes CMake run npm tools in the same way that developers (should) run them, which is more progress towards supporting native usage of tools.

kotfic:
Does it make sense to briefly document the way developers "should" run these tools with links out to tool documentation?

brianhelba:
Not sure how I missed those places. I'll fix.

@kotfic Developers don't *need* to do anything differently under normal testing yet. Other PRs have and will unshackle our tools from CMake, but even in the final state, our tests should be kicked off by Tox or an `npm run` task in the typical case. Accordingly, I don't think Girder needs to include usage info for individual tools, as it would be duplicative of upstream tool docs. However, I do think it would be positive to document an inventory of our tools with links to upstream info, but I'd rather wait until the updates to these tools makes it more clear how they all fit together.

brianhelba:
@jbeezley These places were already covered in the current PR. Did you grep from the correct branch?

jbeezley:
Yeah, I was on the wrong branch.  :disappointed:

PR #2550, opened 2018-01-04
Allow ESLint to be run as a native tool

brianhelba:
The major motivation for this change is to enable the invocation of `eslint` as a natively-run CLI tool (which allows the use of helpful options like `--fix`). Additionally, these changes will allow ESLint settings for plugin tests to be automatically detected by many editors and IDEs.

This has the perhaps-more-controversial side benefits of upgrading ESLint and removing some CMake options. Note:
* The latest major release of ESLint is more strict about line spacing, so downstream external plugins may notice some new failures.
  * However, these can now be fixed automatically with `npx eslint --fix`!
* The removed CMake options should never cause downstream plugins to experience explicit failures. It should also generally not affect testing functionality, since most downstreams will already have placed their `.eslintrc.json` and `.eslintignore` files in sane places, where ESLint will automatically pick them up.

Relevant documentation about these changes for downstreams can be found in the changelog and in updates to the plugin developer guide,

A full list of commits is:
* Upgrade ESLint packages to latest releases
* Use ESLint's built-in mechanism for configuration cascading
  * This removes the ESLINT_CONFIG_FILE CMake option.
  * Also remove the ESLINT_IGNORE_FILE CMake option, in favor of globally ignoring the "web_client/extra/" directory and file-level exclusion.
* Correct usage of "eslint-disable"
* Update ES6 style of Grunt "version-info" task
* Fix whitespace style issues found by new ESLint version
* Enable new ESLint rules to detect common defects
  * These do not actually cause any new test failures, and should be uncontroversial.
* Add an ESLint rule to make ternary-if statements use consistent newlines
  * ESLint's stricter indentation rule may force ternary-if statements to not align properly. This rule ensures that the statements will typically use only either 1 or 3 lines, ensuring their readability.
* Remove unused ESLint disable directives


tests/JavascriptTests.cmake
```diff
@@ -27,25 +27,10 @@ function(add_eslint_test name input)
...
   endif()
 
-  set(_args ESLINT_IGNORE_FILE ESLINT_CONFIG_FILE)
-  cmake_parse_arguments(fn "${_options}" "${_args}" "${_multival_args}" ${ARGN})
-
-  if(fn_ESLINT_IGNORE_FILE)
-    set(ignore_file "${fn_ESLINT_IGNORE_FILE}")
-  else()
-    set(ignore_file "${PROJECT_SOURCE_DIR}/.eslintignore")
-  endif()
```
jbeezley: Unfortunately, I think you'll have to add the ignore file configuration option back.  `eslint` only looks for an ignore file in `CWD`, and there is no way to extend it by adding ignore files in subdirectories like you can with config files.

.eslintignore
```diff
@@ -1,4 +1,4 @@
-docs/_build/
-clients/web/dev/
+build/
 clients/web/static/built/
 clients/web/test/lib/
+docs/_build/
```
jbeezley: Could you add `plugins/*/web_client/extra/`?  Many plugins use [this directory](https://github.com/girder/large_image/blob/master/.eslintignore) to store vendored dependencies.  Given the limitation on ignore files (mentioned in my other comment), this will at least allow running `eslint` standalone to work in *most* cases.

jbeezley:
Should `npm run lint` just run `eslint .` now rather than specifying individual paths?

tests/JavascriptTests.cmake
brianhelba: Yes, you're right, I misread the docs. I'll fix.

.eslintignore
brianhelba: Sure.

brianhelba:
Yes, the intent is that the tool can now be run once at the top level. CMake is still running multiple tests, to avoid being more disruptive than necessary in this PR.

brianhelba:
@jbeezley PTAL. I've updated the docs and added a new entry to the `.eslintignore` file.

Issue #2385, opened 2017-10-06
Retire old/unsupported plugins


mgrauer:
We'll need to decide which ones of the built-in plugins should be removed as part of the Girder 3.0 release.

danlamanna:
Current candidates for removal are:
- celery_jobs
- metadata extractor
- geospatial
- Mongo search
- Vega
- Item preview

@girder/developers should chime in on any built in plugins they're using or know are being used from the list and plan migration paths or remove them from the list.

brianhelba:
I'd add `jquery_widgets`  to the list.

zachmullen:
I know we have downstreams actively using geospatial, as well as metadata extractor. So for those two, at least, we should migrate them out of built-in status but keep them around under the girder org and only maintain them when the need arises.

jcfr:
👍 

Can we add a section in the doc (http://girder.readthedocs.io/en/latest/plugins.html) mentioning when each plugin was removed and which girder core functionality could be used as a replacement ...

danlamanna:
Updated list:

- celery_jobs
- jquery_widgets
- metadata extractor (to be migrated out of core)
- geospatial (to be migrated out of core)
- Mongo search
- Vega
- Item preview

manthey:
The vega plugin was updated recently, so perhaps it shouldn't be removed.

brianhelba:
Should we perhaps move `vega` out of core instead? Having a plugin which only activates when the magic `vega` metadata field is set seems pretty niche, and the actual code / logic within the plugin is extremely minimal (it just serves to inject Vega elements into the item page).

jcfr:
To extract commit history, you could look at what was done in https://github.com/Slicer/SlicerLegacyModules#how-to-extract-module-history-from-slicer-and-include-it-in-this-extension-

waxlamp:
We could instead detect that a given JSON file is in fact a Vega spec,
which can be done by examining the `$schema` top-level field. The magic
field could instead be used to disable visualizing a particular Vega spec.

mgrauer:
Also in the discussion: be more explicit about what plugins we want to include and why, any categories (experimental, blessed), what decision framework to use to decide this (funding, interest, sprawl of features). Possibly add these flags to the python package metadata.

jeffbaumes:
@danlamanna can we consider this closed?

jeffbaumes:
[...]

PR #2552, opened 2018-01-04
Add a registry for loading and querying plugins

jbeezley:
This adds a module meant to replace the functional contents of `utilities/plugin_utilities.py` for the girder 3.0 migration.  This module is not yet used to maintain backwards compatibility.  I have added it as an independent PR to collect feedback before moving on with the larger effort to replace core girder code with this new module, which will require backwards incompatible changes and migration of existing plugins.

I don't know how useful this is as a standalone PR, but it is the most I can achieve before I start breaking compatibility.  Alternatively, I can make this a WIP and start pushing breaking changes here so others can get a better idea for where this is headed.

### DIFF
--- a/girder/_plugin.py
+++ b/girder/_plugin.py
@@ -1,0 +1,266 @@
+###############################################################################
+#  Copyright Kitware Inc.
+#
+#  Licensed under the Apache License, Version 2.0 ( the "License" );
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+###############################################################################
+
+"""
+This module defines functions for registering, loading, and querying girder plugins.
+"""
+
+import abc
+import traceback
+
+from pkg_resources import iter_entry_points
+import six
+
+from girder import logprint
+
+NAMESPACE = 'girder.plugin'
+_pluginRegistry = None
+_pluginFailureInfo = {}
+
+
+@six.add_metaclass(abc.ABCMeta)
+class GirderPlugin(object):
+    """
+    This is a base class for describing a girder plugin.  A plugin is registered by adding
+    an entrypoint under the namespace ``girder.plugin``.  This entrypoint should return a
+    class derived from this class.
+
+    Example ::
+        class Cats(GirderPlugin):
+            '''
+            This string will be displayed to the user as a plugin description.
+            '''
+            URL = 'https://girder-cats.com'
+            DEPENDENCIES = ['animals']
+
+            def load(self, info):
+                super(Cats, self).load()
+                import rest  # register new rest endpoints
+    """
+    URL = ''
+    DEPENDENCIES = []
+    WEB_DEPENDENCIES = []
+
+    def __init__(self, entryPoint):
+        self._entryPoint = entryPoint
+        self._loaded = False
+
+    @property
+    def name(self):
+        """Return the plugin name defaulting to the entrypoint name."""
+        return self._entryPoint.name
+
+    @property
+    def description(self):
+        """Return the plugin description defaulting to the classes docstring."""
+        return self.__class__.__doc__ or ''
+
+    @property
+    def url(self):
+        """Return a url reference to the plugin (usually a readthedocs page)."""
+        return self.URL
+
+    @property
+    def version(self):
+        """Return the version of the plugin automatically determined from setup.py."""
+        return self._entryPoint.dist.version
+
+    @property
+    def dependencies(self):
+        """Return a list of plugins that this plugin requires."""
+        return list(self.DEPENDENCIES)
+
+    @property
+    def webDependencies(self):
+        """Return a list of plugins containing web client assets this plugin uses."""
+        return list(self.WEB_DEPENDENCIES)
+
+    @property
+    def loaded(self):
+        """Return true if this plugin has been loaded."""
+        return self._loaded
+
+    @abc.abstractmethod
+    def load(self, info):
+        """Execute any code necessary to load the plugin.
+
+        This method must be overridden by derived classes and call the superclass method.
+        """
+        self._loaded = True
+
+
+def _findPlugins():
+    """Iterate through entrypoints to discover installed plugins."""
+    global _pluginRegistry
+    if _pluginRegistry is not None:
+        return
+
+    _pluginRegistry = {}
+    for entryPoint in iter_entry_points(NAMESPACE):
+        pluginClass = entryPoint.load()
+        plugin = pluginClass(entryPoint)
+        _pluginRegistry[plugin.name] = plugin
+
+
+def _walkPluginTree(func, plugins, dependencyGetter, handled, nodes):
+    """Walk through the plugin dependency tree depth first.
+
+    This is a recursive function performing a specialized topological sorting of the
+    plugin dependency tree.  Each time a new plugin is encountered the provided function
+    is called on the plugin config object.  This function will be called at most once
+    for each plugin and if it raises any exception, an error message will be recorded and
+    the walk cancelled.
+
+    :param func: A function taking the plugin configuration object.
+    :param plugins: A list of plugin names to traverse.
+    :param dependencyGetter: A function that returns a list of dependencies of a plugin.
+    :param handled: A set of plugin names that have already been handled.
+    :param nodes: A set of plugin names in the current tree traversal.  This is used
+                  to detect cycles in the dependency graph.
+    """
+    # loop over a sorted list of plugins to ensure a consistent traversal order
+    for pluginName in sorted(plugins):
+        if pluginName in handled:
+            continue
+
+        # handle dependency cycles
+        if pluginName in nodes:
+            logprint.error('Cyclic dependencies encountered while processing plugins')
+            raise Exception('Cyclic dependencies encountered while processing plugins')
+
+        # generate a set of nodes traversed up to the root for cycle detection
+        pluginNodes = {pluginName}
+        pluginNodes.update(nodes)
+
+        plugin = getPlugin(pluginName)
+
+        # handle missing dependencies
+        if plugin is None:
+            logprint.error('ERROR: Plugin %s not found', pluginName)
+            raise Exception('Required plugin missing')
+
+        # traverse into sub-dependencies
+        try:
+            _walkPluginTree(func, dependencyGetter(plugin), dependencyGetter, handled, pluginNodes)
+        except Exception:
+            logprint.error('ERROR: Dependency failure while processing %s', pluginName)
+            _pluginFailureInfo[pluginName] = {
+                'traceback': '',
+                'message': 'Dependency error'
+            }
+            handled.add(pluginName)
+            raise
+
+        # call the handler function and record it
+        func(plugin)
+        handled.add(pluginName)
+
+
+def _getToposortedPlugins(plugins=None, dependencyGetter=None):
+    """Return a toposorted list of plugins with a custom child-node accessor.
+
+    :param plugins: A list of plugins to include (defaults to all plugins)
+    :param dependencyGetter: A function taking a plugin object and returning a list of dependencies
+    """
+    dependencyGetter = dependencyGetter or (lambda p: p.dependencies)
+    pluginNames = []
+    if plugins is None:
+        plugins = allPlugins()
+
+    def appendPluginName(plugin):
+        pluginNames.append(plugin.name)
+
+    _walkPluginTree(appendPluginName, plugins, dependencyGetter, set(), set())
+    return pluginNames
+
+
+def getToposortedPlugins(plugins=None):
+    """Return a toposorted list of plugins.
+
+    By default this function will return a toposorted list of all detected plugins.  If a list of
+    plugin names is provided, it will only return a minimal list of plugins required for the
+    given top-level plugins.
+    """
+    return _getToposortedPlugins(plugins)
+
+
+def getToposortedWebDependencies(plugins=None):
+    """Return a toposorted list of plugin web dependencies.
+
+    By default this function will return a toposorted list of all detected plugins.  If a list of
+    plugin names is provided, it will only return a minimal list of plugins required for the
+    given top-level plugins.
+    """
+    return _getToposortedPlugins(
+        plugins,
+        dependencyGetter=lambda p: p.webDependencies
+    )
+
+
+def getPlugin(name):
+    """Return a plugin configuration object or None if the plugin is not found."""
+    _findPlugins()
+    if name not in _pluginRegistry:
+        return None
+    return _pluginRegistry[name]
+
+
+def getPluginFailureInfo():
+    """Return an object containing plugin failure information."""
+    return _pluginFailureInfo
+
+
+def loadPlugin(name, root, appconf, apiRoot=None):
+    """Load a plugin and all of its dependencies in topological order."""
+    def callPluginLoad(plugin):
+        if not plugin or plugin.loaded:
+            return
+        info['dependencies'] = list(plugin.dependencies)
+        try:
+            plugin.load(info)
+        except Exception:
+            logprint.exception('ERROR: Failed to execute load method for %s', plugin.name)
+            _pluginFailureInfo[plugin.name] = {
+                'traceback': traceback.format_exc()
+            }
+            raise
+
+    info = {
+        'config': appconf,
+        'serverRoot': root,
+        'apiRoot': apiRoot
+    }
+    try:
+        _walkPluginTree(callPluginLoad, [name], lambda p: p.dependencies, set(), set())
+    except Exception:
+        return None
+    return getPlugin(name)
+
+
+def loadPlugins(names, root, appconf, apiRoot=None):
+    """Load a list of plugins and their dependencies in topological order."""
+    loadedPlugins = {}
+    for name in names:
+        pluginObject = loadPlugin(name, root, appconf, apiRoot)
+        if pluginObject:
+            loadedPlugins[name] = pluginObject
+    return loadedPlugins
+
+
+def allPlugins():
+    """Return a list of all detected plugins."""
+    _findPlugins()
+    return list(_pluginRegistry.keys())
--- a/girder/test/test_plugin_registry.py
+++ b/girder/test/test_plugin_registry.py
@@ -1,0 +1,315 @@
+###############################################################################
+#  Copyright Kitware Inc.
+#
+#  Licensed under the Apache License, Version 2.0 ( the "License" );
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+###############################################################################
+
+import re
+
+import mock
+import pytest
+
+from girder import _plugin as plugin
+
+
+@pytest.fixture
+def logprint():
+    with mock.patch.object(plugin, 'logprint') as logprintMock:
+        yield logprintMock
+
+
+def assertNoDuplicates(loadOrder):
+    __tracebackhide__ = True
+    assert len(loadOrder) == len(set(loadOrder))
+
+
+def assertPluginLoadOrder(plugin, loadOrder):
+    __tracebackhide__ = True
+
+    assertNoDuplicates(loadOrder)
+    pluginIndex = loadOrder.index(plugin.name)
+    for dep in plugin.dependencies:
+        depIndex = loadOrder.index(dep)
+        assert depIndex < pluginIndex
+
+
+def assertAllPluginsLoadOrder(loadOrder):
+    __tracebackhide__ = True
+
+    assertNoDuplicates(loadOrder)
+    for pluginName in loadOrder:
+        assertPluginLoadOrder(plugin.getPlugin(pluginName), loadOrder)
+
+
+class MockedPlugin(plugin.GirderPlugin):
+    _side_effect = None
+
+    def __init__(self, *args, **kwargs):
+        self._mockLoad = mock.Mock()
+        super(MockedPlugin, self).__init__(*args, **kwargs)
+
+    def load(self, *args, **kwargs):
+        super(MockedPlugin, self).load(*args, **kwargs)
+        self._mockLoad(*args, **kwargs)
+        if self._side_effect:
+            raise self._side_effect
+
+
+def mockPluginGenerator(deps, webDeps, url, doc, side_effect=None):
+    class GeneratedMockPlugin(MockedPlugin):
+        DEPENDENCIES = deps
+        WEB_DEPENDENCIES = webDeps
+        URL = url
+        _side_effect = side_effect
+        __doc__ = doc
+
+    return GeneratedMockPlugin
+
+
+class MockEntryPoint(object):
+    def __init__(self, name, version, pluginClass):
+        self.name = name
+        self.dist = mock.Mock()
+        self.dist.version = version
+        self.load = mock.Mock(return_value=pluginClass)
+        self.pluginClass = pluginClass
+
+
+def mockEntryPointGenerator(name, version='0.1.0', deps=None, webDeps=None, side_effect=None):
+    if deps is None:
+        deps = []
+    if webDeps is None:
+        webDeps = []
+
+    url = 'url for %s' % name
+    doc = 'doc for %s' % name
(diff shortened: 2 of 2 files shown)


girder/_plugin.py
```diff
@@ -0,0 +1,266 @@
...
+
+    @property
+    def name(self):
+        """Return the plugin name defaulting to the entrypoint name."""
+        return self._entryPoint.name
+
+    @property
+    def description(self):
+        """Return the plugin description defaulting to the classes docstring."""
+        return self.__class__.__doc__ or ''
```
brianhelba: Could this be taken from the description in `setup.py`?

girder/_plugin.py
```diff
@@ -0,0 +1,266 @@
...
+
+    @property
+    def url(self):
+        """Return a url reference to the plugin (usually a readthedocs page)."""
+        return self.URL
```
brianhelba: Could this be taken from the URL in `setup.py`?

girder/_plugin.py
```diff
@@ -0,0 +1,266 @@
...
+            '''
+            URL = 'https://girder-cats.com'
+            DEPENDENCIES = ['animals']
+
+            def load(self, info):
+                super(Cats, self).load()
+                import rest  # register new rest endpoints
+    """
+    URL = ''
+    DEPENDENCIES = []
```
brianhelba: Could this be expressed as ordinary `install_requires` in `setup.py`?

girder/_plugin.py
```diff
@@ -0,0 +1,266 @@
...
+        handled.add(pluginName)
+
+
+def _getToposortedPlugins(plugins=None, dependencyGetter=None):
+    """Return a toposorted list of plugins with a custom child-node accessor.
+
+    :param plugins: A list of plugins to include (defaults to all plugins)
+    :param dependencyGetter: A function taking a plugin object and returning a list of dependencies
+    """
+    dependencyGetter = dependencyGetter or (lambda p: p.dependencies)
```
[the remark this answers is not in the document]
jbeezley: I wanted to avoid exposing the `dependencyGetter` argument publicly because it exposes too much of the internal implementation that we may want to replace.

For the `plugins` argument, do you mean duplicate the
```python
if plugins is None:
    plugins = allPlugins()
```
in the public functions?  I could do that, but I don't see a real benefit.

girder/_plugin.py
jbeezley: It possible, but the implementation would be somewhat complicated.  The problem is that there isn't a one-to-one relationship between python packages and girder plugins.  A plugin can depend on a package that isn't a plugin, a package can contain multiple plugins, and a plugin can depend on a package that contains a plugin without actually depending on the plugin itself.

It would be possible as a default behavior to examine the entrypoints and get the packages that contain them to generate this list automatically for the typical use case.  Plugins requiring more complicated behavior could override it.  I'm undecided if doing this is a bad idea or not.

brianhelba: After in-person discussion with @jbeezley today, I expressed the opinion that:
> A plugin can depend on a package that isn't a plugin

This hopefully won't be an issue if we're able to delegate dependency tree resolution to pip, then only look at what ends up getting installed and having a girder plugin entry point (not necessarily caring how it was resolved).

> a package can contain multiple plugins

I think this should not be allowed. It needlessly complicates the situation, without offering any unique benefit (that I can think of, but I'd be happy to hear use cases).

> a plugin can depend on a package that contains a plugin without actually depending on the plugin itself

Currently, this is `staticWebDependencies`, and we only support it for plugins' web components. In the future, we could continue to support that same functionality by having a plugin depend on another in only its `package.json` (allowing it to `require` / ES6-`import` from the other), but not reference the other with a pip dependency. However, for both the Python and web worlds, it is probably best if developers reorganize their code to distribute common components in a non-Girder-requiring library package which all plugins can depend on directly.

girder/_plugin.py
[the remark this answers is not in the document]
kotfic: I was not suggesting exposing dependencyGetter in getToposortedPlugins or getToposortedWebDependencies.  Only that you might move the default functionality defined in _getToposortedPlugins  into getToposortedPlugins:

```python
def getToposortedPlugins(plugins=None):
     """Return a toposorted list of plugins.
 
     By default this function will return a toposorted list of all detected plugins.  If a list of
     plugin names is provided, it will only return a minimal list of plugins required for the
     given top-level plugins.
     """
     return _getToposortedPlugins(
         plugins,
         dependencyGetter=lambda p: p.dependencies
    )
```

Equally you might consider moving the "defaulting" functionality of ```plugins``` into the public functions. That way _getToposortedPlugin  accepts plain arguments rather than keyword arguments and is only responsible for doing its work.  This makes testing the "defaulting" behavior explicitly apart of the public API and an easier target for testing.  But also... it is just a nitpick :smile:

danlamanna:
Related: #2385

jbeezley:
I'm working on deriving plugin dependency information from setup.py, and I'm finding it a little awkward with the current approach.  I'm toying with the idea of making it the responsibility of plugins themselves to load dependencies as necessary.  This would look something like:
```python
# in plugin1 that depends on plugin2
def load(info):
    girder.plugin.findPlugin(plugin2).load(info)
   # load self
```
The load functions would include a mechanism preventing themselves from being called twice. This would allow plugins more flexibility in handling errors and even open the possibility for optional dependencies.  It would also remove a lot code in this PR related topological sorting and error handling.

The main issue with this approach is that it removes the ability to get a static list of toposorted plugins which is required for our web client build as it exists now.  Depending on where we go with it, this may be a non-issue.  I *think* the approach ultimately will be that the new equivalent of `girder-install web` will build *all* installed plugins rather than attempt to decipher a minimal list of requirements.  I also don't think it will require a topological sorting of the plugins because it won't involve multiple invocations of webpack for each plugin as we do now.  @brianhelba does this match your vision for the refactored client build?

danlamanna:
> I'm working on deriving plugin dependency information from setup.py, and I'm finding it a little awkward with the current approach.

What is awkward exactly?

> The main issue with this approach is that it removes the ability to get a static list of toposorted plugins which is required for our web client build as it exists now.

Doesn't this also prevent the UI from knowing which dependent plugins to enable?

jbeezley:
> What is awkward exactly?

It increases the complexity of `_walkPluginTree` logic because it not only needs to look for a list of plugins but also reference the packages those plugins are contained in.  This add a whole new layer of failure conditions that need to be handled.  The more I think about it, it seems like duplication of python's package dependency logic for no good reason.  Removing this logic also reduces the size of this PR **significantly**.

> Doesn't this also prevent the UI from knowing which dependent plugins to enable?

The UI doesn't really need to know the dependent plugins.  It would only enable the single plugin in the database.  When the server reloads, dependent plugins would load automatically.  When rerendering the plugin admin page, it would get the list of loaded plugins rather than attempting to get the list of "enabled" plugins in the database.  This could also satisfy @brianhelba's desire to distinguish between plugins that were explicitly enabled and those that were enabled by the dependency mechanism.

jbeezley:
I pushed a change that implements the new behavior.  I think it is dramatically better.  The only complicated bit is a function wrapper to the load method that handles logging and ensuring the load method is only called once.

I still have to finish the testing, but a plugin descriptor class would now look like this:
```python
from girder import plugin

class Cats(plugin.GirderPlugin):
   def load(self, info):
        plugin.getPlugin('Pets').load(info)  # load dependency

        try:
            plugin.getPlugin('Food').load(info)  # try to load an optional dependency
        except Exception:
             pass

        # get loaded plugins to generate optional behavior
        pluginList = plugin.loadedPlugins()

        # register events and endpoints as usual...
```
For now, all plugin information is determined through the python package, but we might need to add a cross reference to an npm package name for the client build.

brianhelba:
I like where @jbeezley is going, but let me restate somethings to help clarify my own thinking. It seems like there are 3 cases here:
1. A plugin depends on another's for assets (code, files, etc.), but doesn't need the other to actually run
2. A plugin depends on another to also run, but the load ordering doesn't matter
3. A plugin depends on another to run, and requires the other to load first

To satisfy 1, we:
* In general, refactor common assets into a non-Girder-plugin library package, and depend on that.
* In the specific case of client-side plugins (what we now call `staticWebDependencies`), depend on the other in `package.json` only. However, the library package approach should be encouraged where possible.

To satisfy 2, we:
* To the extent necessary, do the steps in 1
* Add the other plugin as a requirement in `setup.py`.

To satisfy 3, we:
* Do the steps in 2
* Add `plugin.getPlugin('OtherPlugin').load(info)` to the depending plugin's `load`. This has the added benefit of allowing the actual order of dependant plugin loading to be explicitly controlled, which I believe was non-deterministic before.
* **Open question**: how do we ensure a load order on the client side?
  * If we had a more traditional JS workflow, we'd just have the depending plugin `import OtherPlugin;` at the top of it's on entry file, forcing the dependent plugin to run first (Girder core would also have something like `import OtherPlugin;` / `require('OtherPlugin');`, but whichever ran second would be a no-op.
  * Since we rely on the HTML `<script>` tag ordering to determine load order, I'm not sure how we get the load order information to whatever builds the HTML (which is currently a Python-Mako loop, but will hopefully be Webpack in the future).

Note, a side effect of this is that we no longer distinguish between a plugin being "installed" and "enabled". If it's installed, its entry point triggers, and it gets enabled.

---

To address @jbeezley 's earlier question about my vision for the client build, my strong hope is that the client-side build process can something look like:
* `girder-init-web`: A Python script to generate a `package.json` script, containing pip-installed plugins as dependencies.
* ~`npm install`~ `yarn install`
* ~`npm run build`~ `yarn run build`: Use Grunt+Webpack to build JS and CSS files.

As a further dream extension, we could consider the merits of both using Webpack to generate HTML files (allowing cache-busting, hot reloading, and other goodies). Relatedly, we could get rid of code-splitting and the `Dll*Plugin` (allowing case 3 from above to be more easily satisfied), but I admittedly don't understand the current rationale for code-splitting well enough to judge whether this would be an improvement (let's discuss this bit in-person, since I don't want to derail this thread).

jbeezley:
> Open question: how do we ensure a load order on the client side?

That's a good question.  I'm tempted to make the claim that it girder shouldn't have to resolve the load order on the client in the same way that it isn't necessary on the server.  My plan for this branch is to minimize the changes to the client, which will probably mean adding a hacky solution just to get it to work.  I suspect this will involve explicitly resolving the load order in some way, but it should much simpler than what I was doing before.  I think I can get away with tracking the order from `load` wrapper in the metaclass.

> Note, a side effect of this is that we no longer distinguish between a plugin being "installed" and "enabled". If it's installed, its entry point triggers, and it gets enabled.

I'm not against this, but I don't think it is necessary to accomplish the goals of this PR.  It may be controversial, so we would have to discuss doing so with the larger group.

jbeezley:
This is about where I want it now.  I think I'm going to start on breaking changes from here on.

jbeezley:
I'm relatively happy with the current state of the plugin registry in regards to the server side infrastructure.  I have migrated the jobs and worker plugins to installable packages.  As of now, the biggest missing piece is related to the client build.  I have started experimenting by manually constructing a `webpack.config.js` and `package.json` in a staging directory.  Once I settle on a solution, I'll describe it here.

For the server components, I think things have progressed enough that we can start farming out tasks to interested parties.  These are a few things I think could be started now before issues with the client are solved.
1. Finish migrating the other plugins' server code.
2. Fix testing server components for plugins.
3. Think about deployment, release, and CI infrastructure.  This is a little premature, but it would be useful to have someone think about answers to questions like:
   1. How does this break ansible roles across the company?  Is there a simple migration strategy?
   2. Do these changes allow (or force) us to redesign how we do CI testing?
   3. How will we handle releases now that there are many more steps involved?

danlamanna:
I branched off of this to see how the server-side testing for plugins could work, see https://github.com/girder/girder/compare/pip-installable-plugins...pip-installable-plugins-pytest. Running `pytest` from the jobs directory there should work.

zachmullen:
@danlamanna I notice you actually got the model singletons to reload via 

```
for m in model_base._modelSingletons:
        del m.__class__._instance
```

When I do that I get a ton of weird errors that make it seem like the database is not getting dropped between test runs... did you encounter that also?

danlamanna:
> @danlamanna I notice you actually got the model singletons to reload via 
>
> ```
> for m in model_base._modelSingletons:
>         del m.__class__._instance
> ```
>
> When I do that I get a ton of weird errors that make it seem like the
> database is not getting dropped between test runs... did you encounter
> that also?

Are you talking about the Girder test suite or in the jobs plugin tests?
The Girder test suite is somewhat broken on this PR as a result of some
of the breaking changes.

When looking at just the jobs tests, doing something like this:
pip install pytest-repeat && pytest --count=2 test

Shows failures unless the model singleton fix is added.

zachmullen:
I was talking about the Girder test suite, which was breaking for me when I added the model singleton reset logic. I haven't tracked down exactly why, but the symptoms seem to indicate that after the models are reinitialized, all the tests are pointing to the same database rather than getting their own unique one.

jbeezley:
I just got an extremely preliminary, but minimally viable client build process working with plugins.  You can try the following (probably best in a fresh virtualenv):
```
pip install -e git://github.com/jbeezley/girder_plugin_jobs.git#egg=girder_plugin_jobs -e git://github.com/girder/girder.git@pip-installable-plugins#egg=girder
girder-build
girder-serve
```
That will install a ported version of the jobs plugin I put in an external repo to show that this all works without copying plugin or symlinking the code to any specific place.

There are a couple of technical details of note:
1. This creates a staging directory and static directory under `<sys.prefix>/share/girder`.  When executing `girder-build` it generates a `package.json` and symlinks the build infrastructure to the staging area.  It then runs `npm install` and `grunt`.  No `npm install` step is done inside grunt any more.  In a development environment, the client code is symlinked from the source directory, but in a production environment it will be fetched from npm.
2. The static files are served from under the built path in the `sys.prefix` rather than from inside girder's python package.
3. There is some ugliness in symlinking node packages as usual.  I found that including the node flag `--preserve-symlinks` to the grunt command makes it work as expected.

There are at least a couple of issues I need to address:
1. There needs to be a way to pass information from a plugin's client code to girder's build system to answer questions like "What is the entrypoint?" "What is the name of the plugin (as opposed to the package name)?" "Is there a webpack helper function?".  I'm thinking of adding a `girder` key in `package.json` for this purpose.
2. I want to remove grunt or at least reduce the webpack task to a single config with multiple entrypoints.  This will probably mean dropping the DLL plugin in favor of more traditional code splitting, but I haven't explored it much.  This will have the benefit that running webpack as a standalone script will work (including `webpack watch` :fireworks:).

I think my next step will be to try and port over candela, which has more complicated build requirements, to see if there are any specific issues with this approach.

zachmullen:
@jbeezley that sounds great! One change I would suggest is that any reference to `grunt` should be replaced with a call to an npm script, e.g. `npm run build` or `npm run girder_build`. Whenever we do the web client rewrite, we'll be abandoning grunt if all goes well.

zachmullen:
Ignore my comment about grunt, I think I misunderstood how the build was working. As long as it's just an implementation detail, I don't mind.

jbeezley:
I made some more progress on the client build including converting more plugins (`item_tasks` and `candela`).  In particular, I fixed DLL references between plugins and solved issues related to requiring 3rd party packages.  On the client build, there are still a few more issues unresolved:
* The `webpack.helper.js` hook allowing plugins to override the default webpack configuration.  (The `dicom` plugin will be a good candidate to work this out.  It turns out that with the changes I made, candela no longer needs it.)
* Add a flag to `girder-build` to enable webpack watch mode.
* Add some mechanism for client-side testing.  My preliminary plan is to copy (or link) the test files into the static root and load them from there rather than mounting the `/plugins` directory as it is done now.  This is the one major TODO that I don't have a clear idea of how to accomplish.

Just to keep track, these are some other random issues I have in mind that still need to be resolved:
* Add a hook in the testing infrastructure to register a "fake" plugin that isn't installed via entrypoint.  This will replace the current `testPlugin` marker.
* Clarify references to plugin names in the code, and unify how to cross-reference them.  Currently plugins are referenced by several different "names", which is sure to cause confusion if they aren't disambiguated.
  * The displayed name (`Remote Worker`).  This isn't implemented in the new infrastructure yet.
  * The entrypoint name (`worker`).  This is a valid javascript and python token used by both the client and the server to reference the plugin internally.  This is the name that is currently displayed on the "plugin" page.
  * The python package name (`girder-plugin-worker`)
  * The npm package name (`@girder/worker`)
* Define the procedure for publishing.  (Do we publish girder with all of the builtin plugins all at once or are plugins published on their own?)  Because there will now be many packages to publish, it will be important to make this process automated.
* Test a production deployment with packages fetched directly from pypi and npm.

jbeezley:
My latest push has added a pytest marker for enabling plugins at run time.  It also allows us to inject plugins temporary into the test context without installing them.  This is based on some of the mocks that I had in place in `test/test_plugin_registry.py` and replaces the `testPlugin` marker.  All pytest tests now pass.

More importantly, this adds necessary infrastructure for writing server-side tests in plugins.  As an example,
```python
@pytest.mark.plugin('jobs')
def testWithJobsPluginEnabled(server):
    # The jobs plugin will be loaded on server start
```

jbeezley:
In the last commit, I started on a [migration guide](https://github.com/girder/girder/blob/f633a10340ea14b452e07e3a877961549c9d04c1/docs/migration-guide.rst) for plugin maintainers.  I'm sure I forgot many breaking changes, but this should be enough for core Girder developers to understand how to migrate builtin plugins.

jbeezley:
@girder/developers I'm going to be pushing on this branch again thanks to @aashish24 approving the use of GUMBO funding.  I'll continue describing my progress here as I work on it.  Hopefully, I will be able to get it into a merge-able state in the next month.

My latest commit adds a massive diff that will be difficult to revert later, so it would be good for interested parties to look at it now.  In summary, it moves all of the web assets inside of the python packages.  See the commit message for details.

jcfr:
> Remove unsupported plugins

Do you also plan to create repository for there two plugins ? 

* metadata extractor (to be migrated out of core)
* geospatial (to be migrated out of core)

I think it would be useful to keep an entry in the documentation indicating when each plugin was removed, or how to get similar functionality.

jbeezley:
> Do you also plan to create repository for there two plugins ?

The geospatial plugin is currently being rewritten here at <https://github.com/OpenGeoscience/girder_geospatial>.  I'll probably use the metadata_extractor plugin as a test case for an "external plugin".  For now, I wanted to remove them so I don't forget and waste time porting them.

> I think it would be useful to keep an entry in the documentation indicating when each plugin was removed, or how to get similar functionality.

That's a good point.  I'll put in a place holder for now.

jbeezley:
@girder/developers I just removed the WIP tag on this PR, which means... I think it just might be ready.  I just made push to update the documentation, which as we all know could use improvement.  I hope at least the new [plugin development](https://github.com/girder/girder/blob/7b6c214c56c1f73be4a27bbc9f0ef4ba3693c218/docs/plugin-development.rst) and [migration guide](https://github.com/girder/girder/blob/7b6c214c56c1f73be4a27bbc9f0ef4ba3693c218/docs/migration-guide.rst#2x--3x) docs are sufficient for core developers to understand.

There are a couple of outstanding TODO's in the code that are mostly related to the unittest -> pytest transition rather than specific to this branch.  Before we release Girder v3, we should make an effort to port all server side tests to pytest and deprecate the CMake test harness, but that is out of scope here.  We have our quarterly PR cleanup tomorrow morning.  I'm hoping we can discuss this PR and more broadly opening the merge window for Girder 3 breaking changes.

jbeezley:
I migrated covalic to work with this branch to see how it works with an external plugin.  The results are in https://github.com/girder/covalic/pull/249.

kotfic:
I was able to get this up and running with almost no disruption to the traditional workflow,  plugin enable/disable appears to be working fine all of which is very exciting.  I will take a look at converting a non-trivial plugin over next.

The one hiccup I encountered was installing girder with the "plugins" extra fails because it can't find girder-table-view on pypi.  Are we planning on adding CI steps that push "default" (or "core" or what are we calling these now?)  plugins out to the proper distribution mechanisms?

jbeezley:
Yes, we should release the builtin plugin packages to pypi once this is finalized.  You got the error with `girder-table-view` because I forgot to add it to `requirements-dev.txt`.

The issue of local dev installs vs. installs from pypi is probably going to be the biggest source of problems.  If you want a dev install of all plugins from the git repo, you have to ensure that they are installed in topological order.  This is what `requirements-dev.txt` does for you, but if you install anything else first, pip will ignore the request to install from the local package.

jcfr:
This will help increase adoption of Girder. Thanks for working on this @jbeezley :+1: 

This should probably be mentioned in the SciPy talk at the plenary session.

Cc: @thewtex @mgrauer

jbeezley:
For reviewers, here are a few commands to get started.  Make sure you start with a fresh virtualenv.  Upgrading on old Girder installation probably won't work.

```bash
# base install with no plugins
pip install -e .
girder build # see --help for options
girder serve

# install plugins individually (dependent plugins *must* be installed first to prevent pip from attempting an install from PyPI)
pip install -e . -e plugins/jobs
girder build
girder serve

# install all plugins + dev dependencies for running the tests
pip install -e . -r requirements-dev.txt
girder build --dev
```

jbeezley:
I just found an issue with our CI config that isn't easy to solve, and also not really related to this branch.  I'll quote my last commit message:

> This is not ideal.  For one, the cache should be invalidated if any
> plugin setup.py changes.  Second, we should not be caching the
> virtualenv's containing editable installs.
>
> The issue this fixes is related to pypa/pip#4537
>
> The downgrade of celery from #2761 triggers a behavior (maybe a bug?) in
> pip that only occurs when the "egg-info" directory for an editable
> install does *not* exist.  A second invocation of `pip install` will
> run successfully.
>
> To reproduce this problem outside of CI, in a fresh virtualenv:
> ```
> pip install -e . -e ./plugins/jobs -e ./plugins/remote_worker
> pip install celery==4.2
> rm -fr ./plugins/remote_worker/*.egg-info
> pip install -e ./plugins/remote_worker
> ```
>
> One alternative option is to turn off caching of the virtualenv.

jcfr:
> I just found an issue with our CI config that isn't easy to solve

I wonder if pipenv would avoid such issue ... https://pipenv.readthedocs.io/en/latest/

.circleci/config.yml
```diff
@@ -162,9 +162,15 @@ jobs:
           command: echo ". $CIRCLE_WORKING_DIRECTORY/girder_env/bin/activate" >> $BASH_ENV
       - restore_cache:
           key: npm-{{ arch }}-{{ checksum "girder/package-lock.json" }}
+      - run:
+          name: Install top level node modules
```
brianhelba: Can you change the name to make it clear that this is installing the `@girder/lint` package?

girder/plugin.py
```diff
@@ -0,0 +1,263 @@
...
+    #: The name of the web client package on npm.
+    NPM_PACKAGE_NAME = None
+
+    #: The version of the web client package to use.  Accepts any npm version identifier.
+    NPM_PACKAGE_VERSION = '*'
+
+    #: The path of the plugin's web client source code.  This path is given relative to the python
+    #: package.  This property is used to link the web client source into the staging area while
+    #: building in development mode.
+    CLIENT_SOURCE_PATH = 'web_client'
```
brianhelba: Could we maybe make `CLIENT_SOURCE_PATH` the primary determinant of whether a package has a web client location? This would have two benefits:
* Allow both `NPM_PACKAGE_NAME` and `NPM_PACKAGE_VERSION` to be hidden and unimportant to most developers.
* Allow the `CLIENT_SOURCE_PATH` path to be more transparent to all developers, by making it explicitly specified and no longer have a magical default.

The result is that developers still need to specify one property (now `CLIENT_SOURCE_PATH` instead of `NPM_PACKAGE_NAME`), but Girder takes care of figuring out `NPM_PACKAGE_NAME` itself from the plugin's `package.json` file.

jbeezley:
As planned, we are on track for merging this branch tomorrow (Sept. 19).  As mentioned in other channels, I have [moved](https://github.com/girder/girder_worker/pull/311) the `remote_worker` plugin over to the `girder_worker` package.  Because we have other plugins in this repo that depend on the worker plugin, I had to ensure that we [install a specific branch](https://github.com/girder/girder/blob/4f0f23e8c765f6310f3b5fb9ca12d79bfe6fcec0/requirements-dev.txt#L5) in the testing environment.  This can (and should) go away later on in the merge window once we have prereleases deployed.

Speaking of prereleases, I will plan on generating publishing prelease packages for everything in this repository before merging to catch any last minute issues with the release process.  This is probably the last opportunity to change package names before they are finalized.  The new python packages added follow a consistent naming schema based on the entrypoint they register:
```
girder-audit-logs
girder-authorized-upload
girder-autojoin
girder-candela
girder-curation
girder-dicom-viewer
girder-download-statistics
girder-google-analytics
girder-gravatar
girder-hashsum-download
girder-hdfs-assetstore
girder-homepage
girder-item-licenses
girder-item-tasks
girder-jobs
girder-ldap
girder-oauth
girder-provenance
girder-table-view
girder-terms
girder-thumbnails
girder-treeview
girder-user-quota
girder-virtual-folders
```
Existing plugins not in that list have been temporarily removed, but may be added back (or moved to a new repository) before finalizing the 3.0 release.

jeffbaumes:
Looks good. Only suggestion is to rename `girder-provenance` to `girder-metadata-history` (`girder-metadata-provenance` would be fine too if someone likes that any better). We may want to reserve the name `girder-provenance` for something more robust, such as implementing the [PROV standard](https://www.w3.org/2001/sw/wiki/PROV).

PR #2554, opened 2018-01-04
Revert #2542 and use the latest flake8-docstrings

jbeezley:
There was just a new release of flake8-docstrings that fixes the error
that #2542 addressed.  This reverts that change and requires the latest
release with the fix in place.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,14 +2,11 @@
 coverage
 flake8
 flake8-blind-except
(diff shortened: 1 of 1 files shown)


brianhelba:
@jbeezley Have you tested this with a downstream project?

jbeezley:
Yes, it was causing errors in large_image.  It no longer throws an exception, but now raises a style error.

PR #2560, opened 2018-01-09
Run Tox tests from packaged and installed Girder

brianhelba:
This utilizes Tox's default behavior of building and installing packages to be tested. This will provide significantly greater assurance that Girder works when installed from an sdist package.

* Move Pytest "test" directory out of the "girder" pacakge
* Update Pytest's configuration to point to the new location for tests
* Prevent the "test" directory from being a Python module
  * This affects the way that the 'python path' is built by Pytest, per https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code
* Move test plugins to "test/test_plugins"
* Update legacy tests to use new "test_plugins' location
* Update Pytest tests to use new 'test_plugins' location and multiple markers
* Run Tox tests from packaged and installed Girder
* Remove "packaging_verify_contents" test
  * This is now implicitly tested by Tox using the installed package.
* Include Python coverage from Tox-installed packages
* Include auto-reformatng of package.json


brianhelba:
For future reference, it is critical that an `__init__.py` file **not** be added under the `test/` directory (or wherever Pytest tests are stored). This would cause Pytest to switch to a different import behavior, whereby the repository "girder" package would be found by `import girder` instead of the installed "girder" package.

However, we don't want to just add a test to assert that the "girder" being imported from the test environment is the installed version, as this would fail when Pytest is run natively outside of Tox.

Finally, in the future, if making the test directory into a module (i.e. adding `__init__.py`) is desired, then the "girder" package directory could be moved to "src/girder" instead. However, this is likely a more controversial change (though it should not raise any compatibility issues).

For more information, see:
* https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code
* https://hynek.me/articles/testing-packaging/
* https://blog.ionelmc.ro/2014/05/25/python-packaging/

brianhelba:
@girder/developers This is still a WIP because I need to figure out where we want to put test plugins, and how we want to resolve them. Regardless, it is also apparent that even without this change, the current mechanism of `pytest.mark.testPlugins` will not work once "pytest_girder" is distributed and imported by plugins in an independent testing context, as the current `test_plugins` are not included in the installed "girder" package.

zachmullen:
> Move Pytest "test" directory out of the "girder" pacakge

I don't feel comfortable approving this change since I wasn't responsible for the original decision. @danlamanna is probably the right person to review this concept.

> Remove "packaging_verify_contents" test \ This is now implicitly tested by Tox using the installed package.

I'm in favor of this in theory. Is there a reason to keep the rest of the packaging tests around after this?

danlamanna:
> I don't feel comfortable approving this change since I wasn't responsible for the original decision. @danlamanna is probably the right person to review this concept.

I had discussed moving them outside of the package with @jbeezley but we never settled on anything. The initial reason for moving them into the package was so downstream could use our testing utilities, however this is now covered by the `pytest-girder` package.

jbeezley:
Yes, the primary motivation for including tests *inside* the package being tested is so that they can import each other to share code.  This avoids the situation where you have to add something to PYTHONPATH to make it work.  Having the `pytest-girder` package mitigates this issue because shared code should just placed there.

brianhelba:
@zachmullen Yes, my goal is to remove all the packaging tests eventually. This PR removes some of them, but there's still some ”smoke test" functionality of our CLIs that the packaging tests have. In future work, we can port that elsewhere or determine that it's already adequately covered elsewhere.

zachmullen:
Sounds good.

setup.cfg
```diff
@@ -74,4 +74,4 @@ ignore: D100,D101,D102,D103,D104,D105,D106,D107,D200,D201,D202,D203,D204,D205,D4
 [tool:pytest]
 addopts = --verbose --strict --showlocals --cov=girder --cov-report=""
 cache_dir = build/test/pytest_cache
-testpaths = girder/test
```
danlamanna: The `MANIFEST.in` will also have to be updated.

danlamanna:
> For future reference, it is critical that an __init__.py file not be added under the test/ directory (or wherever Pytest tests are stored). This would cause Pytest to switch to a different import behavior, whereby the repository "girder" package would be found by import girder instead of the installed "girder" package.

@brianhelba Did you discover how nesting directories underneath test is handled?

brianhelba:
@danlamanna @girder/developers PTAL now.

brianhelba:
I'm seeing server coverage drop for some reason here. I'll investigate more, but please let me know if you spot any issues.

brianhelba:
@danlamanna I never fully explored sub-folder modules in `test/`. Feel free to look at it yourself, but I may not have a chance to dig into it under this PR.

brianhelba:
@danlamanna I've fixed the coverage issues with an additional commit. PTAL.

setup.cfg
```diff
@@ -3,8 +3,13 @@ data_file = build/test/coverage/python_temp/.coverage
...
 omit =
     girder/external/*
-    girder/test/*
+    test/*
 parallel = True
+[coverage:paths]
+# Include sources from installed package in Tox's {envsitepackagesdir}
+girder =
+    girder/
+    build/test/tox/*/lib/*/site-packages/girder/
```
danlamanna: Just to clarify the intent here - this is so that running `pytest` will always work and generate coverage for the currently checked out version of Girder, and running `tox` will always work and generate coverage against the installation of the currently checked out version of Girder.

brianhelba: Yes, that's the correct intent.

The mechanism for doing so is actually the `--cov={envsitepackagesdir}/girder` argument to `pytest` in `tox.ini`. This particular config setting just ensures that Coverage's reports map all files back to the correct paths (so it will have no effect if running from the uninstalled code).

Let me know if you'd like to see additional code documentation around this, in case its non-intuitive from reading the config files.

PR #2562, opened 2018-01-09
Swagger doesn't recognize application/zip as binary data.

### LINKED ISSUE #2561: Swagger UI generates corrupted zip file
As we know, when an item has multiple files, the item/download will produce a zip file. However, using the download link provided in the Swagger UI will produce a corrupted zip file. 
@manthey found out it is because there are extra bytes for UTF-8 being inserted into the file between chunks.

manthey:
Swagger injects BOM markers to non-binary data when it buffers it for downloading.  Since Swagger 2.x only recognizes images, pdfs, and application/octet-stream as binary, any other content dispositions can be corrupted by this process.  Change the value stored in the `produces` field for Swagger so that any data that might be binary (notably zip files) is specified as application/octet-stream.

Fixes #2561.

### DIFF
--- a/girder/api/describe.py
+++ b/girder/api/describe.py
@@ -28,6 +28,7 @@ import jsonschema
 import os
 import six
 import cherrypy
+from collections import OrderedDict
 
 from girder import constants, events, logprint
 from girder.api.rest import getCurrentUser, getBodyJson
@@ -135,7 +136,17 @@ class Description(object):
             resp['consumes'] = self._consumes
 
         if self._produces:
-            resp['produces'] = self._produces
+            # swagger has a bug where not all appropriate mime types are
+            # considered to be binary (see
+            # https://github.com/swagger-api/swagger-ui/issues/1605).  If we
+            # have specified zip format, replace it with
+            # application/octet-stream
+            #   Reduce the list of produces values to unique values,
+            # maintaining the order.
+            produces = list(OrderedDict.fromkeys([
+                'application/octet-stream' if item in ('application/zip', )
+                else item for item in self._produces]))
+            resp['produces'] = produces
 
         if self._deprecated:
             resp['deprecated'] = True


girder/api/v1/collection.py
```diff
@@ -113,7 +113,9 @@ def getCollectionDetails(self, collection):
         .modelParam('id', model=CollectionModel, level=AccessType.READ)
         .jsonParam('mimeFilter', 'JSON list of MIME types to include.', requireArray=True,
                    required=False)
-        .produces('application/zip')
+        # Tell swagger that we produce binary data; it doesn't treat
+        # application/zip as binary
+        .produces('application/octet-stream')
```
zachmullen: I think we should keep this as `application/zip` since that's the correct content-type. Seems like we can fix this by just adding some logic within `produces` itself that converts internally to `application/octet-stream` if the value doesn't match a whitelist. That way, once swagger fixes this bug, we can remove that logic in a central location and everything should work.

manthey: Yes, that is a better way to do this.

girder/api/describe.py
```diff
@@ -135,7 +136,15 @@ def asDict(self):
...
         if self._produces:
-            resp['produces'] = self._produces
+            # swagger has a bug where not all appropriate mime types are
+            # considered to be binary (see
+            # https://github.com/swagger-api/swagger-ui/issues/1605).  If we
+            # have specified zip format, replace it with
+            # application/octet-stream
+            produces = list(OrderedDict.fromkeys([
+                'application/octet-stream' if item in ('application/zip', )
+                else item for item in self._produces]))
```
zachmullen: Why `list(OrderedDict())`?

manthey: I wanted to remove duplicates from the list but not change its order.  This is the most compact pythonic way I know to do that.

zachmullen: Can you add a comment that expresses that intention? Otherwise someone might come along and not realize you are trying to preserve the order and change it to use `set()` or something.

zachmullen:
Let's squash this so the second commit doesn't contain a revert of the first, then I will approve.

PR #2563, opened 2018-01-09
WIP: Use ESDoc2, to fix package security warnings

brianhelba:
ESDoc2 is a fork of ESDoc with more updated dependencies, including the "marked" package, which is causing security warnings.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,12 @@
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
+        "@types/node": {
+            "version": "9.3.0",
+            "resolved": "https://registry.npmjs.org/@types/node/-/node-9.3.0.tgz",
+            "integrity": "sha512-wNBfvNjzsJl4tswIZKXCFQY0lss9nKUyJnG6T94X/eqjRgI2jHZ4evdjhQYBSan/vGtF6XVXPApOmNH2rf0KKw==",
+            "dev": true
+        },
         "abab": {
             "version": "1.0.4",
             "resolved": "https://registry.npmjs.org/abab/-/abab-1.0.4.tgz",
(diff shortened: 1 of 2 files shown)


brianhelba:
It looks like the versions of ESDoc2 that are on GitHub vs npm are a bit misleading. This PR isn't ripe yet.

manthey:
In GeoJS I explicitly added the latest version of `marked` to `packages.json`; the modules that used it then used this version and the security warning went away.  Although it adds a package that shouldn't be an explicit dependency to the package list, it does eliminate GitHub's security warning (and, assumably, reduces some potential XSS issue).

brianhelba:
@manthey I don't think is approach will be effective in Girder.

In GeoJS, the dependency on an insecure version of `marked` is coming from `jsdoc`. [Your GeoJS PR](https://github.com/OpenGeoscience/geojs/pull/756) upgrades `jsdoc` from `^3.4` to `^3.5`, but both those versions of `jsdoc` depend on marked in the same way: as `~0.3.6` (see [here](https://github.com/jsdoc3/jsdoc/blob/3.4.3/package.json#L22) and [here](https://github.com/jsdoc3/jsdoc/blob/3.5.5/package.json#L22)), which currently resolves to `0.3.12` (where the security issue is fixed). Thus, even without the PR, new installs of GeoJS would not have installed the insecure `marked`; the real value of the PR is to refresh `package-lock.json`, which prevents development installs from not being pinned to the older insecure versions.

Girder's case is not quite the same, as Girder uses `esdoc`, where even the latest version (1.0.4) [depends on `marked` at exactly `0.3.6`](https://github.com/esdoc/esdoc/blob/cf2c92b2d97e211284437fee3412f29b68d40d40/package.json#L34)(which is an insecure version). Adding a newer version of `marked` as an explicit dependency in Girder's `package.json` would not resolve the issue, because, unlike pip, npm allows multiple versions of a dependency to be installed simultaneously; thus, we'd just be pointlessly installing a newer version of `marked`, but also installing the older version, and `esdoc` should be given the older version when it `require`d it at runtime (as this is a "feature" of npm).

The solution in this sort of case is to either push for the upstream package to update or (ideally) loosen their dependencies, stop using the upstream package, or tolerate the security warnings. There is no way to force npm to do what we really want in this case.

The good news is that the security warnings don't indicate any actual security vulnerability in Girder, as `esdoc` is only run at development build-time, so an XSS vulnerability would only allow a malicious JS code comment to add code to the output HTML documentation, which we are not worried about (as the ability to alter Girder's JS code comments implies the ability to just alter Girder's JS code itself).

zachmullen:
This is a case where a warning is not indicative of an error. Hopefully GitHub allows us to resolve this warning so it doesn't appear in the UI while still showing us other ones that happen in the future.

mgrauer:
Never getting fixed. Also the state of auto-doc generation in JS is sad.

PR #2564, opened 2018-01-10
Modify ansible role to upgrade setuptools as well as pip

danlamanna:
The version of setuptools on our default deployment targets was too
old to handle the changes introduced in #2519.

### DIFF
--- a/devops/ansible/roles/girder/tasks/pip.yml
+++ b/devops/ansible/roles/girder/tasks/pip.yml
@@ -1,10 +1,13 @@
 - block:
     - name: Update Pip
       pip:
-        name: pip
+        name: "{{ item }}"
         state: latest
         virtualenv: "{{ girder_virtualenv }}"
         virtualenv_python: "{{ girder_python | default(ansible_python.executable) }}"
+      with_items:
+        - pip
+        - setuptools
 
     - name: Install Girder and plugin requirements
       pip:
@@ -21,9 +24,12 @@
 - block:
     - name: Update Pip
       pip:
-        name: pip
+        name: "{{ item }}"
         state: latest
         executable: "{{ girder_pip | default(omit) }}"
+      with_items:
+        - pip
+        - setuptools
 
     - name: Install Girder and plugin requirements
       pip:


brianhelba:
LGTM.

However, now that I think about it, an inline comment explaining why we update these would be good.

devops/ansible/roles/girder/tasks/pip.yml
```diff
@@ -1,10 +1,13 @@
...
     - name: Update Pip
       pip:
-        name: pip
+        name: "{{ item }}"
         state: latest
         virtualenv: "{{ girder_virtualenv }}"
         virtualenv_python: "{{ girder_python | default(ansible_python.executable) }}"
+      with_items:
+        - pip
+        - setuptools
```
manthey: Why is using `with_items` preferable to a list in the `name` field? (I have no issue, I am just curious)

danlamanna: I didn't even realize a list in the name field was possible, apparently it was added specifically to the `pip` module in 2.2 (note for instance the `apt` module doesn't have an analog).

I actually think the `pip` module shouldn't allow this, because Ansible goes to great lengths to ensure the return values are useful. For instance it can successfully indicate which packages changed state[1], it can return the result output of each invocation inside an array using the `register` syntax, etc. The only reason I can think of for the `pip` module allowing this is because `pip install x y` can result in different things than `pip install x; pip install y` can.

[1] It looks like the `pip` module with a list of names just blanket reports whether *any* of the desired packages had their state changed, see https://github.com/ansible/ansible/commit/f874c29dff30875d2b8513138537d8247927aff3#diff-62bf642191d16cdc80c0db31388c8cdfR411.

manthey: ansible *does* allow it with `apt` (though it doesn't say a list is allowed in http://docs.ansible.com/ansible/latest/apt_module.html).  Maybe it shouldn't.  I had the vague idea that it was somewhat faster to use the list form than the with_items form, since it would call pip once rather than several times.

brianhelba: In deployment, I'd happily trade speed for more granular error reporting.

PR #2565, opened 2018-01-10
Force Girder to bind to localhost instead of 0.0.0.0 by default

danlamanna:


kotfic:
This will break systems where girder runs in docker containers and port forwards to other systems.  See: https://github.com/girder/girder_worker/pull/233 for how it affected girder worker's test infrastructure

msmolens:
Following this change, Girder is no longer accessible at http://localhost:9080/ after running a fresh `vagrant up`. The documentation at http://girder.readthedocs.io/en/latest/dev-installation.html#virtual-machine could suggest to modify `girder.local.cfg`, or perhaps some configuration could happen automatically in a vagrant deployment.

danlamanna:
> perhaps some configuration could happen automatically in a vagrant
> deployment.

This is definitely the way to go, I'm taking a look at our
Vagrant/Ansible configuration currently so I'll fix this.

PR #2566, opened 2018-01-11
Alter girder_user to only control the system daemon

danlamanna:
This PR changes `girder_user` to only be in control of the daemon 
which Girder runs under. This means that users deploying with a 
sudo-user (necessary due to system packages, system daemons, etc) 
can now run Girder under an unprivileged system user.

This is part of an effort to remove all unnecessary usage of escalating 
to root in our ansible roles.

### DIFF
--- a/devops/ansible/roles/girder/tasks/main.yml
+++ b/devops/ansible/roles/girder/tasks/main.yml
@@ -18,11 +18,7 @@
   file:
     path: "{{ girder_path }}"
     state: directory
-    group: "{{ girder_user|default(ansible_user_id) }}"
-    owner: "{{ girder_user|default(ansible_user_id) }}"
     mode: 0755
-  become: yes
-  become_user: root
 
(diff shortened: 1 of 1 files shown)


zachmullen:
> Since the Girder daemon must self-modify its files when rebuilding the web root at runtime

What are you referring to here?

brianhelba:
> What are you referring to here?

[`/system/web_build`](https://github.com/girder/girder/blob/master/girder/api/v1/system.py#L491)

zachmullen:
If that is a security concern for a deployment, I would recommend *not* granting write permission to the daemon user, which would force all rebuilds to happen via the deploy user. We could provide a more graceful failure message for those attempting to rebuild at runtime by testing permission on the directory prior to starting the subprocess.

danlamanna:
This PR is going to be superseded by a different branch.

PR #2540, opened 2017-12-19
Make MongoDB 3.2 the minimum supported version

brianhelba:
* Make MongoDB 3.2 the minimum supported version
* Run CI with MongoDB's ephemeralForTest storage engine
  * There is significant variability in the time taken by the "serverTest" CI steps, but running with this flag on should be theoretically faster, and has empirically been 0.5-2 minutes faster.

Part of #2516

PR #2570, opened 2018-01-15
Add support for --host argument to girder-server

danlamanna:
This will allow certain cases (such as Docker containers) to override the host at the CLI level instead of in configuration files. This should help ease some of the pain introduced by #2565 (@kotfic).

This PR also upgrades the `docker-compose.yml` to use the new argument, and upgrades the version of Mongo to our minimally supported version 3.2 (see #2540).


zachmullen:
Thanks @danlamanna

PR #2571, opened 2018-01-16
Issue log message on upload

zachmullen:
This will allow for better log-based forensics for files uploaded into Girder.

### DIFF
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -21,7 +21,7 @@ import datetime
 import six
 from bson.objectid import ObjectId
 
-from girder import events
+from girder import events, logger
 from girder.api import rest
 from girder.constants import SettingKey
 from .model_base import Model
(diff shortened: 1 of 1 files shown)


girder/models/upload.py
```diff
@@ -243,6 +243,9 @@ def finalizeUpload(self, upload, assetstore=None):
         if '_id' in upload:
             self.remove(upload)
 
+        logger.info('Upload complete. Upload=%s File=%s User=%s' % (
+            upload['_id'], file['_id'], upload['userId']))
```
manthey: Should we also add the size and/or the name of the file?  When trying to determine files of interest, these could help.

zachmullen: > I didn't include any actual metadata about the file itself in the log, it seems more useful to just retain its ID and we could lookup the full metadata from the database via a script if needed.

Since much of the metadata is mutable, I figured we would be better off just using this as a way to do lookups into the database. Probably any analysis we'll want to perform will involve data munging from disparate sources.

PR #2572, opened 2018-01-17
Drop support for Python3 <3.5

### LINKED ISSUE #2516: Officially drop support for Python3 <3.5 and MongoDB <3.2
Officially drop support for Python3 <3.5 (but keeping Python 2.7), because:
* It's the [minimum version supported by Ansible](https://docs.ansible.com/ansible/2.4/dev_guide/developing_python3.html#minimum-version-of-python-3-x-and-python-2-x) (and Ansible will fail on Python <=3.4)
* Girder is not able to fully test Python 3.4, as the CircleCI machine environment requires Ansible to provision itself
* It's the [default Python on the current Ubuntu LTS (16.04)](https://packages.ubuntu.com/zesty/python3)
  * Ubuntu 14.04 ships with Python 2.7 as the default interpreter, which Girder will continue to support
* It's installable on nearly any platform [via `pyenv`](https://github.com/pyenv/pyenv)

Also, officially drop support for MongoDB <3.2 because:
* MongoDB 2.7 dropped [support on October 2016](https://www.mongodb.com/support-policy)
* MongoDB 3.2 contains [a number of desirable features](https://docs.mongodb.com/v3.2/release-notes/3.2/) that are not available in 3.0:
  * WiredTiger storage engine enabled by default
  * The `$lookup` (aka JOIN) aggregation stage (which will be critical if Girder is ever going to do scalable access filtering in-database)
  * Indexes on aggregation stages
  * Improved support for case-insensitive search (allowing Girder to eventually drop `lowerName`)

If this moves forward, we'll update the Girder documentation and test environment to use these versions as their minimums. No other changes should be necessary.

brianhelba:
Per the rationale in #2516, this makes Python 3.5 the minimum version of Python3 for Girder.

* Remove documentation references to Python 3.4
* Remove Python 3.4 from Tox
* Add package metadata to specify compatible Python versions
* Simplify code checks for Python versions <3.5
* Fix test_hash_state to work on Python 3.6
* Prevent test_hash_state from printing to stdout on import
* Identify the repo root for Codecov client, to silence warnings
* Improve variable naming in test_hash_state
* Clarify docs on supported Python 3 versions
* Do not install "software-properties" packages in Docker
  * It's unknown what these packages do, but their absence does not prevent Girder from building and running in Docker.
* Upgrade to Node 8 and Debian Stretch for the Docker base image
  * This makes the Docker "py3" image use Python 3.5.
* Fix a Pip version conflict in the Docker base image

Fixes #2516

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,7 @@ EXPOSE 8080
 RUN mkdir /girder
 RUN mkdir /girder/logs
 
-RUN apt-get update && apt-get install -qy software-properties-common python-software-properties && \
-  apt-get update && apt-get install -qy \
+RUN apt-get update && apt-get install -qy \
     build-essential \
     git \
     libffi-dev \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:6
+FROM node:8-stretch
 MAINTAINER Kitware, Inc. <kitware@kitware.com>
 
 EXPOSE 8080
@@ -31,4 +31,4 @@ COPY README.rst /girder/README.rst
 RUN pip install -e .[plugins]
 RUN girder-install web --all-plugins
 
-ENTRYPOINT ["python", "-m", "girder"]
+ENTRYPOINT ["python2", "-m", "girder"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN apt-get update && apt-get install -qy \
     libffi-dev \
     libsasl2-dev \
     libldap2-dev \
-    libpython-dev && \
+    libpython2.7-dev && \
   apt-get clean && rm -rf /var/lib/apt/lists/*
 
 RUN wget https://bootstrap.pypa.io/get-pip.py && python get-pip.py
@@ -28,7 +28,7 @@ COPY setup.py /girder/setup.py
 COPY package.json /girder/package.json
 COPY README.rst /girder/README.rst
 
-RUN pip install -e .[plugins]
+RUN pip install --upgrade --editable .[plugins]
 RUN girder-install web --all-plugins
 
 ENTRYPOINT ["python2", "-m", "girder"]
--- a/docs/deploy.rst
+++ b/docs/deploy.rst
@@ -244,7 +244,7 @@ Elastic Beanstalk
 
 Girder comes with pre-packaged configurations for deploying onto Elastic Beanstalk's
 `Python platform <http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/concepts.platforms.html#concepts.platforms.python>`_
-(both 2.7 and 3.4).
+(both 2.7 and 3.6).
 
 The configurations live within ``devops/beanstalk`` and are designed to be copied into your working Girder directory
 at deploy time.
--- a/docs/prerequisites.rst
+++ b/docs/prerequisites.rst
@@ -19,7 +19,7 @@ recommend setting up `Postfix <http://www.postfix.org/documentation.html>`_.
 See the specific instructions for your platform below.
 
 .. note:: We perform continuous integration testing using Python 2.7 and Python 3.5.
-   The system *should* work on other versions of Python 3 as well, but we do not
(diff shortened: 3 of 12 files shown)


docs/deploy.rst
```diff
@@ -244,7 +244,7 @@ Elastic Beanstalk
 
 Girder comes with pre-packaged configurations for deploying onto Elastic Beanstalk's
 `Python platform <http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/concepts.platforms.html#concepts.platforms.python>`_
-(both 2.7 and 3.4).
+(both 2.7 and 3.6).
```
brianhelba: @danlamanna Is there anything else we should do to verify this?

setup.py
```diff
@@ -66,7 +66,7 @@ def run(self, *arg, **kw):
     'CherryPy<11.1',
     'click',
     'filelock',
-    'funcsigs ; python_version < \'3.5\'',
+    'funcsigs ; python_version < \'3\'',
```
zachmullen: If I understand, this is kind of going out of our way to break compatibility with python between 3.0 and 3.4?

brianhelba: Indeed, this is making an additional change (which I've structured as a separate final commit, in case we don't like it) to eliminate special cases around Python <3.5. However, if we don't support Python 3.4, it will only be a matter of time before something breaks under it, and going forward, this change somewhat simplifies some of the existing code in Girder.

@manthey I recall speaking to you about this, but I don't recall what your opinion was?

brianhelba:
@danlamanna Yes, I've seen this issue with the hash serialization. I've got a fix (short and long term) in the works.

brianhelba:
@danlamanna I've fixed the tests under Python 3.6, though we'll require some additional work (which I plan to do at some point) to start testing Python 3.6 on CircleCI.

@girder/developers This PR is ready for additional review / discussion.

test/test_hash_state.py
```diff
@@ -20,29 +20,46 @@
...
-        checksum.update(chunk)
-        state = hash_state.serializeHex(checksum)
+@pytest.mark.parametrize('algorithm', [
+    hashlib.md5,
+    hashlib.sha1,
+    hashlib.sha224,
+    hashlib.sha256,
+    hashlib.sha384,
+    hashlib.sha512
+])
```
danlamanna: Mind adding the `ids` argument here for clearer test output?

docs/prerequisites.rst
```diff
@@ -3,7 +3,7 @@ System Prerequisites
 
 The following software packages are required to be installed on your system:
 
-* `Python 2.7 or 3.4+ <https://www.python.org>`_
+* `Python 2.7 or 3.5+ <https://www.python.org>`_
```
danlamanna: Should line 22 be changed to "work on **newer** versions of Python 3 as well"?

test/test_hash_state.py
brianhelba: Refactored to accomplish the same thing.

docs/prerequisites.rst
brianhelba: Done.

brianhelba:
@danlamanna @cjh1 Upgraded the Docker images to use a newer Debian with Python 3.5.

Dockerfile
```diff
@@ -1,13 +1,12 @@
-FROM node:6
+FROM node:8-stretch
 MAINTAINER Kitware, Inc. <kitware@kitware.com>
 
 EXPOSE 8080
 
 RUN mkdir /girder
 RUN mkdir /girder/logs
 
-RUN apt-get update && apt-get install -qy software-properties-common python-software-properties && \
```
brianhelba: It doesn't seem that these packages are needed now (they go all the way back to the first Dockerfile), but let me know if there's some crucial reason that they are required.

danlamanna: I think these are an artifact of originally using a PPA to install node.

danlamanna:
> Also - could we add this change to the removals section of the changelog?

@brianhelba ^

PR #2574, opened 2018-01-17
TEST: WIP: Post a Codecov detail comment on PRs with coverage changes

brianhelba:


manthey:
We had a PR from @danlamanna that specifically disabled comments from codecov about a year ago.

danlamanna:
> We had a PR from @danlamanna that specifically disabled comments from codecov about a year ago.

#1813 - I believe the justification for this was a consensus on IRC that the comments/emails are rather annoying.

brianhelba:
@manthey Yup, I'm not yet sure if this will ultimately be a net positive; I'm just using this PR to run some experiments.

My hope is that due to recent improvements to Girder's coverage handling, the information will be significantly more meaningful than it was with the old defaults. Also, the line `require_changes: true` should result in comments only being posted if changes occur at all (disabling them for things like documentation edits). Of course, that relies on coverage being absolutely stable from run-to-run, but fixing coverage determinism would be a worthwhile goal in itself.

mgrauer:
Vue-Mooted.

brianhelba:
Not mooted by Vue, but rather by the state of CircleCI <-> GitHub interfacing.

PR #2576, opened 2018-01-17
Use Girder's custom JsonEncoder to serialize JSON

brianhelba:
* Use Girder's custom JsonEncoder to serialize JSON
* Explicitly add support for JSON serialization of ObjectIds
* Add support for JSON serialization of general iterables
* Fail if an object can't be meaningfully JSON serialized

### DIFF
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -28,6 +28,7 @@ from .model_base import AccessControlledModel
 from girder import events
 from girder.constants import AccessType
 from girder.exceptions import ValidationException, GirderException
+from girder.utility import JsonEncoder
 from girder.utility.progress import noProgress, setResponseTimeLimit
 
 
@@ -708,7 +709,7 @@ class Folder(AccessControlledModel):
                 yield (filepath, file)
         if includeMetadata and metadataFile and doc.get('meta', {}):
             def stream():
-                yield json.dumps(doc['meta'], default=str)
+                yield json.dumps(doc['meta'], cls=JsonEncoder)
             yield (os.path.join(path, metadataFile), stream)
 
     def copyFolder(self, srcFolder, parent=None, name=None, description=None,
--- a/girder/models/item.py
+++ b/girder/models/item.py
@@ -29,7 +29,7 @@ from girder import events
 from girder import logger
 from girder.constants import AccessType
 from girder.exceptions import ValidationException, GirderException
-from girder.utility import acl_mixin
+from girder.utility import acl_mixin, JsonEncoder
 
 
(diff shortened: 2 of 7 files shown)


brianhelba:
I want to do more investigation of the final commit, and see if there are any cases where coercing an unknown object to a string actually produces a true serialization of the object state (and where we can't know and add native support for the object type in advance).

However, I do think that the past behavior of blindly coercing things to a string violates the design of the base `json.JsonEncoder`, and is the equivalent of just silencing real errors and offloading the failure to the end user without warning. This is particularly problematic in a RESTful design, when a consumer ideally ought to be able to fetch, de-serialize, serialize, and re-save the same data without losing information.

brianhelba:
@girder/developers This is ready for review, but I'd like to get more feedback (particularly from the `item_tasks` crowd) on whether there are any Python objects that may fail to be serialized going forward. If we can identify any, I think the solution is to add true serialization support, instead of essentially silencing the error.

jbeezley:
I don't see any major problems with this.  I just have a couple of questions.
1. `json.dumps` is used all over the place in the code base.  Why have you used the json serializer in these particular places and not others?
2. Are we concerned about deserialization as well?  At lease when serializing with dumps, you have to explicitly handle non-serializeable objects, which provides a hint when deserializing.

I think if we want a more generic solution, we should think about `jsonpickle`, but that would cause backwards compatibility issues in many places.  For example, `ObjectId('...') -> '{"py/state": {...}, "py/object": "bson.objectid.ObjectId"}'` rather than just the string representation.

danlamanna:
> 1. json.dumps is used all over the place in the code base. Why have you used the json serializer in these particular places and not others?

I wasn't able to spot any places outside of tests, which is mostly used for POSTing requests to the server.

> I think if we want a more generic solution, we should think about jsonpickle, but that would cause backwards compatibility issues in many places. For example, ObjectId('...') -> '{"py/state": {...}, "py/object": "bson.objectid.ObjectId"}' rather than just the string representation.

I think this PR is mostly concerned with fixing areas where HTTP responses bypass our custom `JsonEncoder` class, in which case `jsonpickle` wouldn't make much sense.

> However, I do think that the past behavior of blindly coercing things to a string violates the design of the base json.JsonEncoder, and is the equivalent of just silencing real errors and offloading the failure to the end user without warning.

+1 

@brianhelba Could you add relevant test(s) to `test_rest_utils.py`, in particular it might be useful to have one testing the condition for a non-json serializable response.

mgrauer:
@brianhelba thinks this should be relatively easy to get this merged in. It will introduce new errors intentionally.

PR #2577, opened 2018-01-18
Refactor Python documentation building

brianhelba:


tox.ini
```diff
@@ -11,6 +11,24 @@ commands = pytest \
...
+    sphinx_rtd_theme
+# readthedocs.org runs "python setup.py install" (which installs no dependencies); with the
+# limitations of Tox (not being able to use a different install_command for deps installation and
+# package installation) and the desire to install cleanly via pip (instead of directly with
+# setuptools), this seems to be the best configuration possible
+skip_install = true
+commands =
+    pip install --no-deps --editable . --editable clients/python
+    sphinx-build \
+        -W \
```
danlamanna: Adding `--color` here lets `tox` output warnings/errors in red which is nice.

tox.ini
```diff
@@ -11,6 +11,24 @@ commands = pytest \
...
+        -b html \
+        -d {envtmpdir}/doctrees \
+        docs \
+        build/docs/python
```
danlamanna: Does it make sense to pass `{posargs`} here?

tox.ini
brianhelba: Will do.

tox.ini
brianhelba: Good question. What other options would someone want to inject though? Running a different builder would be a common use case, but you'd have to retype the command manually, since a builder is already hardcoded. The only few that I see are:
`-E                don't use a saved environment, always read all files`
`-j N              build in parallel with N processes where possible`
`-n                nit-picky mode, warn about all missing references`
`-v                increase verbosity (can be repeated)`
Some of these actually may be interesting enough to turn on all the time, but I feel like the others are getting so low-level that if a developer wants to specify them, it's probably time to run the command manually.

danlamanna: Looking at that list, I would vote for `-n` to be added to the defaults. It seems difficult to justify it since I don't build the docs quite often, but any/all of those options and more (full tracebacks, pdb, only showing warnings, configuration settings, etc).

zachmullen:
> +137 −847

😹 🥂 🎉

tox.ini
danlamanna: Unrelated, but for anyone passing by - I recently discovered https://github.com/GaretJax/sphinx-autobuild which makes the process of editing docs with sphinx significantly less painful.

brianhelba:
@girder/developers Ready for review. PTAL.

This fixes the broken RTD build.

girder/models/__init__.py
```diff
@@ -27,10 +27,6 @@
 
 _dbClients = {}
 
-if pymongo.version_tuple < (3,):
-    raise Exception('Your pymongo version (%s) is too old. Please update to '
-                    'pymongo 3.x.' % pymongo.version)
```
zachmullen: This isn't related to this PR, let's make this a separate PR.

.circleci/config.yml
```diff
@@ -137,7 +137,7 @@ jobs:
           working_directory: girder_build
       - run:
           name: Run tox
-          command: tox -e circleci-py35
+          command: tox -e circleci-py35,docs
```
zachmullen: Would it make sense to do this as a separate step in the workflow? It would be pretty fast, right?

brianhelba:
@zachmullen Oops, I disabled CI builds of documentation while I was developing this (since I had to push repeatedly for RTD to build) and forgot to re-add it. This is fixed now.

.circleci/config.yml
brianhelba: Eventually, I could see nearly all of our test process getting launched through `tox`, and I don't know if we'd want to re-invoke it on a per-environment basis then. For now though, I don't have a strong feeling. @danlamanna do you know if there's a compelling reason to do things either way?

tox.ini
brianhelba: @danlamanna We'll have to address this in the future. See #2643 and #2646 .

girder/models/__init__.py
brianhelba: This code will fail when `pymongo` is mocked in a Sphinx-build environment. However, this logic is now obsolete (the correct `pymongo` version will now be ensured by `setup.py` when `girder` is pip-installed, unlike in the past when `girder` was not pip-installed and we just hoped users had used the `requirements.txt` file), so I thought it was safe to remove here. Do you think this still necessitates a separate PR?

BTW, thanks for pushing back on out-of-context changes being slipped into PRs. I think it's good to avoid, I'm just not sure if that's the case here.

jcfr:
Great to see the documentation improved :+1: 

Two questions:

* Can we still easily generate the documentation locally ? In project, I like to have two make targets `docs` and `docs-only`. The first generate doc and open local browser, the other only generate the doc. For example, see [here](https://github.com/scikit-build/scikit-build/blob/d5fa09fb28caddad73f7a7f141969cdda28266b4/Makefile#L60-L68)

* Can the documentation framework be used to generate doc of only one plugin independently of the complete girder ? Quite useful when a plugin doesn't belong to the core.

girder/models/__init__.py
zachmullen: I didn't realize this change was related to the sphinx build stuff. In light of that, I'd be OK with doing it on this PR, but could you explain how the mocking of packages works? Is that something we configure somewhere, or is it just magically handled for us? And if it's magic, why did this work before and not now? Or was this the cause of all the build failures?

brianhelba:
@jcfr You can definitely generate the documentation locally. As [it's now documented](https://github.com/girder/girder/pull/2577/files#diff-570609f10b140ce5703d5c50a5313725R4), just run `tox -e docs`; as a bonus, this now uses a virtualenv to build the docs in the same sort of mocked environment that RTD will use in production.

I would object to opening a GUI web browser as a side-effect of running a developer CLI command, on the same principle that I would object to a GUI text editor being opened as a side effect of running a developer CLI command to build the software in debug mode. The docs are now output to the easily-reachable location `build/docs/python` (and this fact is emitted at the end of the build command's output), which developers can open with the tools of their choice.

I would like to eventually see the documentation for built-in plugins be decentralized out of `docs/plugins.rst`, and moved into files within each built-in plugin's folder. Girder's documentation build process should then scrape built-in plugin folders and concatenate the text to produce the final page(s). Under this system, the output may look nearly identical, but the source code will be more modular. However, I'd object in principal to making a Girder include a system for building stand-alone documentation for downstream plugins; it's far simpler for downstreams to bootstrap Sphinx for their own custom needs.

If you'd like to discuss either of these ideas further, can you create separate issues for them, to make it easier to track separately from this PR?

girder/models/__init__.py
brianhelba: The mocking is done by the [`autodoc_mock_imports` option](http://www.sphinx-doc.org/en/stable/ext/autodoc.html#confval-autodoc_mock_imports). This option was not used by Girder in the past, but is a fairly significant part of `sphinx.ext.autodoc`, so I don't think it needs independent documentation here (as we assume developers of `docs/conf.py` understand Sphinx and its common extensions).

zachmullen: :shipit:

brianhelba:
Merging now.

There's [one discussion point](https://github.com/girder/girder/pull/2577#discussion_r172670514) where I'd welcome a follow-up from @danlamanna , but otherwise, I believe all review comments have been addressed.

jcfr:
> run tox -e docs; as a bonus, this now uses a virtualenv to build the docs in the same sort of mocked environment that RTD will use in production.

Great. Is the documentation build directory cached between run ? 

As a side note, worth noting that the a similar approach is used to [generate](https://github.com/tox-dev/tox/blob/9a9ea31922c75c275fc2e28ff842b3aad5664257/tox.ini#L20-L25) tox documentation.

> can you create separate issues for them, to make it easier to track separately from this PR?
yes. see reference above


> The docs are now output to the easily-reachable location build/docs/python (and this fact is emitted at the end of the build command's output), which developers can open with the tools of their choice.

Great. the full path would be displayed last so that we could suggest developer to add an alias like:

```
alias builddocs='tox -e docs |& tee /tmp/log && (docdir=$(tail -n1 /tmp/log) && xdg-open file://$docdir/index.html)'
```

brianhelba:
@jcfr 
> Great. Is the documentation build directory cached between run ?

Yes, the virtualenv is cached by Tox, and the Sphinx files are also provided with a place to be cached (the `-d` option to `sphinx-build`). One theoretical opportunity for speedup is in the fact that `girder` and `girder_client` are re-installed by Tox on every run, but there's a code comment in the `tox.ini` explaining why this is too hard to work around for now.

> As a side note, worth noting that the a similar approach is used to generate tox documentation.

Indeed! My goal is to make Girder as unsurprising as possible, and to attempt to follow conventions and practices established by the broader Python ecosystem and upstream tools.

> Great. the full path would be displayed last so that we could suggest developer to add an alias like:...

Cool. You're welcome to add this yourself, but honestly, the attempt to automatically parse the output seems overly generalized. Sphinx docs will always be output to `build/docs/python`; this is (by design) not configurable.

PR #2578, opened 2018-01-19
Add item and folder hrefs.

manthey:
This adds hrefs in the item and folder lists.  By preventing defaults on the links, these won't affect the existing hierarchy widget behavior.

Note that this neither depends on nor changes PR #2489 in any way.

We tried to do this in PR #1434, but in a much heavier way.  This doesn't use the href to do routing.

### DIFF
--- a/clients/web/src/templates/widgets/folderList.pug
+++ b/clients/web/src/templates/widgets/folderList.pug
@@ -3,7 +3,7 @@ ul.g-folder-list
     li.g-folder-list-entry(public=(folder.get('public') ? 'true' : 'false'))
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-folder-cid=folder.cid)
-      a.g-folder-list-link(g-folder-cid=folder.cid, href='#folder/' + folder.id)
+      a.g-folder-list-link(g-folder-cid=folder.cid, href=`#folder/${folder.id}`)
         i.icon-folder
         = folder.get('name')
         i.icon-right-dir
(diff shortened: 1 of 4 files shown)


clients/web/src/templates/widgets/folderList.pug
```diff
@@ -3,7 +3,7 @@ ul.g-folder-list
     li.g-folder-list-entry(public=(folder.get('public') ? 'true' : 'false'))
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-folder-cid=folder.cid)
-      a.g-folder-list-link(g-folder-cid=folder.cid)
+      a.g-folder-list-link(g-folder-cid=folder.cid, href='#folder/' + folder.id)
```
brianhelba: Can you use ES6 templating instead of string concatenation in these templates?

manthey:
There are cases where we use the hierarchy widget without navigating, such as when using it to select an item via a dialog.  In this case, the href will never reflect what selecting the item will do, but, I would argue, has more value precisely because of that -- the user can right click the link and examine the item in another tab, thus helping to determine if they want to select that item in the dialog.

We already allow plugins to override pretty much anything, so how do you see adding an explicit function to override the href as beneficial?

brianhelba:
I just discussed this with @manthey . While I agree that the callback approach is necessary for widgets to work as importable, reusable, composable elements, there are other factors that are currently blocking that from being a practical approach for downstreams.

So, if we merge this now, we can give end users the immediate benefits of hrefs. In the future, if / when we make Girder widgets practically reusable, we will have to switch to a callback approach, but that can be done as an additional non-breaking feature.

PR #2581, opened 2018-01-23
Convert some more core tests to pytest

zachmullen:


girder/__init__.py
```diff
@@ -226,8 +226,8 @@ def _setupLogger():
 logger = _setupLogger()
 
 
-def logStdoutStderr():
-    if _originalStdOut == sys.stdout:
+def logStdoutStderr(force=False):
```
danlamanna: Why is this flag required?

zachmullen: pytest hijacks sys.stdout and sys.stderr, so `_originalStdOut` in this module is no longer equal to `sys.stdout` at test time. This flag seemed like the simplest way to workaround that.

PR #2582, opened 2018-01-23
Workaround broken circleci/mongo:3.2-ram image

zachmullen:
See https://github.com/circleci/circleci-images/issues/143


brianhelba:
You could instead just remove the command entirely for now.

zachmullen:
Tried that first, same error.

brianhelba:
We could just just the non-RAM image for now.

I don't feel strongly right now though.

zachmullen:
I plan to revert whatever we do on this topic as soon as possible, so I think we should just merge this since it fixes the issue.

zachmullen:
@brianhelba did your fix get upstreamed to docker hub yet? If so I will make a revert PR.

brianhelba:
The fix is merged: https://github.com/circleci/circleci-images/pull/144

I'd suggest opening a revert PR, and seeing if it passes, as there's a slim possibility that rebuilds and cache flushing needs more time to occur.

PR #2585, opened 2018-01-25
Upgrade UglifyJS plugin to latest version

zachmullen:
The motivation for this is to take advantage of a new feature in the UglifyJsPlugin that is not available in the current major version of webpack that Girder uses. Namely, the ability to handle ES6+ syntax in the output.

This allows me to build my paraviewweb + vtk.js plugin in prod mode, which was otherwise impossible since their upstream build process produces outputs with some ES6 features in the output, such as hex numeric notation.

### DIFF
--- a/grunt_tasks/build.js
+++ b/grunt_tasks/build.js
@@ -22,6 +22,7 @@ const extendify = require('extendify');
 const webpack = require('webpack');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const GoogleFontsPlugin = require('google-fonts-webpack-plugin');
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 
 const isTrue = (str) => !!str && !['false', 'off', '0'].includes(str.toString().toLowerCase());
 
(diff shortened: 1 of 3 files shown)


brianhelba:
Why can't / shouldn't the dependencies be piped through Babel?

More generally, can the upstream start releasing packages which use [the `module:` field in `package.json`](https://github.com/rollup/rollup/wiki/pkg.module) (which [Webpack supports too](https://webpack.js.org/configuration/resolve/#resolve-mainfields)), so that Girder can be the one to import, transpile, and bundle their code?

zachmullen:
@brianhelba I will investigate the `module` field approach, but we can consider that independently from this proposal. That is, do we consider this change OK, regardless of the motivation? If someone were to intentionally build a Girder plugin that produces ES6 output, they'd currently be unable to uglify it with Girder's build infrastructure. This change would support such a use case.

brianhelba:
I agree that latter proposal is a totally separate topic. We can ignore that for now.

How about the first question? Why don't you transpile the dependencies through Babel (like the rest of Girder's own code)?

I'm trying to understand the use case here, I don't fundamentally disagree with this change.

zachmullen:
I did transpile the dependency with `babel-loader` with the `env` preset, for some reason that didn't remove all the ES6 syntax. Maybe I'm misunderstanding or misusing the babel plugin in the upstream build?

PR #2591, opened 2018-01-29
Pin parts of vega.

manthey:
The most recent version of vega depends on vega-canvas, but that doesn't build with webpack.  See https://github.com/vega/vega-canvas/issues/1. This pins enough of vega to avoid the problem.

This can be reverted when candela publishes a work around or the problem is fixed in vega-canvas.

### DIFF
--- a/plugins/candela/package.json
+++ b/plugins/candela/package.json
@@ -2,6 +2,10 @@
     "name": "candela-plugin",
     "version": "0.1.0",
     "dependencies": {
+        "vega": "3.0.9",
+        "vega-scenegraph": "2.1",
+        "vega-wordcloud": "2.0.2",
+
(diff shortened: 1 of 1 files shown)


zachmullen:
Thanks @manthey

brianhelba:
Why can't the pinning be done upstream in Candela, and do a quick release of that?

Given, the the way that npm dependency resolution works, sub-dependency pinning is fragile at best and only works in limited cases (luckily, perhaps we *could* trick npm in this particular case).

manthey:
@brianhelba It has been done in candela, but candela's tests are failing, and therefore pushing a release is harder than it could be.  This is a temporary workaround until it is fixed somewhere upstream.

PR #2592, opened 2018-01-30
Use ES6 arrow functions for lambdas, instead of _.bind

brianhelba:


brianhelba:
@girder/developers PTAL, before this gets stale.

PR #2593, opened 2018-01-30
Improve SearchFieldWidget support for long-running searches

brianhelba:
* Show an animated icon for long-running searches
* Fix a race condition when `SearchFieldWidget` goes to results while pending


Pierre-Assemat:
Thanks @zachmullen! 

This looks great @brianhelba, LGMT!!

PR #2594, opened 2018-01-30
Improvements to SearchResultsView 

brianhelba:
* Return a stable ordering for unknown search result types
  * This also fixes bugs with custom search modes returning fewer types.
* Use hrefs in the search results page
* Remove unused variables in `SearchResultsView`
* Simplify parameters for `SearchResultsTypeView`
* Refactor `SearchResultsView` to add a pending spinner icon
* Adjust the title position of the search results page

### DIFF
--- a/clients/web/src/templates/body/searchResults.pug
+++ b/clients/web/src/templates/body/searchResults.pug
@@ -1,16 +1,15 @@
 .g-search-results-header-container
   .g-search-results-title-container
     i.g-search-results-title-element.icon-search
-    if length === 1
-      span.g-search-results-title.g-search-results-title-element Result : #{query}
-    else
-      span.g-search-results-title.g-search-results-title-element Results : #{query}
+    span.g-search-results-title.g-search-results-title-element Results : #{query}
   .g-search-results-header
     .g-search-results-advance-search
 
 .g-search-results-container
 
-if !length
-  .g-search-no-results.disabled
-    i.icon-block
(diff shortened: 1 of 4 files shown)


clients/web/src/templates/body/searchResults.pug
```diff
@@ -1,16 +1,15 @@
 .g-search-results-header-container
   .g-search-results-title-container
     i.g-search-results-title-element.icon-search
-    if length === 1
-      span.g-search-results-title.g-search-results-title-element Result : #{query}
-    else
-      span.g-search-results-title.g-search-results-title-element Results : #{query}
+    span.g-search-results-title.g-search-results-title-element Results : #{query}
```
Pierre-Assemat: So we don't need to make the difference for only one result found ?

Pierre-Assemat:
I checkout this on my machine, this looks really great! I like the pending search stuff!

clients/web/src/templates/body/searchResults.pug
brianhelba: I don't think it matters in this context, since it's a header for any number of items below, as opposed to part of a direct label (like the user count).

brianhelba:
@Pierre-Assemat , can you review the commits in this branch one-by-one? If you think all the code changes make sense, I'd be comfortable merging from your approval.

brianhelba:
Cool, thanks!

PR #2596, opened 2018-01-30
Unify girder-* commands into top level girder command

danlamanna:
The primary objective of this PR is to add a top level `girder` console script which can be used to centralize the various scripts housed in Girder (`girder-install`, `girder-server`, `girder-shell`, `girder-sftpd`, etc). Note that all existing console scripts continue to function as they did before.

This will provide:
1) A unified CLI interface
    As flags need to get added or modified, we can add them to the top level `girder` command rather than reimplementing them in each subcommand's parser.
2) A centralized index for users
    With this PR, typing `girder` or `girder --help` shows an index of subcommands and descriptions that can be executed:
```
Usage: girder [OPTIONS] COMMAND [ARGS]...

Options:
  --help  Show this message and exit.

Commands:
  serve  Run the Girder server.
  sftpd  Run the Girder SFTP service.
  shell  Run a Girder shell.
```

`girder-install` has been left out due to it's likely significant changes in Girder 3

Future work:
- It would be nice to move the console scripts to a `bin` or `cli` module, rather than having some in `utility`, some in `api`, and some in the main `girder` package.
- There has been discussion about new commands for config file modification and schema migrations.
- Other pip installable packages could (if they want) use an entrypoint to create their own subcommands to the `girder` command.

### DIFF
--- a/docs/sftp.rst
+++ b/docs/sftp.rst
@@ -10,16 +10,16 @@ with their Girder login name and password. This protocol can make it much easier
 nested datasets with many individual files, and is tolerant to network failure since it supports
 resuming interrupted downloads of entire data hierarchies.
 
-After installing the Girder package via pip, you should now see the ``girder-sftpd`` executable
-in your PATH. Running it will start the SFTP server using the same database configuration as the
-main Girder HTTP server.
+After installing the Girder package via pip, you should now see the ``girder`` executable
+in your PATH. Running ``girder sftpd`` will start the SFTP server using the same database configuration
+as the main Girder HTTP server.
 
 The SFTP server requires a private key file for secure communication with clients. If you do
 not pass an explicit path to an RSA private key file, the service will look for one at
 ``~/.ssh/id_rsa``. It's recommended to make a special key just for the SFTP service, e.g.::
 
     ssh-keygen -t rsa -b 2048 -f my_key.rsa -N ''
-    girder-sftpd -i my_key.rsa
+    girder sftpd -i my_key.rsa
 
 You can control the port on which the server binds by passing a ``-p <port>`` argument to the
 server CLI. The default port is 8022.
--- a/girder/api/sftp.py
+++ b/girder/api/sftp.py
@@ -273,7 +273,7 @@ class SftpServer(socketserver.ThreadingTCPServer):
               help='The port to bind to')
 def main(identity_file, port, host):  # pragma: no cover
     """
-    This is the entrypoint of the girder-sftpd program. It should not be
+    This is the entrypoint of the girder sftpd program. It should not be
     called from python code.
     """
     try:
--- a/girder/api/sftp.py
+++ b/girder/api/sftp.py
@@ -19,6 +19,7 @@
 
 from __future__ import print_function
 
+import click
 import os
 import paramiko
 import six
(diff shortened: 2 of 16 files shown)


jeffbaumes:
Not sure how this cool feature was added in 10 net lines of code. Click wizardry ...

zachmullen:
Any good way to unit test this?

brianhelba:
I don't think this is *unit* testable, but switching some of the smoke / packaging tests or CI build steps to use this would be good.

danlamanna:
PTAL @brianhelba, this PR also includes a commit that should have been added in #2332.

brianhelba:
@danlamanna Could you fix merge conflicts? I'll be happy to review afterwards.

danlamanna:
> @danlamanna Could you fix merge conflicts? I'll be happy to review
> afterwards.

Done.

docs/sftp.rst
```diff
@@ -10,7 +10,7 @@ with their Girder login name and password. This protocol can make it much easier
 nested datasets with many individual files, and is tolerant to network failure since it supports
 resuming interrupted downloads of entire data hierarchies.
 
-After installing the Girder package via pip, you should now see the ``girder-sftpd`` executable
+After installing the Girder package via pip, you should now see the ``girder sftpd`` executable
```
brianhelba: This should be just `girder`.

girder/api/sftp.py
```diff
@@ -261,32 +262,25 @@ def shutdown_request(self, request):
         pass
 
 
-def _main():  # pragma: no cover
+@click.command(name='sftpd', short_help='Run the Girder SFTP service.')
+@click.option('-i', '--identity-file', help='path to identity (private key) file')
+@click.option('-p', '--port', default=DEFAULT_PORT, type=int)
```
brianhelba: Might as well take this chance to add `help` to the `--port` and `--host` options. It should be fine to copy the strings from `girder serve`.

girder/utility/shell.py
```diff
@@ -42,7 +42,8 @@ def _launchShell(context):
         return code.interact(banner=header, local=context)
 
 
-@click.command(context_settings=dict(help_option_names=['-h', '--help']))
+@click.command('shell', short_help='Run a Girder shell.',
+               context_settings=dict(help_option_names=['-h', '--help']))
```
brianhelba: Why is `help_option_names` here? If there's a good reason, can you add a code comment? I realize you didn't add this here, but it's confusing as-is.

girder/api/sftp.py
```diff
@@ -261,32 +262,25 @@ def shutdown_request(self, request):
...
+@click.option('-H', '--host', default='localhost')
+def _main(identity_file, port, host):  # pragma: no cover
```
brianhelba: Should this be `main`? Should everything else be `_main`? I don't really know, but it ought to be consistent.

girder/cli.py
```diff
@@ -0,0 +1,14 @@
+from click_plugins import with_plugins
+from pkg_resources import iter_entry_points
+import click
+
+
+@with_plugins(iter_entry_points('girder.cli_plugins'))
+@click.group()
```
brianhelba: Want to give this a `short_help`? If you can't think of anything better, you could always recycle `'Girder: data management platform for the web.'`.

brianhelba:
Nevermind about `girder-cli`, I see the comment.

girder/api/sftp.py
danlamanna: It doesn't matter to me. I think there's consensus that CLI functions in Python aren't/shouldn't be part of our API surface. Perhaps when all of the entrypoints are moved into a central module (#2612) we can just mark that module as outside of the API surface.

danlamanna:
@brianhelba PTAL

girder/api/sftp.py
```diff
@@ -261,32 +262,29 @@ def shutdown_request(self, request):
         pass
 
 
-def _main():  # pragma: no cover
+@click.command(name='sftpd', short_help='Run the Girder SFTP service.',
```
brianhelba: Why are both `short_help` and `help` required? Would putting the help into a docstring help?

girder/cli.py
```diff
@@ -0,0 +1,15 @@
+from click_plugins import with_plugins
+from pkg_resources import iter_entry_points
+import click
+
+
+@with_plugins(iter_entry_points('girder.cli_plugins'))
+@click.group(help='Girder: data management platform for the web.',
+             context_settings=dict(help_option_names=['-h', '--help']))
```
brianhelba: Could you add a comment, just explaining that the point of this is to add an `-h` option, which is not there by default? It's not obvious from the source.

girder/api/sftp.py
```diff
@@ -261,32 +262,29 @@ def shutdown_request(self, request):
...
+               help='Run the Girder SFTP service.')
+@click.option('-i', '--identity-file', show_default=True,
+              default=os.path.expanduser(os.path.join('~', '.ssh', 'id_rsa')),
+              help='The identity (private key) file to use')
+@click.option('-H', '--host', show_default=True, default='localhost',
+              help='The interface to bind to')
+@click.option('-p', '--port', show_default=True, default=DEFAULT_PORT, type=int,
+              help='The port to bind to')
+def _main(identity_file, port, host):  # pragma: no cover
```
brianhelba: Could you make this `main` then, for consistency? I agree we can still consider it excluded from the API surface.

girder/cli.py
danlamanna: `help_option_names=['-h', '--help'])` as an input to a CLI command is as obvious as it gets (without adding a comment).

girder/api/sftp.py
danlamanna: Done.

PR #2598, opened 2018-01-30
Prevent value injection into DICOM Javascript query

brianhelba:


brianhelba:
Previously, searching for a query with a backslash (`\`) at the end would cause the MongoDB query Javascript to escape improperly. At best, this would cause query syntax errors, but it could also be used to rewrite the query more maliciously.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -206,19 +207,24 @@ def dicomSubstringSearchHandler(query, types, user=None, level=None, limit=0, of
     """
     if types != ['item']:
         raise RestException('The dicom search is only able to search in Item.')
+    if not isinstance(query, six.string_types):
```
brianhelba: This just ensures that a non-string `query` (like a list or dict) doesn't slip in. [`six.string_types` is just a way to refer to strings](https://pythonhosted.org/six/#six.string_types) in both Python 2 and Python 3.

plugins/dicom_viewer/server/__init__.py
```diff
@@ -206,19 +207,24 @@ def dicomSubstringSearchHandler(query, types, user=None, level=None, limit=0, of
...
             return Object.keys(dicomMeta).some(
                 function(key) {
                     return (key.toLowerCase().indexOf(queryKey) !== -1)  ||
                         dicomMeta[key].toString().toLowerCase().indexOf(queryValue) !== -1;
                 })
             }
-        """ % {'queryKey': query, 'queryValue': query}
+    """ % {
+        # This could eventually be a separately-defined key and value
+        'query': json.dumps(query)
```
brianhelba: The `dumps` serialization will escape any backslashes and output a JSON-safe string (which is the same as a Javascript string literal).

PR #2599, opened 2018-01-30
Add support for alternative assetstore models

zachmullen:
This change allows models besides the core assetstore model to be used to initialize assetstore adapters. This is desirable because it decouples assetstore *management* (an explicitly admin-only capability) from the actual service provided by the assetstore adapter. This will allow us to have user-managed "Database" models with their own custom UI and set of REST endpoints in the downstream database_assetstore plugin.

@manthey @jeffbaumes  PTAL

### DIFF
--- a/girder/models/file.py
+++ b/girder/models/file.py
@@ -161,19 +161,30 @@ class File(acl_mixin.AccessControlMixin, Model):
                 raise ValidationException(
                     'Linked file URL must start with http: or https:.',
                     'linkUrl')
-        if doc.get('assetstoreModel'):
+        if doc.get('assetstoreType'):
             # If assetstore model is overridden, make sure it's a valid model
-            try:
-                self.model(doc['assetstoreModel'], plugin=doc.get('assetstoreModelPlugin'))
-            except Exception:
-                raise ValidationException('Invalid assetstore model: %s (plugin=%s).' % (
-                    doc['assetstoreModel'], doc.get('assetstoreModelPlugin')))
+            self._getAssetstoreModel(doc)
         if 'name' not in doc or not doc['name']:
             raise ValidationException('File name must not be empty.', 'name')
 
         doc['exts'] = [ext.lower() for ext in doc['name'].split('.')[1:]]
 
         return doc
+
+    def _getAssetstoreModel(self, file):
+        from .assetstore import Assetstore
+
+        if file.get('assetstoreType'):
+            try:
+                if isinstance(file['assetstoreType'], six.string_types):
+                    return self.model(file['assetstoreType'])
+                else:
+                    return self.model(*file['assetstoreType'])
+            except Exception:
+                raise ValidationException(
+                    'Invalid assetstore type: %s.' % (file['assetstoreType'],))
+        else:
+            return Assetstore()
 
     def createLinkFile(self, name, parent, parentType, url, creator, size=None,
                        mimeType=None, reuseExisting=False):
@@ -279,8 +290,7 @@ class File(acl_mixin.AccessControlMixin, Model):
         }, field='size', amount=sizeIncrement, multi=False)
 
     def createFile(self, creator, item, name, size, assetstore, mimeType=None,
-                   saveFile=True, reuseExisting=False, assetstoreModel=None,
-                   assetstoreModelPlugin=None):
+                   saveFile=True, reuseExisting=False, assetstoreType=None):
         """
         Create a new file record in the database.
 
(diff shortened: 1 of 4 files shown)


test/test_assetstore_model_override.py
```diff
@@ -0,0 +1,49 @@
...
+
+    yield
+
+    assetstore_utilities.removeAssetstoreAdapter('fake')
+
+
+def testAssetstoreModelOverride(fakeModel, fakeAdapter, admin):
+    file = File().createFile(
+        creator=admin, item=None, name='a.out', size=0, assetstore=fakeModel,
+        assetstoreModel='fake', assetstoreModelPlugin='fake_plugin')
```
cjh1: Can you help me understand why we have to pass in the model 'name' ( 'fake') as well as the model instance here? Couldn't  the name be derived from the information provided in the ```registerModel``` call?

zachmullen: The variable name `fakeModel` is misleading, the fixture is for registering an actual model class, but what the fixture actually returns is a document of that model, which is just a dict. Due to this stringly-typed document representation in Girder, we have to pass around model types out-of-band.

zachmullen: I could always change the fixture so that it doesn't create a Fake document, and do that in the test function, if that would be cleaner.

cjh1: Ok, thanks for the explanation. fakeModel => fakeAssetstore, might make things a little clearer.

girder/models/file.py
```diff
@@ -309,6 +323,10 @@ def createFile(self, creator, item, name, size, assetstore, mimeType=None,
             'itemId': item['_id'] if item else None
         }
 
+        if assetstoreModel:
+            file['assetstoreModel'] = assetstoreModel
+            file['assetstoreModelPlugin'] = assetstoreModelPlugin
```
kotfic: Does it make sense for this metadata to be assigned to files even if they don't have a non-standard assetstoreModel/assetstoreModelPlugin ?

kotfic: Also,  if we just passed in the class (or the model object) rather than a string,  couldn't we enforce an API on assetstore models that returns this data.  That way createFile only has to query the assetstore model object (or class or whatever)  for this information?  

maybe this is going to far,  but if the assetstore argument was an actual object (instead of a dict)  then all of this could be pulled off of that object right?

zachmullen: > Does it make sense for this metadata to be assigned to files even if they don't have a non-standard assetstoreModel/assetstoreModelPlugin ?

We could do that, though preserving the old mode also ensures that legacy files will work without any need for data migration.

> if the assetstore argument was an actual object (instead of a dict) then all of this could be pulled off of that object right?

Not at this time, no. The registration of models to a string is currently a one-way pointer. The model itself doesn't know under what string it's registered, though that might be a nice future enhancement.

kotfic: i guess there have been some conversations about schema migration.  Don't let my concerns hold things up.  I was just curious as I was looking through.  I assume you want @manthey to make sure this is going to work for the database assetstore downstream right?

girder/models/file.py
```diff
@@ -161,6 +161,13 @@ def validate(self, doc):
                 raise ValidationException(
                     'Linked file URL must start with http: or https:.',
                     'linkUrl')
+        if doc.get('assetstoreModel'):
+            # If assetstore model is overridden, make sure it's a valid model
+            try:
+                self.model(doc['assetstoreModel'], plugin=doc.get('assetstoreModelPlugin'))
```
manthey: Elsewhere we have `attachedToType` where the type is either a string with the name of the model OR a list that has two entries that can be passed to `self.model(*file['attachedToType'])`.  In the utilities search module, we also handle model types specified by a dot-separated string.  We should pick a method to be the standard (and possible modify the `ModelImporter.model` method to handle all of these directly so we can always just call `ModelImported.model(<string or dot string or list or tuple with one or two items>)`).

aashish24:
thanks this will help with database_assetstore plugin!

girder/models/file.py
zachmullen: Yep, that's correct.

girder/models/file.py
zachmullen: Good point... what do you think we should do specifically for this PR?

manthey: I think, to match elsewhere in the code, this should be a single parameter called `assetstoreType` (rather than `assetstoreModel` and `assetstoreModelPlugin`).  Which means that you'd have to handle the string or tuple cases here (unless you want to make `model()` handle them generically, but I figured that would be better as a separate PR).

zachmullen: So if it's a plain string, it would be a core model, but if it's a tuple we assume it's of the form `(modelName, pluginName)`?

manthey: That's what we've done with `attachedToType`.

zachmullen:
This is ready for final review

Issue #2589, opened 2018-01-26
Add tests asserting the behavior of "hidden folders"


jbeezley:
Related to the behavior described in #2588.  As a supported use case, we should add explicit assertions that hidden folder access control works correctly.  Of particular concern are the `/resource/path/:id` and `/resource/lookup` endpoints.

PR #2600, opened 2018-01-31
Add tests asserting the behavior of 'hidden folders'

Pierre-Assemat:
The scope of this PR is to add consistent tests on the behavior of `/resource/path/:id` and `/resource/lookup` endpoints.

- [x] Unit test on [`parentsToRoot` function](https://github.com/girder/girder/blob/13903bcc01fb314ee40b28f287616a11de1fb933/girder/models/folder.py#L558)
- [x] Unit test on [`getResourcePath` function](https://github.com/girder/girder/blob/13903bcc01fb314ee40b28f287616a11de1fb933/girder/utility/path.py#L223)
- [x] Unit test on [`lookUpPath` function](https://github.com/girder/girder/blob/13903bcc01fb314ee40b28f287616a11de1fb933/girder/utility/path.py#L127)

Fixes: #2589

### DIFF
--- a/test/test_folder.py
+++ b/test/test_folder.py
@@ -46,16 +46,11 @@ def parentChain(admin):
         'privateFolder': privateFolder,
         'folder4': F4
     }
-    # Delete de parent chain
+    # Delete the parent chain
     Folder().remove(F1)
 
 
 def testParentsToRootAdmin(parentChain, admin):
-    """
-    Demonstrate that parentsToRoot works even if the user has missing right access
-    on one or more folder in the full path.
-    This tests for a user with right access, a user without and a none user.
-    """
     # Get the parent chain for a user who has access rights
     parents = Folder().parentsToRoot(parentChain['folder4'], user=admin)
     assert parents[1]['object']['name'] == 'F1'
--- a/test/test_folder.py
+++ b/test/test_folder.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#  Copyright 2014 Kitware Inc.
+#
+#  Licensed under the Apache License, Version 2.0 ( the "License" );
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+###############################################################################
+
+import pytest
+from bson.objectid import ObjectId
+
+from girder.exceptions import AccessException
+from girder.models.folder import Folder
+from pytest_girder.assertions import assertStatus, assertStatusOk
+
+def _createParentChain(admin):
+    # Create the parent chain
+    F1 = Folder().createFolder(
+        parent=admin, parentType='user', creator=admin,
+        name='F1', public=True)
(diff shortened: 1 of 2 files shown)


Pierre-Assemat:
What do we do with that? Do we still need it ? 
If yes I think this is ready for review @brianhelba

jbeezley:
Thanks!  The assertions here are exactly what we are looking for, but I think it would be better to add these as new "pytest-style" tests instead of the old testing infrastructure.  See for example `test/test_access.py`.

danlamanna:
Docs are [here](http://girder.readthedocs.io/en/latest/developer-cookbook.html#adding-automated-tests-with-pytest) and [here](http://girder.readthedocs.io/en/latest/development.html#running-the-tests-with-pytest) for that @Pierre-Assemat.

Pierre-Assemat:
Yes of course, it make total sense! 
I'm on it!

Pierre-Assemat:
@jbeezley I think this is ready for review.

test/test_folder.py
```diff
@@ -0,0 +1,137 @@
...
+                          method='GET', user=None,
+                          params={
+                              'path': '/user/%s/%s/%s/%s/%s' % (
+                                  admin['login'],
+                                  F1['name'],
+                                  F2['name'],
+                                  privateFolder['name'],
+                                  F4['name'])
+                          })
+    assertStatus(resp, 400)
```
jbeezley: This should be three separate tests.  You might consider [parametrizing](https://docs.pytest.org/en/latest/example/parametrize.html) them to deduplicate the code.

test/test_folder.py
```diff
@@ -0,0 +1,137 @@
...
+    assert resp.json == '/user/%s/%s/%s/%s/%s' % (
+        admin['login'], F1['name'], F2['name'], privateFolder['name'], F4['name'])
+    resp = server.request(path='/resource/%s/path' % F4['_id'],
+                          method='GET', user=user,
+                          params={'type': 'folder'})
+    assertStatus(resp, 403)
+    resp = server.request(path='/resource/%s/path' % F4['_id'],
+                          method='GET', user=None,
+                          params={'type': 'folder'})
+    assertStatus(resp, 401)
```
jbeezley: Three separate tests.

test/test_folder.py
Pierre-Assemat: I didn't succeed to use `admin` or `user` as a parameters with `pytest.mark.parametrize`. Do you think it's an issue to have some duplicate code in testing?

test/test_folder.py
```diff
@@ -38,100 +40,119 @@ def _createParentChain(admin):
...
+    # Delete de parent chain
+    Folder().remove(F1)
 
 
-def testParentsToRoot(admin, user):
+def testParentsToRootAdmin(parentChain, admin):
     """
     Demonstrate that parentsToRoot works even if the user has missing right access
     on one or more folder in the full path.
     This tests for a user with right access, a user without and a none user.
```
jbeezley: Could you update or remove this docstring?

test/test_folder.py
jbeezley: That's okay if it's too awkward in this case.  It's just a good trick to keep in mind for the future.

test/test_folder.py
```diff
@@ -38,100 +40,119 @@ def _createParentChain(admin):
...
         parent=privateFolder, parentType='folder', creator=admin,
         name='F4', public=True)
-    return F1, F2, privateFolder, F4
+    yield {
+        'folder1': F1,
+        'folder2': F2,
+        'privateFolder': privateFolder,
+        'folder4': F4
+    }
+    # Delete de parent chain
```
jbeezley: `s/de/the`

jbeezley:
Looks good, @Pierre-Assemat!

PR #2602, opened 2018-02-01
Allow cancellation of raw Celery

cjh1:
Fixes girder/girder_worker#254. It appear that then the UI component was written there was no support for cancelling RUNNING celery jobs.


mgrauer:
As Method Man says, "When you give it to me, ayy, give it to me raw". 

That is from the song "Bring The Pain", which maybe doesn't bode so well for this PR...

cjh1:
@mgrauer Thanks for musical interlude :smile:

PR #2605, opened 2018-02-06
Virtual folders

zachmullen:
This is the minimum viable feature set for core support for virtual folders. This is a backend-only change; as written, it would be up to plugins to use this feature by creating their own virtual folders.

@manthey PTAL.

### DIFF
--- a/clients/web/src/stylesheets/widgets/hierarchyWidget.styl
+++ b/clients/web/src/stylesheets/widgets/hierarchyWidget.styl
@@ -73,6 +73,9 @@
     padding 0
     list-style-type none
 
+    .g-virtual-folder-icon
+      color #2ba560
+
     >li
       border-top 1px solid #efefef
       box-sizing border-box
--- a/clients/web/src/templates/widgets/folderList.pug
+++ b/clients/web/src/templates/widgets/folderList.pug
@@ -4,7 +4,10 @@ ul.g-folder-list
       if checkboxes
         input.g-list-checkbox(type="checkbox", g-folder-cid=folder.cid)
       a.g-folder-list-link(g-folder-cid=folder.cid, href=`#folder/${folder.id}`)
-        i.icon-folder
+        if folder.get('isVirtual')
+          i.icon-folder-empty.g-virtual-folder-icon
+        else
+          i.icon-folder
         = folder.get('name')
         i.icon-right-dir
       .g-folder-privacy
--- a/girder/api/v1/folder.py
+++ b/girder/api/v1/folder.py
@@ -96,7 +96,7 @@ class Folder(Resource):
 
             return list(self._model.childFolders(
                 parentType=parentType, parent=parent, user=user,
-                offset=offset, limit=limit, sort=sort, filters=filters))
+                offset=offset, limit=limit, sort=sort, filters=filters, virtualQuery=True))
         elif text:
             return list(self._model.textSearch(
                 text, user=user, limit=limit, offset=offset, sort=sort))
--- a/girder/api/v1/item.py
+++ b/girder/api/v1/item.py
@@ -81,8 +81,7 @@ class Item(Resource):
         user = self.getCurrentUser()
 
         if folderId:
-            folder = Folder().load(
-                id=folderId, user=user, level=AccessType.READ, exc=True)
+            folder = Folder().load(id=folderId, user=user, level=AccessType.READ, exc=True)
             filters = {}
             if text:
                 filters['$text'] = {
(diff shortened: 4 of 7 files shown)


mgrauer:
@zachmullen @manthey 

Can one of you (or both if you disagree ⚖️) please describe what a virtual folder is and what the use case is?

zachmullen:
From d1c9d4b

> These folders cannot contain subfolders or items in the normal
way, but can be used to display an "alternate" view of the
hierarchy via a custom database query to control what folders
and items are listed as their children. Those folders and items
must live somewhere else within the hierarchy.

We should add something about this to the docs, I'm just not sure where yet. Hopefully that will fall out in QA.

Also a changelog entry is due.

danlamanna:
It seems like another way to describe this would be equivalent to a "view" in the database world. It appears like an actual table (folder) but is really just running a query dynamically to present itself.

jbeezley:
What happens when a virtual folder contains a query that matches itself (or even a parent folder)?  Do all of our endpoints that recurse through the folder hierarchy handle this?

zachmullen:
I probably should have put a WIP tag on this, because there's a lot still to address, and a lot more unit testing to add to verify that this doesn't do anything bad. I'm going to omit the tag, though, to invite as much review as possible. I'll leave a comment here when I think this is ready for a final pass.

manthey:
Another way to think of a virtual folder is as a "smart playlist".  It contains folders and items subject to rules rather than an explicit list.  This is not functionally different than thinking of it as a database view.

zachmullen:
I like "view", but I'd want to make sure it's not too esoteric for those who have no background in databases. "Smart playlist" is too evocative of a music player for my tastes. I've heard others refer to this like a "cart" of data, but that also suffers from the problem of being too evocative of a different domain.

cjh1:
There seems to be [good prior art](https://en.wikipedia.org/wiki/Virtual_folder)  for the term.

kotfic:
It seems like this is adding a lot of conditionals/flags to handle virtual folder functionality. I'm wondering what the trade offs are for introducing a new model type/endpoints?

manthey:
Conceptually, I don't think the default should be to NOT list the contents of a virtual folder.  I think that unless a client is aware, they should appear like normal folders.  This means that if you walk a virtual folder and its children it would be best if it couldn't enter a loop (but file systems allow this with bind mounts and such, so perhaps it should be a client-beware).

zachmullen:
@manthey I think I agree, however I'm concerned about breaking assumptions downstreams might make when using `childItems` or `childFolders`; if someone is using those to modify a subtree, for instance, things could go badly for them if we use the virtual behavior by default.

manthey:
Is there a way to make this work as a "dumb playlist" as well?  If the query is `{_id: {$in: [<list of ids, possibly some duplicates>]}}`, we get a result that contains specific items or folders, but I'm not sure how to use the mongo sort to force that to be in the list's order (including the possible duplicates).

The specific use for HistomicsTK, for instance, would be a researcher making a list of items of interest to examine later.  There, duplicates probably aren't necessary, but it would be nice to handle them, too (so I could ask for items A, B, C, A, D in that order).

zachmullen:
On the subject of cycles: right now I'm inclined to leave it up to users of this feature not to shoot themselves in the foot. The main thing I want to make sure is mechanically impossible is an infinite recursion on the server side, and right now, I believe that's not possible. None of the recursive methods in the folder model itself are using the virtual query. There are some recursive methods in the folder model (e.g. `updateSize`) that do actually call `childItems` / `childFolders`, which is why I wanted to disable `virtualQuery` by default, so that users have to opt-in to using the virtual functionality at all.

zachmullen:
> Is there a way to make this work as a "dumb playlist" as well?

I think a `$in` in the query should be fine to use with a `sort` parameter. Duplicate values in `$in` will not affect the result; that represents a set of values to match against, and a document will only be returned once in a query.

manthey:
I'm sure the query part will work to get the limited list, and I know we can sort to a particular field, but I don't know how to sort to a list.  The ability to list the same item multiple times could be useful.

zachmullen:
I'm not sure how to get the same item to show up multiple times, however if you need a truly custom sort order, it's best to add an `ordinal` field to the document as a piece of metadata. We do this in COVALIC and use a drag and drop widget in the browser to let people define the order of their challenge phases.

manthey:
I think having to alter the existing data negates a lot of the value of the virtual-ness of the folders.  Each ordered virtual folder would need its own ordinal field in all of its items.

If we supported aggregations instead of simple queries, we don't need the ordinal field.  For instance, if the `itemsQuery` was a dict, we pass it to `find`, but it if is a list, we pass it to `aggregate`.    See here for an aggregation that returns the results in the same order as the `$in` list: http://www.kamsky.org/stupid-tricks-with-mongodb/using-34-aggregation-to-return-documents-in-same-order-as-in-expression

This doesn't show to repeat an item (but it might be possible with some other aggregation).

zachmullen:
@manthey I like that idea, but maybe we can do that on a subsequent topic to decouple the discussion?

manthey:
Absolutely.

zachmullen:
I'll go ahead and call this ready for review now.

zachmullen:
@girder/developers I added this as a topic onto our agenda for Thursday's meeting

PR #2607, opened 2018-02-06
Allow plugins to customize parts of REST API Documentation page

msmolens:
Allow plugins to customize parts of the REST API Documentation (aka Swagger) page using Mako template inheritance [1]. The Mako template defines two named blocks—"docsHeader" and "docsBody"—that may be overridden in an inheriting template.

For example, a plugin can add its own template directory and specify a template filename in its `load` function:

    templateDir = os.path.join(getPluginDir(), 'cats', 'server')
    info['apiRoot'].addTemplateDirectory(templateDir)
    info['apiRoot'].templateFilename = 'cats_api_docs.mako'

The plugin template can inherit from the default template and override the blocks:

    <%inherit file="api_docs.mako"/>

    <%block name="docsBody">
    <p>Warning: This API is not a sandbox; don't get scratched.</p>
    </%block>

To avoid hardcoding the default template filename, pass it as a variable:

    baseTemplateFilename = info['apiRoot'].templateFilename
    info['apiRoot'].updateHtmlVars({
        'baseTemplateFilename': baseTemplateFilename
    })

and get the value in the template like:

    <%inherit file="${context.get('baseTemplateFilename')}"/>

[1] http://docs.makotemplates.org/en/latest/inheritance.html

### DIFF
--- a/docs/plugin-development.rst
+++ b/docs/plugin-development.rst
@@ -508,6 +508,41 @@ login via a username and password. This allows alternative authentication
 modes to be used instead of core, or prior to attempting core authentication.
 The event info contains two keys, "login" and "password".
 
+Customizing the Swagger page
+****************************
+
+To customize text on the Swagger page, create a
+`Mako template <http://www.makotemplates.org/>`_ file that inherits from the
+base template and overrides one or more blocks:
+
+.. code-block:: html+mako
+
+    <%inherit file="${context.get('baseTemplateFilename')}"/>
+
+    <%block name="docsHeader">
+      <span>Cat programming interface</span>
+    </%block>
+
+    <%block name="docsBody">
+      <p>Manage your cats using the resources below.</p>
+    </%block>
+
+Install the template by registering its directory and setting its filename on
+``apiRoot``:
+
+.. code-block:: python
+
+    def load(info):
+        # Set base template filename variable
+        info['apiRoot'].updateHtmlVars({
+            'baseTemplateFilename': info['apiRoot'].templateFilename
+        })
+
+        # Set custom API docs template
+        templateDir = os.path.join(info['pluginRootDir'], 'server')
+        info['apiRoot'].addTemplateDirectory(templateDir)
+        info['apiRoot'].templateFilename = 'custom_api_docs.mako'
+
 .. _client-side-plugins:
 
 Extending the Client-Side Application
--- a/docs/plugin-development.rst
+++ b/docs/plugin-development.rst
@@ -513,7 +513,8 @@ Customizing the Swagger page
 
 To customize text on the Swagger page, create a
 `Mako template <http://www.makotemplates.org/>`_ file that inherits from the
-base template and overrides one or more blocks:
+base template and overrides one or more blocks. For example,
+``plugins/cats/server/custom_api_docs.mako``:
 
 .. code-block:: html+mako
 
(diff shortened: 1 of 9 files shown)


zachmullen:
Good feature, could you add a pytest test for this?

msmolens:
@zachmullen Sure. But first, I'd like to see this test pass:
```
$ pytest test/test_webroot.py
...
test/test_webroot.py::testAccessWebRoot FAILED
```
Note that it does pass if an earlier test used the `server` fixture.

zachmullen:
That indicates that we need to update the `server` fixture to also reset the state of the webroot on teardown.

zachmullen:
Also, we should add a section in the developer docs for this

girder/utility/webroot.py
```diff
@@ -57,6 +56,25 @@ def updateHtmlVars(self, vars):
         self.vars.update(vars)
         self.indexHtml = None
 
+    def addTemplateDirectory(self, templateDir):
+        """
+        Add a directory in which to search for templates.
+        """
+        self._templateLookup.directories.append(templateDir)
```
brianhelba: Do you think it's safe to mutate the `TemplateLookup` instance variable `directories` directly like this? The constructor [does a little bit more work](https://github.com/zzzeek/mako/blob/master/mako/lookup.py#L186) (mostly calling [`posixpath.normpath`](https://docs.python.org/2/library/os.path.html#os.path.normpath)) before placing entries into `directories`.

girder/utility/webroot.py
```diff
@@ -57,6 +56,25 @@ def updateHtmlVars(self, vars):
...
+        """
+        The filename of the template to render. The file should be located in a
+        directory specified by addTemplateDirectory().
+        """
+        return self._templateFilename
+
+    @templateFilename.setter
+    def templateFilename(self, value):
+        self._templateFilename = value
+        self.indexHtml = None
```
brianhelba: After https://github.com/girder/girder/pull/2274/commits/a5b6d83c10cbe91fcff0becb39f0e0f436589890 , it looks like `self.indexHtml` is a dead variable. I think you should remove it everywhere.

girder/utility/webroot.py
```diff
@@ -57,6 +56,25 @@ def updateHtmlVars(self, vars):
...
+        """
+        self._templateLookup.directories.append(templateDir)
+
+    @property
+    def templateFilename(self):
+        """
+        The filename of the template to render. The file should be located in a
+        directory specified by addTemplateDirectory().
+        """
+        return self._templateFilename
```
brianhelba: If `self.indexHtml` is gone (which I think it should be), does the getter / setter do anything? Could you just make `self._templateFilename` -> `self.templateFilename` and eliminate the methods?

docs/plugin-development.rst
```diff
@@ -508,6 +508,41 @@ login via a username and password. This allows alternative authentication
...
+Customizing the Swagger page
+****************************
+
+To customize text on the Swagger page, create a
+`Mako template <http://www.makotemplates.org/>`_ file that inherits from the
+base template and overrides one or more blocks:
+
+.. code-block:: html+mako
+
+    <%inherit file="${context.get('baseTemplateFilename')}"/>
```
brianhelba: Why do you need to use `${context.get('baseTemplateFilename')}`?  Since `self.vars` is injected directly into the `template.render` call, can't you just reference it as `${baseTemplateFilename}`?

docs/plugin-development.rst
```diff
@@ -508,6 +508,41 @@ login via a username and password. This allows alternative authentication
...
+      <p>Manage your cats using the resources below.</p>
+    </%block>
+
+Install the template by registering its directory and setting its filename on
+``apiRoot``:
+
+.. code-block:: python
+
+    def load(info):
+        # Set base template filename variable
```
brianhelba: This example is a little unclear. More comments would help to reflect the fact that:
1. When `info['apiRoot']` was already instantiated by Girder to render `api_docs.mako`, its template directories include the path to `api_docs.mako`
2. `info['apiRoot'].templateFilename` now just stores the string `'api_docs.mako'`
3. setting `info['apiRoot'].addTemplateDirectory(os.path.join(info['pluginRootDir'], 'server'))` is only necessary because the new `info['apiRoot'].templateFilename` is not fully qualified
4. due to 1., the effective lookup `<%inherit file="api_docs.mako"/>` will succeed

Perhaps this example could be restructured to better separate the two distinct steps happening here: the setting of a new `info['apiRoot'].templateFilename` and the use of `%inherit` to reference the old `api_docs.mako`.

docs/plugin-development.rst
msmolens: Mako treats directives like `<%inherit>` differently from the body content. http://docs.makotemplates.org/en/latest/runtime.html#context-variables describes that
>  variable names that are referenced within the function [...] are pulled from the Context object’s dictionary
Looking at the compiled template, the function generated for the inherit directive doesn't pull variables out of the context, so it has to be done manually:
```python
def _mako_inherit(template, context):
    _mako_generate_namespaces(context)
    return runtime._inherit_from(context, (context.get('baseTemplateFilename')), _template_uri)
```

msmolens:
@brianhelba What do you think about replaced `addTemplateDirectory()` with a function that accepts a full path to the template. This would be consistent with the `WebrootBase` constructor and would simplify setting up a custom template. To continue support injecting the base template filename into custom templates, the filename could still be available as a readonly property. This could look like:
```python
def setTemplatePath(templatePath):
    # split path into dir, filename
    # add dir to list
    # make filename available through property
    # invalidate TemplateLookup instance
```

PR #2614, opened 2018-02-15
Add a findWithPermissions method.

### LINKED ISSUE #1603: Allow fetching of total number of items for paging widgets

manthey:
For access controlled models and models that use the acl-mixin, add a `findWithPermissions` method that uses mongo to filter by permissions either by extra query parameters (for ac models) or via an aggregate pipeline (for acl-mixin models).  For models that use the acl-mixin and get permissions from a model that itself uses the acl-mixin or use `attachedToType` rather than `resourceColl`, this falls back to using `filterResultsByPermission`.  If an aggregate pipeline is used, augment the resulting CommandCursor with a `count` method that uses a similar pipeline to get the count (but only on demand).

The where clauses used for groups are different than for other models.

For rest endpoints, support returning generators, Mongo Cursors, and Mongo Command Cursors.  For each of these, the result is coerced to a list as part of the standard endpoint handling.  For Cursors and Command Cursors with a count method, a `Girder-Total-Count` header to the response.  Adding this header resolves #1603.

If the Mongo server is less than version 3.4, this falls back to `filterResultsByPermission`.

Things that could be done:

* For models that use the acl-mixin and refer to a model that uses an acl-mixin, we could do multiple joins (`$lookup`) instead of falling back to filtering in python.

* Contribute to the mongo proxy upstream so that we can use it directly.

* When `findWithPermissions` is called without a user and asking for more than READ permission, a special query is constructed that returns an empty result set.  This results in a consistent result type (a Cursor or CommandCursor).  Is there a better way to do this?

Areas that need testing:

* The mongo proxy with a command cursor.

* We may want to test with Mongo version 3.4 (and Python 2 with Mongo >=3.4).

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -455,7 +455,7 @@ class filtermodel(object):  # noqa: class name
 
             if isinstance(val, _MONGO_CURSOR_TYPES):
                 if callable(getattr(val, 'count', None)):
-                    cherrypy.response.headers['X-Total-Count'] = val.count()
+                    cherrypy.response.headers['Girder-Total-Count'] = val.count()
                 return [model.filter(m, user, self.addFields) for m in val]
             elif isinstance(val, (list, tuple, types.GeneratorType)):
                 return [model.filter(m, user, self.addFields) for m in val]
@@ -605,7 +605,7 @@ def _mongoCursorToList(val):
     # be callable.
     if isinstance(val, _MONGO_CURSOR_TYPES):
         if callable(getattr(val, 'count', None)):
-            cherrypy.response.headers['X-Total-Count'] = val.count()
+            cherrypy.response.headers['Girder-Total-Count'] = val.count()
         val = list(val)
     return val
 
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -24,12 +24,16 @@ import datetime
 import inspect
 import json
 import posixpath
+import pymongo
 import six
 import sys
 import traceback
+import types
 import unicodedata
 
 from dogpile.cache.util import kwarg_function_key_generator
+from girder.external.mongodb_proxy import MongoProxy
+
 from . import docs
 from girder import auditLogger, events, logger, logprint
 from girder.constants import SettingKey, TokenScope, SortDir
@@ -44,6 +48,8 @@ from six.moves import range, urllib
 
 # Arbitrary buffer length for stream-reading request bodies
 READ_BUFFER_LEN = 65536
+
+_MONGO_CURSOR_TYPES = (MongoProxy, pymongo.cursor.Cursor, pymongo.command_cursor.CommandCursor)
 
 
 def getUrlParts(url=None):
@@ -447,7 +453,11 @@ class filtermodel(object):  # noqa: class name
 
             user = getCurrentUser()
 
-            if isinstance(val, (list, tuple)):
+            if isinstance(val, _MONGO_CURSOR_TYPES):
+                if callable(getattr(val, 'count', None)):
+                    cherrypy.response.headers['X-Total-Count'] = val.count()
+                return [model.filter(m, user, self.addFields) for m in val]
+            elif isinstance(val, (list, tuple, types.GeneratorType)):
                 return [model.filter(m, user, self.addFields) for m in val]
             elif isinstance(val, dict):
                 return model.filter(val, user, self.addFields)
@@ -583,6 +593,23 @@ def _logRestRequest(resource, path, params):
     })
 
 
+def _mongoCursorToList(val):
+    """
+    If the specified value is a Monog cursor, convert it to a list.
+    Otherwise, just return the passed values.
+
+    :param val: a value that might be a Mongo cursor.
+    :returns: a list if val was a Mongo cursor, otherwise the original val.
+    """
+    # This needs to be before the callable check, as mongo cursors can
+    # be callable.
+    if isinstance(val, _MONGO_CURSOR_TYPES):
+        if callable(getattr(val, 'count', None)):
+            cherrypy.response.headers['X-Total-Count'] = val.count()
+        val = list(val)
+    return val
+
+
 def endpoint(fun):
     """
     REST HTTP method endpoints should use this decorator. It converts the return
@@ -606,6 +633,8 @@ def endpoint(fun):
             if 'Content-Range' in cherrypy.response.headers:
                 cherrypy.response.status = 206
 
+            val = _mongoCursorToList(val)
+
             if callable(val):
                 # If the endpoint returned anything callable (function,
                 # lambda, functools.partial), we assume it's a generator
@@ -617,6 +646,9 @@ def endpoint(fun):
             if isinstance(val, cherrypy.lib.file_generator):
                 # Don't do any post-processing of static files
                 return val
+
+            if isinstance(val, types.GeneratorType):
+                val = list(val)
 
         except RestException as e:
             val = _handleRestException(e)
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -595,7 +595,7 @@ def _logRestRequest(resource, path, params):
 
 def _mongoCursorToList(val):
     """
-    If the specified value is a Monog cursor, convert it to a list.
+    If the specified value is a Mongo cursor, convert it to a list.
     Otherwise, just return the passed values.
 
     :param val: a value that might be a Mongo cursor.
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -45,6 +45,37 @@ _allowedFindArgs = ('cursor_type', 'allow_partial_results', 'oplog_replay',
 _modelSingletons = []
 
 
+def _permissionClauses(user=None, level=None, prefix=''):
+    """
+    Given a user and access level, return a list of clauses that can be used as
+    part of a Mongo find query or aggregate match step.
+
+    :param user: The user to check policies against.
+    :type user: dict or None
+    :param level: The access level.  Explicitly passing None skips doing
+        permissions checks.
+    :type level: AccessType
+    :param prefix: an optional string to append to the keys used in the
+        clauses.
+    :type prefix: str
+    :returns: A query dictionary with an '$or' entry which consists of a list
+        of match clauses, any one of which implies validation.
+    """
+    permissionClauses = [
+        {prefix + 'public': True},
+    ]
+    if user:
+        permissionClauses.extend([
+            {prefix + 'access.users': {'$elemMatch': {
+                'id': user['_id'],
+                'level': {'$gte': level}}}},
+            {prefix + 'access.groups': {'$elemMatch': {
+                'id': {'$in': user.get('groups', [])},
+                'level': {'$gte': level}}}},
+        ])
+    return {'$or': permissionClauses}
+
+
 class _ModelSingleton(type):
     def __init__(cls, name, bases, dict):
         super(_ModelSingleton, cls).__init__(name, bases, dict)
@@ -94,6 +125,7 @@ class Model(ModelImporter):
         typically not have to call this method.
         """
         db_connection = getDbConnection()
+        self._dbserver_version = tuple(db_connection.server_info()['versionArray'])
         self.database = db_connection.get_database()
         self.collection = MongoProxy(self.database[self.name])
 
@@ -311,6 +343,27 @@ class Model(ModelImporter):
         kwargs = {k: kwargs[k] for k in kwargs if k in _allowedFindArgs}
         return self.collection.find_one(query, projection=fields, **kwargs)
 
+    def _textSearchFilters(self, query, filters=None, fields=None):
+        """
+        Return a set of filters and fields used in the text search.
+
+        :param query: The text query. Will be stemmed internally.
+        :type query: str
+        :param filters: Any additional query operators to apply.
+        :type filters: dict
+        :param fields: A mask for filtering result documents by key, or None to return the full
+            document, passed to MongoDB find() as the `projection` param.
+        :type fields: `str, list of strings or tuple of strings for fields to be included from the
+            document, or dict for an inclusion or exclusion projection`.
+        :returns: (filters, fields) to be passed to the query.
+        """
+        filters = filters or {}
+        fields = fields or {}
+
+        fields['_textScore'] = {'$meta': 'textScore'}
+        filters['$text'] = {'$search': query}
+        return filters, fields
+
     def textSearch(self, query, offset=0, limit=0, sort=None, fields=None,
                    filters=None, **kwargs):
         """
@@ -333,11 +386,7 @@ class Model(ModelImporter):
         :returns: A pymongo cursor. It is left to the caller to build the
             results from the cursor.
         """
-        filters = filters or {}
-        fields = fields or {}
-
-        fields['_textScore'] = {'$meta': 'textScore'}
-        filters['$text'] = {'$search': query}
+        filters, fields = self._textSearchFilters(query, filters, fields)
 
         cursor = self.find(filters, offset=offset, limit=limit,
                            sort=sort, fields=fields)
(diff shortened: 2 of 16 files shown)


manthey:
@brianhelba We discussed the possibility of deprecating the `list` method.  ACL models have a `list` method, but the base non-acl model does not.  Several non-acl models also have a `list` method (assetstore and upload for two).  In these cases, the signature of `list` is different from the acl version.  `findWithPermissions` is a superset of `list`, with a `query` parameter (and `fields`, `timeout`, and `level).

Perhaps it would make more sense to alter the acl `list` method (it would just be the current `findWithPermissions` method except with the args in a different order to be compatible with the current order in `list`).

mgrauer:
@manthey What is the status of this PR? Is it still WIP?

manthey:
@mgrauer It needs more tests.  Apparently no one has an issue with the approach or changes, so I hope to add tests "soon" and then it will not longer be WIP.

manthey:
This now has tests and is ready for review.

girder/api/rest.py
```diff
@@ -618,6 +647,9 @@ def endpointDecorator(self, *path, **params):
                 # Don't do any post-processing of static files
                 return val
 
+            if isinstance(val, types.GeneratorType):
+                val = list(val)
```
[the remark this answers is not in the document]
manthey: This is new functionality.  Before (and still), if something wanted to stream results, it returned a function which would return a generator.  Now, it can alternately return a generator and it will be coerced into a list.

There were several places where we were manually coercing generators into lists (for instance, in find assetstores, getAssetstoreFiles, listing groups, etc).  This avoids having to wrap those in `list()` in individual endpoints.

If we want to further add a streaming style to output a generator to a json list, that would be new functionality (which should be another PR).

girder/api/rest.py
```diff
@@ -447,7 +453,11 @@ def wrapped(*args, **kwargs):
 
             user = getCurrentUser()
 
-            if isinstance(val, (list, tuple)):
+            if isinstance(val, _MONGO_CURSOR_TYPES):
+                if callable(getattr(val, 'count', None)):
+                    cherrypy.response.headers['X-Total-Count'] = val.count()
```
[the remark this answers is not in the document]
manthey: Sure.  Now that I look into it, it has been discouraged since at least 2012 via [RFC 6648](https://tools.ietf.org/html/rfc6648).

girder/models/model_base.py
```diff
@@ -45,6 +45,37 @@
...
+    """
+    Given a user and access level, return a list of clauses that can be used as
+    part of a Mongo find query or aggregate match step.
+
+    :param user: The user to check policies against.
+    :type user: dict or None
+    :param level: The access level.  Explicitly passing None skips doing
+        permissions checks.
+    :type level: AccessType
+    :param prefix: an optional string to append to the keys used in the
```
zachmullen: prepend*

girder/models/model_base.py
```diff
@@ -45,6 +45,37 @@
 _modelSingletons = []
 
 
+def _permissionClauses(user=None, level=None, prefix=''):
```
zachmullen: Is there someone downstream who we know will use this `prefix` feature? It doesn't look like we're using it in this repo.

girder/models/model_base.py
```diff
@@ -1511,10 +1581,52 @@ def prefixSearch(self, query, user=None, filters=None, limit=0, offset=0,
...
+            permissions checks.
+        :type level: AccessType
+        :returns: A pymongo Cursor or CommandCursor.  If a CommandCursor, it
+            has been augmented with a count function.
+        """
+        # If no user is specified and greater than READ access is requested,
+        # we won't return anything.  So as to always return a cursor for
+        # consistency, we make a query that will return no results
+        if user is None and level is not None and level > AccessType.READ:
+            query = {'__matchnothing': 'nothing'}
```
zachmullen: Is there a better way to fake this that won't hit the database at all? I worry that a query like this might actually be slower than leaving this condition out, but I'd have to do some reading/experiments to be sure.

girder/utility/acl_mixin.py
```diff
@@ -193,9 +201,162 @@ def prefixSearch(self, query, user=None, filters=None, limit=0, offset=0,
...
+            aggregation is used.
+        :type aggregateSort: List of (key, order) tuples.
+        :returns: A pymongo Cursor, CommandCursor, or an iterable.  If a
+            CommandCursor, it has been augmented with a count function.
+        """
+        # If no user is specified and greater than READ access is requested,
+        # we won't return anything.  So as to always return a cursor for
+        # consistency, we make a query that will return no results
+        if user is None and level is not None and level > AccessType.READ:
+            query = {'__matchnothing': 'nothing'}
```
zachmullen: same here

girder/utility/acl_mixin.py
```diff
@@ -193,9 +201,162 @@ def prefixSearch(self, query, user=None, filters=None, limit=0, offset=0,
...
+                {'$match': query},
+                {'$lookup': {
+                    'from': self.resourceColl,
+                    'localField': self.resourceParent,
+                    'foreignField': '_id',
+                    'as': '_access'
+                }},
+                {'$match': self.permissionClauses(user, level, '_access.')},
+            ]
+            countPipeline = initialPipeline[:] + [
```
zachmullen: Why copy `initialPipeline` here and line 328?

girder/utility/acl_mixin.py
```diff
@@ -193,9 +201,162 @@ def prefixSearch(self, query, user=None, filters=None, limit=0, offset=0,
...
+            ]
+            if sort is not None or aggregateSort is not None:
+                fullPipeline.append({'$sort': collections.OrderedDict(sort or aggregateSort)})
+            # limit should immediately follow sort for efficiency
+            if limit:
+                fullPipeline.append({'$limit': limit + (offset or 0)})
+            if offset:
+                fullPipeline.append({'$skip': offset})
+            if fields is not None:
+                if any([isinstance(v, bool) for v in fields.values()]):
```
zachmullen: Prefer `any(isinstance(v, bool) for v in fields.values())`

girder/models/model_base.py
```diff
@@ -1511,10 +1581,52 @@ def prefixSearch(self, query, user=None, filters=None, limit=0, offset=0,
...
+        :param offset: The offset into the results
+        :type offset: int
+        :param limit: Maximum number of documents to return
+        :type limit: int
+        :param timeout: Cursor timeout in ms. Default is no timeout.
+        :type timeout: int
+        :param fields: A mask for filtering result documents by key, or None to return the full
+            document, passed to MongoDB find() as the `projection` param.
+        :type fields: `str, list of strings or tuple of strings for fields to be included from the
+            document, or dict for an inclusion or exclusion projection`.
```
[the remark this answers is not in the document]
manthey: I copied this from `find`.  I'll fix a bunch of docstrings.

girder/models/model_base.py
manthey: The acl_mixin uses a prefix of `'_access.'`.

girder/utility/acl_mixin.py
manthey: I'm going to assume I was either delusional or it was because I did something different a long time ago and failed to change it.

girder/models/model_base.py
manthey: The virtue of faking it this way is that we return an actual pymongo `Cursor`, so it behaves as expected.  I'm sure it takes more time than returning a iterator with a synthetic count function, but it is probably a fairly rare condition -- no user asking for write or better access.  Returning something else means that other code could need a special code path.

The other alternative would be to generate a Mongo Cursor that is more faked.  For instance, after the `self.find()` returns a Cursor, it could be monkey-patched by reaching in and saying `<cursor>._Cursor__empty = True` or perhaps calling `<cursor>._Cursor__die()`, which would prevent any round-trip to Mongo from occurring.  This seems hackier (and less stable -- though it could be done if the `_Cursor__die` function exists and not if it doesn't).

girder/utility/acl_mixin.py
```diff
@@ -193,9 +201,162 @@ def prefixSearch(self, query, user=None, filters=None, limit=0, offset=0,
...
+        :param aggregateSort: A sort order to use if `sort` is None and an
+            aggregation is used.
+        :type aggregateSort: List of (key, order) tuples.
+        :returns: A pymongo Cursor, CommandCursor, or an iterable.  If a
+            CommandCursor, it has been augmented with a count function.
+        """
+        # If no user is specified and greater than READ access is requested,
+        # we won't return anything.  So as to always return a cursor for
+        # consistency, we make a query that will return no results
+        if user is None and level is not None and level > AccessType.READ:
```
brianhelba: I'd cut this, as a premature optimization.

girder/models/model_base.py
```diff
@@ -45,6 +45,37 @@
...
+        permissions checks.
+    :type level: AccessType
+    :param prefix: an optional string to prepend to the keys used in the
+        clauses.
+    :type prefix: str
+    :returns: A query dictionary with an '$or' entry which consists of a list
+        of match clauses, any one of which implies validation.
+    """
+    permissionClauses = [
+        {prefix + 'public': True},
```
zachmullen: Since `public` only applies to `level=AccessType.READ`, should this be conditioned on the `level` parameter?

girder/utility/acl_mixin.py
```diff
@@ -193,9 +201,162 @@ def prefixSearch(self, query, user=None, filters=None, limit=0, offset=0,
...
+                    **kwargs)
+
+            query = query or {}
+            initialPipeline = [
+                {'$match': query},
+                {'$lookup': {
+                    'from': self.resourceColl,
+                    'localField': self.resourceParent,
+                    'foreignField': '_id',
+                    'as': '_access'
```
brianhelba: Can we call this `__parent` instead of `_access`?

girder/utility/acl_mixin.py
```diff
@@ -193,9 +201,162 @@ def prefixSearch(self, query, user=None, filters=None, limit=0, offset=0,
...
+            # limit should immediately follow sort for efficiency
+            if limit:
+                fullPipeline.append({'$limit': limit + (offset or 0)})
+            if offset:
+                fullPipeline.append({'$skip': offset})
+            if fields is not None:
+                if any(isinstance(v, bool) for v in fields.values()):
+                    fullPipeline.append({'$project': fields})
+                else:
+                    fullPipeline.append({'$addFields': fields})
```
brianhelba: This `$addFields` case should be cut, as injecting constant data is not something we need to support. We should expect that `fields` is always either, at most:
* A Python `set` of fields that should all be kept
* A Python `dict`, where all values are `bool`
* A Python `dict`, where all values are `int` of values `1` or `0`

I'd be open to further restricting the allowed types of `fields`.

zachmullen:
```
if any(isinstance(v, bool) for v in fields.values()):
    fullPipeline.append({'$project': fields})
else:
    fullPipeline.append({'$addFields': fields})
```

Let's replace this with just

    fullPipeline.append({'$project': fields})

zachmullen:
Timeline:

* Reviewers should have initial comments made by end of day tomorrow Sep 5.
* @manthey  will hopefully address the comments with new commits by end of day Sep 6.
* Reviewers please do a final review pass before end of day Sep 7; if you fail to make new comments or mark approval, the PR will be merged.

girder/utility/acl_mixin.py
manthey: This actually gets used as part of the `_textSearchFilters` where the field `_textScore` is set to `{'$meta': 'textScore'}`.  This means that this should probably be changed to:
```
            if fields is not None:
                if isinstance(fields, (abc.Sequence, abc.Set)):
                    fields = dict.fromkeys(fields, 1)
                if any(not isinstance(v, abc.Mapping) for v in fields.values()):
                    fullPipeline.append({'$project': {
                        k: v for k, v in six.iteritems(fields)
                        if not isinstance(v, abc.Mapping)}})
                if any(isinstance(v, abc.Mapping) for v in fields.values()):
                    fullPipeline.append({'$addFields': {
                        k: v for k, v in six.iteritems(fields)
                        if isinstance(v, abc.Mapping)}})
```

manthey:
With aggregations, we can use the `allowDiskUse` flag which will allow sorting arbitrarily large results (with enough diskspace).  Using an ordinary cursor, we can't sort non-indexed fields if all involved documents exceed 32Mb.  For `textSearch`, we have a value `TEXT_SCORE_SORT_MAX`, and when there are less than that number of documents, we sort.

Since when the user is an admin, we don't *need* to use an aggregation for permissions, this means that  for AC Models and for admin users, we don't sort text searches that return any significant number of results.  But, for ACL Mixin models and *non-admin* users, we *can* (and do) sort text searches.

It seems counterintuitive to have an admin get worse results than a non-admin.  We could *always* use aggregations (but that must have some performance implications).  We could avoid sorting numerous results from non-admins (which takes extra effort).

As a real world example, I have a collection with ~7000 items.  These items have names like "TCGA-02-0001-01Z-00-DX1.83fce43e-42ac-4dcd-b156-2908e75f2e47.svs".  When I do a text search based on fragments of this name from the topbar of Girder's web api (for instance, for "TCGA-EI-6883"), the non-admin user will receive a list where the first item starts with the typed string is at the top.  The admin user receives a random jumble of items that contain at least one of the substrings ("TCGA", "EI", "6883"), none of which are the item that starts with the string I typed.

This PR improves the behavior on the master branch in at least non-admins get items back from a text search in a sorted order.

zachmullen:
Anyone know what's up with the coverage?

girder/models/model_base.py
```diff
@@ -45,6 +45,44 @@
...
+    permissionClauses = []
+    if level is None or (user and user['admin']):
+        # Without a level or with an admin user, match everything.
+        permissionClauses.append({})
+    elif level <= AccessType.READ:
+        permissionClauses.append({prefix + 'public': True})
+    elif not user:
+        # If we have no user and asked for higher than read access, make a
+        # query that will fail
+        permissionClauses.append({'__matchnothing': 'nothing'})
```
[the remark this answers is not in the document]
zachmullen: Also since we are `$or`ing these together, doesn't that mean this won't change any behavior?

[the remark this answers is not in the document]
manthey: Something has to return no data when there is no user and a above read level.  We can certainly use a different permission clause, but we have to have *some* permission clause or have a special case where sometimes we return a cursor and sometimes we don't.  Instead of this being a special case wherever permission clauses might be needed, this is now in a single location where we generate the permission clause.

manthey:
codecov switched from heroku to google cloud.  Older branches seem to have been prejudicially affected.

girder/models/model_base.py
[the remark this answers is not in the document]
manthey: In this case there will be exactly one clause that gets `$or`ed, which requires that at least one clause applies.  So it will get used.

girder/models/model_base.py
```diff
@@ -45,6 +45,44 @@
...
+    :param prefix: an optional string to prepend to the keys used in the
+        clauses.
+    :type prefix: str
+    :returns: A query dictionary with an '$or' entry which consists of a list
+        of match clauses, any one of which implies validation.
+    """
+    permissionClauses = []
+    if level is None or (user and user['admin']):
+        # Without a level or with an admin user, match everything.
+        permissionClauses.append({})
```
zachmullen: Don't we want to `return` here? If `level=None` but we have a user, the below block will execute with `'$gte': None`, which seems like it would cause an error.

girder/models/model_base.py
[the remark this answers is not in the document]
zachmullen: To make that more explicit / readable, can we `return` in this block instead of appending to the list of clauses?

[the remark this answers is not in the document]
manthey: Will do.

girder/models/model_base.py
manthey: I'll change this to return here, too.

manthey:
As per the stated timeline (and two approvals), I'm merging this.

PR #2615, opened 2018-02-16
Add some new client Model methods

brianhelba:
* Add a `FolderModel.removeContents` method to the client library
* Add an `ItemModel.getFiles` method to the client library
* Move an unused utility method to a more appropriate place
  * This moves the method from `UserAccountView.temporaryPassword` to `UserModel.fromTemporaryToken`, and updates the logic to return a promise and no longer perform global routing.
* Start using `UserModel.fromTemporaryToken`, reducing code duplication


brianhelba:
I'm using these in a downstream project, but since they utilize Girder API methods, I wanted to contribute them upstream.

PR #2616, opened 2018-02-16
Move CLI commands to a "cli" module

brianhelba:
* Move CLI commands to a "cli" module
* Deprecate legacy "python -m ..." methods for starting Girder
  * The CLI console scripts should be used instead.
* Don't use warnings.warn for runtime server code  …
  * The logger should be used directly, to ensure the output is available for all log handlers.
* Document a coverage exclusion policy for Python code

### DIFF
--- a/clients/python/girder_client/__main__.py
+++ b/clients/python/girder_client/__main__.py
@@ -18,8 +18,8 @@
 ###############################################################################
 
 import warnings
-from girder_client import cli  # pragma: no cover
+from girder_client import cli
 
 if __name__ == '__main__':
     warnings.warn('Deprecation notice: Use "girder-client" to run the CLI.', DeprecationWarning)
-    cli.main()  # pragma: no cover
+    cli.main()
--- a/clients/python/girder_client/__main__.py
+++ b/clients/python/girder_client/__main__.py
@@ -17,9 +17,9 @@
 #  limitations under the License.
 ###############################################################################
 
-import warnings
+import click
 from girder_client import cli
 
 if __name__ == '__main__':
-    warnings.warn('Deprecation notice: Use "girder-client" to run the CLI.', DeprecationWarning)
+    click.echo('Deprecation notice: Use "girder-client" to run the CLI.', err=True)
     cli.main()
--- a/clients/python/girder_client/__main__.py
+++ b/clients/python/girder_client/__main__.py
@@ -17,7 +17,9 @@
 #  limitations under the License.
 ###############################################################################
 
+import warnings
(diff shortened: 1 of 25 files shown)


clients/python/girder_client/__main__.py
```diff
@@ -17,7 +17,9 @@
...
 ###############################################################################
 
-from girder_client import cli  # pragma: no cover
+import warnings
+from girder_client import cli
 
-
-cli.main()  # pragma: no cover
+if __name__ == '__main__':
+    warnings.warn('Deprecation notice: Use "girder-client" to run the CLI.', DeprecationWarning)
```
jbeezley: I tried running `python -m girder_client`, and I'm not seeing this warning.  Maybe I have something in my environment disabling it...

girder/__main__.py
```diff
@@ -17,42 +17,23 @@
...
-    server.setup(testing)
-
-    cherrypy.engine.start()
-    cherrypy.engine.block()
-
-
-if __name__ == '__main__':  # pragma: no cover
-    main()
+if __name__ == '__main__':
+    girder.logger.warning('Deprecation notice: Use "girder serve" to start Girder.')
```
jbeezley: I see the warning here in `~/.girder/logs/error.log` but not on the terminal.  Is that expected?

clients/python/girder_client/__main__.py
brianhelba: Fixed to always print.

girder/__main__.py
brianhelba: Fixed.

clients/python/girder_client/__main__.py
```diff
@@ -17,9 +17,9 @@
 #  limitations under the License.
 ###############################################################################
 
-import warnings
+import six
 from girder_client import cli
 
 if __name__ == '__main__':
-    warnings.warn('Deprecation notice: Use "girder-client" to run the CLI.', DeprecationWarning)
+    six.print_('Deprecation notice: Use "girder-client" to run the CLI.')
```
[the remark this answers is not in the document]
brianhelba: Done. I was trying to be consistent with the behavior of `girder.logprint`, but I agree that `stderr` is more appropriate.

brianhelba:
Thanks to a hint from @danlamanna , I've switched the girder_client warnings to use `click.echo` to `stderr`. The server will continue to use `girder.logprint` (which does go to `stdout`), to ensure that it's goes to the logger, which is important when the server is run as a daemon.

@jbeezley Can you re-approve?

PR #2617, opened 2018-02-20
pydicom has renamed the importable library to "pydicom", from "dicom"

brianhelba:
See: http://pydicom.github.io/pydicom/stable/transition_to_pydicom1.html

### DIFF
--- a/plugins/dicom_viewer/requirements.txt
+++ b/plugins/dicom_viewer/requirements.txt
@@ -1,1 +1,1 @@
-pydicom==0.9.9
+pydicom>=1.0.1rc1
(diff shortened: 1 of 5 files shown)


brianhelba:
@girder/developers PTAL ASAP. Until this goes in, Girder is currently broken on master. The tests are only currently passing due to a stale cache of Python dependencies.

zachmullen:
This touches a reasonable amount of code in two different plugins; if we want a quick fix to get in ASAP, my vote would be to instead open a PR that pins us to the *old* version of pydicom (which should just be a 2 line change) and then we can review this PR less hastily.

brianhelba:
@zachmullen That's a fair point: https://github.com/girder/girder/pull/2619

brianhelba:
Note, I also manually tested the `dicom_viewer` and `thumbnails` plugins. Both still decode DICOM images.

PR #2620, opened 2018-02-21
Virtual folders plugin

zachmullen:
This has been moved to a plugin, but I've scoped it down so that a virtual folder can only contain items, not subfolders, which obviates any headaches related to infinite recursion. This doesn't preclude us from supporting virtual subfolder lookups in the future, but as long as just the items lookup meets our needs, this should be a lot less scary.


zachmullen:
Ready for review.

plugins/virtual_folders/server/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+    if 'virtualItemsQuery' in doc:
+        try:
+            json.loads(doc['virtualItemsQuery'])
+        except (TypeError, ValueError):
+            raise ValidationException(
+                'The virtual items query must be valid JSON.', field='virtualItemsQuery')
+
+    if 'virtualItemsSort' in doc:
+        try:
+            json.loads(doc['virtualItemsSort'])
```
manthey: Should we go further and ensure that this could be used in a mongo query?  Or, just give an example in the param description (I never remember that it is of the form `[[(field), (direction)], [(field), (direction)]]`.

zachmullen: I think that'd be a good feature, although I might want to do it on a different PR.

plugins/virtual_folders/server/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+        return  # This is not a child listing request
+
+    user = self.getCurrentUser()
+    folder = Folder().load(params['folderId'], user=user, level=AccessType.READ)
+
+    if not folder.get('isVirtual') or 'virtualItemsQuery' not in folder:
+        return  # Parent is not a virtual folder, proceed as normal
+
+    limit, offset, sort = self.getPagingParameters(params, defaultSortField='name')
+    q = json.loads(folder['virtualItemsQuery'], object_hook=json_util.object_hook)
```
manthey: Should we be using `bson.json_util.loads` rather than passing `json_util.object_hook` to `json.loads`?  It uses `object_hook_pairs` rather than `object_hook` (at least in the versions of Python we care about).

PR #2621, opened 2018-02-21
Remove the deprecated and insecure "DELETE /user/password" route

brianhelba:


mgrauer:
@brianhelba Should this PR get a security tag and be back-ported?

brianhelba:
I've gone ahead and tagged it as security. I don't personally think backporting is necessary, unless others feel strongly.

zachmullen:
I'd vote for not backporting this one.

PR #2631, opened 2018-02-27
Make build type available to front-end code

msmolens:
Use DefinePlugin to set `process.env.NODE_ENV` to `"development"` or `"production"`. This allows writing debug code that can be eliminated when building for production. Frameworks such as Vue.js use this variable to decide whether to build in production mode.

### DIFF
--- a/grunt_tasks/build.js
+++ b/grunt_tasks/build.js
@@ -119,6 +119,15 @@ module.exports = function (grunt) {
         });
     }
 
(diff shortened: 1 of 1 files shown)


jbeezley:
FYI, I'm working on fixing the assetstore test failure.

msmolens:
Rebased on top of assetstore test fix.

PR #2633, opened 2018-02-27
Add a getLocalFilePath method to the File model.

manthey:
The method `getLocalFilePath` is added to the File model, and will return `None` if there is no such path.

The same method name is used in the AbstractAssetstoreAdapter and inherited by the other adapters.  The filesystem assetstore adapter overrides this to provide a path.  The Fuse plugin overrides the AbstractAssetstoreAdapter to report a fuse path if one exists.

This allows any consumer (such as the worker or metadata extractor plugins) to ask for `File().getLocalFilePath(file)`, and, if it isn't `None`, to use that path.

If the Fuse plugin is enabled, the metadata extractor will now work on any file.  This also fixes a bug where a file that was attached to a resource (rather than on an Item) would throw an exception in the metadata extractor.

### DIFF
--- a/girder/utility/abstract_assetstore_adapter.py
+++ b/girder/utility/abstract_assetstore_adapter.py
@@ -21,7 +21,7 @@ import six
 
 from girder.api.rest import setResponseHeader, setContentDisposition
 from girder.constants import SettingKey
-from girder.exceptions import GirderException, ValidationException
+from girder.exceptions import GirderException, ValidationException, FilePathException
 from girder.models.setting import Setting
 from girder.utility import progress, RequestBodyStream
 from .model_importer import ModelImporter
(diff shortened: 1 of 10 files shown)


manthey:
@kotfic We discussed this, if you'd care to take a look.

zachmullen:
I think I'm OK with this, but I want to hear from the segment of our developers who are more opposed to core feature creep than I am. @brianhelba can you take a quick look today?

brianhelba:
I can later today.

girder/utility/abstract_assetstore_adapter.py
```diff
@@ -440,3 +440,13 @@ def open(self, file):
...
+
+    def getLocalFilePath(self, file):
+        """
+        If an assetstore adapter supports it, return a path to the file on the
+        local file system.
+
+        :param file: The file document.
+        :returns: a local path to the file or None if no such path is known.
+        """
+        return None
```
brianhelba: Can this raise an exception? This would be preferable for two reasons:
* The return type could be a single type (`str`). IMO, Girder abuses Python's dynamic typing a bit too much, such that conditional checks are necessary to handle lots of seemingly-simple function outputs.
* In cases where a path is expected to "just work" you can call the method without having to validate the output. In those cases, an uncaught exception would actually be desirable, as it indicates a failure of downstream assumptions.
* In cases where a path may not be available, it allows more information on the cause of the failure to be returned.
* It allows more complex implementations of `getLocalFilePath` (e.g. those with nested function calls) to efficiently abort when they fail.

kotfic:
This has been hanging for awhile now and i don't want responsibility for review to get lost in the shuffle. If David's latest change resolves your issues @brianhelba  will you make sure to approve.  Thanks!

PR #2634, opened 2018-03-01
Add a pytest marker to enable plugins

jbeezley:
This commit makes it possible to write pytest style tests for girder plugins.  It is based on the interface that will be provided by pip installable plugins in girder 3.  The new marker is incompatible with the "testPlugin" marker, which mocks the normal plugin load mechanism.  This could in theory be fixed, but it isn't clear there is a use case for enabling both "test plugins" and "real plugins" in a single test.

I added a new test to the jobs plugin as a proof-of-concept.  For the moment, each plugin with pytest tests needs to be added to the flags in `setup.cfg`.  I didn't check if globbing works in these flags. Maybe @danlamanna or @brianhelba could suggest an alternative configuration.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,7 +9,9 @@ parallel = True
 # Include sources from installed package in Tox's {envsitepackagesdir}
 girder =
     girder/
+    plugins/*/server/
     build/test/tox/*/lib/*/site-packages/girder/
+    build/test/tox/*/lib/*/site-packages/girder/plugins/*/server/
 [coverage:html]
 directory = build/test/artifacts/python_coverage
 title = Girder Coverage Report
@@ -77,6 +79,6 @@ exclude: girder/external/*
 ignore: D100,D101,D102,D103,D104,D105,D106,D107,D200,D201,D202,D203,D204,D205,D400,D401,D402,E123,E226,E241,N802,N803,N806,N812,W503
 
 [tool:pytest]
-addopts = --verbose --strict --showlocals --cov=girder --cov-report=""
+addopts = --verbose --strict --showlocals --cov=girder --cov=plugins/jobs/server --cov-report=""
 cache_dir = build/test/pytest_cache
 testpaths = test
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,14 +4,20 @@ branch = True
 omit =
     girder/external/*
     test/*
+include =
+    girder/*
+    plugins/*/server/*
+    build/test/tox/*/lib/*/site-packages/girder/*
+    build/test/tox/*/lib/*/site-packages/girder/plugins/*/server/*
 parallel = True
 [coverage:paths]
 # Include sources from installed package in Tox's {envsitepackagesdir}
 girder =
     girder/
-    plugins/*/server/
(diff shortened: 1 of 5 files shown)


jbeezley:
Apparently, I need to fix coverage reporting on CI.

setup.cfg
```diff
@@ -77,6 +80,6 @@ exclude: girder/external/*
 ignore: D100,D101,D102,D103,D104,D105,D106,D107,D200,D201,D202,D203,D204,D205,D400,D401,D402,E123,E226,E241,N802,N803,N806,N812,W503
 
 [tool:pytest]
-addopts = --verbose --strict --showlocals --cov=girder --cov-report=""
+addopts = --verbose --strict --showlocals --cov=girder --cov=plugins/jobs/server --cov-report=""
 cache_dir = build/test/pytest_cache
-testpaths = test
+testpaths = test plugins/jobs/test
```
zachmullen: I was thinking about this the other day ... would it make sense to move the built-in plugin tests underneath the main "tests" directory, or do you think it is best to keep them under the plugin?

jbeezley: That's a good point.  For now, it would probably make more sense to move them to `test/` to avoid this complication.  For Girder 3, I'm less convinced it is a good idea.

zachmullen: We may also be able to use a globbing expression to avoid enumerating every plugin in the `--cov` flag.

jbeezley: I just confirmed globbing expressions don't work for the `--cov` flag.

brianhelba:
The `--cov` flag evaluates to the `source` option of the Coverage API, which is:
> a list of file paths or package names. Only code located in the trees indicated by the file paths or package names will be measured.

It may actually be possible to drop using this entirely, and use `include` / `omit` in the config file.

I'll take a look at this possibility. I'd also like a chance to comment on this final PR before it's merged.

zachmullen:
Looks like the coverage reporter can't find the plugins source files...

jbeezley:
Yeah, there are issues with merging the coverage files between tox and cmake.  Tox is using the installed package where plugins are *inside* the girder package, where cmake is using the source directory where plugins are *outside* the girder package.  I **thought** the `[coverage:paths]` options were supposed to resolve this, but I was unable to make it work.

In regards to @brianhelba's comment about the `--cov` argument versus `include` and `source` options, there is some black magic that `pytest-cov` does to the configuration with the `--cov` option that will actually disable the other options.  I think the current config is close to what you are describing modulo the broken `coverage combine` step.

jbeezley:
I think I resolved the coverage problem in my last commit.  We'll see it works on CI.  :crossed_fingers:

jbeezley:
There was a failure in the tests on merge to master with the coverage reporting.  I'm not sure if it represents a randomly occurring problem or was just a fluke.  It's possible we may need to revert this if it starts to cause problems.

I have made an [issue](https://bitbucket.org/ned/coveragepy/issues/649/explicit-path-alias-matching-order-for) on coveragepy's repo to fix our problem for real rather than hack around it.  Even if a patch is accepted, it will probably take a while before it lands in a release.

PR #2636, opened 2018-03-05
Reinstate py34

zachmullen:
@girder/developers I'd like you to consider this PR. I'm not sure if it's a good idea myself, but I think there needs to be a bit more discussion on this topic before we come to a final conclusion about python 3.4.

While I generally approve of the spirit of #2572, I had some trepidation about the parts of the code where we went out of our way to make sure Girder would break if someone tried to install it on python 3.4. I understand why those changes were made, but given that we have downstream users who still would like to use python 3.4, we should reconsider whether that was a good or a bad decision.

Given that we don't strictly need to do anything in our code that will work on python 2.7 and 3.5 but not 3.4, I see no harm in preserving the *present* support for python 3.4, even though we will continue to state in our docs that it is not supported.

@brianhelba @jamiesnape @mgrauer @danlamanna PTAL

### DIFF
--- a/girder/api/describe.py
+++ b/girder/api/describe.py
@@ -19,6 +19,10 @@
 
 import bson.json_util
 import dateutil.parser
+try:
+    from inspect import signature, Parameter
+except ImportError:
+    from funcsigs import signature, Parameter
 import inspect
 import jsonschema
 import os
@@ -37,11 +41,6 @@ from girder.utility.webroot import WebrootBase
 from girder.utility.resource import _apiRouteMap
 from . import docs, access
 from .rest import Resource, getApiUrl, getUrlParts
-
-if six.PY3:
-    from inspect import signature, Parameter
-else:
-    from funcsigs import signature, Parameter
 
 """
 Whenever we add new return values or new options we should increment the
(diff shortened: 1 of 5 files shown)


mgrauer:
I'm ok with this change from the perspective of supporting a downstream, though if we don't have CI on 3.4 we may face future breakages obviously, and we probably don't want 3.4 CI if we are claiming it isn't supported.

For more background here, AWS Elastic Beanstalk has a single [preconfigured Docker container for Python](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/concepts.platforms.html#concepts.platforms.dockerpreconfig), which is Python 3.4/nginx/uWSGI. Once they add support for another version of Python here we can really ignore Python 3.4 from now on. Beanstalk has [Python/Apache/WSGI](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/concepts.platforms.html#concepts.platforms.python) support for versions 2.7, 3.4 and 3.6. Most current Girder deployments that I am aware of are Python 2.7.

brianhelba:
I'd like to clarify the use case of this downstream user. We've discovered some issues where Girder may not work well on wsgi anyway, so we should clarify our level of Beanstalk+wsgi support more directly.

zachmullen:
For the sake of argument, let's assume we get our house in order on WSGI deployments and that they will work great in the near future.

PR #2642, opened 2018-03-05
Add disable_event_daemon option to girder.dist.cfg

zachmullen:
This option wasn't really documented anywhere. We may want to include setting this somewhere in our deploy docs, but I'm not sure exactly where.


mgrauer:
How about in the beanstalk and wsgi sections, would that make sense @danlamanna ?

I was actually surprised to learn we had a wsgi deploy section--is anyone using this deployment type?

zachmullen:
I'd really like to just have the wsgi module disable the event daemon automatically. As part of that change I'd like to remove the construction of `girder.events.daemon` as an import side effect and instead make that a function that gets called from within `girder.utility.server.setup`. That way we wouldn't have to document the config file option or hack around with import order.

I'll try that on this topic, but if it ends up breaking too many assumptions we can back off that plan and just document this setting.

zachmullen:
@brianhelba @danlamanna PTAL at e52c477 in particular. I think it's certainly a net improvement, but it breaks some undocumented assumptions about `girder.events.daemon`.

brianhelba:
Since this option wasn't previously documented, I think it'd be safe to reconsider the name at this point. Rather than having some options that are 'negative' (e.g. `disable_event_daemon`) and others that are 'positive' (e.g. `cherrypy_server`), it might be simpler to instead form all options in a positive way. The defaults for each could still be anything; but it would make it easier to see which things were turned on / off based on their values.

zachmullen:
I don't have strong feelings either way. If we decide to rename, I'd like to do that on a subsequent PR to decouple the discussion.

girder/events.py
```diff
@@ -317,8 +316,12 @@ def trigger(eventName, info=None, pre=None, async=False, daemon=False):
...
+
 
-if config.getConfig()['server'].get('disable_event_daemon', False):
-    daemon = ForegroundEventsDaemon()
-else:
-    daemon = AsyncEventsThread()
+def setupDaemon():
+    global daemon
+    if config.getConfig()['server'].get('disable_event_daemon', False):
+        daemon = ForegroundEventsDaemon()
```
danlamanna: Since the synchronous event daemon is defined on import, should this just be a noop? Otherwise at run-time the import-time daemon is clobbered?

zachmullen: The idea here is for this method to be idempotent since it's doing stuff at server startup and affects stuff in module scope. We could do some type checking, but these are light objects and they'll be garbage collected, so I wasn't worried about the clobbering.

danlamanna: My concern about the clobbering is that any downstream binding to events at import time (rather than in a `load` function for instance) would have their bindings removed, right? (I don't think we should necessarily accomodate that, just checking).

zachmullen: No, that would be fine, the event bindings live in module scope under the `_mapping` global. The event daemons are stateless in that regard.

zachmullen:
@danlamanna I made one more change on a78b15b, PTAL.

PR #2655, opened 2018-03-09
Add support for one-time passwords (two-factor authentication)

### LINKED ISSUE #2724: autoDescribeRoute doesn't support required header parameters
The logic for `autoDescribeRoute` does not support required header parameters. Attempting to add a `.param` with both `paramType='header'` and `required=True` causes the route to always return a 400 error with the message `Parameter "..." is required.`.

brianhelba:
This now has basic functionality. Still TODO:
* [x] Allow OTP to be sent when logging into the SFTP server
* [x] Only prompt for an OTP on the login dialog if necessary
* [x] Refactor some logic out of the API layer and into the User model
* [x] Add `Girder-OTP` to default CORS headers
* [x] Server testing
* [x] Client testing
* [x] Resolve #2724
* [x] Add client confirmation / error message dialogs
* [x] Final styling pass
* [x] Resolve #2730

Additional work that may be out of scope for this initial PR:
* [ ] Rate limit login requests
* [ ] Switch caching of the last-used token to be stored in MongoDB ([see here](https://docs.openstack.org/keystone/pike/contributor/caching-layer.html) for possibly-helpful resources)
* [ ] Option to mandate use of 2FA by all users
* [ ] Allow user to recover account access with HOTP-based codes (currently, a site admin must disable a user's OTP to recover access)
* [ ] Use application secrets (once they exist) when generating TOTP keys
* [ ] Set OTP issuer when `User._TotpFactory` is created (requires more robust server hostname detection)

Technical reference: http://passlib.readthedocs.io/en/stable/narr/totp-tutorial.html

---

Current login dialog:
image

Before 2FA enrollment:
image

2FA enrollment process:
image

### DIFF
--- a/girder/models/user.py
+++ b/girder/models/user.py
@@ -20,12 +20,15 @@
 import datetime
 import os
 import re
+from passlib.totp import TOTP, TokenError
 
 from .model_base import AccessControlledModel
+from .setting import Setting
 from girder import events
 from girder.constants import AccessType, CoreEventHandler, SettingKey, TokenScope
 from girder.exceptions import AccessException, ValidationException
 from girder.utility import config, mail_utils
+from girder.utility._cache import cache
 
 
 class User(AccessControlledModel):
@@ -52,6 +55,13 @@ class User(AccessControlledModel):
         self.exposeFields(level=AccessType.ADMIN, fields=(
             'size', 'email', 'groups', 'groupInvites', 'status',
             'emailVerified'))
+
+        # To ensure compatibility with authenticator apps, other defaults shouldn't be changed
+        self._TotpFactory = TOTP.using(
+            issuer=Setting().get(SettingKey.BRAND_NAME),
+            # TODO: Add an application secret here
+            wallet=None
+        )
 
         events.bind('model.user.save.created',
                     CoreEventHandler.USER_SELF_ACCESS, self._grantSelfAccess)
@@ -127,7 +137,7 @@ class User(AccessControlledModel):
 
         return doc
 
-    def authenticate(self, login, password):
+    def authenticate(self, login, password, otpToken=None):
         """
         Validate a user login via username and password. If authentication fails,
         a ``AccessException`` is raised.
@@ -158,6 +168,11 @@ class User(AccessControlledModel):
 
         if not Password().authenticate(user, password):
             raise AccessException('Login failed.')
+
+        if self.hasOtp(user):
+            self.verifyOtp(user, otpToken)
+        elif otpToken:
+            raise AccessException('The user has not enabled one-time passwords.')
 
         # This has the same behavior as User.canLogin, but returns more
         # detailed error messages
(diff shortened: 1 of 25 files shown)


mgrauer:
@jtomeck Screenshots please? Whenever you are done with the current round of work.

jtomeck:
I worked on styling the UI of the one time password widget. Here is a screenshot of my approach.

screencapture-localhost-9080-2018-05-11-16_39_03

jeffbaumes:
I like the styling @jtomeck! Peanut gallery item: Should the bottom Cancel/Enable buttons be together (say both on the left) with the same size / font size, not unlike the "Close and comment" and Comment buttons on GitHub comments?

image

danlamanna:
More peanut gallery items:
- Will the bottom brown be affected by the brand color settings, and is that good/bad? (Not sure how that's implemented)
- Can we grep this whole branch and ensure that "Two-Factor" is punctuated as such everywhere? I saw the original PR had both "2-factor" and "two-factor".

danlamanna:
Also, most MFA implementations I've used have a modal dialog for username/password and then a separate dialog for the second factor. Is this a workflow we should consider?

jtomeck:
@jeffbaumes I think that's a good idea. I had a thought process behind why I did it the way I did, but I don't think the result really does anything to help the user on their path through the UI.

@danlamanna Right now that bottom bar just has the the top bar hex value for its color. I don't think it would be good if that matched the brand color because it could hurt the visibility of the buttons in the bar. I think this raises the question of if we want to change the color of the bottom bar, as well as the other blue elements I've designed into the layout. Maybe there are some stylus color variables somewhere in the CSS that I can play around with.

girder/models/user.py
```diff
@@ -127,7 +138,16 @@ def validate(self, doc):
 
         return doc
 
-    def authenticate(self, login, password):
+    def filter(self, doc, user, additionalKeys=None):
```
brianhelba: @girder/developers I'm not sure if I like what's being done here, but it's the only way I see to created 'derived' values when a document is filtered.

girder/models/user.py
```diff
@@ -156,9 +179,33 @@ def authenticate(self, login, password):
         if user is None:
             raise AccessException('Login failed.')
 
+        # Validate OTP preconditions
+        if self.hasOtp(user):
+            if otpToken is None:
+                raise AccessException(
+                    'User authentication must include a one-time password '
```
brianhelba: @girder/developers There are security implications to what the `AccessException` error message is. Previously, regardless of whether a user was not found or a password validation failed, the message `'Login failed.'` was returned. After this change, if a user is found who has OTP enabled, but no OTP is sent, the longer message `'User authentication must include a one-time password (typically in the "Girder-OTP" header).'` is send with the 401 error. This 'leaks' two pieces of information:
1. A user of that login / email addresses exists
2. That user has OTP enabled
If we want to consider any of this information to be secret, we can make some trade-offs in the order of the the various checks to leak some different information about password correctness instead. Unfortunately, to support the feature where not every user is prompted to enter an OTP token, we need to leak at least some information about the user existing and having OTP enabled.

Feel free bring up this with me directly, and I'll explain in more detail.

brianhelba:
@girder/developers I've done a lot more work here. There are a few final elements to wrap up (as detailed in the TODO) of the PR description, but feel free to try it out now, as it should be working end-to-end.

girder/models/user.py
brianhelba: @zachmullen Let me know if you have any thoughts.

girder/models/user.py
```diff
@@ -53,6 +57,13 @@ def initialize(self):
             'size', 'email', 'groups', 'groupInvites', 'status',
             'emailVerified'))
 
+        # To ensure compatibility with authenticator apps, other defaults shouldn't be changed
+        self._TotpFactory = TOTP.using(
+            issuer=Setting().get(SettingKey.BRAND_NAME),
+            # TODO: Add an application secret here
```
danlamanna: Can you elaborate?

brianhelba:
@girder/developers Review is complete and this is ready to merge. However, we need to fix #2730 before testing will pass in CI.

mgrauer:
💕

PR #2656, opened 2018-03-09
Use new-style Python 2 classes.

manthey:
Code of the form `class <name>:` or `class <name>():` define old-style classes in Python 2.  Use new-style classes instead by using `class <name>(object):`.  In Python 3, this results in no change.

### DIFF
--- a/girder/__init__.py
+++ b/girder/__init__.py
@@ -78,7 +78,7 @@ class LogFormatter(logging.Formatter):
         return super(LogFormatter, self).format(record, *args, **kwargs)
 
 
-class StreamToLogger:
+class StreamToLogger(object):
     """
     Redirect a file-like stream to a logger.
     """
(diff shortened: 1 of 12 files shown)


zachmullen:
Is there a linter rule we can enable for this?

manthey:
Not with the linters we currently use.  There is a flake8 plugin that adds this ([hacking](https://pypi.python.org/pypi/hacking) - https://github.com/openstack-dev/hacking -- rule H238).  That plugin comes with a fair amount of stuff we probably don't want, though, so I expect that if we added it we'd turn off everything that doesn't pass automatically.

zachmullen:
No problem, was just curious.

manthey:
The hacking module is pinned to old versions of mccabe, flake8, and pyflakes, so we can't use it without other changes in any case.

PR #2659, opened 2018-03-13
[WIP] Always use virtual addressing style for AWS S3

mgrauer:
This will support buckets in other regions

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -605,6 +605,10 @@ def makeBotoConnectParams(accessKeyId, secret, service=None, region=None, inferC
         if not service.startswith('http://') and not service.startswith('https://'):
             service = 'https://' + service
         params['endpoint_url'] = service
+    else:
+        params['config'] = params['config'].merge(botocore.client.Config(s3={
+            'addressing_style': 'virtual'
+        }))
 
     return params
 


girder/utility/s3_assetstore_adapter.py
```diff
@@ -605,6 +605,10 @@ def makeBotoConnectParams(accessKeyId, secret, service=None, region=None, inferC
         if not service.startswith('http://') and not service.startswith('https://'):
             service = 'https://' + service
         params['endpoint_url'] = service
+    else:
+        params['config'] = params['config'].merge(botocore.client.Config(s3={
```
mgrauer: This didn't help in my case (which went into the `inferCredentials` block and was `region='us-west-1'`) , because the `else` block wasn't entered as `service` had a value of 's3.amazonaws.com'.

When I changed this by adding the virtual addressing style to the `if service` block hence

```
if service:
     if not service.startswith('http://') and not service.startswith('https://'):
         service = 'https://' + service
     params['endpoint_url'] = service
     params['config'] = params['config'].merge(botocore.client.Config(s3={
         'addressing_style': 'virtual'
     }))
else:
     params['config'] = params['config'].merge(botocore.client.Config(s3={
         'addressing_style': 'virtual'
     }))
```            
  

I'm not saying that is the right code change because I don't know that we always want to have 'addressing_style': 'virtual' inside the `if service` block, just saying it worked for my case.

jbeezley: It's possible this is related to the testing fix I made in #2632.

zachmullen: I'm pretty certain this is the result of a recent change in either upstream botocore/boto3 or a policy change in AWS S3. If it's botocore/boto3, it seems likely that it's the same thing that caused your issue in #2632.

zachmullen:
I think I found a solution that will work in general. Closing this, will reopen with a new commit.

zachmullen:
Superseded by #2660

PR #2660, opened 2018-03-13
Do not specify default service for S3 assetstore type

zachmullen:
@mgrauer this is the correct fix for the S3 issue you encountered.

Connecting to buckets in non-default regions requires either virtual addressing style i.e. `bucketname.s3.amazonaws.com` or using the region-specific endpoint with path-style addressing e.g. `s3-eu-west-1.amazonaws.com/bucketname`.

If you explicitly pass `s3.amazonaws.com` as the service, it overrides the intelligent behavior of botocore that would otherwise handle this for you and use the region-specific endpoint.

In your case, you could have also solved this problem by setting your service to `s3-us-west-1.amazonaws.com` in your configuration, but that would be redundant since we have a region field.

TLDR: we should never set service explicitly to `s3.amazonaws.com`. The service field of an assetstore should only be set if you're using something other than AWS S3.

### DIFF
--- a/devops/ansible/roles/girder/library/girder.py
+++ b/devops/ansible/roles/girder/library/girder.py
@@ -257,9 +257,10 @@ options:
 
                    service:
                        required: false
-                       default: s3.amazonaws.com
+                       default: ''
                        description:
                            - The S3 service host (for S3 type)
+                           - Only set this if you're not using AWS S3.
                            - This can be used to specify a protocol and port
                              -  use the form [http[s]://](host domain)[:(port)]
                            - Do not include the bucket name here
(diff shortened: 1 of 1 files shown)


zachmullen:
Reference: https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html

mgrauer:
Great, thanks for the fix. I will test it out. One question I have for you @danlamanna @brianhelba , is how does this fit in with our Terraform and Ansible playbooks/Ansible-Girder role?

In our Terraform playbooks (what is the right word?) we can specify the AWS region for a deployment. How should that get pushed down through Ansible into the Assetstore setup?

Maybe I'm conflating separate issues though.

mgrauer:
Disregard that comment about Ansible, there is already a means to pass in the region to an S3 assetstore in the Girder Ansible client, I just needed to use that.

@zachmullen This worked for me. Have you tried it in the default AWS region? If not, I'd like to test that before we merge, but I won't have time to do it until tomorrow.

zachmullen:
Not with ansible, but I've tried it with Girder, and setting the service to the empty string when using the default region works there.

PR #2661, opened 2018-03-13
Better clean up after a fuse mount error.

manthey:
When we fail to make a fuse mount, don't leave the entry in the mount list.

### DIFF
--- a/plugins/fuse/server/server_fuse.py
+++ b/plugins/fuse/server/server_fuse.py
@@ -410,10 +410,14 @@ def unmountServerFuse(name):
 
 
 class FUSELogError(fuse.FUSE):
-    def __init__(self, name, operations, mountpoint, *args, **kwargs):
+    def __init__(self, name, onError, operations, mountpoint, *args, **kwargs):
         """
         This wraps fuse.FUSE so that errors are logged rather than raising a
         RuntimeError exception.
+
+        :param name: key for the mount point.
+        :param onError: a function that is called with `name` if initialization
+            fails.
         """
         try:
             super(FUSELogError, self).__init__(operations, mountpoint, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


plugins/fuse/server/server_fuse.py
```diff
@@ -121,7 +121,8 @@ def _getPath(self, path):
         except ValidationException:
             raise fuse.FuseOSError(errno.EROFS)
         except Exception:
-            self.log.exception('ServerFuse server internal error')
+            if self.log:
+                self.log.exception('ServerFuse server internal error')
```
zachmullen: Out of scope of the PR, but would it make more sense to assume `log` is always a logger? Seems easier to enforce that and control logging through configuration rather than always having to use `if self.log`.

plugins/fuse/server/server_fuse.py
```diff
@@ -422,6 +423,7 @@ def __init__(self, operations, mountpoint, *args, **kwargs):
                 'it empty?  Does the user have permission to create FUSE '
                 'mounts?  It could be another FUSE mount issue, too.' % (
                     mountpoint, ))
+            _fuseMounts.pop(name, None)
```
zachmullen: This seems like an odd place to mutate global state. Is there a way we could do this in the outer thread, maybe by passing an `onError` callback?

plugins/fuse/server/server_fuse.py
manthey: Yes.  This change was for consistency with the rest of the logging -- a future PR could just remove these clauses and add a proviso that `self.log` must be a logger.

plugins/fuse/server/server_fuse.py
manthey: I suppose, ideally, this would trigger an async event so that the error handler could also join the thread.  This would involve having the FUSELogError class do an event.daemon.trigger.  The plugin would need to bind to the event (at which point it could just call `unmountGirderFuse` , possibly with an extra flag to not bother calling umount).

manthey:
@zachmullen I mentioned in a now outdated comment that perhaps this would be better handled with asynchronous events.  In practice, it probably doesn't matter much, as if the setup of a fuse mount is misconfigured, whoever is starting Girder will probably fix it and restart.

manthey:
@zachmullen I rebased this on master to get the changes from #2663.  Can you reapprove, please?

plugins/fuse/server/server_fuse.py
```diff
@@ -422,6 +427,18 @@ def __init__(self, operations, mountpoint, *args, **kwargs):
                 'it empty?  Does the user have permission to create FUSE '
                 'mounts?  It could be another FUSE mount issue, too.' % (
                     mountpoint, ))
+            onError(name)
```
zachmullen: It probably makes sense to also pass the Exception to the error handler

manthey: The Fuse exception is particularly unhandy (just a generic RuntimeError).  While we can pass the exception to the handler, it won't actually be used.

zachmullen: I was thinking of API stability, but I suppose it doesn't matter for this plugin. We can add it later if we want.

PR #2663, opened 2018-03-14
Restrict python-dateutil to <2.7 for botocore 1.9.8

jbeezley:
The dev requirement botocore=1.9.8 restricts the version of
python-dateutil to <2.7 to maintain support for python 2.6[1].  This is
breaking how we do development installs on CI because pip doesn't fully
resolve package dependencies.  I also had to add the restriction to
setup.py because we use the `-U` flag in the install command in
`tox.ini`.  It's ugly, but hopefully will go away with future updates to
botocore.

An alternative solution would be to pin botocore to 1.9.7 (as well as
boto which requires botocore>=1.9.8) because we don't actually care
about python 2.6 support.  I chose to do this because it is more likely
that botocore will make important changes.

[1] https://github.com/boto/botocore/pull/1402

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -10,5 +10,6 @@ moto[server]
 pytest
 pytest-cov
 pytest-xdist
+python-dateutil<2.7 # required for botocore=1.9.8
 tox
 virtualenv
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ installReqs = [
     'pymongo>=3.5',
     'PyYAML',
     'psutil',
-    'python-dateutil',
+    'python-dateutil<2.7',  # required for compatibility with botocore=1.9.8
     'pytz',
     'requests',
     'shutilwhich ; python_version < \'3\'',


requirements-dev.txt
```diff
@@ -10,5 +10,6 @@ moto[server]
 pytest
 pytest-cov
 pytest-xdist
+python-dateutil<2.7 # required for botocore=1.9.8
```
zachmullen: I'm trying to figure out why it's also necessary to add this under requirements-dev. Does one of our other dev requirements upgrade python-dateutil?

jbeezley: It might not matter.  It depends on the order in which people install things.  For example, it is necessary (I think, might need verification) if someone does the following:
```
pip install -U -e .
pip install -U -r requirements-dev.txt
```

In any case, the *correct* place for the pinning is in the dev requirements because girder itself doesn't even depend on boto.  It just happens tox does the installation in the opposite order.

zachmullen: Girder does depend on botocore, but indirectly. It's a requirement of boto3.

jbeezley: Right.  It's an indirect dependency, which is primarily the reason why pip isn't correctly resolving the `python-dateutil<2.7` restriction correctly.

zachmullen: Ok, I was just a bit confused by having a non-dev requirement listed in requirements-dev, but I just wasn't thinking about the way we use requirements-dev.txt. It actually does make sense.

PR #2669, opened 2018-03-19
Run pip install in a subprocess

jbeezley:
This fixes errors running girder in certain environments that arose from the release of pip 9.0.2.  Running pip install in a subprocess is the method recommended by the pip maintainers, and the only way that will be supported post 10.0.

The `server_fuse.server_fuse` test is failing on this branch.  I still haven't been able to track down why.


jbeezley:
@manthey and I tacked on two other commits that solve other unrelated issues causing random failures.  This should be good to go now.

PR #2676, opened 2018-03-23
Adding a method that gets metadata.

AaronV77:
This PR is for adding a method to the Python Client for retrieving metadata. After looking through the source code of the python client for some time, I think getting the metadata of files, items, folders, and collections would be useful. There are options to add metadata but not getting it.


zachmullen:
@AaronV77 in the interest of not proliferating too many methods in the client API, and since this can be done in one line already, I don't think we will add this to the client at this time. Thank you for your contribution, though.

PR #2677, opened 2018-03-26
Bind mount node_modules in vagrant by default

### LINKED ISSUE #2569: Vagrant shared folders cause problems with certain npm packages
https://github.com/girder/girder/pull/2545#issuecomment-356754342 revealed a general problem with how certain npm packages can cause issues when `node_modules` is shared through Vagrant. 

@brianhelba can you link to the npm issue(s) you mentioned about this?

jbeezley:
Using a shared node_modules directory is unreliable.  This sets the
default behavior to bind mount node_modules from outside the shared
directory for all hosts, not just windows.

Fixes #2569

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -12,7 +12,7 @@ Vagrant.configure("2") do |config|
   ansible_example_name = ENV.fetch("GIRDER_EXAMPLE", "girder-dev-environment")
   is_testing = true?(ENV.fetch("ANSIBLE_TESTING", false))
   is_client_testing = true?(ENV.fetch("ANSIBLE_CLIENT_TESTING", false))
-  bind_node_modules = true?(ENV.fetch("BIND_NODE_MODULES", Vagrant::Util::Platform.windows?))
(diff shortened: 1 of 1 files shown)


brianhelba:
I agree this is the best path now, but it's unfortunate that this functionally is not available to developers anymore.

Can you open / reopen a bug about this issue, to remind us to try and re-enable the shared `node_modules` if npm fixes its race conditions (or we switch to Yarn, etc.)?

jbeezley:
Well, you can still enable it if you want :P.  From my perusing of npm issues, it's just a fact that shared directories don't behave as they are supposed to.  It's really more of an issue with virtualbox rather than npm.

https://github.com/npm/npm/issues/992#issuecomment-1262556

jbeezley:
I would also say that sharing `node_modules` with the host is bad practice anyway.  Compiled modules are not compatible across OS's, and the structure of `node_modules` is dependent on the specific version of `npm` used.

PR #2678, opened 2018-03-26
Check girder worker status

### LINKED ISSUE #1977: Add "Test connection" ability to Jobs plugin
The user should be able to test the connection to their broker from within the configuration panel.

Pierre-Assemat:
fixes #1977 

Steps:

- [x] Server API
    - [x] Add endpoint to check the workers status
    - [x] Handle the case where the broker can't be accessed
- [x] Client
    - [x] Add `view worker status` button on the girder_worker settings page
    - [x] Create a new view to display the data
- [x] Testing   
    - [x] Add Server testing
    - [x] Add Client testing


danlamanna:
Could you expand on what your use case is? I've wanted to implement something similar (ping) for #1977 but never got around to it.

@kotfic may also want to chime in on how to approach exposing control commands via the REST API.

manthey:
@danlamanna I've asked for this, as anytime someone deploys HistomicsTK manually, they seem to fail to deploy girder_worker such that girder can communicate with it.  Ideally, I would like to know how many worker nodes are able to be reached from girder.  If we can learn any more status about them, all the better.

kotfic:
This is looking really great!  I'm wondering if this belongs "closer" to the Jobs page,  rather than in the on the settings page?

One place to look for frontend inspiration would be at [Flower's screenshots](http://flower.readthedocs.io/en/latest/screenshots.html) which is celery's "built-in" tool for tracking this kind of information.   I'm wondering if @jtomeck might get involved w.r.t a strong design for displaying this info? @jeffbaumes not sure what jared's schedule looks like?

manthey:
I agree with @kotfic, this is looking great but should have a way to reach it from the jobs page.

I agree, the ping result isn't useful.

I think it could also use a "refresh" button, so after I submit a job I can refresh the display and see the results again.  When it is waiting for results, there should be something that indicates you are waiting for results.  If you navigate away before the results are returned, the UI jumps back to the page when they arrive; we probably want to ignore the results if the user is no longer on the status page.

@kotfic when the task is "girder_worker.run()" (e.g., "girder_worker.run () 7b465521-db07-4c3e-ac1d-14687899ab8e"), is there a way to correlate that back to a specific job in the Girder database?

Pierre-Assemat:
I took into account your thoughts @manthey, and started to add some css following the idea in the link @kotfic give me ([Flower's screenshots](http://flower.readthedocs.io/en/latest/screenshots.html)).

---
* The page is loading:
screenshot-2018-4-4 girder 2
---
* Worker found, the refresh button is working:
screenshot-2018-4-4 girder
---
* No Worker found:
screenshot-2018-4-4 girder 1
---

The time to loading the workers status is around 5seconds. I'm not sure it enough to allow the user to leave the page. But we can make the UI go back to this view after loading is completed anyway.

I though to add a link/button on the worker name to display the list of current tasks, and maybe other info, like the queue...

Let me know if you have more thoughts.

manthey:
This looks nicer.

I liked seeing what tasks were running -- @kotfic says with the new style girder worker tasks we can easily have human-friendly titles on the tasks.  Perhaps we could list the names of the running tasks in another column.

Do we know how many tasks are queued but not running on any workers?  And, if so, can we list those somewhere?

Pierre-Assemat:
There is an API endpoint to get the list of active_queues but this is not what we expect, I didn't find any way to list the task queued but not running.

About 
> @kotfic says with the new style girder worker tasks we can easily have human-friendly titles on the tasks. Perhaps we could list the names of the running tasks in another column.

Do you mean the name of tasks? Because the name is already display by clicking on the worker row of interest.

kotfic:
I believe determining what tasks are queued requires popping them off the queue. I don't believe there is a non-intrusive way to determine what tasks are queued (at least with rabbitmq as a broker)

zachmullen:
True, although you can apparently hack together your own peek behavior if you want: https://stackoverflow.com/questions/4700292/using-rabbitmq-is-there-a-way-to-look-at-the-queue-contents-without-a-dequeue

manthey:
This is looking wonderful.

I think it might be worth adding a column showing the pool max-concurrency.  I believe that this is the number of tasks that could be processed at one time (so you could tell if the worker is at full capacity).

manthey:
Regarding listing the number of queued but unscheduled items, if I do 
```
celery.task.control.inspect().reserved()
```
it looks like it lists the queued tasks.

mgrauer:
@Pierre-Assemat Please work with @manthey to update the CHANGELOG file before this PR gets merged, it seems like a significant feature.

kotfic:
`celery.task.control.inspect().reserved()` actually tracks the number of tasks that have been pre-fetched by the worker.  In high throughput situations one of the bottlenecks that arises is the time it takes to request a new task from celery.  Celery allows for 'pre-fetching'  a group of tasks rather than requesting them one at a time.  In this case the tasks have been "reserved" by that worker,  but have not yet moved into the "running" state (i.e.  they are waiting to be dispatched by the main worker process to one of the concurrent pool processes)

plugins/worker/server/api/worker.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class Worker(Resource):
+    def __init__(self):
+        super(Worker, self).__init__()
+        self.resourceName = 'worker'
+        self.route('GET', ('status',), self.getWorkerStatus)
+
+    @autoDescribeRoute(
+        Description('Get workers status and tasks information.')
```
manthey: I think this would read better as `'Get worker status and task information.'`

plugins/worker/web_client/views/taskStatusView.js
```diff
@@ -0,0 +1,107 @@
...
+                'report': reportTmp,
+                'stats': statsTmp | 0,
+                'concurrency': concurrencyTmp | 0,
+                'ping': pingTmp,
+                'active': active[worker],
+                'reserved': reserved[worker]
+            });
+        });
+        if (!this.workers.length) {
+            this.errorMsg = 'No task informations.';
```
manthey: This should be `'No task information.'`.

manthey:
@Pierre-Assemat Thanks for all your work on this.

Pierre-Assemat:
Thank you @manthey! It was very interesting to work around worker and use celery!

PR #2680, opened 2018-03-28
Fix where testing.min.css is loaded from.

manthey:
Also, we don't generate girder_app.min.css, so don't try to load it.


manthey:
I notice that in our tests, we were asking to load css files that didn't exist or were located in a different location.  Besides those mentioned in this commit, if you import a plugin via `girderTest.importPlugin` and that plugin doesn't have css (such as the gravatar and provenance plugins), we will ask for a non existent css file.

We could add a flag to `importPlugin` to skip the css when it doesn't exist.  There might be a nicer alternative.

Askign for non-existent css files doesn't directly cause problems, but it makes debugging other problems more complicated.

PR #2681, opened 2018-03-30
Oauth provider

zachmullen:
@brianhelba this is ready for preliminary testing from your client. You'll need to use `POST /api/v1/oauth_client` to create your client, there is no GUI for that yet. Once you've got your client set up, send your users to URLs of the form:

    http://my.girder.com/#/oauth_provider/authorize?clientId=5abe7c9e37887976639ff4ab&redirect=http://my.client.server.com/callback&state=foo&scope=core.user_info.read

You should see a login screen if you're not logged into Girder, otherwise you'll go straight to a consent screen like the following:

screen shot 2018-03-30 at 3 33 45 pm


zachmullen:
When you get your code back from the user's authorization confirmation, you exchange it for a token using `POST oauth_client/token`, whose documentation should be self-explanatory.

screen shot 2018-03-30 at 3 35 23 pm

PR #2684, opened 2018-04-04
Improve exceptions when no token.

manthey:
Querying an endpoint that needs a token should throw an `AccessException` if there is no token.  Before, this would throw an `AttributeError` (and the rest call would return a 500 error).

You can see this behavior by calling the GET `/job/:id` endpoint without an auth token.

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -645,10 +645,10 @@ def ensureTokenScopes(token, scope):
     :type scope: `str or list of str`
     """
     tokenModel = Token()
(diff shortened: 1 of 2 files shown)


zachmullen:
Would it make sense to support `token=None` is `Token.hasScope`?

manthey:
Probably.  I will do so.

manthey:
Looking, `Token().hasScope()` handles the `token=None` case.   The real problem was with how the exception was raised.

manthey:
@zachmullen Updated.  Unless you want the error message to be different, this should be ready.

PR #2685, opened 2018-04-04
Garbage collect before unmounting fuse.

manthey:
If a file has been placed in the queue for garbage collection, it can still be open until python actually does the garbage collection.  If the file is on a fuse mount, fuse can't unmount the file system due to the open file.  By running `gc.collect()` prior to the unmount call, files that should be closed don't prevent the unmounting.

Prior to this change, there were some wait loops in the tests to try to clean up that still sometimes could fail due to the file being open and the mount not closing.

### DIFF
--- a/plugins/fuse/server/server_fuse.py
+++ b/plugins/fuse/server/server_fuse.py
@@ -1,6 +1,7 @@
 import atexit
 import errno
 import fuse
+import gc
 import os
 import shutil
 import six
@@ -388,6 +389,9 @@ def unmountServerFuse(name):
 
     :param name: a key within the list of known mounts.
     """
+    # Have python do a garbage collection.  This ensures any file handle that
+    # could be closed really is closed, and therefore won't block unmounting.
+    gc.collect()
     with _fuseMountsLock:
         entry = _fuseMounts.pop(name, None)
         if entry:
--- a/plugins/fuse/plugin_tests/server_fuse_test.py
+++ b/plugins/fuse/plugin_tests/server_fuse_test.py
@@ -524,6 +524,7 @@ class ServerFuseTestCase(base.TestCase):
         fh = op.open(self.publicFileName, os.O_RDONLY)
         self.assertTrue(isinstance(fh, int))
         self.assertIn(fh, op.openFiles)
+        op.release(self.publicFileName, fh)
         path = os.path.dirname(self.publicFileName)
         fh = op.open(path, os.O_RDONLY)
         self.assertTrue(isinstance(fh, int))
--- a/plugins/fuse/plugin_tests/server_fuse_test.py
+++ b/plugins/fuse/plugin_tests/server_fuse_test.py
@@ -30,14 +30,7 @@ def tearDownModule():
     curConfig = config.getConfig()
     tempdir = curConfig['server_fuse']['path']
     base.stopServer()
-    retries = 0
-    while retries < 50:
-        try:
-            os.rmdir(tempdir)
-            break
-        except OSError:
-            retries += 1
-            time.sleep(0.1)
+    os.rmdir(tempdir)
 
 
 class ServerFuseTestCase(base.TestCase):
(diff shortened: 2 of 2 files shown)


plugins/fuse/server/server_fuse.py
```diff
@@ -388,6 +389,9 @@ def unmountServerFuse(name):
 
     :param name: a key within the list of known mounts.
     """
+    # Have python do a garbage collection.  This ensures any file handle that
+    # could be closed really is closed, and therefore won't block unmounting.
+    gc.collect()
```
zachmullen: This seems pretty dirty, and relying on implementation details of the interpreter.

Where are the file handles leaking from?

zachmullen: Ah, I see. I'm going to read a bit more about `gc.collect` before approving.

zachmullen: Are you sure about using the context manager form of `open` not closing when the context manager exits? I would think that would make an explicit call to `f.close()` immediately.

manthey: No, I'm not completely sure, but any code that does `fh = open(<file>)` without an explicit close will have this situation, too.

I think it is better to have the `gc.collect` than to expect that all code with close file handles promptly.  As I am using the fuse plugin with large_image, I see this after the tile source cache is freed (the tile source destructors DO close the file handles, but those destructors are only called when gc'ed).

I do see one test in `testFunctionOpen` where there is a line `fh = op.open(...)` without a companion `op.release`, which is at least one source of the non-garbage collected file references.

zachmullen: Well, my concern is just that this call could be masking / working around bad practices in other parts of the code where we aren't tearing things down explicitly when we should be. If all it takes to actually address this problem is to close our file handles, I'd rather do that than rely on the garbage collector. If it's in some deep and unreachable part of a third party library, maybe this is an acceptable workaround.

manthey: My current concern is that in large_image, we cache tile sources, which could have a number of open file handles.  When the tile source is discarded from cache (either because the cache has been flushed or because it has been evicted), the tile source reference is abandoned by the generic caching mechanism.  When python GCs the tile source object, its destructor explicitly closes the file handle(s).  I suppose we could overload the `__del__` method of the [LRUCache](http://cachetools.readthedocs.io/en/latest/) that we use, but that is a less general solution than the `gc.collect` method (and perhaps just as hacky).

jbeezley: If I understand the use cases for this correctly, this is only something that will generally happen when shutting down the server.  In this case, I don't think it's too terrible.  We certainly wouldn't want to run explicit garbage collections frequently.

I will add though that this won't solve all problems with unmounting... A file handle could be stored in a global variable or in an object that can't be collected.  We can consider this a problem of the code using this plugin, but I feel like this issue is representative of a bigger problem with what is being done in the fuse plugin.  Mounting/unmounting is a system-level operation that really shouldn't be handled by a web server.  My fear is that as we move on to more esoteric use cases, these kinds of problems will continue to occur.

I think this PR is low impact enough that it's okay to go in, but I would like consider the idea of transitioning the management of the fuse mounts to the system daemon before we build too many assumptions on how it currently works.

zachmullen:
> If the __exit__ step of the built-in open context manager is not calling close

I actually think it does call close, or at least I haven't found any evidence yet that it doesn't. If someone knows better than me or has looked at the source code, please enlighten us.

> if it's important that file handles be closed when they are evicted from a cache, then it certainly makes sense to hook onto a cache delete operation to do so.

+1 for cleaning these up explicitly. The discussion about large_image's caching behavior is probably out of scope of this PR, but in general, I'm not sure it's a good idea to cache file descriptors since it makes the cache no longer able to be shared across systems. If we must do so, though, I'd like to make sure we clean them up appropriately via an event handler or something.

brianhelba:
@manthey
> No, I'm not completely sure, but any code that does fh = open(<file>) without an explicit close will have this situation, too.

So, assuming that the context manager indeed does not call `close` within `__exit__`:

> Much as with open(<file>) as fh: is a nice construct, using fh = open(<file>) and then letting it get closed by GC is not prohibited.

What utility is there to using a context manager at all with `open`? Python variable scoping is still function-level at it's most granular, so what's the point of a context manager over simple assignment of the file handle?

manthey:
The context manager does close the handle when done (I was mistaken in using it as an example).

manthey:
I'm closing this -- I'll open another PR with a less contentious way of making sure the file handles are all closed on unmount.

PR #2686, opened 2018-04-05
Report if we can't unmount fuse.

manthey:
If any Girder file is still open when unmount is called, unmount will fail.  Instead of quietly trying to end, report the error code so the unmounting caller can then choose what to do.

Release a file handle in one of the tests.  Prior to this change, there were some wait loops in the tests to try to clean up that still sometimes could fail due to the file being open and the mount not closing.

### DIFF
--- a/plugins/fuse/server/server_fuse.py
+++ b/plugins/fuse/server/server_fuse.py
@@ -387,17 +387,21 @@ def unmountServerFuse(name):
     mount.
 
     :param name: a key within the list of known mounts.
+    :return: None for success or an error number from the unmount process
     """
     with _fuseMountsLock:
-        entry = _fuseMounts.pop(name, None)
+        entry = _fuseMounts.get(name)
         if entry:
             events.trigger('server_fuse.unmount', {'name': name})
             path = entry['path']
             # Girder uses shutilwhich on Python < 3
             if shutil.which('fusermount'):
-                subprocess.call(['fusermount', '-u', os.path.realpath(path)])
+                result = subprocess.call(['fusermount', '-u', os.path.realpath(path)])
             else:
-                subprocess.call(['umount', os.path.realpath(path)])
+                result = subprocess.call(['umount', os.path.realpath(path)])
+            if result:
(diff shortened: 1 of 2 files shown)


plugins/fuse/server/server_fuse.py
```diff
@@ -387,17 +387,21 @@ def unmountServerFuse(name):
     mount.
 
     :param name: a key within the list of known mounts.
+    :return: None for success or an error number from the unmount process
```
brianhelba: Since a failure to unmount is an error condition, would it be better to raise an exception in that case? The calling function could always choose to do nothing, but it would force the caller to at least acknowledge that an error is occurring (as opposed to being able to just never examining the return value). Additionally, it removes a heterogeneously-valued return type.

This is really just a microexample of C-style error handling vs C++-style, and I think we should prefer the latter.

plugins/fuse/server/server_fuse.py
```diff
@@ -375,29 +375,37 @@ def destroy(self, path):
...
     with _fuseMountsLock:
-        entry = _fuseMounts.pop(name, None)
+        entry = _fuseMounts.get(name)
         if entry:
             events.trigger('server_fuse.unmount', {'name': name})
             path = entry['path']
             # Girder uses shutilwhich on Python < 3
             if shutil.which('fusermount'):
-                subprocess.call(['fusermount', '-u', os.path.realpath(path)])
+                result = subprocess.call(['fusermount', '-u', os.path.realpath(path)])
```
brianhelba: As a minor nit, [`subprocess.check_call`](https://docs.python.org/2/library/subprocess.html#subprocess.check_call) may be a slightly more straightforward way of doing this, since we'd just catch and re-raise its exception as another type, instead of having the "logic" of figuring out whether to raise one ourselves.

manthey: A valid point.  I expect that this will change when we switch to "girder mount" instead of a plugin, so I'll leave it for now.

Issue #2687, opened 2018-04-05
Move FUSE binding lifecycle management outside girder server process


zachmullen:
ping @girder/developers please outline the technical plan in comments.

manthey:
@aashish24 Doing this is a much better solution to using mapnik with FUSE than forking to process mapnik tiles.

manthey:
The bulk of the Girder fuse plugin becomes `girder/cli/mount.py`.  In think the remaining functionality in Girder should move to core (not in a plugin that requires enabling).  This functionality extends `File().getLocalFilePath()` and adds a `getFuseFilePath` method (currently accessed via the fuse plugin namespace).  This latter function I would propose adding to the `File()` model.

The actual creation of the mount command is easy -- handling the tests will take the bulk of the effort.

aashish24:
> @aashish24 Doing this is a much better solution to using mapnik with FUSE than forking to process mapnik tiles.

thanks, this looks like a cleaner approach.

manthey:
Resolved by #2691.

PR #2691, opened 2018-04-11
Move fuse from a plugin to a mount command.

manthey:
The converts the FUSE plugin to a `girder mount` command.

The FUSE plugin had some unused features for mounting as a specific user.  This didn't really have great utility, since any process that can access the Girder user's filesystem could access the full feature mount, too.  These features were removed.

The mount command creates separation between the FUSE process and the Girder server process, making it harder to hang the FUSE process (since killing the Girder server process will not affect the mount process).

For those who want a `mount -t girder` option, a comment at the end of the `girder/cli/mount.py` file explains how to create one.


manthey:
Resolves #2687.

brianhelba:
A lot of this functionally is now "leaking" outside of the plugin and into core. Can you provide some elaboration / rationale on this?

manthey:
I'm not sure I understand the question -- the plugin has been removed, so of course what is left is no longer in a plugin.  The rationale is that it separates the mount process from the server process.

There is a change to the File model so that, if such a mount exists, core girder can make use of it.  The key feature of having a FUSE mount is that we can access Girder files via a filesystem.  The File() method `getLocalFilePath` has to be able to know if there is a mount so we can return such a path.  This has to be in the running server code (not just the mount code), since it is the process that needs the path.

Further, there are times when having a "natural" path is needed over having just any filesystem path (discussed briefly in the docs).  Specifically, if you have two files `foo.bar` and `foo.baz`, some C libraries expect them to be side-by-each with predictable extensions, and not as widely disparate hashed names (often the library is only given one base name and finds the rest from there).

manthey:
Additionally, the mount process needs to communicate the mount point to the server process.  I have done this via a setting (since both the mount and server processes *must* have access to the database).  If you think there is a preferable way for this communication, please suggest it.

brianhelba:
Sorry, I thought some of the plugin code still remained, so my question was even more unclear. To rephrase: could / should this continue to be implemented as a plugin (which, in principle, ought to be able to expose new CLIs too) or is there a strong reason to put it in "core"?

manthey:
I thought, as a group, we thought it best to not be a plugin (of course maybe I misunderstood).

The code is already out of the server except for the changes to the File model.  Having code checked into the main girder repo is itself "core", so whether this was in a plugin or as a cli doesn't affect the server code in a substantive way.

As a plugin, conceptually there is an extra step that (I think) is surprising.  You would have to install the plugin to get the mount command, then enable the plugin for the mount command to do anything useful, and do the mount command itself.  If the plugin was installed but not enabled, the mount command would exist, expose a filesystem to non-Girder applications, but not be usable by Girder applications.  Or, I suppose, the mount command would refuse to work with the plugin disabled.

Do you think there a strong reason to have it as a plugin (which is in core) instead of a cli in core?  Is forcing the potential user to turn on a plugin of any utility?  Is the drag of checking if the mount exists when a local file path is requested for non-filessystem assetstores too high?

zachmullen:
The purpose plugins is to provide opt-in functionality. By virtue of this being a separate command, and having its third party package dependencies listed in an extras_require field, this achieves that, so I'm OK with it as written.

brianhelba:
> I thought, as a group, we thought it best to not be a plugin (of course maybe I misunderstood).

My understanding was that the key goal was to move it out of the server process and into a separate CLI. I originally assumed this world not change its plugin vs core status.

> As a plugin, conceptually there is an extra step that (I think) is surprising. You would have to install the plugin to get the mount command, then enable the plugin for the mount command to do anything useful, and do the mount command itself. If the plugin was installed but not enabled, the mount command would exist, expose a filesystem to non-Girder applications, but not be usable by Girder applications. Or, I suppose, the mount command would refuse to work with the plugin disabled.

Indeed, as with the previous behavior, you'd have to install / enable the plugin for it to work, but that could be seen as a feature which allows the optional functionality to stay totally encapsulated.

> Do you think there a strong reason to have it as a plugin (which is in core) instead of a cli in core? Is forcing the potential user to turn on a plugin of any utility? Is the drag of checking if the mount exists when a local file path is requested for non-filessystem assetstores too high?

I think the calculus of plugin vs core ought to be exactly the same here as before and with other plugins. Putting someone in a plugin allows for encapsulation of optionally functionality; the goal is to streamline the user / developer experience in all the cases where it's not being used. It also makes maintainability easier by placing inter-dependent code into an common subdirectory.

In this instance, my original expectation was that for this functionally (like virtual folders), it'd be safer and cleaner to encapsulate it within a plugin. Particularly since this pushes the boundaries of what a traditional web server does (which certainly isn't a bad thing!), it'd be lower risk to put it in a plugin, particularly if we can avoid being bound by API stability requirements in core.

I'm open to reshuffling functionality into / out of core (including in this case), but I was concerned that two orthogonal changes (CLI vs server process; core vs plugin) were being needlessly conflated in one PR.

brianhelba:
> The purpose of plugins is to provide opt-in functionality.

I'd argue we don't need a plugin system for this; it'd be way s simpler to do with feature flags and settings. My view is that the primary virtue (with respect to 'built-in' plugins) is encapsulation, increasing the simplicity and auditability of core, and providing a place for technically risky or immature features to incubate.

zachmullen:
Girder's plugin system as it exists today is dependent on the girder server process -- plugins get loaded during server initialization. Since this doesn't depend on the server, I'm having trouble seeing how it would fit in as a plugin. I could certainly see it as its own pip package, or at least its own module, but right now it seems an ill fit for a traditional Girder plugin.

brianhelba:
That's a reasonable point, I think I'm thinking more about keeping it in its own package, which for now, would sort of just be a plugin directory (but pip-installable plugins will fix this, naturally).

manthey:
I think it is unexpected to have double opt-in requirements.  The plugin version did, which I view as a negative (and thus, this is an improvement).  

@brianhelba Regarding two conflating changes: I think it would be much messier to make this as two PRs, one which moved the code to core integration, and then one that deintegrated from core to a CLI.  Or, one that has the messiness of being the first plugin CLI addition and then removing that to move it to the CLI directory.

We don't have the other CLI code in plugins.  For instance, the sftp CLI adds to the girder/api directory.  I think that making this a special one-off as the only plugin to be a CLI adds more development burden rather than less.

setup.py
```diff
@@ -112,6 +112,9 @@ def run(self, *arg, **kw):
 extrasReqs['sftp'] = [
     'paramiko',
 ]
+extrasReqs['mount'] = [
```
zachmullen: Does it make more sense to name this extra "fuse" instead of "mount"?

manthey: My thought was the dependency is for the mount command.  Just like we name the paramiko dependency sftp, since that it what it is used with.

zachmullen: Well, SFTP is the capability, paramiko is an implementation detail. I don't understand enough about FUSE to say whether this is appropriate -- could there ever be a "mount" context that doesn't involve "fuse"?

manthey: I was trying to match conceptually.  There are different python fuse libraries.  And, I suppose, one could write a kernel driver rather than use fuse at all.

zachmullen: Ok, fair enough. I'm fine with `mount`.

PR #2695, opened 2018-04-17
Add --upgrade-strategy eager to all uses of pip install -U

jbeezley:
The default behavior of `pip install -U` has changed as of version 10.
The default only-if-needed strategy only upgrades direct dependencies of
the packages being installed.  This is likely to break automated
upgrades such as ansible and CI deployments that use a previously
generated virtual environment.

Both upgrade strategies can lead to invalid virtual environment states because 
One could argue that the new behavior is superior because it makes
minimal changes to the environment; however, my guess is that the eager
strategy will cause less confusion and is identical to what pip had done
prior to version 10 with no flag.

See also:

https://github.com/pypa/pip/issues/5274

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
           command: pip install --upgrade pip setuptools virtualenv
       - run:
           name: Install Girder
-          command: pip install --upgrade --editable .[plugins,sftp] clients/python --requirement requirements-dev.txt
+          command: pip install --upgrade  --upgrade-strategy eager --editable .[plugins,sftp] clients/python --requirement requirements-dev.txt
           # Until https://github.com/pypa/pip/pull/4208 is available in pip 10, the install of
           # "requirements-dev.txt" must be run from the "girder" working directory
           working_directory: girder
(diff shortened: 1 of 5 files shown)


brianhelba:
Is including this flag with pip 9.x just silently ignored?

manthey:
pip 9.x honors the flag.  pip 8 fails if it is specified.

jbeezley:
The flag was added in 9, the default was changed in 10.

brianhelba:
I think this change should definitely go in right now.

@jbeezley @danlamanna 
However, perhaps we should reconsider our aggressive use of the `--upgrade` flag in multiple places, in light of the path that pip seems to be taking. In particular, we could:
* In CI, just cache `~/.cache/pip`, and reinstall the virtualenv every time. We'd still save on download time, but we'd need to start spending the full time to install all required Python packages.
* In Tox, yield to the fact that Tox's intended behavior is to cache virtualenvs once they're created, and only reinstall dependencies when strictly necessary (which is why Tox has a separate `deps` config section). We seem to be fighting this intention (and slowing down Tox) by including the `--upgrade` flag now. If Tox users (who we assume are more sophisticated developers of upstream Girder) need to ensure that newer dependencies are installed, they can run Tox with the `--recreate` flag.
* In Docker, we should probably be installing Girder into a virtualenv regardless, instead of system-wide. Since this virtualenv would be created on every Docker re-build, nothing actually needs to be upgraded.

PR #2696, opened 2018-04-20
Add a text index to the file model.

manthey:
Allow files to be searched via `resource/search`.


zachmullen:
It looks like your test is asserting this, but just to sanity check: this obeys access control, right?

manthey:
Yes, definitely.

PR #2697, opened 2018-04-20
Add retry logic to the girder-client cli

jbeezley:
The retry logic is added following the description in this blog post:

https://www.peterbe.com/plog/best-practice-with-retries-with-requests

In the first commit, I had to do some refactoring of how sessions are injected by `girder_client.cli`.  It turns out sessions were being closed before the request was performed, which in turn closed all attached adapters.  The changes in `girder_client.lib` ensure that all requests now go through the `sendRestRequest` method.

For testing, I'm only asserting the retry adapter is attached to the session.  I originally tried to assert that the retries actually occurred, but simulating server failures turned out to be too difficult.  httmock can't be used for this purpose because it also mocks out the lower level retry logic.

### DIFF
--- a/clients/python/girder_client/cli.py
+++ b/clients/python/girder_client/cli.py
@@ -18,6 +18,7 @@
 ###############################################################################
 import click
 import requests
+from requests.adapters import HTTPAdapter
 import sys
 import types
 from girder_client import GirderClient, __version__
@@ -30,7 +31,8 @@ class GirderCli(GirderClient):
     """
 
     def __init__(self, username, password, host=None, port=None, apiRoot=None,
-                 scheme=None, apiUrl=None, apiKey=None, sslVerify=True, token=None):
+                 scheme=None, apiUrl=None, apiKey=None, sslVerify=True, token=None,
+                 retries=None):
         """
         Initialization function to create a GirderCli instance, will attempt
         to authenticate with the designated Girder instance. Aside from username, password,
@@ -80,6 +82,7 @@ class GirderCli(GirderClient):
         interactive = password is None
 
         self.sslVerify = sslVerify
+        self.retries = retries
 
         if token:
             self.setToken(token)
@@ -92,6 +95,8 @@ class GirderCli(GirderClient):
     def sendRestRequest(self, *args, **kwargs):
         with self.session() as session:
             session.verify = self.sslVerify
+            if self.retries:
+                session.mount(self.urlBase, HTTPAdapter(max_retries=self.retries))
             return super(GirderCli, self).sendRestRequest(*args, **kwargs)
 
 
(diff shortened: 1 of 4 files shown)


cjh1:
Could this retry logic be added to the ```GirderClient``` class rather than just the CLI? That way users of the client API would get it as well?

jbeezley:
I thought about that.  The problem is then there are endless options that go with configuring the retry logic and possible conflicts with adapters the consumer of the library might have added.  If a user of the library wants retry logic, this is minimal code to do it:
```python
from girder_client import GirderClient
from requests.packages.urllib3.util import Retry
from requests.adapters import HTTPAdapter

gc = GirderClient()
with gc.session() as session:
    session.mount(gc.urlBase, HTTPAdapter(max_retries=Retry.DEFAULT))

    # any rest calls made here will retry with the default options
``` 

It's really just one extra line of code, and you have the full power of urllib3 to customize and monitor the behavior.  My fear is if we try to wrap this behavior in girder client, then it will be difficult to do anything more advanced.

cjh1:
Fair enough, I can cut and paste.

clients/python/girder_client/cli.py
```diff
@@ -97,10 +97,14 @@ def sendRestRequest(self, *args, **kwargs):
...
             session.verify = self.sslVerify
             if self.retries:
-                session.mount(self.urlBase, HTTPAdapter(
-                    max_retries=Retry(total=self.retries, backoff_factor=1,
-                                      method_whitelist=False)
-                ))
+                # create a retry handler that:
+                #  1. uses exponential backoff
+                #  2. retries on all methods
+                #  3. retries on default status codes (503, 413, and 429)
```
zachmullen: I realize this is an upstream decision, but I'm curious about 413... is that something services typically send back due to high load?

jbeezley: The spec says this

> The HTTP 413 Payload Too Large response status code indicates that the request entity is larger than limits defined by server; the server might close the connection or return a Retry-After header field.

I don't know if urllib3 looks at the `Retry-After` header, but since girder doesn't use 413 I don't think it matters much.

zachmullen: From the docs about 413:

> If the condition is temporary, the server SHOULD generate a Retry-After header field to indicate that it is temporary and after what time the client MAY try again.

`Retry` has a parameter:

```
:param bool respect_retry_after_header:
 |      Whether to respect Retry-After header on status codes defined as
 |      :attr:`Retry.RETRY_AFTER_STATUS_CODES` or not.
```

Which defaults to `True`.

As such, is it prudent to specify `status_forcelist`?

danlamanna:
For future reference, setting up a proxy that would randomly inject failures was much easier than I thought, `npm install toxy` and run https://gist.github.com/danlamanna/0633256594c25788547ba8583ca09397.

Then `export HTTP_PROXY=http://localhost:3000` and run `girder-client`.

@jbeezley Mind adding a changelog entry?

cjh1:
Could toxy be used to test this retry logic? Or is it not worth it as you would really be testing requests/urllib3's retry logic?

danlamanna:
> Could toxy be used to test this retry logic? Or is it not worth it as you would really be testing requests/urllib3's retry logic?

I think if we wanted to test our instrumentation we could just raise a `ConnectionError` in one of our mocks, and assert that the CLI tries the same request again.

jbeezley:
I tried a number of ways to mock a connection error.  The problem is that mocking libraries (httmock, responses, etc.) bypass the actual request, so it never actually passes through urllib3.  The only way I can think of to do it is to mock it at the socket layer.  The amount of effort doesn't seem worth it given we'd really just be testing the behavior of `urllib3.util.Retry`.

zachmullen:
Sounds good. It's worth noting that while we mostly just have to consider Girder's HTTP behavior, we also have to consider that of any reverse proxy that sits in front of Girder that might inject its own behavior. As such, one other status I'd like us to consider is 408, which is something that a reverse proxy might send under certain reasonable configurations.

jbeezley:
I updated the branch.  I think it's most appropriate just to use the defaults for now.  We can add additional configuration if the need arises.

PR #2698, opened 2018-04-23
Fix the units shown on the job's time series graph.

manthey:
The job's time series graph didn't show units when the seconds were over 1000.  For instance, instead of showing '1.4k' as the timing history displays, it was just showing '1.4'.  Partly this was due to plotting the time in milliseconds.  This fixes the display.

Modified a line of code in the timing history vega specification for consistency.

Moved vega to version ^3.0.0 to be consistent with the candela plugin.


manthey:
Before this change, the timing history shows time like so:
image
But the time series showed it (note the highest value is `1.8` rather than `1.8k`):
image

Issue #2530, opened 2017-12-15
[RFC] Non-global server lifecycle support


zachmullen:
**Problem**: Girder was designed under the assumption that the lifecycle of the cherrypy/girder server was the same as the lifecycle of its python process. Now that we are using the server within pytest as a fixture, that assumption no longer holds.

**Requirement**: We need to be able to spin the server up and down over and over without leaking any information from the previous invocation into the global context, including when loading plugins.

**Current issues**:

* Any side effects that happen on import should go away. We can try to move the side effect to server startup time if it relates to the server.
* State that is stored in python module scope should either be centralized into a single object so that it can be reset easily, or moved somewhere besides global scope.
* We should make decisions about the lifecycle of `girder.events` and how that relates to the server lifecycle.

8bdfb27 is a workaround, but it's a spaghetti-wise solution where we just have to plug these holes as they appear. It would obviously be preferable to do a refactor so we can start to make guarantees about the state of the server based on a straightforward, documented API.

ping @girder/core @girder/developers

zachmullen:
I put this under the 3.0 milestone because I think it may involve breaking changes, but I'm not totally sure.

danlamanna:
> I put this under the 3.0 milestone because I think it may involve breaking changes, but I'm not totally sure.

The removal of starting an event daemon on import of `girder.events` is a breaking change in itself.

danlamanna:
I have a local branch that addresses this issue, I'll push it for feedback when it's functioning correctly.

zachmullen:
awesome!

danlamanna:
This has been removed from the 3.0 milestone but I was asked to reference my PoC branch with a little documentation.

The example branch for switching search mode registries to such a system is here: https://github.com/girder/girder/compare/app-lifecycle-management.

That branch uses Flask application contexts to allow multiple application objects to exist at the same time. Flask documentation describes the how and why quite well in their documentation:
- http://flask.pocoo.org/docs/1.0/appcontext/
- http://flask.pocoo.org/docs/1.0/patterns/appfactories/

The individual commits in that branch demonstrate converting a single global registry to an application context compliant one, and adds tests which are able to create their own application with isolated registries.

Motivating example of the problem:

events.py
```python
      event_bindings = {}

      def add_event_binding(key, fn):
	  if key in event_bindings:
	      event_bindings.append(fn)
	  else:
	      event_bindings[key] = [fn]

      def get_event_bindings(key):
	  return event_bindings.get(key, [])
[...]

PR #2701, opened 2018-05-01
Ensure events are unbound from server fixture

alesgenova:


cjh1:
Looks like a lot of test are failing. It looks like this approach is removing all the binding used internally as well. We may just wanted to add some cleanup code to our test case.

cjh1:
@girder/developers Any thought on this one. Basically if you write tests for a plugin that binds to any events, each time the server fixture is used the event handler get bound again causing interesting results! It would be nice if the server fixture cleaned out its event map. Other than making a copy of the internal map ```events_mapping``` and restoring it in the fixture clean up I'm not sure how this can be achieved. Or may be we could make bind the same handler a no-op?

danlamanna:
I talked with @cjh1 about this since it's another instance of #2530. Later this week I'll clean up/push my branch which deals with global state in Girder more generally - though that branch will likely require a good bit of discussion and review.

jeffbaumes:
> I'll clean up/push my branch which deals with global state in Girder more generally

@danlamanna is there more code you plan on pushing here?

PR #2702, opened 2018-05-02
Audit trail

zachmullen:
@girder/developers please comment on the architecture so far. If we are OK with it, I'll add tests.


danlamanna:
I can't remember if we discussed this, but should this be a plugin? Auditing seems to be a special case of something we might not want to let people toggle on and off.

zachmullen:
I think logging should be configurable. Saving extremely detailed transaction records might not be a feature everyone wants, since it can grow the database quite quickly.

zachmullen:
Tests are complete, this is ready for review

test/audit_logs/test_audit_log.py
```diff
@@ -0,0 +1,94 @@
...
+
+@pytest.fixture
+def recordModel():
+    from girder.plugins.audit_logs import Record
+    yield Record()
+
+
+@pytest.fixture
+def resetLog():
+    yield auditLogger
```
danlamanna: Should we be mocking the actual logger object instead? Rather than our tests taking a `resetLog` they would take a `mockAuditLog` or something similar?

zachmullen: Depends what we want to assert. I was more focusing on asserting that our handler was working, your strategy would be for asserting that the log messages are being dispatched.

brianhelba: In principle, this test could be written as a unit-ish tests with multiple independent parts: one which:
  1. Mocked the audit logger
  2. Made some HTTP requests
  3. Asserted that entries were added to the mocked log
and another which:
  1. Mocked the audit logger
  2. Directly added some entries to the mocked log
  3. Asserted that the plugin handled these entries as expected

Note that without mocking, these two parts would not be parallelizable with Pytest, as both parts interacting with the common global logger could cause failures.

Of course, as the current test does not examine the log data structure directly, is more of an end-to-end test, not a unit test (which is fine). However, although the non-mocking fixture is fine under the current structure, I think it's important that (wherever possible) we use fixtures to mock, rather than mutate-as-a-side-effect. Firstly, it will make this fixture usable as-is if we ever want to do more direct unit testing of this code. Secondofly, people are very likely going to copy-paste these sort of fixures as examples when writing additional tests, so we should prevent potential anti-patterns from spreading.

zachmullen:
Thanks for the comments. No one has requested changes on this branch, is there anything that still needs to be done, or is it ready to merge?

brianhelba:
I'd like to see the fixture changed as discussed. Otherwise, I've looked at the rest of the PR, and I'd be happy to approve it after that.

test/audit_logs/test_audit_log.py
zachmullen: > Note that without mocking, these two parts would not be parallelizable with Pytest, as both parts interacting with the common global logger could cause failures.

In my experience, when pytest is run in parallel, it does not run multiple tests in parallel in the same python process. It forks off several subprocesses and schedules the tests onto them, but they are run serially within each process. Is there some multithreaded pytest execution mode I'm not aware of?

jcfr: > Is there some multithreaded pytest execution mode I'm not aware of?

That would be available after installing the concurrent plugin:

https://github.com/reverbc/pytest-concurrent

zachmullen:
Just to clarify, for each of the end to end tests that are currently written, you would like me to write two versions, one which mocks each side of the logger & handler? It kind of feels like we'd be working against the rationale of unit tests (either being faster, more simplistic, or cheaper to maintain than integration tests), but if people feel strongly about it I can make that change.

brianhelba:
I think having a single fixture should be fine, it's just that behavior should be to provide a mocking, rather than 'reset' the default handler.

@danlamanna What would you suggest here?

zachmullen:
@brianhelba I took a pass at your approach last week but ran into trouble -- it's not simple to use a pytest fixture to mock the global logging handler at the right time, i.e. during plugin initialization. Maybe someone knows of a way to work around that?

brianhelba:
I'm going to withdraw my objections for now, because I realize there are two harder problems here that I don't necessary know the 'right' answer to:
* what's the best practice for mocking Python logger during testing (in all Python apps, not just Girder)
* how do we address the fact that the act of loading plugins can have disruptive side effects on the state of Girder's Python process, which may not be cleaned up just by tearing down the `server` fixture?

I'd definitely appreciate community input, so I'll create an issue for this.

zachmullen:
Any other comments on this branch? Needs approval before it can be merged.

zachmullen:
The main thing missing from this is that it doesn't log creation of new documents and their ID. If we add that, it will be possible to trace the entire lifecycle of a document.

test/audit_logs/test_audit_log.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@pytest.fixture
+def recordModel():
+    from girder.plugins.audit_logs import Record
+    yield Record()
+
+
+@pytest.fixture
+def freshLog():
```
danlamanna: Rename to `funkyFreshLog` plz.

zachmullen:
@danlamanna could you please re-add review? I added a test for the last feature.

PR #2707, opened 2018-05-08
Upgrade node to version 8.x in ansible role

zachmullen:
Currently, girder-install web --all-plugins is breaking; some of our upstream packages are using newer ES features than node 6 supports.

### DIFF
--- a/.circleci/girder_test_py2/Dockerfile
+++ b/.circleci/girder_test_py2/Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER Kitware, Inc. <kitware@kitware.com>
 # Don't use "sudo"
 USER root
 
-# Install Node.js 6
+# Install Node.js 8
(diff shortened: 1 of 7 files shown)


zachmullen:
@fbudin69500 will review.

brianhelba:
Let's upgrade this across the board please, including CI and documentation.

zachmullen:
@brianhelba done.

brianhelba:
To be clear, if you'd like to push the Node 10 upgrades to a subsequent PR, I'd be fine with that, but I think it should be upgraded eventually.

fbudin69500:
I tested the Debian one locally and it solved my problem. Thanks.

zachmullen:
@brianhelba changes addressed, PTAL

brianhelba:
Note, I didn't test locally myself.

zachmullen:
Thanks for the reviews

PR #2708, opened 2018-05-09
Fix plugin webroots not working under WSGI deployment

danlamanna:
Prior to this, we were exposing the primary webroot as the WSGI
application. In the case of plugin webroots we mount multiple webroots
onto cherrypy.tree which itself is a WSGI-compliant application.

Steps to test:
1) `girder-install plugin -s test/test_plugins/has_webroot`
2) Run Girder via uWSGI
    `uwsgi -p4 --lazy-apps --http :8081 --module girder.wsgi --check-static clients/web` 
3) Enable the "has webroot plugin" and restart the uwsgi process.
4) Verify http://localhost:8081/has_webroot displays "some webroot" (prior to this PR it will 404)


danlamanna:
@zachmullen can you reapprove?

PR #2710, opened 2018-05-15
Fix Python client uploading existing files to item

LucasGandel:
As discussed on discourse.girder,
https://discourse.girder.org/t/uploading-dicom-files-to-item-how-to-bypass-the-limit/123
the python client fails on checking for already existing files because
of request return limit.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -836,8 +836,7 @@ class GirderClient(object):
         :param filename: name of file to look for under the parent item.
         :param filepath: path to file on disk.
         """
-        path = 'item/' + itemId + '/files'
-        itemFiles = self.get(path)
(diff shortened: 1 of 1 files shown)


LucasGandel:
@jbeezley Here is a small patch for the bug we found today. Hope that helps! Thanks again!

jbeezley:
@LucasGandel Thanks for the patch!  The failing tests are because our data server is flaky at the moment.  I'll restart the tests once we have it resolved.

LucasGandel:
@jbeezley Thanks for reviewing! Can we integrate this change?

mgrauer:
Sorry @LucasGandel , our normal procedure is to approve the change and have the patch submitter (you in this case) merge, but you didn't have merge access. I've merged it now.

PR #2712, opened 2018-05-17
Allow the notification stream to be disabled, based on setting

### LINKED ISSUE #2639: Add a config option to disable web client notifications
This is mainly to preserve server side resources, ideally this logic could have a small API surface area. 

Perhaps just some quick error handling and cleanup on the client side as well.

brianhelba:
Fixes #2639.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,9 @@ Changes
 Added Features
 --------------
 
+* Added a setting to disable the notification stream. This may improve Girder's performance in runtime environments with
+  fewer threads (`#2712 <https://github.com/girder/girder/pull/2712>`_).
+
 Python Client
 ^^^^^^^^^^^^^
 * Added a ``--token`` option to the girder-client command line interface to allow users to specify
--- a/clients/web/src/utilities/EventStream.js
+++ b/clients/web/src/utilities/EventStream.js
@@ -4,8 +4,6 @@ import Backbone from 'backbone';
 
 import { getApiRoot } from 'girder/rest';
 
-var timestamp = null;
-
 /**
  * The EventStream type wraps window.EventSource to listen to the unified
  * per-user event channel endpoint using the SSE protocol. When events are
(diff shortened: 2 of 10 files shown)


waxlamp:
`requestAnimationFrame()` is similar to `setTimeout()` in that it will execute in a fresh stack; used recursively as in Girder, it constitutes a sort of trampoline (and thus won't grow the stack).

brianhelba:
@zachmullen @waxlamp The call stack is not the same as the event loop, so infinite recursion is indeed happening here. Feel free to drop in a breakpoint and see for yourself.

It is true that each callback to `_timeout` executes on a new pass of the event loop. However, note that the call stack frame of an asynchronous call includes a closure with the entire scope of the asynchronously-called function's definition. In the case, the asynchronously-called function is:
```
() => { this._heartbeat(); }
```
and it is re-defined within the currently-running call to `EventStream._heartbeat`. Each additional time `_heartbeat` is called, it defines a brand-new lambda, which is invoked on top of the current call stack. Accordingly, the call stack for this sequence of calls continues to grow unbounded.

brianhelba:
@zachmullen Is there a particular reason that `window.requestAnimationFrame` is used to determine when the window goes out of focus, instead of the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API)?

zachmullen:
My guess is that at the time this feature was written, the page visibility API was not widely adopted. requestAnimationFrame was still state-of-the-art.

Seems like we should switch, though.

brianhelba:
@girder/developers PTAL now.

brianhelba:
Just to clarify, tests need to be fixed, because they were touching private aspects of `EventStream`. I'll fix them, but feel free to review the rest of the updates now.

brianhelba:
The test code in [`widgetsSpec.js`](https://github.com/girder/girder/blob/master/clients/web/test/spec/widgetsSpec.js#L58) is still failing, but I'm having a difficult time understanding the structure of that test.

@manthey I think you wrote it originally, can you provide any feedback?

manthey:
@brianhelba It is failing asked we ask the event stream to stop by setting the heartbeat timeout to 1 ms.  This was the timeout so that when a tab lost focus we would stop asking for the event stream -- it defaulted to 5 seconds, and in the test we made it shorter so as to not need to wait so long.

It looks like you've changed this to monitor the document visibility, so the tests have to change to follow suit.

manthey:
I'll note that the eventStream was originally written before there was widespread browser support of the document `visibilitychange` event.

brianhelba:
@manthey What I'm struggling with is the interaction of `_setProgress`-set server behavior with the client-side behavior under test. Maybe we could discuss it in-person.

danlamanna:
Server-side LGTM, can you add a changelog entry?

clients/web/src/utilities/EventStream.js
```diff
@@ -14,103 +12,164 @@ var timestamp = null;
...
+    let obj;
+    try {
+        obj = window.JSON.parse(e.data);
+    } catch (err) {
+        console.error('Invalid JSON from SSE stream: ' + e.data + ',' + err);
+        this.trigger('g:error', e);
+        return;
     }
+    EventStream.setLastTimestamp(obj._girderTime);
+    this.trigger('g:event.' + obj.type, obj);
```
zachmullen: Can we simplify this control flow to:

```
try {
    const obj = window.JSON.parse(e.data);
    EventStream.setLastTimestamp(obj._girderTime);
    this.trigger('g:event.' + obj.type, obj);
} catch (err) {
    console.error('Invalid JSON from SSE stream: ' + e.data + ',' + err);
    this.trigger('g:error', e);
}

Comment (brianhelba) [reply clients/web/src/utilities/EventStream.js]: I only want to catch JSON decoding errors, and Javascript doesn't have an `else` block for `try` statements (unlike Python).

Comment (zachmullen) [reply clients/web/src/utilities/EventStream.js]: Fair enough, if we anticipate exceptions from the other two lines.

File: girder/api/v1/notification.py [hunk 2]
```diff
@@ -71,8 +74,12 @@ def __init__(self):
         .produces('text/event-stream')
         .errorResponse()
         .errorResponse('You are not logged in.', 403)
+        .errorResponse('The notification stream is not enabled.', 503)
     )
     def stream(self, timeout, params):
+        if not Setting().get(SettingKey.ENABLE_NOTIFICATION_STREAM):
+            raise RestException('The notification stream is not enabled.', code=503)
```
zachmullen: While "service unavailable" does seem to fit the semantics of our response, I think using status code 503 might not be a good idea here. I believe 5xx status codes typically indicate that something is *wrong* with the server, whereas in this case, nothing is wrong, it's just intentionally configured to be disabled. I wasn't going to object until I read the description [on wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_errors) of 503 in particular:

> The server is currently unavailable (because it is overloaded or down for maintenance). Generally, this is a temporary state.

As written, this isn't a temporary state, and it's not due to being overloaded or due to maintenance. We could add features in the future to temporarily throttle or disable the notifications due to load, and in that case I think 503 would be appropriate, but in this case I feel like status 410 may be our best bet.

brianhelba: While a 500 indicates that something is probably wrong with the server, not all 5xx codes have the same semantics. From [RFC7231](https://tools.ietf.org/html/rfc7231#section-6.6):
> The 5xx (Server Error) class of status code indicates that the server is aware that it has erred or **is incapable of performing the requested method**.

For reference, here are definitions of the two codes in question:
* [410 Gone](https://httpstatusdogs.com/img/410.jpg)
> The 410 (Gone) status code indicates that access to the target resource is no longer available at the origin server and that this condition is likely to be permanent.  If the origin server does not know, or has no facility to determine, whether or not the condition is permanent, the status code 404 (Not Found) ought to be used instead.
>
> The 410 response is primarily intended to assist the task of web maintenance by notifying the recipient that the resource is intentionally unavailable and that the server owners desire that remote links to that resource be removed.  Such an event is common for limited-time, promotional services and for resources belonging to individuals no longer associated with the origin server's site. It is not necessary to mark all permanently unavailable resources as "gone" or to keep the mark for any length of time -- that is left to the discretion of the server owner.
>
> A 410 response is cacheable by default; i.e., unless otherwise indicated by the method definition or explicit cache controls.

* [503 Service Unavailable](https://httpstatusdogs.com/img/503.jpg)
> The 503 (Service Unavailable) status code indicates that the server is currently unable to handle the request due to a temporary overload or scheduled maintenance, which will likely be alleviated after some delay.  The server MAY send a Retry-After header field (Section 7.1.3) to suggest an appropriate amount of time for the client to wait before retrying the request.
>
> Note: The existence of the 503 status code does not imply that a server has to use it when becoming overloaded.  Some servers might simply refuse the connection.

As I understand it, the notification stream being disabled is:
* Caused by an action of the server, not the client
  * Accordingly, with a 503, the client was not 'at fault' for a perfectly reasonable, well-formed request, it's the server that's decided not to provide the resource.
  * To me, the implication of 410 Gone seems to be 'this resource isn't here, you shouldn't have asked for it, and you shouldn't ever ask for it again'.
* Not perminant
  * The decision to disable the notification stream is made by altering a runtime setting, and the functionality could be re-enabled at any time. Indeed, the actual policy of web clients is to attempt to reconnect to the notification stream every time the web app reloads.
  * While the server may send a `Retry-After` header with a 503, it is optional, and I see nothing wrong with a service being "temporarily" down, even if when or if it comes back up is uncertain, so long as the service *could* come back up (machines are very patient).
  * To restate my point from [our previous discussion about this](https://github.com/girder/girder/issues/2639):
    > In my understanding, "permanent" status codes (like 308 Permanent Redirect) are more for things where a policy is changed, and clients should ideally update themselves to never issue a request to the old route again.

zachmullen: I still think this is more of a policy change than a temporary setting -- I think temporary conditions come with the expectation that they'll be "fixed" at some point. That said, I don't want to hold up this branch due to bikeshedding, so we can merge it as-is.

brianhelba:
@zachmullen @danlamanna Added a CHANGELOG entry. PTAL and re-approve.

brianhelba:
@girder/developers Can any developer PTAL at the CHANGELOG entry and re-approve this PR? All other components have already been approved by @zachmullen and @danlamanna .

CHANGELOG.rst
```diff
@@ -17,6 +17,9 @@ Changes
 Added Features
 --------------
 
+* Added a setting to disable the notification stream. This may improve Girder's performance runtime environments with a
```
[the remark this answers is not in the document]
jcfr: `Added Features` -> `New Features`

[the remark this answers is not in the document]
mgrauer: Sorry, I may have been unclear.

I meant that in a section called `Added Features` I don't like saying `Added a setting` but prefer `Setting to disable`.

Again, this is not something that anyone should bother paying attention to as it is really unimportant.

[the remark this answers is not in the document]
brianhelba: @mgrauer Apparently I forgot how to use prepositions, so I'll fix the missing *to*.

However, since a number of threads is a countable quantity, not a matter of size or degree, the adjective *fewer* should be used: http://www.gmatpill.com/more-than-vs-greater-than-vs-less-than-fewer-than/

Thanks for the feedback. For many developers, the CHANGELOG is the primary feed of Girder updates, so I'm happy to spend time making it polished and easy to follow.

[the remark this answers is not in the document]
mgrauer: Right, it would be `fewer threads` because threads are countable [1] but you said `a fewer number of threads`. It should be either a `smaller number of threads` or `fewer threads`.

And it turns out I may be wrong about `smaller number`, it could end up being `lower number` or `lesser number`, see [here](https://english.stackexchange.com/questions/26471/lower-number-vs-smaller-number). As English is descriptive rather than prescriptive, perhaps that blue line in the top chart wins the day and it is `smaller number` after all? Maybe `fewer threads` is best as it sidesteps this sticky wicket.

🥀 

How's that for burnishing the bikeshed!

[1] good thing we are not using greenlets which are uncountable [2]
[2] joking

brianhelba:
Fixed.

PR #2714, opened 2018-05-18
Allow uploading of empty files in python client

zachmullen:
@msmolens discovered an ancient bug in girder_client that disallowed uploading of an empty file, which is supported by the server.


msmolens:
@zachmullen Can you add a changelog entry for this bug fix?

PR #2715, opened 2018-05-21
Upgrade to Vega3

waxlamp:
This PR upgrades Girder's Vega dependency from 2.x to 3.x. This includes the following updates:
- the `vega` plugin now works with Vega 3 specifications
- the `jobs` plugin uses Vega 3 for its two visualizations: a bar chart and a line chart
  - a bug in the line chart, where the wrong time was shown on hover, has been fixed

TODO:
- [ ] verify that Girder no longer depends on Vega 2
- [ ] add a test for the `vega` plugin update

@manthey PTAL.

### DIFF
--- a/plugins/jobs/web_client/views/JobGraphWidget.js
+++ b/plugins/jobs/web_client/views/JobGraphWidget.js
@@ -1,7 +1,7 @@
 import _ from 'underscore';
 import $ from 'jquery';
 import { parse,
-         View as VegaView } from 'vega';
+    View as VegaView } from 'vega';
 import moment from 'moment';
 
 import View from 'girder/views/View';
@@ -74,7 +74,7 @@ const JobGraphWidget = View.extend({
             config.height = this.$('.g-jobs-graph').height();
             config.data[0].values = vegaData;
 
-            const minval = Math.min(0, Math.min.apply(this, vegaData.map(d => d.elapsed === undefined ? 10 : d.elapsed)) / 1000);
+            const minval = Math.min(0, Math.min.apply(this, vegaData.map((d) => d.elapsed === undefined ? 10 : d.elapsed)) / 1000);
             config.data[1].values = [minval < -86400 ? -86400 : minval];
 
             config.scales[1].type = this.yScale;
--- a/plugins/jobs/web_client/views/JobGraphWidget.js
+++ b/plugins/jobs/web_client/views/JobGraphWidget.js
@@ -1,6 +1,7 @@
 import _ from 'underscore';
 import $ from 'jquery';
-import vg from 'vega';
+import { parse,
+         View as VegaView } from 'vega';
 import moment from 'moment';
 
 import View from 'girder/views/View';
@@ -67,7 +68,7 @@ const JobGraphWidget = View.extend({
             let withForEachJob = width / numberOfJobs;
             // if the width for each job is less than 20px, remove axe labels
             if (withForEachJob < 20) {
-                config.axes[0].properties.labels.opacity = { value: 0 };
+                config.axes[0].encode.labels.opacity = { value: 0 };
             }
             config.width = width;
             config.height = this.$('.g-jobs-graph').height();
@@ -79,13 +80,13 @@ const JobGraphWidget = View.extend({
             config.scales[3].domain = this.collection.pluck('_id');
             config.scales[3].range = this.collection.pluck('title');
 
-            vg.parse.spec(config, (chart) => {
-                var view = chart({
-                    el: this.$('.g-jobs-graph').get(0),
-                    renderer: 'svg'
-                }).update();
-                view.on('click', openDetailView(view));
-            });
+            const runtime = parse(config);
+            const view = new VegaView(runtime)
+                .initialize(document.querySelector('.g-jobs-graph'))
+                .renderer('svg')
+                .hover()
+                .run();
+            view.addEventListener('click', openDetailView(view));
 
             let positiveTimings = _.clone(this.timingFilter);
             this.timingFilterWidget.setItems(positiveTimings);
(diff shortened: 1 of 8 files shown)


manthey:
If I turned off all phases but one, sometimes the graph doesn't appear.

waxlamp:
Let's hold off on the Candela change for a future PR.

@manthey could you give the phase/graph behavior another try?

manthey:
There is a minor style difference between the existing legend and this branch.  The shape of the legend dots isn't important (though I like them not having a border around each dot), but there should be a little more space between the rightmost data bar and the legend:
Existing branch:
image
Current PR:
image

manthey:
One strangeness (which this PR did not introduce and hasn't changed), is that if you uncheck the phase that corresponds to a job's final status (typically "Success", "Error", or "Canceled"), the marker above the job name and below the chart turns to the "inactive" color.  I would think that either the marker would vanish or it would be some color that isn't represented by any entry in the legend.

waxlamp:
I'll work on spacing the legend today.

As for the state indicator turning to "inactive" when the phase is disabled -- I actually put a hack into this PR to recover that behavior to mimic what the old chart did :). I agree that it shouldn't simply change to gray, but I didn't like it simply vanishing either. However, I don't have strong opinions on this. Here are the three options I would choose from to move forward, in order of effort required:

1. Leave it as is, and let a future PR change the behavior (since this PR was only meant to upgrade to Vega 3, rather than fix any problems with the chart).
2. Remove the hack so that the indicator just vanishes.
3. Change the "off" color to black or a very dark gray.

Option 3 seems pretty simple, but my hack was to place inactive-colored boxes under the status indicators, so that when they do disappear, it looks like they have turned gray. The problem is that when you roll over those boxes currently, their opacity drops to 0.5. This actually causes a bit of a difference from the old chart, since now you will see a 0.5 opaque status color mixed with the backing inactive-gray. Changing it to black will further change the color seen when rolling over. It's probably not a big deal but I wanted to make you aware of the mechanism.

@manthey let me know what you think.

manthey:
For now, lets go with option 1 (let a future PR change this).  I like option 3 just fine (but it should be a future PR).

manthey:
@waxlamp Can you switch the dependencies in the vega and job plugins from `vega` to `vega-lib`?

manthey:
I'm not seeing the tooltips that should appear on the data bars (I see them on the status marks).

manthey:
For the time series graph, is using seconds rather than milliseconds, the displayed time always has the form of xx.000s or xx:xx.000s (the milliseconds are always zero).  Sub-second resolution could be useful, so we probably need to put this back in the `datetime` millisecond field (but if for some reason we don't, then we should remove the `.%L` part of the format string).

As a separate PR, I'd love to see the format drop the leading zeros and remove the `s` suffix when minutes and hours are included.

manthey:
And, so it is recorded, if the Table View plugin is enabled, its use of the datalib package can cause issues with vega-3 due to a d3 time format conflict.

manthey:
@waxlamp I don't know if we care, but switching to Vega 3 (without adding a polyfill) breaks Girder's support for Internet Explorer 11.  Vega has polyfills to support IE 11.

waxlamp:
@zachmullen is it ok if we break support for IE11? Do we have a general policy on IE support?

zachmullen:
I read that as well, but I'm wondering more about their de facto support for new features that enter the standard -- it feels like they've deprecated IE altogether and new features are really only being added to Edge. We can discuss this at a future planning meeting, but I would be more comfortable supporting Edge and officially stating that we do not support IE.

manthey:
I agree that new features are not being added to IE.  The switch to vega3 (without adding a polyfill) is the first breaking change that prevents a core plugin from working on IE (technically already done by the candela plugin).

waxlamp:
@manthey, I think I have fixed the time formatting problem from [your comment above](https://github.com/girder/girder/pull/2715#issuecomment-391711109); PTAL.

I've also opened #2726 and #2727 to capture future TODOs from this thread.

manthey:
On the time series graph, the data points are not above their labels.  This branch:
image
Master:
image

waxlamp:
@manthey, should be fixed now.

waxlamp:
I need some help understanding this test failure: https://circleci.com/gh/girder/girder/10940. I am not sure how the last commit I made could possibly have caused a new test failure in the "collection terms" test.

manthey:
The terms test is flaky -- rerunning it is likely to make it pass.

plugins/jobs/web_client/views/JobGraphWidget.js
```diff
@@ -67,25 +68,29 @@ const JobGraphWidget = View.extend({
             let withForEachJob = width / numberOfJobs;
             // if the width for each job is less than 20px, remove axe labels
             if (withForEachJob < 20) {
```
manthey: I realize this isn't part of the change on this PR, but can you rename `withForEachJob` to `widthForEachJob`?

waxlamp:
Thanks for the help with that test, @manthey. I made the change you requested; hopefully the tests all pass this time without a hitch :)

plugins/vega/web_client/views/VegaWidget.js
```diff
@@ -29,12 +30,11 @@ var VegaWidget = View.extend({
...
                 .done(function (spec) {
-                    vg.parse.spec(spec, function (chart) {
-                        chart({
-                            el: '.g-item-vega-vis',
-                            renderer: 'svg'
-                        }).update();
-                    });
+                    let runtime = parse(spec);
+                    let view = new VegaView(runtime)
+                        .initialize(document.querySelector('.g-item-vega-vis'))
```
brianhelba: @waxlamp @manthey Just pointing out that we should be using jQuery (ideally, Backbone-view-instance-scoped jQuery) to do DOM selection. Using `document.querySelector` is inconsistent with Girder / Backbone style. If you need a raw HTMLElement, you can always use `[0]` on the jQuery selection.

There are a number of places in this PR with this issue. Ideally, these should be cleaned up to discourage reuse.

PR #2720, opened 2018-05-24
Allow setting admin=False on users

### LINKED ISSUE #2719: User API no longer supports revoking site admin privileges
The `PUT /user/{id}` endpoint, which updates a user's information, no longer supports revoking site admin privileges by setting the `admin` parameter to `False`:
https://github.com/girder/girder/blob/f53e907c653e4a4c59b5dc6850401d4c978987c9/girder/api/v1/user.py#L238-L242

This behavior changed in https://github.com/girder/girder/commit/1a0fa0a71bf0cfa1c19ee2c05bdbb9dad6cdb8b8.

zachmullen:
Fixes #2719

ping @msmolens

### DIFF
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -235,11 +235,11 @@ class User(Resource):
         user['email'] = email
 
         # Only admins can change admin state
-        if admin is True:
+        if admin in (True, False):
             if self.getCurrentUser()['admin']:
                 user['admin'] = admin
-            elif not user['admin']:
-                raise AccessException('Only admins may enable admin state.')
+            elif user['admin'] != admin:
+                raise AccessException('Only admins may change admin status.')
 
         # Only admins can change status
         if status is not None and status != user.get('status', 'enabled'):
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -238,7 +238,7 @@ class User(Resource):
         if admin in {True, False}:
             if self.getCurrentUser()['admin']:
                 user['admin'] = admin
-            elif user['admin'] != admin:
+            elif user['admin'] is not admin:
                 raise AccessException('Only admins may change admin status.')
 
         # Only admins can change status
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -235,7 +235,7 @@ class User(Resource):
         user['email'] = email
 
         # Only admins can change admin state
-        if admin in (True, False):
+        if admin in {True, False}:
             if self.getCurrentUser()['admin']:
                 user['admin'] = admin
             elif user['admin'] != admin:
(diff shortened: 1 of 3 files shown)


girder/api/v1/user.py
```diff
@@ -235,11 +235,11 @@ def updateUser(self, user, firstName, lastName, email, admin, status):
         user['email'] = email
 
         # Only admins can change admin state
-        if admin is True:
+        if admin in (True, False):
```
brianhelba: Actually, can you use a set `{ ... }` not a tuple here? It's more appropriate for the semantics.

girder/api/v1/user.py
```diff
@@ -235,11 +235,11 @@ def updateUser(self, user, firstName, lastName, email, admin, status):
         user['email'] = email
 
         # Only admins can change admin state
-        if admin is True:
+        if admin in {True, False}:
             if self.getCurrentUser()['admin']:
                 user['admin'] = admin
-            elif not user['admin']:
-                raise AccessException('Only admins may enable admin state.')
+            elif user['admin'] != admin:
```
brianhelba: Since we're dealing with boolean values, it's more Pythonic to compare with `is not`.

girder/api/v1/user.py
```diff
@@ -235,11 +235,11 @@ def updateUser(self, user, firstName, lastName, email, admin, status):
         user['email'] = email
 
         # Only admins can change admin state
-        if admin is True:
+        if admin in {True, False}:
```
msmolens: Doesn't `autoDescribeRoute` already coerce `admin` to a boolean, when provided? Can this be `if admin is not None:`?

brianhelba:
It's slightly concerning that this wasn't caught by previous tests. Maybe this case should be added?

girder/api/v1/user.py
msmolens: Could this always raise the exception if the `admin` parameter is provided but the current user is not an admin?

girder/api/v1/user.py
```diff
@@ -235,11 +235,11 @@ def updateUser(self, user, firstName, lastName, email, admin, status):
...
 
         # Only admins can change admin state
-        if admin is True:
+        if admin is not None:
             if self.getCurrentUser()['admin']:
                 user['admin'] = admin
-            elif not user['admin']:
-                raise AccessException('Only admins may enable admin state.')
+            elif user['admin'] is not admin:
+                raise AccessException('Only admins may change admin status.')
```
msmolens: When a non-admin specifies the `admin` parameter, I suggest raising this exception regardless of the existing value. What do you think?

zachmullen: For backward compatibility, I'd prefer not to change that behavior at this time. The `admin` field of user documents is exposed to everyone, so if someone attempts to naively save the user document with all of its fields that were originally sent down to them as params, it would fail after the proposed change.

msmolens: That makes sense, thanks.

zachmullen:
@msmolens PTAL

msmolens:
There could be a template to ensure a consistent structure for each release.

PR #2725, opened 2018-05-29
Skip autoDescribeRoute behavior for header params

zachmullen:
@brianhelba this is actually option 2 from my proposal on #2724, just to show the very simple fix, which is to do nothing with header params. We can merge this if you think it's a worthwhile stop-gap support for header params, otherwise it will take a bit more time to get the "option 1" solution logic in place.


brianhelba:
@zachmullen I'm going to merge this now, so I can incorporate the fix into https://github.com/girder/girder/pull/2655 .

PR #2728, opened 2018-05-30
Add privacy notice link

### LINKED ISSUE #2723: GDPR support Privacy Notice link
See #2722.

This issue is to create a Privacy Notice link at the footer of Girder, where the link target can be set as a system config property in the Admin UI.

This requirement should be relatively straightforward and well scoped compared to some of the other GDPR needs, so I wanted to break it out here.

Pierre-Assemat:
fixes #2723

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,6 +28,9 @@ Python Client
 
 Added Features
 --------------
+* Added a privacy notice link to the footer which can be set on the Server Configuration view of the web client (
+  `#2728 <https://github.com/girder/girder/pull/2728>`_).
+
 * Added a setting to disable the notification stream. This may improve Girder's performance in runtime environments with
   fewer threads (`#2712 <https://github.com/girder/girder/pull/2712>`_).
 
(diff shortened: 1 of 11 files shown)


girder/constants.py
```diff
@@ -205,6 +206,7 @@ class SettingDefault(object):
             'users': []
         },
         SettingKey.CONTACT_EMAIL_ADDRESS: 'kitware@kitware.com',
+        SettingKey.PRIVACY_NOTICE: 'https://www.kitware.com/privacy',
```
mgrauer: I'm not sure if we want to point the default to Kitware's privacy policy, but it makes as much sense as pointing the default contact email to kitware@kitware.com, which we currently do. I think this is OK.

mgrauer:
I don't see any problems.

@zachmullen Can you do a QA pass with an actual dev env?

mgrauer:
I forgot, please add a change log entry.

CHANGELOG.rst
```diff
@@ -28,6 +28,9 @@ Python Client
 
 Added Features
 --------------
+* Added a privacy notice link to the footer which can be configure on the girder setting view (
```
mgrauer: `configure on the girder setting view` => `set on the Server Configuration view of the web client`

PR #2730, opened 2018-06-01
Ensure models connect to current pytest database

danlamanna:
This fixes an issue where tests run serially were not reconnecting to the correct new database each time.


brianhelba:
We should remove the calls to `.remove` in each of the model fixtures, since they are now no longer all be written to the same database.

Also, I think it'd be even safer if we destroyed the model singleton instances each time, instead of just calling `.reconnect` on the existing instances. However, this doesn't seem to work properly, so perhaps a little more investigation is needed.

brianhelba:
@danlamanna It turns out it's never safe to destroy and re-create model singletons, so I've updated the comments to indicate so.

danlamanna:
@brianhelba best changes ever.

PR #2732, opened 2018-06-04
Add file upload functionality to server fixture

dorukozturk:


zachmullen:
Would it be worthwhile to port a test that makes use of this as part of this branch?

pytest_girder/pytest_girder/utils.py
```diff
@@ -250,3 +312,107 @@ def buildHeaders(headers, cookie, user, token, basicAuth, authHeader):
...
+            'parentType': parentType,
+            'parentId': str(parent['_id']),
+            'name': name,
+            'size': len(contents),
+            'mimeType': mimeType
+        })
+
+    fields = [('offset', 0), ('uploadId', resp.json['_id'])]
+    files = [('chunk', name, contents)]
+    resp = multipartRequest(
```
msmolens: It would be simpler to pass the data in the request body instead of using `multipart/form-data` encoding, especially considering this is intended only for small files. See: https://github.com/girder/girder/blob/6d9cc4d3cdbc13843f2eb74cca6aa08dd7649672/girder/api/v1/file.py#L235-L237

The upload could even be done in a single request by including the data in the `POST /file` request:
https://github.com/girder/girder/blob/6d9cc4d3cdbc13843f2eb74cca6aa08dd7649672/girder/api/v1/file.py#L77-L80

zachmullen: @msmolens this utility is just for testing legacy multipart mode, which is deprecated. You're right that it shouldn't be used in production, but we do want to test it to make sure it works for legacy clients.

msmolens: If it is indeed for testing the legacy functionality, I think that should be noted on the fixture.

dorukozturk: Sorry it took me a long time to respond. I only copied and pasted code from the old test suite to the pytest_girder package. I am working on a plugin and I am using pytest_girder package. As part of my testing I wanted to test file uploads and I just copy-pasted functionality from the legacy test suite as I mentioned. Do you guys think it is better to solve issues you mentioned in a separate PR in the future?

msmolens: That makes sense. I think it would be useful to note in the commit message or PR description that the code is copied from the legacy test suite.

dorukozturk: @msmolens I added a note to the commit message. Let me know if that is descriptive enough.

msmolens: LGTM, and I'll defer to others for PR approval

zachmullen: I think if we're going to implement only one mode in this helper, it should be the sanctioned mode (body + query params). I fear that otherwise, this code will not be touched for a long time and we'll forget that it's even using multipart internally. Using the normal upload mode also has the benefit of not duplicating arcane, ugly code, i.e. this PR will be much shorter and more palatable. Here's an example of uploading a chunk the modern way, taken from the old testing framework:

```
resp = self.request(
            path='/file/chunk', method='POST', user=self.user, body=chunk2, params={
                'offset': resp.json['offset'],
                'uploadId': uploadId
            }, type='text/plain')
```

The `body` param can be passed as an encoded string or a text type; if it's passed as a text type we will send it as UTF-8.

This change should be fairly straightforward. Let me know if you run into problems with it.

dorukozturk:
@zachmullen This is much more elegant. I tried it and it works for my use case.

PR #2733, opened 2018-06-04
Fix global nav to fade out

jtomeck:
This commit will remove an element responsible for creating a jagged edge where the sidebar fades out, and properly style it for a smooth fade.

### DIFF
--- a/clients/web/src/stylesheets/layout/globalNav.styl
+++ b/clients/web/src/stylesheets/layout/globalNav.styl
@@ -3,9 +3,7 @@ $emptyBorder=1px solid transparent
 $minHeight=500px
 
 .g-global-nav-main
-  -webkit-box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
-  -moz-box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
-  box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
+  box-shadow inset 0 -35px 45px -10px #fff
   background-image linear-gradient(to left, #f4f4f4 0, #fff 7px)
   min-height $minHeight - 12px
   padding 15px 0 20px 6px
--- a/clients/web/src/stylesheets/layout/globalNav.styl
+++ b/clients/web/src/stylesheets/layout/globalNav.styl
@@ -3,15 +3,12 @@ $emptyBorder=1px solid transparent
 $minHeight=500px
 
 .g-global-nav-main
-  box-shadow inset -1px 0 0 #eee
+  -webkit-box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
+  -moz-box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
+  box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
   background-image linear-gradient(to left, #f4f4f4 0, #fff 7px)
   min-height $minHeight - 12px
   padding 15px 0 20px 6px
-
-.g-global-nav-fade
-  height 12px
-  box-shadow inset -1px 0 0 #f5f5f5
-  background-image linear-gradient(to left, #f4f4f4 0, #fff 4px)
 
 .g-nav-link
   display block
@@ -92,5 +89,3 @@ ul.g-global-nav
     box-shadow none
     background-image none
     min-height 0
-  .g-global-nav-fade
-    display none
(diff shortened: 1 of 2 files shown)


clients/web/src/templates/layout/layoutGlobalNav.pug
```diff
@@ -5,4 +5,3 @@
         a.g-nav-link(g-target=navItem.target, g-name=navItem.name, href=`#${navItem.target}`)
           i(class=navItem.icon)
           span= navItem.name
-.g-global-nav-fade
```
brianhelba: @zachmullen You apparently added this `div` in https://github.com/girder/girder/commit/84604f16485e6032a19f14656dc3428707383bf5#diff-1b1a121412de63cbd86dd297a34998d5 . Is it fine to remove?

jtomeck: Interesting. Must have been something we did that added it back. That was the first thing I removed. Yes you can remove.

clients/web/src/stylesheets/layout/globalNav.styl
```diff
@@ -3,16 +3,13 @@ $emptyBorder=1px solid transparent
 $minHeight=500px
 
 .g-global-nav-main
-  box-shadow inset -1px 0 0 #eee
+  -webkit-box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
```
brianhelba: The Nib CSS utility library (which Girder always [imports via Stylus](https://github.com/girder/girder/blob/ba1544a009c6abf437a82810fa1a9f0529c16fa9/grunt_tasks/webpack.config.js#L105) should [provide vendor prefixes for `box-shadow` automatically](https://github.com/stylus/nib/tree/master/docs#transparent-mixins), so the rule should only need to defined once.

clients/web/src/stylesheets/layout/globalNav.styl
```diff
@@ -3,16 +3,13 @@ $emptyBorder=1px solid transparent
...
+  -moz-box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
+  box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
```
brianhelba: Would it be simpler to set the color as `white`, `#fff`, `#ffffff` instead of `rgba(255, 255, 255, 1)`?

clients/web/src/templates/layout/layoutGlobalNav.pug
brianhelba: @jtomeck This PR does indeed remove the `.g-global-nav-fade`. Since it was originally added by @zachmullen , I'm now asking whether there's still a reason why he thinks it should remain.

zachmullen: The element is fine to be removed, it was a hack for styling because I am not as good at CSS as @jtomeck  :)

I think I copied the implementation from someplace else, might even have been github from many years ago.

clients/web/src/stylesheets/layout/globalNav.styl
```diff
@@ -3,16 +3,13 @@ $emptyBorder=1px solid transparent
 $minHeight=500px
 
 .g-global-nav-main
-  box-shadow inset -1px 0 0 #eee
+  -webkit-box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
+  -moz-box-shadow inset 0 -35px 45px -10px rgba(255, 255, 255, 1)
```
zachmullen: Our stylus loader automatically includes the browser-specific version of rules at build time. Hence, we can omit the `-webkit` and `-moz` lines in the source files and just keep the `box-shadow` rule.

clients/web/src/stylesheets/layout/globalNav.styl
zachmullen: We can replace `rgba(255, 255, 255, 1)` with just `white` or `#fff`.

brianhelba:
Oh, it looks like I had the same feedback as @zachmullen .

clients/web/src/stylesheets/layout/globalNav.styl
jtomeck: Yes, good point. I was experimenting with different opacities, ended up going with fully opaque and never changed it. Thanks!

clients/web/src/stylesheets/layout/globalNav.styl
jtomeck: Thanks Brian, This is good to know moving forward!

PR #2734, opened 2018-06-05
Upgrade Candela plugin dependencies

waxlamp:
This uses recent versions of Candela and vega-loader.

@manthey, this should (finally) eliminate the canvas building problem that originated this effort.


manthey:
The tests are failing because PhantomJS isn't ES6 compliant.  We shim PhantomJS in the tests via core-js/shim, but evidently that isn't enough to handle things like `eval('class foo')`.  We either need to change how candela is transpiled or how we polyfill/shim PhantomJS.

waxlamp:
Thanks for your help, @manthey!

PR #2740, opened 2018-06-15
Use consistent field to return S3 HTTP request information

msmolens:
To perform a direct-to-S3 upload, a client makes a series of HTTP requests to various /file endpoints. The responses include information for the client to make an HTTP request to send data to S3 using a pre-signed URL.

For a multi-part (chunked) upload, the `POST /file` and `POST /file/chunk` endpoints return the request information in the `s3.request` field. In contrast, the `POST /file/completion` endpoint returns the request information in a `s3FinalizeRequest` field.

To make `POST /file/completion` consistent with the earlier steps in the multi-part direct-to-S3 upload workflow, this commit changes the response to use the `s3.request` field instead of `s3FinalizeRequest`.

### DIFF
--- a/clients/web/src/utilities/S3UploadHandler.js
+++ b/clients/web/src/utilities/S3UploadHandler.js
@@ -267,7 +267,7 @@ prototype._finalizeMultiChunkUpload = function () {
             root.appendChild(partEl);
         });
 
-        var req = resp.s3FinalizeRequest;
+        var req = resp.s3.request;
         var xhr = new XMLHttpRequest();
 
         xhr.open(req.method, req.url);
@@ -278,7 +278,7 @@ prototype._finalizeMultiChunkUpload = function () {
 
         xhr.onload = () => {
             if (xhr.status === 200) {
-                delete resp.s3FinalizeRequest;
+                delete resp.s3;
                 this.trigger('g:upload.complete', resp);
             } else {
                 this.trigger('g:upload.error', {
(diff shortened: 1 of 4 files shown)


zachmullen:
Definitely an improved specification, however this is a breaking change that will require changes in downstream girder clients that support S3 behavior. This could go into Girder 3, though.

msmolens:
I agree that it's technically a breaking change. However, practically, AFAIK the web client is the only downstream client, and the specification isn't documented.

zachmullen:
The nascent girder-vue client also implements this behavior, and some of our downstreams implement their own web front ends, so it's possible that this specification has proliferated pretty far.

mgrauer:
This is a good change, and it breaks the API, so we will wait until the 3.0 release window to merge it.

zachmullen:
Once merged, I will change the girder_web_components impl.

jeffbaumes:
@brianhelba you are assigned to review, can you review this soon?

jeffbaumes:
@brianhelba ping, waiting on review, could it be completed by Dec 7?

zachmullen:
I am voting that we do not merge this, at least not now. Consistency is desirable, but I don't think it's worth breaking existing upload clients just to have a consistent naming convention in this case.

Maybe in some future where we have a larger batch of API-breaking changes, it would make sense to include this change as well.

Issue #2741, opened 2018-06-18
improve logging


mgrauer:
Move from our own logging to standard Python logging, which will make our instances easier to maintain and debug under WSGI.

zachmullen:
@mgrauer could you clarify this a bit? What is non-standard about our logging that is problematic on WSGI?

manthey:
In the current Girder default, we capture stdout and stderr to logs (as well as letting them go to stdout and stderr).  By default, stdout gets logged at "INFO" level and stderr gets logged at "ERROR" level.  

I like to see all of Girder logs in one place, rather than split between error and info logs.  We have a setting for our standard logging to set the level that goes to the "info" log:
```
[logging]
log_max_info_level = "CRITICAL"  # send everything to the info log
```

Note that this doesn't stop ERROR-level messages from going to our default error log (but you can point the error and info logs to the same file to change that).

You can prevent stdout and stderr from reaching the actual stdout and stderr by adding:
```
log_quiet = True
```

I've noticed that when multiple processes are writing to the same log file, nothing bad happens except when log file rotation occurs, at which point each process will often perform the rotation (even though only one should do so). 

We could (as a plugin or as part of core), add additional log handlers.  For instance, we could add a log handler that would use syslog.  If we did this in lieu of the file log handler, it would be a way to better handle multiple processes.

zachmullen:
The problem here is that `_setupLogger` is called as an import side effect when importing anything under the girder package. `_setupLogger` adds these handlers that are problematic in the wsgi entry point.

Proposed solution: the import side effect should simply be `logger = logging.getLogger('girder')` rather than `logger = _setupLogger()`. It should be up to each entrypoint to configure the handlers, so the call to handler configuration could be moved to `girder.cli.serve:main`, and then wsgi could have its own call. That's certainly not the only way to fix it, we could alternatively keep the present behavior and just make the handlers more configurable via the config files.

brianhelba:
@manthey I agree, we should have one log file with everything, as this is far easier to capture and analyze using any sort of automated tools (it's much easier to filter one file than merge two). I think this is a change we can make to Girder right now, and hopefully it may allow us to eliminate some lines of code and associated config fields.

@zachmullen The change to call `_setupLogger` as part of the CLI startup sounds great, and would help to clarify the fact that the *only* responsibility of `_setupLogger` is to attach a set of handlers to the central logger (accordingly, [this line](https://github.com/girder/girder/blob/master/girder/__init__.py#L168)
[...]

PR #2742, opened 2018-06-20
More flexible logging configuration

zachmullen:


zachmullen:
@mgrauer 1e49b61 is the minimum change set described in #2742. With that commit, handlers are no longer attached to the logger at import time. What do we need to do for wsgi mode?

This doesn't remove the stdout/stderr logging configuration that is called during `server.setup` as I think that change is more controversial and probably better done by @manthey .

mgrauer:
I’ll leave it to @brianhelba or @danlamanna to advise on what we need for wsgi. 

I know we will want the ability to not write to std out/err from anywhere in Girder for this.

zachmullen:
Some of the material I was reading seemed to indicate that we *should* be sending log messages to stdout/stderr, which causes the messages to go into the WSGI container log. However, based on your comments in #2741 it seems like having them rolled into the WSGI container service's logs is undesirable. What would be preferred?

test/test_logging_config.py
```diff
@@ -44,7 +44,8 @@ def configureLogging(logConfig={}, oneFile=False):
         os.unlink(cfg['logging']['error_log_file'])
     cfg['logging'].update(logConfig)
 
-    girder.logger = girder._setupLogger()
+    girder.logger.setLevel('INFO')
```
brianhelba: We generally should never be setting the level of the central `girder.logger`, rather we should prefer to set individual handlers. Why is this being done for this test?

tests/cases/system_test.py
```diff
@@ -496,11 +496,15 @@ def testLogRoute(self):
         del config.getConfig()['logging']
 
     def testLogLevel(self):
-        from girder import logger
+        from girder import logger, _attachFileLogHandlers
+        _attachFileLogHandlers()
+        logger.setLevel('INFO')
```
brianhelba: Again, why is it necessary to set the level of the logger itself here?

girder/wsgi.py
```diff
@@ -18,14 +18,24 @@
...
-config.loadConfig()  # Read the config files first, so we can override some values
 cherrypy.config.update({'engine.autoreload.on': False,
                         'environment': 'embedded'})
 cherrypy.config['server'].update({'cherrypy_server': False,
                                   'disable_event_daemon': True})
 
+# TODO The below line can be removed if we do away with girder.logprint
+girder._quiet = True  # This means we won't duplicate messages to stdout/stderr
+_formatter = girder.LogFormatter('[%(asctime)s] %(levelname)s: %(message)s')
+_handler = logging.handlers.StreamHandler()  # Send logs to stderr
```
brianhelba: Per the (slightly bitter) descriptions [here](https://modwsgi.readthedocs.io/en/develop/user-guides/application-issues.html#writing-to-standard-output) and [here](https://modwsgi.readthedocs.io/en/develop/user-guides/debugging-techniques.html#apache-error-log-files) it looks like we should be writing to the WSGI environ value `'wsgi.errors'`, which is a stream. However, while CherryPy does provide this via `request.wsgi_environ['wsgi.errors']`, the value is going to reset on every request, it doesn't allow us to log messages that occur outside of a request (like server startup), and it's probably going to be awkward within Girder to attach a new handler at the start of every request and detach one once it's over (though it's probably possible if we dig into CherryPy's wildly-undocumented [WSGI middleware capabilities](https://github.com/cherrypy/cherrypy/blob/master/cherrypy/_cpwsgi.py).

Accordingly, I think the second best we can do is to log to `sys.stderr` here. It looks like we should attempt to never write to `sys.stdout` if we want Apache's mod_wsgi to be totally happy (and that's what's used by AWS Beanstalk by default).

girder/wsgi.py
```diff
@@ -18,14 +18,24 @@
...
+_handler.setFormatter(_formatter)
+girder.logger.addHandler(_handler)
+cherrypy.log.access_log.addHandler(_handler)
+cherrypy.log.error_log.addHandler(_handler)
```
brianhelba: Do we know what's being sent to `cherrypy.log.access_log` and `cherrypy.log.error_log`. I think it'd be ideal if could ignore these and prefer `girder.logger` as the one logger to rule them all.

test/test_logging_config.py
zachmullen: Because by default, the level of a logger is 'WARNING' (inherited from the default of the root logger), and our tests expect 'INFO'-level messages to appear in the logs.

girder/wsgi.py
zachmullen: Internal stuff from cherrypy, I'm assuming. I'm fine either way, was just copying some of the logic that happens inside the old method.

girder/wsgi.py
zachmullen: I can't tell if this comment is requesting a change here or affirming that I made the correct decision :)

test/test_logging_config.py
brianhelba: Interesting point. Given the [behavior of `logging.Logger.setLevel`](https://docs.python.org/3/library/logging.html#logging.Logger.setLevel), I think we should call
`logging.getLogger('girder').setLevel(logging.DEBUG)` immediately after the base Girder logger is created: https://github.com/girder/girder/blob/0c28e5be9f5eaee1f5220930d142908818d2b234/girder/__init__.py#L41

The alternative could be that Girder 'takes over' the Python root logger (`logging.getLogger('girder')`) and sets its level to `logging.NOTSET`. However, particularly in the short term, I think it's good if we discipline ourselves to ensure that everything is logged to `logging.getLogger('girder')` or a descendant logger.

girder/wsgi.py
brianhelba: We should change it to log to `sys.stderr` only.

I honestly don't understand how to interpret this [statement from the docs](https://docs.python.org/2/library/logging.handlers.html#streamhandler) means:
> If *stream* is specified, the instance will use it for logging output; otherwise, *sys.stderr* will be used.

but I think we should just have:
```
_handler = logging.handlers.StreamHandler(sys.stderr)
```
instead.

zachmullen: It means `sys.stderr` is the default if none is passed, so your proposed change won't affect behavior, but it's worthwhile for readability.

girder/wsgi.py
```diff
@@ -18,14 +18,24 @@
...
 
-config.loadConfig()  # Read the config files first, so we can override some values
 cherrypy.config.update({'engine.autoreload.on': False,
                         'environment': 'embedded'})
 cherrypy.config['server'].update({'cherrypy_server': False,
                                   'disable_event_daemon': True})
 
+# TODO The below line can be removed if we do away with girder.logprint
+girder._quiet = True  # This means we won't duplicate messages to stdout/stderr
+_formatter = girder.LogFormatter('[%(asctime)s] %(levelname)s: %(message)s')
```
brianhelba: Is it possible or desirable to follow the [Common Log Format](https://en.wikipedia.org/wiki/Common_Log_Format) here?

girder/wsgi.py
brianhelba: Oops, somehow I totally misread those docs, and thought it was talking about `sys.stdout` also in some way.

brianhelba: However, I've dug into CherryPy a bit more, and they do provide a [`WSGIErrorHandler`](https://github.com/cherrypy/cherrypy/blob/master/cherrypy/_cplogging.py#L441), which we could instantiate and attach ourselves to the base Girder logger (CherryPy also has an API to attach one to the CherryPy loggers). However, as mentioned, this would still not capture log output from times outside of a request context.

What things does Girder log outside of a request? Obviously plugin enablement, but is there anything else? Additionally, how much do we care about Apache being able to associate Girder's logs with the particular virtual host that it's running under? Do we anticipate running Girder with multiple virtual host web apps on the same Apache server?

girder/__init__.py
```diff
@@ -136,13 +138,12 @@ def getLogPaths():
     }
 
 
-def _setupLogger():
+def _attachFileLogHandlers():
```
brianhelba: Is this function expected to be called multiple times? There is some logic within that removes all handlers from the main Girder logger, but I wonder if it's necessary:
https://github.com/girder/girder/blob/0c28e5be9f5eaee1f5220930d142908818d2b234/girder/__init__.py#L180-L185

If tests need to reset / remove loggers, they should do so explicitly.

girder/wsgi.py
zachmullen: I think that's a good idea. Would it be too disruptive to also change it in the main girder server context?

girder/__init__.py
zachmullen: I'll remove that logic and see what breaks :)

girder/wsgi.py
zachmullen: Actually it looks like that format only covers access logs, i.e. basic request info, not general purpose logging. I vote we leave formatting changes for a future branch.

zachmullen:
@manthey looking again at the logic inside `_attachFileLogHandlers`, I believe the filters will still make it so that only 'INFO' and above get logged with the default log level, so I don't expect you'd see any behavior changes if you're not using wsgi. However, it's probably worth testing that theory to make sure.

girder/wsgi.py
zachmullen: > What things does Girder log outside of a request? Obviously plugin enablement, but is there anything else? 

This is kind of a design question -- are the logs meant to be usable in all contexts of Girder, or only when it's being used as a web server? I could go either way on this.

> Additionally, how much do we care about Apache being able to associate Girder's logs with the particular virtual host that it's running under? Do we anticipate running Girder with multiple virtual host web apps on the same Apache server?

I'm not sure, but it would be good to support this in the most first-class way possible.

zachmullen: @brianhelba @girder/developers thoughts on this? Should I attach this handler instead of the StreamHandler?

zachmullen:
This needs a final review, @brianhelba or @danlamanna

mgrauer:
@zachmullen please refresh this.

@brianhelba is ok with merging this.

We should have a larger discussion about logging approaches in a Girder planning meeting.

zachmullen:
Rebased, awaiting final review.

PR #2745, opened 2018-06-20
Use JQuery in setting up Vega visualizations

waxlamp:
This addresses @brianhelba's [concern](https://github.com/girder/girder/pull/2715#pullrequestreview-129258827) from #2715.

### DIFF
--- a/plugins/vega/web_client/views/VegaWidget.js
+++ b/plugins/vega/web_client/views/VegaWidget.js
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import _ from 'underscore';
 
 import { parse,
     View as VegaView } from 'vega-lib';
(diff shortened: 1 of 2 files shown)


plugins/vega/web_client/views/VegaWidget.js
```diff
@@ -29,13 +30,13 @@ var VegaWidget = View.extend({
             restRequest({
                 url: `item/${this.item.id}/download`
             })
-                .done(function (spec) {
+                .done(_.bind(function (spec) {
                     let runtime = parse(spec);
                     let view = new VegaView(runtime)
-                        .initialize(document.querySelector('.g-item-vega-vis'))
+                        .initialize($('.g-item-vega-vis')[0])
```
brianhelba: Can this be `this.$`? It's nice if we can scope these queries, but I realize sometimes the element to be selected is not a direct child of the current view's `this.el`.

waxlamp: I believe the latter is the case here.

waxlamp:
@brianhelba is this good to go or do you want to address any deeper issues here?

waxlamp:
@brianhelba I sort of lost the thread on this; is it still ok to merge, and we can clean up concerns about `$`-scoping later?

PR #2746, opened 2018-06-20
Also catch TypeErrors from GeoJSON parsing

### LINKED ISSUE #2744: Correct syntax for GeoJSON for the Geospatial Plugin
I recently installed the Geospatial plugin for Girder and was excited to plot GeoJSON coordinates onto an image. I went to the REST API and went to PUT /item/{id}/geospatial under the item header. 
image

However, when I put GeoJSON coordinates that were known to work in geojson.io into the text field, I got a TypeError (string indices must be integers):
image

The geojson.io coordinates are here:
image

And here is the code for the coordinates I used:
`{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              110,
              55.32914440840507
            ],
            [
              89.20898437499999,
              55.32914440840507
            ],
            [
              89.20898437499999,
              61.897577621605016
            ],
            [
              70.048828125,
              61.897577621605016
            ],
            [
              110,
              55.32914440840507
            ]
          ]
        ]
      }
    }
  ]
}`

Is there something I wrote incorrectly above? I also changed the coordinates to integers but I still got the same "string indices must be integers" issue. I'm not sure what that refers to.

zachmullen:
Fixes #2744 

It's quite possible that the design of the REST endpoints in the geospatial plugin is fundamentally bad, but I don't know. In either case, though, we need to catch this exception type.

### DIFF
--- a/plugins/geospatial/server/geospatial.py
+++ b/plugins/geospatial/server/geospatial.py
@@ -326,7 +326,7 @@ class GeospatialItem(Resource):
             if v:
                 try:
                     GeoJSON.to_instance(v, strict=True)
-                except ValueError:
+                except (ValueError, TypeError):
                     raise RestException('Geospatial field with key %s does not'
                                         ' contain valid GeoJSON: %s' % (k, v))
 


zachmullen:
@girder/developers could someone review this small change? It should be fairly uncontroversial.

PR #2751, opened 2018-06-27
Existing folder and item names are considered valid.

### LINKED ISSUE #2750: Changing permissions on a folder with a duplicate name
Although we try to avoid duplicate item and folder names, we don't strictly do so (see issue #1997).  If you do have duplicate names on a folder, changing the permissions on such a folder fails because the validation step rejects it due to the duplicate name.

One fix would be to skip checking for duplicate names if an existing folder or item is saved without changing its name.

manthey:
Don't validate folder and item names for existing entries for uniqueness.  If the folder or item doesn't change its name and is not unique, changing non-name properties should validate.

Fixes #2750.

### DIFF
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -79,9 +79,11 @@ class Folder(AccessControlledModel):
                                   doc['parentCollection'],
                                   'girder.models.folder.invalid-parent-type')
         name = doc['name']
+        # If the folder already exists with the current name, don't check.
+        checkName = '_id' not in doc or not self.findOne({'_id': doc['_id'], 'name': name})
         n = 0
         itemModel = Item()
-        while True:
+        while checkName:
             q = {
                 'parentId': doc['parentId'],
                 'name': name,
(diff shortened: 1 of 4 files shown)


girder/models/folder.py
```diff
@@ -79,9 +79,11 @@ def validate(self, doc, allowRename=False):
                                   doc['parentCollection'],
                                   'girder.models.folder.invalid-parent-type')
         name = doc['name']
+        # If the folder already exists with the current name, don't check.
+        checkName = '_id' not in doc or not self.findOne({'_id': doc['_id'], 'name': name})
```
zachmullen: I'm assuming the query engine is smart enough that this will not be asymptotically slower than a normal `_id` query? If it is for some reason slower, we should just do the findOne and then check the name in python.

brianhelba: Both `name` and `_id` have indexes, and there's no strange compound logic here. I'd be very surprised if this didn't scale properly.

zachmullen: I would also be very surprised; just the presence of a unique index field (_id) in a findOne query should mean it's nearly identical in query time to an _id alone. Just wanting to sanity check.

PR #2760, opened 2018-07-10
Pytest live server fixture

dorukozturk:
Adds a live server fixture to be used for integration tests. This fixture starts a girder server on a random
port and returns a girder client instance which points to the girder server. It also does not use any 
mocked fixtures such as bicrypt to test the real behavior of a live server.

### DIFF
--- a/pytest_girder/pytest_girder/fixtures.py
+++ b/pytest_girder/pytest_girder/fixtures.py
@@ -5,8 +5,8 @@ import mongomock
 import os
 import pytest
 import shutil
-
 import socket
+import subprocess
 import time
 import warnings
 try:
@@ -15,6 +15,8 @@ try:
 except ImportError:
     from urllib.request import urlopen
     from urllib import URLError
+
+from girder_client import GirderClient
 from .utils import uploadFile, MockSmtpReceiver, request as restRequest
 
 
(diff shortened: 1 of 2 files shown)


manthey:
@dorukozturk I have a similar desire to have a pytest girder server that can receive actual http calls, specifically to allow it to work with girder_worker in testing.  In this branch, you've yielded a GirderClient.  Rather than do this, we could instead yield a `server` object like the `server` fixture.  That is:

This is mostly the `server` fixture moved to a generator function.
```
def _yieldServer(request, exposePort=False):
    # The event daemon cannot be restarted since it is a threading.Thread object, however
    # all references to girder.events.daemon are a singular global daemon due to its side
    # effect on import. We have to hack around this by creating a unique event daemon
    # each time we startup the server and assigning it to the global.
    import girder.events
    from girder.api import docs
    from girder.constants import SettingKey
    from girder.models.setting import Setting
    from girder.utility.server import setup as setupServer

    girder.events.daemon = girder.events.AsyncEventsThread()

    enabledPlugins = []
    hasPluginMarkers = request.node.get_closest_marker('plugin') is not None
    pluginRegistry = PluginRegistry()

    with pluginRegistry():
        if hasPluginMarkers:
            for pluginMarker in request.node.iter_markers('plugin'):
                pluginName = pluginMarker.args[0]
                shouldEnable = pluginMarker.kwargs.pop('enabled', True)
                if len(pluginMarker.args) > 1:
                    pluginRegistry.registerTestPlugin(
                        *pluginMarker.args, **pluginMarker.kwargs
                    )
                if shouldEnable:
                    enabledPlugins.append(pluginName)

        Setting().set(SettingKey.PLUGINS_ENABLED, enabledPlugins)

        cherrypy.config['server.socket_port'] = 9080
        server = setupServer(test=True, plugins=enabledPlugins)
        server.request = restRequest
        server.uploadFile = uploadFile
        if exposePort:
            port = _getFreeTcpPort()
            cherrypy.config['server.socket_port'] = port
            server.exposedPort = port

        # I don't know if subscribing twice causes problems (from multiple tests), so always unsubscribe and resubscribe if necessary
        cherrypy.server.unsubscribe()    
        if exposePort:
            cherrypy.server.subscribe()
        cherrypy.config.update({'environment': 'embedded',
                                'log.screen': False,
                                'request.throw_errors': True})
        cherrypy.engine.start()

        yield server

        cherrypy.engine.unsubscribe('start', girder.events.daemon.start)
        cherrypy.engine.unsubscribe('stop', girder.events.daemon.stop)
        cherrypy.engine.stop()
        cherrypy.engine.exit()
        cherrypy.tree.apps = {}
        docs.routes.clear()


@pytest.fixture
def server(db, request):
    for server in _yieldServer(request):
        yield server

@pytest.fixture
def exposedServer(db, request):
    for server in _yieldServer(request, exposePort=True):
        yield server
```

Then, individual pytests can use `requests.get('http://127.0.0.1:%d/api/v1/...' % liveServer.exposedPort)` as well as `liveServer.request` if they don't need to go through actual http.

manthey:
Closing this is its was satisfied by #2904, though not quite in the same way.

PR #2762, opened 2018-07-13
If a DICOM metadata has a binary value, it could prevent parsing.

manthey:
We don't store values we don't understand, but we were trying to store arbitrary binary values.  This fails if we can't coerce the value to UTF-8 to stick it in Mongo.  If we actually want to store binary data, we need to encode it different.  This skips such binary data, including any data that doesn't encode to UTF-8 or has a null in it.

This uses a pydicom test file for testing.

### DIFF
--- a/plugins/dicom_viewer/server/__init__.py
+++ b/plugins/dicom_viewer/server/__init__.py
@@ -134,6 +134,15 @@ def _removeUniqueMetadata(dicomMeta, additionalMeta):
 
 
 def _coerceValue(value):
+    # For binary data, see if it can be coerced further into utf8 data.  If
+    # not, mongo won't store it, so don't accept it here.
+    if isinstance(value, six.binary_type):
+        if b'\x00' in value:
+            raise ValueError('Binary data with null')
+        try:
+            value.decode('utf-8')
+        except UnicodeDecodeError:
+            raise ValueError('Binary data that cannot be stored as utf-8')
     # Many pydicom value types are subclasses of base types; to ensure the value can be serialized
     # to MongoDB, cast the value back to its base type
     for knownBaseType in {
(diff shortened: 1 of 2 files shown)


plugins/dicom_viewer/server/__init__.py
```diff
@@ -134,6 +134,15 @@ def _removeUniqueMetadata(dicomMeta, additionalMeta):
 
 
 def _coerceValue(value):
+    # For binary data, see if it can be coerced further into utf8 data.  If
+    # not, mongo won't store it, so don't accept it here.
+    if isinstance(value, six.binary_type):
```
zachmullen: Are these values always coming in as `six.binary_type` or could they be other things, e.g. `bytearray`?

manthey: I don't know, but we reject things that aren't subclasses of `six.binary_type` or other known types below and raise a `ValueError`.

manthey: We can do that, which will also need to add `bytearray` to the `knownBaseType` set below.  If you have a `bytearray` as a value, does it insert into Mongo?  Must it be unicode like a `six.binary_type`?

manthey: Looking at pydicom, it uses `bytearray` for pixel data and some other data.  I'd say since we are many extracting data for use in searches, extracting `bytearray` or true binary data is not particularly desirable.  Even though we could do so, I'm going to recommend we don't (or at least not as part of this PR).  This PR really just tries to get back to the functionality we had of displaying DICOM files regardless of what metadata was in them; the failure to parse binary strings was one  such issue.

zachmullen: Fair enough

PR #2765, opened 2018-07-20
Move proxy settings docs to appropriate place

zachmullen:
@felixveysseyre


zachmullen:
@girder/developers PTAL, this fixes a bug in the location of a section of documentation.

PR #2767, opened 2018-07-23
Fix a client error raised when deleting a collection

jbeezley:
This was introduced by a bad copy/paste in 681547c932.


manthey:
This looks good to me, but should we add a test that covers this line?

zachmullen:
Given the moribund state of the Backbone client, unless the test can be added very quickly, I'm fine with merging without a test.

PR #2768, opened 2018-07-24
Add an endpoint to poll for notifications

jbeezley:
This adds a notification listing endpoint at `/notification`.  It returns the same information provided by `/notification/stream`, but without the issues with long-lived request threads.  It would be up to the client to fall back to manual long polling.

The implementation is simple, but I'll add a test if everyone agrees this is a good idea.

### DIFF
--- a/girder/api/v1/notification.py
+++ b/girder/api/v1/notification.py
@@ -121,6 +121,8 @@ class Notification(Resource):
     def listNotifications(self, params):
         user, token = self.getCurrentUser(returnToken=True)
         since = params.get('since')
+        setResponseHeader('Date', str(time.time()))
+
         if since is not None:
             since = datetime.utcfromtimestamp(since)
         return list(NotificationModel().get(user, since, token=token))
--- a/girder/api/v1/notification.py
+++ b/girder/api/v1/notification.py
@@ -55,7 +55,7 @@ class Notification(Resource):
         super(Notification, self).__init__()
         self.resourceName = 'notification'
         self.route('GET', ('stream',), self.stream)
-        self.route('GET', (), self.list)
+        self.route('GET', (), self.listNotifications)
 
     @access.cookie
     @access.token
@@ -118,7 +118,7 @@ class Notification(Resource):
         .errorResponse()
         .errorResponse('You are not logged in.', 403)
     )
-    def list(self, params):
+    def listNotifications(self, params):
         user, token = self.getCurrentUser(returnToken=True)
         since = params.get('since')
         if since is not None:
--- a/girder/api/v1/notification.py
+++ b/girder/api/v1/notification.py
@@ -121,13 +121,13 @@ class Notification(Resource):
     def listNotifications(self, params):
         user, token = self.getCurrentUser(returnToken=True)
         since = params.get('since')
+        timestamp = time.time()
 
         if since is not None:
             since = datetime.utcfromtimestamp(since)
         notifications = list(NotificationModel().get(user, since, token=token))
-        if not notifications:
-            timestamp = time.time()
-        else:
+
+        if notifications:
             timestamp = notifications[0]['updatedTime']
             for notification in notifications:
                 timestamp = max(timestamp, notification['updatedTime'])
--- a/girder/api/v1/notification.py
+++ b/girder/api/v1/notification.py
@@ -55,6 +55,7 @@ class Notification(Resource):
         super(Notification, self).__init__()
         self.resourceName = 'notification'
         self.route('GET', ('stream',), self.stream)
+        self.route('GET', (), self.list)
 
     @access.cookie
     @access.token
@@ -105,3 +106,21 @@ class Notification(Resource):
 
                 time.sleep(wait)
         return streamGen
+
+    @access.cookie
+    @access.token
+    @autoDescribeRoute(
+        Description('List notification events')
+        .notes('This endpoint can be used for manual long-polling when '
+               'SSE support is disabled or otherwise unavailable.')
+        .param('since', 'Filter out events before this time stamp.',
+               dataType='integer', required=False)
+        .errorResponse()
+        .errorResponse('You are not logged in.', 403)
+    )
+    def list(self, params):
+        user, token = self.getCurrentUser(returnToken=True)
+        since = params.get('since')
+        if since is not None:
+            since = datetime.utcfromtimestamp(since)
+        return list(NotificationModel().get(user, since, token=token))
--- a/girder/api/v1/notification.py
+++ b/girder/api/v1/notification.py
@@ -118,9 +118,8 @@ class Notification(Resource):
         .errorResponse()
         .errorResponse('You are not logged in.', 403)
     )
-    def listNotifications(self, params):
+    def listNotifications(self, since):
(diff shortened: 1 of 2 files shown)


zachmullen:
+1

manthey:
Can the server return its current epoch (in a header?) so that the client can use it as the next `since` parameter to ensure it neither misses anything nor gets things twice?

girder/api/v1/notification.py
```diff
@@ -105,3 +106,21 @@ def streamGen():
...
+    @autoDescribeRoute(
+        Description('List notification events')
+        .notes('This endpoint can be used for manual long-polling when '
+               'SSE support is disabled or otherwise unavailable.')
+        .param('since', 'Filter out events before this time stamp.',
+               dataType='integer', required=False)
+        .errorResponse()
+        .errorResponse('You are not logged in.', 403)
+    )
+    def list(self, params):
```
brianhelba: Can you use a better name for this resource handler?

brianhelba:
@manthey Is there a standard HTTP header that you're aware of that would hold this sort of thing? I'm reluctant to invent our own just for this endpoint.

brianhelba:
@manthey Alternatively, do each of the included events include a timestamp? Would an application be able to use the latest timestamp of all events as the `since` value?

manthey:
We could use `Last-Event-ID`, or use `Last-Modified` or `Date`.

brianhelba:
`Last-Event-ID` seems to be a _request_ header associated with SSE / EventSource reconnection, so I don't think that would be appropriate for the server to set in a response to a normal endpoint.

`Last-Modified` is associated with HTTP caching (`If-Modified-Since`, etc.), which Girder has generally avoided worrying about. While not strictly inappropriate here, I'd avoid this header until we're ready to consider HTTP caching headers more comprehensively.

`Date` makes total sense, as it's just an informative statement of the server's current date (I assume this is what we want, the epoch is 1970). It wouldn't seem unreasonable to include it with every HTTP response, but setting it just in this endpoint for now would be fine too.

TLDR: +1 to add a `Date` header.

girder/api/v1/notification.py
```diff
@@ -105,3 +106,23 @@ def streamGen():
...
+               'SSE support is disabled or otherwise unavailable.')
+        .param('since', 'Filter out events before this time stamp.',
+               dataType='integer', required=False)
+        .errorResponse()
+        .errorResponse('You are not logged in.', 403)
+    )
+    def listNotifications(self, params):
+        user, token = self.getCurrentUser(returnToken=True)
+        since = params.get('since')
+        setResponseHeader('Date', str(time.time()))
```
zachmullen: As currently written, there is a chance that notifications would be duplicated during polling if they get written between the execution of line 124 and 128. In the stream version, the last update time is set as the maximum of the timestamps on the fetched batch of events. Let's adopt the same approach here, or at least some solution that relies on the database as the authoritative source of truth.

zachmullen: (As an aside, those sorts of race conditions will exist regardless because do not generate and write timestamps to the database atomically. If we wanted to fix that, we can use mongo's `$currentDate` operator on writes.)

jbeezley: I fixed it the best I could.  We can't always use the same method as the notification stream because there won't always be a notification to send back.  As you said though, these race conditions are possible regardless.

girder/api/v1/notification.py
```diff
@@ -121,8 +121,16 @@ def streamGen():
...
         user, token = self.getCurrentUser(returnToken=True)
         since = params.get('since')
-        setResponseHeader('Date', str(time.time()))
 
         if since is not None:
             since = datetime.utcfromtimestamp(since)
-        return list(NotificationModel().get(user, since, token=token))
+        notifications = list(NotificationModel().get(user, since, token=token))
+        if not notifications:
+            timestamp = time.time()
```
zachmullen: This `time.time()` should be computed prior to the query so we don't miss any notifications.

girder/api/v1/notification.py
```diff
@@ -121,13 +121,13 @@ def streamGen():
...
             since = datetime.utcfromtimestamp(since)
         notifications = list(NotificationModel().get(user, since, token=token))
-        if not notifications:
-            timestamp = time.time()
-        else:
+
+        if notifications:
             timestamp = notifications[0]['updatedTime']
             for notification in notifications:
                 timestamp = max(timestamp, notification['updatedTime'])
```
zachmullen: Lines 131-133 can be rewritten as simply:

`timestamp = max(n['updatedTime'] for n in notifications)`

girder/api/v1/notification.py
```diff
@@ -105,3 +106,31 @@ def streamGen():
...
+        .notes('This endpoint can be used for manual long-polling when '
+               'SSE support is disabled or otherwise unavailable.')
+        .param('since', 'Filter out events before this time stamp.',
+               dataType='integer', required=False)
+        .errorResponse()
+        .errorResponse('You are not logged in.', 403)
+    )
+    def listNotifications(self, params):
+        user, token = self.getCurrentUser(returnToken=True)
+        since = params.get('since')
```
zachmullen: If you make a `since` kwarg you won't have to read it out of the params dict.

brianhelba:
@jbeezley @zachmullen I think there's a problem with this PR as-is. Please see my inline comment.

girder/api/v1/notification.py
```diff
@@ -105,3 +106,28 @@ def streamGen():
...
+    def listNotifications(self, since):
+        user, token = self.getCurrentUser(returnToken=True)
+        timestamp = time.time()
+
+        if since is not None:
+            since = datetime.utcfromtimestamp(since)
+        notifications = list(NotificationModel().get(user, since, token=token))
+
+        if notifications:
+            timestamp = max(n['updatedTime'] for n in notifications)
```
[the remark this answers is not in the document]
zachmullen: We could just control this completely on the client side without the need to send any headers at all. The client can simply poll this endpoint, and if it receives nothing in response, it just polls with the same parameters as before. If it receives a non-empty list in response, it iterates through the list and pulls the max timestamp from it, and uses that on the next poll.

If we choose to do it that way, we may want to support ISO 8601 dates instead of (or in addition to) epoch times for the `since` parameter.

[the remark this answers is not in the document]
brianhelba: @zachmullen Yup, I'd be fine if we want to implement this on the client-side instead. As I mentioned before:
> Alternatively, do each of the included events include a timestamp? Would an application be able to use the latest timestamp of all events as the since value?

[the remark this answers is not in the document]
brianhelba: @jbeezley @zachmullen Also, the current use of Unix times as the value of the `Date` HTTP header is not compliant with the RFC, the value must be encoded as an RFC-1123 date.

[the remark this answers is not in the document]
zachmullen: I'll put in a follow up PR shortly that will simplify this endpoint.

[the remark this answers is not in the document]
zachmullen: refs #2769

PR #2769, opened 2018-07-25
Change behavior of notification polling endpoint

zachmullen:
* Rather than pass an HTTP header to clients indicating the next
  "since" parameter to poll with, it's left to clients to infer
  it from the response.
* The "since" parameter for this endpoint now supports date strings
  in addition to epoch timestamps.
* The results of the endpoint are now guaranteed to be returned
  sorted chronologically; this means clients can discover the next
  "since" value to use in O(1) time rather than O(n).

@brianhelba @jbeezley let me know what you think, I'll write tests when we agree on the behavior.

### DIFF
--- a/girder/api/v1/notification.py
+++ b/girder/api/v1/notification.py
@@ -18,7 +18,6 @@
 ###############################################################################
 
 import cherrypy
-import dateutil.parser
 import json
 import time
 from datetime import datetime
@@ -115,23 +114,11 @@ class Notification(Resource):
         .notes('This endpoint can be used for manual long-polling when '
                'SSE support is disabled or otherwise unavailable. The events are always '
                'returned in chronological order.')
-        .param('since', 'Filter out events before this date. Pass as either epoch seconds or '
-               'an ISO-8601 formatted datetime.', required=False)
+        .param('since', 'Filter out events before this date.', required=False, dataType='dateTime')
         .errorResponse()
         .errorResponse('You are not logged in.', 403)
     )
     def listNotifications(self, since):
         user, token = self.getCurrentUser(returnToken=True)
-
-        if since is not None:
-            try:
-                since = dateutil.parser.parse(since)
-            except ValueError:
-                try:
-                    since = datetime.utcfromtimestamp(float(since))
-                except ValueError:
-                    raise RestException(
-                        'The "since" parameter must be a valid date string or epoch timestamp.')
-
         return list(NotificationModel().get(
             user, since, token=token, sort=[('updated', SortDir.ASCENDING)]))
--- a/girder/api/v1/notification.py
+++ b/girder/api/v1/notification.py
@@ -18,13 +18,14 @@
 ###############################################################################
 
 import cherrypy
+import dateutil.parser
 import json
 import time
 from datetime import datetime
 
 from ..describe import Description, autoDescribeRoute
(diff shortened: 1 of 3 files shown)


jbeezley:
LGTM, but I'll let @brianhelba do the official approval.

girder/api/v1/notification.py
```diff
@@ -112,22 +113,25 @@ def streamGen():
...
-            since = datetime.utcfromtimestamp(since)
-        notifications = list(NotificationModel().get(user, since, token=token))
-
-        if notifications:
-            timestamp = max(n['updatedTime'] for n in notifications)
-
-        setResponseHeader('Date', str(timestamp))
-        return notifications
+            try:
+                since = dateutil.parser.parse(since)
```
brianhelba: Should we use `girder.utility.parseTimestamp`? It seems to provide a little more robust handling of timezones (if they were included, as ISO-8601 allows).

girder/api/v1/notification.py
```diff
@@ -112,22 +113,25 @@ def streamGen():
...
+            except ValueError:
+                try:
+                    since = datetime.utcfromtimestamp(float(since))
```
brianhelba: All other parts of Girder serialize and expect to parse times in ISO-8601. The most source of the `since` parameter will be if an application extracts the most recent `updated` (which is returned as ISO-8601) from an old list of notifications.

Javascript can also easily create ISO-8601 representations of current or arbitrary times: [`new Date().toISOString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString).

So, why are we supporting epoch seconds in this one place (but not the rest of Girder) at all? I'd vote to simplify and expect everything as ISO format.

zachmullen: Fine by me.

girder/api/v1/notification.py
```diff
@@ -112,22 +112,13 @@ def streamGen():
     @autoDescribeRoute(
         Description('List notification events')
         .notes('This endpoint can be used for manual long-polling when '
-               'SSE support is disabled or otherwise unavailable.')
-        .param('since', 'Filter out events before this time stamp.',
-               dataType='integer', required=False)
+               'SSE support is disabled or otherwise unavailable. The events are always '
+               'returned in chronological order.')
+        .param('since', 'Filter out events before this date.', required=False, dataType='dateTime')
```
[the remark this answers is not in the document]
zachmullen: I think that's a fine idea, though it's out of scope for this PR.

test/test_notification.py
```diff
@@ -16,54 +16,43 @@
...
 import pytest
 from pytest_girder.assertions import assertStatus, assertStatusOk
-
 from girder.models.notification import Notification
 
-
-def assertApproximateTimestamp(time1, time2, delta=1):
-    __tracebackhide__ = True
-    assert abs(float(time1) - float(time2)) < delta
+OLD_TIME = datetime.datetime.utcnow() - datetime.timedelta(days=3)
```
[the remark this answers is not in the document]
brianhelba: At this point, I'm only curious about Pytest best practices.

zachmullen:
I'd like to get this merged today since it's reverting some behavior already in master. Anything blocking that?

zachmullen:
Thanks for the reviews!

PR #2771, opened 2018-08-03
Fix up vagrant and ansible

cjh1:
I ran into two problems:

- The pip ansible module seem to have been [broken.](https://github.com/ansible/ansible/issues/37912). Added a workaround suggested in the issue. 
- On older distros such as our 14.04 image pip complains when trying to update six and requests. I have added a step to use the --ignore-installed option to reinstall these packages. It not particularly nice as it done regardless of the distro, but it seem alot of work to determine exactly which targets have this problem.

### DIFF
--- a/devops/ansible/roles/girder/tasks/pip.yml
+++ b/devops/ansible/roles/girder/tasks/pip.yml
@@ -33,9 +33,8 @@
 
     - name: Install Girder and plugin requirements
       pip:
-        name: ".[plugins]"
+        name: "{{ girder_path }}[plugins]"
         extra_args: "-e"
-        chdir: "{{ girder_path }}"
         executable: "{{ girder_pip | default(omit) }}"
 
     - name: Find girder-install executable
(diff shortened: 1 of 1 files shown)


devops/ansible/roles/girder/tasks/pip.yml
```diff
@@ -9,6 +9,16 @@
...
+    - name: Force install of requests and six
+      pip:
+        name: "{{ item }}"
+        extra_args: "-I"
+        virtualenv: "{{ girder_virtualenv }}"
+        virtualenv_python: "{{ girder_python | default(ansible_python.executable) }}"
+      with_items:
+        - six
+        - requests
+
```
cjh1: I think this block is not actually needed as things seem to work in the virtualenv case.

danlamanna: Can you remove this block and replace the other `-I` with `--ignore-installed`?

cjh1: Sure

cjh1: Done

PR #2776, opened 2018-08-15
Include plugin version information.

### LINKED ISSUE #1444: Include plugins' git sha values when possible
When non-core plugins are added to Girder, it would be nice if their information included the gitinfo sha value if the plugin is in a git repo.

manthey:
If plugins are checked out from git, add their version information to girder-version.json.  This also adds the lastCommitTime to the available information.

This resolves #1444.


manthey:
This changes the version information to include plugin information, like so:
```
{
    "git": true,
    "SHA": "b0b1ae7a67ff5323829d4115f2eeda64079b4343",
    "shortSHA": "b0b1ae7",
    "lastCommitTime": "2018-08-15T15:13:23+0000",
    "date": "2018-08-15T15:48:44+0000",
    "apiVersion": "2.5.0",
    "plugins": {
        "large_image": {
            "SHA": "3ccc9627204c5417a3029e7abb9cfb5557622e55",
            "shortSHA": "3ccc962",
            "lastCommitTime": "2018-08-15T13:08:39+0000"
        }
    }
}
```

This has the virtue that you can tell the version of any git based plugin in girder from the `system/version` endpoint.

zachmullen:
I'm really hoping that the dependence of the `system/version` endpoint on a file built via the web client build process can go away altogether in the future; the backend ideally should not depend on the web client build, especially since it's not really showing the current git version, but rather the git version at the last time the web client was built. I'm hesitant to further entrench this practice, especially in light of the pip-installable-plugins changes @jbeezley is making.

How critical is this feature to your users @manthey ? If knowing the git version of installed plugins via the REST API is a requirement for your users, I would prefer to do the lookup on the server side at the time of the request to make sure the info is up to date. If your users are actually interested in just knowing that their web client code is built from a certain version, I would prefer not to go through the REST API for that information at all.

manthey:
This is a feature to make it easier as a developer to see if the user has actually deployed the expected versions or not.  In many ways, a server-side lookup would be preferable, as you can change the version that is installed while the server is running.

I recognize that this technique will go away with pip-installable-plugins.

Would it be preferred for the REST endpoint to make a series of git version calls for each plugin and for Girder itself?  Is there much downside to that (clearly it is slower, but we could cache the result since it shouldn't change unless Girder is restarted)?

zachmullen:
> Would it be preferred for the REST endpoint to make a series of git version calls for each plugin and for Girder itself? Is there much downside to that?

I think so. If we make it optional via a flag to that endpoint, we wouldn't even need to worry about the caching IMO.

manthey:
Okay.  I'll refactor this.

girder/utility/install.py
```diff
@@ -130,6 +132,41 @@ def _pipeOutputToProgress(proc, progress):
             proc.wait()
 
 
+def getGitVersions():
```
zachmullen: To channel @brianhelba here, I think we should make this symbol private.

manthey: Sure.

PR #2777, opened 2018-08-15
Sort imports when importing into a filesystem assetstore.

manthey:
Sort files and directories when importing them into a filesystem assetstore.  Try to import files first, as if we are importing to a collection or user this will throw an exception right away instead of possibly importing some of the directories and then failing.

By having a predictable sort order, if the task is interrupted due to a failure or restart it is more obvious how far it got.

### DIFF
--- a/girder/utility/filesystem_assetstore_adapter.py
+++ b/girder/utility/filesystem_assetstore_adapter.py
@@ -407,12 +407,17 @@ class FilesystemAssetstoreAdapter(AbstractAssetstoreAdapter):
                 listDir, params=params)
             return
 
+        # Sort listDir to process files before directories.  Files and
+        # directories are alphabetized as well.
+        listDir = [entry[-1] for entry in sorted([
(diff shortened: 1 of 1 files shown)


manthey:
There are two concerns here. One is failing quickly (or prohibiting in the first place) importing files into collections and users.  A user has complained that when there is a mix of files and folders that are imported, rather than failing right away, an arbitrary set of folders are imported and then it fails.

Regarding knowing how far importing had progress: we had a condition where a corrupt file was causing a third-party library to seg-fault, which restarted Girder.  It would have been much easier to diagnose if we knew what order the files were being imported.  The third-party library was used as part of parsing data from the file on import from the large_image plugin.

Instead, we could log what file is being imported.  If we logged at debug level than this wouldn't spam the logs in ordinary use.

zachmullen:
+1 for the logging solution, since in your case it was fundamentally about tracing what caused an error

manthey:
Closing this in favor of #2783 and #2784.

PR #2778, opened 2018-08-20
Add a setting to disable public access.

manthey:
Note that plugins with @access.public endpoints may be broken if public access is disabled.  For those endpoints, it may be necessary to change to `@access.public(alwaysPublic=True)`.

The pytest access have been refactored to ensure that they use a clean database, as the setting for public access now affects endpoints marked public.


brianhelba:
Can we discuss this at the next Girder meeting? I'd like to better understand the implications on the permission system.

manthey:
Closed in favor of #2797.

PR #2782, opened 2018-08-21
Fixes #2781. S3 SSE support

### LINKED ISSUE #2781: S3 assetstore support for encryption at rest
https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html

This would simply be a new option on S3 assetstores that would add the `x-amz-server-side-encryption` header on the appropriate requests if it was enabled.

zachmullen:
This will allow Girder to keep objects encrypted at rest when
using an S3 assetstore.


brianhelba:
This could be tested within `moto` (to assert that the correct flag is passed), but currently is not tested.

zachmullen:
I'd like to get this merged, @manthey since you've looked at this code recently could you provide a review?

manthey:
@zachmullen We now only are testing the with s3 encryption flag checked.  Should we test with it both on and off?  Otherwise, this looks good.

zachmullen:
Added regression test for SSE disabled

PR #2786, opened 2018-08-27
Pin to boto3 < 1.8

manthey:
Boto3 1.8 breaks our tests using moto.  See https://github.com/spulec/moto/issues/1793 .  We can unpin when moto is updated.

This also has to add boto-core as a requirement to allow moto to install.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,8 +59,8 @@ with open('README.rst') as f:
     readme = f.read()
 
 installReqs = [
-    'boto3>=1.7,<1.8',
-    'botocore<1.11.0',
+    'boto3>=1.7,<1.8',  # TODO: unpin once moto works with boto3>=1.8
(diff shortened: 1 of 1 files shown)


zachmullen:
Thanks for looking into this. Could you add a TODO comment inline to say "this should be removed once XXX is released"?

manthey:
@zachmullen Is that sufficient?

fxfitz:
@manthey Any idea when we could get a release for this?

manthey:
@fxfitz I've just pinned versions of boto3 and boto-core in our the girder project.  I think you want to comment on the actual moto issue.

fxfitz:
Oh woops! You're entirely right! My bad; thanks @manthey, can't believe I missed that.

PR #2787, opened 2018-08-27
Audit log cleanup

zachmullen:
I've just deployed the audit log plugin for the first time, and received a sysadmin request to be able to remove old log entries, so I've provided a script that can be run in a cron job to do so, which can filter them by type. Additionally, I've added the ability to decorate endpoints so that calls to them are not added to the audit log, and made the notification endpoints use it.


plugins/audit_logs/scripts/cleanup.py
```diff
@@ -0,0 +1,26 @@
...
+
+    girder shell --plugins=audit_logs cleanup.py -- --days=30 --types=rest.request
+"""
+
+import click
+import datetime
+from girder.plugins.audit_logs import Record
+
+
+@click.command()
```
danlamanna: Untested, but I think if this plugin were to have a `setup.py` it could use the `girder.cli_plugins` entrypoint and be invoked with something like `girder audit-logs cleanup`.

zachmullen: In a previous incarnation of this PR, I had a comment at the top of the script that said something like "this can be restructured after pip installable plugins lands", and I think that still holds; it'll be more enticing to move to that once we're forced to have a setup.py anyway.

plugins/audit_logs/girder_audit_logs/cleanup.py
```diff
@@ -0,0 +1,26 @@
...
+@click.option('--days', type=click.INT, help='How many days to preserve records', default=90)
+@click.option('--types', help='Which record types to remove as a comma separated list. If not '
+              'provided, removes all record types.')
+def cleanup(days, types):
+    filter = {'when': {'$lt': datetime.datetime.utcnow() - datetime.timedelta(days=days)}}
+
+    if types:
+        filter['type'] = {'$in': types.split(',')}
+
+    print('Deleted %d log entries.' % Record().collection.delete_many(filter).deleted_count)
```
waxlamp: What happens if the deletion operation fails?

girder/api/rest.py
```diff
@@ -582,15 +582,27 @@ def _handleValidationException(e):
     return val
 
 
+def disableAuditLog(fun):
```
danlamanna: I think I missed this in the original PR for the audit logs plugin. I thought everything was contained within a plugin, but it appears that some of the audit logging code is in Girder core?

zachmullen: The emission of log entries comes from core (since that's what we are interested in logging). The handling of those events is in the plugin.

plugins/audit_logs/girder_audit_logs/cleanup.py
zachmullen: An exception would be raised

danlamanna: Nitpick, maybe use `click.echo` http://click.palletsprojects.com/en/7.x/utils/#printing-to-stdout.

plugins/audit_logs/setup.py
```diff
@@ -46,6 +46,10 @@ def prerelease_local_scheme(version):
     entry_points={
         'girder.plugin': [
             'audit_logs = girder_audit_logs:AuditLogsPlugin'
+        ],
+        'girder.cli_plugins': [
+            'audit_logs_cleanup = girder_audit_logs.cleanup:cleanup',
```
danlamanna: Can these be hyphenated?

test/audit_logs/test_audit_log.py
```diff
@@ -100,13 +95,36 @@ def testDownloadLogging(server, recordModel, freshLog, admin, fsAssetstore):
...
+
+@pytest.mark.plugin('audit_logs')
+@pytest.mark.parametrize('args,expected', [
+    ([], 0),
+    (['--days=0'], 4),
+    (['--days=0', '--types=rest.request'], 1),
+    (['--days=0', '--types=document.create'], 3)
+])
+def testCleanupScript(server, freshLog, args, expected):
+    user = User().createUser('admin', 'password', 'first', 'last', 'a@a.com')
```
danlamanna: Can this use the `user` fixture?

plugins/audit_logs/girder_audit_logs/cleanup.py
zachmullen: I love this part of the documentation for that function:

> and it will never fail (except in Python 3

zachmullen:
@danlamanna I think I addressed everything

plugins/audit_logs/girder_audit_logs/cleanup.py
```diff
@@ -0,0 +1,26 @@
...
+
+    girder audit_logs_cleanup --days=30 --types=rest.request
+"""
+
+import click
+import datetime
+from girder_audit_logs import Record
+
+
+@click.command(name='audit_logs_cleanup')
```
danlamanna: Should this and the docstring be changed as well? (for both files)

zachmullen: Fixed

PR #2788, opened 2018-08-30
Change permissions required for quota endpoints

zachmullen:
@brianhelba PTAL, this is scoped down from part of https://github.com/girder/girder/pull/2779

### DIFF
--- a/plugins/user_quota/server/quota.py
+++ b/plugins/user_quota/server/quota.py
@@ -174,37 +174,31 @@ class QuotaPolicy(Resource):
                            dictionary as if the JSON was already decoded.
         :returns: a validate policy dictionary.
         """
-        validKeys = []
-        for key in dir(self):
-            if key.startswith('_validate_'):
-                validKeys.append(key.split('_validate_', 1)[1])
-        for key in list(policy):
-            if key.startswith('_'):
-                del policy[key]
+        validKeys = [k[10:] for k in dir(self) if k.startswith('_validate_')]
+        policy = {k: v for k, v in six.iteritems(policy) if not k.startswith('_')}
+
         for key in policy:
             if key not in validKeys:
                 raise RestException(
                     '%s is not a valid quota policy key.  Valid keys are %s.' %
                     (key, ', '.join(sorted(validKeys))))
-            funcName = '_validate_' + key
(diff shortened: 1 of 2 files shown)


brianhelba:
I'll take a look.

zachmullen:
@brianhelba any objections to this change?

plugins/user_quota/server/quota.py
```diff
@@ -174,63 +174,55 @@ def _validatePolicy(self, policy):
...
         """
-        validKeys = []
-        for key in dir(self):
-            if key.startswith('_validate_'):
-                validKeys.append(key.split('_validate_', 1)[1])
-        for key in list(policy):
-            if key.startswith('_'):
-                del policy[key]
+        validKeys = [k[10:] for k in dir(self) if k.startswith('_validate_')]
+        policy = {k: v for k, v in six.iteritems(policy) if not k.startswith('_')}
```
brianhelba: @manthey Can you provide any insight as to [why this check was added](https://github.com/girder/girder/commit/0bb4cee20be050c99b17450fff51c719ef388af8#diff-9e7feac579f760b14c6480f3d8477a83R184)? It'd be good to have a code comment here explaining it.

plugins/user_quota/server/quota.py
```diff
@@ -279,8 +271,11 @@ def _getBaseResource(self, model, resource):
                  type, either 'user' or 'collection'., and 'resource' is the
                  base resource document or the id of that document.
         """
-        if isinstance(resource, tuple(list(six.string_types) + [ObjectId])):
-            resource = self.model(model).load(id=resource, force=True)
+        if isinstance(resource, six.string_types + (ObjectId,)):
```
brianhelba: I'd like to see more strongly typed parameters here; i.e. `resource` should just always be a document, and it should be the responsibility of the caller to load it if necessary.

However, that's out of scope of this PR.

plugins/user_quota/server/quota.py
manthey: If I recall, we add _ prefixed values to resource dictionaries (with, for instance, `_currentFileSizeQuota`), but we don't want to write that back to the database, so this strips out those private values.

PR #2790, opened 2018-09-07
Make audit log separate from girder log

zachmullen:
Prior to this, everything sent to the audit log was also logged to the girder log because we registered the audit log as a sub-log of the girder log by virtue of their names. This change makes the audit log separate from the main girder log.

### DIFF
--- a/girder/__init__.py
+++ b/girder/__init__.py
@@ -226,6 +226,7 @@ def _setupLogger():
 
 logger = _setupLogger()
 auditLogger = logging.getLogger('girder_audit')
+auditLogger.setLevel(logging.INFO)
 
 
 def logStdoutStderr(force=False):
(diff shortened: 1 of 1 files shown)


danlamanna:
Could the same be achieved by setting `propagate` to `False`?

zachmullen:
@danlamanna presumably

zachmullen:
@manthey could you please re-add review, I had to explicitly set the log level of the audit log now that it is separated.

brianhelba:
@zachmullen @danlamanna I think it's a cleaner and better accepted pattern to manage with a single hierarchy of `girder.*` loggers, and set `propagate=False`. I think we should do this instead.

zachmullen:
@brianhelba why is that preferable? The fact that we would have to add `propagate=False` indicates that we do not want the audit logger to be treated as a child of the main logger, which reflects the intent of them being two separate channels, not a parent-child relationship.

PR #2793, opened 2018-09-11
Scripting mode for girder-shell

zachmullen:
This allows a script to be passed into the `girder-shell` or
`girder shell` command which will execute with the same context
as though its code were run inside the interactive shell. Namely,
its `__name__` will be `'__main__'`, it will have the `appconf`
and `webroot` globals set, and the necessary plugins loaded. Its
`sys.argv` will be whatever was passed after `--` of the `girder-shell`
invocation, e.g.

    girder shell --plugins=audit_logs ./audit_report.py -- --format=csv

There may exist some `click` cleverness to make the `--` not necessary, but I wasn't able to find it if so.

### DIFF
--- a/girder/cli/shell.py
+++ b/girder/cli/shell.py
@@ -19,6 +19,8 @@
 
 import click
 import girder
+import six
+import sys
 
 from girder.utility.server import configureServer
 
@@ -42,15 +44,24 @@ def _launchShell(context):
         return code.interact(banner=header, local=context)
 
 
-@click.command('shell', short_help='Run a Girder shell.', help='Run a Girder shell.')
+@click.command('shell', short_help='Run a Girder shell.', help='Run an interactive Girder shell '
+               'or a script in the Girder environment.')
 @click.option('--plugins', default=None, help='Comma separated list of plugins to import.')
-def main(plugins):
+@click.argument('script', type=click.Path(exists=True, dir_okay=False), required=False)
+@click.argument('args', nargs=-1, required=False)
+def main(plugins, script, args):
     if plugins is not None:
         plugins = plugins.split(',')
 
     webroot, appconf = configureServer(plugins=plugins)
 
-    _launchShell({
-        'webroot': webroot,
-        'appconf': appconf
-    })
+    if script is None:
+        _launchShell({
+            'webroot': webroot,
+            'appconf': appconf
+        })
+    else:
+        globals_ = {k: v for k, v in six.viewitems(globals()) if k != '__name__'}
+        sys.argv = [script] + list(args)
+        six.exec_(open(script, 'rb').read(), dict(
+            webroot=webroot, appconf=appconf, __name__='__main__', **globals_))
--- a/girder/cli/shell.py
+++ b/girder/cli/shell.py
@@ -61,7 +61,8 @@ def main(plugins, script, args):
             'appconf': appconf
         })
     else:
-        globals_ = {k: v for k, v in six.viewitems(globals()) if k != '__name__'}
+        globals_ = {k: v for k, v in six.viewitems(globals()) if k not in {'__file__', '__name__'}}
         sys.argv = [script] + list(args)
         six.exec_(open(script, 'rb').read(), dict(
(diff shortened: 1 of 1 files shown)


zachmullen:
As a demonstration, I converted https://github.com/girder/girder/pull/2787 to use this new mode.

zachmullen:
ping @subdavis

girder/cli/shell.py
```diff
@@ -42,15 +44,24 @@ def _launchShell(context):
         return code.interact(banner=header, local=context)
 
 
-@click.command('shell', short_help='Run a Girder shell.', help='Run a Girder shell.')
+@click.command('shell', short_help='Run a Girder shell.', help='Run an interactive Girder shell '
+               'or a script in the Girder environment.')
 @click.option('--plugins', default=None, help='Comma separated list of plugins to import.')
-def main(plugins):
+@click.argument('script', type=click.Path(exists=True, dir_okay=False), required=False)
```
danlamanna: If you want, this could be `click.File` which would pass an open file handle.

girder/cli/shell.py
```diff
@@ -42,15 +44,24 @@ def _launchShell(context):
...
-        'webroot': webroot,
-        'appconf': appconf
-    })
+    if script is None:
+        _launchShell({
+            'webroot': webroot,
+            'appconf': appconf
+        })
+    else:
+        globals_ = {k: v for k, v in six.viewitems(globals()) if k != '__name__'}
```
danlamanna: Why does every global have to be involved?

manthey: To exec a file in a similar manner as `python <file>`, you have to pass the globals.

Pedantically, this is fine *but* won't work for a file that uses the multiprocessing module.  Multiprocessing is a special case where you actually have to mutate the `globals()` object rather than copy it to a new dictionary.

For that purpose, this could be
```
    exec_ = six.exec_
    globals_ = globals()
    originalGlobals = globals_.copy()
    # if we don't want the receiving file to already have our imports, we may want to remove any key from globals_ that doesn't start with _
    globals_.update({
        '__name__': '__main__', 
        '__file__': script,
        'webroot': webroot,
        'appconf': appconf,
    })
    exec_(open(script, 'rb').read(), globals_)
    globals._clear()
    globals_.update(originalGlobals)

File: girder/cli/shell.py [hunk 3]
```diff
@@ -42,15 +44,24 @@ def _launchShell(context):
...
+        sys.argv = [script] + list(args)
+        six.exec_(open(script, 'rb').read(), dict(
+            webroot=webroot, appconf=appconf, __name__='__main__', **globals_))
```
manthey: You probably also want to specified `'__file__': script`.

girder/cli/shell.py
manthey: See https://github.com/manthey/pyexe/blob/master/pyexe.py#L76-L114 for even more thorough support (python files may be zip archives, for instance).

girder/cli/shell.py
zachmullen: I did that at first, but it turns out we need the actual path inside the function.

girder/cli/shell.py
zachmullen: > Pedantically, this is fine but won't work for a file that uses the multiprocessing module. Multiprocessing is a special case where you actually have to mutate the globals() object rather than copy it to a new dictionary.

Is this documented somewhere? It seems like a bug

manthey: I think it has to do with this code https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Lib/multiprocessing/__init__.py#L22.  A contributor to the Octave project made me aware of it.  I wasn't able to find  an issue in the Python issue tracker (but there very well could be one).

zachmullen:
@manthey I think I'd like to defer more obscure support (zip files & multiprocessing idiosyncrasies) to future PRs, if they are not show stoppers in your opinion.

manthey:
The addition of __file__ was all I really cared about.  The rest is obscure functionality that most people won't care about.  I'm fine with this.

Issue #2794, opened 2018-09-12
Coverage is broken


zachmullen:
Not sure if this is a codecov issue or a local issue, but it seems likely that it's local.

manthey:
I see that the coverage file sent to codecov on on a successful coverage branch from late August was [600kb](https://circleci.com/gh/girder/girder/12853), whereas a recent submission from the same step was only [450kb](https://circleci.com/gh/girder/girder/13190).

manthey:
In the circleci environment, the last working coverage used pytest 3.7.3 and pytest-cov 2.5.1.  The broken coverage has pytest 3.8.0 and pytest-cov 2.6.0.  A crude test of pinning pytest <3.8 and pytest-cov <2.6 makes coverage work again.

zachmullen:
Good to know, at least we can pin those temporarily until we can sort out what's really wrong.

manthey:
I see that working versions store coverage in a file called something like `girder/build/test/coverage/python_temp/.coverage.692a993cea2b.3978.292632`, whereas the newer version store coverage in `girder/build/test/coverage/python_temp/.coverage`.

manthey:
I'll check if we have to pin pytest (I assume we have to pin pytest-cov).  Pinning both definitely works.

manthey:
Outdated.

PR #2795, opened 2018-09-12
Fix coverage by pinning pytest-cov.

manthey:


manthey:
Fixes #2794, but we should probably figure out how to use the latest pytest-cov.

PR #2797, opened 2018-09-13
Add a plugin that, when enabled, disables public endpoints.

manthey:
When enabled, this makes most public endpoints into user endpoints.

This also alters the user find with permissions methods to only show users that the current user has write-access to, effectively hiding most users from non-admin users.

Users can still leak -- for instance groups will show user names.

This replaces #2778.


manthey:
Non-admin's won't be able to find and add other users to groups.

It would need some additional work to allow exact name matches to produce work or to accept name/email addresses/ids directly without searching for the matching user.

jeffbaumes:
Ok, yes it's a balance between discoverability/collaboration and privacy. This plugin as is might not be the perfect happy medium, but that's why it's a plugin. In the future, we could add more options to this plugin or make alternate plugins with different behavior if needed.

manthey:
Yes.  Also, the user-listing restrictions could be separated from disabling public access to endpoints, which would either introduce settings or split this into multiple plugins.

manthey:
This doesn't belong in core, so closing this.

PR #2798, opened 2018-09-13
Restore some scripts that were deleted.

manthey:
The install_cmake.sh and install_mongo.sh are convenience scripts for downstream CI.

These were removed as part of PR #2222.

Perhaps they shouldn't exist in Girder's repo, but they are being used in CI by both large_image and HistomicsTK.

### DIFF
--- a/scripts/install_cmake.sh
+++ b/scripts/install_cmake.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+PREFIX="$CACHE/cmake-$CMAKE_VERSION"
+if [[ ! -f "$PREFIX/bin/cmake" || -n "$UPDATE_CACHE" ]] ; then
+    rm -fr "$PREFIX"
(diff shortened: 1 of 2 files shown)


manthey:
@brianhelba Can we put these back for now, and remove them once the Girder 3 merge window opens?

brianhelba:
I'll copy my comment from the other PR:

> I'd suggest copying the scripts to downstreams, if it's required that they install that software outside of a package manager. Girder's supported deployment methods (Ansible roles, Docker images, documented install environments) don't include systems that would need to use such scripts, so I don't think Girder should be responsible for maintaining them, particularly as we transition away from cloning the Git repo for deployments.

That said, I'm fine to wait until 3 to remove these. So we don't forget, can you open up another PR / issue (tagged as 3) to revert this one and re-remove the scripts?

manthey:
@brianhelba Thanks.

PR #2800, opened 2018-09-13
Use the latest pre-built CTest in Python 3 testing

brianhelba:
The version added by #2222 was an older version and a source tarball, so binaries were not available.


brianhelba:
Since this is a change to the testing Docker image, it will only be built into `latest` once this branch is merged into `master`.

So, we'll have to merge it before it actually fixes CI testing. :roll_eyes:  However, I've tested this locally, and it builds properly and lets you run `cmake` now.

brianhelba:
Once this build finishes: https://hub.docker.com/r/girder/girder_test/builds/br4fbrjwu6suyr2uhtkhmdq/ , CI should work again.

PR #2801, opened 2018-09-14
Decouple the 2.x-maintenance branch from master

jbeezley:


brianhelba:
We need to think about documentation. The policy on RTD has always been that the main docs track the bleeding edge master, and I think it's fine if that continues. However, we should probably also configure a 'v2' docs to track `2.x-maintenance` too.

@girder/developers Thoughts?

jbeezley:
There is a "stable" tag on RTD, which automatically tracks the latest release.  I could change the default (redirect from '/') to that.

PR #2802, opened 2018-09-18
Pin fusepy to < 3.0 to prevent deadlocks in tests

jbeezley:
This fixes a deadlock in [py3_integrationTests](https://circleci.com/gh/girder/girder/13463).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -114,7 +114,7 @@ extrasReqs['sftp'] = [
     'paramiko',
 ]
 extrasReqs['mount'] = [
-    'fusepy>=2.0.4',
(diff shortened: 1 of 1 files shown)


zachmullen:
Is the underlying issue a fusepy bug or something we are doing wrong? Does it only affect the tests, or would a deadlock be possible in the fuse plugin runtime?

jbeezley:
I don't know.  I looked at the diff in fusepy.  It's mostly a bunch of formatting changes.  There aren't any changelogs indicating what backwards incompatible things happened.  Hopefully @manthey can look at it when he gets back.

PR #2807, opened 2018-09-20
Fix ansible role galaxy deployment on circle

jbeezley:
The ansible role deployment failed after #2552 was merged.  This should fix that; however as we discovered yesterday, there are backwards incompatible changes in the ansible role as well.  I don't know if anything can/should be done to the deployment to handle that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -466,6 +466,9 @@ jobs:
           path: girder
       # Since this cache must have been created in a previous step of this workflow, we can get
       # a virtualenv without having to create one or install anything.
+      - run:
+          name: Generate python environment checksum file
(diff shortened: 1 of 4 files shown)


jbeezley:
I'm attempting to push a compatibility change to the ansible role.  Something like current code might work, but I'm still testing.  I'm hoping others can tell me if this is worth doing or not.

zachmullen:
Let me know once you want this merged and I'll approve.

jbeezley:
I was able to get this to work with `girder_version = 2.x-maintenance`.  I did encounter a botocore version conflict when trying to start up the deployed version, but that is unrelated I think.  It would be best if someone with a real use case could try it.   Failing that, we can just merge and hope for the best sense it's broken now anyway.

PR #2810, opened 2018-09-27
Make event bindings with the same handler idempotent

zachmullen:
This makes it so that handler strings corresponding to event bindings are unique. Binding to the same event twice with the same handler previously would make the handler be invoked twice. After this change, it will instead overwrite the previous handler function, only making a single call, and emit a warning to the log. This will also cause the new binding to move to the end of the handler list for the given event.

@danlamanna PTAL

### DIFF
--- a/girder/events.py
+++ b/girder/events.py
@@ -43,8 +43,10 @@ receive the Event object as its only argument.
 
 import contextlib
 import girder
+import six
 import threading
 
+from collections import OrderedDict
 from girder.utility import config
 from six.moves import queue
 
@@ -225,16 +227,15 @@ def bind(eventName, handlerName, handler):
     :type handler: function
     """
     if eventName in _deprecated:
-        girder.logger.warning('event "%s" is deprecated; %s' %
-                              (eventName, _deprecated[eventName]))
+        girder.logger.warning('event "%s" is deprecated; %s' % (eventName, _deprecated[eventName]))
 
     if eventName not in _mapping:
-        _mapping[eventName] = []
+        _mapping[eventName] = OrderedDict()
 
-    _mapping[eventName].append({
-        'name': handlerName,
-        'handler': handler
-    })
+    if handlerName in _mapping[eventName]:
+        girder.logger.warning('Event binding already exists: %s -> %s' % (eventName, handlerName))
+    else:
+        _mapping[eventName][handlerName] = handler
 
 
 def unbind(eventName, handlerName):
@@ -246,13 +247,7 @@ def unbind(eventName, handlerName):
     :param handlerName: The name that identifies the handler calling bind().
     :type handlerName: str
     """
-    if eventName not in _mapping:
-        return
-
-    for handler in _mapping[eventName]:
-        if handler['name'] == handlerName:
-            _mapping[eventName].remove(handler)
-            break
+    _mapping.get(eventName, {}).pop(handlerName, None)
 
 
 def unbindAll():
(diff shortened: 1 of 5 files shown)


girder/events.py
```diff
@@ -225,16 +227,17 @@ def bind(eventName, handlerName, handler):
...
+        _mapping[eventName] = OrderedDict()
 
-    _mapping[eventName].append({
-        'name': handlerName,
-        'handler': handler
-    })
+    if handlerName in _mapping:
+        # We delete from the OrderedDict to move the handler to the end of the ordering
+        del _mapping[handlerName]
+        girder.logger.warning('Overwriting event binding: %s -> %s' % (eventName, handlerName))
```
danlamanna: Doesn't this mean that a plugin could replace a core binding without intending to do so? Should subsequent calls to bind instead no-op and print the warning that they aren't being used?

zachmullen: I'm fine with either behavior, I'll make it a no-op unless someone prefers it this way.

zachmullen: After thinking about it some more, I actually prefer the behavior I had written at first, because it is more in line with how a `dict` works, and it also has additional power, in that it would allow plugin authors to intentionally overwrite event handlers from either core or their upstream plugins.

zachmullen:
@girder/developers PTAL, this is a policy change that should be reviewed by multiple stakeholders.

girder/events.py
```diff
@@ -223,18 +225,23 @@ def bind(eventName, handlerName, handler):
...
 
-    _mapping[eventName].append({
-        'name': handlerName,
-        'handler': handler
-    })
+    if handlerName in _mapping[eventName]:
+        # We delete from the OrderedDict to move the handler to the end of the ordering
+        del _mapping[eventName][handlerName]
+        if not overwrite:
+            girder.logger.warning('Overwriting event binding: %s -> %s' % (eventName, handlerName))
```
jbeezley: I think it might be appropriate to raise an exception in this case because we are providing a mechanism to get around it.

jbeezley:
I can see an argument for simply not providing an API for overwriting event handlers, i.e. double binding would result in an exception.  Providing a API for it implicitly gives it our blessing.  Because it can be dangerous and/or cause weird/difficult to understand bugs that depend on `iter_entry_points` order, I'm not sure that we should.

We can't prevent plugin authors from doing stupid and dangerous things.  They could simply unbind the previous handler first if they really need to do it.  In fact, this is what they would have to do in Girder 2.  Is there an existing rather than theoretical use case for overriding handlers?

zachmullen:
The motivation for this is actually that event bindings were getting duplicated in our testing infrastructure. The way our fixtures are loaded and unloaded does not play nicely with the global nature of the event registry; simply calling `events.unbindAll()` in the teardown step doesn't work because then core events do not ever get re-bound.

When discussing candidate solutions to that issue, we realized this was a simple one, and @danlamanna and I both thought this idempotent behavior was actually superior to the old behavior independently of the fact that it happens to solve the testing issue.

zachmullen:
> Is there an existing rather than theoretical use case for overriding handlers?

This is only theoretical right now, though I can imagine some uses.

jbeezley:
> The motivation for this is actually that event bindings were getting duplicated in our testing infrastructure. The way our fixtures are loaded and unloaded does not play nicely with the global nature of the event registry; simply calling events.unbindAll() in the teardown step doesn't work because then core events do not ever get re-bound.

Doesn't this imply that we are okay with leaving in an internal warning message when a plugin inevitably rebinds itself in a test?

I understand this solves an immediate need, and I don't want to get in the way of that.  My main concern is that this is *new* behavior added as a workaround for testing.  Do we want to bless that behavior with a public API in the live server as well?

zachmullen:
> My main concern is that this is new behavior added as a workaround for testing. Do we want to bless that behavior with a public API in the live server as well?

While this was motivated by a testing workaround, I would not have made this PR if that was the only reason. I actually think this is a superior behavior in general. I'd like us to consider that question independently of the fact that it happens to make our testing development easier.

zachmullen:
> Doesn't this imply that we are okay with leaving in an internal warning message when a plugin inevitably rebinds itself in a test?

For the moment, yes. @danlamanna both agree that the long-term solution for the testing issue is to have our event registry bound to a context other than a module scoped variable.

jbeezley:
> I actually think this is a superior behavior in general.

I agree that silently double binding events is bad, so I'm fine with changing this.  I would be find with one of the following alternatives:

1. Use @danlamanna's suggestion to keep the original handler and emit a warning.
2. Remove the `overwrite` keyword argument until we find a real use case for it.

zachmullen:
I'll do both of those things. Downstreams who want to overwrite can use `unbind` like you said.

zachmullen:
Please take a final look, I think this is ready.

PR #2811, opened 2018-09-27
RFC: Always remove node_modules after building the client

jbeezley:
There is a problem with how we do the web client build that manifests itself when trying to uninstall Girder.  Try running the following:
```
pip install --pre girder
girder build
pip uninstall girder
```
The last command will take several minutes.  This is caused because the `node_modules` directory created by `girder build` is not tracked by pip, and it lists every non-tracked file as a warning.

I wanted to see what others thought about this.  The solution in this PR just removes `node_modules` on every build... This mean that every `girder build` command needs to regenerate it from scratch.  I just ran a test and `npm install` from scratch is about 1 minute vs. 6 seconds the second time.  This extra time includes the `--watch` commands which used to be nearly instant.

Possible other solutions:
1.  Only delete `node_modules` after a production build.
2.  Leave it as is and tell people to use `pip uninstall -y girder`, which is fast but still leaves the untracked files around.
3.  Move `node_modules` elsewhere and set the `--prefix` flag (and maybe `NODE_PATH` environment vars).  This option is somewhat likely to cause problems with edge cases in module path resolution.
4.  Use a staging area outside of the Girder python package.  This is what I wanted to do originally, but it caused too many problems.  See my commit message in 517f7e668c0398ece3bebc7c13adfbe43ddef50f for some details about this.

### DIFF
--- a/girder/cli/build.py
+++ b/girder/cli/build.py
@@ -74,7 +74,8 @@ def main(dev, watch, watch_plugin):
     try:
         check_call(buildCommand, cwd=staging)
     finally:
-        shutil.rmtree(os.path.join(staging, 'node_modules'), ignore_errors=True)
+        if not dev:
+            shutil.rmtree(os.path.join(staging, 'node_modules'), ignore_errors=True)
 
 
 def _collectPluginDependencies():
--- a/girder/cli/build.py
+++ b/girder/cli/build.py
@@ -21,6 +21,7 @@ import json
 import os
 from pkg_resources import resource_filename
 from subprocess import check_call
+import shutil
 import sys
 
 import click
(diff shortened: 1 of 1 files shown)


jeffbaumes:
I wonder how JupyterLab resolves this problem.

jbeezley:
Looks like jupyterlab basically does (4).  I'm still trying to figure out how (or if) they handle in place builds for development.   There is a lot of [custom code](https://github.com/jupyterlab/jupyterlab) that mucks around with setuptools internals that I would be afraid to touch.  One thing I noticed that is odd is that they [vendor yarn](https://github.com/jupyterlab/jupyterlab/tree/master/jupyterlab/staging) for `jupyter lab build`.

zachmullen:
Looks like phantomjs is hanging in CI, are you seeing that locally as well?

zachmullen:
What's the verdict on this PR? If we consider this a temporary / hacked solution that we want to replace, there is some risk in merging it that it will exist longer than we'd like. It speeds up performance at uninstall time, but (if I understand correctly) slows down performance at build time, which may not be a desirable tradeoff.

Is option 4 intractable, or just a lot of work? Are we waiting on some other changes that will make this problem moot?

jbeezley:
In my last change, I switched to option (1) because we need `node_modules` in place when running tests.  In development mode, there isn't much difference from before.  We have had to tell users to "delete node_modules" to fix build issues in the past, so maybe automatically doing so in production isn't the worst idea.

> What's the verdict on this PR? 

I'm still undecided; I wanted to submit this to let people know of the problem and get feedback.  I don't think the issue is big enough that we need to worry about it immediately.

> Is option 4 intractable, or just a lot of work?

It's certainly possible.  My impression from looking at jupyter lab is that it would be a bigger maintenance burden than we want to take on.

PR #2813, opened 2018-10-02
Update to fusepy 3.x.

manthey:
fusepy 3 added support for ioctls, but this doesn't work in Python 3 without extra work.  Since we don't need to handle them, explicitly don't handle them.

Also, fusepy 3 has changed an internal time format for consistency with the base fuse library.  Before it coerced float epochs itself; that is now deprecated and it prefers nanosecond epochs instead.

Reference #2802.

### DIFF
--- a/girder/cli/mount.py
+++ b/girder/cli/mount.py
@@ -49,6 +49,8 @@ class ServerFuse(fuse.Operations):
     extended to expose metadata and other resources by extending the available
     paths.  Files can also be reached via a path shortcut of /file/<id>.
     """
+    use_ns = True
+
     def __init__(self, stat=None):
         """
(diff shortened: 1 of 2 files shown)


girder/cli/mount.py
```diff
@@ -49,6 +49,8 @@ class ServerFuse(fuse.Operations):
     extended to expose metadata and other resources by extending the available
     paths.  Files can also be reached via a path shortcut of /file/<id>.
     """
+    use_ns = True
```
zachmullen: Should this be grouped with the other attributes below?

manthey: Conceptually they are different.  The other are specifically file system methods we don't want to handle.  This is a flag for fusepy's Operations class.

PR #2447, opened 2017-10-27
WIP: Add a script to get Girder public Python symbols

brianhelba:
Based on work by @danlamanna in https://github.com/girder/girder/commit/da3bb202ec5c9093400818be3dc342125504ac7b

PR #2814, opened 2018-10-02
Require via CI commits to publicNames.txt when public symbols are introduced.

kotfic:
This PR replaces #2447  and introduces a new CI job "public_symbols"  The public_symbols job fails if public symbols have been introduced that are not committed to the `scripts/publicNames.txt` file.  To update scripts/publicNames.txt run:

```sh
python scripts/publicNames.py > scripts/publicNames.txt
```

from the root of your local repository.  Committing any changes introduced will cause the CI test to pass.  Effectively this means all PRs will include a list of the public symbols they introduced (or removed). 

It is hoped this will encourage contributors and reviewers to think more critically about the public symbols that are being introduced. 

Note that I have intentionally left the symbols table unaltered so reviewers can see what the error output of the job looks like. If approved the following additional steps will be necessary:

1. Update scripts/publicNames.txt
2. push `girder/ctags:latest` and update the image used for the `public_symbols` job from `kotfic/ctags:latest`
3. Close #2447


zachmullen:
We'll want to add something to CONTRIBUTING.md about this too.

kotfic:
This will rely on https://github.com/girder/girder/pull/2827 going in and switching the container before being merged

scripts/publicNames.py
```diff
@@ -6,7 +6,7 @@
 
 Tree = lambda: collections.defaultdict(Tree)
 
-EXCLUDE_DIRS = ['test', 'tests', 'plugin_tests', '.*egg-info']
+EXCLUDE_DIRS = ['test', 'tests', 'plugin_tests', '.*egg-info', 'web_client']
```
zachmullen: Rather than blacklisting, can we just not add any symbol that isn't part of a python package? That way things like `mail_templates` or whatever other package_data doesn't get caught up in this list.

kotfic:
@zachmullen can you take a look at the new publicNames.txt and see if this meeting your expectations?  This should only index folders that have at least one python file in them. We still need to blacklist folders which have python files we don't want to add (e.g. tests,  docs) and files we don't want to add (e.g. setup.py)

scripts/publicNames.txt
```diff
@@ -0,0 +1,1452 @@
...
+                GirderCli
+                    sendRestRequest
+                help
+                main
+devops
+    ansible
+        roles
+            girder
+                library
+                    girder
```
zachmullen: Are these python symbols part of our API surface?

CONTRIBUTING.rst
```diff
@@ -56,12 +56,49 @@ These results will list any failed tests. Coverage reports and any screenshots
 from failed web client tests will be attached to the build as artifact files. You can reach your
 build by clicking the build status link on your GitHub PR.
 
+Confusing failing test message "AttributeError: 'module' object has no attribute 'x_test'"
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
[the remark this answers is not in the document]
kotfic: sorry, looks like a bad rebase conflict

scripts/publicNames.txt
kotfic: do you mean the girder ansible library or specifically the roles/girder/library symbols?

kotfic:
Note: once this looks good we'll need to merge https://github.com/girder/girder/pull/2827 and update the reference to `kotfic/girder_test:latest-py3` in the CI yaml

scripts/publicNames.txt
zachmullen: The latter. The documented API is the YAML interface that it exposes to ansible users, just wondering whether stability of the python symbols inside the library actually matters to us.

kotfic: no i don't think it matters to anyone,  i'll remove it

CONTRIBUTING.rst
```diff
@@ -56,12 +56,25 @@ These results will list any failed tests. Coverage reports and any screenshots
...
+Tracking Public Symbols
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Public symbol addition and removal is tracked in Girder through the ``scripts/publicNames.txt`` file
+and any PR that adds new public symbols must add those symbols to this file. This is done by running
+the following script::
+
+    python scripts/publicNames.py > scripts/publicNames.txt
+
+Any changes to the file should be committed as apart of the PR or not all CI tests will pass.
```
zachmullen: apart -> a part

CONTRIBUTING.rst
```diff
@@ -56,12 +56,25 @@ These results will list any failed tests. Coverage reports and any screenshots
 from failed web client tests will be attached to the build as artifact files. You can reach your
 build by clicking the build status link on your GitHub PR.
 
+Tracking Public Symbols
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Public symbol addition and removal is tracked in Girder through the ``scripts/publicNames.txt`` file
```
zachmullen: I would start with

"Adding new public symbols to Girder's python library should only be done intentionally, as doing so increases the surface of the API and introduces a maintenance burden. Public symbols are packages, modules, and symbols within those modules that do not start with an underscore character. To help with this goal, public symbol addition and removal is tracked automatically as part of our CI process, with the full list of symbols residing in ``scripts/publicNames.txt``."

scripts/test_names.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
manthey: Do we want this to be `#!/usr/bin/env bash` instead for slightly more cross-platform compliance?  And then, ideally this script would be marked executable and the tox.ini command wouldn't need to specify `/bin/bash`

manthey:
This fails for me in my regular build environment because I create my python virtualenv within the girder repo directory (via `virtualenv env`), and it tries to parse python files in that directory.  I wonder if whitelisting the directories would be better than blacklisting particular directories.

kotfic:
My fear is explicitly whitelisting all directories will led to whole areas of the API being left out by accident as new or different packages and modules are added.  Is there a way we can support your personal local workflow without having to add _another_ level of white_listing on top of the current black_listing?

manthey:
I'd love to see if something is `.gitignore`d that it wouldn't include it (i.e., don't include any file listed by `git check-ignore **/*`).

I'm not advocating, but the blacklist regex can be abused to be a white list, e.g., adding `'^(?!(clients|girder|plugins))'` as an entry will work.

kotfic:
I don't hate using the regex to only look at certain top-level folders. With a comment i think its fine actually.

@danlamanna is pytest_girder apart of the public API?  we're not including tests,  but pytest_girder is used by downstream plugins right?

manthey:
I like the addition of pytest fixtures to this list, as removing or changing them will be a more obvious change, but if others feel differently, I'm okay with not doing so.

kotfic:
@brianhelba I'm happy to make changes if there are things you'd like to see fixed once you get a chance to take a look at this.  For now i'm going to merge it

PR #2815, opened 2018-10-02
Set Access-Control-Expose-Headers on cross origin requests

zachmullen:


zachmullen:
@girder/developers PTAL, we haven't had a need to set this until now since we have a non-simple header that clients will want to inspect on XHR responses. I didn't bother adding it to the web client UI, I doubt anyone will want to change the default value anytime soon, and if they do they can use the REST API.

manthey:
Good.  I'd like to see this in the 2.x-maintenance branch as well.

zachmullen:
@manthey is `Girder-Total-Count` on 2.x-maintenance?

manthey:
@zachmullen Yes, it was the last PR merged before we decoupled from master.

zachmullen:
@manthey if you mark approval, I will merge this.

Issue #2613, opened 2018-02-14
Rename async property of events daemon


danlamanna:
In Python 3.7, `async` will become a reserved keyword and Girder will not start due to [this usage](https://github.com/girder/girder/blob/31fee2e47c8b771f34692dee9efb358ce23c8151/girder/events.py#L73). We should probably rename to `async_` or something similar.

zachmullen:
Good find. We should make sure to address this for 3.0

jeffbaumes:
This was fixed by #2817.

Issue #946, opened 2015-09-05
Add a mechanism to migrate the database


brianhelba:
As Improvements and bugfixes continue to be made to the Model layer, the database may occasionally need to be updated too. Rather than have lots of inline checks for legacy database formats, it would be nice to have a system for one-time migration operations that run against the database.

This differs from the existing system consistency check in that migration would address very specific issues that we know need to be changed, would be run only once per issue, and would be mandatory to run before a new version of Girder is deployed.

To enable this, we may need to add a database version field to the database (perhaps as a hidden Setting). When Girder starts up, it will check the database version against the current value in the migration mechanism; if the database is older than the latest migration, than all migrations between the database version and the code version are run once.

For safety, each migration should be idempotent, but in theory they will only be run once. Additionally, migrations should generally be written using raw Pymongo operations, since we don't know which version of the model layer will be present when a migration is finally run.

It is possible that for large databases, migrations will take some time, so the proposed mechanism of running them automatically on startup will need to be well-documented, with migrations only being added on minor releases (major releases is probably too slow, since migrations could be bugfixes, that don't come with a new API). Alternatively, we could try to follow Django's typical workflow ( https://docs.djangoproject.com/en/1.8/topics/migrations/ ) and run them manually, but this would necessitate a lot more compatibility code and defeat a large part of the reason for having migrations.

brianhelba:
Some sample PRs which should have accompanying migrations:
* https://github.com/girder/girder/pull/810
* https://github.com/girder/girder/pull/945

zachmullen:
:+1: , this should be pretty straightforward to achieve.

sebastienbarre:
Feel free to hit me up if you decide to discuss it in a meeting. We have migrated the KWiK database schema 226 (!) times as of today, and most often live during the day while you guys are using the app.

PR #2817, opened 2018-10-03
Alias async to async_ to allow 3.7 compatibility

giesberge:
In 3.7 async and await are protected keywords.
Renamed async to asynq and added an alias so changes shouldn't be breaking for users using <3.7

### DIFF
--- a/girder/events.py
+++ b/girder/events.py
@@ -315,7 +315,7 @@ def trigger(eventName, info=None, pre=None, async_=False, daemon=False):
     :param daemon: Whether this was triggered via ``girder.events.daemon``.
     :type daemon: bool
     """
-    e = Event(eventName, info, async=async)
+    e = Event(eventName, info, async_=async_)
     for name, handler in six.viewitems(_mapping.get(eventName, {})):
         if daemon and not async_:
             girder.logprint.warning(
--- a/girder/events.py
+++ b/girder/events.py
@@ -45,10 +45,25 @@ import contextlib
 import girder
 import six
 import threading
+from functools import wraps
 
 from collections import OrderedDict
 from girder.utility import config
 from six.moves import queue
+
+def deprecated_async(func):
+    """A decorator, that let's us keep our old API, but deprecate it"""
+    @wraps(func)
+    def inner(*args, **kwargs):
+        if 'async' in kwargs:
+            if 'asynq' in kwargs:
+                raise ValueError('cannot use both async and asynq '
+                                 'keyword arguments! the latter obsoletes the first.')
+            warnings.warn('async keyword argumnt is deprecated, '
+                          'use asynq instead', DeprecationWarning)
+            kwargs['asynchronous'] = kwargs.pop('async')
+        return func(*args, **kwargs)
+    return inner
 
 
 class Event(object):
@@ -71,7 +86,8 @@ class Event(object):
         'currentHandlerName'
     )
 
-    def __init__(self, name, info, async=False):
+    @deprecated_async
+    def __init__(self, name, info, asynq=False):
         self.name = name
         self.info = info
         self.propagate = True
@@ -274,7 +290,7 @@ def bound(eventName, handlerName, handler):
     finally:
         unbind(eventName, handlerName)
 
-
+@deprecated_async
 def trigger(eventName, info=None, pre=None, async=False, daemon=False):
     """
     Fire an event with the given name. All listeners bound on that name will be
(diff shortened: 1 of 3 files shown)


giesberge:
Need to do some small work, development environment is screwy on current computer

plugins/jobs/server/models/job.py
```diff
@@ -264,7 +264,7 @@ def createJob(self, title, type, args=(), kwargs=None, user=None, when=None,
             'log': [],
             'meta': {},
             'handler': handler,
-            'async': async,
```
danlamanna: What do @girder/developers think about changing the schema here? It is more consistent, but since Girder currently lacks any notion of database migrations (#946) - this could cause issues in the future for existing sites.

girder/events.py
```diff
@@ -44,10 +44,26 @@
...
+    """A decorator, that let's us keep our old API, but deprecate it"""
+    @wraps(func)
+    def inner(*args, **kwargs):
+        if 'async' in kwargs:
+            if 'asynq' in kwargs:
+                raise ValueError('cannot use both async and asynq '
+                                 'keyword arguments! the latter obsoletes the first.')
+            warnings.warn('async keyword argumnt is deprecated, '
+                          'use asynq instead', DeprecationWarning)
+            kwargs['asynchronous'] = kwargs.pop('async')
```
danlamanna: Is this supposed to be "asynchronous"?

giesberge: No my bad, updating now, still trying to get my dev environment up on Windows.
Also I decided on asynq since that what I saw in a few other fixes. Though official convention is async_
We can change it to whatever we think might work.

plugins/jobs/server/models/job.py
manthey: I think that the move to Girder 3 is a fine time to change this.  We could use `job.get('asynq', job.get('async'))` to read the old `async` key if no `asynq` key exists.

I'd mildly prefer `async_` to `asynq`, as I suspect I will always mistype `asynq`.

danlamanna:
fixes #2613

plugins/jobs/server/models/job.py
giesberge: Do we want to switch it over to async_ ? Or would another convention be better

zachmullen:
Excellent contribution, thanks @giesberge !

In addition to the deprecation warnings, we should add a note in the [migration guide](https://github.com/girder/girder/blob/master/docs/migration-guide.rst) about this general change, and a little bullet point enumeration of each of the places where it changed.

zachmullen:
Also, could you please change this to be against the master branch rather than 2.x-maintenance?

giesberge:
@zachmullen I'm just going to submit a new pull request and close this one just to make the merge cleaner if that works for you.

giesberge:
@zachmullen nevermind, rebased it and did a push --force

danlamanna:
A couple of notes:
- So it appears we're deprecating the usage of `async` in Girder 2 and 3, and will only remove the ability to pass `async` in Girder 4? Is that right? I think we should be able to make these sorts of breaking changes in master right now.
- We're changing the job document format to rename the `async` field to `async_`. While we could always employ `job.get('async_', job.get('async'))` when dealing with job documents - is this something we should "migrate" as part of the move to Girder 3? Alternatively, do we want to leave the document format alone entirely? It adds inconsistency which is unfortunate, but there is no technical reason it needs to be escaped.

zachmullen:
> So it appears we're deprecating the usage of async in Girder 2 and 3, and will only remove the ability to pass async in Girder 4? Is that right? I think we should be able to make these sorts of breaking changes in master right now.

I don't have strong opinions on this. I know there are some people in the Girder community who have argued that symbols that are being deprecated should be deprecated for some long period of time or for a full major release before removal, but I'm not arguing for that personally.

> Alternatively, do we want to leave the document format alone entirely? It adds inconsistency which is unfortunate, but there is no technical reason it needs to be escaped.

+1 for leaving the documents as-is. These fields are always referenced as strings, so it's OK to have one called `async`. One reasonable argument for changing it might be that we anticipate using an ODM at some point that would expose these keys to attributes, but if we do something like that, it would be on some other major version, and we could cross the data migration bridge at that point.

giesberge:
Chiming in. On the first point I'm of the opinion that the overhead associated with the function is minimal. The developer base is small enough now that it may not matter.

On the second point internal consistency is nice and avoiding using async is good in general.

jeffbaumes:
@zachmullen this looks clear to merge. Anything stopping us from doing so or is more discussion required?

PR #2818, opened 2018-10-03
Add Discourse to CONTRIBUTING.rst

giesberge:
Also removed mailing list
https://public.kitware.com/pipermail/girder-users/2017-September/000107.html


zachmullen:
@giesberge if it's not too much trouble, could you also add this change to the `master` branch?

PR #2820, opened 2018-10-03
Removed NODE_PATH addition to scripts from package.json

giesberge:
NPM run already sets the local node_modules path when its run. Wasn't sure why this was in there as it's one of the few lines blocking windows compatibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,8 +73,8 @@
         "whatwg-fetch": "^2.0.4"
     },
     "scripts": {
-        "build": "NODE_PATH=$PWD/node_modules grunt",
-        "watch": "NODE_PATH=$PWD/node_modules grunt --watch",
+        "build": "grunt",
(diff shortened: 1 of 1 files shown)


manthey:
@giesberge This was done to handle symlinked plugins.  It is no longer present in the master branch.

To not break symlinked plugin support on non-Windows systems, you could make this `"NODE_PATH=$PWD/node_modules grunt || grunt"`.

We aren't directly supporting Windows, but to get it working, I also had to change a line in girder/utility/install.py from `proc = subprocess.Popen(cmd, cwd=wd)` to `proc = subprocess.Popen(cmd, cwd=wd, shell=hasattr(sys, 'getwindowsversion'))`.

giesberge:
@manthey I just used the --npm flag to point to my npm.cmd to solve the second issue.

PR #2821, opened 2018-10-03
Windows compatibility improvements

giesberge:
Updated build to use windows shell if necessary, also updates fontella grunt task to properly unzip in Windows.

### DIFF
--- a/girder/cli/build.py
+++ b/girder/cli/build.py
@@ -58,7 +58,7 @@ def main(dev, watch, watch_plugin, reinstall):
         if os.path.exists(npmLockFile):
             os.unlink(npmLockFile)
 
-        check_call(['npm', 'install'], cwd=staging, shell=hasattr(sys, 'getwindowsversion'))
+        check_call(['npm', 'install'], cwd=staging)
 
     quiet = '--no-progress=false' if sys.stdout.isatty() else '--no-progress=true'
     buildCommand = [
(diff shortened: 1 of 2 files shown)


girder/cli/build.py
```diff
@@ -58,7 +58,7 @@ def main(dev, watch, watch_plugin, reinstall):
         if os.path.exists(npmLockFile):
             os.unlink(npmLockFile)
 
-        check_call(['npm', 'install'], cwd=staging)
+        check_call(['npm', 'install'], cwd=staging, shell=hasattr(sys, 'getwindowsversion'))
```
zachmullen: @giesberge could you help me understand this change? What problem is being solved here for Windows that is fixed by running in a shell?

manthey: @zachmullen Without using the shell, python subprocess reports that it can't find the `npm` command, even though on Windows, it is available via `where`.  With the shell it is resolved properly.   Oddly, `subprocess.check_call("where npm")` reports where it is, but `subprocess.check_call("npm")` can't find it, though `subprocess.check_call("npm", shell=True)` can.

manthey: It turns out this is because nodejs uses a `.com` file for `npm.com`, and `.com` files can't be executed directly, only through a shell.  If it was `npm.exe`, it would work as expected.

zachmullen: This feels like an npm (or windows?) bug, however we did used to support this via the now removed `--npm` option.

I think the best thing would be to add back `--npm`, I can't remember why it was removed.

zachmullen:
@giesberge we have a PR that should be merged today that will take care of this issue for you in a more cross-platform way: https://github.com/girder/girder/pull/2826

Once that goes in, you can just pass the path to npm either as an argument or you can set it as an environment variable.

Thanks!

giesberge:
@zachmullen still need to merge in the the proper delimiters inside the fontella.js grunt script.

zachmullen:
Good point, could you scope this down to that and then I will merge it?

PR #2825, opened 2018-10-04
Decimate plugins

zachmullen:
This also fulfills some TODOs related to the migration guide.


docs/migration-guide.rst
```diff
@@ -197,6 +197,31 @@ handler, you must call :py:func:`girder.events.unbind` on the existing mapping f
...
+* metadata_extractor
+* mongo_search
+* provenance
+* treeview
+* vega
+
+The following plugins were **moved to different repositories**:
+
+* geospatial
+* worker
```
zachmullen: I plan to change these to links. @dorukozturk where is the new geospatial plugin, and is it compatible with the old one in terms of data model?

docs/migration-guide.rst
```diff
@@ -197,6 +197,31 @@ handler, you must call :py:func:`girder.events.unbind` on the existing mapping f
     # Prints 'a' and 'b' in Girder 2, but only 'a' in Girder 3
     events.trigger('an_event')
 
+Removed or moved plugins
+++++++++++++++++++++++++
+
+Many plugins were either deleted from the main repository, or moved to other repositories. Plugins
+that were moved to other repositories will no longer be installed via the ``[plugins]`` extra when
+installing the ``girder`` python package, but can be installed by
+``pip install girder_plugin_[name]``. If you were depending on a plugin that was deleted
```
jcfr: should that be `pip install girder-plugin-[name]` ? this would be more consistent with python convention

zachmullen:
Ok, this is ready to go. My diffstats are now unstoppable.

danlamanna:
fixes #2385

plugins/.gitignore
```diff
@@ -5,25 +5,18 @@
 !audit_logs/
```
zachmullen: Should I get rid of this file altogether? We no longer copy external plugins into this directory, right?

jbeezley: You still need to if you want to run cmake tests for an external plugin.  There may be a way to hack around this limitation, but I haven't explored it yet.

zachmullen:
Any objections to this being merged? Just need someone to mark approval

PR #2829, opened 2018-10-07
Upgrade docker-compose.yml to version 2

ariarijp:


zachmullen:
If we are updating this version, shouldn't we upgrade to v3, which is the latest and recommended version?

ariarijp:
Version 3 is latest and I recommend it, but version 2 is still popular.

I think we can skip version 2 for now.

zachmullen:
What's the motivation for the change?

ariarijp:
Hi @zachmullen ,

I prefer to use `depends_on` instead of version 1 style `link` and `ports`.

Version 2 and 3 style is easy to understand dependencies of services.

Thank you,

Issue #1933, opened 2017-04-19
Generate version info on server side rather than client build time


zachmullen:
See discussion on #1931

PR #2832, opened 2018-10-08
Skip asking for the git revision information in production mode.

manthey:
This is cleaner than always failing and then reassuring the user that failure was expected.

### DIFF
--- a/girder/web_client/grunt_tasks/build.js
+++ b/girder/web_client/grunt_tasks/build.js
@@ -23,7 +23,6 @@ const webpack = require('webpack');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const GoogleFontsPlugin = require('google-fonts-webpack-plugin');
 const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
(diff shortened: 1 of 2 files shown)


jbeezley:
The issue here is that you can trigger error message when running:
```
pip install girder --pre
girder build --dev
```

That is, the development mode referred to by the message is the python package mode not the node environment.  Perhaps the message could be clarified, but I don't think it should go away.

manthey:
I didn't take away the message.  I stopped try to get git revision information in production mode and remove the part of the message that read `This is normal for non-development builds of girder`.

jbeezley:
IMO that was the important part of the message.  I added it because that command emits (even when caught) a message that says:
```
fatal: not a git repository (or any parent up to mount point /)
```
The message was intended to reassure people paying attention that this did not represent a fixable problem with their build.

If we want to go ahead with this, I'm okay with it, but it does mean that the version information in a production instance will always be `git: false, SHA: null`.  In light of #1933, we might be better just removing this entirely.

manthey:
You've convinced me this should stay, though I still wish we were clever enough not to print a fatal warning that we expect.

PR #2836, opened 2018-10-08
Updated ansible role for Girder 3

danlamanna:
This updates the ansible role to work with Girder 3. Overall, the direction of the role is shifting towards production installations done entirely with pip. There's support for passing a path so development installs can be managed, but it's not the primary goal.

The most notable changes:

- We only support Ubuntu 16+
- All testing/linting is now done with Molecule (@kotfic this includes your library tests)
  ```
  cd devops/ansible-role-girder
  pip install python-vagrant docker molecule
  molecule test --all
  ```
- Installing Girder against system python no longer works, a virtualenv has to be given
  This has the added benefit of implicitly handling Python 2 or 3.

Left to do:
- [x] Fix CircleCI to use upgraded version of role
- [ ] Remove galaxy
  - [ ] Removing all associated credentials/secret environment variables

I was also looking through `devops/` in an attempt to clean up some things. @manthey  It appears the `mongo-shard-rs` and `mongo-replica-set` Vagrant setups aren't working, and haven't been modified in a while. Are these being used anywhere and are they still useful to keep around?


devops/ansible-role-girder/meta/main.yml
```diff
@@ -1,20 +1,17 @@
+---
 galaxy_info:
   author: Kitware, Inc.
   description: Install the Girder data management platform.
   company: Kitware, Inc.
   issue_tracker_url: https://github.com/girder/girder/issues
-
-  license: Apache
-  min_ansible_version: 2.0
+  license: Apache 2.0
```
brianhelba: I was wrong, this should be `Apache`

devops/ansible-role-girder/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+
+# Setup systemd service
+- block:
+    - name: Install Girder service
+      template:
+        src: "daemon/girder.service.j2"
+        dest: "/etc/systemd/system/girder.service"
+      notify: Restart Girder
+
+    - name: Enable/Disable Girder service on boot
```
brianhelba: Should just be "Enable"

brianhelba:
I've reported issues for both: #2996 #2997

PR #2839, opened 2018-10-09
Throw exception for interrupted file downloads

zachmullen:
Prior to this change, if a connection was closed partway through a download, the python client code did not raise any exception, leaving downstream users to deal with the fallout of unknowingly having a partial file. See [this article](https://blog.petrzemek.net/2018/04/22/on-incomplete-http-reads-and-the-requests-library-in-python/) for more information about the counter-intuitive behavior of the requests library in this scenario.

Because this condition typically indicates ephemeral network or service failures, I think it's good to have it raise a distinct exception type that carries those semantics, allowing downstreams to catch this type and handle it differently than other types of HTTP errors, namely by adding retry logic, whereas we don't necessarily want to retry if we receive an HTTP error status but an otherwise intact response.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -98,6 +98,13 @@ class HttpError(requests.HTTPError):
         return super(HttpError, self).__str__() + '\nResponse text: ' + self.responseText
 
 
+class IncompleteResponseError(requests.HTTPError):
+    def __init__(self, message, expected, received, response=None):
+        super(IncompleteResponseError, self).__init__('%s (%d of %d bytes received)' % (
+            message, received, expected
+        ), response=response)
+
+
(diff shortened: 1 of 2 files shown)


danlamanna:
Do we want to handle this at the `downloadFile` function level or at the `get` function level?

zachmullen:
I don't want to close the door on adding it more generally in the future, but I would prefer not to do so on this PR, because the solutions for streamed responses (like this one) and non-streamed responses (like most other REST requests) are totally different. That is to say, a general solution that goes into `get` would likely only be able to run if `stream=False`, meaning we'd still need this specialized solution in place anyway.

zachmullen:
Also, we're kind of already covered for any request where `json=true` since the JSON parse will fail in the case of a partial response.

PR #2842, opened 2018-10-16
Add more clarity to migration docs

jeffbaumes:
For people less familiar with constructing python packages.


zachmullen:
I restarted but it failed in the same way. It looks like web client coverage reporting is broken. Is this based on latest master?

ping @manthey is this a known issue related to the recent package dependency changes?

jbeezley:
It shouldn't...  We use `--dev` in all of the circle environments.  I'll poke around in the build container.

PR #2843, opened 2018-10-16
Use a compatible version of nyc for coverage reporting

jbeezley:


zachmullen:
Our devDependencies in package.json express the nyc version as follows:

    "nyc": "^11.3.0"

I have no understanding of how or if we use npm lockfiles anymore, but would it make sense to simply change the version in package.json rather than inside the script?

jbeezley:
The problem is that with the reorg, it isn't using nyc from that `package.json` at all.  Instead, it's installing a new one (or using whatever happens to be in the path).  I went ahead and removed it from the girder package because it wasn't necessary there.  The toplevel package.json is probably the least surprising place to put it.

zachmullen:
Should we do the same for `phantomjs-prebuilt`?

jbeezley:
Yeah, probably.  Let me make sure it works.

zachmullen:
To put it more generally, what is the rationale for a test-time dependency to go in one file vs. the other? Ping @manthey

PR #2845, opened 2018-10-18
Change behavior of ModelImporter

### LINKED ISSUE #2377: Remove importlib guesswork behavior from ModelImporter
Require models to use the new explicit registration behavior for those who need to look them up using strings.

zachmullen:
Fixes #2377. This removes the ability of the ModelImporter to
infer the class of a model based on the string(s), instead requiring
each model to be explicitly registered to a string if it is needed
for model loading.

This also changes the API for registering models to the ModelImporter;
rather than registering the model instance, one now only registers the
model type. This is cleaner in that it can defer model instantiation,
and also makes the ModelImporter decoupled from the fact that models
are singletons.

This is a breaking change.

### DIFF
--- a/girder/utility/model_importer.py
+++ b/girder/utility/model_importer.py
@@ -17,44 +17,7 @@
 #  limitations under the License.
 ###############################################################################
 
-import importlib
-import six
-
-from . import camelcase
-from girder import logger
-
-# We want the models to essentially be singletons, so we keep this centralized
-# cache of instantiated models that have been lazy-loaded.
-_modelInstances = {}
-
-
-def _loadModel(model, module, plugin):
-    className = camelcase(model)
-
-    try:
-        imported = importlib.import_module(module)
-    except ImportError:
-        logger.exception('Could not load model "%s".' % module)
-        raise
-
-    try:
-        constructor = getattr(imported, className)
-    except AttributeError:
-        raise Exception('Incorrect model class name "%s" for model "%s".' % (
-            className, module))
-
-    _modelInstances[plugin][model] = constructor()
-
-
-def reinitializeAll():
-    """
-    Force all models to reconnect/rebuild indices (needed for testing).
-
-    .. deprecated
-    """
-    for pluginModels in list(six.viewvalues(_modelInstances)):
-        for model in list(six.viewvalues(pluginModels)):
-            model.reconnect()
+_modelClasses = {}
 
 
 class ModelImporter(object):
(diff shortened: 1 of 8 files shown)


zachmullen:
@girder/developers PTAL. This PR lumps together two related but independent changes, if there is controversy I can separate them or back out the secondary change.

1. We now require models to be explicitly registered, we won't attempt to guess at how to import a model class based on the string. This fixes #2377.
2. Rather than registering the model instance, the API has been changed to instead accept the model class. That means the singleton nature of the models is now more of an internal detail rather than something that is reflected in the ModelImporter API, which I think is better. If you guys approve of this change, I will need to change some of the documentation as well.

jbeezley:
Is there a particular reason we are keeping `ModelImporter` around at all?  It's not too bad as an encapsulation of the registry itself, but it's original purpose has been more or less discouraged.

It looks to me like there are many places in the code that use `ModelImporter.load` in ways that are unnecessary, and only exist for historical reasons.  As an example, see [api/v1/system.py](https://github.com/girder/girder/blob/master/girder/api/v1/system.py).  All the usages there would be easily expressed as direct imports.

Removing this class and its use as a base class would force cleanup in all of these places, and hopefully cause people to examine when dynamically dereferencing a model from its name is truly necessary.

zachmullen:
> It looks to me like there are many places in the code that use ModelImporter.load in ways that are unnecessary, and only exist for historical reasons. As an example, see api/v1/system.py. All the usages there would be easily expressed as direct imports.

Good find, those should definitely be replaced with direct imports. They fell through the cracks of the refactor because I did a `git grep \.model\('` rather than a totally comprehensive search.

> Removing this class and its use as a base class would force cleanup in all of these places

I considered removing it as a mixin to some of the base classes on this PR, but it felt out of scope for addressing #2377. I would probably get behind that change, but on a follow-up PR to this if that's OK.

jbeezley:
Sure, that works for me.

girder/utility/model_importer.py
```diff
@@ -63,53 +53,44 @@ class ModelImporter(object):
...
         """
         Call this to get the instance of the specified model. It will be
         lazy-instantiated.
 
-        :param model: The name of the model to get. This is the module
-                      name, e.g. "folder". The class name must be the
-                      upper-camelcased version of that module name, e.g.
-                      "Folder".
+        :param model: The name of the model to get. This must have been
+            registered using the :py:ref
```
jbeezley: incomplete reference

girder/utility/model_importer.py
```diff
@@ -17,44 +17,34 @@
 #  limitations under the License.
 ###############################################################################
 
-import importlib
-import six
+_modelClasses = {}
+_coreModelsRegistered = False
```
jbeezley: This is another case of global context that might cause issues in testing.  Should we reset it in the db fixture?

zachmullen: Not sure. We don't currently have any tests that depend on the core models being unregistered, so I'm inclined to think YAGNI, but I could go either way.

zachmullen:
@jbeezley this is ready for final review

PR #2846, opened 2018-10-18
RFC: Remove MacOS instructions from docs

zachmullen:
Follow-on from our discussion about platform support. Since MacOS isn't a tested platform, it should only be used by the brave. As such, it probably shouldn't coexist on the (literal) same level as the Linux tabs.

How to install on MacOS should probably still be captured somewhere since many developers like to use it for Girder, but that place could be somewhere like discourse, or somewhere else entirely.

### DIFF
--- a/docs/installation-quickstart.rst
+++ b/docs/installation-quickstart.rst
@@ -57,20 +57,6 @@ Basic System Prerequisites
 
          sudo yum -y install openldap-devel cyrus-sasl-devel
 
-   .. group-tab:: macOS
-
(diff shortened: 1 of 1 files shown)


manthey:
I think this sort of documentation is useful, if only for developers that rarely use OSX but want to test there.  Would it be appropriate to have the removed section in a file called something like `installation-unsupported.rst` with a caveat at the top like: `Note: The systems listed here are not supported.  Some or all features may not work.`

PR #2850, opened 2018-10-24
Fix flake8 errors related to linting packages update (ported to 2.x)

jbeezley:


jbeezley:
For 2.x, we also have to disable the async/await check because we don't have #2817.

zachmullen:
Was there a plan to backport #2817 ?

manthey:
We can probably merge this despite the coverage test failure.

PR #2853, opened 2018-10-31
[WIP] Better error sanitization in prod mode

zachmullen:
Prior to this change, unexpected exceptions that occurred during requests would be sent to the user with some of the original exception information intact (namely the type and message field).

This PR makes two changes related to this.

1. In prod mode, users will no longer receive such exception information in REST responses. It will appear in the log as before, but the message returned to the user is now generic.
2. In order to more easily match user reported exceptions to logged exceptions, a UID is generated for each such incident and written to both the response and the corresponding log message.

If you approve of these changes I'll write verification tests.

@jcfr @girder/developers


girder/api/rest.py
```diff
@@ -661,15 +662,19 @@ def endpointDecorator(self, *path, **params):
...
-            val = {'message': '%s: %s' % (t.__name__, repr(value)),
-                   'type': 'internal'}
-            curConfig = config.getConfig()
-            if curConfig['server']['mode'] != 'production':
-                # Unless we are in production mode, send a traceback too
+            val = dict(type='internal', uid=uid)
+
+            if config.getConfig()['server']['mode'] == 'production':
+                # Sanitize errors in production mode
+                val['message'] = 'An unexpected error occurred on the server.'
```
jcfr: I suggest a more friendly message with a recommended action.

```
We are sorry, an unexpected error occurred on the server. Please, contact the administrator at xxxx and share the identifier yyyy to help them lookup related information.
```

aashish24: +1 I would recommend passing UUID so that users can share that info with admins.

danlamanna:
I could see UID generation occurring at the earliest point in an HTTP request and being stored on the request local thread. That way we could easily trace all logs belonging to an individual request, though maybe our logs store the thread id, I can't remember.

girder/api/rest.py
zachmullen: The UID is passed in a `uid` field of the response document, rather than in-band within the message. I think it's good to keep it there and have display of the UID be a client-side consideration and keep the `message` field to strictly human-readable stuff.

jcfr: > have display of the UID be a client-side consideration 

Make sense

Should the update of the client side be part of the PR ?

danlamanna:
It also seems common to put a request identifier in response headers, for example: https://stackoverflow.com/a/27174552.

zachmullen:
@danlamanna I think that's a good idea, though it should be on a future PR. It won't break anything that is done on this PR.

girder/api/rest.py
zachmullen: I don't think I want to update the Backbone client with this PR. Most of our error reports involve people posting the full JSON response of the failing request anyway.

Since this is a UX concern I'd defer to @jtomeck about the best way to present this to users, though it is likely different for each Girder-based front end.

zachmullen:
There's a lot of discussion about the UID features on this, to the point that I think it merits its own PR so that we don't hold up the security fix that motivated this one. I'll isolate this one just to part 1 of the description and try to implement @danlamanna 's suggestion on a follow up PR.

zachmullen:
Ready for a final review

PR #2855, opened 2018-11-06
Add the mock dependency to pytest-girder

danlamanna:
This was being brought in transitively, which was causing breakage in other scenarios.

@jbeezley


danlamanna:
@jbeezley

jbeezley:
Do you want to increment the version number?

PR #2857, opened 2018-11-07
Use latest release of boto/botocore

jbeezley:
After some experimenting I found the boto pinning to be mostly unnecessary with the latest moto.  It does appear that dummy credentials are now required (https://github.com/spulec/moto/issues/1924), but it's unclear if this is intentional or a bug.

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -473,7 +473,7 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
         disp = 'attachment; filename="%s"' % file['name']
         mime = file.get('mimeType') or ''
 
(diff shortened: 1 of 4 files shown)


girder/utility/s3_assetstore_adapter.py
```diff
@@ -473,7 +473,7 @@ def fileUpdated(self, file):
         disp = 'attachment; filename="%s"' % file['name']
         mime = file.get('mimeType') or ''
 
-        if key['ContentType'] != mime or key['ContentDisposition'] != disp:
+        if key.get('ContentType') != mime or key.get('ContentDisposition') != disp:
```
jbeezley: I'm a little unsure about this change.  These keys are now missing in the `server_file` test, but I'm not certain why.

PR #2859, opened 2018-11-08
Support directory outputs from slicer cli's (2.x)

jbeezley:


manthey:
Is there a way to choose an existing folder?  Or do you have to type a name when picking a folder?

Issue #2864, opened 2018-11-26
DICOM images out of order


manthey:
DICOM image slices within an item are presented out of order.  The item's dicom metadata correctly lists the files in order, but the Backbone files collections sorts them alphabetically by name, which is incorrect.  Instead of slices 1, 2, 3, ..., 9, 10, 11, ..., you get 1, 10, 11, ..., 19, 2, 20, ...

manthey:
Fixed by #2865.

PR #2865, opened 2018-11-26
Fix the order that slices are shown in the DICOM viewer.

manthey:
This also stops wrapping long file names, as that causes the controls to jump around.


manthey:
Fixes #2864.

manthey:
It looks like pytest>=4 breaks our tests.

PR #2869, opened 2018-11-29
Allow the S3 assetstore to work with google storage.

manthey:
This works without credentials and in a read-only manner.

This also adjusts the presigned url, which may allow it to work with read/write or credentials, but those have not been tested.

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -134,7 +134,15 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
         """
         url = self.client.generate_presigned_url(*args, **kwargs)
         if getattr(self.client, '_useGoogleAccessId', False):
-            url = url.replace('AWSAccessKeyId', 'GoogleAccessId')
+            awskey, gskey = 'AWSAccessKeyId', 'GoogleAccessId'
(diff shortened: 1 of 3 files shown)


manthey:
I've manually tested this with a public Google Storage site, but not with a credentialed site.

The approach is not elegant, but I think it could be used until we have motivation to do a GS assetstore.

girder/utility/s3_assetstore_adapter.py
```diff
@@ -117,6 +125,18 @@ def _getRequestHeaders(self, upload):
...
+    def _generatePresignedUrl(self, *args, **kwargs):
+        """
+        Wrap self.client.generate_presigned_url to allow it work with
+        Google Cloud Storage.
+
+        See https://gist.github.com/gleicon/2b8acb9f9c0f22753eaac227ff997b34
+        """
+        url = self.client.generate_presigned_url(*args, **kwargs)
+        if getattr(self.client, '_useGoogleAccessId', False):
+            url = url.replace('AWSAccessKeyId', 'GoogleAccessId')
```
zachmullen: Sorry for the confusion, this is actually where I was talking about using `urlparse` rather than `str.replace`, though it's also an improvement where you changed it up above.

manthey:
@zachmullen Can you approve again?  I had to resolve conflicts with master.

PR #2879, opened 2018-12-07
[RFC] Archive access

manthey:
This isn't intended to be merged as is, it is a proof-of-concept for further discussion.

This branch contains two things based on the 2.x-maintenance branch:

1. A plugin called "Archive access".  It modifies the `File()` model with two extra methods `archiveList` and `archiveOpen`, plus three rest endpoints `file/{id}/archive`, `file/{id}/archive/download?path={archive path}`, and `file/{id}/archive/download/{name}?path={archive path}`.

  `archiveList` checks if a Girder file is a zip or tar (possibly compressed) file, and, if so, returns a list of paths within the archive.
  'archiveOpen' returns a file-like object to a path within a Girder file that can be opened as an archive.  It can be used as a context in the same manner as `File().open` using a `ArchiveFileHandle` class.

2. The DICOM plugin has been modified so that, if the archive access plugin is enabled, it can used archive files that contain DICOM files.

Questions:

* Should this be a plugin or just added to Girder's File model directly?  
* If a plugin, should it be in core (in Girder 3), or external?  If not a plugin, should this have a setting to enable it?
* Reading from compressed tar files can tax the server (and, to a lesser extent, so can zip files).  Should this ability be an option or removed?
* Having DICOMs work with compressed files is handy.  DICOM is in core, so if the archive access is a plugin, it seems like for proper separation the changes to DICOM should be in a plugin as well, but this seems fairly convoluted.  Can anyone propose a clean way to do that?

### DIFF
--- a/plugins/archive_access/plugin.json
+++ b/plugins/archive_access/plugin.json
@@ -1,0 +1,7 @@
+{
+    "name": "Archive access",
+    "description": "Provide utiltiies for accessing internal files of archive (zip and tar) files.",
+    "version": "0.1.0",
+    "dependencies": []
+}
+
--- a/plugins/archive_access/server/__init__.py
+++ b/plugins/archive_access/server/__init__.py
@@ -1,0 +1,301 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import cherrypy
+import os
+import tarfile
+import zipfile
+
+from girder.api import access
+from girder.api.describe import Description, autoDescribeRoute, describeRoute
+from girder.api.rest import boundHandler, setResponseHeader, setContentDisposition
+from girder.constants import AccessType, TokenScope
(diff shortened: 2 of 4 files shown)


kotfic:
It would be nice to understand the relationship between this PR and [girder/tape_archive](https://github.com/girder/tape_archive). Is there a larger feature/plugin here that generically works with archives?

manthey:
> It would be nice to understand the relationship between this PR and [girder/tape_archive](https://github.com/girder/tape_archive). Is there a larger feature/plugin here that generically works with archives?

They are orthogonal to each other.  The tape_archive plugin imports data from a tar file and exports to a tar file.  This PR allows reading internal files from zip or tar files as python file-like objects.

manthey:
This was made into a stand-alone plugin, so I'm closing this PR.

PR #2881, opened 2018-12-11
Disable failing test

zachmullen:
The lack of lockfiles for plugin npm deps has led to a hairy situation with the jobs plugin dependency tree. Something underneath the vega-lib subtree moved, and is causing some ES6 syntax to be injected into the phantomjs environment.

For now I've chosen to disable the test because I think that's the least expensive path forward; it's difficult to discover what actually changed in this very large package subtree to cause the issue, but if someone wants to dig around to try and find it, they're welcome to.

This is just another example of why we should decentralize the building of client code and not have it be something managed by Girder core at all.


zachmullen:
@girder/developers could someone review? All merging will be blocked until this goes in, or some alternative to this that actually fixes the problem.

PR #2882, opened 2018-12-12
Fix the failing web client jobs test

zachmullen:
This required two changes:

* Add a webpack helper for the jobs plugin to babel-load its third party dependencies
* Include babel-polyfill in the test environment runtime (for regeneratorRuntime)

@jbeezley PTAL, this re-enables the test that was temporarily disabled.


plugins/jobs/girder_jobs/web_client/webpack.helper.js
```diff
@@ -0,0 +1,15 @@
+module.exports = function (config) {
+    config.module.rules.push({
+        resource: {
+            test: /\.js$/,
+            include: [/node_modules\//]
+        },
```
jbeezley: This will add babel to all js files in node_modules globally.  If that's the intent, you might as well just add this rule to the top level webpack config.

zachmullen: On my machine this appears to only affect the build of the jobs plugin target, not any other plugins or the core build.

zachmullen:
Strange, this is working for me locally, but all the tests are broken on CircleCI.

plugins/jobs/girder_jobs/web_client/webpack.helper.js
jbeezley: Yeah, I guess it generate a separate webpack config object for each plugin.  This is fine then.

zachmullen:
Alles in Ordnung

PR #2885, opened 2018-12-18
Fix the ctest javascript coverage report.

manthey:
Tell nyc where we stored the coverage data so the test passes.  With this change, ctest successfully passes locally.  Without it, it fails.

### DIFF
--- a/tests/JavascriptTests.cmake
+++ b/tests/JavascriptTests.cmake
@@ -12,7 +12,7 @@ function(javascript_tests_init)
   add_test(
     NAME js_coverage_combine_report
     WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
-    COMMAND npx nyc report
+    COMMAND npx nyc report -t "${PROJECT_SOURCE_DIR}/build/test/coverage/web_temp"
   )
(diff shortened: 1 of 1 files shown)


tests/JavascriptTests.cmake
```diff
@@ -12,7 +12,7 @@ function(javascript_tests_init)
   add_test(
     NAME js_coverage_combine_report
     WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
-    COMMAND npx nyc report
+    COMMAND npx nyc report -t "${PROJECT_SOURCE_DIR}/build/test/coverage/web_temp"
```
zachmullen: This option didn't exist in major version 11 of nyc, which is what's listed in our package.json. Should we upgrade our nyc dependency to 13.x?

manthey: It is present in the long form of `--temp-directory` in nyc 11.8.0 (and earlier).  13.1 prefers `--temp-dir` and added `-t`.  I'll up the dependency to 13.1 and move to the `--temp-dir`.

zachmullen:
Looks like you forgot to check in the `package-lock.json` changes

PR #2888, opened 2019-01-09
Add "get metadata" endpoints for item and folder

waxlamp:
This adds two new endpoints:
- `GET /folder/{id}/metadata`
- `GET /item/{id}/metadata`

along with tests.

### DIFF
--- a/girder/api/v1/folder.py
+++ b/girder/api/v1/folder.py
@@ -39,6 +39,7 @@ class Folder(Resource):
         self.route('DELETE', (':id', 'contents'), self.deleteContents)
         self.route('GET', (), self.find)
         self.route('GET', (':id',), self.getFolder)
+        self.route('GET', (':id', 'metadata'), self.getFolderMetadata)
         self.route('GET', (':id', 'details'), self.getFolderDetails)
         self.route('GET', (':id', 'access'), self.getFolderAccess)
         self.route('GET', (':id', 'download'), self.downloadFolder)
@@ -103,6 +104,17 @@ class Folder(Resource):
                 text, user=user, limit=limit, offset=offset, sort=sort)
         else:
             raise RestException('Invalid search mode.')
+
+
+    @access.public(scope=TokenScope.DATA_READ)
+    @autoDescribeRoute(
+        Description('Get folder metadata.')
+        .modelParam('id', model=FolderModel, level=AccessType.READ)
+        .errorResponse()
+        .errorResponse('Read access was denied on the folder.', 403)
+    )
+    def getFolderMetadata(self, folder):
+        return folder.get('meta', {})
 
     @access.public(scope=TokenScope.DATA_READ)
     @autoDescribeRoute(
--- a/girder/api/v1/item.py
+++ b/girder/api/v1/item.py
@@ -38,6 +38,7 @@ class Item(Resource):
         self.route('DELETE', (':id',), self.deleteItem)
         self.route('GET', (), self.find)
         self.route('GET', (':id',), self.getItem)
+        self.route('GET', (':id', 'metadata'), self.getItemMetadata)
         self.route('GET', (':id', 'files'), self.getFiles)
         self.route('GET', (':id', 'download'), self.download)
         self.route('GET', (':id', 'rootpath'), self.rootpath)
@@ -220,6 +221,18 @@ class Item(Resource):
                     yield data
             yield zip.footer()
         return stream
+
+
+    @access.public(scope=TokenScope.DATA_READ)
+    @autoDescribeRoute(
+        Description('Get item metadata.')
+        .modelParam('id', model=ItemModel, level=AccessType.READ)
+        .errorResponse()
+        .errorResponse('Read access was denied on the folder.', 403)
+    )
+    def getItemMetadata(self, item):
+        return item.get('meta', {})
+
 
     @access.public(scope=TokenScope.DATA_READ)
     @filtermodel(model=File)
--- a/tests/cases/folder_test.py
+++ b/tests/cases/folder_test.py
@@ -305,6 +305,22 @@ class FolderTestCase(base.TestCase):
             resp.json['message'],
             'Parameter metadata must be valid JSON.')
 
+        resp = self.request(
+            path='/folder', method='POST', user=self.admin, params={
+                'name': 'My public subfolder without meta',
+                'parentId': publicFolder['_id']
+            }
+        )
+        self.assertStatusOk(resp)
+        folder = resp.json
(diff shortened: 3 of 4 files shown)


brianhelba:
@waxlamp Are you still interested in pursuing this PR? If so, can you provide some rationale for it.

waxlamp:
I don't remember my specific use case for this, but I needed to retrieve folder/item metadata, and was surprised by the lack of API for doing so. I'm sure I worked around it by taking the longer route, but these seemed simple enough to add so I added them.

brianhelba:
`GET /folder/{id}` and `GET /item/{id}` return metadata within the `meta` field.

Is there a case to be made for not using these?

brianhelba:
Nevermind, I see the issue here. New folders, etc. are not created with an empty `meta` field, so the field is initially nonexistent in the API response. Setting and clearing some metadata will then cause the API to return `"meta": {}` in future API responses.

I'd support changing some internal code to always add a `meta` field to new folders, etc., though this would not fix the issue for any folders, etc. in an existing database.

For better or worse, this sort of schema-less nature is an intrinsic part of Girder's current design. Trying to comprehensively make any sort of strong guarantees about what fields exist is awkward and arbitrary at best. The solution seems to typically be to defensively code around many things that come out of the database, including `meta` in this case.

waxlamp:
You're right, that's why I originally did this (lack of `meta` field altogether if metadata had never been set).

I'm fine with closing this PR without merging, but I did end up opening it for a genuinely puzzling interaction around this behavior. If it's documented somewhere and I missed it, then certainly there's not much argument for merging this PR.

brianhelba:
> If it's documented somewhere and I missed it

Yeah, my point is that the lack of a schema makes things undocumentable in principle. We could add some ad-hoc text about the `meta` field somewhere in the docs where few people will ever see it, but that doesn't help much, IMO.

I'd suggest filing a bug to request that the `meta` field be added to newly created folders and items automatically; I know someone who will have time to fix that bug within the next 2-3 weeks. That at least makes things more user-friendly in the future, though again, it's not something we can robustly rely on to always work, since old databases will still lack the field.

waxlamp:
In that case, can I hear the argument *against* merging this PR?

Regardless of whether this is merged, I agree that we should open an issue to add an empty `meta` field to new resources, but adding a new endpoint that paves over this schemalessness problem at least provides a uniform way to see what the metadata values are. Even though Mongo is a schemaless DB, that doesn't mean our "schema" (whatever technical meaning that may have in context) doesn't include metadata on items and folders--Girder's model is definitively that it does.

zachmullen:
> In that case, can I hear the argument against merging this PR?

I would argue that adding new endpoints to return a specific sub-field of a conceptual document is probably not a desirable solution to the problem as stated, which if I understand correctly, is that the schema of those documents, as returned by the REST API, does not always contain a consistent set of keys.

It seems to me that a cleaner alternative would be to change the existing `GET /item/{id}` and `GET /folder/{id}` endpoints to ensure that the `meta` field is always sent back, even if it doesn't exist on the documents stored in mongo. (In addition to making sure `meta` is always set on newly created documents moving forward.)

waxlamp:
@zachmullen, if the constraint is "try not to add new endpoints" then I think your solution is perfect.

I can work on converting this PR to do what you suggest. Is that the right way forward?

brianhelba:
> adding a new endpoint that paves over this schemalessness problem at least provides a uniform way to see what the metadata values are

My objection is that it's a lot of redundant code and API surface to only help the issue for exactly one field. If you want to see `meta` added in a backwards-compatible way, just add another 'migration' ([which we already have several of](https://github.com/girder/girder/blob/master/girder/models/folder.py#L122)) to the Folder and Item `load` method to dynamically add a `meta` field to those documents when they're loaded.

> Even though Mongo is a schemaless DB, that doesn't mean our "schema" (whatever technical meaning that may have in context) doesn't include metadata on items and folders--Girder's model is definitively that it does.

You bring up a really good point. Indeed, a schemaless database does not imply a schemaless app, since apps can provide their own validation / enforcement mechanisms (Girder does this somewhat selectively). Also, you're right in that every app does end up with some sort of *de facto* schema, but without enforcement at some level, the schema becomes much more complicated to describe.

However, my point is just that, for better or worse, the current situation is an almost-inevitable outcome of the way that Girder is designed. If it wasn't the `meta` field, it's bound to be something else. We could (and maybe some would argue we should) continue to throw more special-case code at the problem, but we'll be fighting the design and make things more difficult to maintain with so many LOC. Of course, if someone wants to significant re-architect Girder to make some guarantees about schemas generally, I'd be all onboard.

brianhelba:
Oops, I was writing my last response at the same time @zachmullen posted his.

If you want to address this for existing databases, please take the approach I suggested (`Folder.load`), instead of trying to change it at the API layer. Feel free to let me know if you disagree, @zachmullen .

Also, as I sort of hinted, we'll have a new Girder developer starting next week, and this is a perfect onboarding issue. If you'd rather not be bothered with this now @waxlamp , I'd be happy to have our new dev take it on within the next 2 weeks.

waxlamp:
> We could (and maybe some would argue we should) continue to throw more special-case code at the problem, but we'll be fighting the design and make things more difficult to maintain with so many LOC.

The gist of it seems to be not to add new endpoints to fix one-off issues stemming from the schema, because it becomes something of a slippery slope. That makes sense to me.

This whole thing is an interesting lesson in the spectrum of schemaless-to-strict-type-checking (or something like that). To oversimplify, it seems like embracing schemalessness wholeheartedly enables fast development but leads to issues such as this, while enforcing a schema from the get-go would have opposite effects. Of course, if you get the schema *wrong* from the start, then you have a whole other order of problems...

waxlamp:
@brianhelba, please do have the new Girder dev work on this. Let me know if I can help in any way.

brianhelba:
> To oversimplify, it seems like embracing schemalessness wholeheartedly enables fast development but leads to issues such as this, while enforcing a schema from the get-go would have opposite effects. Of course, if you get the schema _wrong_ from the start, then you have a whole other order of problems...

I also think there's a lesson in here about architecture-driven development too. If you skimp on architecting your app and allow it to organically grow, you gain immediate dividends in allowing new developers to contribute with minimal onboarding and gain the ability to rapidly change the app's requirements, since there's no objective standard to say a feature is inconsistent, incoherent, or out of scope. However, you will eventually pay for it down the road.

brianhelba:
> Let me know if I can help in any way.

Just close this PR and file a bug please.

waxlamp:
Closing in favor of #3035

Issue #2889, opened 2019-01-09
Bad plugins break Girder 3


manthey:
In Girder 3, we need to guard against bad plugins.  That is, suppose I have a plugin whose `__init__.py` script contains:

```
import unknown_module        # This is the problem
from girder.plugin import GirderPlugin

class MyPlugin(GirderPlugin):
  ...
```

with an appropriate entrypoint in its setup.py.  When Girder calls `_getPluginRegistry`, the `entryPoint.load()` call throws an uncaught exception and Girder never starts.  These failures occur before the plugin can be added to the _pluginRegistry (since we never learn its internal name).

We need to catch and log this exception.

Should we also add an approximate value to plugin registry?  That is, should we do something like
```
for entryPoint in iter_entry_points(_NAMESPACE):
        try:
            pluginClass = entryPoint.load()
            plugin = pluginClass(entryPoint)
            _pluginRegistry[plugin.name] = plugin
        except Exception:
            try:
                plugin = GirderPlugin(entryPoint)
                plugin._pluginFailureInfo = sys.exc_info()
                _pluginRegistry[entryPoint.name] = plugin
                logprint.exception('Failed to instantiate plugin %s' % entryPoint.name)
            except Exception:
                logprint.exception('Failed to parse plugin %s' % entryPoint.name)
```

As a further minor bug, Girder 3 no longer shows the failure messages from failed plugins on the web client (an easy change to the pug file will fix it).

jbeezley:
This throws an exception like this:
```python
Traceback (most recent call last):
  File "/home/jbeezley/git/girder3/env/bin/girder", line 11, in <module>
    load_entry_point('girder', 'console_scripts', 'girder')()
  File "/home/jbeezley/git/girder3/env/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/jbeezley/git/girder3/env/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/jbeezley/git/girder3/env/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jbeezley/git/girder3/env/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #2890, opened 2019-01-10
Fix showing plugin errors in the web client.

manthey:
This bug was due to a change between Girder 2 and 3.  The error button also showed the intended tooltip as a result title rather than as a tooltip.


manthey:
Based on discussions this closes #2889.

PR #2893, opened 2019-01-15
Remove static root setting

brianhelba:
The base URL where static assets are served (known in Webpack as the `publicPath`) was never designed to be determined at Javascript runtime. While Webpack includes a limited tool (the magic `__webpack_public_path__` variable) to defer setting `publicPath` within the Webpack config, it intractably breaks several types of asset loading to try to set this value at JS runtime (as opposed to just within the JS build process).

Previous attempts to work around this are summed up within the code comment:
> `publicPath` would normally be set in the Webpack config file, but is not known at compile time. Note that in theory, an ES6-style import of a file asset that occurred earlier in the import resolution sequence than this module could fail to have its `publicPath` set at all, but the typical style rules for ordering ES6-style imports ensure that this module will be loaded very early in the import sequence. However, if `App` startup code modifies the `staticRoot`, then any ES6-style imports will probably have the incorrect path. Ultimately though, most file asset imports will occur in Pug files via require-style imports, which happen at runtime, so they will always succeed.

However, as mentioned in the comment, the previous strategy did not work for some types of asset imports. Worse, it fundamentally broke using the Girder JS code with any other build system (which would not expect an attempt to set `__webpack_public_path__` at runtime), making it impossible to reuse Girder's client components in externally-built projects or to modernize external projects with non-Girder build systems.

The static root setting was added very early in Girder's development, long before Webpack was introduced. At this point in time, it is likely that Webpack technology is simply not compatible with a dynamic `staticRoot` setting.

If Girder's static asset libraries need to be served from an alternative location (e.g. a CDN), then Webpack's `output.publicPath` value should be changed at JS build configuration time, ideally via a plugin that modifies the state of Grunt's config and the output of Girder's server-side Mako templates.

An alternative to this change would be to keep the `staticRoot` setting, but require that it be passed to Grunt on every build. However, this would require the client build process to have database or API server access for any build, and would require that the web client be rebuilt after any change of the setting's value. For a setting that is changed as rarely and infrequently as `staticRoot`, this extra complexity is likely not worthwhile.

### DIFF
--- a/girder/utility/server.py
+++ b/girder/utility/server.py
@@ -21,7 +21,6 @@ import cherrypy
 import mako
 import mimetypes
 import os
-import posixpath
 import six
 
 import girder.events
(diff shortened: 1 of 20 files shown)


brianhelba:
This is still slightly WIP, but ready for initial comments and review.

zachmullen:
> If Girder's static asset libraries need to be served from an alternative location (e.g. a CDN), then Webpack's output.publicPath value should be changed at JS build configuration time, ideally via a plugin that modifies the state of Grunt's config and the output of Girder's server-side Mako templates.

We could just use an environment variable as the static root for both cases. I don't have strong feelings either way; future versions of Girder (beyond 3) will not contain any HTML rendering or static content building & serving logic anyway.

girder/utility/server.py
```diff
@@ -154,7 +117,6 @@ def configureServer(test=False, plugins=None, curConfig=None):
         'serverRoot': root,
         'serverRootPath': routeTable[constants.GIRDER_ROUTE_ID],
         'apiRoot': root.api.v1,
-        'staticRoot': routeTable[constants.GIRDER_STATIC_ROUTE_ID]
```
brianhelba: @girder/developers Should we continue to provide `info['staticRoot']` to plugins? It will always have the value `/static`.

girder/utility/server.py
```diff
@@ -218,7 +180,7 @@ def setup(test=False, plugins=None, curConfig=None):
         girderWebroot, str(routeTable[constants.GIRDER_ROUTE_ID]), appconf)
 
     # Mount static files
-    cherrypy.tree.mount(None, routeTable[constants.GIRDER_STATIC_ROUTE_ID],
+    cherrypy.tree.mount(None, '/static',
                         {'/':
                          {'tools.staticdir.on': True,
                           'tools.staticdir.dir': os.path.join(constants.STATIC_ROOT_DIR),
```
brianhelba: Below on line 235, could the conditional be simplified? Is there anything in `routeTable` that's not in `pluginWebroots` other than `constants.GIRDER_ROUTE_ID`?

zachmullen:
(And of course, if the env var isn't set, we can default to `"/static"`)

brianhelba:
Closing in favor of the approach in #2898, which preserves this setting.

PR #2896, opened 2019-01-16
Cache resource path lookups in girder mount.

manthey:
Resource path lookups hit the mongo database.  This caches them for up to 1 second.  A longer cache duration could be useful if we implemented a cache invalidation method, but this is a simple first improvement.

### DIFF
--- a/girder/cli/mount.py
+++ b/girder/cli/mount.py
@@ -100,9 +100,14 @@ class ServerFuse(fuse.Operations):
             else:
                 logger.debug('<- %s (length %d) %r', op, len(ret), ret[:16])
 
+    _getPathCacheMaxEntries = 100
+    _getPathCacheExpiryDuration = 1  # seconds
+    _getPathCache = {}
+
     def _getPath(self, path):
         """
-        Given a fuse path, return the associated resource.
+        Given a fuse path, return the associated resource.  This uses a simple
+        cache to avoid repeated lookups.
 
         :param path: path within the fuse.
         :returns: a Girder resource dictionary.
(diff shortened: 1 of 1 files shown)


girder/cli/mount.py
```diff
@@ -112,18 +117,30 @@ def _getPath(self, path):
...
+            resource = cacheEntry['result']
+            cacheEntry['hits'] += 1
+        else:
+            try:
+                # We can't filter the resource, since that removes files'
+                # assetstore information and users' size information.
+                resource = path_util.lookUpPath(
+                    path.rstrip('/'), filter=False, force=True)
+                if len(self._getPathCache) > self._getPathCacheMaxEntries:
+                    self._getPathCache.clear()
```
zachmullen: LRU behavior is probably preferable, but would require an additional data structure, so might not be worth the complexity.

Is there a reason not to make use of the cache utils present in Girder core?

manthey: Sure, LRU with TTL would be nicer, but this is simple and sufficient. 

I didn't use the dogpile Memory backend because it seemed like it was more complex, and, also, I don't think it would have a virtue to allow it to use a non-memory backend.

zachmullen: This is fine, as long as a sawtooth performance pattern is acceptable for users.

My one concern would be whether this behavior is sensible as a default, or whether this is something that should be configurable (e.g. env vars).

Do you have any measurements of the performance gain for your use cases?

danlamanna: > I don't think it would have a virtue to allow it to use a non-memory backend.

Like what?

manthey: It saves 10-11ms per served tile in one test using large_image.

manthey: Looking into using dogpile a bit further: 

One could add:
```
from dogpile.cache import make_region

_getPathCache = make_region().configure(backend='dogpile.cache.memory', expiration_time=1)

class ServerFuse ...

    @_getPathCache.cache_on_arguments()
    def _getPath(self, path):
        ...
```
which works.  But, dogpile's memory backend has [caveats](https://dogpilecache.sqlalchemy.org/en/latest/api.html):

"There is no size management, and values which are placed into the dictionary will remain until explicitly removed. Note that Dogpile’s expiration of items is based on timestamps and does not remove them from the cache."

which seem like a big enough issue not to use it out-of-the-box.

zachmullen: I guess it's too much to ask for a caching library to implement basic eviction policies in one of its core backends...

manthey: Others have discussed using cachetools with dogpile ([see here](https://bitbucket.org/zzzeek/dogpile.cache/pull-requests/32/add-a-cachetools-lru-lfu-in-memory-backend/diff)), but that went nowhere.  Since then dogpile.cache has been moved to sqlalchemy's organization and nothing changed.

brianhelba: I'm concerned that we're starting on the path to implementing and being forced to maintain a new Girder-specific code stack for what should be delegated to a mature upstream library. If the "cache" was just a global dict, then perhaps it wouldn't be worthwhile to pull in heavier dependencies, but even this implementation has eviction policies, size limits, etc. I suspect that we'll want even more sophistication and usage of caching in the future, leading to a situation where Girder's Python client and Python server have totally divergent caching technology.

If a well-established upstream library is lacking a necessary feature, I don't think the first-line solution should be to re-implement the entire system from scratch. We should probably implement a new Dogpile backend with the necessary feature(s), and perhaps publish it in a separate package / repo, so all of our projects (and even other external users) can benefit.

manthey: As always, there is a balance -- There are greater costs in developing things for other libraries.

How does dogpile's backend managing multi-threading, for instance?  Here, I am using a simple dictionary with specific access methods and can see at a glance that it will be robust in a multi-threaded environment without locks (assuming cpython, of course).  I know you can use a lock with dogpile, but acquiring and releasing a lock has a time cost, too.

We added dogpile specifically to try to offload caching behavior to an upstream library, but I don't observe that anything has actually used it since.  Were we premature in selecting it?  Should we have gone with plain cachetools instead?  If so, that requires using locks for thread safety which then involves local code anyway.

If we are trying to reduce maintenance burden, adding another repo is more cognitive burden.  This is worth it if it gets used multiple places but not if it ends up only being used by one spot in one project.

zachmullen: We shouldn't need to use locks for an in-memory cache implementation with cpython; even if we need to implement synchronized data structures, I think doing so in a way that we are confident does not release the GIL would be fairly straightforward. I tend to agree that the desire for caching utilities is general enough that we should create / leverage a more central solution instead of ad hoc versions all over the place (especially since we already kind of have an API for this in core). I would find it hard to believe if we had to implement LRU behavior ourselves rather than using a third party library, but if that's the case I'd have fun creating that abstraction.

manthey:
As a commentary on why caching in some form would be helpful: the openslide library is doing a getattr call on a file each tile it extracts a region (for large_image, a tile).  I've isolated it to the openslide C library itself.  Within a girder mount, getattr is passed the fuse file path, which is a Girder resource path.  This path is queried via girder (and hence via Mongo) to get the associated item and its attributes (size and date, for instance).

manthey:
There are more modern ways to do this.  Closing -- it may come back under a different form.

PR #2898, opened 2019-01-18
Set the static root at JS build time

brianhelba:
This is a follow-up to #2893:
> The base URL where static assets are served (known in Webpack as the `publicPath`) was never designed to be determined at Javascript runtime. While Webpack includes a limited tool (the magic `__webpack_public_path__` variable) to defer setting `publicPath` within the Webpack config, it intractably breaks several types of asset loading to try to set this value at JS runtime (as opposed to just within the JS build process).
>
> Previous attempts to work around this are summed up within the code comment:
>> `publicPath` would normally be set in the Webpack config file, but is not known at compile time. Note that in theory, an ES6-style import of a file asset that occurred earlier in the import resolution sequence than this module could fail to have its `publicPath` set at all, but the typical style rules for ordering ES6-style imports ensure that this module will be loaded very early in the import sequence. However, if `App` startup code modifies the `staticRoot`, then any ES6-style imports will probably have the incorrect path. Ultimately though, most file asset imports will occur in Pug files via require-style imports, which happen at runtime, so they will always succeed.
>
> However, as mentioned in the comment, the previous strategy did not work for some types of asset imports. Worse, it fundamentally broke using the Girder JS code with any other build system (which would not expect an attempt to set `__webpack_public_path__` at runtime), making it impossible to reuse Girder's client components in externally-built projects or to modernize external projects with non-Girder build systems.

This PR now causes `girder build` to pass the static root (as determined by the Girder database setting) to Webpack at build time.

The ability to set the static root at JS run time (via the `setStaticRoot` API or via a `<div id="g-global-info-staticroot">`) has been removed.

Finally, the `girder.utility.server.getStaticRoot` method has been changed to return an absolute URL. This is necessary, since when Webpack's `publicPath` is a relative URL, assets referenced within CSS files (e.g. some of the JSON metadata editor assets) are incorrectly resolved relative to the referring file.

### DIFF
--- a/docs/migration-guide.rst
+++ b/docs/migration-guide.rst
@@ -159,6 +159,19 @@ place (in the Girder python package) in both development and production installs
 The built assets are installed into a virtual environment specific static path
 ``{sys.prefix}/share/girder``.
 
+Static root is required during web client build
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The static root, indicating the base URL where web client files are served from,
+is now required when the web client is built. The primary implication is that if the static root
+setting is changed, the web client must be immediately rebuilt. Also, note the API changes:
+
+* In the web client, ``girder.rest.staticRoot``, ``girder.rest.getStaticRoot``, and
+  ``girder.rest.setStaticRoot`` have been removed
+* The ability to set the web client static root via the special element
+  ``<div id="g-global-info-staticroot">`` has been removed
+* In the server, the ``girder.utility.server.getStaticRoot()`` function now returns an absolute path
+  unless set otherwise (which is not recommended).
+
 Server changes
 ++++++++++++++
 
--- a/girder/web_client/test/testEnv.pug
+++ b/girder/web_client/test/testEnv.pug
@@ -8,7 +8,6 @@ html(lang='en')
 
   body
     #g-global-info-apiroot.hide %HOST%#{apiRoot}
-    #g-global-info-staticroot.hide %HOST%#{staticRoot}
 
     each js in jsFiles
       script(src=js)
--- a/girder/cli/build.py
+++ b/girder/cli/build.py
@@ -29,6 +29,7 @@ import six
 
 from girder.constants import STATIC_ROOT_DIR
 from girder.plugin import allPlugins, getPlugin
+from girder.utility.server import getStaticRoot
 
 # monkey patch shutil for python < 3
 if not six.PY3:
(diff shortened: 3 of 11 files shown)


brianhelba:
Note that if the static root is changed via the web GUI and the web client is not immediately rebuild, then the web client will break until an administrator with shell access runs `girder build`.

docs/migration-guide.rst
```diff
@@ -159,6 +159,19 @@ place (in the Girder python package) in both development and production installs
 The built assets are installed into a virtual environment specific static path
 ``{sys.prefix}/share/girder``.
 
+Static root is required during web client build
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The static root, indicating the base URL where web client files are served from,
+is now required when the web client is built. The primary implication is that the static root
```
jcfr: `The primary implication is that the static root` 
->
`The primary implication is that if the static root`

brianhelba: Will fix.

girder/web_client/test/testEnv.pug
```diff
@@ -8,7 +8,6 @@ html(lang='en')
 
   body
     #g-global-info-apiroot.hide %HOST%#{apiRoot}
-    #g-global-info-staticroot.hide %HOST%#{staticRoot}
```
zachmullen: This may be the reason all the web client tests are failing. We could just hard code for the test environment.

zachmullen:
I have an editable install of Girder locally, but updates I make to the web client files do not show up when I run `girder build && girder serve`, leading me to believe the paths aren't set up correctly. Here's the invocation:

```
grunt "--static-path=/Users/zach/.virtualenvs/girder/bin/../share/girder/static" "--static-url=/static" "--no-progress=false" "--env=prod"
```

Any idea what's wrong?

girder/web_client/test/testEnv.pug
brianhelba: This `<div>` should now be meaningless. The value of `/static` just needs to be provided to Webpack, which it ostensibly is.

I haven't had a chance to figure out why the tests actually are failing though.

brianhelba:
@zachmullen The failure was due to `updateWebpackConfig` clobbering the `output.filename` option. This fixed by [the subsequent commit](https://github.com/girder/girder/pull/2898/commits/da3465c22c2a3040a5513de9176a681184407001).

@girder/developers All tests are now passing, and this is ready for final review.

PR #2899, opened 2019-01-21
Split the core web client into its own package.json

brianhelba:
This will allow the code within the core web client to be published and reused by other projects with different build systems.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -455,18 +455,21 @@ should be done:
 
        npm install -g 'npm@>=5.6'
 
-2. Update ``dependencies`` or ``devDependencies`` in ``girder/web_client/package.json.template``
-   to add a new *abstract* specifier for the package:
+2. Update ``girder/web_client/package.json.template`` or ``girder/web_client/src/package.json`` to
+   add a new *abstract* specifier for the package:
 
-  * Packages that are bundled into the web client should generally use the
+  * Packages that are bundled into the web client must be listed under the ``dependencies`` field
+    of ``girder/web_client/src/package.json`` and should generally use the
     `tilde range <https://www.npmjs.com/package/semver#tilde-ranges-123-12-1>`_
     to specify versions.
-  * Packages that are part of the build or testing process should generally use the
+  * Packages that are part of the build or testing process should be listed under either the
+    ``dependencies`` or ``devDependencies`` fields of ``girder/web_client/package.json.template``
+    and should generally use the
     `caret range <https://www.npmjs.com/package/semver#caret-ranges-123-025-004>`_
     to specify versions.
 
-If updating node libraries related to linting or documentation, you should instead modify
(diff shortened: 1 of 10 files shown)


docs/development.rst
```diff
@@ -455,18 +455,21 @@ should be done:
...
        npm install -g 'npm@>=5.6'
 
-2. Update ``dependencies`` or ``devDependencies`` in ``girder/web_client/package.json.template``
-   to add a new *abstract* specifier for the package:
+2. Update ``girder/web_client/package.json.template`` or ``girder/web_client/src/package.json`` to
+   add a new *abstract* specifier for the package:
 
-  * Packages that are bundled into the web client should generally use the
+  * Packages that are bundled into the web client must be listed under the ``dependencies`` field
+    of ``girder/web_client/src/package.json`` and should generally use the
```
zachmullen: Should this now just be done with the `npm add` command?

brianhelba: There is no `npm add`, just an `npm install --save`. We don't want that, because there's no need to create a `node_modules` or `package-lock.json` within the `girder/web_client/src/` directory, which the command will always do. Since we don't need the functionality to keep those things in sync, hand-editing the file is going to be much easier than working with the statefulness of `npm install`.

zachmullen: Makes sense, thanks.

brianhelba:
@girder/developers I'm having trouble reproducing the failing tests locally.

zachmullen:
The `girder build --dev` command fails silently (which is a separate bug).

screen shot 2019-01-22 at 8 45 46 am

girder/web_client/grunt_tasks/webpack.config.js
```diff
@@ -19,12 +19,9 @@
  * the DllPlugin for dynamic loading, each individual bundle has its own config options
  * that can extend these.
  */
-var path = require('path');
```
zachmullen: This line needs to be restored

brianhelba:
@zachmullen For some odd reason, [another build of the same code](https://circleci.com/gh/girder/girder/16752) (missing the `require('path')`), had the build step succeed without that error.

I've just fixed that particular issue, which was introduced when I fixed an issue with my original definition for `loaderPaths`.

At this point, even with the fix, the build is still failing. I'll continue to investigate, but again, any help is appreciated.

brianhelba:
I've managed to fix an issue affecting test loading. Without this fix, the build works for normal usage, but not most tests.

However, the tests still seem to be broken on the "integrationTests" builds. I'm investigating now to see what's different about that environment to cause the web client build to be different (`girder_lib.min.js` is a smaller size when built in that environment).

brianhelba:
I think the problem is that the machine executors on CircleCI are using NodeJS 6. Upgrading to Node 10 seems to fix the build.

brianhelba:
@girder/developers Tests are now all passing.

PR #2904, opened 2019-01-23
Add a fixture to create a server with an exposed port.

manthey:
This will have a new exposed port per instance.  It does not spawn a process.

The main expected utility of this is to do pytest with a running girder-worker so we can get results back from girder-worker.

This supercedes PR #2760, though it is not an exact replacement.  Rather, this still runs in the same process so it has less isolation and less overhead than #2760.


pytest_girder/pytest_girder/fixtures.py
```diff
@@ -130,9 +138,40 @@ def server(db, request):
...
+def server(db, request):
+    """
+    Require a CherryPy embedded server.
+
+    Provides a started CherryPy embedded server with a request method for
+    performing local requests against it. Note: this fixture requires the db
+    fixture.
+    """
+    for server in _yieldServer(request):
+        yield server
```
zachmullen: Is the loop necessary, doesn't `_yieldServer` only yield a single thing?

pytest_girder/pytest_girder/fixtures.py
```diff
@@ -130,9 +138,40 @@ def server(db, request):
...
+
+    Provides a started CherryPy server with an exposed port and a request
+    method for performing local requests against it. Note: this fixture
+    requires the db fixture.  The returned value has an `exposedPort` property
+    identifying where the server can be reached.  The server can then be
+    accessed via http via an address like
+    `'http://127.0.0.1:%d/api/v1/...' % exposedServer.exposedPort`.
+    """
+    for server in _yieldServer(request, exposePort=True):
+        yield server
```
zachmullen: Same here?

pytest_girder/pytest_girder/fixtures.py
manthey: Somehow, pytest doesn't actually do things correctly unless the pytest.fixture uses the yield keyword itself.  Pythonically, `return _yieldServer(request)` works, but pytest doesn't work for me when I do that.  Maybe I'm missing something.

zachmullen: What about `yield _yieldServer(request)`?

manthey: That returns a generator of a generator not a generator itself.

manthey: And `next(_yieldServer(request))` never calls the cleanup code of _yieldserver since it doesn't ask for the final (non existent) element.

zachmullen: Hm, carry on then!

manthey: One could do:
```
server = _yieldServer(request)
yield next(server)
next(server)
```

pytest_girder/pytest_girder/fixtures.py
```diff
@@ -130,9 +138,40 @@ def server(db, request):
...
+    Provides a started CherryPy embedded server with a request method for
+    performing local requests against it. Note: this fixture requires the db
+    fixture.
+    """
+    for server in _yieldServer(request):
+        yield server
+
+
+@pytest.fixture
+def exposedServer(db, request):
```
zachmullen: Maybe `boundServer` would be a better name?

manthey: Renamed.

manthey:
@zachmullen I renamed the fixture, so you'll need to approve this again.

PR #2908, opened 2019-01-24
Include web client fonts via an npm package, instead of a Webpack plugin

brianhelba:


brianhelba:
For reference, I chose to reference this file within `App.js` because:
* The `Open Sans` fonts are referenced via `global.styl`, which itself is referenced within `App.js`
* All other global CSS rules (for Boostrap, etc.) are referenced within `App.js`

PR #2909, opened 2019-01-24
Distribute Fontello files via an npm package

brianhelba:


brianhelba:
This still needs documentation on the update and release process for `@girder/fontello`, but I'd like to get feedback from @girder/developers that this is a good idea at all.

jbeezley:
The effectively reverts https://github.com/girder/girder/pull/1550, which removed grunt-fontello due to how many problems we had with it.  I don't recall all of the details, but it resulted frequently failing builds.  I'm fine with trying it again to see if it is better now, but we should be prepared to change back if necessary.

brianhelba:
@jbeezley Not quite. The difference is that we run `grunt-fontello` once, as part of the build step of `@girder/fontello`, then we push the resulting font files to npm. `@girder/core` (or really `@girder/meta-build` will fetch the font files via `npm install`. Compared to the current system, this has the advantage of using npm's CDN instead of `data.kitware.com` and a much simpler installation process.

jbeezley:
Oh, I see.  Yes, this is better.

brianhelba:
@jbeezley I've added a `README.md` file to the new `@girder/fontello` package. Can you QA it and re-review?

brianhelba:
Fixed spelling issues in README.

PR #2911, opened 2019-01-25
Remove already-deprecated support for "path" parameter in restRequest

brianhelba:
* Fix an outdated usage of restRequest arguments
* Remove already-deprecated support for "path" parameter in restRequest

### DIFF
--- a/docs/migration-guide.rst
+++ b/docs/migration-guide.rst
@@ -139,6 +139,9 @@ Other backwards incompatible changes affecting plugins
         def load(self, info):
             ModelImporter.registerModel('job', Job, 'jobs')
 
+* In the web client, ``girder.rest.restRequest`` no longer accepts the deprecated ``path``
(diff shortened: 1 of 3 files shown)


docs/migration-guide.rst
```diff
@@ -139,6 +139,9 @@ Other backwards incompatible changes affecting plugins
         def load(self, info):
             ModelImporter.registerModel('job', Job, 'jobs')
 
+* In the web client, ``girder.rest.restRequest`` no longer accepts the deprecated ``path``
```
zachmullen: We should put this under a new heading for web client changes

brianhelba: @zachmullen I agree, but can we do this in another PR for reorganizing the migration guide? Right now, this change belongs under "Other backwards incompatible changes affecting plugins" as much as changes to mail template detection, etc.

PR #2912, opened 2019-01-25
Replace Javascript imports of "girder" with "@girder/core"

brianhelba:
* Replace Javascript imports of "girder" with "@girder/core"
* Document the change of "girder" to "@girder/core"
* Update the client to only use the "girder": "@girder/core" alias for plugins

Since the npm package containing the Girder core Javascript code is now named `@girder/core`, all internal references should use this new name. Otherwise, third party packages which depend on `@girder/core` will fail to build it without adding their own alias rules.

However, note that for now, the rule aliasing `girder` to `@girder/core` remains in place for plugins, making this a deprecation and not a breaking change for downstreams. New Javascript code within core Girder will have to use the new syntax, which is intended.

Considering that downstream plugins already have to change the JS import syntax for other plugins, perhaps it's better to force all of the changes at once in the Girder 3 release, in which case we should remove the alias for plugins now too.

### DIFF
--- a/girder/web_client/grunt_tasks/build.js
+++ b/girder/web_client/grunt_tasks/build.js
@@ -282,6 +282,12 @@ module.exports = function (grunt) {
             module: {
                 rules: [babelRule]
             },
+            resolve: {
+                alias: {
+                    // This is deprecated, but kept in place for compatibility
+                    'girder': '@girder/core'
+                }
+            },
             plugins: [
                 new webpack.DllPlugin({
                     path: path.join(grunt.config.get('builtPath'), 'plugins', name, 'plugin-manifest.json'),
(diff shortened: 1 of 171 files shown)


brianhelba:
@girder/developers Can someone please review this?

girder/web_client/grunt_tasks/build.js
```diff
@@ -282,6 +282,12 @@ module.exports = function (grunt) {
             module: {
                 rules: [babelRule]
             },
+            resolve: {
+                alias: {
+                    // This is deprecated, but kept in place for compatibility
+                    'girder': '@girder/core'
```
zachmullen: Does this mean this is not actually a breaking change?

brianhelba: Not for downstream plugins.

zachmullen: Sorry, you're saying it *is* breaking for downstream plugins, or it's *not*?

brianhelba: However, as noted, we might want to intentionally make this a breaking change, if we decide that the 2->3 release is a more opportune time to require that downstreams start using the 'correct' core package name (since they're already changing plugins).

PR #2913, opened 2019-01-26
Fix warnings for deprecated "Collection.remove" PyMongo method

brianhelba:
The `Cursor.count` method is also deprecated, but that will be significantly more work to resolve.


brianhelba:
@girder/developers Can someone please review this?

PR #2914, opened 2019-01-27
[WIP] Support web client tests in external plugins

zachmullen:
As of 4e5436b, the following test in core now works:

```python
import os
import pytest
from girder.constants import PACKAGE_DIR
from pytest_girder.utils import runWebClientTest


@pytest.mark.parametrize('spec', [
    'versionSpec.js'
])
def testWebClient(spec):
    runWebClientTest(os.path.join(PACKAGE_DIR, 'web_client', 'test', 'spec', spec))
```

I don't know if it's within scope of this PR to destroy the CTest-harnessed web client testing infrastructure, but I'd be happy to do so. That would involve rethinking how to divide up our tests across CircleCI slots.


zachmullen:
On second thought, I don't want to try to destroy the CTest infrastructure on this PR, it would be a pretty big undertaking and probably not worth it.

zachmullen:
I created a [minimal test plugin](https://data.kitware.com/#item/5c4f575d8d777f072b1b52d2) to demonstrate this working. If you install the pytest-girder package from this branch, the following should work:

```bash
cd test-webclient
pip install -e .
girder build --dev
pytest
```

pytest_girder/pytest_girder/utils.py
```diff
@@ -1,16 +1,28 @@
...
 
+from girder.api import access
+from girder.api.describe import describeRoute, Description
+from girder.api.rest import Resource
+from girder.constants import ROOT_DIR, registerAccessFlag
+from girder.exceptions import RestException
+from girder.models.folder import Folder
+from girder.models.upload import Upload
+from girder.plugin import getPlugin
+from girder.utility.progress import ProgressContext
```
jbeezley: In the past, we had to put girder imports inside functions in pytest_girder otherwise the coverage numbers were incorrect.  Is this still the case?

pytest_girder/pytest_girder/fixtures.py
```diff
@@ -77,70 +74,18 @@ def db(request):
         pymongo.MongoClient = realMongoClient
 
 
-def _findFreePort():
-    with contextlib.closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as sock:
-        sock.bind(('', 0))
-        return sock.getsockname()[1]
+def _getPluginsFromMarker(request, registry):
+    plugins = []
```
jbeezley: What is this for?

zachmullen: It's a helper function for extracting the enabled plugins from pytest markers on a node

jbeezley: Oh, I see.  The diff is split.  I just saw this line and was confused.

jbeezley:
I don't really like putting this here because it adds to the confusion between unit and end to end tests.  I think much of the intent in the movement to pytest was to simplify what our tests do and to make them more "unit-like".

Part of me wants to insist on this being moved to a different python package, but I'm also uncomfortable adding new packages before we have automated releases built into CI.  As a compromise, I would suggest that anything related to web client testing should be moved to a new file to make the distinction more clear.

pytest_girder/pytest_girder/utils.py
jbeezley: You won't see it in a plugin.  This issue is that the import is executed *before* pytest starts tracking coverage.  Then when the test infrastructure to started up and these packages are imported by core, the cached module prevents the top-level code (and anything executed as a side-effect) from showing up as covered.

zachmullen: Is this a documented limitation of pytest-cov? It seems like a pretty big hassle...

jbeezley: Not really, we just have a weird use case.  We are using the library we are testing inside a pytest plugin that we are using to test it.

jbeezley: We might be able to get around it by adding global fixture that deletes `sys.modules['girder']` before running the test.  I'm not sure what other consequences this might have.

zachmullen:
> think much of the intent in the movement to pytest was to simplify what our tests do and to make them more "unit-like".

Another major reason for this switch was to move away from CTest as a harness, which is what this PR is about. This package makes the most sense *right now* as the place for meeting the downstream needs of end-to-end web client testing that *right now* is the norm.

I'm happy to put the web client test code in a separate module in pytest_girder, and certainly happy to have it go away in the future. I'd prefer not to have to deal with the associated overhead of having a separate package for this. I think once we finally reach a world where we realize the ideal of more lightweight testing, pytest_girder will be changed a great deal anyway, as a reflection of changes to Girder itself.

pytest_girder/pytest_girder/utils.py
zachmullen: Oh I see, so this pattern would only be necessary for stuff in `pytest_girder`, not in the test files themselves. I'm fine with that, I will change these back.

PR #2916, opened 2019-01-28
Require that downstream plugins use '@girder/core' for JS imports

brianhelba:
Previously, plugins building with Girder's JS build system could use the legacy `girder` package name for imports too.

### DIFF
--- a/docs/migration-guide.rst
+++ b/docs/migration-guide.rst
@@ -101,17 +101,17 @@ plugin:
 * Migrate all imports in Python and Javascript source files.  The old plugin module paths are no longer
   valid.  Any import reference to:
 
-  * ``girder.plugins`` in Python should be changed to the actual installed module name
+  * ``girder.plugins`` in Python must be changed to the actual installed module name
 
     * For example, change ``from girder.plugins.jobs.models.job import Job`` to
       ``from girder_jobs.models.job import Job``
 
-  * ``girder_plugins`` in Javascript should be changed to the actual installed package name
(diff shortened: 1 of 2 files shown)


zachmullen:
It should be noted that this is an RFC, designed to generate discussion, not something designed to be merged necessarily.

brianhelba:
@girder/developers This is a separate PR from #2912 to provide an RFC on the breaking change here.

As noted in the last PR:
> Considering that downstream plugins already have to change the JS import syntax for other plugins, perhaps it's better to force all of the changes at once in the Girder 3 release, in which case we should remove the alias for plugins now too.

manthey:
I am for this change.

On a related note, I've found that in some instances I seem to need to add "@girder/core" to the package.json dependencies of a plugin.  It might work (though I haven't tried it) to add it to the peerDependencies.  Specifically, this happens if I try to do `girder build --dev` in a tox generated environment; if it is not listed it as a dependency, the build fails due to failure to find node modules that are required by @girder/core (like moment or backbone).

jbeezley:
Yes, if a plugin uses anything from `@girder/core` it should include it as peer dependencies, otherwise there is a possibility that the core modules will be duplicated in the plugin's bundle and wrapping views will be broken.  This means all plugins should put it in their `package.json`, including those in this repo.

This isn't actually an issue with this change set, rather it comes from https://github.com/girder/girder/pull/2899.  We should go through all G3 compatible plugins (as well as the cookiecutter) and make the change.

brianhelba:
@girder/developers It seems the consensus in favor of this. Merging.

PR #2917, opened 2019-01-28
Update CI Docker image

brianhelba:
* Fetch CMake from GitHub when building CI Docker image
* Install Yarn in CI docker image

GitHub is now the official location for new CMake releases and should have higher availability  then the existing cmake.org servers. Yarn can be useful as a tool for downstreams building CI on Girder's stack.


brianhelba:
@girder/developers PTAL.

PR #2922, opened 2019-01-30
Refactor and simplify Girder version handling

### LINKED ISSUE #1933: Generate version info on server side rather than client build time

brianhelba:
This makes the following significant changes:
* Only the base "release" and full Git "sha" are included with version information on both web client and server. The web client also includes its build date.
* The web client and server are now fully decoupled. Particularly, the server no longer depends on a web client build to report its own version.
* The web client can now be used without errors by external build systems which don't provide version information.
* Building the web client in a non-Git environment no longer emits error messages.

Fixes #1933


brianhelba:
This is still a WIP, as I need to finish adding Git SHA support to the server.

This PR also brings us much to being able to do continuous release of at least the Python package. However, that would likely change and simplify the version specification even further, by always reporting the version as a combined base release number with (if using a non-release commit) an appended short SHA. The reporting of the full Git SHA component of the version would likely be removed.

brianhelba:
@girder/developers This is ready for review. PTAL.

#2932 is necessary for the failing test to pass.

girder/constants.py
```diff
@@ -41,18 +43,22 @@
...
+        ['git', 'rev-parse', 'HEAD'],
+        cwd=ROOT_DIR
+    ).decode().strip()
+except subprocess.CalledProcessError:
+    # "gitSha" is None
     pass
+VERSION = {
+    # Set by __init__.py
+    'release': girder.__version__,
+    # TODO: set this
```
zachmullen: TODO?

brianhelba: That's no longer TODO, but I forgot to remove the comment. Fixed.

brianhelba:
@girder/developers PTAL again.

PR #2929, opened 2019-02-01
Move setup_database into the pytest_module.

manthey:
This renames most parameters to make them private.  This can still be used by non-pytest tests, but can now be used within a pytest.

For instance,
```
with warnings.catch_warnings():
    warnings.filterwarnings('ignore', 'setup_database.*')
    from pytest_girder.setup_database import setupDatabase

def testWebClientWithAnnotation(boundServer, fsAssetstore, db):
    spec = <path to specFile.js>
    setupFile = os.path.splitext(spec)[0] + '.yml'
    if os.path.exists(setupFile):
        setupDatabase(setupFile)
    runWebClientTest(boundServer, spec)

```

### DIFF
--- a/pytest_girder/pytest_girder/setup_database.py
+++ b/pytest_girder/pytest_girder/setup_database.py
@@ -33,23 +33,23 @@ warnings.warn(
 )
 
 #: A prefix for all relative file paths
-prefix = '.'
+_prefix = '.'
 
 
-def resolvePath(path):
+def _resolvePath(path):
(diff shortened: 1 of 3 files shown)


jbeezley:
With the pytest infrastructure, writing custom fixtures to do this is pretty straightforward and much more flexible.  I would vote against including it here.

manthey:
I'm persuaded to drop this approach.

PR #2930, opened 2019-02-01
Setting params should be specified in request body

zachmullen:


danlamanna:
I pushed a change, won't be able to test it for a little bit.

zachmullen:
Thanks. FWIW, this shouldn't be a breaking change, since our endpoints don't typically care how the params are encoded, it really just affects swagger.

zachmullen:
@danlamanna can you approve this PR?

PR #2932, opened 2019-02-08
Use universal-ctags to parse public symbols

brianhelba:
Unlike exuberant-ctags, universal-ctags is actively maintained and parses Python files more accurately. In particular, keyword arguments are no longer parsed as variable declarations.

### DIFF
--- a/scripts/publicNames.txt
+++ b/scripts/publicNames.txt
@@ -73,7 +73,6 @@ clients
             cli
                 GirderCli
                     sendRestRequest
-                help
                 main
 girder
     LogFormatter
@@ -343,9 +342,7 @@ girder
     auditLogger
     cli
         build
-            help
             main
-        context_settings
         main
         mount
             FUSELogError
@@ -363,17 +360,13 @@ girder
                 readdir
                 release
                 use_ns
-            help
             main
             mountServer
-            show_default
             unmountServer
         serve
             main
         sftpd
             DEFAULT_PORT
-            default
-            help
             main
         shell
             main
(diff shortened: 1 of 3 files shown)


brianhelba:
@girder/developers PTAL.

Note, [the final commit](https://github.com/girder/girder/pull/2932/commits/4edcfc5a078b66a7d61885faf093de9dc1499c48) will be removed prior to merging, due to the way that the CI Docker image is built and used.

scripts/publicNames.txt
```diff
@@ -507,7 +500,9 @@ girder
         mongodb_proxy
             EXECUTABLE_MONGO_METHODS
             Executable
+            MongoClient
             MongoProxy
+            MongoReplicaSetClient
```
zachmullen: I'm dubious that these belong here since they are spiritually the  same as imports, here's the code snippet in question:

```python
try:
    # will fail to import from older versions of pymongo
    from pymongo import MongoClient, MongoReplicaSetClient
except ImportError:
    MongoClient, MongoReplicaSetClient = None, None
```

brianhelba: Yes, I've seen, and I'd consider it a false positive. However, the tool is emitting them as ordinary variables (without any additional type flags), so unless we want to build a whitelisting mechanism (at which point, we've started building our own linter from scratch), we are stuck with them.

The old tool was emitting keyword arguments as variables too, and nobody noticed. It seems like the point of the list is mostly to flag deltas, and is less so an absolute reference.

zachmullen: > It seems like the point of the list is mostly to flag deltas, and is less so an absolute reference.

This is true. There will be a human in the loop to review if these change anyway.

PR #2934, opened 2019-02-11
Use the real path for web builds.

manthey:
In a virtual Python 2.7 environment, with Girder installed in non-editable mode, the web_client path contains `<env root>/local/lib/python2.7/site-packages/girder/web_client`, but `local/lib` is a symlink to lib.  When npm installs packages, it fails to put the packages in the expect spot, so, for instance, underscore ends up in `.../web_client/node_modules/@girder/core/node_modules/underscore` rather
than at `.../web_client/node_modules/underscore`.  This causes the web client build to fail, as webpack fails to locate some modules.

### DIFF
--- a/girder/cli/build.py
+++ b/girder/cli/build.py
@@ -35,7 +35,7 @@ from girder.utility.server import getStaticRoot
 if not six.PY3:
     import shutilwhich  # noqa
 
-_GIRDER_BUILD_ASSETS_PATH = os.path.realpath(resource_filename('girder', 'web_client'))
+_GIRDER_BUILD_ASSETS_PATH = resource_filename('girder', 'web_client')
 
 
 @click.command(name='build', help='Build web client static assets.')
--- a/girder/cli/build.py
+++ b/girder/cli/build.py
@@ -35,7 +35,7 @@ from girder.utility.server import getStaticRoot
 if not six.PY3:
     import shutilwhich  # noqa
 
-_GIRDER_BUILD_ASSETS_PATH = resource_filename('girder', 'web_client')
+_GIRDER_BUILD_ASSETS_PATH = os.path.realpath(resource_filename('girder', 'web_client'))
 
 
 @click.command(name='build', help='Build web client static assets.')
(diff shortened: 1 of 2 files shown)


jbeezley:
While this change should be fine, I don't think it actually fixes the underlying problem.   Code using underscore from `@girder/core` will resolve it just fine in either place.  If something else is using underscore (a plugin or the build meta-package), then it should request it is a dependency as well.  You can't rely on any npm behavior that moves a dependency outside of `node_modules` in the package that requests it.

manthey:
`girder/web_client/Gruntfile.js` uses underscore, so perhaps it should be in the primary package.json rather than in the src/package.json.  However, doing that just made the build process complain about not finding jquery.

jbeezley:
That's due to [this line](https://github.com/girder/girder/blob/master/girder/web_client/grunt_tasks/webpack.config.js#L205).  Being pedantic, the correct fix is to always specify jquery as a "peerDependency" rather than use a resolve alias.  Realistically though, I don't think that will work for a third-party jquery plugin (if it includes jquery as a normal dependency).  I think the best way to fix this is just to add jquery to the meta-package as well.

manthey:
Reverting the path change and adding underscore and jquery to the package template then fails on `moment` in the jobs plugin, so this still isn't sufficient.

manthey:
And, if `moment` is added to the meta-package, then an external plugin (large_image) with fails to resolve `backbone`, even though the plugin requires `@girder/core` as a peer dependency.

jbeezley:
So it's a house of cards then... This is probably going to happen for a lot of plugins that assume they are using dependencies from core girder.  The rules for how things get moved around in `node_modules` isn't stable AFAIK.  The only way to guarantee a package is resolvable is to include it in the every `package.json` using it.

We can go back to installing from the real path. There is no guarantee that the same won't happen again on some OS/npm/node combination for some other reason, but we can tackle it if/when it occurs.

manthey:
Ugh.  I've reverted to just modifying the path via `realpath`.

PR #2943, opened 2019-02-28
Lazy-compute git version info

zachmullen:
This also fixes an issue where stderr from the git subprocess
was sent to the parent process stderr. Now it is ignored.


zachmullen:
Closing this, it's an improvement but we can do something totally different that will be friendlier.

PR #2944, opened 2019-02-28
RFC: Remove git version computation

zachmullen:


zachmullen:
The Backbone client now grabs the git sha at build time anyway, so I don't even know where this is being used.

zachmullen:
@girder/developers @girder/core this has sat here for a little while now. Does anyone object to it? If not, can we mark approval?

danlamanna:
I was waiting for @brianhelba to chime in since he changed it last.

jcfr:
If you plan to distribute python package, I suggest to use https://github.com/warner/python-versioneer

It also support nicely the integration of the version information in the source distribution

jbeezley:
@jcfr Would you recommend python-versioneer over [setuptools-scm](https://github.com/pypa/setuptools_scm/)?  It looks like python-versioneer doesn't have any recent commits.  Also, setuptools-scm is hosted in pypa's github org, so it would seem to be more official.

zachmullen:
If anyone likes this change and thinks it should be merged, go ahead and mark approval. If I don't hear any disapproval before the end of the day today, I will merge.

PR #2946, opened 2019-03-04
Fix item/{id}/copy endpoint when folderId is specified.

### LINKED ISSUE #2945: Issue with POST item/{id}/copy
`POST item/{id}/copy` doesn't work as expected when a `folderId` is specified via swagger.  Specifically, the `folderId` parameter is listed as a path parameter, but the route only takes an item `{id}` parameter.  Either the `folderId` should be specified as a `paramType` other than path (probably `query`), or another route should be created as `item/{id}/copy/{folderId}`.

manthey:
This fixes #2945.


manthey:
@zachmullen I think this bug was introduced back when the endpoint was switched from `describeRoute` to `autoDescribeRoute`.

PR #2953, opened 2019-03-08
Fixes #2952. Use static root correctly in build

### LINKED ISSUE #2952: "girder build" is failing to set static root
Following #2948, running `girder build` gives stdout looking like:
```bash
$ girder build 
...
> @girder/meta-build@ build /home/ubuntu/.venvs/girder/lib/python2.7/site-packages/girder/web_client
> grunt "--static-path=/home/ubuntu/.venvs/girder/share/girder/static" "--static-url=None" "--no-progress=false" "--env=prod"
```
and Webpack actually emits files with internal URL references like `None/built/assets/open-sans-latin-300-521d17bc.woff`

This is on a deployment (with a running database, installed `girder`, etc.) where `girder build` was working properly prior to pulling in #2948.

zachmullen:


zachmullen:
I assumed any bugs in #2948 would have been caught by regression tests, but it seems none of our tests depend on any static assets besides just the javascript and CSS files being in the right place. Not sure if it's worth trying to add a test for that at this point, we've removed the [packaging tests](https://github.com/girder/girder/blob/2.x-maintenance/tests/packaging/pip_install_girder.sh) that might have been a logical place for that kind of thing.

brianhelba:
@zachmullen A packaging test would not catch this, nor would one that checks whether the assets referenced by HTML are downloadable (since the HTML references to `staticRoot` are filled by the Python server template). It's only within inter-asset references (the best examples are `url(...)` expressions in CSS and `import` of image assets within JS) that `girder build` (as currently implemented) needs `staticRoot` to set Webpack's `publicPath`.

zachmullen:
True. It'd have to be checked in phantom or we could do some hacky inspection of the bundle... I'm not interested in doing either at this moment.

PR #2956, opened 2019-03-08
Publish python packages from CI

jbeezley:
This is pretty straightforward.  It adds:
* a bunch of boilerplate added to each setup.py file
* a tox environment for generating and pushing the releases
* a CI step that is run after tests on master and tags matching `/^v.*/` that runs `tox -e release`

I also had to add my own pypi credentials to the circle ci environment... There doesn't seem to be any kind of api key or token based auth.  :(

I verified this works as well as I can, but the only test it is to merge.

### DIFF
--- a/clients/python/setup.py
+++ b/clients/python/setup.py
@@ -18,8 +18,25 @@
 ###############################################################################
 
 import os
-import re
+
 from setuptools import setup, find_packages
+
+
+def prerelease_local_scheme(version):
+    """Return local scheme version unless building on master in CircleCI.
+    This function returns the local scheme version number
+    (e.g. 0.0.0.dev<N>+g<HASH>) unless building on CircleCI for a
+    pre-release in which case it ignores the hash and produces a
+    PEP440 compliant pre-release version number (e.g. 0.0.0.dev<N>).
+    """
+
+    from setuptools_scm.version import get_local_node_and_date
+
+    if os.getenv('CIRCLE_BRANCH') == 'master':
+        return ''
+    else:
+        return get_local_node_and_date(version)
+
 
 install_reqs = [
     'click>=6.7',
@@ -31,16 +48,11 @@ install_reqs = [
 with open('README.rst') as f:
     readme = f.read()
 
-init = os.path.join(os.path.dirname(__file__), 'girder_client', '__init__.py')
-with open(init) as fd:
-    version = re.search(
-        r'^__version__\s*=\s*[\'"]([^\'"]*)[\'"]',
-        fd.read(), re.MULTILINE).group(1)
-
 # perform the install
 setup(
     name='girder-client',
-    version=version,
+    use_scm_version={'root': '../..', 'local_scheme': prerelease_local_scheme},
+    setup_requires=['setuptools-scm'],
     description='Python client for interacting with Girder servers',
     long_description=readme,
     author='Kitware, Inc.',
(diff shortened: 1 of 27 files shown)


clients/python/setup.py
```diff
@@ -18,9 +18,27 @@
...
+    This function returns the local scheme version number
+    (e.g. 0.0.0.dev<N>+g<HASH>) unless building on CircleCI for a
+    pre-release in which case it ignores the hash and produces a
+    PEP440 compliant pre-release version number (e.g. 0.0.0.dev<N>).
+    """
+
+    from setuptools_scm.version import get_local_node_and_date
+
+    if 'CIRCLE_BRANCH' in os.environ and \
+       os.environ['CIRCLE_BRANCH'] == 'master':
```
zachmullen: if you wanted to avoid the backslash, this could just be 

```python
if os.environ.get('CIRCLE_BRANCH') == 'master':

File: girder/__init__.py [hunk 2]
```diff
@@ -29,6 +36,7 @@
 import sys
 import traceback
 
+
```
zachmullen: This extra newline may or may not cause pep8 failure

pytest_girder/setup.py
```diff
@@ -1,8 +1,29 @@
...
+    PEP440 compliant pre-release version number (e.g. 0.0.0.dev<N>).
+    """
+
+    from setuptools_scm.version import get_local_node_and_date
+
+    if 'CIRCLE_BRANCH' in os.environ and \
+       os.environ['CIRCLE_BRANCH'] == 'master':
+        return ''
+    else:
+        return get_local_node_and_date(version)
```
zachmullen: Could we put this function as a util inside the girder package to avoid repeating it in every plugin?

manthey:
What does adding setuptools-git do?  I note that we aren't using it in girder_worker (and I didn't in large_image).

pytest_girder/setup.py
jbeezley: Not really without going to crazy town.  That would require adding girder to `setup_requires`.  And the import to the function would have to be guarded in a `try` block.

girder/__init__.py
jbeezley: It's required as far as I know.

jbeezley:
I added it a few PR's ago.  It is necessary for girder plugins that include web_client assets that don't have a manifest file.

girder/__init__.py
zachmullen: It's not, but it looks like it's fine by the linter.

pytest_girder/setup.py
zachmullen: Makes sense. Should we add this block to the plugin cookiecutter template?

girder/__init__.py
jbeezley: Oh, I was looking at the wrong line... I'll amend it.

pytest_girder/setup.py
jbeezley: We could, but the cookiecutter doesn't include any CI/release infrastructure anyway.  I'm relatively indifferent.

jbeezley: There is the possibility of a `setuptools-scm` plugin to deduplicate that function, but we can do that later.

manthey: I don't think it belongs in the cookiecutter template because I don't think plugins should default to using setuptools-scm and circle-ci.

zachmullen: Agreed. Let's not add it there.

clients/python/setup.py
jbeezley: Changed to `os.getenv`.

PR #2957, opened 2019-03-09
Execute CI workflow on git tags for release

jbeezley:
PyPI publishing is working, but we need to execute the workflow when tags are pushed to generate new releases.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -506,7 +506,7 @@ jobs:
           name: Activate virtual environment
           command: echo ". $CIRCLE_WORKING_DIRECTORY/girder_env/bin/activate" >> $BASH_ENV
       - run:
-          name: Install twine
+          name: Install tox
           command: pip install --upgrade pip tox
       - run:
           name: Publish python packages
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -545,26 +545,53 @@ workflows:
   version: 2
   test_all:
     jobs:
-      - py2_serverInstall_serverTest
-      - py2_integrationTests
+      - py2_serverInstall_serverTest:
+          filters:
+            tags:
+              only: /^v.*/
```
zachmullen: Maybe `/^v[0-9]+.*/` for specificity?

(Minor nit, it seems like either the `^` or the `.*` should be unnecessary, depending on the matching behavior on CircleCI.)

zachmullen:
I think it may have been #2956 that introduced this problem, but the docker build is failing related to setuptools-scm:

```
Step 17/19 : RUN pip install --upgrade --upgrade-strategy eager --editable .
---> Running in f723032903e0
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Obtaining file:///girder
Complete output from command python setup.py egg_info:
File "<string>", line 1, in <module>
File "/girder/setup.py", line 139, in <module>
'build = girder.cli.build:main'
File "/usr/local/lib/python2.7/dist-packages/setuptools/__init__.py", line 145, in setup
return distutils.core.setup(**attrs)
File "/usr/lib/python2.7/distutils/core.py", line 111, in setup
_setup_distribution = dist = klass(attrs)
File "/usr/local/lib/python2.7/dist-packages/setuptools/dist.py", line 447, in __init__
k: v for k, v in attrs.items()
File "/usr/lib/python2.7/distutils/dist.py", line 287, in __init__
self.finalize_options()
File "/usr/local/lib/python2.7/dist-packages/setuptools/dist.py", line 739, in finalize_options
ep.load()(self, ep.name, value)
File "/girder/.eggs/setuptools_scm-3.2.0-py2.7.egg/setuptools_scm/integration.py", line 23, in version_keyword
dist.metadata.version = get_version(**value)
File "/girder/.eggs/setuptools_scm-3.2.0-py2.7.egg/setuptools_scm/__init__.py", line 144, in get_version
parsed_version = _do_parse(config)
File "/girder/.eggs/setuptools_scm-3.2.0-py2.7.egg/setuptools_scm/__init__.py", line 111, in _do_parse
"use git+https://github.com/user/proj.git#egg=proj" % config.absolute_root
LookupError: setuptools-scm was unable to detect version for '/girder'.
Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.
For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /girder/
Removing intermediate container f723032903e0
The command '/bin/sh -c pip install --upgrade --upgrade-strategy eager --editable .' returned a non-zero code: 1
```

(This may not be the right place for this, does it make sense to fix on this topic or should it be done separately?)

jbeezley:
I'll make a different PR for the docker build.  We may need to discuss how to solve it in more depth.

PR #2958, opened 2019-03-10
Separate DICOM search by key and value with equal sign

finetjul:
xxx will search xxx in key or value
xxx=yyy will search xxx in key and yyy in key value
xxx\=yyy will search xxx=yyy in key or value


finetjul:
@jbeezley @zachmullen can you please review ?

mgrauer:
Also @brianhelba as we have a customer relying on DICOM search.

zachmullen:
+1 to have @brianhelba review this since he is the original designer of the DICOM search features.

brianhelba:
Will do.

finetjul:
@brianhelba I don't think my PR is making the tests fails. Feel free to have a look. Thanks.

PR #2963, opened 2019-03-19
Mention changing allowed host when running in docker

subdavis:
`-p 8080:8080` isn't a useful flag without `--host 0.0.0.0`

### DIFF
--- a/docs/deploy.rst
+++ b/docs/deploy.rst
@@ -117,7 +117,7 @@ information, see the
 `Docker Hub Page <https://registry.hub.docker.com/u/girder/girder/>`_. Since the
 container does not run a database, you'll need to run a command in the form: ::
 
-   $ docker run -p 8080:8080 girder/girder -d mongodb://db-server-external-ip:27017/girder
+   $ docker run -p 8080:8080 girder/girder -d mongodb://db-server-external-ip:27017/girder --host 0.0.0.0
 
 Google Container Engine
 -----------------------


zachmullen:
> but given this page is about deployment, it makes sense to add.

It wouldn't be a good idea to open up Girder directly in a production deployment though, would it?

jbeezley:
Isn't exposing it publicly the point of deployment?

zachmullen:
I guess, but we should include a disclaimer that it won't have TLS enabled.

jbeezley:
That seems like an orthogonal concern.  The nginx or apache examples on this page don't enable TLS either.

I assume anyone deploying for real will understand what the `--host` flag means, but maybe this can be moved to a note after the command example.  I don't particularly care either way.

zachmullen:
Yeah, this is more of a philosophical question about how much we should try and document production deployments. Like, should we show nginx configurations that include TLS only, etc.

In any case, we should be explicit about what exactly is being provided by the given commands. Like "this will make the server visible to your network, but be careful because no support for TLS is included in this example, you are responsible for that".

I'm OK with merging this, my position is that people trying to deploy to production need to know what they're doing anyway, including understanding docker networking if they're using docker in production.

subdavis:
I actually didn't have devops people in mind with this fix.

This is the only line of documentation I could find that exists about running girder with docker, which I found by searching `docker` in readthedocs (it's also the second result for google search `girder docker`), so anyone wanting to test out the image will likely end up here.

I think any reasonable user will understand `--host 0.0.0.0`, but only someone with familiarity with girder would be able to guess why `8080` isn't responding after running `docker run -p 8080:8080 girder/girder -d mongodb://db-server-external-ip:27017/girder`

This change probably wouldn't be necessary if https://hub.docker.com/r/girder/girder/ provided some helpful getting-started style docs since that's where a docker user will look for them.  I'd be happy to log an issue and write something tailored for that page.

zachmullen:
Maybe I'm misunderstanding... if you omit the `--host 0.0.0.0`, and go to `localhost:8080`, you don't see Girder? Or are you talking about going to `0.0.0.0:8080` and not seeing it? I assumed that `localhost:8080` would work just by passing `-p 8080:8080`.

subdavis:
@zachmullen correct.  Without `--host`, `localhost:8080` returns `Site can't be reached.`

zachmullen:
Thanks for clarifying. In that case, we should definitely merge this.

Issue #1692, opened 2016-12-01
Path-based routing


ronichoudhury:
Would love to get to an item/folder/collection by forming a URL like `/#collection/MyCollectionName/foo/bar/baz` even if that URL simply redirects to the canonical hash-based URL.

jbeezley:
`/resource/lookup` returns an id from a path string.  It wouldn't be too hard to perform the routing from that.  The catch is that you have to know a priori what kind of model it is so it knows what collection to query.

kotfic:
+1

PR #2964, opened 2019-03-21
Add a resource/path/download/+path endpoint.

manthey:
This adds the ability to have a path with a variable number of components at the end by making the last component +<name>.  Note that this shows up in swagger as a single parameters {name}, which doesn't
work properly in the swagger interface.

### DIFF
--- a/girder/api/docs.py
+++ b/girder/api/docs.py
@@ -36,7 +36,7 @@ def _toRoutePath(resource, route):
     """
     # Convert wildcard tokens from :foo form to {foo} form
     convRoute = [
-        '{%s}' % token[1:] if token[0] == ':' else token
+        '{%s}' % token[1:] if token[0] in (':', '+') else token
         for token in route
     ]
 
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -1011,6 +1011,20 @@ class Resource(object):
                     break
             else:
                 return route, handler, wildcards
+        # Handle routes with variable arguments
+        for routelen in range(len(path), 0, -1):
+            for route, handler in self._routes[method][routelen]:
+                wildcards = {}
+                if route[-1].startswith('+'):
+                    wildcards[route[-1][1:]] = path[routelen - 1:]
+                    for routeComponent, pathComponent in six.moves.zip(
+                            route[:-1], path[:routelen - 1]):
+                        if routeComponent[0] == ':':  # Wildcard token
+                            wildcards[routeComponent[1:]] = pathComponent
+                        elif routeComponent != pathComponent:  # Exact match token
+                            break
+                    else:
+                        return route, handler, wildcards
 
         raise RestException('No matching route for "%s %s"' % (method.upper(), '/'.join(path)))
 
--- a/girder/api/v1/resource.py
+++ b/girder/api/v1/resource.py
@@ -17,6 +17,7 @@
 #  limitations under the License.
 ###############################################################################
 
+import cherrypy
 import six
 
 from ..describe import Description, autoDescribeRoute
@@ -24,6 +25,7 @@ from ..rest import Resource as BaseResource, setResponseHeader, setContentDispos
 from girder.constants import AccessType, TokenScope
 from girder.exceptions import RestException
 from girder.api import access
+from girder.models.file import File
 from girder.utility import parseTimestamp
 from girder.utility.search import getSearchModeHandler
 from girder.utility import ziputil
@@ -49,6 +51,7 @@ class Resource(BaseResource):
         self.route('GET', (':id', 'path'), self.path)
         self.route('PUT', (':id', 'timestamp'), self.setTimestamp)
         self.route('GET', ('download',), self.download)
+        self.route('GET', ('path', 'download', '+path'), self.pathDownload)
         self.route('POST', ('download',), self.download)
         self.route('PUT', ('move',), self.moveResources)
         self.route('POST', ('copy',), self.copyResources)
@@ -204,6 +207,51 @@ class Resource(BaseResource):
                             doc=doc, user=user, includeMetadata=includeMetadata, subpath=True):
                         for data in zip.addFile(file, path):
                             yield data
+            yield zip.footer()
+        return stream
+
+    @access.public(scope=TokenScope.DATA_READ)
+    @autoDescribeRoute(
+        Description('Download a resource based on its path')
+        .param('path', 'The path of the resource.', paramType='path')
+        .errorResponse('Path is invalid.')
+        .errorResponse('Path refers to a resource that does not exist.')
+        .errorResponse('Read access was denied for the resource.', 403)
+    )
+    def pathDownload(self, path):
+        user = self.getCurrentUser()
+        path = '/'.join(path)
+        resource = path_util.lookUpPath(path, self.getCurrentUser())['document']
+        if resource['_modelType'] == 'file':
+            single_file = resource
+        else:
+            model = self._getResourceModel(resource['_modelType'], 'fileList')
+            single_file = None
+            for (path, file) in model.fileList(doc=resource, user=user, subpath=True, data=False):
+                if single_file is None:
+                    single_file = file
+                else:
+                    single_file = False
+                    break
+        if single_file is None:
+            return None
+        if single_file is not False:
+            offset, endByte = 0, None
+            rangeHeader = cherrypy.lib.httputil.get_ranges(
+                cherrypy.request.headers.get('Range'), single_file.get('size', 0))
+            if rangeHeader and len(rangeHeader):
+                offset, endByte = rangeHeader[0]
+            single_file = File().load(single_file['_id'], user=user, level=AccessType.READ)
+            return File().download(single_file, offset, endByte=endByte)
+        setResponseHeader('Content-Type', 'application/zip')
+        setContentDisposition(resource.get('name', 'Resources') + '.zip')
+
+        def stream():
+            zip = ziputil.ZipGenerator()
+            for (path, file) in model.fileList(
+                    doc=resource, user=user, subpath=True):
+                for data in zip.addFile(file, path):
+                    yield data
             yield zip.footer()
         return stream
 
--- a/girder/cli/mount.py
+++ b/girder/cli/mount.py
@@ -509,7 +509,7 @@ def mountServer(path, database=None, fuseOptions=None, quiet=False, plugins=None
 # will work.  If you specify a string that doesn't contain :// as the database
 # uri, then it will use the default girder configuration (e.g., use
 # "mount -t girder girder <path>").
-# If you have girder installed in a virtualenv, ifrequently prepending the
+# If you have girder installed in a virtualenv, frequently prepending the
 # virtualenv's bin directory to the path is enough to use it, so the
 # mount.girder file becomes
 #   #!/usr/bin/env bash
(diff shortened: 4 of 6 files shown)


girder/api/v1/resource.py
```diff
@@ -207,6 +210,51 @@ def stream():
...
+    )
+    def pathDownload(self, path):
+        user = self.getCurrentUser()
+        path = '/'.join(path)
+        resource = path_util.lookUpPath(path, self.getCurrentUser())['document']
+        if resource['_modelType'] == 'file':
+            single_file = resource
+        else:
+            model = self._getResourceModel(resource['_modelType'], 'fileList')
+            single_file = None
```
brianhelba: Girder uses `camelCase`, not `snake_case` for its Python code.

girder/api/v1/resource.py
```diff
@@ -49,6 +51,7 @@ def __init__(self):
         self.route('GET', (':id', 'path'), self.path)
         self.route('PUT', (':id', 'timestamp'), self.setTimestamp)
         self.route('GET', ('download',), self.download)
+        self.route('GET', ('path', 'download', '+path'), self.pathDownload)
```
brianhelba: This does essentially the same type of lookup as `resource/lookup`, yet the API is totally different. Could we do anything to unify this?

girder/api/v1/resource.py
```diff
@@ -207,6 +210,51 @@ def stream():
             yield zip.footer()
         return stream
 
+    @access.public(scope=TokenScope.DATA_READ)
+    @autoDescribeRoute(
+        Description('Download a resource based on its path')
+        .param('path', 'The path of the resource.', paramType='path')
```
brianhelba: There's no documentation here on the use of this parameter (for instance, that you must prepend `'/user/'` or `'/collection'`).

girder/api/rest.py
```diff
@@ -1011,6 +1011,20 @@ def _matchRoute(self, method, path):
                     break
             else:
                 return route, handler, wildcards
+        # Handle routes with variable arguments
```
brianhelba: This new `+` syntax appears to be a new invention for Girder (as opposed to something inherited from CherryPy). Should we add developer documentation? Should we add more explicit testing?

brianhelba: It also looks like there's some logic here (e.g. support for both `:` and `+` in a route) for cases which are not covered implicitly by the use of `resource/path/download/+path`. Do we have any validation that it works? The addition of this new code significantly increases the cyclomatic complexity of `_matchRoute`.

girder/api/docs.py
```diff
@@ -36,7 +36,7 @@ def _toRoutePath(resource, route):
     """
     # Convert wildcard tokens from :foo form to {foo} form
     convRoute = [
-        '{%s}' % token[1:] if token[0] == ':' else token
+        '{%s}' % token[1:] if token[0] in (':', '+') else token
```
brianhelba: Sets are better than tuples for testing membership. :nerd_face:

girder/api/v1/resource.py
```diff
@@ -207,6 +210,51 @@ def stream():
...
+            model = self._getResourceModel(resource['_modelType'], 'fileList')
+            single_file = None
+            for (path, file) in model.fileList(doc=resource, user=user, subpath=True, data=False):
+                if single_file is None:
+                    single_file = file
+                else:
+                    single_file = False
+                    break
+        if single_file is None:
+            return None
```
brianhelba: Sometimes this returns 200 with a `null` body?

girder/api/v1/resource.py
```diff
@@ -207,6 +210,51 @@ def stream():
...
+        else:
+            model = self._getResourceModel(resource['_modelType'], 'fileList')
+            single_file = None
+            for (path, file) in model.fileList(doc=resource, user=user, subpath=True, data=False):
+                if single_file is None:
+                    single_file = file
+                else:
+                    single_file = False
+                    break
+        if single_file is None:
```
brianhelba: So `single_file` could be either `None`, or `False`, or a `dict`, with significantly different behavior for each. This might be abusing dynamic typing a little much, but that's just my opinion.

zachmullen:
Since there are several review comments, I'm going to engage the branch prediction failure protocol. I'll revert and we can re-open for further review.

manthey:
For a third party utility, we want to supply a route to a file (ask @matthewma7 for the specific use case).  Prior to this, we have the behavior that if you download a one-file-item, you get the file, but if you download a multi-file-item you get a zip file.  This endpoint makes use of the that behavior, but extends it so if you ask to download any resource that contains one file you get that file.  Probably for consistency, downloading a zero-file resource should return a zip file (not None).

By exposing a path based download endpoint, some external applications can treat one-file-items as if they are files.

Fundamentally, the swagger (and openapi) spec doesn't permit variable length routes, which is why there is an innovation of `+path`.  This *cannot* by compliant with the swagger spec.  If there is something I've missed that could allow this to work in swagger, I would have much preferred that.

girder/api/v1/resource.py
matthewma7: @brianhelba Do you mean the difference of the API of the endpoint? The use case of mine is looking for downloading girder item/file with relative path. Passing the path as a param won't work. And, the ability to download girder resource with a relative path will benefit not just my case. Such as OBJ, mtl, image files, .shp, .shx, .dbf, and .prj files. That's the reason David implement this API.

girder/api/v1/resource.py
manthey: Why is the use of a language feature abuse?  The alternative is to use multiple variables, which, in my opinion, reduces readability rather than increases it.

girder/api/v1/resource.py
manthey: In retrospect, this should return a zip file with no files in it for consistency.

matthewma7:
Yes, the use case of mine requires a relative path for looking up sibling files for downloading. And I see this relative path ability is beneficial to Girder in general and people have asked for it. https://github.com/girder/girder/issues/1692

matthewma7:
I had a need for this feature when worked with OBJ file but I found a workaround. but this time, I am working with Cesium.js 3DTileset, it expects to take a single JSON file and will download files referenced in the JSON file by requesting at the same URL path of the JSON file. I have no control over this behavior and this PR made it possible.

kotfic:
If the goal is to support Cesium's requirements shouldn't this functionality be implemented in a plugin? Maybe i've missed something critical here, is there something about new endpoints/event handlers that is insufficiently powerful enough to support the functionality needed for this use-case?

zachmullen:
> it expects to take a single JSON file and will download files referenced in the JSON file by requesting at the same URL path of the JSON file

And presumably the files referenced in the JSON file are not a flat list, but contain `/` in the identifiers? Is it possible to use a query string for this part?

matthewma7:
@kotfic I and David think this is a good feature for Girder to have. For things like obj, mtl, image file set and .shp, .shx, .dbf, .prj file set. Did you like the feature in https://github.com/girder/girder/issues/1692?

matthewma7:
@zachmullen I don't fully see using a query string for this. The library will take the path of the JSON file and replace the JSON with the filenames defined in the JSON. Here is an example. The foo and bar are something I added arbitrarily. And the JSON and b3dm files are generated. 

image

zachmullen:
Are there any known cases where the path is nested rather than flat? If not, we don't need to add this variadic path argument support to core. We could put all of the files into a single item (or folder if preferred) and add an endpoint (maybe in a plugin at first) like:

`GET /item/{id}/file_download_by_name/tileset.json`

Would that work?

kotfic:
Sorry,  i'm doing my best to keep up here, maybe you can explain to me why this specific implementation to meet the needs of Cesium.js is also a good general implementation for other uses cases? 

As you said,  you've found workarounds for this in the past.  I'm sure others have had this problem also, In what cases will it make sense to go back and re-implement hacky old solutions with the capabilities added by this feature? Are there other existing features (e.g. `resource/lookup`) we need to re-consider/think about removing in light of this feature?

zachmullen:
@kotfic there are other tools that make assumptions about names mentioned in an index file being relative paths to that index file. We actually have the [same behavior](https://github.com/girder/girder_pacs/blob/master/girder_stroke/__init__.py#L250) in the DICOM thumbnail viewer that uses paraviewweb/interactive thumbnail functionality.

Other example would be, as Matthew mentioned, an OBJ file that references MTL files that should be loaded by model viewers.

matthewma7:
@zachmullen Some more complex case of 3Dtileset will be nested. For example this [one](https://cesiumjs.org/NewYork/?view=-74.01881302800248%2C40.69114333714821%2C753.2406554180401%2C21.27879878293835%2C-21.343905508724625%2C0.0716951918898415), it is `"content":{"url":"6/25/28.b3dm"}`
Just find out it's part of [the spec](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification#uris) of 3d tiles.

zachmullen:
If we need to support nested names, we could use the query string, assuming that the client is smart enough to support the path being in the query string.

`GET /item/{id}/file_download_by_name?name=/tileset.json`

Internally, that endpoint could do some logic to resolve the name to a urlencoded version of it within the item. If we don't want to keep the tileset all in a single item (maybe it isn't semantically all the same dataset), we could do actual nesting with folders. I'd be hesitant to do that for fear of making the performance intolerably slow (multiple folder lookups for each tile would be dicey).

I'm trying to avoid adding features to our REST API that cannot be expressed in OpenAPI. I haven't checked myself, but it seems like there were some claims earlier in the thread that this isn't supported in OpenAPI/swagger.

manthey:
As far as my reading of the OpenAPI spec, variable length paths are not supported.  You could register every possible path length with placeholders /foo/:part1, foo/:part1/:part2, foo/:part1/:part2/:part3, etc., with each length appearing as a different endpoint, but that seems too awful to do.

I think that this is a legitimate way to satisfy multiple use cases where OpenAPI is insufficient.

matthewma7:
@kotfic I had had more control when working with OBJ with VTK.js, but if relative path was supported, it would be easier.

zachmullen:
In that case it seems we would need variadic path args. [Here](https://github.com/OAI/OpenAPI-Specification/issues/892) is the relevant philosophical discussion that will probably need to be echoed within the Girder community to figure out whether we're OK with the concept.

Side note: even if we do support that, I doubt you'd want to use the download by path endpoint as written here for loading map tiles. Each tile fetch will incur O(m+n) database lookups where m is the depth of the index.json and n is the nesting level of the tile within the dataset, that seems like it would be a performance killer. If Girder was internally represented more like an object store where each object contained its full absolute path, that would be different.

matthewma7:
@zachmullen Yea, performance is a concern, but the tileset that could be used is not as complex as the New York city example, which contains a handful of items at most.

matthewma7:
Ultimately, from the interface point of view, URL and URI has a built-in concept of directory and path (and relative path). Girder has built-in support of directory. It feels natural to me to map URL directory to girder directory.

zachmullen:
I'm going to pull this discussion into an email thread.

matthewma7:
PS: Swagger doesn't support this, but the swagger functionally works fine with this endpoint.

girder/api/v1/resource.py
jcfr: Instead of variable length path, couldn't the implementation be similar to these existing end points:

image

But in the case of `/resource/path/download` it would accept an absolute path instead of an ID.  It would basically save a call to `/resource/lookup` and then `/resource/download`

PR #2967, opened 2019-03-26
Prevent some false positive alerts in logs

brianhelba:
This resolves some issues which can trigger false positive alerts with log analysis tools.

* ~~Move the logger to 'girder.app' and the audit logger to 'girder.audit'~~
* Don't create special log entries for AccessExceptions

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -558,7 +558,6 @@ def _handleAccessException(e):
         cherrypy.response.status = 401
     else:
         cherrypy.response.status = 403
-        logger.exception('403 Error')
     val = {'message': e.message, 'type': 'access'}
     if e.extra is not None:
         val['extra'] = e.extra


girder/api/rest.py
```diff
@@ -558,7 +558,7 @@ def _handleAccessException(e):
         cherrypy.response.status = 401
     else:
         cherrypy.response.status = 403
-        logger.exception('403 Error')
+        logger.debug('403 Error')
```
zachmullen: This represents a policy change that should be considered carefully by Girder stakeholders. I don't think just changing this to `debug` is the correct solution. A better solution might be to divert 403 errors into a separate logging channel.

I think we should separate out the `propagate = False` change from this one, since I think we want to merge that one quickly and it's probably not controversial.

zachmullen:
Can you elaborate on "false positives"? What is the analytics tool finding that isn't correct? Why does unifying the loggers under a root logger fix that?

girder/api/rest.py
brianhelba: https://docs.python.org/3/library/logging.html#logging.Logger.exception
> Logs a message with level ERROR on this logger. The arguments are interpreted as for debug(). Exception info is added to the logging message. This method should only be called from an exception handler.

Girder denying access to a resource is not an application-level ERROR, and should not be logged with that severity.

If you'd like a separate logging channel for 403 or similar security-adjacent events, that sounds like a great new feature, but is out of scope of this PR and the main `girder` logger.

manthey: We almost don't use log level `debug` in Girder (it looks like it only gets used in a few places in the filesystem assetstore).  Perhaps we want to log this at the `info` level.

zachmullen: > If you'd like a separate logging channel for 403 or similar security-adjacent events, that sounds like a great new feature, but is out of scope of this PR and the main girder logger.

Since 403 events aren't "supposed" to happen under normal operation of Girder, we log them because they may represent malicious attacks, and the inclusion of their stack trace is important in that regard. I agree that they don't belong in the same channel as 500-level exceptions since they're client errors and not server errors. However, I don't want to merge this part until we develop the alternative place to put them, stack trace included; I don't want to create a gap where these things went from being logged *somewhere* to not being logged at all (setting them as "debug" events is essentially throwing them away). I contend that `exception` is the appropriate level for the event, since an exception was thrown and we are handling it, and the exception information is desired in the log.

brianhelba:
I am using [Sentry](https://sentry.io/) to monitor log files from a production Girder server. Having a single application logger and not emitting events at inappropriate levels will make it much easier to monitor Girder.

zachmullen:
How does sentry handle the audit log case, in which the message alone is basically meaningless, and the real information is contained in structured extra arguments? (That was the rationale for making the audit log separate -- it's not intended to work like a normal text-oriented log.)

zachmullen:
Here's a candidate approach. Three loggers:

1. `'girder.app'`, which would take the place of the current 'girder' log (`girder.logger`).
1. `'girder.security'`, we could divert the 403 exception messages here, still under the `exception` level. Other security related stuff could go in here as needed. We could introduce a new python API symbol for this like we have already for the main log and the audit log.
1. `'girder.audit'`, assuming tools like Sentry intelligently handle the events. If not, I'd take the position that this should remain a separate root log.

We wouldn't have to set `propagate=False` on any of them, which is good because that smells fishy.

Thoughts?

brianhelba:
Based on a conversation with @zachmullen , I've reworked this as follows:
* We don't log 403 exceptions to the main log any more. When Girder is used as a pure REST API, it's possible for 403 errors to be returned without any nefarious behavior. For example, a user could be allowed to call `GET /folder/:id` for something they had `READ` access on, but then call `GET /folder/:id/access` and be denied.
  * A future feature could independently decide to log 403s in another log stream, as part of something like a security warning logger.
* We rename the `'girder'` logger to `'girder.app'`, and the `'girder_audit'` to `'girder.audit'`, which future-proofs us to continue to proliferate new loggers under an organized namespace, while also making it clear that the application logger is not privileged at the top of the hierarchy (it's just sometimes the most semantically interesting).

brianhelba:
I'm dropping the changes for changing the logger `girder` -> `girder.app`. This is causing the work done by `_attachFileLogHandlers` to break, and the current system is too complex for me to fix now without breaking it worse.

PR #2969, opened 2019-03-27
Add a resource/path/download/+path endpoint.

zachmullen:
This adds the ability to have a path with a variable number of
components at the end by making the last component +<name>.  Note that
this shows up in swagger as a single parameters {name}, which doesn't
work properly in the swagger interface.


zachmullen:
This is #2964 re-opened for further review. Let's leave this open until the end of this week for commentary.

manthey:
@zachmullen I think this can be closed.  A plugin with this functionality was created at https://github.com/girder/resource_path_tools.

PR #2971, opened 2019-04-03
When specifying an image in item tasks, return a single file

manthey:
For image types, the user picks an item, from which the first file is returned.  If the item contains metadata that includes a fileId in a subdictionary, that file is returned instead of the first one.

### DIFF
--- a/plugins/item_tasks/server/rest.py
+++ b/plugins/item_tasks/server/rest.py
@@ -7,7 +7,6 @@ from girder.api.describe import autoDescribeRoute, Description
 from girder.api.rest import ensureTokenScopes, filtermodel, Resource
 from girder.constants import AccessType, TokenScope, SortDir
 from girder.exceptions import ValidationException
-from girder.models.file import File
 from girder.models.item import Item
 from girder.models.token import Token
 from girder.plugins.jobs.models.job import Job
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -75,6 +75,7 @@ def run(self, *arg, **kw):
     'pymongo>=3.5',
     'PyYAML',
     'psutil',
+    'pyOpenSSL',
```
jbeezley: Why is pyOpenSSL needed here?

manthey: Circle is failing with `pkg_resources.DistributionNotFound: The 'pyOpenSSL>=0.14; extra == "security"' distribution was not found and is required by requests`.  We specify pyOpenSSL explicitly in master's setup.py, so I thought this was the most harmonious way to update the 2.x-maintenance branch.

plugins/item_tasks/server/rest.py
```diff
@@ -130,8 +131,25 @@ def _transformInputs(self, inputs, token):
...
+                    # if we asked for a file, we may have been given an item,
+                    # which case get the first file within it.  If the item's
+                    # metadata includes a dictionary that includes fileId, use
+                    # that file id.
+                    if rtype != 'file':
+                        raise
+                    item = Item().load(
+                        v['id'], level=AccessType.READ, user=self.getCurrentUser(), exc=True)
+                    fileIds = [v['fileId'] for k, v in six.iteritems(item)
+                               if isinstance(v, dict) if v.get('fileId')]
```
jbeezley: I assume this is to get the large image `fileId` if it exists.  Is there any other use case for this behavior or would referencing `item.get('largerImage')` be sufficient?

manthey: `item.get('largeImage')` would work, but it feels like that should be in the large_image plugin (which would then need to check if the item_tasks plugin exists).

plugins/item_tasks/server/rest.py
```diff
@@ -130,8 +131,25 @@ def _transformInputs(self, inputs, token):
...
+                    if len(fileIds) == 1:
+                        resource = File().load(fileIds[0], force=True)
+                    else:
+                        resource = list(Item().childFiles(
+                            item, limit=1, sort=[('_id', SortDir.ASCENDING)]))[0]
```
jbeezley: I think we should prevent the `IndexError` here if the item has no files.

setup.py
zachmullen: Does `requests[security]` bring this in? If so, would it be better just to specify that?

jbeezley: It would not surprise me if it didn't.  There are all sorts of edge cases related to pip resolving extra requirements that are requested through `install_requires`.  This is probably why it ended up in master.

jbeezley:
This should probably first go in https://github.com/girder/girder-item-tasks before merging here.

plugins/item_tasks/server/rest.py
jbeezley: I agree it would be best to have an extensibility hook here and add the code in large_image, but that would clearly require more work.  Failing that, I would rather just encode the large_image behavior directly here for simplicity.

setup.py
manthey: I'll move this to its own PR.

plugins/item_tasks/server/rest.py
manthey: If you've asked for an image and the item has no files, what should be the behavior?  Perhaps `ValidationException('Item %s has no files')` would be appropriate.

manthey:
As this has some CI, and https://github.com/girder/girder-item-tasks doesn't currently, I think I'd rather have it start here and then update there.

zachmullen:
What's the story here? This is not being targeted at an integration branch

manthey:
As far as I'm concerned, this could be merged in to the girder 2.x branch, but I was awaiting approval.  The 2.x maintenance branch could also use #2980, which started as part of this PR but moved to its own.

PR #2972, opened 2019-04-03
Update migration script for 2->3

mgrauer:
@jbeezley This is based on the work today migrating monkeybrains.

### DIFF
--- a/docs/migration-guide.rst
+++ b/docs/migration-guide.rst
@@ -56,8 +56,6 @@ plugin:
   named ``example_plugin``.
 * Move your ``web_client`` directory under the directory created in the previous step,
   which was ``example_plugin`` in the previous step.
-* Delete the ``plugin.json`` file at the root of your plugin. Move the ``dependencies`` from that file to
-  the ``package.json`` file created in the following step.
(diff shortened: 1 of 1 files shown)


docs/migration-guide.rst
```diff
@@ -54,18 +54,25 @@ plugin:
 * Move your plugin's python source code from the ``server`` directory to the package name defined
   in your ``setup.py``. In this example, you would move all python files from ``server`` to a new directory
   named ``example_plugin``.
-* Create a ``package.json`` file inside your ``web_client`` directory defining an npm package.  A minimal
-  example is as follows:
+* Move your ``web_client`` directory under the directory created in the previous step,
+  which was ``example_plugin`` in the previous step.
+* Delete the ``plugin.json`` file at the root of your plugin. Move the ``dependencies`` from that file to
```
manthey: Perhaps this should go after the creation of `package.json`, and the comment on the dependencies in that file should say to copy them from `plugin.json`.

mgrauer: An earlier draft I wrote did exactly that. I'll make that change.

mgrauer:
@manthey fixed, PTAL.

Issue #2974, opened 2019-04-04
PACKAGE argument to add_standard_plugin_tests is undocumented


brianhelba:
https://github.com/girder/girder/commit/1439ea2a670f2cff699e13a264f39ee496e2e3ac

PR #2975, opened 2019-04-05
Add changes to add_standard_plugin_tests to the migraiton guide

jbeezley:
I also removed the comment at the top about upgrading in place.  It is duplicated in the 1->2 section and is no longer relevant.


jbeezley:
Fixes #2974

PR #2976, opened 2019-04-05
Fix girder client to work with pre 3.0 girder server

jbeezley:
This was reported in [discord](https://discourse.girder.org/t/girder-client-3-0-0a6-and-girder-2-5-keyerror-release-on-upload/231).  I don't know if we ever concluded that we would absolutely support pre 3.0 Girder servers in the client, but the fix here is trivial.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -376,8 +376,14 @@ class GirderClient(object):
         :return: The API version as a list (e.g. ``['1', '0', '0']``)
         """
         if not self._serverVersion or not useCached:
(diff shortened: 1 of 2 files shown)


brianhelba:
@girder/developers An open question is the degree to which we want `girder_client` v3 to support Girder servers from v2. Considering we maintain the `2.x-maintenance` branch, I think we should not attempt to do so. However, I could see this PR / endpoint as a special case, since getting the version of a Girder server is critical to establish whether a user's installed `girder_client` is compatible with that server.

zachmullen:
I agree with this change, this is one function that should always work the same.

PR #2977, opened 2019-04-05
Remove the access.cookie decorator.

### LINKED ISSUE #2936: Remove access.cookie decorator
Instead, make `cookie` a kwarg to the other access decorators.

### LINKED ISSUE #2966: Bug: girderToken cookie takes precedence over Girder-Token header when both are set
Exactly what the title says. This is pretty easy to reproduce with a couple `curl`s.
Steps to reproduce:

1.  Retrieve your girder api key. Set `GIRDER_API_KEY` equal to it in your shell. 
2. Find a file you want to download. Grab the download url. Set `DOWNLOAD_URL` equal to it.
3. Run the following script.
```
GIRDER_TOKEN1=$(curl -sSL -X POST "https://data.kitware.com/api/v1/api_key/token?key=${GIRDER_API_KEY}&duration=1" -H "Content-Length: 0" | jq .authToken.token)

GIRDER_TOKEN2=$(curl -sSL -X POST "https://data.kitware.com/api/v1/api_key/token?key=${GIRDER_API_KEY}&duration=1" -H "Content-Length: 0" | jq .authToken.token)

curl -vsSL "${DOWNLOAD_URL}" \
        -H "Cookie: girderToken=${GIRDER_TOKEN1}" \
        -H "Girder-Token: ${GIRDER_TOKEN2}" \
        -o "outfile"
```
Which will use `GIRDER_TOKEN1`. This causes difficult to diagnose behavior if `GIRDER_TOKEN1` has expired.
4. Cry silently

manthey:
There can be conflicts on how endpoints are authenticated when combining the cookie access decorator with other access decorators.  By incorporating this as an option to the other decorators, the problem is
avoided.

This fixes #2966 and adds a test for that case.

This resolves #2936.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Changes
 * Move minimum node version to 8.x due to upstream packages using newer ES features.
   (`#2707 <https://github.com/girder/girder/pull/2707>`_).
 * Require MongoDB 3.4+ (`#2900 <https://github.com/girder/girder/pull/2900>`_)
+* Removed the `@access.cookie` decorator; this is now an option on the other access decorators.
 
 Bug Fixes
 ---------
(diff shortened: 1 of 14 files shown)


zachmullen:
Thanks @manthey !

Issue #2973, opened 2019-04-04
user listing API endpoint does not require user login


florianlink:
The current girder implementation has a public endpoint to list all users (and their login), see for example:

https://challenge.kitware.com/girder/api/v1/user

I think this is a big security problem, since it opens doors for password brute-forcing.
The user list should be only available to users which have a valid login.
Or it should at least be configurable in the settings if the user listing is available to the public.

jbeezley:
#2252 is related.  We ran into complications when attempting to make the users endpoint admin only.  It might be more tractable to require just a normal login for that endpoint.  There are still edge cases (like should the search box return users) that would require consideration.

zachmullen:
I'd be OK with making this endpoint `@access.user` if that's a straightforward change. To be clear, I don't think doing so would mitigate password brute-forcing in any meaningful way, but I can definitely see the sensibility of not making it public by default.

jbeezley:
Besides the endpoint decorator, we may need to tweak the web client handling for the `Users` section.  Unless there are other complications, I think this would be a worthwhile change to make before the v3 release.

zachmullen:
Fixes in #2978

PR #2978, opened 2019-04-05
Only expose user listing endpoint to logged-in users

zachmullen:
Refs #2973


zachmullen:
This is ready for review. It doesn't change the resource search endpoint, so users can still be searched without being logged in. If we want to disable that too, I'd prefer to do it on a follow-up PR.

Issue #2559, opened 2018-01-08
Girder fails to start when pyopenssl is installed (python 3)


jbeezley:
This isn't actually a Girder bug, but I thought I would make an issue in case someone else encounters it.  I believe this is specific to Python 3 only.  With the following packages installed:
```
asn1crypto==0.24.0
bcrypt==3.1.4
boto3==1.5.9
botocore==1.8.23
certifi==2017.11.5
cffi==1.11.2
chardet==3.0.4
cheroot==6.0.0
CherryPy==11.0.0
click==6.7
cryptography==2.1.4
docutils==0.14
filelock==3.0.0
girder==2.4.0
idna==2.6
jmespath==0.9.3
jsonschema==2.6.0
Mako==1.0.7
MarkupSafe==1.0
more-itertools==4.0.1
portend==2.2
psutil==5.4.3
pycparser==2.18
pymongo==3.6.0
pyOpenSSL==17.5.0
python-dateutil==2.6.1
pytz==2017.3
PyYAML==3.12
requests==2.18.4
s3transfer==0.1.12
six==1.11.0
tempora==1.10
urllib3==1.22
```
Starting Girder results in:
```
Running in mode: development
Connecting to MongoDB: mongodb://localhost:27017/girder2
Resolving plugin dependencies...
[08/Jan/2018:09:48:51] ENGINE Bus STARTING
Started asynchronous event manager thread.
[08/Jan/2018:09:48:51] ENGINE Started monitor thread '_TimeoutMonitor'.
[08/Jan/2018:09:48:51] ENGINE Started monitor thread 'Autoreloader'.
[08/Jan/2018:09:48:52] ENGINE Serving on http://0.0.0.0:8080
[08/Jan/2018:09:48:52] ENGINE Bus STARTED
[08/Jan/2018:09:48:52] ENGINE Error in background task thread function <bound method Autoreloader.run of <cherrypy.process.plugins.Autoreloader object at 0x7fc0c9962b70>>.
AttributeError: cffi library '_constant_time' has no function, constant or global variable named '__loader__'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jbeezley/.virtualenvs/girder-bare/lib/python3.5/site-packages/cherrypy/process/plugins.py", line 519, in run
    self.function(*self.args, **self.kwargs)
  File "/home/jbeezley/.virtualenvs/girder-bare/lib/python3.5/site-packages/cherrypy/process/plugins.py", line 651, in run
[...]

PR #2980, opened 2019-04-09
Add pyOpenSSL to the requirements.

manthey:
We've already done this in master.  Without it, the requests library doesn't work.

Also, pin Sphinx for documentation builds.

Without this change, Circle is failing with `pkg_resources.DistributionNotFound: The 'pyOpenSSL>=0.14; extra == "security"' distribution was not found and is required by requests`.

### DIFF
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -1,6 +1,6 @@
 # See tox.ini for documentation on this file.
 funcsigs ; python_version < '3'
 six
-Sphinx
+Sphinx<2
 sphinx_rtd_theme
 sphinx-tabs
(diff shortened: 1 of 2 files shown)


danlamanna:
Note to self to check if this influences or makes https://github.com/girder/girder/issues/2559 irrelevant.

docs/requirements-docs.txt
```diff
@@ -1,6 +1,6 @@
 # See tox.ini for documentation on this file.
 funcsigs ; python_version < '3'
 six
-Sphinx
+Sphinx<2
```
jbeezley: Was Sphinx 2.0 causing an issue?  I don't see this in the master branch.

manthey: Yes, but only on the 2.x branch.

PR #2981, opened 2019-04-10
Redirect early in OAuth callback if any event prevents the default

Xarthisius:
As per [Discourse discussion](https://discourse.girder.org/t/rfc-reusing-oauth-login-to-integrate-with-3rd-party-apps/236) this PR allows to break OAuth callback flow early if any of the `oauth.auth_callback.{after,before}` events prevents default behavior.


zachmullen:
LGTM, thanks @Xarthisius

Issue #2434, opened 2017-10-17
Track which plugins were explicitly enabled


manthey:
Currently, when a user turns on a plugin that requires other plugins, all of those plugins are turned on as well.  If the user then turns off the first plugin, none of the others are turned off.  It would be better if plugins that were only on because of requirements were so marked so that they could be disabled when the requiring plugin is disabled.

plugins that are on because of a dependency should disable the on-off slider or change it (always on)-(on when required).  As part of this, such plugins could list what plugin(s) required them to be on as a tool tip or in some other way.  The UI experience of not being able to turn off something and having no explanation of why is poor.

danlamanna:
@brianhelba Can this be closed now?

brianhelba:
Issue is now obsolete.

Issue #2316, opened 2017-08-31
Guard against plugin webpack errors


manthey:
If a plugin's `webpack.helper.js` script throws an error, none of the plugins will build.  At the very least, there should be a guard around where `webpackHelper` function is executed.  The offending plugin should probably be marked failing in some manner so that it can't be enabled.

The guard is straightforward:
```
try {
    var newConfig = webpackHelper(baseConfig, helperConfig);
} catch (err) {
    grunt.log.error(`Failed to run webpack helper for ${plugin}: ${err.message}`);
    return;
}
```

I'm not sure how to disable the plugin in this case.

jbeezley:
The bigger problem is that you also need to disable anything that depends on the broken plugin.  The dependency resolution happens in python, so there isn't a clear path to figure that out.  Maybe just logging the error as done in your code block as a hint to the developer is enough?

PR #2983, opened 2019-04-12
Remove the ability to selectively enable plugins

### LINKED ISSUE #1438: Ability to enable plugins without restarting the server
This will be fairly easy to achieve; we can still require restart if a plugin is turned off (since we don't have an unloading process).

brianhelba:
Currently, the option to enable / disable plugins serves as a barrier to automated deployment of Girder. Instead of installing a Girder with plugins by simply:
* pip install girder
* pip install plugin
* start server

it instead currently requires:
* pip install girder
* pip install plugin
* start server
* create an admin user via HTTP API
* use admin credentials to enable plugin via HTTP API
* restart server

This PR changes the default behavior to always enable all installed plugins during Girder's normal operation.

---

Commits:
* Remove the ability to selectively enable plugins via the API
  * In normal operation, all installed plugins are now enabled. Within the test infrastructure, plugins may still be selectively enabled.
* Remove the "enabled" option from "pytest.mark.plugin"
* Remove the restart server button from the plugins page

Closes #1438 , #2434


girder/api/v1/system.py
```diff
@@ -194,37 +191,6 @@ def getVersion(self):
...
-        setting = Setting()
-        routeTable = setting.get(SettingKey.ROUTE_TABLE)
-        oldPlugins = setting.get(SettingKey.PLUGINS_ENABLED)
-        reservedRoutes = {GIRDER_ROUTE_ID, GIRDER_STATIC_ROUTE_ID}
-
-        routeTableChanged = False
-        removedRoutes = (
-            set(oldPlugins) - set(plugins) - reservedRoutes)
-
-        for route in removedRoutes:
```
brianhelba: This logic is lost with this PR, but it only covered the case where a webroot was named exactly the same as the plugin, which is not enforced and doesn't cover plugins with multiple webroots.

mgrauer:
I'm not endorsing this approach, but if it goes it, it should also update these [docs](https://github.com/girder/girder/pull/2979).

Does this approach include installing/enabling dependent plugins through pip?

manthey:
I'm for this (but haven't reviewed the changes yet).

Having installed plugins be the enable plugins matches other python libraries (like flake8, for instance).  In the principle of least surprise, I think it is a distinct improvement.  It would be even better if we could determine when `girder build` needs to be done, so that adding or removing a plugin would involve pip and `girder serve` and no other command.

And, to @mgrauer's question, assuming plugins specify their dependencies properly, then pip already should install dependent plugins.

danlamanna:
I gave this a +1.

I don't think there's significant benefit to runtime configuration of plugins, and I've personally experienced the difficulty in deployment as a result. To speak for @kotfic, he originally laid out a pretty good proposal in favor of this in the girder-developers thread 'Girder 3 Proposal: Plugin Page'.

zachmullen:
I think we need a lot of eyes on this, it's a big enough change that I'm a little bit concerned about slipping it in just a few days before we are planning to cut the release, because we may not understand all the implications for a while...

@girder/core @girder/developers

jbeezley:
I'm in favor of this change.  It will make a big improvement to deployment scenarios.

The biggest potential downside of this is that it makes it difficult to publish a python package that is primarily a python library, but also contains a girder plugin.  The prime example of this girder worker.  There will be no way to use girder-worker as a library inside a girder environment without also enabling the girder worker plugin.

For girder worker, I don't forsee any major problems because the girder plugin part is pretty small and minimally invasive.  I can imagine downstream plugins  (but do not know of any) that do similar things for which this would be a bigger problem.  E.g. if we hadn't refactored large image.

based64god:
This does have an impact on downstream vision projects, but I think it's ultimately a good one (as it requires us to audit plugins that aren't really being maintained anymore). However, I can't imagine deployment cases where you're not at least using venv to stand up girder, so +1 from me.

kotfic:
I also support this PR in principal,  though wish it would go further and completely remove the feature to enable/disable plugins through the interface. My reasoning goes like this:

1. Girder plugins don't currently have a systematic way of guaranteeing compatibility with one another.
2. What combinations conflict among the power set of possible plugins is largely unknown, even to core Girder developers.
3. What combinations of plugins are intended to be used together and do not conflict is largely tribal knowledge within D&A (i.e. is not systematically documented anywhere).
4. Given an arbitrary set of plugins, determining compatibility requires a review of the various plugins' source code.
5. Girder's plugin page exposes the ability to enable and disable plugins through a web interface.
6. This means girder administrators without a "source-code level" of understanding are empowered to enable and disable plugins whose interactions are likely unknown (even among core contributors). 

I would like to see us convert the plugin page to display only pip installed plugins and remove front end and back end code that implement the enable/disable functionality of these plugins.  Short of that,  I think this is a positive step in the right direction.

jbeezley:
@kotfic If I understand correctly, this PR already does what you are suggesting.  It only leaves the concept of enabled/disabled plugins for the purposes of testing.  There is no longer a way to disable plugins when starting the server normally.

jcfr:
Does this mean that all plugins shipped with girder will always be enabled ? 

https://github.com/girder/girder/tree/master/plugins

Or does it mean that "core" plugins having their source conveniently  hosted along side girder will have to be explicitly "pip installed" to be enabled ?

jbeezley:
The base package contains no plugins.  Any plugins (including those in this repo) have to be explicitly installed.

brianhelba:
I edited @kotfic 's comment to indicate the context that it was from an earlier email thread about a related feature. It's not a direct comment on this PR.

jcfr:
> Any plugins (including those in this repo) have to be explicitly installed.

The change makes sense. It will simplify the deployment of girder instance using the "infrastructure as code" approach.

For the sake of [dogfooding](https://en.wikipedia.org/wiki/Eating_your_own_dog_food), where is the source repo hosting the code used to deploy data.kitware.com ? will that change have any implication for this instance ?

zachmullen:
@jcfr our deployment IaC repositories are private, I can point you to them offline.

jbeezley:
It may actually be best to stop handling errors at plugin load time with this change.  The prior justification for it was to avoid killing the server via an endpoint.

matthewma7:
A potential benefit of this remove runtime plugin support with this PR. 

I don't know if this is a legitimate need or recommended pattern, I just happen to run into a need to monkey patch some core girder functionality (overwrite Item model initialize() method to add more indexes) in a plugin, but currently, plugin initialization happens after core girder initialization, making it impossible to monkey patch. I am wondering, after this PR, if we could change it to initialize the plugins first then initialize core girder, opening up more possibility to modify core girder within plugins.

zachmullen:
> overwrite Item model initialize() method to add more indexes

I don't think we'll change the load ordering for 3.0, but you shouldn't need any monkey patching to add database indexes. You can just call `SomeModel().ensureIndex(...)` in your plugin `load`. [Here](https://github.com/girder/girder_pacs/blob/master/girder_stroke/__init__.py#L253) is an example.

danlamanna:
This would close https://github.com/girder/girder/issues/1438.

brianhelba:
> It may actually be best to stop handling errors at plugin load time with this change. The prior justification for it was to avoid killing the server via an endpoint.

@jbeezley Sounds good, but let's put that into a follow-up PR. This change is separable.

jbeezley:
@brianhelba agreed.

danlamanna:
This may fix/obsolesce https://github.com/girder/girder/issues/2434 and https://github.com/girder/girder/issues/2316.

brianhelba:
> Shouldn't we keep this around for those using v2?

@cjh1 It's going to be impossible for the same Ansible code to serve both v2 and v3, particularly in the area of plugin management. We will continue to maintain a `2.x-maintaince` branch where the old Ansible code continues to live, and can be used by downstreams still on v2 Girder.

brianhelba:
> If we go through with this, I feel like we should also remove [this code](https://github.com/girder/girder/blob/master/setup.py#L70-L90).

@zachmullen I'd like to do that in a follow-up PR, since it's possible that some may object to the loss of convenience, and I don't want to delay this PR any more.

brianhelba:
@girder/developers I think all feedback has now been addressed and the consensus seems to be strongly in favor of this change. Can someone perform a code review so we can merge this?

brianhelba:
As a preview, I have 3 follow-up PRs awaiting this change:
* Remove the ability to restart Girder server via the HTTP API
* Remove plugin load exception handling
* Remove the "plugins" pip extra

All depend on this PR, but are not strictly necessary in themselves, so I want to review them separately to facilitate feedback.

zachmullen:
This is a significant change, and is not without risk, as I mentioned early on in the discussion. I approve of the idea, but I want to share the responsibility of this change across as much of the core Girder team as possible. I will mark a final approval once I'm satisfied that we are committed to sharing any pain that might arise from this change; to that end, I would love to get as many formal approvals on this PR from the core team as possible before we merge and release it into the wild shortly thereafter.

@jbeezley @danlamanna @mgrauer @manthey @kotfic

wphicks:
Overall, I think this is a good change. It does add some additional challenges for the specific use case of wanting to (easily) temporarily toggle plugins off and on to disable certain functionality. That's something we do quite a bit on my main project. That being said, I think the benefits probably outweigh the loss here.

brianhelba:
I now see 5 approvals, which is great. Has anyone done a code review?

zachmullen:
I have.

mgrauer:
after the fact, approved.

PR #2984, opened 2019-04-12
Rework static root setting

### LINKED ISSUE #2982: Girder doesn't work behind a subpath reverse proxy
The girder web client and the API page was using relative paths like these
```html
<script src="static/built/girder_lib.min.js"></script>
<script src="../static/built/swagger/swagger-ui.min.js"></script>
```
It worked when being put under a subpath by a reverse proxy.
such as http://localhost/subpath/

However, the latest master, uses absolute path
```html
<script src="/static/built/girder_lib.min.js"></script>
<script src="/static/built/swagger/swagger-ui.min.js"></script>
```
It doesn't work behind reverse proxy. becuase the the resource doesn't exists at `http://localhost/static/built/girder_lib.min.js` 
It actually exist at 
`http://localhost/subpath/static/built/girder_lib.min.js`
(Font and the API page still won't work even if I remove the leading slash by monkey patching)

Setting core.route_table like these doesn't help 
```json
[{"key":"core.route_table","value":{"core_girder":"/","core_static_root":"/subpath/static"}}]
```
Becuase even ouptut changed to
```html
<script src="/subpath/static/built/girder_lib.min.js"></script>
```
the resource now actually exists at
`http://localhost/subpath/subpath/static/built/girder_lib.min.js`

I didn't pinpoint which commits changed the behavior, it might be caused by a few commits.
But girder at this hash works. 7327d0039d760114a09aafd6d102b89561144f63
The reverse proxy I was using is a node application, the test script is [here](https://gist.github.com/matthewma7/052c994f09fa11768af07f5424aad07e).

brianhelba:
Girder now serves all static content from `/static`.

The `GIRDER_STATIC_ROUTE_ID` / `core_static_root` settings have been removed.

A new setting, `STATIC_PUBLIC_PATH` / `core.static_public_path`, now controls the prefix from where static content is expected to be fetched by HTTP clients. A reverse proxy or CDN must be used if this setting is used.

Fixes #2982

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,6 @@ COPY . /girder/
 # TODO: Do we want to create editable installs of plugins as well?  We
 # will need a plugin only requirements file for this.
 RUN pip install --upgrade --upgrade-strategy eager --editable .
-RUN girder build --static-root /static
+RUN girder build --static-public-path /static
 
 ENTRYPOINT ["girder", "serve"]
--- a/Dockerfile.py3
+++ b/Dockerfile.py3
@@ -27,6 +27,6 @@ ENV LANG=C.UTF-8
 # TODO: Do we want to create editable installs of plugins as well?  We
 # will need a plugin only requirements file for this.
 RUN pip install --upgrade --upgrade-strategy eager --editable .
-RUN girder build --static-root /static
+RUN girder build --static-public-path /static
 
 ENTRYPOINT ["girder", "serve"]
--- a/devops/ansible/roles/girder/library/test/test_setting.yml
+++ b/devops/ansible/roles/girder/library/test/test_setting.yml
@@ -60,7 +60,6 @@
           key: "core.route_table"
           value:
             core_girder: "/girder"
-            core_static_root: "/static"
 
     - name: Set multiple settings for email
       girder:
--- a/docs/deploy.rst
+++ b/docs/deploy.rst
@@ -28,7 +28,7 @@ example, we have the following:
 
     [server]
     api_root = "/girder/api/v1"
-    static_root = "/girder/static"
+    static_public_path = "/girder/static"
 
 .. note:: If your chosen proxy server does not add the appropriate
    ``X-Forwarded-Host`` header (containing the host used in http requests,
--- a/docs/deploy.rst
+++ b/docs/deploy.rst
@@ -28,7 +28,6 @@ example, we have the following:
 
     [server]
     api_root = "/girder/api/v1"
-    static_public_path = "/girder/static"
 
 .. note:: If your chosen proxy server does not add the appropriate
    ``X-Forwarded-Host`` header (containing the host used in http requests,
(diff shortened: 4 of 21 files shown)


manthey:
I haven't looked at these changes in any depth, but is there a reason we can't use relative paths to the static content?  That is, could the static path be `static` instead of `/static` and let the browser prefix the reverse proxy path automatically instead of using a setting?

matthewma7:
Thank you @brianhelba for this PR. The initial test worked well for me. 
For the case I described in #2982 
If I `girder build --static-public-path /subpath/static` then set the new setting static content root to '/subpath/static' it would work. for my localhost/subpath

But I think what @manthey said made sense. Is it possible to support both relative path and absolute path? like [vue-cli does](https://cli.vuejs.org/config/#publicpath)?
(I actually tried to forcefully set it to just `static` in the database, and `girder build --static-public-path static`, but I think only the first one worked)

brianhelba:
Webpack doesn't always work properly with relative paths for `publicPath`. If we could reliably get it working, we could loosen the restriction on the setting. We still need a setting, to support CDNs and more exotic mounting and build usages.

I think that should be explored in another PR, however.

matthewma7:
I almost got it working with `girder build --static-public-path '../'` except I noticed that the girder image logo doesn't work well. I wonder how it was set up before. 
Another thing I noticed I don't know if is important, due to the [migration](https://github.com/girder/girder/pull/2984/files#diff-f94a7a2e0c5304caf743b4878a5f70fdR145) here once the client database has run this version, reverting to an older version won't work, becuase `gurder serve` will throw error saying `KeyError: 'core_static_root'`

matthewma7:
It would be great if we can make relative path work again, which would make deployment easier.

brianhelba:
> It would be great if we can make relative path work again, which would make deployment easier.

As soon as this is merged, we can open a follow-up PR to try to relax the restriction, but I'll want to do a lot more testing to ensure it works in all cases.

Prior to my own several recent attempts to fix static root-related issues, there were a number of corner cases that were broken.

matthewma7:
This works for me since the current master doesn't work behind a reverse proxy. 
But I am not familiar with specific changes, so I am thinking I'll let others have a chance to give feedback. If nobody responds, I'll approve it. 
Thank you for addressing this so quickly.

brianhelba:
@matthewma7 Feel free to mark your approval as soon as you 'approve' that it fixes #2982 

I need to wait and get a code review from another stakeholder before merging this though.

brianhelba:
@zachmullen Nothing needs to be migrated for Girder 2->3. The name of a setting has changed, but that setting never fully worked before, so it's unlikely anyone was relying upon it.

brianhelba:
Also, I'd like to avoid creating clutter with low-level changes in our release docs. There are much important changes that must be highlighted.

zachmullen:
> Nothing needs to be migrated for Girder 2->3. The name of a setting has changed, but that setting never fully worked before, so it's unlikely anyone was relying upon it.

It appears that a command line option to `girder build` was changed, which should be documented.

There's also a change in deploy.rst that appears to refer to a new / changed config file variable, but it's unclear to me if or how that value is used by the system.

brianhelba:
Updated the docs. Indeed, there were a few out of date things there, thanks @zachmullen .

PR #2985, opened 2019-04-12
Fix ansible role for Girder v2

cjh1:
The current role on galaxy seems to be broken. This PR fixes up some of the issues.


cjh1:
@danlamanna Thanks for the review

PR #2988, opened 2019-04-16
Remove the ability to store girder configuration inside the package

danlamanna:
This will allow Girder to be upgraded via `pip install --upgrade` without concerns that it would overwrite custom configuration.

### DIFF
--- a/girder/utility/config.py
+++ b/girder/utility/config.py
@@ -21,6 +21,7 @@ import cherrypy
 import os
 import six
 
+import girder
 from girder.constants import PACKAGE_DIR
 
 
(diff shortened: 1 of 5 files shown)


girder/utility/config.py
```diff
@@ -41,14 +42,20 @@ def _loadConfigsByPrecedent():
...
     """
-    configPaths = [os.path.join(PACKAGE_DIR, 'conf', 'girder.dist.cfg')]
-
-    if 'GIRDER_TEST_DB' not in os.environ:
-        # we don't want to load the local config file if we are running tests
-        configPaths.append(os.path.join(PACKAGE_DIR, 'conf', 'girder.local.cfg'))
-
-    configPaths.append(os.path.join('/etc', 'girder.cfg'))
-    configPaths.append(os.path.join(os.path.expanduser('~'), '.girder', 'girder.cfg'))
+    # TODO: Remove in Girder 4.0
```
zachmullen: Can you add the string "deprecated" to this comment, I tend to grep for that

girder/utility/config.py
```diff
@@ -41,14 +42,20 @@ def _loadConfigsByPrecedent():
...
+    def _printConfigurationWarning():
+        girder.logprint.warning('Detected girder.local.cfg, this location is no longer supported.')
+        girder.logprint.warning('For supported locations, see https://girder.readthedocs.io/en/stable/configuration.html#configuration')  # noqa
```
zachmullen: This should be just one call to the logger rather than two

danlamanna:
@zachmullen rebased, can you reapprove?

PR #2991, opened 2019-04-17
Remove the ability to restart Girder server via the HTTP API

### LINKED ISSUE #2349: Add system/shutdown endpoint and event
We should have a PUT `system/shutdown` endpoint, much like the `system/restart` endpoint.

Further, both of these should emit an event so that plugins can listen for them and release resources or do other clean up if appropriate.  We can hook into the rest event (`rest.put.system.restart.before`), but since any plugin could elect to cancel this event, it might not be appropriate to perform shutdown actions.  cherrypy might emit a usable event, but a grider-specific event would be more appropriate.

brianhelba:
This functionality is normally only necessary when changing configured webroot paths, but it doesn't work on some deployment environments, requires a significant code burden, and may not reliably work if CherryPy is in an abnormal state.

Closes #2349


brianhelba:
Ping @zachmullen . Need your re-review. I fixed one import causing test failures.

PR #2993, opened 2019-04-17
Remove plugin load exception handling and rework plugin tests

### LINKED ISSUE #2316: Guard against plugin webpack errors

brianhelba:
Closes #2316

### DIFF
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -24,7 +24,6 @@ import json
 import six
 import os
 import logging
-import traceback
 
 from girder.api import access
 from girder.constants import TokenScope, ACCESS_FLAGS, VERSION
@@ -167,18 +166,10 @@ class System(Resource):
                 'version': p.version
             }
 
-        plugins = {
+        return {
             'all': {name: _pluginNameToResponse(name) for name in plugin.allPlugins()},
             'loaded': plugin.loadedPlugins()
         }
-        failureInfo = {
-            plugin: ''.join(traceback.format_exception(*exc_info))
-            for plugin, exc_info in six.iteritems(plugin.getPluginFailureInfo())
-        }
-
-        if failureInfo:
-            plugins['failed'] = failureInfo
-        return plugins
 
     @access.public
     @autoDescribeRoute(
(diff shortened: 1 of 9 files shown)


manthey:
This will probably affect development more than anything else.  When we are writing a plugin and have the server in development mode so it restarts when the python files change, if you break the plugin, the girder server currently restarts without the plugin enabled.  Fixing your code then requires restarting the server either via the endpoint or by touching another non-plugin file (or stopping it and starting it again as done originally).

With this change, the server will stop and have to be restarted via the appropriate command.

I don't object to this, but it could change how we work on plugins.

danlamanna:
> When we are writing a plugin and have the server in development mode so it restarts when the python files change, if you break the plugin, the girder server currently restarts without the plugin enabled.

This is a good point. Does it make sense to have custom logic for development mode here? In production I can see not wanting Girder to start if a plugin fails to load as that presents a fatal error in the deployment scenario. In development however, syntax errors are commonplace and shouldn't involve more intervention than fixing the syntax, if possible.

I don't want Girder to be littered with `if mode == development` logic, but it sounds like the extremely common cycle of write some code, test, repeat could be impacted a bit.

zachmullen:
Syntax errors in a plugin are already fatal to Girder startup, because we only catch exceptions that happen during plugin load, not during `entrypoint.load` of the plugin. We should do all-or-nothing behavior there.

brianhelba:
@girder/developers PTAL. This now removes exception handling from both `_loadPlugins` and `_wrapPluginLoad`. This also significantly rewrites the plugin tests to (hopefully) cover more of the cases that the current plugin system actually supports.

PR #2999, opened 2019-04-23
Fix test for latest moto version

zachmullen:
This fixes the broken test that is currently on master.

This accommodates two changes in the behavior of moto that changed in v1.3.8. First, it enforces bucket name minimum length. Second, it uses domain-name addressing for buckets.

### DIFF
--- a/tests/cases/file_test.py
+++ b/tests/cases/file_test.py
@@ -878,11 +878,11 @@ class FileTestCase(base.TestCase):
     @moto.mock_s3
     def testS3Assetstore(self):
         botoParams = makeBotoConnectParams('access', 'secret')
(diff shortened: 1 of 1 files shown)


tests/cases/file_test.py
```diff
@@ -878,11 +878,11 @@ def testGridFsAssetstore(self):
     @moto.mock_s3
     def testS3Assetstore(self):
         botoParams = makeBotoConnectParams('access', 'secret')
-        mock_s3.createBucket(botoParams, 'b')
+        mock_s3.createBucket(botoParams, 'bname')
```
brianhelba: What's wrong with the old bucket name?

zachmullen: Too short. Min length is 3

PR #3000, opened 2019-04-23
Remove multipart upload

### LINKED ISSUE #2998: Remove deprecated multipart upload support

zachmullen:
Fixes #2998

### DIFF
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -230,6 +230,9 @@ class File(Resource):
         server and client agree on the number of bytes sent/received.
         """
         if 'chunk' in params:
+            # If we see the undocumented "chunk" query string parameter, then we abort trying to
+            # read the body, use the query string value as chunk, and pass it through to
+            # Upload().handleChunk. This case is used by the direct S3 upload process.
             chunk = params['chunk']
         else:
             chunk = RequestBodyStream(cherrypy.request.body)
--- a/girder/api/v1/file.py
+++ b/girder/api/v1/file.py
@@ -19,7 +19,6 @@
 
 import cherrypy
 import errno
-import os
 import six
 
 from ..describe import Description, autoDescribeRoute, describeRoute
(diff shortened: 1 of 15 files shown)


zachmullen:
Depends on #2999

zachmullen:
Rebased, ready for review.

jcfr:
Could the client report an exception if `self.getServerVersion() < ['3', '0']` ?

zachmullen:
@jcfr the client will work with bith new and old versions of the server, so no exception is necessary.

jcfr:
Since the server <= 2.2 only supports multipart uploads, should we still report an exception ?

zachmullen:
Our general rule is that your client major version should match your server major version. (At one point, this was documented officially, and might still be...)

In light of that, I don't think we need version checks across major versions.

jcfr:
I suggest that the documentation is updated to reflect that then. 

Reading https://girder.readthedocs.io/en/stable/python-client.html#versioning

> The design intent is that every version of the client should work with every version of the server, though this is not an explicit guarantee. Generally you should use the latest version of the Girder python client that is installable from pip.

Generally speaking, if a clear error message can save emails and frustration, I like to favor that. I am just anticipating the time where data from an older Girder instance will need to be migrated ...

zachmullen:
Agreed, we should update the docs to make it explicit that you may run into errors if your major version of server and client don't match.

Migrating data from older Girder should work fine, since this only affects uploads.

girder/api/v1/file.py
```diff
@@ -232,24 +228,11 @@ def readChunk(self, upload, offset, params):
...
         upload. The passed offset is a verification mechanism for ensuring the
         server and client agree on the number of bytes sent/received.
-
-        This method accepts both the legacy multipart content encoding, as
-        well as passing offset and uploadId as query parameters and passing
-        the chunk as the body, which is the recommended method.
-
-        .. deprecated :: 2.2.0
         """
         if 'chunk' in params:
```
brianhelba: When would this case still occur?

zachmullen: It happens in the direct to S3 upload behavior. Which is not an ideal design, but that's what it's for.

tests/base.py
```diff
@@ -538,12 +534,11 @@ def request(self, path='/', method='GET', params=None, user=None,
             body = self.getBody(response)
             try:
                 response.json = json.loads(body)
-            except ValueError:
-                raise AssertionError('Did not receive JSON response')
+            except Exception:
```
brianhelba: This should be able to stay as just `ValueError`, since `json.loads` raises a `ValueError` in Python 2 and a [`JSONDecodeError`](https://docs.python.org/3/library/json.html#json.JSONDecodeError) (which is a subclass of `ValueError`) in Python 3.

Is there some other type of exception that needs to be handled here?

girder/api/v1/file.py
brianhelba: Can you add a code comment to explain this?
* If we see the undocumented `chunk` query string parameter, then we abort trying to read the body, use the query string value as `chunk`, and pass it through to `Upload().handleChunk`
* This case is typically used by the direct S3 upload process..

PR #3001, opened 2019-04-25
Enable linting for PyTest tests

brianhelba:


brianhelba:
The remaining test failures here are all `[W605] invalid escape sequence '\/'` in [`test_path_utilities.py`](https://github.com/girder/girder/blob/e980778c1d5a6d88d9878ace6c50c31a94223104/test/test_path_utilities.py). @zachmullen these are all blamed to you; can you fix them?

zachmullen:
I'm [not the original author of those lines](https://github.com/girder/girder/commit/f1fde7344f5daa56ff7d7b8858c9b5fd322e75d7) of those lines but I'm happy to make the fix.

PR #3002, opened 2019-04-25
Remove gitignore for girder.local.cfg

brianhelba:
This should have been removed by #2988


brianhelba:
@girder/developers PTAL

Issue #2996, opened 2019-04-22
MongoDB sharding tests do not work with newer versions of MongoDB


brianhelba:
As shown by [this build result](https://circleci.com/gh/girder/girder/19335), upgrading MongoDB to version 3.6 causes the `server_upload.testGridFSShardingAssetstoreUpload` and `web_client_data_gridfsshard` tests to fail.

[Reverting to MongoDB 3.2](https://github.com/girder/girder/pull/2836/commits/c13770944c007737830311bd8e0683a8b03ec20f) allows the tests to pass.

It's likely that some of the MongoDB sharding changes made [in 3.4](https://docs.mongodb.com/manual/release-notes/3.4/#sharded-cluster) and/or [in 3.6](https://docs.mongodb.com/manual/release-notes/3.6/#sharded-clusters) need to be incorporated into these tests.

Since both tests run as machine-executor "integration tests", this MongoDB version restriction does not affect most of the main CI tests.

manthey:
This may be a mongo installation issue -- locally with Mongo 4.0.9 these tests pass.  Circle is timing out from allowing ssh login right now.  @brianhelba Did the tests pass locally for  you?

PR #3010, opened 2019-04-25
Test with MongoDB 3.6 in integration tests

brianhelba:
This should work now that #2996 is resolved.


brianhelba:
Indeed, this now works. Thanks @manthey for fixing the underlying issue.

Issue #3008, opened 2019-04-25
Expose user details endpoint to log in users


matthewma7:
At this moment, [the `user/details` endpoint is limited to only admin users.](https://github.com/girder/girder/blob/master/girder/api/v1/user.py#L212)
But I don't see why since anonymous users were able to list all user.
(only logged-in user can list users now. https://github.com/girder/girder/pull/2978)

Expected result: Open the user details user open to any logged-in user. 

PS: This girder/girder_web_components#119 would use this.

brianhelba:
The reason is that since `User` is an `AccessControlledModel`, it's non-trivial to count the total number of apparent users that a given non-admin requester is able to see.

However, since `user/details` was first added, we now have `findWithPermissions`. If you can use `User.findWithPermissions` to generate the counts, you should be able to allow non-admins to call the endpoint.

brianhelba:
To clarify, typically, all users are `public=True` and therefore are all visible. However, plugins make it possible to have a `public=False` private user, so the `user/details` endpoint shouldn't dig us further into the assumption that all users are always public.

matthewma7:
Thank you @brianhelba for explaining the background.
Ok, cool, I'll update the existing code to do so.

Issue #3009, opened 2019-04-25
Add a collection details endpoint


matthewma7:
We have `folder/{id}/details`, `collection/{id}/details`, `user/{id}/details`, `user/details`, the only thing missing is `collection/details`, which would return the total number of collections.
It would be very similar to https://github.com/girder/girder/blob/master/girder/api/v1/user.py#L217-L219

PS: This girder/girder_web_components#119 would use this.

PR #3012, opened 2019-04-25
Add and improve details endpoints

matthewma7:
This PR improve `user/details` endpoint and add a new `collection/details` endpoint.

Resolve https://github.com/girder/girder/issues/3008 https://github.com/girder/girder/issues/3009

### DIFF
--- a/tests/cases/collection_test.py
+++ b/tests/cases/collection_test.py
@@ -404,3 +404,20 @@ class CollectionTestCase(base.TestCase):
         # Changes should have been saved to the database
         coll = collModel.load(coll['_id'], force=True)
         self.assertEqual(acl, coll['access'])
+
+    def testCollectionsDetails(self):
+        """
+        Test that the collections count is correct.
+        """
+        Collection().createCollection('private collection', public=False)
+        resp = self.request(path='/collection/details', user=self.admin, method='GET')
+        self.assertStatusOk(resp)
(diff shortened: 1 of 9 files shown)


tests/cases/collection_test.py
```diff
@@ -404,3 +404,20 @@ def testMissingAclRefs(self):
         # Changes should have been saved to the database
         coll = collModel.load(coll['_id'], force=True)
         self.assertEqual(acl, coll['access'])
+
+    def testCollectionsDetails(self):
```
zachmullen: New test cases should be implemented in pytest rather than the old unittest suite. Could you move this case over there?

matthewma7: Ah, I didn't know that. I'll move it.

matthewma7: Hm, I think I didn't find much-existing pytests to learn from. I am not so familar to move this test with a new test framework. Is that ok to have it here this time? If not, I can ask someone to help me take a look offline.

zachmullen: I just pushed a commit to this branch that moves the test case to pytest. I wanted to parametrize over two fixtures, so had to hack that a bit using another fixture, due to limitations of pytest. There may be a more clever way to do it, but I couldn't figure out how to essentially pass a fixture as a parametrized argument.

matthewma7: Thank you! I'll merge this now.

PR #3013, opened 2019-04-26
Reduce npm warnings with girder-build

manthey:
Every time we run girder build, npm warns that various fields aren't set in girder/meta-build.  This sets those fields to get rid of the warnings.

### DIFF
--- a/girder/web_client/package.json.template
+++ b/girder/web_client/package.json.template
@@ -1,7 +1,7 @@
 {
     "name": "@girder/meta-build",
     "description": "Girder meta-build",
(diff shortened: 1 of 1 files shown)


girder/web_client/package.json.template
```diff
@@ -1,5 +1,8 @@
 {
     "name": "@girder/meta-build",
+    "description": "Girder meta-build",
+    "license": "SEE LICENSE IN OTHER MODULES",
```
brianhelba: All of Girder's own code is `"Apache-2.0"`, so this can be too.

PR #3015, opened 2019-04-28
Validate email addresses more comprehensively

### LINKED ISSUE #2282: Validate email addresses
Email addresses are used in 3 different ways:

* Every user account has an email address that's associated with it. These are validated using the [`email_regex` config option](https://github.com/girder/girder/blob/d99f46c11367f8253d4a9cc66c8f87a56f754423/girder/models/user.py#L98) before being saved. These are not transformed before being used for outgoing mails or included in a web page.

* The `EMAIL_FROM_ADDRESS` setting is set by site admins, and is [used to send outgoing emails](https://github.com/girder/girder/blob/d99f46c11367f8253d4a9cc66c8f87a56f754423/girder/utility/mail_utils.py#L112) This is [not validated (except to ensure it's not empty)](https://github.com/girder/girder/blob/d99f46c11367f8253d4a9cc66c8f87a56f754423/girder/models/setting.py#L277) before being saved. This is not transformed before being used for outgoing mails or included in web pages.

* The `CONTACT_EMAIL_ADDRESS` setting is set by site admins, and is used to display "Contact" hyperlinks, for users to contact the instance admins for support. This is not validated (except to ensure it's not empty) before being saved. This is not transformed before being used for outgoing mails or included in web pages.

---

Since all of these are fundamentally an email address, they should be validated in the same way, likely using [a robust regex](https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address). The `email_regex` config option should be removed, as whether an email address is valid should not vary from instance to instance. Allowing selective registration to only users from a certain email domain is a separate problem, and should probably be handled by a normal instance setting within a plugin, possibly [the `autojoin` plugin](https://github.com/girder/girder/tree/d99f46c11367f8253d4a9cc66c8f87a56f754423/plugins/autojoin).

Additionally, email addresses often need to be transformed or encoded before being used or rendered on a webpage. Because this transformation is typically not idempotent, it should be done just-in-time before rendering, rather than in the setting validator. Accordingly:
* Anything in a `mailto: ` link should be encoded per [RFC6068](https://tools.ietf.org/html/rfc6068)
* Other HTML-embedded links probably need to be encoded per [RFC5322](https://tools.ietf.org/html/rfc5322).
* Encoding of addresses into MIME headers should also be considered

brianhelba:
Fixes #2282

### DIFF
--- a/girder/utility/mail_utils.py
+++ b/girder/utility/mail_utils.py
@@ -18,6 +18,7 @@
 
 import cherrypy
 import os
+import re
 import six
 import smtplib
 
@@ -26,6 +27,25 @@ from mako.lookup import TemplateLookup
 from girder import events
 from girder import logger
 from girder.constants import SettingKey, PACKAGE_DIR
+
+
+def validateEmailAddress(address):
+    """
+    Determines whether a string is a valid email address.
+
+    This implements the grammar from 4.10.5.1.5 of the HTML Standard.
+
+    :param address: The string to test.
+    :type address: str
+    :rtype: bool
+    """
+    # https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
+    return re.match(
+        r'^[a-zA-Z0-9.!#$%&\'*+/=?^_`{|}~-]+'
+        r'@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?'
+        r'(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$',
+        address
+    ) is not None
 
 
 def getEmailUrlPrefix():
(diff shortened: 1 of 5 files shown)


girder/utility/mail_utils.py
```diff
@@ -28,6 +29,26 @@
...
+    :type address: str
+
+    :rtype: bool
+    """
+    # https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
+    emailRegex = re.compile(
+        r"^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+"
+        r"@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?"
+        r"(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$"
+    )
```
zachmullen: Should we make this a module-level constant so it doesn't need to be recompiled each time this is called?

brianhelba: I've just inlined everything into one function call. The method is called infrequently enough that there's no need to try and optimize it.

brianhelba:
@girder/developers PTAL.

PR #3018, opened 2019-04-29
Centralize core settings

### LINKED ISSUE #3006: Move setting enum definitions, defaults, and validators to the same place
Part of this will involve ordering the existing definitions, and fixing a few contradictions and omissions that have crept into the settings over time. This is related to #2994 

An open question is where should the new setting data reside:
* within `girder/models/setting.py`
  * this makes the API for using settings and the definitions all in the same place, making it easier to import and inspect
* within a new `girder/settings.py`
  * this separates the definitions of core settings from the model API (which is independently used by plugins) and keeps the model file in line with the other models

If desired, we could import the new location back into `girder/constants.py`, providing backwards compatibility for downstreams importing from the old location.

brianhelba:
* Move all SettingKey to settings.py
* Alphabetize SettingKey definitions
* Move all SettingDefault to settings.py
* Alphabetize SettingDefault definitions
* Move setting_utilities.default functions under SettingDefault as private staticmethods
* Move setting_utilities.validator functions under a SettingValidator class as private staticmethods
* Ensure ValidationExcepton is called with a second argument in validators
* Document setting constant changes
* Update import references to SettingDefault and SettingKey

Fixes #3006


brianhelba:
This is the first of several PRs. I've confined this PR to mostly move and reorder operations in order to simplify review.

brianhelba:
@girder/developers PTAL.

PR #3019, opened 2019-04-30
Remove license headers from individual source files

brianhelba:
The LICENSE file at the repository root is sufficient to cover the entire repository. The copyright headers were never consistently applied and are just a source of cruft.

Third party source files still retain their own LICENSE files or license headers.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -1,21 +1,4 @@
 # -*- coding: utf-8 -*-
-
-###############################################################################
-#  Copyright Kitware Inc.
-#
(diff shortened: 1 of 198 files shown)


brianhelba:
@girder/developers PTAL.

brianhelba:
For some additional precedent, see https://github.com/Kitware/CMake/commit/86578eccf2e82286248796bad1032cd0e3a5e1e2#diff-e70cd26a4ba6127283781c59f9d492d1 . Although that change leaves a shorter stub header in most of CMake's files, Girder's source files are always distributed within the context of a package which contains explicit license metadata.

zachmullen:
@brianhelba I'm jelly of these diffstats

PR #3022, opened 2019-04-30
Fixes #2773. Support empty uploadFromFile

### LINKED ISSUE #2773: Upload.uploadFromFile returns incorrectly when passed an empty file
When passed an empty (`size==0`), [`Upload.uploadFromFile`](https://github.com/girder/girder/blob/master/girder/models/upload.py#L55) does not create a `File`, and instead returns an `Upload` object.

The correct behavior is for a `File` object (with an empty file) to be created and returned.

This bug is due to the fact that [`Upload.handleChunk`](https://github.com/girder/girder/blob/master/girder/models/upload.py#L108) is never called.

zachmullen:


brianhelba:
Code looks good, I just need to test it locally.

PR #3023, opened 2019-05-02
Add missing SettingDefault values

brianhelba:


brianhelba:
@girder/developers PTAL.

PR #3024, opened 2019-05-02
Pre-3.0 deprecated symbol cleanout

zachmullen:
`git grep -i deprecate` was used to find lingering deprecations. This does not remove *all* deprecated things comprehensively, and in one case, actually un-deprecates a behavior, which may be controversial.

Most of these things won't be missed, but this should be looked over carefully.

### DIFF
--- a/docs/migration-guide.rst
+++ b/docs/migration-guide.rst
@@ -304,7 +304,9 @@ handler, you must call :py:func:`girder.events.unbind` on the existing mapping f
 Async keyword arguments and properties changed to async\_ PR #2817
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-In version 3.7 of python ``async`` is a `reserved keyword argument <https://www.python.org/dev/peps/pep-0492/#deprecation-plans/>`_. To mitigate any issues all instances of ``async`` in the codebase has changed to ``async_``. If the functions are called with ``async`` a deprecation warning will be given. This affects:
+In version 3.7 of python ``async`` is a `reserved keyword argument <https://www.python.org/dev/peps/pep-0492/#deprecation-plans/>`_.
+To mitigate any issues all instances of ``async`` in the codebase has changed to ``asynchronous``.
+If the functions are called with ``async`` a deprecation warning will be given. This affects:
 
  * The event framework ``girder/events.py``
  * The built-in job plugin ``plugins/jobs/girder_jobs/models/job.py``
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -825,6 +825,8 @@ class Resource(object):
                 handler = registeredHandler
                 del nLengthRoutes[i]
                 break
+        else:
+            raise GirderException('No such route: %s %s' % (method, '/'.join(route)))
 
         # Remove the api doc
         if resource is None:
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -805,11 +805,9 @@ class Resource(object):
                 'WARNING: No access level specified for route %s %s' % (
                     method, routePath))
 
-    def removeRoute(self, method, route, handler=None, resource=None):
+    def removeRoute(self, method, route, resource=None):
         """
         Remove a route from the handler and documentation.
-
-        .. deprecated :: 2.3.0
 
         :param method: The HTTP method, e.g. 'GET', 'POST', 'PUT'
         :type method: str
@@ -817,9 +815,6 @@ class Resource(object):
                       resource root. Elements of this list starting with ':'
                       are assumed to be wildcards.
         :type route: tuple[str]
-        :param handler: The method called for the route; this is necessary to
-                        remove the documentation.
-        :type handler: Function
         :param resource: the name of the resource at the root of this route.
         """
         self._ensureInit()
--- a/girder/events.py
+++ b/girder/events.py
@@ -44,7 +44,7 @@ class Event(object):
 
     # We might have a lot of events, so we use __slots__ to make them smaller
     __slots__ = (
-        'async_',
+        'asynchronous',
         'info',
         'name',
         'propagate',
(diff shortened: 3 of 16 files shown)


zachmullen:
@brianhelba is going to strip out the `deprecatedAsync` stuff on a separate PR.

girder/models/model_base.py
```diff
@@ -1416,10 +1416,7 @@ def load(self, id, level=AccessType.ADMIN, user=None, objectId=True,
         :raises ValidationException: If an invalid ObjectId is passed.
         :returns: The matching document, or None if no match exists.
         """
-        # Warn of str type deprecation for `fields` param
         if isinstance(fields, six.string_types):
-            logger.warning('String data type for fields param is deprecated, \
```
brianhelba: The underlying PyMongo's `projection` parameter only accepts a `list` or a `dict`. Personally, I think it encourages code sloppiness to have internal APIs that attempt to be overly permissive in what they'll tolerate, as it leads to needing defensive programming code and validations / coercions in internally-called methods. I'd rather internal APIs just have well-documented preconditions, so we can make stronger assumptions about variable types and parameter usage.

Note, these preconditions can still be abstract and yet rigorous. For example:
> parameter `x` is an `Iterable`

instead of 

> parameter `X` can be a `list` or a `tuple`; and if it's a `str`, we'll convert it to a `list` for you; and if it's a `dict`, we'll take its `.keys()` for you; and if it's some other iterable, we probably didn't consider that, so the behavior is undefined-ish...

Practically though, I think that tolerating `fields` as a `str` exists in a few different places, so if we're going to remove the coercion, it would be a bigger change than this one place.

girder/api/rest.py
```diff
@@ -805,21 +805,16 @@ def route(self, method, route, handler, nodoc=False, resource=None):
...
-
         :param method: The HTTP method, e.g. 'GET', 'POST', 'PUT'
         :type method: str
         :param route: The route, as a list of path params relative to the
                       resource root. Elements of this list starting with ':'
                       are assumed to be wildcards.
         :type route: tuple[str]
-        :param handler: The method called for the route; this is necessary to
-                        remove the documentation.
-        :type handler: Function
```
brianhelba: In the code below, what if `route` is not found within `nLengthRoutes`? Then `handler` will never be defined, yet the code afterwards still references it.

@danlamanna may hate this, but this a perfect case for a `for ... else` expression.

girder/web_client/src/rest.js
```diff
@@ -137,28 +137,9 @@ let restRequest = function (opts) {
         delete args.girderToken;
     }
 
-    let jqXHR = Backbone.$.ajax(args);
-    jqXHR.error = function () {
```
brianhelba: @zachmullen Just to confirm, you grep-ed for usages of `.fail` on `restRequest` promises? In my experience, this has been one of the most stubborn to remove.

girder/models/model_base.py
zachmullen: What's your conclusion? I don't have a preference either way.

girder/web_client/src/rest.js
zachmullen: I didn't, thanks for reminding me, I assumed whoever deprecated it had done so. I'll go through and fix them, unless you think this is too risky and we should just leave the fallback in.

zachmullen: I was confused, it's actually `.error` that needs to be removed. I could not find any instances of that remaining.

zachmullen:
This is ready for final review.

girder/events.py
```diff
@@ -70,7 +53,6 @@ class Event(object):
         'currentHandlerName'
     )
 
-    @_deprecatedAsync
     def __init__(self, name, info, async_=False):
```
brianhelba: We should just call this parameter `asynchronous`. `async_` is arbitrary and doesn't follow Girder's naming conventions.

Other upstream libraries have also used `asynchronous`:
* https://github.com/celery/celery/pull/4863
* https://github.com/celery/kombu/pull/839

girder/models/model_base.py
brianhelba: Can you un-un-deprecate this particular location, then let's address it more comprehensively and decide in another PR?

zachmullen:
@brianhelba this should be ready

plugins/jobs/girder_jobs/models/job.py
```diff
@@ -248,7 +233,7 @@ def createJob(self, title, type, args=(), kwargs=None, user=None, when=None,
             'log': [],
             'meta': {},
             'handler': handler,
-            'async_': async_,
+            'async_': asynchronous,
```
brianhelba: Is there a reason we can't name this model property `'asynchronous'` too, so we have consistency?

zachmullen: I was trying to avoid yet another check [like this](https://github.com/girder/girder/blob/master/plugins/jobs/girder_jobs/models/job.py#L327), but I guess it probably doesn't matter. I'll change it in the database.

brianhelba: I think we can omit the check for `'async_'` entirely, since that field only exists within 3.0 pre-release instances. The check for `'async'` is perhaps more reasonable, to accommodate still-pending jobs in a 2.x -> 3 upgrade, but given the fact that `scheduleJob` is typically called immediately after creating a new Job document, I think we can safely remove that compatibility code too.

zachmullen: I changed it, I kept in the `async` fallback check for now.

docs/migration-guide.rst
```diff
@@ -304,7 +304,9 @@ handler, you must call :py:func:`girder.events.unbind` on the existing mapping f
 Async keyword arguments and properties changed to async\_ PR #2817
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-In version 3.7 of python ``async`` is a `reserved keyword argument <https://www.python.org/dev/peps/pep-0492/#deprecation-plans/>`_. To mitigate any issues all instances of ``async`` in the codebase has changed to ``async_``. If the functions are called with ``async`` a deprecation warning will be given. This affects:
+In version 3.7 of python ``async`` is a `reserved keyword argument <https://www.python.org/dev/peps/pep-0492/#deprecation-plans/>`_.
+To mitigate any issues all instances of ``async`` in the codebase has changed to ``asynchronous``.
+If the functions are called with ``async`` a deprecation warning will be given. This affects:
```
brianhelba: I realized this last statement is not true. Calling with `async` is now invalid, not deprecated.

Issue #2987, opened 2019-04-16
Small feature request on girder client: retrieve the current user id


CharlesGueunet:
There is no native function to retrieve the current user id using the girder-client.
Suggestions:
* The `authenticate` method could return the user id (instead of dropping it).
* A small utility could be created like `getCurrentUser()`, calling a `get` on `user/me`

For now, the solution is to build the request like this:

```
currentUserId = client.get('user/me')['_id']
```

zachmullen:
👍  for having `authenticate` return the user document
👎 for a new method

PR #3025, opened 2019-05-03
improve the girder-client authenticate method: return the id of the connected user

CharlesGueunet:
This PR is linked to issue #2987 .

With this PR, the `authenticate` method (girder-client) returns the id of the connected user instead of nothing.
`authenticate` does not return the full user as response given by the server is not consistent between
login+passwd   &  apiKey authentication methods, which would be error prone.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -323,6 +323,12 @@ class GirderClient(object):
 
             self.setToken(resp['authToken']['token'])
 
+        if not 'user' in resp.keys():
+            print('Error trying to retrieve the user id during authentication')
+            return None
+
+        return resp['user']['_id']
+
     def setToken(self, token):
         """
         Set a token on the GirderClient instance. This is useful in the case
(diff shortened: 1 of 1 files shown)


clients/python/girder_client/__init__.py
```diff
@@ -323,6 +323,12 @@ def authenticate(self, username=None, password=None, interactive=False, apiKey=N
 
             self.setToken(resp['authToken']['token'])
 
+        if not 'user' in resp.keys():
```
zachmullen: Let's remove this block, if we got here we can assume that the response contains a `user` field.

CharlesGueunet:
> Actually, I think it would make sense to return the whole `user` dictionary instead of just the ID. Do you mind making that small change?

Done, however as written on the original commit, thus content of this dictionary is not the same depending on the type of authentication used (login/passwd or apiKey)

zachmullen:
> Done, however as written on the original commit, the content of the user dictionary is not the same depending on the type of authentication used (login/passwd or apiKey)

I had forgotten that. If we want to fix that, we should change it on the server IMO. This is the correct behavior for the client.

PR #3026, opened 2019-05-06
Move plugin settings into settings.py files

brianhelba:
Also, use a consistent naming for plugin settings and ensure every plugin setting has a default and a validator.


brianhelba:
@girder/developers PTAL.

plugins/google_analytics/girder_google_analytics/settings.py
```diff
@@ -0,0 +1,17 @@
...
+from girder.utility import setting_utilities
+
+
+class PluginSettings(object):
+    TRACKING_ID = 'google_analytics.tracking_id'
+
+
+@setting_utilities.default(PluginSettings.TRACKING_ID)
+def _defaultTrackingId():
+    return ''
```
zachmullen: For static values, should we use a function? I guess this is attempting to define a normative pattern, but an alternative would be:

```python
SettingDefault.defaults[PluginSettings.TRACKING_ID] = ''
```

Also, the default here was changed from `None` to `''`, is there a reason for that?

brianhelba: My intention was to avoid having plugin load mutate `SettingDefault.defaults`, and move that towards holding core setting defaults only.

As for the typing, it's a partial attempt to move towards trying to have stricter type validations for settings. In this instance, I should also update `_validateTrackingId` to require a string type.

brianhelba: > As for the typing, it's a partial attempt to move towards trying to have stricter type validations for settings. In this instance, I should also update `_validateTrackingId` to require a string type.

Actually, I'll try to deal with this more comprehensively in another PR.

PR #3027, opened 2019-05-06
Remove the "test" parameter from "resource/lookup"

brianhelba:
This functionality violates RESTful design and can easily be provided by client-side logic.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -533,17 +533,15 @@ class GirderClient(object):
 
         return self.get(route)
 
-    def resourceLookup(self, path, test=False):
+    def resourceLookup(self, path):
         """
         Look up and retrieve resource in the data hierarchy by path.
 
         :param path: The path of the resource. The path must be an absolute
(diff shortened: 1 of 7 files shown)


zachmullen:
I approve of this removal. Going to wait a while to add the check mark so that others can weigh in (and the tests can be fixed).

brianhelba:
@zachmullen This is now passing.

As a meta-process note, I'll assert that if a reviewer approves of a PR, they should simply mark their approval. In the ordinary case, the "branch prediction / reversion" policy provides a mechanism for negative feedback to affect the PR; my understanding is that this policy doesn't call for code reviewers to attempt to micro-manage the merge process by selectively timing their approvals. In this particular case, I try to make it my personal habit to leave non-trivial PRs open at least a day or so before merging regardless, to encourage feedback from the community.

zachmullen:
> my understanding is that this policy doesn't call for code reviewers to attempt to micro-manage the merge process by selectively timing their approvals

It doesn't, but either the reviewer or the author is free to be cautious with merging based on their evaluation of the likelihood and cost of a branch prediction miss.

PR #3029, opened 2019-05-06
Use "Authorization" as the default header for Basic Auth

brianhelba:
"Authorization" is a standard RFC7235 header, which is automatically supported by nearly any third-party HTTP client. Using "Authorization" as the preferred header makes it easier and less surprising to use third-party clients to authenticate with Girder.

"Girder-Authorization" remains supported by the server for compatibility.

### DIFF
--- a/docs/developer-cookbook.rst
+++ b/docs/developer-cookbook.rst
@@ -19,10 +19,9 @@ login process requires the client to make an HTTP ``GET`` request to the
 ``api/v1/user/authentication`` route, using HTTP Basic Auth to pass the user
 credentials. For example, for a user with login "john" and password "hello",
 first base-64 encode the string ``"john:hello"`` which yields ``"am9objpoZWxsbw=="``.
-Then take the base-64 encoded value and pass it via the ``Girder-Authorization``
-header (The ``Authorization`` header will also work): ::
+Then take the base-64 encoded value and pass it via the ``Authorization`` header: ::
 
-    Girder-Authorization: Basic am9objpoZWxsbw==
+    Authorization: Basic am9objpoZWxsbw==
 
 If the username and password are correct, you will receive a 200 status code and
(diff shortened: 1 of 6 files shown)


zachmullen:
Should we also change the existing [401 response handler](https://github.com/girder/girder/blob/master/girder/api/rest.py#L541) to also set an appropriate `WWW-Authenticate` header?

brianhelba:
For historical note, the use of `Girder-Authorization` was added by #858, to support HTTP Basic auth from multiple servers (Apache and Girder) simultaneously for a particular downstream deployment.

In general, I think this sort of scheme violates the separation of concerns between the web server and a reverse proxy and should be discouraged in typical deployments. For those deployments that may still want to provide this sort of setup, it is still possible to do via careful configuration of the proxy server.

brianhelba:
> Should we also change the existing [401 response handler](https://github.com/girder/girder/blob/master/girder/api/rest.py#L541) to also set an appropriate `WWW-Authenticate` header?

Yes, I think so, since RFC7235 4.1 states:
> A server generating a 401 (Unauthorized) response MUST send a WWW-Authenticate header field containing at least one challenge.

However, I think the practical utility of this header may be limited, since it seems to typically just cause browsers to pop up a Basic Auth dialog. I suspect that it's going to have no impact on Ajax or user agents like `curl`.

I'll add it now.

zachmullen:
For the login endpoint the type should be `Basic`, for others it should maybe be `Bearer`, even though we aren't OAuth per se, it does fit with that protocol.

brianhelba:
@zachmullen Given some uncertainty about how to handle certain details of a `WWW-Authenticate` header and the non-critical importance of such a change, I've moved it into #3030. I do think we should attempt to follow up on that issue soon, however.

I'll consider this PR ready for review now as-is.

PR #3031, opened 2019-05-07
Run Molecule tests for the Ansible role in CI

### LINKED ISSUE #2997: Molecule tests of Ansible library need to be run in CI
Converting these to use Docker instead of Vagrant may make this easier.

brianhelba:
This also identifies and fixes two bugs with the Ansible role:
* Remove an obsolete CLI option from the Ansible role
* Set important environment variables for the Girder service

Fixes #2997

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -36,6 +36,17 @@ commands =
         docs \
         build/docs/python
 
+[testenv:ansible]
+skip_install = true
+skipsdist = true
+changedir = devops/ansible-role-girder
+passenv = DOCKER_*
+deps =
+  molecule
+  docker
+commands =
+  molecule test
+
 [testenv:release]
 skip_install = true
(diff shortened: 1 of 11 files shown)


brianhelba:
@danlamanna @girder/developers PTAL.

tox.ini
```diff
@@ -36,6 +36,17 @@ commands =
         docs \
         build/docs/python
 
+[testenv:ansible]
+skip_install = true
+skipsdist = true
+changedir = devops/ansible-role-girder
+passenv = DOCKER_*
```
danlamanna: These are set from circle?

brianhelba: Yes, and the `setup_remote_docker` CircleCI sets up a remote Docker daemon to run commands.

PR #3032, opened 2019-05-07
Fix YAML formatting in several places

brianhelba:
Command:
  `yamllint -d '{extends: default, rules: {line-length: false}}' .`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,4 @@
+---
 version: 2
 
 jobs:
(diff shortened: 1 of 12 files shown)


jeffbaumes:
If we care about this, should we add a `yamllint` checking test to CI?

brianhelba:
I considered it. The Ansible code is / will all be linted when we run Molecule in CI. The other files are primarily scattered config files, associated with different ecosystems.

I wouldn't object to adding a CI `yamllint` check if it could fit cleanly into an existing modern test harness (i.e. Tox), but I didn't consider it a priority. To get it working robustly, there will need to be a lot of extra configuration to have it exclude YAML in all of our ignored build directories.

If anyone wants to at least file a bug, I'd be fine with that too.

jeffbaumes:
I'm totally ok with an occasional cleanup like this, if it's not completely trivial to add a test. Not worth much effort IMO.

PR #3033, opened 2019-05-08
Cleanup some devops code style

brianhelba:
* Ensure Ansible specifies "become_user" when appropriate
* Pass Ansible role variables under "vars"
* Cleanup comment spelling in CircleCI

### DIFF
--- a/.circleci/playbook.yml
+++ b/.circleci/playbook.yml
@@ -27,7 +27,8 @@
         mongodb_version: 3.6
 
     - role: ansible-role-girder
-      girder_daemonize: false
+      vars:
+        girder_daemonize: false
 
   post_tasks:
     # Needed for replicaset/sharded tests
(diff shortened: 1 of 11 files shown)


brianhelba:
@girder/developers PTAL

.circleci/playbook.yml
```diff
@@ -26,11 +27,13 @@
...
+      vars:
+        girder_daemonize: false
 
   post_tasks:
     # Needed for replicaset/sharded tests
     - name: Install mongos
       apt:
         name: mongodb-org-mongos
       become: true
+      become_user: root
```
manthey: Is there a reason to set parameters like `become_user` to their default values?

brianhelba: In the case of `become_user: root`, it is important in the case where other parts of a playbook use `become_user` to execute steps as a different user (e.g. the original SSH user is not the same user as the one that's pip-installing the package). I'm still trying to figure out best practices for how to install Girder to a more secure daemon user, and until we have a comprehensive pattern in place, I'd like to be explicit everywhere.

I could foresee globally removing this again in the future (for the sake of concision), once things settle down.

PR #3036, opened 2019-05-09
Remove insecure sha512 password hashing

### LINKED ISSUE #1167: Switch to Passlib to abstract away password hashing
We're going to need more support for additional password hashing schemes in the future, and [Passlib](https://pythonhosted.org/passlib/) is a well-established way to abstract away the details of integrating each hashing library directly.

Additionally, Passlib [has a utility](https://pythonhosted.org/passlib/lib/passlib.hash.bcrypt_sha256.html) for dealing with the 72-character length limitation in bcrypt, which Girder currently does not address.

brianhelba:
This also switches to use the high-level passlib library for all password hashing. This improves security and provides mechanisms to seamlessly migrate user databases to use other hash algorithms (like Argon2) in the future.

Fixes #1167


pytest_girder/pytest_girder/fixtures.py
```diff
@@ -21,9 +21,9 @@ def bcrypt():
     """
     Mock out bcrypt password hashing to avoid unnecessary testing bottlenecks.
     """
-    with mock.patch('bcrypt.hashpw') as hashpw:
-        hashpw.side_effect = lambda x, y: x
-        yield hashpw
+    from passlib.hash import bcrypt
+    with mock.patch('girder.models.user.bcrypt', new=bcrypt.using(rounds=4)) as fastBcrypt:
```
zachmullen: Is 4 rounds very fast?

brianhelba: It's the lowest we can go: https://passlib.readthedocs.io/en/stable/lib/passlib.hash.bcrypt.html#passlib.hash.bcrypt

danlamanna:
There certainly should be some way to fully mock bcrypt. It's called for
every user fixture, combined with parametrized tests it can add significant
time to the test suite.

zachmullen:
We're only mocking `hashpw`, so I guess we're expecting that `verify` calls that internally?

brianhelba:
@danlamanna I'll see if I can mock out with [`passlib.hash.plaintext`](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.plaintext.html) instead.

pytest_girder/pytest_girder/fixtures.py
jbeezley: Is there reason to run the hashing inside the test at all rather than make it a no-op?

zachmullen: We should probably have one test that does run it, but the rest should mock it with a no-op IMO.

brianhelba:
> Is there reason to run the hashing inside the test at all rather than make it a no-op?

@jbeezley We cannot mock out `hashlib` to be a no-op. It does password validation, returns new hashes, etc. However, as stated, I'll see if we can use `passlib.hash.plaintext` to minimize the computational cost of the internal hashing step.

brianhelba:
Here's some values for how long various hash functions take (using iPython `%%timeit` on my laptop's VM):
* `bcrypt.using(rounds=12)`: 258ms
* `bcrypt.using(rounds=4)`: 1.16ms
* `plaintext`: 563ns

I also looked at the practical `user` fixture performance (using `pytest --durations=0 -k testInitializeOtp`). There was no practical difference between 4 round bcrypt and plaintext: both took around ~700ms of fixture setup time.

jbeezley:
I would say that's fast enough.

brianhelba:
@zachmullen @danlamanna @jbeezley I updated the tests to use `plaintext`, but more importantly, I switched to using the higher-level `CryptContext` API, which makes everything even more maintainable now, and provides functionality to [migrate to other hashes in the future](https://passlib.readthedocs.io/en/stable/narr/context-tutorial.html#deprecation-hash-migration) (though we'd have to implement 2-3 lines of extra logic to utilize that).

brianhelba:
Also this now does mutation instead of mocking, which is more maintainable, given the way the previous mocked imports functioned.

girder/models/user.py
```diff
@@ -85,6 +90,10 @@ def validate(self, doc):
             # an email address from a login
             raise ValidationException('Login may not contain "@".', 'login')
 
+        if 'hashAlg' in doc:
```
danlamanna: Should we add some form of structured comment in the event that we implement #946?

brianhelba: The simple existence of your comment here just created a reference on that issue page :astonished:

PR #3037, opened 2019-05-09
Allow ansible-role-girder to implicitly create a Python 3 virtualenv

brianhelba:
* Allow ansible-role-girder to implicitly create a Python 3 virtualenv
  * This makes the role significantly more friendly to use in the default case, while not changing its API surface area or removing the ability to use an existing virtualenv.
* Allow ansible-role-girder to create a virtualenv when possible
  * This puts the previous change into effect for internal playbooks.
* Don't use ansible-role-girder for CI integration tests
  * The CircleCI machine executor, used by the Girder integration tests, runs on Ubuntu 14.04 and uses nvm to manage NodeJS, neither of which are supported by the `girder.girder` or `girder.mongodb` roles. Worse, the use of pyenv makes it impossible to run Ansible under the pyenv-managed Python interpreter, due to the fact that the `python-apt` package is not available (in a working version) on PyPI and does not propagate from it's APT-installed location to the pyenv. Since we now have first-class tests for the Ansible role, trying to use it in the integration tests provides no net benefit.


brianhelba:
I just verified that this also works when provisioning a fresh environment via Vagrant.

brianhelba:
@girder/developers PTAL

devops/ansible-role-girder/molecule/ansible-module/playbook.yml
```diff
@@ -3,24 +3,9 @@
   hosts: all
 
   vars:
-    girder_virtualenv: "{{ ansible_user_dir }}/.venvs/girder"
+    girder_virtualenv: "{{ ansible_user_dir }}/.virtualenvs/girder"
```
danlamanna: Is there any reason to define these in the tests anymore?

brianhelba: In this playbook, it's because we reference `girder_virtualenv` later in `post_tasks`. In other playbooks, where we don't need to work with `girder_virtualenv` after the role completes, we accept the role default.

PR #3038, opened 2019-05-10
Move the login regex out of the config file

brianhelba:
This also fixes a Pytest warning from the CherryPy config file:
"DeprecationWarning: invalid escape sequence \d"

### DIFF
--- a/girder/conf/girder.dist.cfg
+++ b/girder/conf/girder.dist.cfg
@@ -40,12 +40,6 @@ disable_event_daemon = False
 # log_max_info_level = "CRITICAL"
 
 [users]
-# Regular expression that logins must match. All logins are lower()ed before validation.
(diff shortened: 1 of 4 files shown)


brianhelba:
@zachmullen The rationale is that this level of configuration is probably never necessary for deployments.
* Allowing the user `login` to vary widely prevents clients from making assumptions about the regex rules.
* Worse, since the setting value is practically never changed, clients may start making assumptions about `login` names, which would break if the regex is ever changed too significantly.
* Finally, its presence in the setting file causes clutter and distracts from much more important settings that admins should review and consider overriding.

If a specific use case does call for limiting `login` regexes (e.g. to coordinate with some out-of-band auth system), this could be done by creating a plugin. For instance, the `oauth` plugin simply auto-generates `login` names using its own internal logic.

PR #3039, opened 2019-05-15
Fix the client build.

manthey:
Some upstream package used in building the client has changed.  It now seems necessary to use the same version of underscore in the girder package.json as in the web client src package.json and to include babel-polyfill as part of the build.

This also backports the changes from #2999 to allow the build to pass with the latest moto version.

Further, we have to pin bootstrap-switch or it doesn't disable in a test properly.


manthey:
I feel this is the wrong approach, and that pinning something else should resolve this problem, but I haven't figured out a smaller change set that works.  This PR (or something equivalent) is needed to build working plugins with web clients (including those in the main repo).

PR #3040, opened 2019-05-15
Remove path from auth token cookies

zachmullen:
Somehow I only just ran into this bug, possibly a result of a recent Chrome update.

On my localhost:8080 domain, I had a girderToken cookie set *without* a path field, and the ones sent by the server on login did contain a path field. The one without the path field (the old, expired one) was taking precedence when reading `document.cookie`, which means that a hard page refresh reverted to a logged out state every time.

The correct behavior is not to set a path value at all.

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -1132,7 +1132,6 @@ class Resource(object):
 
         cookie = cherrypy.response.cookie
         cookie['girderToken'] = str(token['_id'])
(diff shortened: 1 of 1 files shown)


zachmullen:
It seems we might just be doing some wonky stuff with cookies in the backbone client. It merits a deeper investigation. Closing this until I can look into it more.

PR #3041, opened 2019-05-16
Add empty meta field on new item creation, and direct item load if not present

### LINKED ISSUE #3035: Always return `meta` field from resource APIs, even if it doesn't explicitly exist in the database
This issue asks for two remediations:
1. Augment `GET /folder/{id}` and `GET /item/{id}` to always return a `meta` field, which should just be an empty dict if the `meta` value does not exist on the raw resource document (see @zachmullen's [comment](https://github.com/girder/girder/pull/2888#issuecomment-490657071)).
2. Always place an empty `meta` dict on new item and folder resources when they are created.

AlmightyYakob:
Fixes #3035

### DIFF
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -142,6 +142,11 @@ class Folder(AccessControlledModel):
                 self.update({'_id': doc['_id']}, {'$set': {
                     'lowerName': doc['lowerName']
                 }})
+            if 'meta' not in doc:
+                doc['meta'] = {}
+                self.update({'_id': doc['_id']}, {'$set': {
+                    'meta': {}
+                }})
 
             self._removeSupplementalFields(doc, fields)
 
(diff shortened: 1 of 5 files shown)


AlmightyYakob:
@girder/developers This actually only fixes this issue for items, not folders, but I'm about to do the same for folders.

girder/models/folder.py
```diff
@@ -544,6 +550,17 @@ def updateFolder(self, folder):
         # Validate and save the folder
         return self.save(folder)
 
+    def filter(self, doc, user=None, additionalKeys=None):
```
AlmightyYakob: Like in item.py, this function solves the issue of the meta field not showing up on a folder with no metadata when hitting the base /folder endpoint. However, the empty meta field is not saved here, so while consistent to the user, internally under this circumstance the meta tag can't be relied on. If you guys have any thoughts on if we should change this they're welcome.

zachmullen:
LGTM, seems like a reasonable way to ensure the meta field always shows up via the REST API.

@brianhelba any idea what's wrong with the ansible tests?

girder/models/folder.py
brianhelba: I agree that this isn't ideal, but I can't think of anything better myself either.

@girder/developers Let us know if you have better suggestions.

PR #3045, opened 2019-05-22
Remove the "default" argument to Setting.get

### LINKED ISSUE #2994: Remove `default` arg for `Setting.get`

zachmullen:
Fixes #2994

### DIFF
--- a/girder/api/v1/system.py
+++ b/girder/api/v1/system.py
@@ -11,7 +11,7 @@ import logging
 from girder import plugin
 from girder.api import access
 from girder.constants import TokenScope, ACCESS_FLAGS, VERSION
-from girder.exceptions import GirderException, ResourcePathNotFound, RestException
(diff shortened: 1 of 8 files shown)


zachmullen:
If we want to preserve the behavior of fetching the defaults, I'd be happy to add in another setting endpoint that returns default values.

zachmullen:
@brianhelba PTAL, there's only one potentially controversial change here, which is to the REST API; to simplify things, I've removed the ability to fetch the default setting values via the REST API, which mostly just affects the admin UI in the Backbone client.

brianhelba:
@zachmullen Thanks for taking this on; this is exactly what I wanted to do.

In the future, if we ever want to try and enhance the settings page, we could try to differentiate between unset settings and those that are set but equal the default. However, until the front-end has the technology to save settings on an individual basis, I don't think we're missing anything if we stop returning the default values separately.

PR #3047, opened 2019-05-22
Start Girder in production mode by default

brianhelba:
This optimizes Girder for a production environment (via "pip install") over a development environment (via "git clone"). This should significant improve security by default.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -26,6 +26,13 @@ For more options for building the web client, run: ::
 
     girder build --help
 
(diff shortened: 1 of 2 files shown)


brianhelba:
@girder/developers PTAL.

danlamanna:
+1 - I see this as an extension of https://github.com/girder/girder/pull/2565.

manthey:
There should be a note about this in deploying for development.

PR #3048, opened 2019-05-22
Do not run pytest tests via CTest

zachmullen:
Reverts 56348ee549badd80ba0637b2d452cea005195ffb

@manthey what was the original rationale for doing this? It was causing our pytest suite to run twice in a row in CI.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -62,8 +62,6 @@ if(RUN_CORE_TESTS)
   add_python_test(py_client.cli BIND_SERVER RESOURCE_LOCKS py_client_test_dir)
   add_python_test(py_client.lib BIND_SERVER RESOURCE_LOCKS py_client_test_dir)
 
-  add_pytest_test(core)
-
   add_web_client_test(data_filesystem "${PROJECT_SOURCE_DIR}/girder/web_client/test/spec/dataSpec.js" ASSETSTORE filesystem)
   add_web_client_test(data_gridfs "${PROJECT_SOURCE_DIR}/girder/web_client/test/spec/dataSpec.js" ASSETSTORE gridfs)
   add_web_client_test(data_gridfsrs "${PROJECT_SOURCE_DIR}/girder/web_client/test/spec/dataSpec.js" ASSETSTORE gridfsrs RESOURCE_LOCKS replicaset REQUIRED_FILES "${MONGOD_EXECUTABLE}")
--- a/tests/PythonTests.cmake
+++ b/tests/PythonTests.cmake
@@ -12,8 +12,6 @@ else()
   set(_separator ":")
 endif()
 
-find_program(PYTEST_EXECUTABLE NAMES pytest)
-
 function(python_tests_init)
   add_test(
     NAME py_coverage_reset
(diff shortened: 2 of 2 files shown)


manthey:
The rational was so you could just run ctest and test everything.  That way, as a developer, you could run one test command rather than two.  We were cherry-picking which ctests to run in CI, so we weren't running pytests via ctest in CI.

manthey:
It makes local testing easier.

zachmullen:
Since pytest provides its own harness, I think it makes sense to pull this out of CTest. (We have on the road map to eventually remove all use of CMake/CTest in Girder, and use language-native harnesses instead.)

manthey:
I'd rather we keep this functionality and explicitly skip running pytest in ctest is there is some reason to run it separately.  Or not run it separately.  Or run ctest via pytest.  One test command, please, not two.

zachmullen:
If we move to a single harness, IMO it should be `tox`, which could have an env that runs ctest internally. That said, I see no reason to promise a single harness that could run all tests if we have a codebase that is half javascript/web and half python. If someone really desires that, it's easy enough to make a shell alias for it.

manthey:
We already *have* a single harness that runs all of the tests.  We are moving away from that with this PR. 

Of course it is easy to have a shell script to run multiple harnesses -- but we also set up tooling to make things easier for development, even if not strictly necessary.  We don't have to have any code to download test data -- we could require developers to curl it themselves.  It is all a balance between maintaining tools and providing tools to ease development.

zachmullen:
If I go ahead and add a testenv to tox that runs ctest, and make our CI use that, will that work for you @manthey ?

manthey:
That would be great -- my real desire is to have tests easy enough to run that most developers run them *before* pushing branches.

brianhelba:
Ideally, I think Girder would be most maintainable and approachable for developers if we had two systems for running tasks (including tests): `tox` and `npm`, because we are fundamentally dealing with two different ecosystems.

If we are trying to make things easier for developers, let us not forget that the developer community who care about easily invoking tests are the exact same people that are burdened by the difficult-to-interpret outputs, overly-complex wrapper scripts, and deviations from typical community practices that plague our current test environment. End users do not need to run tests, so any additional maintenance burden of a single command approach will fall only on the audience it is supposed to benefit.

However, I certainly agree that using both `cmake` and `tox` to invoke tests is not ideal. We shouldn't have three task runners. However, if separating them explicitly for now gives us greater flexibility to port specific cases, make things run more cleanly, etc., then I'd support it. If we don't make some short term trade-offs, I think we'll be stuck with CMake indefinitely. Accordingly, do we have a plan (or can we make one) to move specific high-impact tests (`flake8` is a great example) out of CMake and into Tox?

brianhelba:
> That would be great -- my real desire is to have tests easy enough to run that most developers run them _before_ pushing branches.

I wholeheartedly agree with this statement.

I'll just comment as one data point that I personally often find running `ctest` to be too burdensome for my personal tolerance; I would typically let the tests be run by CI, and run locally only if strictly necessary. I imagine that many other non-core developers have an even lower tolerance.

Tox is much more friendly to running locally, since it provides almost complete self-bootstrapping and isolates my top-level virtualenv from the changes it makes.

zachmullen:
> Accordingly, do we have a plan (or can we make one) to move specific high-impact tests (flake8 is a great example) out of CMake and into Tox?

The plan is to get everything out of CMake. Flake8 (and any other python tooling) will go under tox. Normally web stuff should be harnessed by an npm script, however the way we currently run them (with a real live python server) means they are part of the python environment for now. We've already got the infrastructure in place to run web client tests through pytest, so porting the rest of those shouldn't be that much work.

I had kind of assumed that we would move the remaining python unittest tests out of CTest at the same time as we ported them to use pytest rather than unittest, but that isn't even strictly necessary if we'd rather jettison cmake sooner. The only thing we really need to consider is the ExternalData capabilities. Anyone have thoughts on that?

manthey:
> The only thing we really need to consider is the ExternalData capabilities. Anyone have thoughts on that?

Yes.  I'm already doing that for large_image on the girder-3 branch: https://github.com/girder/large_image/blob/girder-3/test/utilities.py#L14-L46

zachmullen:
I'd like to start working on the comprehensive merging, but I'd also like to make sure we aren't duplicating pytest runs in CI in the near term. Can we merge this and I can open a follow-up PR for the larger harness-unification efforts?

brianhelba:
> The only thing we really need to consider is the ExternalData capabilities. Anyone have thoughts on that?

[Git-LFS](https://help.github.com/en/articles/configuring-git-large-file-storage)

PR #3049, opened 2019-05-23
Remove "nginx" Molecule scenario

brianhelba:
This functionality is now tested and provided by https://github.com/girder/ansible-role-girder-nginx


brianhelba:
This does not impact the Girder 3 release candidate. It's just removing dead code.

PR #3053, opened 2019-05-28
Make Ansible always use system Python 3 as the base for implicit virtualenvs

brianhelba:
`ansible_python_interpreter` may not always be set, and using alternate variables for the current Python interpreter on the target could yield another virtualenv, pyenv, etc. Users needing more control over the base of the virtualenv can always create one explicitly.


danlamanna:
> ansible_python_interpreter may not always be set

Source?

brianhelba:
> > ansible_python_interpreter may not always be set
> 
> Source?

It turns out this is a user-set variable, used to instruct Ansible which interpreter to always use. If it's not set and Ansible chooses the interpreter, there are different facts (primarily `ansible_python.executable`) expressing which variable is used.

PR #3057, opened 2019-06-04
Harden Ansible tests to always run under Python 3

brianhelba:


brianhelba:
FYI, installing NodeJS also installs Python 2, since `node-gyp` still requires it. This means we always get Python 2 even if we don't want it. Previously, the Testinfra tests were running under Python 2, but the provisioning was happening under Python 3.

PR #3062, opened 2019-06-06
Add --dev option to girder serve command

AlmightyYakob:
This adds a --dev flag for setting development mode on when running girder serve, as the default is now production. In my implementation, I've given testing mode priority in the event that they're both set. I'm not sure what the best approach is here, but that seemed reasonable. Another approach could be to show a console statement (and exit) if both flags are passed when calling serve.


AlmightyYakob:
Should the dev flag also be included [here](https://github.com/girder/girder/pull/3062/files#diff-f94a7a2e0c5304caf743b4878a5f70fdR56)? I'm not sure where this option normally gets set but it seems like it does for development.

danlamanna:
Another possibility is adding a `--mode` flag which takes one of `testing`, `development`, `production`.

brianhelba:
@danlamanna Do you have an opinion as to which should be used?

danlamanna:
I would prefer mode because the internal code in `utility/server.py` would be cleaner if 1 argument was passed around instead of 3. That being said, it would be really nice if we could have consistent flags for the mode between `girder build` and `girder serve`, and `girder build` uses `--dev`. This would require creating some sort of alias at the CLI level in either `build`, `serve`, or both. Does anyone have any thoughts on this?

AlmightyYakob:
I agree with @danlamanna, having the mode argument is cleaner, and would also get rid of the precedence between testing and dev flags. @brianhelba If you agree with this I can make these changes

AlmightyYakob:
It appears that something I've done is causing girder to not read my `~/.girder/girder.cfg` file properly. I'll have to take a look at this.

girder/settings.py
```diff
@@ -114,7 +114,7 @@ def _defaultRouteTable():
     @staticmethod
     @setting_utilities.default(SettingKey.SECURE_COOKIE)
     def _defaultSecureCookie():
-        return config.getConfig()['server']['mode'] == 'production'
+        return config.getServerMode() == PRODUCTION_MODE
```
brianhelba: @AlmightyYakob #3063 removed this setting entirely. Can you merge master into this branch?

girder/utility/server.py
```diff
@@ -137,21 +135,21 @@ def reconcileRouteTable(routeTable):
...
+def setup(mode=PRODUCTION_MODE, plugins=None, curConfig=None):
     """
     Configure and mount the Girder server and plugins under the
     appropriate routes.
 
     See ROUTE_TABLE setting.
 
-    :param test: Whether to start in test mode.
+    :param mode: The server mode to start in.
     :param plugins: List of plugins to enable.
```
brianhelba: Might as well add a `:type mode: string` here.

girder/utility/server.py
```diff
@@ -64,13 +65,10 @@ def configureServer(test=False, plugins=None, curConfig=None):
     mimetypes.add_type('application/font-woff', '.woff')
 
     curConfig.update(appconf)
+    curConfig['server']['mode'] = mode
```
brianhelba: I realized there is a subtle change here that could be the source of future bugs. The question is: "what is the source of truth for the default mode?":
1. The file `girder/conf/girder.dist.cfg`
2. The default parameter of `configureServer` of `girder/utility/server.py`
3. The default parameter of `setup` of `girder/utility/server.py`
4. The default argument value in `girder/cli/serve.py`
5. The default argument value in `girder/cli/build.py`

Put anther way, if we wanted to change the default server mode back to "development", and we didn't change all 5 of these places together, inconsistencies and bugs could arise.

So, I'd propose:
* make the file `girder/conf/girder.dist.cfg` the source of truth
* set the default function arguments to be `mode=None`, and only override values (like `curConfig['server']['mode'] = mode`) if mode is set / truthy
* make the default argument value from the CLIs be `None` / unset

girder/constants.py
```diff
@@ -29,6 +29,11 @@
 STATIC_PREFIX = os.path.join(sys.prefix, 'share', 'girder')
 STATIC_ROOT_DIR = os.path.join(STATIC_PREFIX, 'static')
 
+# Server modes
+PRODUCTION_MODE = 'production'
```
brianhelba: I know this seems minor, but could we scope these under a class like `ServerMode`, then have `ServerMode.PRODUCTION`, etc.?

I'd like to avoid further spamming this namespace with more flat values. However, if you disagree, feel free to push back on this suggestion.

girder/cli/serve.py
```diff
@@ -5,17 +5,23 @@
...
 @click.option('-d', '--database', default=cherrypy.config['database']['uri'],
               show_default=True, help='The database URI to connect to')
 @click.option('-H', '--host', default=cherrypy.config['server.socket_host'],
               show_default=True, help='The interface to bind to')
 @click.option('-p', '--port', type=int, default=cherrypy.config['server.socket_port'],
               show_default=True, help='The port to bind to')
-def main(testing, database, host, port):
+def main(dev, mode, database, host, port):
+    if (dev):
+        mode = DEVELOPMENT_MODE
```
brianhelba: If both `--dev` and `--mode` are set, should we emit an error / warning?

girder/cli/build.py
```diff
@@ -22,27 +22,33 @@
...
-def main(dev, watch, watch_plugin, npm, reinstall):
+def main(dev, mode, watch, watch_plugin, npm, reinstall):
     if shutil.which(npm) is None:
         raise click.UsageError(
             'No npm executable was detected.  Please ensure the npm executable is in your '
             'path, use the --npm flag, or set the "NPM_EXE" environment variable.'
         )
 
+    if (dev):
+        mode = DEVELOPMENT_MODE
```
brianhelba: If both `--dev` and `--mode` are set, should we emit an error / warning?

girder/cli/build.py
```diff
@@ -22,27 +22,33 @@
...
 @click.command(name='build', help='Build web client static assets.')
-@click.option('--dev/--no-dev', default=False,
-              help='Build girder client for development.')
+@click.option('--dev', default=False, is_flag=True, help='Sets the server mode to development')
+@click.option('--mode', type=click.Choice([PRODUCTION_MODE, DEVELOPMENT_MODE]),
+              default=PRODUCTION_MODE, show_default=True, help='Specify the server mode')
 @click.option('--watch', default=False, is_flag=True,
-              help='Build girder library bundle in watch mode (implies --dev --no-reinstall).')
+              help='Build girder library bundle in '
+              'watch mode (implies mode=development --no-reinstall).')
```
brianhelba: Typo: "implies --mode=development" (note the "--")

girder/cli/build.py
```diff
@@ -22,27 +22,33 @@
...
 @click.option('--watch-plugin',
-              help='Build a girder plugin bundle in watch mode (implies --dev --no-reinstall).')
+              help='Build a girder plugin bundle in '
+              'watch mode (implies mode=development --no-reinstall).')
```
brianhelba: Typo: "implies --mode=development" (note the "--")

girder/constants.py
AlmightyYakob: I agree, I'll make those changes

girder/utility/server.py
AlmightyYakob: I agree. This also fixes the issue that I was having around not reading the config file properly.

girder/cli/serve.py
AlmightyYakob: I think so. I'll add a click error to my next commit.

girder/cli/build.py
AlmightyYakob: Same as above

AlmightyYakob:
@brianhelba I've made those changes

girder/cli/serve.py
```diff
@@ -5,17 +5,28 @@
...
+    ]), default=None, show_default=True, help='Specify the server mode')
 @click.option('-d', '--database', default=cherrypy.config['database']['uri'],
               show_default=True, help='The database URI to connect to')
 @click.option('-H', '--host', default=cherrypy.config['server.socket_host'],
               show_default=True, help='The interface to bind to')
 @click.option('-p', '--port', type=int, default=cherrypy.config['server.socket_port'],
               show_default=True, help='The port to bind to')
-def main(testing, database, host, port):
+def main(dev, mode, database, host, port):
+    if (dev and mode):
```
brianhelba: Parens not necessary.

girder/cli/serve.py
```diff
@@ -5,17 +5,28 @@
...
+        raise click.ClickException('Conflict between --dev and --mode')
+    if (dev):
```
brianhelba: Parens not necessary.

girder/utility/server.py
```diff
@@ -64,13 +65,11 @@ def configureServer(test=False, plugins=None, curConfig=None):
     mimetypes.add_type('application/font-woff', '.woff')
 
     curConfig.update(appconf)
+    if (mode):
```
brianhelba: Parens not necessary.

girder/utility/server.py
```diff
@@ -176,8 +176,8 @@ def setup(test=False, plugins=None, curConfig=None):
         if name != constants.GIRDER_ROUTE_ID and name in pluginWebroots:
             cherrypy.tree.mount(pluginWebroots[name], route, appconf)
 
-    if test:
-        application.merge({'server': {'mode': 'testing'}})
+    if (mode and mode != ServerMode.PRODUCTION):
```
brianhelba: Parens not necessary.

brianhelba: Do we need the `mode != ServerMode.PRODUCTION` check? If `setup()` is passed an affirmative `mode`, should we just put it into the `application` config?

More significantly, is this needed here at all? Isn't `configureServer` (which is always called) already doing the same thing (updating `curConfig['server']['mode']` with a passed `mode`)?

girder/utility/server.py
AlmightyYakob: Can you tell I've been writing a lot of JS? :grin:

girder/utility/server.py
AlmightyYakob: I agree, it doesn't look like it's needed, and removing it has no noticeable effect.

AlmightyYakob:
@brianhelba I've made those changes, and also rebased off master to fix failing tests around flake8. Lmk if this all looks good and if so please update your review

danlamanna:
Can you merge master into this? There's a lot of unrelated changes in the diff.

girder/cli/build.py
```diff
@@ -22,27 +22,35 @@
 
 
 @click.command(name='build', help='Build web client static assets.')
-@click.option('--dev/--no-dev', default=False,
-              help='Build girder client for development.')
+@click.option('--dev', default=False, is_flag=True, help='Sets the server mode to development')
```
danlamanna:
```suggestion
@click.option('--dev', default=False, is_flag=True, help='Alias for --mode=development')
```

girder/cli/serve.py
```diff
@@ -5,17 +5,28 @@
 
 from girder import _attachFileLogHandlers
 from girder.utility import server
+from girder.constants import ServerMode
 
 
 @click.command(name='serve', short_help='Run the Girder server.', help='Run the Girder server.')
-@click.option('-t', '--testing', is_flag=True, help='Run in testing mode')
+@click.option('--dev', default=False, is_flag=True, help='Sets the server mode to development')
```
danlamanna:
```suggestion
@click.option('--dev', default=False, is_flag=True, help='Alias for --mode=development')
```

girder/cli/serve.py
```diff
@@ -5,17 +5,28 @@
...
+@click.option('--mode', type=click.Choice([
+    ServerMode.PRODUCTION,
+    ServerMode.DEVELOPMENT,
+    ServerMode.TESTING
```
danlamanna: What runs `girder serve --mode=testing`?

AlmightyYakob: I'm not sure, I was just preserving what was already there.

PR #3063, opened 2019-06-09
Set cookie security from HTTPS status, not production mode

brianhelba:
First, this completely removes the SECURE_COOKIE setting from Girder. Since this setting never had a GUI for setting it, it is unlikely that most end users would have configured it.

Second, this no longer sets Girder's cookie "secure" attribute based on production vs development mode. Since Girder now starts in production mode by default, developers running Girder from a Git checkout for the first time will typically be in production mode with an HTTP connection.

Rather, this causes Girder to now auto-detect the scheme of the incoming request (for which the cookie is being set), and set the "secure" attribute accordingly. In production environments, this will by detected by CherryPy's proxy tools via the "X-Forwarded-Proto" header, which reverse proxies should set. In test mode, the HTTPS status must
simulated differently, since CherryPy proxy tools cannot be easily enabled.

From now on, an absent or misconfigured reverse proxy in production mode will cause Girder to gracefully degrade to using less secure cookies, rather than breaking.

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -1135,7 +1135,9 @@ class Resource(object):
         cookie['girderToken']['path'] = '/'
         cookie['girderToken']['expires'] = int(days * 3600 * 24)
 
-        if Setting().get(SettingKey.SECURE_COOKIE):
+        # CherryPy proxy tools modify the request.base, but not request.scheme, when receiving
+        # X-Forwarded-Proto headers from a reverse proxy
+        if cherrypy.request.scheme == 'https' or cherrypy.request.base.startswith('https'):
             cookie['girderToken']['secure'] = True
 
         return token
--- a/girder/settings.py
+++ b/girder/settings.py
@@ -8,7 +8,7 @@ import six
 
 from girder.constants import GIRDER_ROUTE_ID
 from girder.exceptions import ValidationException
-from girder.utility import config, setting_utilities
+from girder.utility import setting_utilities
 
 
 class SettingKey(object):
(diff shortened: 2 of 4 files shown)


zachmullen:
> gracefully degrade to using less secure cookies, rather than breaking

Silently degrading security is arguably worse than breaking, but I think this change is OK, assuming that our canonical reverse proxy (e.g. nginx) configurations contain the appropriate directives.

brianhelba:
> Silently degrading security is arguably worse than breaking

In general, I'd agree, but given the particular value of the "secure" cookie attribute, I don't think it practically compromises much if it's absent. In normal client usage, `Girder-Token` headers with the same info will be sent by subsequent requests anyway. I suppose an attacker could serve a malicious site which requests a non-preflighted CORS to an `http://` address of a misconfigured Girder server to which the target was already logged in, then also MITM / sniff the unencrypted HTTP request header to grab the `Cookie` value. :man_shrugging: 

The question is, can we effectively warn server administrators that their server is misconfigured? This could be either Girder getting served for true production use over HTTP or where a reverse proxy is not properly informing Girder (via `X-Forwarded-Proto`) that it is indeed wrapped by HTTPS. Without generating a lot of scary false positives in development, I'm not sure how to do this, given the current way that Girder's `development` vs `production` modes actually are configured.

brianhelba:
@jcfr See my comment here in the PR:

> The question is, can we effectively warn server administrators that their server is misconfigured? This could be either Girder getting served for true production use over HTTP or where a reverse proxy is not properly informing Girder (via `X-Forwarded-Proto`) that it is indeed wrapped by HTTPS. Without generating a lot of scary false positives in development, I'm not sure how to do this, given the current way that Girder's `development` vs `production` modes actually are configured.

Unless you have any ideas, I don't see a way to do so accurately. Also, on the scale of security risks, I see this (that is, "secure" cookies) as relatively low; HTTPS is much more important, but it's out of scope for Girder server to handle directly.

zachmullen:
Agree that this isn't a major risk, HSTS is a more important configuration setting and will ensure the same security policy, though defense-in-depth is of course important. As long as our "official" reverse proxy examples show the correct settings, that's sufficient IMO.

PR #3067, opened 2019-06-11
Fix searching in database accept sort field

RemiCecchinato:

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -312,7 +312,7 @@ class Model(object):
 
         return cursor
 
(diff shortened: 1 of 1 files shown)


zachmullen:
@RemiCecchinato what is the objective here? `findOne` is meant to be used in scenarios where the expected document is unique based on the search criteria, so IMO a sort doesn't make sense.

RemiCecchinato:
> @RemiCecchinato what is the objective here? `findOne` is meant to be used in scenarios where the expected document is unique based on the search criteria, so IMO a sort doesn't make sense.

Indeed, I forgot to delete that merge request.

PR #3068, opened 2019-06-14
Remove duplicated MetadataWidget from collection view

jcfr:
Since the HierarchyWidget already takes care of instantiating the
MetadataWidget, this commit removes the one explicitly associated
with the collection.

image

### DIFF
--- a/girder/web_client/src/views/widgets/HierarchyWidget.js
+++ b/girder/web_client/src/views/widgets/HierarchyWidget.js
@@ -294,6 +294,19 @@ var HierarchyWidget = View.extend({
             }
         }
 
+        if (this.parentModel.resourceName === 'collection' && this._showMetadata) {
+            if (!this.metadataWidget) {
+                this.metadataWidget = new MetadataWidget({
+                    item: this.parentModel,
+                    parentView: this,
+                    accessLevel: this.parentModel.getAccessLevel()
+                });
+            }
+            this.metadataWidget.setItem(this.parentModel);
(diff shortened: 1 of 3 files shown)


jcfr:
Would appreciate if someone could QA the change, I don't have my devel env. setup.

zachmullen:
I believe my latest commit resolves the issue. @brianhelba @AlmightyYakob PTAL.

AlmightyYakob:
@zachmullen LGTM, my only comment is that we're displaying collection metadata under the class 'g-folder-metadata', but functionally I think it's good.

jcfr:
Thanks for the help working on this !

zachmullen:
@brianhelba @AlmightyYakob could you guys remove or update your previous reviews so this can be merged? Thanks!

girder/web_client/src/views/widgets/HierarchyWidget.js
```diff
@@ -294,6 +294,19 @@ var HierarchyWidget = View.extend({
             }
         }
 
+        if (this.parentModel.resourceName === 'collection' && this._showMetadata) {
+            if (!this.metadataWidget) {
```
jcfr: To be consistent with the folder initialization, should a function `_initCollectionViewSubwidgets` be added ?

zachmullen: I actually don't like the logic in the block above this. It seems to condition display of the metadata widget on the value of `_showItems`, which seems like buggy behavior. Since it's just one widget in the collection case, I prefer to leave it as written.

PR #3070, opened 2019-06-19
Move ESLint out of CMake

brianhelba:
This removes the CMake `add_eslint_test` function, and configures ESLint to be run only via the `npm run lint` script command or via direct invocation (i.e. `npx eslint ...`).

This also adds a `@girder/eslint-config` npm package (based on the to-be-removed `eslint-config-girder` package), and adds instructions for how use this to add ESLint checks for Girder style rules to external plugins.

This pulls in the history of [the `eslint-config-girder` repository](https://github.com/girder/eslint-config-girder), along with the following commits:
* Set up @girder/eslint-config package
* Run ESLint directly through npm
* Fix style issues found by updated ESLint stack
* Don't allow non-imported jQuery usage


manthey:
Can we add an entry to run this via tox?  I'd like to keep an easy facility to run all of the tests before submitting code.  At one point ctest was the way to do this.

zachmullen:
> Can we add an entry to run this via tox?

I appreciate the desire to have one command to rule them all (to approximate the CI process locally), however I think `tox` and `npm` are siblings in our case; making one a parent of the other is messy. IMO the thing that should wrap them is simply a shell command / script. We could put a small bash script to that effect in the `scripts` directory.

manthey:
As I've mentioned elsewhere, my desire is to make it trivial for a non-core developer to run the tests locally.

@brianhelba Can you add documentation for this in developers.rst?

brianhelba:
I've just updated the documentation for this. Honestly, it's still not as good as it could be for new developers, but I think a major problem right now is that running anything via CMake is so complicated that it overwhelms the rest of the testing documentation. As we continue to clean up the infrastructure, I expect the documentation to shrink, which will actually make it far more useful.

Additionally, I'd like to start referencing https://github.com/girder/cookiecutter-girder-plugin directly from our docs. In many cases, particularly for testing which needs to be bootstrapped only one time ever, It's much easier to show by example than attempt to describe things in paragraphs of prose. I'll be making some downstream PRs to to the cookiecutter repo too, then we can incorporate it into the upstream docs by reference.

brianhelba:
> my desire is to make it trivial for a non-core developer to run the tests locally.

I totally agree. To me, the desired procedure is going to be:
```
tox ; npm run lint
```
I oppose trying to wrap these into a script or test runner together. They are totally separate domains and just directly invoke well-known broadly-accepted tooling without magic. Developers are free to locally alias this line if they find it tedious.

I can confidently say that the biggest obstacle to making tests trivial to run is CTest / CMake. As we try to move things off that harness, it's possible that some things will change in breaking ways, but I think it's worth the short-term inconvenience.

zachmullen:
> `tox ; npm run lint`

That command assumes that we only ever run linting through npm. If the goal is to hide the whole process behind one command, we should have a different npm script alias, in case we add other steps besides just lint.

manthey:
I think we are conflating the difficulty of adding tests to cmake and the ease of using it.  Once the tests are setup in cmake, if you can build in cmake, then testing is really just `ctest` and that is all.  I don't think that is hard to *run*.  It has been hard to add and maintain tests, but easy to run.  Here we are moving to greater primary developer convenience (a good thing), but for an external developer who wants to contribute where they don't have to add tests, we are currently making things harder.

brianhelba:
> > `tox ; npm run lint`
> 
> That command assumes that we only ever run linting through npm. If the goal is to hide the whole process behind one command, we should have a different npm script alias, in case we add other steps besides just lint.

The dream is that we can also run the regression tests via npm too, in which case we'd use `npm run test` (callable as `npm test` too). I expect that the `test` script would also call the `lint` script, in addition to invoking the test runner.

For now, since all we can do via npm is linting, that's the one-and-only command we need. I think it's premature to try to set up a `test` script as an alias too. This PR is the first in a series of improvements, so I'd like to be able to make the low-level changes first before trying to worry about meta-commands and wrapper scripts.

PR #3071, opened 2019-06-19
Respect blacklist in python client recursive upload

mgrauer:
blacklist arg wasn't being respected in subfolders, for recursive upload via the python client.


mgrauer:
Fixes #3072.

PR #3079, opened 2019-06-21
Run pug-lint via npm only

brianhelba:
* Run pug-lint via npm only
* Move pug-lint config to a sharable package


brianhelba:
@danlamanna Yup.

PR #3080, opened 2019-06-21
Run stylint via npm only

brianhelba:
* Run stylint via npm only
* Remove CMake harness for Javascript style testing
* Fix lint ignores for nested node_modules


manthey:
Is there a way to have the stylint rules in an npm package, like we've done for the pug and eslint rules?

brianhelba:
It does not seem so. Stylint seems to be pretty inactive.

I'm updating the cookiecutter repo to make it easier to copy-paste a config downstream though.

PR #3081, opened 2019-06-21
Run Flake8 via Tox

brianhelba:
* Run Flake8 via Tox
* Fix newly found Flake8 errors
* Move toxworkdir to default location
* Cleanup Flake8 config


tox.ini
```diff
@@ -1,7 +1,6 @@
 [tox]
-envlist = py{27,35,36},docs,public_names
+envlist = py{27,35,36},lint,docs,public_names
 skip_missing_interpreters = true
-toxworkdir = {toxinidir}/build/test/tox
```
brianhelba: Is everyone cool with this going back to `.tox`? I'm the one that originally set it to `build/test/tox`, as part of a hope that we could keep all build-time artifacts confined to a single directory, but as things have evolved (particularly with how Pip-installable plugins are built), that's probably not practical to fully realize.

I think it now might be best to let Tox go in its default place, since that's going to be the least surprising to other developers and tooling.

zachmullen:
Is this new lint env actually being run in CircleCI?

tox.ini
```diff
@@ -1,7 +1,6 @@
 [tox]
-envlist = py{27,35,36},docs,public_names
+envlist = py{27,35,36},lint,docs,public_names
```
brianhelba: @zachmullen Since it's part of [the default `envlist`](https://tox.readthedocs.io/en/latest/config.html#conf-envlist), simply running `tox` will kick off the `tox -e lint` test too. However, I realize that CI always asks for specific Tox environments, so you are correct, I need to fix this. Oops. :roll_eyes:

brianhelba:
I also changed the order to make the `lint` step run first in the default Tox ordering, since it's probably the most straightforward to address and most likely to fail for developers. We could also make Tox run before CMake in CI, to minimize the turnaround time to when linting errors are reported (again, the most likely source of failure), but I didn't want to mess with CI config too much in this PR (unless people think this optimization is worthwhile).

PR #3082, opened 2019-06-24
Fail on the first failing linter

manthey:
Otherwise, if any but the last linter fails, `npm run lint` will still return success.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
         "stylint": "^1.5.9"
     },
     "scripts": {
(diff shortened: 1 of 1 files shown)


brianhelba:
I approve. The intent was to run all linters even if one fails, but obviously the prior implementation allows a failing exit status to be swallowed.

Thanks for also fixing the minor whitespace issues.

manthey:
Running all of the linters and reporting any failure could be done with a bit of bash magic, e.g., something like `lint1 && { lint2 } || { lint2 && false }`, but I think simplicity is better here.

PR #3086, opened 2019-06-25
Remove some Flake8 ignored rules and add Flake8 checking of names

brianhelba:
* Remove some Flake8 ignored rules and add Flake8 checking of names
* Fix newly flagged style issues


brianhelba:
PTAL

PR #3092, opened 2019-07-02
Add Flake8-bugbear

brianhelba:
Add [Flake8-bugbear](https://github.com/PyCQA/flake8-bugbear), "a plugin for Flake8 finding likely bugs and design problems in your program".

This was suggested by @manthey , and seems to indeed be finding issues that should be changed.

### DIFF
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -403,15 +403,15 @@ class TestFindWithPermissions(object):
         # Test with fields
         for currentUser in (user, admin):
             query, fields = _model._textSearchFilters('names')
-            assert 'name' in _model.findWithPermissions(
-                query, fields=fields, user=currentUser).next()
-            assert 'field1' in _model.findWithPermissions(
-                query, fields=fields, user=currentUser).next()
+            assert 'name' in next(_model.findWithPermissions(
+                query, fields=fields, user=currentUser))
+            assert 'field1' in next(_model.findWithPermissions(
+                query, fields=fields, user=currentUser))
             query, fields = _model._textSearchFilters('names', fields={'name': True})
-            assert 'name' in _model.findWithPermissions(
-                query, fields=fields, user=currentUser).next()
-            assert 'field1' not in _model.findWithPermissions(
-                query, fields=fields, user=currentUser).next()
+            assert 'name' in next(_model.findWithPermissions(
+                query, fields=fields, user=currentUser))
+            assert 'field1' not in next(_model.findWithPermissions(
+                query, fields=fields, user=currentUser))
 
     def testFindWithPermissions(self, db, admin, user, group, FakeAcModel):
         _model = FakeAcModel()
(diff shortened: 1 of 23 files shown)


brianhelba:
@manthey , in https://github.com/girder/girder/blob/5de05f523e472b5eac4ddb1c8601c0f533ee1380/girder/api/filter_logging.py#L19 can we change the attribute `'logging.filtered'` to something else without a period? We'd like to be able to simply reference something like `record.logging_filtered`, but I'm not sure if `logging.filtered` has any other special behavior. Why did the identifier use this period in the first place?

brianhelba:
@zachmullen , at https://github.com/girder/girder/blob/5de05f523e472b5eac4ddb1c8601c0f533ee1380/test/test_logging_config.py#L35 , the rule `Do not use mutable data structures for argument defaults.` is violated. Do you have a suggested refactor?

brianhelba:
@manthey at https://github.com/girder/girder/blob/5de05f523e472b5eac4ddb1c8601c0f533ee1380/tests/mongo_replicaset.py#L351 , the rule `Do not use mutable data structures for argument defaults.` is violated. Do you have a suggested refactor?

manthey:
> @manthey at
> 
> https://github.com/girder/girder/blob/5de05f523e472b5eac4ddb1c8601c0f533ee1380/tests/mongo_replicaset.py#L351
> 
> , the rule `Do not use mutable data structures for argument defaults.` is violated. Do you have a suggested refactor?

In general, I'd change anything where bugbear reports this to `<parameter>=None` and then in the function have `if <parameter> is None: <parameter> = <default>`.

manthey:
> @manthey , in
> 
> https://github.com/girder/girder/blob/5de05f523e472b5eac4ddb1c8601c0f533ee1380/girder/api/filter_logging.py#L19
> 
> can we change the attribute `'logging.filtered'` to something else without a period? We'd like to be able to simply reference something like `record.logging_filtered`, but I'm not sure if `logging.filtered` has any other special behavior. Why did the identifier use this period in the first place?

I have no recollection of why I used a period here; it certainly doesn't matter.  I'd recommend changing this to something starting with an underscore to emphasize that it isn't intended to be part of the API (e.g., `_logging_filter`).

test/test_model.py
```diff
@@ -403,15 +403,15 @@ def generalTest(self, _model, admin, user):
         # Test with fields
         for currentUser in (user, admin):
             query, fields = _model._textSearchFilters('names')
-            assert 'name' in _model.findWithPermissions(
-                query, fields=fields, user=currentUser).next()
-            assert 'field1' in _model.findWithPermissions(
-                query, fields=fields, user=currentUser).next()
+            assert 'name' in next(_model.findWithPermissions(
```
manthey: This is wrong.  `next()` is a mongo cursor call, not a Python iterator built-in.

brianhelba: After discussion with @manthey in-person, we determined that a `pymongo.Cursor` or a `pymongo.CommandCursor` [is a Python iterator](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/cursor.py#L1162), as it implements the Python iterator protocol (`next` / `__next__` and `__iter__`), so we can advance it in a generalized way with `next()`.

This should be good to stay as-is.

PR #3093, opened 2019-07-04
Remove MongoProxy wrapping of PyMongo objects

brianhelba:
The presence of this wrapping has several negative impacts:
* It breaks certain newer features of PyMongo, such as objects being iterators
* It disrupts the ability to do type introspection of PyMongo objects
* It increases the amount and complexity of code at a critical location
* It has no direct testing
* It was authored by a third-party, but is now unmaintained

This wrapping has minimal realistic benefits, since:
* AutoReconnect exceptions typically only happen when there is network disruption; MongoDB is typically deployed on the same machine or on the same LAN as Girder, where network outages are extremely rare
* Naively re-calling the failed method may be incorrect for many operations:
  * Insert operations may have internally succeeded, and should not be retried
  * Cursor iteration may have internally succeeded, so retries will skip results
  * Cursors may no longer be available on a replica set failover, so retrying iteration can still fail
* I've never actually seen the "'AutoReconnecting, try ..." event appear in Girder's logs
  * It is possible that since the MongoProxy hooks up to its own separate log handler by default, logged events are simply not being emitted


brianhelba:
I've just rebased this.

PR #3094, opened 2019-07-04
Remove support for MongoDB sharding

brianhelba:
This support adds significant code complexity and will likely never be needed for any deployment of Girder. Support for MongoDB replica sets remains intact.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -254,7 +254,7 @@ jobs:
             sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5
             echo "deb [ arch=amd64 ] https://repo.mongodb.org/apt/ubuntu trusty/mongodb-org/3.6 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-3.6.list
             sudo apt-get update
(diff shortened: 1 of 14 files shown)


brianhelba:
@girder/developers PTAL. I'd like to propose merging this prior to 3.0.

zachmullen:
Just to make sure, a deployment could still use sharded mongo for the database or for a gridfs assetstore, right? It's just up to them to configure it, Girder no longer does anything special in that case?

If that's the case, I approve of this. If this change makes it impossible to use the GridFS assetstore in a sharded context, that would be a problem and we should discuss other options.

brianhelba:
All of the removed code was to support Girder's bootstraping a very specific thing: MongoDB ranged-based collection sharding of GridFS assetstores. This is ultimately accomplished via just two MongoDB commands: `enableSharding` and `shardCollection`, which need to be issued only one time.

In the future, database admins remain free to [enable ranged-based collection sharding themselves](https://docs.mongodb.com/manual/core/ranged-sharding/#shard-a-collection) via direct configuration of the database.

For new and existing replica sets, Girder's `database.uri` config setting just needs to be set to the `mongos` instance. The Pymongo client [natively supports connecting to sharded clusters](https://api.mongodb.com/python/current/api/pymongo/mongo_client.html).

PR #3096, opened 2019-07-08
Api key duration

dorukozturk:
Adds an option to set a duration for api keys.

### DIFF
--- a/girder/api/v1/api_key.py
+++ b/girder/api/v1/api_key.py
@@ -49,13 +49,14 @@ class ApiKey(Resource):
                'key will last.', required=False)
         .param('active', 'Whether the key is currently active.', required=False,
                dataType='boolean', default=True)
+        .param('apiKeyDuration', 'Max number of days for api key to be used', required=False)
         .errorResponse()
     )
-    def createKey(self, name, scope, tokenDuration, active):
+    def createKey(self, name, scope, tokenDuration, active, apiKeyDuration):
         if Setting().get(SettingKey.API_KEYS):
             return ApiKeyModel().createApiKey(
                 user=self.getCurrentUser(), name=name, scope=scope, days=tokenDuration,
-                active=active)
+                active=active, apiKeyDuration=apiKeyDuration)
         else:
             raise RestException('API key functionality is disabled on this instance.')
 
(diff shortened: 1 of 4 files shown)


zachmullen:
This is looking good, thanks! We actually can just set the API key duration as a system setting rather than allowing clients to pass it in.

brianhelba:
What's the motivation for this setting?

In most auth models, an API key is essentially equivalent to a password, just intended for programmatic usage. You then use that API key as a credential (just like logging in) to obtain an ephemeral token. In this model, having expiring API keys is like having expiring passwords, which is [generally seen as a bad idea](https://www.sans.org/security-awareness-training/blog/time-password-expiration-die) (and the rationale might be transferable, if the hassle of rotating API keys encourages sloppiness).

For Girder, if you actually want to delegate access for a finite time, why not just issue a token yourself, then pass that token to the semi-trusted delegate? Tokens in Girder last for 180 days by default, and I don't believe there's any actual enforcement if you request an individual token for longer than the `COOKIE_LIFETIME` setting.

brianhelba:
Some of these test failures will be fixed by #3097.

zachmullen:
This is being driven by a specific user's security requirements, it's possible we could do this as a plugin if we don't anticipate it being useful outside their environment. The default for this setting would be no expiration, and I certainly don't anticipate many people needing it, but it does fall in line with password expiration policies.

brianhelba:
> This is being driven by a specific user's security requirements,

So issuing and managing tokens as the shareable credential of access would not satisfy this user?

> it's possible we could do this as a plugin if we don't anticipate it being useful outside their environment.

We already have a precedent for adding one-off settings / policies to satisfy particular users 
(e.g. `ENABLE_PASSWORD_LOGIN`, `API_KEYS` , #2978), so I won't object to this going into core too.

At some point, I hope we could design *a priori* a coherent security model and then have the confidence to offer it as-is; the continual bolting-on of these 'features' doesn't increase Girder's objective security and makes me nervous about possible second-order interactions with plugins, etc.

> it does fall in line with password expiration policies.

I'm curious, would we implement these too if a user asked for it?

zachmullen:
> So issuing and managing tokens as the shareable credential of access would not satisfy this user?

Correct, their tokens are set to expire after 24 hours, this is designed to provide a bit longer utility for a token, and a better user interface to use with client apps. There will be additional safeguards in place on their API key usage, e.g. IP-address whitelisting for API key login.

PR #3097, opened 2019-07-09
Fix a test failure caused by upstream flake8-docstring bugs

brianhelba:
The issue is tracked at: https://gitlab.com/pycqa/flake8-docstrings/issues/36


brianhelba:
We can revert this once upstream releases a new version.

PR #3102, opened 2019-07-26
Fix for  _passArg in the describe API

CharlesGueunet:
The `_passArg` function may add arguments with a given name into the  `params` kwargs but it did notcheck the existence of `params` a priori. This PR fix this behavior.

### DIFF
--- a/girder/api/describe.py
+++ b/girder/api/describe.py
@@ -629,6 +629,8 @@ class autoDescribeRoute(describeRoute):  # noqa: class name
             params = {k: v for k, v in six.viewitems(kwargs) if k != 'params'}
             params.update(kwargs.get('params', {}))
 
+            kwargs['params'] = kwargs.get('params', {})
+
             for descParam in self.description.params:
                 # We need either a type or a schema ( for message body )
                 if 'type' not in descParam and 'schema' not in descParam:


zachmullen:
Was something broken before?

CharlesGueunet:
I am creating unit tests for a girder plugin and have a `KeyError:  'params'` launched by this function, by the `pop(name, None)` call.

CharlesGueunet:
Maybe I do not reach this function using the usual way and so the creation of the `params` if not ensured ?
The backtrace of the exception shows I have reached this point from:
```
     self._passArg(fun, kwargs, name, val)
```
in the `wrapped(*args, **kwargs)` function in **describe.py**
which has been triggered by my own function called in the test.
Maybe we can ensure the existence of the `params` keyword in the `wrapped` function instead ?

zachmullen:
Can you post the code that reproduces the error?

CharlesGueunet:
I cannot post the original code so I have created a minimal example with cookiecutter which is hosted [here](https://github.com/CharlesGueunet/GirderTestPlugin).
If you install it and use pytest, you should have the same KeyError than I have :)

zachmullen:
Thanks, I understand now. In our tests, we don't ever call REST handlers directly like that, we always do it through the `server` fixture to simulate an HTTP request, which is why we haven't encountered this before.

CharlesGueunet:
Is there anything else I can do to help with this MR ?

zachmullen:
Prior to this change, the `params` kwarg was optional for functions decorated with `autoDescribeRoute`. I just want to make sure that is still the case after this change, but I have not had time to test it myself.

finetjul:
@zachmullen any news on this one ? Thanks!

CharlesGueunet:
> Just looked at the code. Could you move this change to line 632 instead? That way we only have to do the check once rather than being inside the loop body.

Done

zachmullen:
Thanks @CharlesGueunet 

Could you please merge latest origin/master into this, I'm hoping that fixes CI

zachmullen:
Tests are passing now, however this change tipped that function over the cyclic complexity threshold, meaning we need to either refactor it, or replace the explicit conditional with an internal conditional, e.g.

```python
kwargs['params'] = kwargs.get('params', {})
```

I'm OK with that workaround if you are, but if you prefer to refactor it, feel free.

zachmullen:
(Sorry for all the trouble... this is a very long and difficult road for such a small code change, I realize :)

CharlesGueunet:
> `kwargs['params'] = kwargs.get('params', {})`

This syntax is more  pythonic, let's use it !

zachmullen:
Thanks @CharlesGueunet !

PR #3105, opened 2019-08-02
Fixes #3104. Allow certificate CLI option to work with --api-url

### LINKED ISSUE #3104: girder-client --certificate option doesn't work with --api-url
This comes from the following check in `cli.py`:

```python
    if certificate and ctx.obj.scheme != 'https':
        raise click.BadArgumentUsage(
            'A URI scheme of "https" is required for option "--certificate"')
```

Most of the time we don't pass the scheme, we use `--api-url` instead.

zachmullen:
This is a breaking change due to a change to the cert-related CLI options, so we should get it in before v3. It's also a bug fix.


zachmullen:
@girder/developers @girder/core anyone know why the integration tests are failing?

PR #3106, opened 2019-08-06
Fix PhantomJS when run with newer OpenSSL

brianhelba:
* Fix PhantomJS when run with newer OpenSSL
* Removed unused code from specRunner.js


brianhelba:
@girder/developers PTAL, this fixes a broken `master`.

PR #3107, opened 2019-08-06
Make Python 3.6 the minimum Py3 version and improve CI Dockerfiles

brianhelba:
* Make Python 3.6 the minimum Python 3 version
* Remove deprecated MAINTAINER command from Dockerfiles
* Clean up formatting in test Dockerfiles
* Use a simpler, constant version of CMake in testing
* Use a Python 3 testing Docker image with NodeJS pre-installed
* Install Tox globally in Docker testing images
* Ensure the correct NodeJS version is installed in the testing Docker

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -486,15 +486,6 @@ jobs:
       - checkout:
           path: girder
       - run:
-          name: Create virtual environment (if necessary)
-          command: if [ ! -d girder_env ]; then virtualenv girder_env; fi
-      - run:
-          name: Activate virtual environment
-          command: echo ". $CIRCLE_WORKING_DIRECTORY/girder_env/bin/activate" >> $BASH_ENV
-      - run:
-          name: Install tox
-          command: pip install --upgrade pip tox
-      - run:
           name: Test public symbols
           command: tox -e public_names
           working_directory: girder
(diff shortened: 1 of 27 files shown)


brianhelba:
Rationale:
* Python 3.6 is preinstalled on Ubuntu 18.04 LTS
  * This is the distro targeted by all of Girder's Ansible roles, and most cloud VMs
* [AWS Elastic Beanstalk uses Python 3.6](https://docs.aws.amazon.com/elasticbeanstalk/latest/platforms/platforms-supported.html#platforms-supported.python)
* Python 3.6 contains many of [the most convenient modern Python features](https://docs.python.org/3/whatsnew/3.6.html), like f-strings, guaranteed dict ordering, variable type annotations, etc.
  * Downstream plugins wishing to take advantage of these features will likely target Python 3.6+
  * I plan to move https://github.com/girder/cookiecutter-girder-plugin and some other downstream repos to all target Python 3.6, meaning that significant parts of the Girder organization CI test infrastructure can be simplified (without introducing a strict dependency)

brianhelba:
@girder/developers PTAL.

mgrauer:
@jamiesnape FYI

kotfic:
I will say that in principle I agree with this PR,  but in terms of blast radius I'll have at least one deployment that will need to be modified/updated.

brianhelba:
@kotfic This is just about testing and support. There is no enforcement mechanism to prevent installation to Python 3.5, and we can't actually use any of the Python 3.6-specific features, since Python 2.7 is still supported.

brianhelba:
This was approved and tests passed. Merging with failing tests because of the way DockerHub builds things, but `master` should pass.

brianhelba:
Actually, `master` may fail on the first CI run, but it should start passing again within the next 30 minutes. I'll monitor it.

brianhelba:
Also, apparently the Python 2 CI Docker image has been failing to build for the past month. This PR fixes it.

PR #3108, opened 2019-08-08
Remove obsolete Flake8 packages from requirements-dev.txt

brianhelba:
Flake8 packages are always installed via Tox.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -25,15 +25,11 @@
 # External dependencies
 configparser==3.5.0
 coverage
-flake8
-flake8-blind-except
-flake8-docstrings>=1.3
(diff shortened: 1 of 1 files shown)


zachmullen:
Is that an explicit/documented guarantee from tox?

brianhelba:
I mean the packages in question are installed from this list: https://github.com/girder/girder/blob/master/tox.ini#L18 , as developers should run Flake8 via `tox -e lint`.

Listing them in `requirements-dev.txt` does nothing, and it's not even the correct packages anymore.

zachmullen:
Sorry, I misunderstood what you meant by "via tox". Thanks.

PR #3111, opened 2019-08-08
Fix an error during ZIP download where cursors expire

brianhelba:


zachmullen:
The idea of the original implementation was to avoid loading a potentially very large subtree into memory. Maybe we are not concerned about that, but I know that sometimes these documents can get pretty big when all the metadata is taken into account. I'd be more comfortable with this if we added a projection filter to the queries to only return the fields we need for the download.

brianhelba:
@zachmullen Note that these ZIP download endpoints have an `includeMetadata` flag, so in some cases, the `meta` field will have to be extracted into memory. For core Girder, this is the only thing large enough to bother dropping with a query projection.

However, it is still possible for plugins to tack very large fields into an item, which we would not need for this query.

zachmullen:
Right, for that reason I'd suggest using a whitelist rather than a blacklist projection operator.

zachmullen:
(Also I'd be fine with dropping the includeMetadata functionality altogether...)

brianhelba:
@zachmullen See this update.

girder/models/item.py
```diff
@@ -501,7 +501,11 @@ def fileList(self, doc, user=None, path='', includeMetadata=False,
         metadataFile = 'girder-item-metadata.json'
 
         fileModel = File()
-        for file in self.childFiles(item=doc):
+        # Eagerly evaluate this list, as the MongoDB cursor can time out on long requests
+        # Don't use a "filter" projection here, since returning the full file document is promised
```
brianhelba: Note this.

girder/models/folder.py
```diff
@@ -691,19 +691,31 @@ def fileList(self, doc, user=None, path='', includeMetadata=False,
         if subpath:
             path = os.path.join(path, doc['name'])
         metadataFile = 'girder-folder-metadata.json'
-        for sub in self.childFolders(parentType='folder', parent=doc, user=user):
+
+        # Eagerly evaluate this list, as the MongoDB cursor can time out on long requests
+        childFolders = list(self.childFolders(
+            parentType='folder', parent=doc, user=user,
+            fields={'name': True, 'meta': includeMetadata}
```
zachmullen: I don't think this will work.

> With the exception of the _id field, you may not have both inclusions and exclusions in the same projection document.

girder/models/item.py
zachmullen: No worries, these are the leaf nodes, so they won't be on the stack for long.

girder/models/folder.py
brianhelba: Oops, you're right.

PR #3113, opened 2019-08-09
Handle when upload chunks are submitted as multipart form parts

brianhelba:


zachmullen:
[We officially removed support for this feature](https://girder.readthedocs.io/en/latest/migration-guide.html#multipart-encoded-upload-chunk-support-has-been-removed) as part of 3.0. With that in mind, do you still want to merge this change?

brianhelba:
@zachmullen This PR is just about preventing 500 errors, it doesn't directly allow / disallow the use of multipart-form uploads.

If we want to rigorously prevent multipart-form uploads, then [this line](https://github.com/girder/girder/blob/57b924bcb0cfa8bda5ffa55a0a2f602226098d07/girder/api/v1/file.py#L219) will need to have an additional check added. Due to the way that Girder makes `params` contain query strings or form fields or form parts, I think we'd have to explicitly look for a multipart form and raise a 400 error.

So, I'd suggest merging this, then filing an issue or follow-up PR to explicitly reject multipart forms.

zachmullen:
Thanks for the clarification. Just so I understand, after merging this, attempting a multipart upload will still cause a 500 error, just a different one? Or is this sufficient to make that respond with 4XX?

brianhelba:
@zachmullen Honestly, I have not tested by sending my own (technically unsupported) multi-part upload at an instance with this patch. It's possible that it still fails somewhere else down the line. Since we don't actually support multi-part uploads, this didn't concern me too much at the time :roll_eyes: .

This was just my attempt to patch a utility function to not throw 500 errors in production (where this occurred several times on data.kitware.com).

zachmullen:
It seems to me there's a two-liner that can fix this properly, at the top of the chunk endpoint function:

```python
if cherrypy.request.headers.get('content-type') == 'multipart/form-data':
    raise RestException('Multipart encoding is not supported.')
```

If you want, I can go ahead and make that change and give it a quick test.

zachmullen:
Feel free to revert this if you want, since #3114 should obviate it.

PR #3118, opened 2019-08-12
Harden the logic when a passwordless user needs to login with OAuth

brianhelba:
Also, fix a bug in some legacy support logic: "Google" -> "google"

### DIFF
--- a/plugins/oauth/girder_oauth/__init__.py
+++ b/plugins/oauth/girder_oauth/__init__.py
@@ -17,15 +17,28 @@ def checkOauthUser(event):
     if user.get('oauth'):
         if isinstance(user['oauth'], dict):
             # Handle a legacy format where only 1 provider (Google) was stored
-            prettyProviderNames = 'Google'
+            userProviders = ['google']
         else:
+            userProviders = [val['provider'] for val in user['oauth']]
+
+        # Silently skip non-installed providers
+        availableUserProviders = [
+            userProvider for userProvider in userProviders
+            if userProvider in providers.idMap
+        ]
+
+        if availableUserProviders:
             prettyProviderNames = ', '.join(
-                providers.idMap[val['provider']].getProviderName(external=True)
-                for val in user['oauth']
+                providers.idMap[userProvider].getProviderName(external=True)
+                for userProvider in availableUserProviders
             )
+            helpMessage = 'Please log in with %s, or use the password reset link.' % \
+                          prettyProviderNames
+        else:
+            helpMessage = 'Please use the password reset link.'
+
         raise ValidationException(
-            "You don't have a password. Please log in with %s, or use the "
-            'password reset link.' % prettyProviderNames)
+            "You don't have a password. %s" % helpMessage)
 
 
 class OAuthPlugin(GirderPlugin):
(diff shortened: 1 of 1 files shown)


brianhelba:
@jeffbaumes I believe you actually experienced this as a user on August 9.

plugins/oauth/girder_oauth/__init__.py
```diff
@@ -17,15 +17,28 @@ def checkOauthUser(event):
...
             )
+            helpMessage = 'Please log in with %s, or use the password reset link.' % \
+                          prettyProviderNames
+        else:
+            helpMessage = 'Please use the password reset link.'
+
         raise ValidationException(
-            "You don't have a password. Please log in with %s, or use the "
-            'password reset link.' % prettyProviderNames)
+            "You don't have a password. %s" % helpMessage)
```
jeffbaumes: I believe this should turn into a single-quote string.

jeffbaumes:
Yes, there was no useful error message when I attempted to log in with a password when it should have been Google OAuth. In 2.x Girder it would give me a hardly-helpful but decipherable error that had the word `Google` in it. This PR fix looks to be much better.

plugins/oauth/girder_oauth/__init__.py
brianhelba: No, [`flake8-quotes`](https://github.com/zheller/flake8-quotes) checks this, and it insists that we use double quotes when a single quote is within the string (to minimize escaping) and single quotes otherwise. The `tox -e lint` will always complain now if we get quotes wrong.

jeffbaumes: Got it, missed the contraction there. Carry on.

zachmullen: I use [black](https://black.readthedocs.io/en/stable/) on other projects, but it's hard to adopt into an already large codebase. Something to consider for the next major of Girder.

brianhelba:
@girder/developers PTAL and review.

plugins/oauth/girder_oauth/__init__.py
brianhelba: Black doesn't handle quotes quite in the way we might want. It defaults to requiring double quotes. The only other option is to turn off quote checking entirely.

For this and other types of checks, Flake8 (and plugins) is still useful in a project with Black.

plugins/oauth/girder_oauth/__init__.py
```diff
@@ -17,15 +17,28 @@ def checkOauthUser(event):
...
-            prettyProviderNames = 'Google'
+            userProviders = ['google']
         else:
+            userProviders = [val['provider'] for val in user['oauth']]
+
+        # Silently skip non-installed providers
+        availableUserProviders = [
+            userProvider for userProvider in userProviders
+            if userProvider in providers.idMap
+        ]
```
zachmullen:
```python
    userProviders = {'google'}
else:
    userProviders = {val['provider'] for val in user['oauth']}

availableUserProviders = set(providers.idMap.keys()) & userProviders
```

What do you think? Feels like the order doesn't matter here.

brianhelba: I was operating as though ordering mattered when I wrote this, but you're right, it doesn't. Set comparison is more natural in this case.

brianhelba: Changed.

PR #3120, opened 2019-08-13
Publish npm packages as part of the release process

### LINKED ISSUE #2365: Move packaging for pypi/npm to CI
This is good for producing packages from a pristine-ish environment, and will ultimately be less burdensome if we end up creating packages for each core plugin.

Additionally this will allow us to take advantage of the various pre-release type features mentioned in #974.

brianhelba:
Fixes #2365


scripts/publish.sh
```diff
@@ -0,0 +1,40 @@
...
+GIT_VERSION=$(git describe --tags)
+PUBLISHED_NPM_PACKAGES=(
+  # Published Fontello is used by all builds, and is critical to publish
+  girder/web_client/fontello
+  # These lint configs are used by downstream plugins
+  girder/web_client/eslint-config
+  girder/web_client/pug-lint-config
+  # The raw JS source is used by some downstream 'external builds'
+  girder/web_client/src
+  # These plugins were published to support downstream exteranl builds, and should be kept updated
```
zachmullen: external*

brianhelba: Fixed.

brianhelba:
@girder/developers PTAL, but please let me do the final merge of this.

scripts/publish.sh
```diff
@@ -0,0 +1,40 @@
...
+  girder/web_client/fontello
+  # These lint configs are used by downstream plugins
+  girder/web_client/eslint-config
+  girder/web_client/pug-lint-config
+  # The raw JS source is used by some downstream 'external builds'
+  girder/web_client/src
+  # These plugins were published to support downstream external builds, and should be kept updated
+  plugins/jobs/girder_jobs/web_client
+  plugins/oauth/girder_oauth/web_client
+  plugins/gravatar/girder_gravatar/web_client
```
zachmullen: I would prefer to publish the absolute minimum set of these packages. Are all these strictly necessary?

brianhelba: Yes.

Also, none of these are new. They are all already published on npm.

tox.ini
```diff
@@ -72,7 +72,7 @@ deps =
   setuptools-scm
   twine
 commands =
-  {toxinidir}/scripts/pypi_publish.sh
+  {toxinidir}/scripts/publish.sh
```
brianhelba: @danlamanna brought up feedback that the `npm publish` shouldn't happen within a Tox command, but should live in an `npm run` script instead. It should be easy to move over, but I want to do that before merging this.

brianhelba: I'll do that in a follow-up PR.

PR #3128, opened 2019-09-11
RFC: Add an option to change the sftp server's initial root path

boucaud:
When using the sftp server to download files from a specific folder, girder users have to navigate through various subdirectories before finding the content they wish to download (typically /user/username/Private/folder).

We would like to add an option to simplify this process, and directly expose to users the contents of the folder they are intended to download from.

These proposed changes add a configurable path parameter to the sftp server, that would let plugin developpers change the server's root if they wish. It is probably not the best way to do this, which is why we would like feedback on this.


zachmullen:
This is an interesting idea. If the plan is to use this in a plugin context, is it possible to do via extension rather than modification of the existing class? Or put another way, is there a way to slightly tweak the base class to make this specific extension straightforward, without having to encode the concept of rootPath or :girderUser in the base class?

boucaud:
Hi, I think we're going to try and do this via a plugin. I will keep this updated should any issues arise. Thanks.

boucaud:
After consideration, it is impossible right now to extend `_SftpServerAdapter` without also changing `_SFTPRequestHandler`'s setup method and `SftpServer`'s constructor, which leads to a bit of duplication. 

It would be useful to change `SftpServer` and `_SftpRequestHandler` so that they can take optional classes in their constructors (and thus allow us to use a modified `_SftpServerAdapter` without reimplementing the other classes).

@zachmullen, would it be ok ?

zachmullen:
Sounds great!

girder/api/sftp.py
```diff
@@ -220,7 +228,7 @@ class SftpServer(socketserver.ThreadingTCPServer):
 
     allow_reuse_address = True
 
-    def __init__(self, address, hostKey):
+    def __init__(self, address, hostKey, *args, **kwargs):
```
zachmullen: I think we should either not make this variadic and only add `sftpServerAdapter=None`, or else pass through all variadic args into `socketserver.TCPServer.__init__` and explicitly document that in the docstring.

boucaud: I added the sftpServerAdapter arg.

boucaud:
Does the _processPath change seem fine as well ?

PR #3132, opened 2019-10-14
Update documentation for Ansible role

### LINKED ISSUE #3131: What python version will this run Girder as?
Issue meta-issue: Issues were not enabled on this repo. I had to enable them before I could create this one. Perhaps that will open the floodgate.

My real issue is what Python version will this run Girder as? It seems like from [here](https://github.com/girder/ansible-role-girder/blob/master/tasks/main.yml#L22-L32) it will run as whatever is the system Python3.

Is there a way this would not be true? If you pass in your own value for `girder_virtualenv`?

I think my confusion stems from this comment in the readme for `girder_virtualenv`:

>Path to a Python virtual environment to install Girder in. Will be implicitly created if using Python 3.

Did this comment make more sense before there was a change to use system Python3 as a default? How should it be updated?

Overall I'd like to have the Readme clarify that system Python3 will be used as a default and have this comment be updated with whatever makes sense.

brianhelba:
Fixes #3131

### DIFF
--- a/devops/ansible-role-girder/README.md
+++ b/devops/ansible-role-girder/README.md
@@ -8,9 +8,9 @@ An Ansible role to install [Girder](https://github.com/girder/girder).
 
 Ubuntu 16.04+.
 
-Running Ansible with Python 3 on the target host is recommended.
-[Setting `ansible_python_interpreter: auto`](https://docs.ansible.com/ansible/latest/reference_appendices/interpreter_discovery.html)
-will enable this behavior.
+Using Python 3 as
+[the target host Python interpreter](https://docs.ansible.com/ansible/latest/reference_appendices/interpreter_discovery.html)
+is recommended. Setting `ansible_python_interpreter: auto` will enable this behavior.
 
 ## Role Variables
 
--- a/devops/ansible-role-girder/README.md
+++ b/devops/ansible-role-girder/README.md
@@ -8,7 +8,7 @@ An Ansible role to install [Girder](https://github.com/girder/girder).
 
 Ubuntu 16.04+.
 
-Running Ansible with Python 3 on the target host is reccomended.
+Running Ansible with Python 3 on the target host is recommended.
 [Setting `ansible_python_interpreter: auto`](https://docs.ansible.com/ansible/latest/reference_appendices/interpreter_discovery.html)
 will enable this behavior.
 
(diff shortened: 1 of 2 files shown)


devops/ansible-role-girder/README.md
```diff
@@ -1,28 +1,71 @@
...
-| girder_package          | no       | girder                                     | Package name to install via ``pip``, can be a path.                                                      |
-| girder_pip_extra_args   | no       |                                            | Any extra arguments to pass to ``pip`` when installing Girder.                                           |
-| girder_version          | no       |                                            | PyPI version of Girder to install.                                                                       |
-| girder_virtualenv       | no       | {{ ansible_user_dir }}/.virtualenvs/girder | Path to a Python virtual environment to install Girder in. Will be implicitly created if using Python 3. |
-| girder_web              | no       | true                                       | Whether to build the Girder web client.                                                                  |
+Running Ansible with Python 3 on the target host is reccomended.
```
mgrauer: reccomended => recommended



`Running Ansible with Python 3 on the target host` => this is confusing to me.

Are you saying you recommend to (1) run Ansible with Python 3, (2) run Ansible with Python 3 on the target host with like an Ansible local_action, (3) run your deployment of Girder with Python 3 on the target host, or (4) something else?

mgrauer:
The question I had wanted to answer in issue #3131 is "what version of Python will my Girder be running under?" if I use a given Ansible playbook that includes the Girder role. It was a question posed to me by @zachmullen about a particular instance and I couldn't answer it easily.

This PR does provide docs that answer the question, but it isn't a direct answer, which is fine. e.g., in the case which motivated that issue, the answer is "The system Python3 for the host OS".

devops/ansible-role-girder/README.md
brianhelba: I've fixed some spelling errors and reworded this section of the docs to hopefully be more clear.

brianhelba:
> The question I had wanted to answer in issue #3131 is "what version of Python will my Girder be running under?" if I use a given Ansible playbook that includes the Girder role. It was a question posed to me by @zachmullen about a particular instance and I couldn't answer it easily.

This question can't be authoritatively answered by the role alone, since it depends on the context of its use within playbook. Hopefully these docs make the behavior more clear for how the role will perform within a playbook.

I don't want the role docs to be evasive, but they need to assume a more comprehensive understanding of how Ansible works as a prerequisite.

> This PR does provide docs that answer the question, but it isn't a direct answer, which is fine. e.g., in the case which motivated that issue, the answer is "The system Python3 for the host OS".

@mgrauer I'd suggest expanding the FAQs under https://girder.readthedocs.io/en/stable/provisioning.html#faq if you want to provide a more straightforward answer to this sort of question for future users, particularly those with less Ansible experience.

PR #3136, opened 2019-10-17
Use standard range request response status

### LINKED ISSUE #3135: Correctly handle range requests from video players
> If girder is asked for the whole file (Range 0-), then it responds with a 200.  Reading https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests, we probably should still send a 206.

zachmullen:
Fixes #3135

As written, this still doesn't add multipart range support, but since we haven't encountered that yet in the wild, I'm OK with pushing that to a future topic.

### DIFF
--- a/girder/utility/abstract_assetstore_adapter.py
+++ b/girder/utility/abstract_assetstore_adapter.py
@@ -3,6 +3,7 @@ import os
 import re
 import six
 
+import cherrypy
 from cherrypy._cpreqbody import Part
 
 from girder.api.rest import setResponseHeader, setContentDisposition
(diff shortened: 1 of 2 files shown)


manthey:
@zachmullen It looks like linting is complaining about some existing code; once CI passes, I'll approve.

Now that I've looked at it, I'd love to see multipart range requests result in a 200 rather than a 206 with the first part, but as that is an existing problem, we can ignore it for now.

zachmullen:
I opened an issue to add multipart support.

manthey (approved):

PR #3139, opened 2019-10-22
Make `--mock-db` work in pytest_girder

zachmullen:


brianhelba:
@zachmullen @kotfic Is it possible to ever use the `User()` model in a test if the `db` fixture has not been injected?

zachmullen:
@brianhelba that is possible, and would also not use the mocked DB nor mocked crypto. To address that, we could do something in conftest (or maybe another autouse session fixture) that intentionally breaks database connectivity *unless* you load the `db` fixture, i.e. monkey patch the connection function to raise an exception that instructs the developer that they must use the fixture. What do you think?

brianhelba:
@zachmullen I was originally thinking about whether it would be possible or desirable to use some of the functionality in `User()` without an actual database connection. However, if we decide that should be impossible, I'm fine with the policy, as long as its intentional and explicit.

It didn't occur to me that tests would successfully connect to another database that happened to be running. Since it looks like we are going to disallow model access when `db` is not injected, I agree that an explicit guard like #3140 is desirable.

PR #3140, opened 2019-10-22
Remove foot-shooting capability in pytest

zachmullen:
This forces any pytest case that uses `pytest_girder` and attempts
to connect to Girder's database to use the `db` fixture. Prior to this change, failure to add the `db` fixture to a test that talked to the database could lead to a test that inadvertently modifies a developer's actual Girder database, which is obviously undesirable.

@brianhelba PTAL, this is a fairly straightforward (if hackish) solution to the problem you brought up earlier. This problem wasn't introduced by #3139, but it seems we just discovered it. The test case I added here verifies the behavior we want with respect to requiring the fixture, which should future-proof this solution against changes inside `getDbConnection`.

### DIFF
--- a/pytest_girder/pytest_girder/fixtures.py
+++ b/pytest_girder/pytest_girder/fixtures.py
@@ -135,6 +135,7 @@ def boundServer(db, request):
         with serverContext(plugins, bindPort=True) as server:
             yield server
 
+
 @pytest.fixture
 def smtp(db, server):
     """
@@ -208,5 +209,4 @@ def fsAssetstore(db, request):
         shutil.rmtree(path)
 
 
-__all__ = ('admin', 'db', 'fsAssetstore', 'server', 'boundServer', 'user', 'smtp',
-           '_disableRealDatabaseConnectivity')
+__all__ = ('admin', 'db', 'fsAssetstore', 'server', 'boundServer', 'user', 'smtp')
--- a/pytest_girder/pytest_girder/fixtures.py
+++ b/pytest_girder/pytest_girder/fixtures.py
@@ -1,4 +1,5 @@
 import hashlib
+import mock
 import mongomock
 import os
 import pytest
@@ -13,6 +14,18 @@ def _uid(node):
     Generate a unique name from a pytest request node object.
     """
     return '_'.join((node.module.__name__, node.cls.__name__ if node.cls else '', node.name))
+
+
+@pytest.fixture(scope='session', autouse=True)
+def _disableRealDatabaseConnectivity():
+    from girder.utility.config import getConfig
+
+    class MockDict(dict):
+        def get(self, *args, **kwargs):
+            raise Exception('You must use the "db" fixture in tests that connect to the database.')
+
+    with mock.patch.dict(getConfig(), {'database': MockDict()}):
+        yield
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


pytest_girder/pytest_girder/fixtures.py
```diff
@@ -191,4 +203,5 @@ def fsAssetstore(db, request):
         shutil.rmtree(path)
 
 
-__all__ = ('admin', 'db', 'fsAssetstore', 'server', 'boundServer', 'user', 'smtp')
+__all__ = ('admin', 'db', 'fsAssetstore', 'server', 'boundServer', 'user', 'smtp',
```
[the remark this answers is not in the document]
zachmullen: I can exclude it from `__all__`, then I'd just have to change [this](https://github.com/girder/girder/blob/master/pytest_girder/pytest_girder/plugin.py#L2) back to importing each symbol manually. If the fixture doesn't get imported into that plugin, it doesn't get run, due to pytest magic.  👋

zachmullen:
@brianhelba PTAL -- it's clear we need some solution this, since we were guilty of forgetting this even in our own core test suite (a3a1fdb).

pytest_girder/pytest_girder/fixtures.py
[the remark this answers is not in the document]
brianhelba: I don't have a preference whether you export it from `__all__`, so if you want to change it back, that's fine with me. I just didn't understand.

pytest_girder/pytest_girder/fixtures.py
```diff
@@ -66,6 +79,11 @@ def db(request):
 
     connection.close()
 
+    # Clear connection cache and model singletons
+    _dbClients.clear()
+    for model in model_base._modelSingletons:
+        model.__class__._instance = None
```
manthey: @zachmullen I hadn't realized that the implications of this before it was merged.

If a singleton binds events, those events will still be bound after the singleton is reset.  This makes it so tests involving such singletons will have to unbind events so that they can be rebound to the new singleton.

If we reset the singletons, should we also somehow unbind the events on that singleton?  Or add a convenience fixture for unbinding events with a particular handler name?

zachmullen: Event binding is now idempotent, so doing the same binding multiple times should be fine. Did you run into a case where this actually breaks?

manthey: Yes -- in large_image_annotation tests.  I bind to an event to watch a settings change.  I see this event triggered on an instance of the model, but another instance of the model accessed from the test.  Since we don't allow rebinding to an event with the same same eventName and handlerName combination, I see warnings about that.  Unbinding all of the large_image specific events (by, for instance, walking through the events and unbinding anything that matches the handlerName) between each test avoids the problem.

zachmullen: > Since we don't allow rebinding to an event with the same same eventName and handlerName combination

Ah, I assumed it would override the existing binding, which may be more desirable behavior. Barring that, I can't think of a robust solution for the test fixture that would automagically unbind all events that were bound during model instantiation.

manthey: To work around this, I'm unbinding all handlers with a specific handlerName.  Unfortunately, this requires reaching in to `events._mapping`, which, being a private member, is clearly not intended.  It seems like we should either make doing that an official method (at least in the pytest utilities) or have rebinding an event replace the existing binding (it could still produce a warning).

zachmullen: I'm in favor of having later bindings replace earlier ones and emitting a warning, that seems a better and more expected behavior than ignoring the later binding.

PR #3143, opened 2019-11-05
Remove python-dateutil requirement.

manthey:
python-dateutil is required by botocore.  By also requiring it directly, we can end up with an incompatible version.

See https://github.com/boto/botocore/issues/1872.

We should probably release once this is merged, as the current version won't install properly without overriding either the python-dateutil or botocore version.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -33,6 +33,7 @@ moto[server]>=1.3.7
 pytest>=3.6
 pytest-cov>=2.6
 pytest-xdist
-python-dateutil
+# python-dateutil is pinned because of github.com/boto/botocore/issues/1872
+python-dateutil<=2.8.0
 tox
 virtualenv
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,8 @@ installReqs = [
     'PyYAML',
     'psutil',
     'pyOpenSSL',
-    'python-dateutil',
+    # python-dateutil is pinned because of github.com/boto/botocore/issues/1872
+    'python-dateutil<=2.8.0',
     'pytz',
     'requests',
     "shutilwhich ; python_version < '3'",


zachmullen:
FWIW I don't understand why the package installation fails; it seems to be treating the unpinned specifier as an instruction that we must use the latest version of the package. Is that how it's supposed to interpret unpinned package identifiers?

manthey:
pip whimsically installs listed packages in some order, rather than trying to resolve dependencies first.  It installs the unpinned version of python-dateutil.  Then it installs botocore, which demands a specific version of python-dateutil, but since the installation is being done in the same pip command, pip doesn't check that the versions match (whereas if installed subsequently, it would).  This is a known defect of pip, and something that is avoided by pinning, freezing, or using conda.

manthey:
The ansible ci task is failing.  Any ideas why?

zachmullen:
The error message is not helpful to me. @brianhelba any ideas?

brianhelba:
The error is this:
> TASK [Wait for Girder to be listening] *****************************************
> fatal: [ubuntu18]: FAILED! => {"changed": false, "elapsed": 300, "msg": "Timeout when waiting for localhost:8080"}

It looks like Girder simply isn't starting. I'm re-running CI against master now, so we'll see whether it's a bug in this branch or an infrastructure issue.

brianhelba:
Ok, I was able to reproduce the failure. Since the Ansible tests build Girder from `master`, and `master` will not build, the test fails. We should ignore the failing test in this case, and merge this branch when we are otherwise satisfied. The CI process to release does not depend in the Ansible tests passing.

For the record, to debug issues in the Ansible test environment, you can just run [Molecule sequence commands](https://molecule.readthedocs.io/en/stable/getting-started.html#run-test-sequence-commands) via Tox. In my case, I ran:
```bash
tox -e ansible -- converge
tox -e ansible -- login --host ubuntu18
tox -e ansible -- destroy
```

zachmullen:
Thanks @brianhelba and @manthey

PR #3151, opened 2020-01-06
Require MongoDB>=3.6 and recommend MongoDB==4.2

brianhelba:
Earlier versions of MongoDB have passed their end of life: https://www.mongodb.com/support-policy

MongoDB 3.6 is also the earliest version to support all of the aggregation operators needed by `findWithPermissions`.


brianhelba:
@zachmullen Just rebased this on the latest master, so tests will pass. No changes made.

PR #3152, opened 2020-01-06
Do not catch redundant exception types in Python 3

brianhelba:
As of Python 3.3, IOError is an alias of OSError.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -1401,7 +1401,7 @@ class GirderClient(object):
         try:
             with open(os.path.join(dest, '.girder_metadata'), 'r') as fh:
                 self.localMetadata = json.loads(fh.read())
-        except (IOError, OSError):
+        except (OSError if six.PY3 else (IOError, OSError)):
             print('Local metadata does not exists. Falling back to download.')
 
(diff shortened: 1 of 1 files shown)


clients/python/girder_client/__init__.py
```diff
@@ -1401,7 +1401,7 @@ def loadLocalMetadata(self, dest):
         try:
             with open(os.path.join(dest, '.girder_metadata'), 'r') as fh:
                 self.localMetadata = json.loads(fh.read())
-        except (IOError, OSError):
+        except (OSError if six.PY3 else (IOError, OSError)):
```
zachmullen: Does this change anything in practice?

brianhelba: Nope, the behavior at runtime should be the same.

brianhelba: I made the change in this way (as opposed to `noqa`) with an eye towards eventually dropping Python 2 support. When the conditional is elided under Python 3, it's now catching the correct exception, as opposed to the redundancy remaining there indefinitely and being ignored.

zachmullen: Fair enough.

PR #3153, opened 2020-01-06
Recommend Ubuntu>=18.04 (Bionic)

brianhelba:
Ubuntu 14.04 has passed its End of Standard Support period: https://wiki.ubuntu.com/Releases

Ubuntu 16.04 ships with Python 3.5.x, and eventual Python 3 support in Girder will likely require Python=>3.6. For now, Ubuntu 16.04 will still function, but its use should be discouraged for new installations.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -223,7 +223,8 @@ jobs:
             - girder_env
 
   py2_integrationTests:
-    machine: true
+    machine:
+      image: ubuntu-1604:201903-01
     working_directory: /home/circleci/project  # as $CIRCLE_WORKING_DIRECTORY
 
     steps:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -223,7 +223,8 @@ jobs:
             - girder_env
 
   py2_integrationTests:
-    machine: true
+    machine:
+      image: ubuntu-1604:201903-01
```
brianhelba: Note, this makes the integration CI tests now run under Ubuntu 16.04, instead of 14.04.

CircleCI does not yet offer a machine image with 18.04: https://circleci.com/docs/2.0/configuration-reference/#available-machine-images

PR #3155, opened 2020-01-07
Use MongoDB 4.2 in integration tests

brianhelba:


brianhelba:
I split this out as a minimal separate PR because apparently the `testGridFSReplicaSetAssetstoreUpload` test breaks with MongoDB 4.2.

@manthey Do you have any ideas? I believe you're the original author of a lot of the relevant code.

manthey:
Mongo 4.2 eliminated some command line options.  The test makes a replicaset via mongod, so these options threw errors.  I think it unsporting for a library to remove options on minor version changes.

brianhelba:
@manthey Excellent, thanks! And yes, it looks like those two options were for MMAPv1, so they are now irreverent.

This LGTM now. Can you mark as approved, so we can merge?

PR #3162, opened 2020-01-10
WIP: Add a new ESLint rule, disallowing "var"

brianhelba:


brianhelba:
This depends on #3173

PR #3163, opened 2020-01-10
WIP: Upgrade jQuery

brianhelba:
See "Deprecating positional selectors and the sunset of Sizzle": https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/

### DIFF
--- a/plugins/jobs/girder_jobs/web_client/views/JobListWidget.js
+++ b/plugins/jobs/girder_jobs/web_client/views/JobListWidget.js
@@ -37,7 +37,9 @@ var JobListWidget = View.extend({
             } else {
                 delete this.jobCheckedStates[jobId];
             }
-            this._renderData();
+            // this._renderData recreates the DOM that this event handler is attached to,
+            // so run it after the conclusion of the handler
+            setTimeout(_.bind(this._renderData, this), 0);
         },
         'click input.g-job-checkbox-all': function (e) {
             e.stopPropagation();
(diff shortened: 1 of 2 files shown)


brianhelba:
This breaks the `jobs` plugin.

When a `.g-job-checkbox` [is clicked](https://github.com/girder/girder/blob/eea2d0597719d93f8ea72c62066d1511b3ba88ed/plugins/jobs/girder_jobs/web_client/views/JobListWidget.js#L33), it causes the large portion of the [DOM to re-render](https://github.com/girder/girder/blob/eea2d0597719d93f8ea72c62066d1511b3ba88ed/plugins/jobs/girder_jobs/web_client/views/JobListWidget.js#L246) during the event handler, including the very element that received the event. With the [fixes to event handling in jQuery 3.4.0](https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/), this now causes Javascript errors (` Error: TypeError: undefined is not an object (evaluating 'result.value')`).

Doing this sort of operation was already very fragile and bug-prone, but unfortunately, it looks like it may be some work to untangle, unless any @girder/developers can find an elegant workaround.

plugins/jobs/girder_jobs/web_client/views/JobListWidget.js
```diff
@@ -37,7 +37,9 @@ var JobListWidget = View.extend({
             } else {
                 delete this.jobCheckedStates[jobId];
             }
-            this._renderData();
+            // this._renderData recreates the DOM that this event handler is attached to,
+            // so run it after the conclusion of the handler
+            setTimeout(_.bind(this._renderData, this), 0);
```
zachmullen: Can we avoid using `_.bind` by using an arrow function instead?

brianhelba:
Closing in favor of #3289

PR #3166, opened 2020-01-10
Upgrade web client dependencies

brianhelba:


zachmullen:
It's weird seeing a change like this without changes to a lockfile. I've forgotten how this all works in Girder.

PR #3167, opened 2020-01-10
WIP: Clean up Docker

### LINKED ISSUE #2959: Docker build is broken due to setuptools-scm
#2956 broke our automated docker builds.  Unfortunately, there aren't any clear solutions.  For some upstream discussion see https://github.com/pypa/setuptools_scm/issues/77.

I can think of three potential solutions:
1. Add the entire repo (with `.git`) inside the docker container rather than just selective paths.
2. Prebuild the sdist and copy that to the container.  Use a production install instead of editable.
3. Pregenerate the `girder.egg-info` by doing an editable install on the host and copy it to the container.

Both (2) and (3) require steps before executing `docker build`, so would require abandoning generating docker images on dockerhub.  It would also be less conventional for people coming to girder expecting `docker build .` to "just work".

(1) would bloat the docker image, but otherwise everything should work more or less as it does now.

Any other ideas?

brianhelba:
Fixes #2959

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,10 @@
-**/node_modules
-*.egg-info
-Dockerfile
-*.Dockerfile
+*
+!LICENSE
+!README.rst
+!setup.py
+!girder/
+girder/**/node_modules
+
+# TODO: remove....
+!.git/
+!plugins/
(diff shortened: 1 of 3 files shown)


brianhelba:
I'd like to see if we can pass an environment variable to the package version to the `setup.py` file at install-time, to see if we can fix #2959 .

I'd also like to explore installing some common, non-invasive plugins (quotas, oauth, etc.) within the published Docker image.

PR #3169, opened 2020-01-10
Prevent a Pytest warning

brianhelba:


brianhelba:
This needs testing to ensure it's compatible with CircleCI.

brianhelba:
Confirmed that this works with CircleCI. Failing PyTest tests show:
image
Expanding the output can give the full context.

PR #3179, opened 2020-01-13
Directly use an Ansible Galaxy API key to publish roles

brianhelba:
This fixes a bug, but also should be slightly more secure to use, as the secret token is just for Ansible Galaxy, not GitHub.


brianhelba:
I've tested this manually on a SSH'd CI instance.

PR #3180, opened 2020-01-14
Use "girder.test" as a TLD for all email addresses in testing

brianhelba:
RFC2606 guarantees that ".test" will never be resolvable with DNS. By using this, instead of domains like "example.com", during testing, it ensures that test misconfiguration will not attempt to send real emails to real external email addresses.


danlamanna:
LGTM. Though I see this as being prone to regressions, I don't see a clear path for preventing it.

brianhelba:
@AlmightyYakob @danlamanna Needed to fix a merge conflict. Please mark again.

PR #3185, opened 2020-01-22
Fixes #3184. Ensure core swagger client does not put passwords in URL

### LINKED ISSUE #3184: Password exposure in Logs
When I change the password, the old and new password appear in the girder logs in plain text like this:
```
172.22.0.4 - - [22/Jan/2020:13:58:40] "PUT /api/v1/user/password?old=myOldPassword&new=myNewPassword HTTP/1.1" 200 32 "http://localhost/" "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:71.0) Gecko/20100101 Firefox/71.0"
```
Besides girder, web proxies also often log the url including parameters. Because of that, the url should not contain sensitive information.

Affected routes are:
* `POST /user`
* `PUT /user/{id}/password`
* `PUT /user/password`

Some routes contain a token. This could also be a problem depending on the properties of that token.
* `PUT /user/{id}/verification`
* `GET /user/password/temporary/{id}`

My recommendation is to put sensitive information in the body of the request to avoid it appearing in logs.

zachmullen:


zachmullen:
@CarstenWalther

PR #3188, opened 2020-01-24
Update girder build to remove old node_modules

AlmightyYakob:
This fixes a bug that can occur when re-running girder build after a dependency change, by removing `girder/web_client/node_modules`, `girder/web_client/src/node_modules`, and the `web_client/node_modules` folder of any installed plugins.

### DIFF
--- a/girder/cli/build.py
+++ b/girder/cli/build.py
@@ -54,7 +54,8 @@ def main(dev, mode, watch, watch_plugin, npm, reinstall):
         reinstall = False
 
     staging = _GIRDER_BUILD_ASSETS_PATH
-    _generatePackageJSON(staging, os.path.join(_GIRDER_BUILD_ASSETS_PATH, 'package.json.template'))
+    pluginDependencies = _collectPluginDependencies()
+    _generatePackageJSON(staging, os.path.join(_GIRDER_BUILD_ASSETS_PATH, 'package.json.template'), pluginDependencies)
 
     if not os.path.isdir(os.path.join(staging, 'node_modules')) or reinstall:
         # The autogeneration of package.json breaks how package-lock.json is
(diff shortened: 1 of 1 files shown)


girder/cli/build.py
```diff
@@ -63,6 +63,19 @@ def main(dev, mode, watch, watch_plugin, npm, reinstall):
         npmLockFile = os.path.join(staging, 'package-lock.json')
         if os.path.exists(npmLockFile):
             os.unlink(npmLockFile)
+
+        # Remove any lingering node_modules to ensure clean install
+        plugin_dirs = [path.replace('file:', '') for path in _collectPluginDependencies().values()]
```
brianhelba: Girder style for Python variables uses camelCase.

brianhelba: `_collectPluginDependencies` is expensive to run, since it ultimately iterates over Python entry points. If possible, it'd be nice to only run it once throughout `girder build`, but it's not a big deal.

brianhelba: `_collectPluginDependencies()` returns a dict of `(package, version)`. Technically, the `file:...` string is a version specifier. Consider renaming the the `path` variable accordingly.

More importantly, the [`GirderPlugin.npmPackages`](https://github.com/girder/girder/blob/4d80f4af14e2476e951f52c3cf869b864d2b5e76/girder/plugin.py#L103) method, which populates this dict doesn't *have* to return a value prefixed by `file:`; this could be overridden by plugins to return something different. So, I'd filter to only use entries that are prefixed by `file:`.

girder/cli/build.py
```diff
@@ -63,6 +63,19 @@ def main(dev, mode, watch, watch_plugin, npm, reinstall):
...
+        dirs = [staging, os.path.join(staging, 'src')]
+        dirs.extend(plugin_dirs)
```
zachmullen: FWIW if you want a more functional version, you can use syntactic sugar:

```python
dirs = [staging, os.path.join(staging, 'src')] + plugin_dirs
```

girder/cli/build.py
```diff
@@ -63,6 +63,19 @@ def main(dev, mode, watch, watch_plugin, npm, reinstall):
...
+        node_module_dirs = [os.path.join(d, 'node_modules') for d in dirs]
```
brianhelba: These ~5 lines of list manipulation could probably be condensed into 2-3 without losing readability, but it's certainly subjective.

girder/cli/build.py
AlmightyYakob: Thanks, forgot about that.

girder/cli/build.py
AlmightyYakob: Wouldn't the filtering be redundant though, as if we run `path.replace` on a path without the `file:` it would just remain unchanged?

brianhelba: Yes, but the later stages of the pipeline, particularly `shutil.rmtree` would all be run on something that's not a path (like `^1.4`).

AlmightyYakob: Gotcha, I've added a filter

PR #3192, opened 2020-02-03
Add lockfiles for ancillary npm packages

brianhelba:
These are built and published as normal packages, outside of the @girder/meta-build package.

### DIFF
--- a/girder/web_client/.gitignore
+++ b/girder/web_client/.gitignore
@@ -1,2 +1,2 @@
 /package.json
-package-lock.json
+/package-lock.json
--- a/girder/web_client/eslint-config/.gitignore
+++ b/girder/web_client/eslint-config/.gitignore
@@ -1,3 +1,2 @@
 /dist
 node_modules
-package-lock.json
--- a/girder/web_client/eslint-config/package-lock.json
+++ b/girder/web_client/eslint-config/package-lock.json
@@ -1,0 +1,5 @@
+{
+  "name": "@girder/eslint-config",
+  "version": "0.0.0",
+  "lockfileVersion": 1
+}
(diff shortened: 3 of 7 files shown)


brianhelba:
This should fix the broken build from `@girder/fontello`. It looks like the latest `3.0.7` published version of that package is internally broken. By pinning the build tooling for `@girder/fontello`, we can prevent future build regressions.

To fully fix things, we'll need to release a `girder` `3.0.8`, to allow this change to be incorporated into the next npm-published build.

brianhelba:
@girder/developers PTAL and review. We'll have merge with the tests failing.

zachmullen:
FYI I've just push `v3.0.8` after the merge.

brianhelba:
Since the build was failing, CI never got to the `publish_release` step, which publishes the new `@girder/fontello` package.

I've just manually built and published a `@girder/fontello@v3.0.8` package.

I've also just restarted the CI pipeline for `master` and the latest `v3.0.8` release. We'll know in about 20 minutes if everything works again.

zachmullen:
Thanks @brianhelba 

> Since the build was failing, CI never got to the publish_release step, which publishes the new @girder/fontello package.

Should we consider a way to address this sort of chicken and egg problem, or is it rare enough that manual fallbacks are acceptable?

brianhelba:
This PR itself should prevent the sort of regressions which cause the problem in question. We just needed this one-off manual intervention to get out of the ditch.

brianhelba:
@girder/developers Things are working again. Downstreams should be able to successfully build either `master` or `v3.0.8`.

Some non-critical npm packages didn't get published with `v3.0.8`, but I'll publish a `v3.0.9` soon, which ought to publish completely cleanly.

PR #3193, opened 2020-02-03
Run curl with the --fail flag, to cause scripts to error on failure

brianhelba:


brianhelba:
As anticipated, this has exposed a pre-existing problem with our CI `publish` process, where the `./.circleci/publish_esdoc.sh` script is failing: https://app.circleci.com/jobs/github/girder/girder/25376

I'll look into a fix.

PR #3194, opened 2020-02-04
Add PyPI package names to plugin docs

brianhelba:


brianhelba:
This uncovered #3195 .

PR #3198, opened 2020-02-07
Fix remarkable complaining about linkify.

manthey:
Otherwise, the console always prints:
```
linkify option is removed. Use linkify plugin instead:

import Remarkable from 'remarkable';
import linkify from 'remarkable/linkify';
new Remarkable().use(linkify)
```

### DIFF
--- a/girder/web_client/src/misc.js
+++ b/girder/web_client/src/misc.js
@@ -1,8 +1,11 @@
 import $ from 'jquery';
 import _ from 'underscore';
-import { Remarkable } from 'remarkable';
 
 import { MONTHS } from '@girder/core/constants';
+
+import { Remarkable } from 'remarkable';
+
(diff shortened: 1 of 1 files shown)


girder/web_client/src/misc.js
```diff
@@ -1,9 +1,12 @@
 import $ from 'jquery';
 import _ from 'underscore';
-import { Remarkable } from 'remarkable';
 
 import { MONTHS } from '@girder/core/constants';
 
+import { Remarkable } from 'remarkable';
```
brianhelba: Is there a reason you moved these down? Third party package imports tend to be clustered together at the top, though we don't have a style rule to enforce this.

manthey: My linter complained when I didn't move them down.  I'm not sure why -- it should be using the rules we have specified.

PR #3199, opened 2020-02-07
Copy items before iterating

cjh1:
Copy items before iterating to prevent 'OrderedDict mutated during iteration' errors. Basically we can have a handler being added while an event is fired.

### DIFF
--- a/girder/events.py
+++ b/girder/events.py
@@ -282,6 +282,8 @@ def trigger(eventName, info=None, pre=None, asynchronous=False, daemon=False):
     :type daemon: bool
     """
     e = Event(eventName, info, asynchronous=asynchronous)
+    # We take a copy of the items here using list(...) in case the dict is
+    # modified while we are iterating
     for name, handler in list(six.viewitems(_mapping.get(eventName, {}))):
         if daemon and not asynchronous:
             girder.logprint.warning(
--- a/girder/events.py
+++ b/girder/events.py
@@ -282,7 +282,7 @@ def trigger(eventName, info=None, pre=None, asynchronous=False, daemon=False):
     :type daemon: bool
     """
     e = Event(eventName, info, asynchronous=asynchronous)
-    for name, handler in six.viewitems(_mapping.get(eventName, {})):
+    for name, handler in list(six.viewitems(_mapping.get(eventName, {}))):
         if daemon and not asynchronous:
             girder.logprint.warning(
                 'WARNING: Handler "%s" for event "%s" was triggered on the daemon, but is '


zachmullen:
@cjh1 thanks for the fix -- since it's not obvious, could you add an inline comment saying why `list` is necessary so some future person doesn't get clever and delete it?

cjh1:
@zachmullen Done

cjh1:
@danlamanna Yes, it happened in our openchemistry work.

psavery:
I gave this a try, and it looks like it fixed the bug we were experiencing.

danlamanna:
Does this occur as a result of mutating events after the server has started
serving requests?

cjh1:
Its the result of binding to an event after the server has started yes.

brianhelba:
> Does this occur as a result of mutating events after the server has started serving requests?

If this is the case, then maybe we don't want to try to support this type of behavior with this PR, particularly if there are safer ways of accomplishing the same result.

cjh1:
@brianhelba Where does it say this is not supported?

psavery:
There is API to [unbind events](https://github.com/girder/girder/blob/2b3973780dc64b04a3d0d75791d000841935d0a8/girder/events.py#L226). Wouldn't that usually happen after the server has started? And therefore, the event bindings are mutable?

zachmullen:
I don't see any reason why we should disallow event binding and unbinding while the server is running, but be advised that you may run into odd behaviors (like this) as it's a pattern that is rarely used. Typically all binding is done at load time and then any dynamic request-time aspects of it are handled within the event handler itself.

Could you point to code or explain your rationale for wanting to bind events during requests?

brianhelba:
> I don't see any reason why we should disallow event binding and unbinding while the server is running

Doing this makes a server inherently stateful. This has negative implications on multi-process deployments (each process has a different state and will respond differently to requests). It also has more general negative impacts on runtime predictability: a server is going to perform differently now, based on some internal state (hysteresis); in other words, every restart of the server (in between requests) could change its behavior.

@danlamanna discusses this as a anti-pattern for testing as well: https://github.com/girder/girder/issues/2530#issuecomment-443228256

If a single downstream does this via workarounds, we can't stop them, but I'm not sure if Girder should change to try and support it further.

zachmullen:
I'd like to figure out exactly the use case here to see if we can recommend a better way or if we might actually want to support it. I still don't know if this is a race condition from a request on one thread that is binding to events while another request is triggering them (and somehow releasing the GIL), or if this is synchronous and just a matter of someone binding an event from within an event handler. Those may both be anti-patterns or bad practices but maybe they're legitimate use cases that we should consider.

alesgenova:
The use case is this basic queue plugin that we use in one of the openchemistry projects: https://github.com/OpenChemistry/mongochemserver/tree/master/girder/queues

In a nutshell, it keeps track of the taskflows that a user has submitted for execution. The queue plugin ensures that at each time only a maximum amount of taskflows per user can be running. We needed the to bind to a taskflow update event, so that when one of the currently running tasks is completed, we can pop one of the queued ones.

Do you have alternative approaches that would be working better?

brianhelba:
@cjh1 @psavery I think the issue in your codebase can be distilled to this line: https://github.com/OpenChemistry/mongochemserver/blob/9a2fba0a2a7cb841dd67bec016e83468105bd5b2/girder/queues/queues/models/queue.py#L141 Here, you're binding a new dynamically constructed (partially applied) functions as event handlers for the generic `'cumulus.taskflow.status_update'` event.

Most concerningly is the fact that these functions are never unbound, which is a resource leak. Over the course of the server's lifetime, it will continue to accumulate dozens, hundrends, etc. of event handlers, each of which is invoked (then quickly returns: https://github.com/OpenChemistry/mongochemserver/blob/9a2fba0a2a7cb841dd67bec016e83468105bd5b2/girder/queues/queues/models/queue.py#L254 ) in response to any `'cumulus.taskflow.status_update'` event.

You could fix this issue by unbinding the handler once it finishes, but I think this came up because Girder doesn't provide an obvious answer to the "where do we put state for async events" question. Answers are:
1. In the event handler, as partially-applied arguments
2. In the `event` object itself
3. In the database

For the reasons in my last post (about mutating server state after start-up), I think it's generally dangerous to do 1; otherwise, that wouldn't intrinsically be a bad approach. Practically though, I think that 2 or 3 are the better solutions. Perhaps you could do 2 and pass `queue` and `user` all the way through to where the event is triggered (so they could be attached to the `event.info` object).

More likely, 3 is best, and you simply store `queue` and `user` (or the necessary pre-requisites to re-fetch / re-compute) them in the database, associated with the `taskflow_id`. When a single (bound at start-up) event handler for `'cumulus.taskflow.status_update'` runs, it can then receive `taskflow_id` in the event, fetch and compute `queue` and `user`, then dispatch to the normal per-taskflow handler behavior.

zachmullen:
I agree that option 3 is the best.

brianhelba:
> There is API to [unbind events](https://github.com/girder/girder/blob/2b3973780dc64b04a3d0d75791d000841935d0a8/girder/events.py#L226). Wouldn't that usually happen after the server has started? And therefore, the event bindings are mutable?

Unbinding events can still be legitimately done during server start-up, as plugins could unbind listeners from Girder's core or other plugins.

However, if it can not be done during start-up, then by definition, it depends on the inputs that the server has received over its running lifetime, and therefore makes the server's code stateful. So, we want to disallow all `bind` / `unbind` after start-up.

zachmullen:
I'm going to close this PR, but some sort of follow-up seems warranted. At the least it seems we should document that event bindings should only be expressed statically. Not sure how exactly to phrase that, but that's the practice we should recommend. We could maybe try to enforce that practice in code somehow, like disabling modifying the event system after the server loop has started, however that seems like it could throw a wrench into testing and might not be worth the maintenance burden.

PR #3202, opened 2020-02-12
Invalidate the settings cache for a key before saving it.

### LINKED ISSUE #3201: Settings cache leads to unexpected behavior
If you've enable the dogpile setting cache and have an event listener on `model.setting.save.after`, the event will get the old value of the setting, not the new value.

Specifically, if you configure the cache via the ini file like:
```
[cache]
enabled: True
cache.global.backend = "dogpile.cache.memory"
```
and, in some code, bind to the settings change:
```
events.bind('model.setting.save.after', 'cool.plugin', onSettingChange)
```
Then in the function `onSettingChange`, `Setting().get(<setting name>)`, will get the old value.  The event has the new value.

Within the setting model, this could be prevented by invalidating the cache before the save (e.g., change:
```
        setting = self.save(setting)
        self._get.set(setting, self, key)
```
to:
```
        self._get.invalidate(self, key)
        setting = self.save(setting)
        self._get.set(setting, self, key)
```
though there will still be race conditions if multiple threads are changing the setting concurrently.

manthey:
In the current behavior, if the setting cache is enabled and a setting is changed, then an event listener bound to `model.setting.save.after` that gets the setting will obtain the old value, not the new value.  The event has the new value in it, but fetching from the setting gets the old one.

This change invalidates the cache before the save.  We can't set it before save, since that would bypass validation.

There is still the potential for a race condition where if two different threads are setting a key nearly simultaneously, then the wrong one could end up in the cache (that is, not the same one that ends up in the database).  This change is still a strict improvement of the current behavior.

Fixes #3201.

### DIFF
--- a/girder/models/setting.py
+++ b/girder/models/setting.py
@@ -118,6 +118,10 @@ class Setting(Model):
         else:
             setting['value'] = value
 
+        # Invlidate the cache so that events that listen for a change in the
+        # setting will be get the right value from a get.  We don't set it here
+        # as it could fail to validate and save.
+        self._get.invalidate(self, key)
         setting = self.save(setting)
 
         self._get.set(setting, self, key)
--- a/girder/models/setting.py
+++ b/girder/models/setting.py
@@ -118,7 +118,7 @@ class Setting(Model):
         else:
             setting['value'] = value
 
-        # Invlidate the cache so that events that listen for a change in the
+        # Invalidate the cache so that events that listen for a change in the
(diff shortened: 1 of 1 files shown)


girder/models/setting.py
```diff
@@ -118,6 +118,10 @@ def set(self, key, value):
         else:
             setting['value'] = value
 
+        # Invlidate the cache so that events that listen for a change in the
```
zachmullen: Invalidate* otherwise LGTM

PR #3204, opened 2020-02-14
Guard when copying from items and folders with no 'meta' field

brianhelba:


girder/models/folder.py
```diff
@@ -801,7 +801,7 @@ def copyFolderComponents(self, srcFolder, newFolder, creator, progress,
 
         # copy metadata and other extension values
         updated = False
-        if srcFolder['meta']:
+        if srcFolder.get('meta'):
```
manthey: Should this be `if 'meta' in srcFolder:` so that empty meta dictionaries are copied?

brianhelba: I specifically chose this way. I see no reason to copy an empty field, but maybe I'm missing an important case.

manthey: I don't know of any use where having an empty meta dictionary would be treated distinctly from not having a meta field.

mgrauer:
I thought this PR made it so items or folders without a `meta` field would have the field added when the document is loaded. If that is the case, why are you encountering items/folders lacking `meta`?

https://github.com/girder/girder/pull/3041

zachmullen:
@mgrauer possibly a legacy document.

girder/models/folder.py
zachmullen: I agree with @manthey , given that this is intended to copy the resource, it should copy it as faithfully as possible, by preserving all keys even if their values are falsy.

mgrauer:
I thought that PR added `meta` to legacy docs, but I may not be understanding the current PR or that other one fully.

girder/models/folder.py
brianhelba: In that case, then I should also remove the `updated` flag, since it will always get set to `True` for non-legacy `srcFolder` documents (as they always have an existent `meta` field, which will be faithfully copied).

brianhelba:
@zachmullen @manthey PTAL.

PR #3207, opened 2020-02-18
Use path relative to plugin for setuptools-scm

zachmullen:
Prior to this change, setuptools-scm was looking for the repository root relative to the working dir, not the plugin dir. This was causing downstream failures when trying to install these plugins from a Girder git repository rather than pypi.


zachmullen:
I have no idea if this works...

zachmullen:
Spoiler alert: it doesn't

Issue #3208, opened 2020-03-02
Can girder work with another OAuth2 provider that isn't Google?


xdoan:
Synapse is an OAuth2 [server](https://docs.synapse.org/articles/using_synapse_as_an_oauth_server.html) and I was wondering if it is possible to use it as an OAuth provider instead of Google as shown on the [docs](https://girder.readthedocs.io/en/stable/plugins.html#oauth2-login)

mgrauer:
Girder currently works with multiple different OAuth2 providers, including Google, Globus, Linkedin, Box, BitBucket, GitHub. 

It isn't too much code to add a new OAuth2 compliant provider, see these [examples](https://github.com/girder/girder/tree/master/plugins/oauth/girder_oauth/providers).

PR #3211, opened 2020-03-14
Add Synapse as a OAuth (OIDC) identity provider

brucehoff:
https://sagebionetworks.jira.com/browse/IT-842

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,9 +23,11 @@ COPY . /girder/
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
-# TODO: Do we want to create editable installs of plugins as well?  We
-# will need a plugin only requirements file for this.
 RUN pip install --upgrade --upgrade-strategy eager --editable .
+
+ARG plugins
+RUN for plugin in $plugins; do pip install --upgrade --upgrade-strategy eager --editable ./plugins/$plugin; done
+
 RUN girder build
 
 ENTRYPOINT ["girder", "serve"]
--- a/plugins/oauth/girder_oauth/providers/synapse.py
+++ b/plugins/oauth/girder_oauth/providers/synapse.py
@@ -34,8 +34,8 @@ class Synapse(ProviderBase):
 
         callbackUrl = '/'.join((getApiUrl(), 'oauth', 'synapse', 'callback'))
 
-        #OIDC claims. The complete list is here: 
-        #https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/oauth/OIDCClaimName.html
+        # OIDC claims. The complete list is here:
+        # https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/oauth/OIDCClaimName.html
         claims = {
             'email': None,
             'given_name': None,
@@ -64,8 +64,8 @@ class Synapse(ProviderBase):
         }
         basicAuth = ':'.join((str(self.clientId), str(self.clientSecret)))
         auth = base64.b64encode(basicAuth.encode('utf8'))
-        resp = self._getJson(method='POST', url=self._TOKEN_URL, 
-            data=params, headers={'Authorization': 'Basic %s' % auth.decode()})
+        resp = self._getJson(method='POST', url=self._TOKEN_URL,
+                             data=params, headers={'Authorization': 'Basic %s' % auth.decode()})
         return resp
 
     def getUser(self, token):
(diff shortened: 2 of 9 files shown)


Dockerfile
```diff
@@ -26,6 +26,7 @@ ENV LANG=C.UTF-8
 # TODO: Do we want to create editable installs of plugins as well?  We
 # will need a plugin only requirements file for this.
 RUN pip install --upgrade --upgrade-strategy eager --editable .
+RUN pip install --upgrade --upgrade-strategy eager --editable ./plugins/oauth
```
zachmullen: This should probably be done in a downstream image rather than here, since none of the other plugins are getting installed in this image.

plugins/oauth/girder_oauth/providers/synapse.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def getToken(self, code):
+        params = {
+            'grant_type': 'authorization_code',
+            'code': code,
+            'redirect_uri': self.redirectUri
+        }
+        basicAuth=':'.join((str(self.clientId), str(self.clientSecret)))
+        auth = base64.b64encode(basicAuth.encode('utf8'))
+        resp = self._getJson(method='POST', url=self._TOKEN_URL, data=params, headers={'Authorization': 'Basic %s' % auth.decode()})
```
zachmullen: `auth.decode()` should just be `auth`, right?

zachmullen:
If this is using OIDC rather than regular OAuth, we should just implement it in its own OIDC plugin. The virtue of that is that we won't need to statically encode each provider.

I've actually written a working OIDC plugin for Girder, but it's subject to a public release process. I'll inquire about that with the relevant stakeholders.

plugins/oauth/girder_oauth/providers/synapse.py
brucehoff: No.  See https://docs.python.org/3/library/stdtypes.html#bytes.decode

Dockerfile
brucehoff: I will update the PR with a parametrized version of the Dockerfile, allowing users to build with whichever plugins they choose (none by default).

brianhelba: I'd like to see what this sort of parameterization looks like, but for now, I think we should keep Girder's primary Docker image free of plugins. Downstreams can easily inherit from this image and install additional plugins if they wish.

brucehoff: > Downstreams can easily inherit from this image and install additional plugins if they wish.

Looking at the original [Dockerfile](https://github.com/girder/girder/blob/master/Dockerfile) it looks like one has to install plugins (pip install ...) before calling 'girder build'.  If the plugins could be added afterwards, that's great:  We could have a Dockerfile starting with 'FROM girder/girder' which adds plugins.  But currently it's not clear how that would work.  Can you show how to do that?

brianhelba: Indeed, downstreams would have something along the lines of:
```
FROM girder/girder
RUN pip install ./plugins/oauth \
  && girder build
```
I agree we should have better support and documentation for this use case though.

brucehoff:
> If this is using OIDC rather than regular OAuth, we should just implement it in its own OIDC plugin. The virtue of that is that we won't need to statically encode each provider.
> 
> I've actually written a working OIDC plugin for Girder, but it's subject to a public release process. I'll inquire about that with the relevant stakeholders.

Note that the Globus and Google auth providers in the current code base leverage the OIDC extension to OAuth, just as this PR does.

This PR was written following the guidance given in https://github.com/Sage-Bionetworks/HTAN-coordination/issues/95.  The comment above suggests that the repo' owners are _not_ receptive to this PR, even though it follows the existing code organization.  Before we respond to the other comments on this PR and make further refinements, can we find out whether the repo' owners are actually receptive to this PR, or will ultimately just close it without merging?

zachmullen:
I'm not able to follow the link you provided, is that to a private repository?

Regardless, I would suggest we stop moving forward with this PR as it stands, with a new provider in the OAuth plugin. The path forward for providers who support the OIDC standard is to instead use a one-size-fits-all OIDC plugin that can be configured to support any OIDC-compliant provider, without the need for special static support for each provider. Eventually such a plugin could supersede this one as the popular providers adopt OIDC.

xdoan:
@zachmullen The guidance came from the answer to this issue https://github.com/girder/girder/issues/3208

zachmullen:
Thanks for opening #3217 , which captures the intended path toward supporting this use case. Contributions to that end are welcome; otherwise we may have time ourselves to add it in the coming months.

PR #3215, opened 2020-03-18
Allow user quotas to be set to 0

dchiquit:
Allow user quotas to be set to non-negative numbers, instead of just
positive numbers. Previously 0, as a falsy value, was assumed to be the
same as null, which means "no quota".

### DIFF
--- a/plugins/user_quota/girder_user_quota/quota.py
+++ b/plugins/user_quota/girder_user_quota/quota.py
@@ -35,7 +35,7 @@ def ValidateSizeQuota(value):
               and a recommended error message or None.
     :rtype: (int or None, str or None)
     """
-    if value is None or value == '' or value == 0:
+    if value is None or value == '':
         return None, None
     error = False
     try:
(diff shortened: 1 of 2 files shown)


plugins/user_quota/girder_user_quota/quota.py
```diff
@@ -353,7 +351,7 @@ def _checkUploadSize(self, upload):
         if resource is None:
             return None
         fileSizeQuota = self._getFileSizeQuota(model, resource)
-        if not fileSizeQuota:
+        if fileSizeQuota == None:
```
brianhelba: Should be `is None`

zachmullen: Hoping our linter would catch that :)

danlamanna: He just did!

zachmullen: 😢

PR #3216, opened 2020-03-19
Fix rendering 0 user quotas in the dashboard

dchiquit:
Fix the dashboard rendering and saving user quotas of 0. I neglected to
check that the UI worked correctly when I changed the user-quota plugin
to allow user quotas to be set to 0.

### DIFF
--- a/plugins/user_quota/girder_user_quota/web_client/templates/configView.pug
+++ b/plugins/user_quota/girder_user_quota/web_client/templates/configView.pug
@@ -13,7 +13,7 @@ form#g-user-quota-form(role="form")
         | quota is explicitly set.
       input.input-sm.g-sizeValue.form-control(
           id=`g-user-quota-${resource.model}-size-value`,
-          type="text", value=resource.sizeValue || '',
+          type="text", value=resource.sizeValue,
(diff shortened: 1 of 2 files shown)


plugins/user_quota/girder_user_quota/web_client/templates/configView.pug
```diff
@@ -13,7 +13,7 @@ form#g-user-quota-form(role="form")
         | quota is explicitly set.
       input.input-sm.g-sizeValue.form-control(
           id=`g-user-quota-${resource.model}-size-value`,
-          type="text", value=resource.sizeValue || '',
+          type="text", value=resource.sizeValue || resource.sizeValue === 0 ? resource.sizeValue : '',
```
brianhelba: From [how this is rendered](https://github.com/girder/girder/blob/ae9b34ae282a10c4c0b7a8e6cd78de2dbb8d33a0/plugins/user_quota/girder_user_quota/web_client/views/ConfigView.js#L52), I think we can assume it's always either a `Number` or `''`, so all this defensive logic could probably go away.

However, if you think the safety value outweighs the added complexity, I'm fine with keeping it.

PR #3219, opened 2020-03-19
Fix a misleading exception in virtual folders

manthey:
When asking for a folder that doesn't exist, this threw an exception while checking if the folder was a virtual folder.  Rather, it should just return and let the parent method handle the issue.

This specifically occurred when trying to find items in a folder that didn't exist via a rest call (item?folderId=<non existant id>).

### DIFF
--- a/plugins/virtual_folders/girder_virtual_folders/__init__.py
+++ b/plugins/virtual_folders/girder_virtual_folders/__init__.py
@@ -83,13 +83,13 @@ def _folderUpdate(self, event):
 def _virtualChildItems(self, event):
     params = event.info['params']
 
-    if 'folderId' not in params:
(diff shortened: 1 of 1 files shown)


zachmullen:
What's the context here? Just wondering if returning silently is the right thing to do, or should it be loading with `exc=True`?

manthey:
Specifically, without the virtual folders plugin, make a query like `item?folderId=<nonexistent id>` returns `"message": "No such folder: 123456789012345678901234"`.  With the plugin it throws an exception.

With this change, it responds the same way as without the plugin.

While poking around just now, I noticed that if you passed `folderId=` without any value, it also behaved differently, so I've pushed a small update.

Issue #3158, opened 2020-01-09
Python 3.8 support


manthey:
It looks like the _hash_state.py module needs to be updated to allow Girder to work with Python 3.8.  I don't know if any other changes are necessary.

zachmullen:
https://pypi.org/project/hashstate/ ?

zachmullen:
For the record I haven't tried building that against cpython 3.8. If it builds I'd like to add another test to verify that the outputs match the normal `hashlib` outputs, and then I'd feel good about using it in Girder.

manthey:
Superficially, even with hashstate built for Python 3.8, this fails when it reaches https://github.com/girder/girder/blob/master/girder/utility/_hash_state.py#L61.  As a crude test, I changed `hashPointer[3]` to `hashPointer[2]`, since the desired reference seemed to now be in a different position and the tox tests pass (I didn't try any other tests, since those include test_hash_state).  How it was determined that we needed `hashPointer[3]` and why it has changed in Python 3.8?

zachmullen:
We're reaching into deeply nested internal C structs that are subject to change at a whim. This is probably a good time to switch to my C extension as a dependency, which would at least find the problem statically rather than at runtime.

manthey:
The relevant structure is in CPython's _hashopenssl.c code.  This PR https://github.com/python/cpython/pull/16023 changed it, so the `hashstate` module probably won't work on Python 3.8.

zachmullen:
Sure, I anticipate some `#define`s are necessary, but it's nonetheless a good idea for us to move from detecting the problem at build time of our bundle rather than having our downstreams hit it at runtime.

Xarthisius:
> Superficially, even with hashstate built for Python 3.8, this fails when it reaches https://github.com/girder/girder/blob/master/girder/utility/_hash_state.py#L61. As a crude test, I changed `hashPointer[3]` to `hashPointer[2]`, since the desired reference seemed to now be in a different position and the tox tests pass (I didn't try any other tests, since those include test_hash_state). How it was determined that we needed `hashPointer[3]` and why it has changed in Python 3.8?

Just spent an hour trying to [figure it out](https://github.com/Xarthisius/girder/commit/fce00c9a434d18296c2cf20ad29986ec9d67aff6) and came here to file an issue. Then I found his report... :-)

manthey:
[...]

PR #3222, opened 2020-03-25
Support Python 3.8.

manthey:
This fixes the hash state function to work with Python 3.7 and adds Python3.8 to the classifiers in all setup.py files.


manthey:
Closes #3158.

PR #3225, opened 2020-03-26
Fix user quotas of 0 rendering as `Unlimited` in the user view

dchiquit:
User quotas of 0 are being rendered in the `User > Actions > Quota and
assetstore policies` modal as `Unlimited` instead of `0`. This is a
simple edge case where `!quota` is true both when `quota` is `null` and
`0`. Change those expressions to `quota === null`.


zachmullen:
Looks like the web client test was depending on the old behavior, so it'll need to be updated.

plugins/user_quota/girder_user_quota/web_client/views/QuotaPoliciesWidget.js
```diff
@@ -137,7 +137,7 @@ var QuotaPoliciesWidget = View.extend({
         sizeInfo = sizeToValueAndUnits(
             this.model.get('quotaPolicy').fileSizeQuota);
         defaultQuota = this.model.get('defaultQuota');
-        if (!defaultQuota) {
+        if (defaultQuota === null) {
             defaultQuotaString = 'Unlimited';
         } else {
             defaultQuotaString = formatSize(defaultQuota);
```
zachmullen: The test failure message `ERROR: TypeError: undefined is not an object (evaluating 'sizeVal.toFixed')` makes me think `defaultQuota` might be `undefined` in the test.

zachmullen:
Could you squash these before merging please

PR #3226, opened 2020-03-31
Drop python 2 support

zachmullen:
See https://discourse.girder.org/t/dropping-python-2-support/317

Before merging, we need to push a release tag that will become our final release that supports py2.


.circleci/config.yml
```diff
@@ -2,18 +2,6 @@
 version: 2.1
 
 executors:
-  py2-mongo:
-    docker:
-      # This image uses the oldest version of many dependencies
-      - image: girder/girder_test:latest-py2
-      # Use the oldest supported MongoDB
-      - image: circleci/mongo:3.6-ram
```
zachmullen: We happened to be leveraging the py2 environment to also test our minimum supported mongodb version. We should come up with an alternative approach. ping @brianhelba

zachmullen:
@girder/core @girder/developers I'm removing the WIP tag so I can get some eyes on this. It's good to go AFAIC, but there's still the issue of minimum mongo version. We can offload that to a follow-up PR if people want.

danlamanna:
Should Python2 support be dropped from the built in plugins?

I did some more miscellaneous fixes on this branch if you want to integrate it: https://github.com/girder/girder/compare/drop-py2-support...more-py2-removals.

zachmullen:
Yes we should remove it from the plugins also, I'll add a commit for that.

zachmullen:
We should also update the plugin cookiecutter template to remove the python 2 classifiers.

brianhelba:
I pushed a commit updating a few more places in the docs.

.circleci/config.yml
brianhelba: Fixed this by always running against the oldest MongoDB. It's far more likely to see breakages due application code using unsupported new features than it is to have the MongoDB API break on a new release (so long as we continue to track supported releases as the oldest acceptable MongoDB version).

brianhelba:
Pushed one more commit to simplify the `girder_test` Docker image. Give it ~5 minutes for things to catch up, then CI should be working. I've updated the [Docker Hub configuration](https://hub.docker.com/repository/docker/girder/girder_test) accordingly.

zachmullen:
@brianhelba your changes LGTM. Anything else you want to add?

brianhelba:
@zachmullen I don't think so. I have some follow-up planned, but that's easier to review in another branch.

zachmullen:
Thanks. I'm going to tag a release and merge this afterward.

PR #3229, opened 2020-04-01
Ensure that login policy is enforced for OAuth users. Fixes #3228

### LINKED ISSUE #3228: OAuth does not respect 'admin approval' policy
When _Registration Policy_ is set to _Admin approval required_ it's still possible to log in to girder using OAuth and get a fully working account, despite it being flagged with status `pending`.

Xarthisius:
This PR ensures that during the OAuth flow we check whether user `canLogin` before returning `girderToken`. I marked it as WIP because I haven't been able to set up testing env locally for girder 3.x  yet and I'm sure it will break all the things. I will add tests as soon as I get it going. Nevertheless it's ready for review/comments.

### DIFF
--- a/girder/models/user.py
+++ b/girder/models/user.py
@@ -204,17 +204,8 @@ class User(AccessControlledModel):
         elif isinstance(otpToken, str):
             raise AccessException('The user has not enabled one-time passwords.')
 
-        # This has the same behavior as User.canLogin, but returns more
-        # detailed error messages
-        if user.get('status', 'enabled') == 'disabled':
-            raise AccessException('Account is disabled.', extra='disabled')
-
-        if self.emailVerificationRequired(user):
-            raise AccessException(
-                'Email verification required.', extra='emailVerification')
-
-        if self.adminApprovalRequired(user):
-            raise AccessException('Account approval required.', extra='accountApproval')
+        # Verify the user account is enabled and auth policies are fulfilled
+        self.verifyLogin(user)
 
         return user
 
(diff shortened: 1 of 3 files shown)


girder/models/user.py
```diff
@@ -425,18 +415,27 @@ def createUser(self, login, password, firstName, lastName, email,
 
         return user
 
-    def canLogin(self, user):
+    def canLogin(self, user, exc=False):
```
brianhelba, line 418: IMO, the pattern of dynamically changing the return type / exception throwing based on arguments makes functions more difficult to understand and maintain.

I think a better pattern would be for this function to return a `(bool, str)`, optionally indicating a reason why login failed, but that would be a breaking change.

Perhaps an alternative is to implement a new `User` method with the desired behavior, then deprecate the old one.

zachmullen: I agree with @brianhelba 

Union types or tuple return types are more familiar patterns.

Xarthisius: Third option is to duplicate the code that raises the exceptions in oauth plugin and don't even touch the girder core...

zachmullen:
It seems to me that all we need here is to make `verifyLogin` have no return, and either raise an exception or not. Then `canLogin` can remain a simple `bool` return type. `canLogin` can still wrap `verifyLogin` but would just be a try/catch around it. This will allow users to choose the behavior they want -- if they want an exception with a message, they can use `verifyLogin` and if they want a boolean they can use `canLogin`. Then we don't have to deprecate anything either.

Does that sound good?

zachmullen (approved):
Thanks, LGTM!

PR #3231, opened 2020-04-02
Clean up CircleCI config naming and formatting

brianhelba:
This should not have any behavioral effects.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,10 +19,8 @@ executors:
       image: ubuntu-1604:201903-01
     working_directory: /home/circleci/project
 
-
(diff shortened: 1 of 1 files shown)


zachmullen:
Is the `-` preferable to `_`? The [canonical sample](https://circleci.com/docs/2.0/sample-config/) seems to use both. I don't have a preference either way.

brianhelba:
@zachmullen I indeed saw both, and I have the vague impression that I saw more instances of `-` than `_` in the CircleCI docs generally.

If we prefer using underscores, it would be trivial to change. I have no preference.

PR #3232, opened 2020-04-02
Remove use of six.PY3

brianhelba:


pytest_girder/pytest_girder/utils.py
```diff
@@ -22,16 +22,6 @@
 class MockSmtpServer(smtpd.SMTPServer):
     mailQueue = queue.Queue()
 
-    def __init__(self, localaddr, remoteaddr, decode_data=False):
-        kwargs = {}
-        if six.PY3:
-            # Python 3.5+ prints a warning if 'decode_data' isn't explicitly
```
brianhelba, line 28: This warning is only on exactly Python 3.5.

For 3.6+, there is no longer a need to wrap `__init__`.

PR #3233, opened 2020-04-02
Replace "six" type aliases with actual types

brianhelba:
This best reviewed by commits. The "Replace ..." commits were generated by globally applying regexes.

### DIFF
--- a/girder/__init__.py
+++ b/girder/__init__.py
@@ -104,7 +104,7 @@ class StreamToLogger(object):
                     and (callable(getattr(stream, key, None))
                          or isinstance(
                              getattr(stream, key, None),
-                             (six.binary_type, str, six.integer_types, bool)))):
+                             (bytes, str, six.integer_types, bool)))):
                 setattr(self, key, getattr(stream, key))
 
     def write(self, buf):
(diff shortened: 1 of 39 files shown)


girder/utility/system.py
```diff
@@ -24,7 +24,7 @@ def _objectToDict(obj):
         for key in dir(obj)
         if not key.startswith('_') and isinstance(
             getattr(obj, key),
-            tuple([float, tuple] + list(str) + list(int)))
```
zachmullen: What on earth was going on here...

girder/__init__.py
```diff
@@ -104,7 +104,7 @@ def __init__(self, stream, logger, level):
                     and (callable(getattr(stream, key, None))
                          or isinstance(
                              getattr(stream, key, None),
-                             (six.binary_type, six.string_types, six.integer_types, bool)))):
```
zachmullen, line 107: I would have assumed this would not be valid, but it turns out `isinstance` flattens sub-tuples, which is kind of gross.

girder/utility/system.py
zachmullen: Oh I see now from the first commit.

zachmullen:
Do you also want to remove `six.BytesIO` and `six.StringIO` on this topic? Trying to figure out what's in scope here. Are you planning to just remove all references to `six` everywhere?

brianhelba:
@zachmullen I will probably attempt it in a follow-up PR at some point.

Issue #2403, opened 2017-10-11
Slim down docker image


cjh1:
Look at using alpine as base image as well as using "FROM as" syntax for build step.

kotfic:
Keep in mind that some downstream projects build on girder/girder:latest.  A change of this magnitude might be considered "breaking"

cjh1:
Yes, we should probably do this for Girder 3

danlamanna:
I added some of my Alpine experimentations to [this branch](https://github.com/girder/girder/compare/slimify-docker-images), running our regular Dockerfile yields an image that's 1.68GB, the Alpine version is 1.02GB.

mgrauer:
We could achieve this by adding an additional lighter weight Docker container based on e.g. Alpine, or using a multi-stage build (though that wouldn't work for downstreams building Girder plugins). Other related issues are that all plugins are built by default and some have huge dependency trees.

subdavis:
``` dockerfile
FROM python:3.7-slim as runtime
EXPOSE 8080

WORKDIR /girder

RUN apt update && apt install -qy \
  gcc \
  git
COPY . /girder/

# TODO: Do we want to create editable installs of plugins as well?  We
# will need a plugin only requirements file for this.
RUN pip install --upgrade --upgrade-strategy eager .

COPY --from=girder/girder:latest /usr/share/girder/static/ /usr/local/share/girder/static/

ENTRYPOINT ["girder", "serve"]
```

This slims us down to 534M.  However, we can shave off another 200M-240M by dropping the production dependency on `git` because when building for production, `setuptools-scm` isn't even being used.

I don't know how to avoid installing `setup_requires` dependencies with pip or how to best rework setup.py, but I think it's worth it to get the image down into the 200 range.

brianhelba:
> I don't know how to avoid installing `setup_requires` dependencies with pip or how to best rework setup.py, but I think it's worth it to get the image down into the 200 range.

@subdavis Agreed. I have some crazy ideas about adding environment variable or config file overrides for the automatic version detection, but they're not fleshed out. I'd be happy to discuss this when you have time.

manthey:
One way to do this is with a single command in the Docker that removes git after use:
```
COPY . /girder/
RUN apt-get update \
  && apt-get install -qy --no-install-recommends \
    git \
 && pip install \
  --no-cache-dir \
  --find-links https://girder.github.io/large_image_wheels \
  && apt-get remove -y git \
  && apt-get autoremove -y \
[...]

PR #3244, opened 2020-04-22
slimmer nodeless dockerfile

subdavis:
A potential solution to #2403

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,4 @@
 **/node_modules
 *.egg-info
+Dockerfile
+*.Dockerfile
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 **/node_modules
+*.egg-info
--- a/slim.Dockerfile
+++ b/slim.Dockerfile
@@ -1,0 +1,19 @@
+FROM python:3.7-alpine as builder
+
+WORKDIR /girder
+RUN apk add --no-cache git gcc musl-dev libffi-dev openssl-dev
+COPY . ./
+RUN pip install --no-cache-dir .
+
+FROM python:3.7-alpine as runtime
+LABEL maintainer="Kitware, Inc. <kitware@kitware.com>"
+EXPOSE 8080
+
+# Set environment to support Unicode: http://click.pocoo.org/5/python3/#python-3-surrogate-handling
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+COPY --from=builder /usr/local/lib/python3.7/site-packages/ /usr/local/lib/python3.7/site-packages/
+COPY --from=girder/girder:latest /usr/share/girder/static/ /usr/local/share/girder/static/
+
+ENTRYPOINT ["python", "-c", "from girder import cli; cli.main()", "serve"]
--- a/slim.Dockerfile
+++ b/slim.Dockerfile
@@ -13,7 +13,10 @@ EXPOSE 8080
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
+RUN apk add --no-cache tini
+
+COPY --from=builder /usr/local/bin/girder /usr/local/bin/girder
 COPY --from=builder /usr/local/lib/python3.7/site-packages/ /usr/local/lib/python3.7/site-packages/
 COPY --from=girder/girder:latest /usr/share/girder/static/ /usr/local/share/girder/static/
 
-ENTRYPOINT ["python", "-c", "from girder import cli; cli.main()", "serve"]
+ENTRYPOINT ["/sbin/tini", "-v", "--", "girder", "serve"]
--- a/docs/deploy.rst
+++ b/docs/deploy.rst
@@ -112,14 +112,20 @@ would be::
 Docker Container
 ----------------
 
-Every time a new commit is pushed to master, Docker Hub is updated with a new
-image of a docker container running Girder. This container exposes Girder at
-port 8080 and requires the database URL to be passed in as an option. For more
+Every time a new commit is pushed to master, Docker Hub is updated with new
+images for running Girder. These containers expose Girder at
+port 8080 and require the database URL to be passed in as an option. For more
 information, see the
 `Docker Hub Page <https://registry.hub.docker.com/u/girder/girder/>`_. Since the
 container does not run a database, you'll need to run a command in the form: ::
 
-   $ docker run -p 8080:8080 girder/girder -d mongodb://db-server-external-ip:27017/girder --host 0.0.0.0
+   $ docker run -p 8080:8080 girder/girder --database mongodb://db-server-external-ip:27017/girder --host 0.0.0.0
+
+The ``girder/girder:latest`` image is quite large and intended for use as a base image for plugin developers.
+If you just need to run Girder, a slim image based on alpine linux is provided without
+many of the build tools needed to compile wheels or run girder client builds: ::
+
+   $ docker run -p 8080:8080 girder/girder:slim --database mongodb://db-server-external-ip:27017/girder
 
 Google Container Engine
 -----------------------
--- a/docs/deploy.rst
+++ b/docs/deploy.rst
@@ -121,9 +121,10 @@ container does not run a database, you'll need to run a command in the form: ::
 
    $ docker run -p 8080:8080 girder/girder --database mongodb://db-server-external-ip:27017/girder --host 0.0.0.0
 
-The ``girder/girder:latest`` image is quite large and intended for use as a base image for plugin developers.
-If you just need to run Girder, a slim image based on alpine linux is provided without
-many of the build tools needed to compile wheels or run girder client builds: ::
+The ``girder/girder:latest`` image is quite large and intended for use as a
+base image for plugin developers. If you just need to run Girder, a slim image
+based on alpine linux is provided without many of the build tools needed to
+compile wheels or run girder client builds: ::
 
    $ docker run -p 8080:8080 girder/girder:slim --database mongodb://db-server-external-ip:27017/girder
 


slim.Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.7-slim as runtime
+EXPOSE 8080
+
+WORKDIR /girder
+
+RUN apt update && apt install -qy gcc git
```
manthey, line 6: I recommend adding `--no-install-recommends` and explicitly installing only the packages that you need.  Also, I've found you can remove the apt cache afterwards via `&& rm -rf /var/lib/apt/lists/*`.  It is better form to you `apt-get` rather than `apt` in scripts.

slim.Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+COPY . /girder/
+
+RUN pip install --upgrade --upgrade-strategy eager .
```
manthey, line 9: Add `--no-cache-dir` so packages aren't cached.

slim.Dockerfile
manthey: I think you're pulling `gcc` just so you get `psutils` to build.  You can use a non-official prebuilt wheel by adding `--find-links https://girder.github.io/large_image_wheels` to the `pip install` command and then you won't need `gcc`.

slim.Dockerfile
subdavis: This should probably go in the regular Dockerfile too?

slim.Dockerfile
subdavis: > It is better form to you apt-get rather than apt in scripts.
Why?

> rm -rf /var/lib/apt/lists/*

neat!

manthey: apt emits `WARNING: apt does not have a stable CLI interface. Use with caution in scripts.`.

slim.Dockerfile
manthey: Yes.

slim.Dockerfile
subdavis: Huh, thought they both did that. Thanks.

subdavis:
@manthey 23b1235 gets it down to 400M -- awesome.

slim.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# avoid psutil GCC dependency by using unofficial large_image wheel
+RUN pip install \
+  --no-cache-dir \
+  --upgrade \
+  --find-links https://girder.github.io/large_image_wheels \
+  --upgrade-strategy eager .
+
+COPY --from=girder/girder:latest /usr/share/girder/static/ /usr/local/share/girder/static/
+
+ENTRYPOINT ["girder", "serve"]
```
brianhelba, line 23: It will be [easier to override this](https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact) by invoking users if this is`:
```suggestion
CMD ["girder", "serve"]
```

If we foresee wanting to invoke a variety of different `girder ...` commands in typical usage (which I do not), we could use:
```
ENTRYPOINT ["girder"]
CMD ["serve"]
```

slim.Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM python:3.7-slim as runtime
+EXPOSE 8080
```
brianhelba, line 2:
```suggestion
LABEL maintainer="Kitware, Inc. <kitware@kitware.com>"
EXPOSE 8080
```

slim.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    libsasl2-dev \
+  && rm -rf /var/lib/apt/lists/*
+COPY . /girder/
+
+# avoid psutil GCC dependency by using unofficial large_image wheel
+RUN pip install \
+  --no-cache-dir \
+  --upgrade \
+  --find-links https://girder.github.io/large_image_wheels \
+  --upgrade-strategy eager .
```
brianhelba, line 19: Since we're installing into a fresh image, there won't be any existing packages to `--upgrade` anyway. You should be able to omit:
```suggestion
  --find-links https://girder.github.io/large_image_wheels \
  .
```

`pip`, `setuptools`, and `wheel` are pre-installed, but since they're not in the transitive dependency tree, they won't be affected by any attempts to upgrade.

slim.Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM python:3.7-slim as runtime
+EXPOSE 8080
+
+WORKDIR /girder
+
+RUN apt-get update \
+  && apt-get install -qy --no-install-recommends \
+    git \
+    libldap2-dev \
+    libsasl2-dev \
```
brianhelba, line 10: Since this is not installing `girder-ldap`, you should be able to omit:
```suggestion
```

slim.Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM python:3.7-slim as runtime
+EXPOSE 8080
+
```
brianhelba, line 3: This should help with a common compatibility issue:

```suggestion

# Set environment to support Unicode: http://click.pocoo.org/5/python3/#python-3-surrogate-handling
ENV LC_ALL=C.UTF-8
ENV LANG=C.UTF-8

```

brianhelba:
This is a great idea. @subdavis if you have time, I'd like to discuss how we could unify these changes with the unfinished work in #3167.

manthey:
If you build from a shallow git checkout, you can save much of the space used by the `.git` directory.

manthey:
One of the biggest python packages is botocore (something like 50 Mb).  I don't know enough about what its data contains, but its data files are pretty-printed json with lots of "documentation" strings.  I think to shrink the image as much as possible, you'd have to start doing somewhat hacky things to that data.

slim.Dockerfile
subdavis: Docker users currently have to do this at a bare minimum:

``` bash
docker run girder/girder --host 0.0.0.0
```

If we changed the entrypoint to a CMD, every docker deployment of girder would break.

subdavis:
> If you build from a shallow git checkout, you can save much of the space used by the .git directory.

I'm not sure if this is a response to my comment about git taking up space, but I mean the actual `git` installation and not the `.git` directory of girder.

I've added a new custom dockerignore to deal with this.

> One of the biggest python packages is botocore (something like 50 Mb). I don't know enough about what its data contains, but its data files are pretty-printed json with lots of "documentation" strings. I think to shrink the image as much as possible, you'd have to start doing somewhat hacky things to that data.

I'd be hesitant to commit to maintaining something like that personally, because I don't know how it would affect plugin authors.  If a plugin author notices `git` is missing, the resulting error would be clear enough to indicate that they should `apt-get install git`.  However, I have no idea what sorts of untraceable and arcane issues might come up if we start performing surgery on individual packages.  You definitely understand the botocore anatomy better than me, so maybe that would be pretty harmless.

> This is a great idea. @subdavis if you have time, I'd like to discuss how we could unify these changes with the unfinished work in #3167.

@brianhelba I would prefer to keep 2 different dockerfiles.  Your build results in an image that will  fully support plugin authors who modify the web client.  I think both the fully featured image and the barebones image are useful, and these PRs can proceed more or less independently.

Not sure if that is what you were suggesting, or just some general coordination between the two.

subdavis:
Down to 273MB, but docker CE 19.03 is required and you have to use

`DOCKER_BUILDKIT=1 docker build -f slim.Dockerfile  -t girder/girder:slim .`

brianhelba:
> @brianhelba I would prefer to keep 2 different dockerfiles.

Me too, I was not proposing to merge the files, but many of the improvements can be applied to and coordinated across both files.

manthey:
> > If you build from a shallow git checkout, you can save much of the space used by the .git directory.
> 
> I'm not sure if this is a response to my comment about git taking up space, but I mean the actual `git` installation and not the `.git` directory of girder.

They were orthogonal.  The `apt-get remove git` I suggested would allow using setuptools-scm to have the girder version without adding git to the image.

When you do `COPY`, you are copying the local directory including the `.git` subdirectory.  The `.git` directory, by default, as the history of Girder in it and is ~22Mb currently.  If, before building, you did a shallow clone (`git clone --depth=1 --single-branch`), this directory will be much smaller.

In your current approach, where you are no longer populating the Girder version from setuptools-scm, this directory isn't used at all and a bare checkout should be done to exclude it.

manthey:
I'll note that you could have also `pip uninstall setuptools-scm` if you wanted to save that space, too, while still having a valid Girder version number.

slim.Dockerfile
brianhelba: One idea, get rid of `--host ...` requirement with: https://github.com/girder/girder/pull/3167/files#diff-3254677a7917c6c01f55212f86c57fbfR30

brianhelba: After discussion, I realize it's better to use `ENTRYPOINT`, even if we eliminate the need to pass `--host` by adding a config file. Users are going to typically need to pass `--database`. Using an `ENTRYPOINT` will allow:
```bash
docker run girder/girder --database mongodb://foo/girder
```
instead of
```bash
docker run girder/girder girder serve --database mongodb://foo/girder
```

Realistically, users are not going to need to run `bash` or other commands in this image under normal usage. If they really do need to run something other than `girder serve`, they have the escape hatch of specifying `--entrypoint`.

slim.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    --no-cache-dir \
+    --find-links https://girder.github.io/large_image_wheels \
+    .
+
+FROM python:3.7-alpine as runtime
+
+COPY --from=builder /usr/local/lib/python3.7/site-packages/ /usr/local/lib/python3.7/site-packages/
+COPY --from=girder/girder:latest /usr/share/girder/static/ /usr/local/share/girder/static/
+
+ENTRYPOINT ["python", "-c", "from girder import cli; cli.main()", "serve"]
```
subdavis, line 26: Since girder is no longer pip installed in this image, it must be started in this way.  The result is identical, as far as I can tell.

.dockerignore
```diff
@@ -1 +1,4 @@
 **/node_modules
+*.egg-info
+Dockerfile
+*.Dockerfile
```
subdavis, line 4: Leaving dockerfiles out allows the build cache to work properly when you modify Dockerfile.

slim.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+RUN apk add --no-cache tini
+
+COPY --from=builder /usr/local/bin/girder /usr/local/bin/girder
+COPY --from=builder /usr/local/lib/python3.7/site-packages/ /usr/local/lib/python3.7/site-packages/
+COPY --from=girder/girder:latest /usr/share/girder/static/ /usr/local/share/girder/static/
+
+ENTRYPOINT ["/sbin/tini", "-v", "--", "girder", "serve"]
```
subdavis, line 22: @brianhelba I don't think `tini` is strictly necessary in our case case because python as an entrypoint will already trap signals properly.  However it will collect zombies if girder creates them, and I don't think it hurts to include it here.

Without tini, it is my understanding that python itself would be pid 1, but girder starts with a python shebang, and that has to be interpreted by a shell, so maybe tini is needed after all....

brianhelba: This isn't necessary. The [shebang is interpreted by the program loader itself](https://stackoverflow.com/a/3009280), not the shell.

I've verified via `docker exec girder ps` that `/usr/local/bin/python` is PID 1 if we execute `girder serve`.

subdavis: Verified that tini is necessary, and girder currently ignores sigterm https://github.com/girder/girder/issues/3247

brianhelba:
@subdavis Could you rebase this on `master`? I believe the tests are failing due to its base on an old version.

After that, I can merge in #3246.

subdavis:
@brianhelba done.

slim.Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM python:3.7-alpine as builder
+
+WORKDIR /girder
+RUN apk add --no-cache git gcc musl-dev libffi-dev openssl-dev
+COPY . ./
+# Both PYTHONDONTWRITEBYTECODE and --no-compile are necessary to avoid creating .pyc files
+RUN PYTHONDONTWRITEBYTECODE=1 pip install --no-cache-dir --no-compile .
```
manthey, line 7: A curious property in how docker composes layers is that even if a file is unchanged between the builder and runtime images, it will still be copied and take up space on two image layers.  As such, when we do `COPY --from=builder /usr/local/lib/python3.7/site-packages/ /usr/local/lib/python3.7/site-packages/`, the preinstalled site-packages end up using space twice.  This could be avoided by doing something like:

```
RUN touch /tmp/timestamp

RUN PYTHONDONTWRITEBYTECODE=1 pip install --no-cache-dir --no-compile .

RUN find /usr/local/lib/python3.7/site-packages/ -type f -a -not -newer /tmp/timestamp -delete
```

which obviously breaks the builder's image (since we just deleted pip and some other modules), but will reduce the size of the copy layer by about 6 Mbytes.

slim.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+FROM python:3.7-alpine as runtime
+LABEL maintainer="Kitware, Inc. <kitware@kitware.com>"
+EXPOSE 8080
+
+# Set environment to support Unicode: http://click.pocoo.org/5/python3/#python-3-surrogate-handling
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+RUN apk add --no-cache tini
```
manthey, line 17: Since `docker run` can be invoked with `--init`, do we really want to use `tini`?  It adds 100kb that education about docker options would make unnecessary.

subdavis: `--init` is not a well known feature of docker, and is not widely used.  100kb weighed against the 15 seconds of "oh right I have to --init this one container" that tini will regularly save me is, IMO, a no-brainer.

I'm all for education, but presently `--init` borders on the esoteric.

It's also worth noting that we don't always have the luxury of controlling args to docker run.  Suppose for example that docker-py doesn't support the `init` arg, or we try to deploy with Rancher and it doesn't expose that option, etc.   I'm not even certain the space savings is worth the amount of time it would take to look to see if it's supported.

subdavis: There's a customer satisfaction aspect here too.  For external customers,  even though they _can_ use `--init`, we probably shouldn't degrade their girder experience if they don't.

slim.Dockerfile
subdavis: On the fence about whether or not this is worth it.  Someone else should weigh in.

slim.Dockerfile
brianhelba: See #3248

PR #3246, opened 2020-04-27
Add additional improvements to slim Dockerfile

brianhelba:


slim.Dockerfile
```diff
@@ -3,7 +3,8 @@ FROM python:3.7-alpine as builder
 WORKDIR /girder
 RUN apk add --no-cache git gcc musl-dev libffi-dev openssl-dev
 COPY . ./
-RUN pip install --no-cache-dir .
+# Both PYTHONDONTWRITEBYTECODE and --no-compile are necessary to avoid creating .pyc files
+RUN PYTHONDONTWRITEBYTECODE=1 pip install --no-cache-dir --no-compile .
```
subdavis, line 7: Since this is in the build stage, does it matter?  Do these `.pyc` files end up in site-packages or elsewhere?

zachmullen: How big are the bytecode files anyway? If they're not too big they're probably worth keeping around for the performance boost.

brianhelba: Yes, they go inside `__pycache__` directories scattered throughout `site-packages`, and will be copied along with `site-packages` otherwise.

slim.Dockerfile
```diff
@@ -18,5 +19,9 @@ RUN apk add --no-cache tini
 COPY --from=builder /usr/local/bin/girder /usr/local/bin/girder
 COPY --from=builder /usr/local/lib/python3.7/site-packages/ /usr/local/lib/python3.7/site-packages/
 COPY --from=girder/girder:latest /usr/share/girder/static/ /usr/local/share/girder/static/
+# Add a config file, to bind the server to all network interfaces inside the container
+RUN \
+  mkdir $HOME/.girder && \
+  echo '[global]\nserver.socket_host = "0.0.0.0"\n' > $HOME/.girder/girder.cfg
```
subdavis, line 25: I'm generally not a fan of using dotfiles in home directories in docker since that pattern only really exists to avoid being a nuisance to users.  If I `exec` into the container to look for something, I don't want to get tricked by hidden files.

What do you think about using `/etc/girder.cfg` instead? https://github.com/girder/girder/blob/4cb9a149b8645a98f29f6e078a61a472ce803b96/girder/utility/config.py#L41

If I were groping around in the dark looking for default configuration, the first place I'd check is `/etc`.

zachmullen: +1 for `/etc/girder.cfg`, it seems like a more appropriate path inside a container.

brianhelba: Sounds good.

slim.Dockerfile
manthey: Unless we prevent it (via an environment variable or python flag), they get regenerated as needed and only have a cost on first use of a module.

slim.Dockerfile
subdavis: Actually, if a user were to run with a different pid, this config could get broken anyway.  https://docs.docker.com/engine/reference/run/#pid-settings---pid

It's best to not have behavior depend on what user the container was built with.

slim.Dockerfile
brianhelba: It reduces the build size by ~15MB, FWIW.

subdavis: :+1: from me, but I don't know python mechanics well enough to make a good argument either way.

zachmullen: We can leave them off then. I guess the point of this image is to be small, not necessarily performant.

PR #3248, opened 2020-04-28
Make Girder natively respond to SIGTERM and SIGHUP

### LINKED ISSUE #3247: Make Girder in cherrypy respond appropriately to SIGTERM
Right now `SIGTERM` sent to `girder serve` does nothing. According to [these cherrypy docs](https://docs.cherrypy.org/en/latest/pkg/cherrypy.process.html#cherrypy.process.plugins.SignalHandler), cherrypy should exit gracefully when SIGTERM is sent, but it doesn't. We should figure out why that is. Are we overriding these defaults somehow? Or is this an upstream bug?

brianhelba:
See: http://docs.cherrypy.org/en/3.3.0/refman/process/plugins/signalhandler.html

After this change, Girder will now natively exit when sent `SIGTERM` and `SIGHUP`. This most immediately impacts Girder's behavior in a container, where it is PID 1 (as PID 1 does not default to being killed when receiving an unhandled `SIGTERM`).

This was tested by using `docker kill -s SIGTERM` on a locally built slim container.

Fixes #3247.

### DIFF
--- a/girder/cli/serve.py
+++ b/girder/cli/serve.py
@@ -34,5 +34,6 @@ def main(dev, mode, database, host, port):
     _attachFileLogHandlers()
     server.setup(mode)
 
+    cherrypy.engine.signal_handler.subscribe()
     cherrypy.engine.start()
     cherrypy.engine.block()


subdavis:
What's the benefit of doing it this way?  I can see several drawbacks, but I'm not sure what we get for those trade-offs.

In order of importance, what about...

* [Proper zombie reaping](https://hackernoon.com/my-process-became-pid-1-and-now-signals-behave-strangely-b05c52cc551c) -- are we certain that all PID 1 responsibilities are handled now? See the `A -> B -> C` diagrams about halfway down.  Even though vanilla girder may not create zombies, a plugin may.
* Edge cases: When girder isn't running properly, such as when it's trying to connect to mongo or rabbitmq, this doesn't work.  `tini` still works in these cases.
* Tests: how to we guarantee that this feature doesn't regress without tests?  `tini` and `dumb_init` have extensive test suites.  We definitely shouldn't spend time writing tests for this, and if we aren't going to test it, we probably shouldn't support it.

I'm having trouble seeing a reason to commit to maintaining a partial solution when we can have a robust solution at no cost.

brianhelba:
I've removed the changes to `slim.Dockerfile`, and rebased this to be merged against `master`. This is still a worthwhile change for all runtime environments, and should have been part of Girder's configuration all along. See: http://docs.cherrypy.org/en/3.3.0/refman/process/plugins/signalhandler.html

@girder/developers PTAL.

brianhelba:
@subdavis I wanted to respond substantively to your comment, just to explain my own understanding and rationale.

> * [Proper zombie reaping](https://hackernoon.com/my-process-became-pid-1-and-now-signals-behave-strangely-b05c52cc551c) -- are we certain that all PID 1 responsibilities are handled now? See the `A -> B -> C` diagrams about halfway down.  Even though vanilla girder may not create zombies, a plugin may.

Indeed, vanilla `girder serve` does not create subprocesses. To install a plugin using one of Girder's Docker images, downstreams must at least inherit from the image and add additional build steps to install their plugin, and can easily add `tini` at this point themselves. Realistically, it may be easier for downstreams to assemble a Docker image from scratch themselves, since the multi-stage build of the upstream slim image makes the image unsuitable for fully re-installing / re-building Girder.

> * Edge cases: When girder isn't running properly, such as when it's failing to connect to mongo or rabbitmq, this doesn't work.  `tini` still works in these cases.

Agreed. See my comments in #3244 for some basic experimentation I did to confirm this.

> * Tests: how to we guarantee that this feature doesn't regress without tests?  `tini` and `dumb_init` have extensive test suites.  We definitely shouldn't spend time writing tests for this, and if we aren't going to test it, we probably shouldn't support it.

This PR's change is a documented, established part of CherryPy's standard config. To the extent it doesn't function, it's a bug with upstream CherryPy. It is also possible that upstream `tini` has bugs which cause it not to function. Either way (with this patch, running under `tini`, or both), Girder does not (and practically can not) test its response to `SIGTERM` as part of its normal regression tests.

PR #3249, opened 2020-04-29
Add girder-readme plugin

### LINKED ISSUE #3242: Add README item rendering to folder view
When a folder contains an item called `README` (or variations like `README.txt` or `README.md`), viewing the folder in the web client should render the contents of the `README` under the metadata widget, similar to how GitHub will render `README`s when browsing repositories.

If possible, it should check for a `README` anywhere in the folder contents, not just the paginated contents that are currently being rendered.

I don't see any downside to having this feature, so I recommend it go into Girder Core. The only reason to provide a setting toggle would be if there is some kind of incompatibility with a plugin (perhaps the monkey brains project?), but fixing it in code would be a one time effort that would be part of this work.

dchiquit:
The girder-readme plugin is used to show README files directly in the
hierarchy widget when viewing a folder, if that folder contains an item
starting with `README`.
All file extensions are valid (`README`, `README.txt`, `README.md`,
`README.rst`, etc.), but they will all be rendered as Markdown for
simplicity.
Behavior is undefined for multiple README items and for multiple files
in the item README; it will simply pick arbitrarily from the available
options.

Fixes #3242


plugins/readme/girder_readme/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+from girder.plugin import GirderPlugin
+from . import rest
+
+
+class ReadmePlugin(GirderPlugin):
+    DISPLAY_NAME = 'README'
+    CLIENT_SOURCE_PATH = 'web_client'
+
+    def load(self, info):
+        info['apiRoot'].folder = rest.FolderReadme()
```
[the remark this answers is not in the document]
zachmullen: e.g.

```python
info['apiRoot'].folder.route('GET', (':id', 'readme'), _getFolderReadme)
```

plugins/readme/setup.py
```diff
@@ -0,0 +1,53 @@
...
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
+    ],
+    include_package_data=True,
+    packages=find_packages(exclude=['plugin_tests']),
+    zip_safe=False,
+    install_requires=[
+        'girder>=3',
+        'girder-jobs>=3',
```
danlamanna, line 47: Why does it require girder-jobs?

danlamanna:
FWIW the best practices around plugin development should be documented (in code) here: https://github.com/girder/cookiecutter-girder-plugin.

plugins/readme/setup.py
dchiquit: I copied all this boilerplate from the thumbnail plugin, missed a spot

plugins/readme/girder_readme/rest.py
```diff
@@ -0,0 +1,30 @@
...
+@autoDescribeRoute(
+    Description('Get the README for a folder, if it exists.')
+    .modelParam('id', model=FolderModel, level=AccessType.READ)
+    .errorResponse()
+    .errorResponse('Read access was denied on the folder.', 403)
+)
+def _getFolderReadme(folder):
+    query = {
+        'folderId': folder['_id'],
+        'lowerName': {'$regex': re.compile('^readme')},
```
zachmullen, line 22: We have a compound index for `folderId` and `name`, but not for `folderId` and `lowerName`. So let's search using `name`, and maybe for now we require all-caps for the README part of the name.

We could also refine this regex to be a little more specific e.g. `^README$|^README\..*`

plugins/readme/plugin_tests/readme_test.py
```diff
@@ -0,0 +1,117 @@
...
+            path=f'/folder/{self.privateFolder["_id"]}/readme',
+            method='GET',
+            user=self.admin,
+            isJson=False,
+        )
+        self.assertStatusOk(resp)
+        readme = b''
+        for b in resp.body:
+            readme += b
+        assert self.readmeMarkdown == readme
```
zachmullen, line 98: This is a raw `assert`. BTW, we could move this testing out of here and use pytest instead, if you prefer. It could go in a subfolder under [here](https://github.com/girder/girder/tree/master/test).

dchiquit: If that's the official place where new tests go, I'd be happy to move it. I copied all this stuff from the `thumbnail` plugin, including any bad practices it might have had.

zachmullen: I'm trying to write all new tests in pytest, we made some effort to transition the test suite out of unittest and into pytest, but it only got partway done.

If it's not too much burden, I think it'd be good to move this over. The `jobs` plugin is already there as an example.

plugins/readme/plugin_tests/readmeSpec.js
```diff
@@ -0,0 +1,51 @@
+girderTest.importPlugin('readme');
```
zachmullen, line 1: Does this spec actually get run? Is it harnessed by CTest or pytest?

dchiquit: AFAICT all the plugin frontend tests are run through CTest. It's run through CTest. You can see the logs where it ran in CI:
```
test 76
      Start 76: web_client_readme.readme

76: Test command: /home/circleci/project/girder_env/bin/python "-m" "unittest" "-v" "tests.web_client_test"
...
76: Loaded plugin "readme"
...
76: . Test the readme UI -> register a user
76: /home/circleci/project/girder/girder/models/user.py:613: DeprecationWarning: count is deprecated. Use Collection.count_documents instead.
76:   return folders.count()
76: /home/circleci/project/girder/girder/api/rest.py:441: DeprecationWarning: count is deprecated. Use Collection.count_documents instead.
76:   cherrypy.response.headers['Girder-Total-Count'] = val.count()
78: . Create and log in to a user for testing -> logout
76: /home/circleci/project/girder/girder/models/folder.py:602: DeprecationWarning: count is deprecated. Use Collection.count_documents instead.
76:   return self.childItems(folder, fields=()).count()
76: /home/circleci/project/girder/girder/models/folder.py:623: DeprecationWarning: count is deprecated. Use Collection.count_documents instead.
76:   return folders.count()
76: . Test the readme UI -> navigate to the user's Public folder
76: . Test the readme UI -> does not render the README when there is no README
76: . Test the readme UI -> upload README.md
76: . Test the readme UI -> renders the README
76: 
76: Test the readme UI: 8 of 8 passed.
76: ----------------
76: Testing Finished
76: ----------------
76: 5 specs, 0 failures in 1.334s.
76: 
...
76: ok
76: 
76: ----------------------------------------------------------------------
76: Ran 1 test in 4.801s
76: 
76: OK
```

zachmullen: Thanks. I forgot that gets pulled in by the standard plugin tests thing.

zachmullen:
🎉  great work!

PR #3250, opened 2020-05-15
Fix CWD file download

Erotemic:
This is a bugfix for an issue in girder-client when you download a file without specifying a prefix directory. 

A MWE example to reproduce this issue is:

```bash
girder-client --api-url https://data.kitware.com/api/v1 download --parent-type file 5dcf0d1faf2e2eed35fad5d1 scallop.jpg
```

The issue is that the system attempts to create the containing directory if it does not exist using `dirname`. However, if there is no prefix path as in this example, then the `dirname` call resolves to `''`, which causes the subsequent makedirs call to fail with the error:

```
  File "/home/joncrall/.local/conda/envs/py38/bin/girder-client", line 11, in <module>
    load_entry_point('girder-client', 'console_scripts', 'girder-client')()
  File "/home/joncrall/.local/conda/envs/py38/lib/python3.8/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/joncrall/.local/conda/envs/py38/lib/python3.8/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/joncrall/.local/conda/envs/py38/lib/python3.8/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/joncrall/.local/conda/envs/py38/lib/python3.8/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/joncrall/.local/conda/envs/py38/lib/python3.8/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/joncrall/.local/conda/envs/py38/lib/python3.8/site-packages/click/decorators.py", line 27, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/home/joncrall/code/girder/clients/python/girder_client/cli.py", line 301, in _download
    gc.downloadFile(parent_id, local_folder)
  File "/home/joncrall/code/girder/clients/python/girder_client/__init__.py", line 1225, in downloadFile
    _safeMakedirs(os.path.dirname(path))
  File "/home/joncrall/code/girder/clients/python/girder_client/__init__.py", line 45, in _safeMakedirs
    os.makedirs(path)
  File "/home/joncrall/.local/conda/envs/py38/lib/python3.8/os.py", line 221, in makedirs
    mkdir(name, mode)
FileNotFoundError: [Errno 2] No such file or directory: ''
```

In contrast if you simply prefix the destination path with `./` then the command works:

```bash
girder-client --api-url https://data.kitware.com/api/v1 download --parent-type file 5dcf0d1faf2e2eed35fad5d1 ./scallop.jpg
```

However, a user should not be required to do that. An easy fix for this issue is to check for an empty path variable before calling `makedirs`, and do nothing in that case. Adding this patch fixes the issue.

### DIFF
--- a/clients/python/girder_client/__init__.py
+++ b/clients/python/girder_client/__init__.py
@@ -41,11 +41,12 @@ def _safeMakedirs(path):
 
     :param path: The directory to create.
     """
-    try:
-        os.makedirs(path)
-    except OSError as exception:
-        if exception.errno != errno.EEXIST:
-            raise
+    if path != '':
+        try:
+            os.makedirs(path)
+        except OSError as exception:
+            if exception.errno != errno.EEXIST:
+                raise
 
 
 class AuthenticationError(RuntimeError):


brianhelba:
I'd like to get this issue resolved, since it appears to be a significant bug. I think the fix in #3264 might be more robust though. We should merge one of these two PRs.

Erotemic:
I assume that `resolve_path=True` uses `realpath`, which I try to avoid if possible. The reason I avoid it is mostly if the path is stored somewhere, typically you want to remember the path the user provided rather than whatever it resolves to on a specific machine. For instance if I point to $HOME/data (which is a symlink I have on multiple machines that points to various places depending on the system), I want to remember the $HOME/data string and not what it expands to so I can maintain portability. The same thing is true for relative paths. However, in this case I don't think the path is ever stored anywhere, its simply used to access a file, so I think the fix in #3264 is fine. 

I think both should be merged. While #3264 will fix this specific issue in the CLI, it will not fix the issue in the Python API, so if you are using "girder_client" as a python module (which I often do) you might still run into this.

brianhelba:
> I assume that `resolve_path=True` uses `realpath`

Correct: https://github.com/pallets/click/blob/97de2c97275f52adc5c6ef8bb8aa993ce997fe7e/src/click/types.py#L578

> The reason I avoid it is mostly if the path is stored somewhere, typically you want to remember the path the user provided rather than whatever it resolves to on a specific machine. For instance if I point to $HOME/data (which is a symlink I have on multiple machines that points to various places depending on the system), I want to remember the $HOME/data string and not what it expands to so I can maintain portability. The same thing is true for relative paths. However, in this case I don't think the path is ever stored anywhere, its simply used to access a file, so I think the fix in #3264 is fine.

I agree with your point about `realpath` in general, but it's indeed not directly relevant here, since this is not an application which stores the path values for later reuse. However, if we ask for the parent directory of a symlink, I'd generally think that the "right" thing is to use the unresolved path.

> if you are using "girder_client" as a python module (which I often do) you might still run into this.

Good point. Of course, this only impacts cases where we use `os.path.dirname` before calling `os.makedirs`, and I realize that neither fix explicitly accommodates under direct Python usage [the fact that](https://docs.python.org/3/library/os.html#os.makedirs):

> Note `makedirs()` will become confused if the path elements to create include `pardir` (eg. “..” on UNIX systems).

Erotemic:
Interesting, I didn't know about the issue with makedirs and pardir. I just tested `os.makedirs('temp/bar/../baz', exist_ok=True)` in an empty folder and it created the folders temp, temp/bar, and temp/baz. I wonder why that doesn't raise an error?

I agree the first point isn't relevant, so #3264 should be merged.

The "check for `''` fix" doesn't break anything though right? It might not fully address the case where you have pardir in an non-existing directory, but that issues exists anyway, and it strictly increases the number of cases that would now work, right? So, I'm still in favor of merging both. 

Alternatively, we could make a call to `realpath` in _safeMakedirs, but that seems less good to me.

PR #3255, opened 2020-05-29
Add Item Tags plugin

### LINKED ISSUE #3251: Tagging plugin
Create a new plugin to support tagging items and searching for tags. This will involve:
- A GUI for adding/removing tags on an item
- An endpoint to set the tags on an item
- A new search mode to search by tag
- An endpoint to search for items with a specific tag
- Appropriate indices for the search optimization
- Optional: a plugin settings page where the current tags in use are listed, and an endpoint to provide that data

Tags should be specified in item metadata under the key `girder_search_tags`.

dchiquit:
A plugin for tagging items and searching for those tags.
Item tags are stored in item metadata.
Item tags can be modified from the item view.
There is also a list of allowed tags that is stored in a setting.
Allowed tags can be modified from the plugin settings.

Closes #3251

### DIFF
--- a/plugins/.gitignore
+++ b/plugins/.gitignore
@@ -11,6 +11,7 @@
 !hashsum_download/
 !homepage/
 !item_licenses/
+!item_tags/
 !jobs/
 !ldap/
 !oauth/
--- a/plugins/item_tags/.editorconfig
+++ b/plugins/item_tags/.editorconfig
@@ -1,0 +1,27 @@
+root = true
+
+[*]
+indent_style = space
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+
+[*.py]
+indent_size = 4
+max_line_length = 100
+
+[*.js]
+indent_size = 4
+
+[*.json]
+indent_size = 4
+
+[*.pug]
+indent_size = 2
+
+[*.styl]
+indent_size = 2
+
+[*.yml]
+indent_size = 2
--- a/plugins/item_tags/.gitignore
+++ b/plugins/item_tags/.gitignore
@@ -1,0 +1,50 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+env/
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+parts/
+sdist/
+var/
+wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.coverage.*
+.cache
+coverage.xml
+*.cover
+.hypothesis/
+.pytest_cache/
+
+# pyenv
+.python-version
+
+# dotenv
+.env
+
+# virtualenv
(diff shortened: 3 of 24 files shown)


jbeezley:
There are some lower-level changes I have for this, but I'll start with some high level comments.

I don't know that I'm convinced placing the tags in the metadata is a good idea.  The way the code works here, it places a lot of the logic of managing the tags on the web client.  Girder tends to be more client agnostic, so I think most of this logic should be moved to the server.  At the very least, the server should provide tag validation, which I think may even be possible keeping it in the `meta` object by attaching to the validation events.  Was there a particular motivation for putting the tags where they are?

In terms of functionality, I can imagine it being a fairly common use case to have the tag list user-editable or even different by collection.  Grouping it all into a single setting also seems limiting if the number of tags becomes large.  I wonder how feasible it would be to have the tag list "dynamic" by doing an aggregation query.  I would assume if we indexed the field it would be relatively quick to compute, but I don't know enough about mongo.

dchiquit:
For some context, this plugin is for the project formerly known as UCONN (Modular design of multiscale models-Kitware). The current solution in place is to put tag strings in the descriptions of items, then piggyback off of the title and description text indices to search for them. The collaborators have already implemented a Python/TKinter GUI that manages those description tags. This plugin is intended to make tagging for that project specifically less bad.

I chose to store tags in metadata because metadata seemed like a simple, easy, and fast place to put information about an item. Given that tags are in metadata, I didn't worry to much about server side validation because users could always find a way to edit the metadata directly anyway.

I wasn't aware of metadata validation. That's definitely something I would have liked to do, but given my time constraints I did not. I can add that if there is no larger systemic problem with the design.

It sounds like you're hoping for a "real" tagging solution that will be universally useful and configurable. Sadly a fully fledged and fully featured tagging solution is out of scope for the UCONN project. I just took a few days to do this much to make things run more smoothly. There are absolutely flaws in this design (server side validation, deleting tags from the whitelist doesn't delete them from elements, editing metadata directly, etc). In the spirit of keeping the official Girder plugin list clean, I can move this plugin out to a separate repository if it is too slapdash as is.

jbeezley:
We try to keep the plugins included in the official repo up to a higher standard.  If a plugin is only ever used in one project, then the maintenance burden of keeping it here can be too high.  I do think that a tagging plugin could go in as an official plugin.  But, I think it would need to provide more structure server-side than this.

Others may have a different opinion, but my approach would be to place the tags in a new top-level field as a list of strings.  We would need new endpoints to create and delete tags as well as append the tag list to the normal item return schema.  For search, I would provide a live aggregation of tags globally assuming it can be done efficiently, otherwise a dedicated list of supported tags might be   best.

It's up to @mgrauer and @brianhelba if they want to go the extra mile with this to make it an official plugin.  Otherwise, I think putting in a project specific repo would be the way to go.

zachmullen:
I'm of the same mind as @jbeezley .

If we want it to be a built-in plugin, I agree we should make it an Array of strings as a top-level field, and expose that field at READ level. I think we should just do away with the system setting altogether and let people add arbitrary (non-empty) tags. Maybe we should make them case-insensitive, but I don't have strong feelings.

mgrauer:
Our project can only afford limited funding here and has a fairly narrow use case. I'd say we should move this out to a third party plugin.

brianhelba:
@dchiquit @mgrauer Should this PR be closed?

dchiquit:
Moved this plugin to https://github.com/girder/item_tags, closing this PR.

PR #3257, opened 2020-05-30
Upgrade plugins to use pydicom 2.0.0

brianhelba:


brianhelba:
@girder/developers pydicom 2.0 was just released yesterday (May 29) and has one breaking change.

This will fix the currently-broken Girder tests.

PR #3259, opened 2020-06-11
Remove six

brianhelba:


girder/api/describe.py
```diff
@@ -465,10 +464,10 @@ def listResources(self, params):
 
         routeMap = _apiRouteMap()
 
-        for resource in sorted(six.viewkeys(docs.routes), key=str):
+        for resource in sorted(docs.routes.keys(), key=str):
```
zachmullen, line 467: For a dict `x`, `in x` is the same as `in x.keys()` so we could get rid of a lot of the calls to `.keys()`, but I don't really care either way.

girder/models/model_base.py
```diff
@@ -735,17 +733,17 @@ def _removeSupplementalFields(doc, fields):
...
         if isinstance(fields, dict):
-            for k, v in six.viewitems(fields):
+            for k, v in fields.items():
                 if not v:
                     doc.pop(k, None)
                 else:
                     whitelist.append(k)
             if whitelist:
-                for k in list(six.viewkeys(doc)):
+                for k in list(doc.keys()):
```
zachmullen, line 742: I'd prefer these as dict comprehensions but don't worry about it if you don't feel like it.

zachmullen: Nevermind, I just realized this has to operate on the input rather than return a new dict. Womp womp.

girder/api/describe.py
brianhelba: Agreed. However, I tried to transform the rest of the code minimally with these changes, so that would be a little out of scope.

PR #3264, opened 2020-06-11
Resolve paths in the CLI before passing them to the client library

brianhelba:
For default paths such as ".", this will expand to an absolute path and allow filesystem traversal in the client library to succeed.

See PR #3250 for a fuller description of this issue.


brianhelba:
I think this is a more robust alternative to #3250, since it supports all CLI commands which take paths and other operations beyond just `os.makedirs`.

@Erotemic let me know what you think.

PR #3265, opened 2020-06-12
Ensure girder_client creates parent directories safely

brianhelba:
The usage of "abspath" is only necessary when "dirname" used to pass a parent reference to "makedirs". Otherwise, "makedirs" always works relative to the CWD.


brianhelba:
This obsoletes #3250 and #3264 .

brianhelba:
@Erotemic PTAL

Erotemic:
LGTM.

PR #3268, opened 2020-06-13
Fix incorrect tests in test_plugin_registry

brianhelba:
By inheriting from another GirderPlugin, then calling super(), the test plugin classes in test_plugin_registry were always being marked as loaded, even if they failed to run.

In the case of ThrowsOnLoad, it was never actually throwing an exception where intended; an invalid use of super() was actually causing the exception.

This refactors the test plugins to no longer inherit from an intermediate GirderPlugin class, making these tests more reliable.

### DIFF
--- a/test/test_plugin_registry.py
+++ b/test/test_plugin_registry.py
@@ -33,52 +33,60 @@ class InvalidPlugin(GirderPlugin):
     pass
 
 
-class NoDeps(GirderPlugin):
+class TestLoadMixin:
     def __init__(self, *args, **kwargs):
-        super(NoDeps, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


test/test_plugin_registry.py
```diff
@@ -33,53 +33,61 @@ class InvalidPlugin(GirderPlugin):
     pass
 
 
-class NoDeps(GirderPlugin):
+class TestLoadMixin:
     def __init__(self, *args, **kwargs):
-        super(NoDeps, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
```
zachmullen, line 38: Can this line just be removed?

brianhelba: No. In practice, this calls through to `GirderPlugin.__init__`, which is mandatory.

PR #3270, opened 2020-06-13
Replace use of "mock" with "unittest.mock"

brianhelba:
The "mock" package is just a backport of "unittest.mock" for Python 2.


zachmullen:
`mock` is still listed in `pytest_girder/setup.py` requirements. Can we remove it from there too?

brianhelba:
@zachmullen Good point. Done.

PR #3278, opened 2020-06-19
Rework production install docs to use an Ansible-driven process

brianhelba:


docs/installation.rst
```diff
@@ -89,31 +93,20 @@ over the instance, so the first thing you should do after starting your instance
 the first time is to register a user. After that succeeds, you should see a link
 appear in the navigation bar that says ``Admin console``.
 
+TODO: SMTP
```
brianhelba, line 96: Email is essential for a production system. This should be filled in.

docs/development.rst
```diff
@@ -10,9 +10,6 @@ the tools used to develop Girder.
 Configuring Your Development Environment
 ----------------------------------------
 
-In order to develop Girder, you should first refer to the :doc:`installation quickstart <installation-quickstart>` and
```
brianhelba, line 13: This probably leaves the development docs with inadequate instructions for prereqs.

devops/production-template/playbook.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Deploy Girder
+  hosts: all
+  vars:
+    ansible_python_interpreter: auto
```
zachmullen, line 5: What does this do? Seems like this would be the default?

docs/development.rst
zachmullen: Agreed. I think we can maybe redo this whole section with the aim of being able to point, say, a new intern at the page and have them up and running with our "preferred" development setup.

devops/production-template/provision.sh
```diff
@@ -0,0 +1,14 @@
...
+set -e
+
+# Download Ansible roles
+ansible-galaxy install \
+  --force \
+  --role-file=./requirements.yml \
+  --roles-path=./roles
+
+# Run Ansible playbook
+export ANSIBLE_HOST_KEY_CHECKING=False
```
zachmullen, line 11: Is it best practice to disable host key verification? Forgive my ansible ignorance -- just wondering what problem this is meant to get around.

devops/production-template/playbook.yml
brianhelba: This is just a meme you need to ensure that Ansible uses Python 3 on the target. Once Ansible 2.12 drops, it will no longer be necessary. See https://docs.ansible.com/ansible/latest/reference_appendices/interpreter_discovery.html

devops/production-template/provision.sh
brianhelba: It removes an interactive prompt for the user to accept, particularly if this is run in CI. Otherwise, The first time a machine is connected to, it will always prompt to accept.

zachmullen: That's what I assumed.

docs/plugins.rst
```diff
@@ -307,6 +307,10 @@ any of the available LDAP servers, they will fall back to normal core
 authentication. Documentation of the LDAP standard in general can be found
 `here <https://en.wikipedia.org/wiki/Lightweight_Directory_Access_Protocol>`_.
 
+Installation of this plugin requires some system libraries:
+* On Ubnutu 18.04, install the ``libldap2-dev`` and ``libsasl2-dev`` APT packages.
+* On RHEL (CentOS) 7, install the ``openldap-devel`` and ``cyrus-sasl-devel`` RPM packages.
```
mgrauer, line 312: Why provide RHEL advice here? Everything else is Ubuntu. I think it's better to not have any RHEL leakage.

docs/development.rst
mgrauer: Can you add a follow up issue to track this?

docs/installation.rst
```diff
@@ -1,86 +1,90 @@
...
+* An email address for the administrator of the system.
 
-.. note:: If you installed Girder into your system ``site-packages``, you may
-   need to run the above commands as root.
+Install Girder With Ansible
+---------------------------
+Download Template Files
++++++++++++++++++++++++
+Download the following files to a fresh directory on the controller machine:
+* `requirements.yml <https://raw.githubusercontent.com/girder/girder/master/devops/production-template/requirements.yml>`_
```
mgrauer, line 37: these four bullet points don't look to be rendered correctly in markdown preview

docs/installation.rst
```diff
@@ -1,86 +1,90 @@
...
+.. note::
+   Spoof ``nginx_registration_email`` at your own risk.
+   The email address is only provided to Let's Encrypt,
+   `to provide warnings in case HTTPS auto-renewal is failing <https://letsencrypt.org/docs/expiration-emails/>`_.
+   Under normal circumstances, no emails should ever be sent.
 
-    girder build
+.. note::
+   When specifying Girder plugins, PyPI package names of published Girder plugin packages should be
+   used whenever possible. See :ref:`plugins` for a list of official Girder plugins and associated
```
mgrauer, line 66:
```
:ref:`Plugins <plugins>`
```

instead of 

```
:ref:`plugins`
```

for consistency.

docs/installation.rst
brianhelba: Will fix.

docs/installation.rst
brianhelba: Sure.

docs/plugins.rst
brianhelba: I don't feel strongly. Perhaps this knowledge was hard-won, and we want to preserve it? Since this isn't the overall installation tutorial, I'd be more ok with including information about other platforms.

@zachmullen is both the author of this plugin and perhaps one of the most likely to need to install on a RHEL system. I'm happy to defer to him.

zachmullen: I don't have strong feelings either. This plugin is unusual in that it requires native (non-python) shared libraries to be available, so to me it's OK that it has these instructions. But if we don't want to enumerate platforms, we can just remove these and say "install the following shared libraries on your system" and then enumerate them. (I think it's just `libldap` and `libsasl` but I can't remember for sure.) At which point it'll be up to each user to google it for themselves.

mgrauer: I withdraw my request, reasoning is: it's possibly better to preserve this info than it is to preserve purity.

docs/installation.rst
brianhelba: Done.

docs/development.rst
brianhelba: Done: #3281

brianhelba:
@zachmullen @mgrauer I've added some additional info to the runtime config section in a subsequent commit. I've also addressed all comments from the prior round of reviews.

This is now ready for final review. PTAL.

docs/plugins.rst
```diff
@@ -307,6 +307,10 @@ any of the available LDAP servers, they will fall back to normal core
 authentication. Documentation of the LDAP standard in general can be found
 `here <https://en.wikipedia.org/wiki/Lightweight_Directory_Access_Protocol>`_.
 
+Installation of this plugin requires some system libraries:
```
zachmullen, line 310: "Installation of this plugin requires LDAP and SASL shared libraries to be installed and available to the Girder process. These may be installed system-wide via package managers in the following way:"

brianhelba: @zachmullen Can you just make an inline "change suggestion", with the GH? I'll be happy to apply it.

zachmullen, line 310:
```suggestion
Installation of this plugin requires LDAP and SASL shared libraries to be installed and available to the Girder process. These may be installed system-wide via package managers in the following way:
```

Is that sufficient? Never used this feature before.

brianhelba: Yup. I've just edited your comment to add some line breaks.

#webide4lyfe

PR #3287, opened 2020-07-02
Update apt cache when installing prereqs

zachmullen:


brianhelba:
I'm taking a look at this, but I'd like to better understand the potential impact on idempotentcy. I just need a little time to refresh my own knowledge.

zachmullen:
The message and comments didn't carry over so I'll add them here for posterity.

As is, if you just try to use this role on a fresh ubuntu 18.04 instance, it will choke because the packages listed cannot be installed until an apt cache update is performed. It doesn't break on our recommended installation procedure, but only because the mongo role happens to run first, and it [happens to update the cache itself](https://github.com/girder/ansible-role-girder-mongodb/blob/master/tasks/main.yml#L17).

brianhelba:
@zachmullen Are you ok with my additional changes? If so, feel free to merge.

zachmullen:
Thanks, LGTM

PR #3288, opened 2020-07-02
Allow additional content in ansible service script

zachmullen:
This is one of only three small changes required to these roles to make it broadly applicable to a "specialized-workflow" style girder deployment that involves a custom front-end app and girder-worker for data processing.

This particular change is useful for scenarios where environment variables are used to control runtime behavior of the Girder server-side application.

This can of course be achieved with `lineinfile` or just overriding the whole file with one's own template, and at first that was my approach. However that is problematic because it removes ansible idempotency purity. The girder.girder template task for the service script always shows as dirty, and then the subsequent overwrite does as well, even though the end state is unchanged. That state of affairs will unnecessarily trigger handlers.

### DIFF
--- a/devops/ansible-role-girder/README.md
+++ b/devops/ansible-role-girder/README.md
@@ -14,16 +14,17 @@ Setting `ansible_python_interpreter: auto` will enable this behavior.
 
 ## Role Variables
 
-| parameter                 | required | default                                      | comments                                                                                  |
-| ------------------------- | -------- | -------------------------------------------- | ----------------------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


danlamanna:
Given a user might want to change any section a systemd service file provides, I think it would be preferable to leave it up to the developer to use a systemd drop in unit: https://coreos.com/os/docs/latest/using-systemd-drop-in-units.html.

zachmullen:
Thanks, I had no idea that was possible.

zachmullen:
Worked like a charm @danlamanna

PR #3290, opened 2020-07-06
Document a note about the GitHub release process

brianhelba:


zachmullen:
Is this not something we can do automatically in CI? The fact that our release instructions are just "push a tag" is incredibly nice, adding even one more step is a significant difference.

brianhelba:
@zachmullen It would have to be https://developer.github.com/v3/repos/releases/#create-a-release

I'd support if someone wants to add this infrastructure, but given how often releases are created, how easy it is to use the GUI, and the fact that we have no prior art for calling the GitHub API from CI, I'm not sure if there's a worthwhile return on investment.

zachmullen:
I think it would be worth it not just for Girder, but generally to establish best practices for how to do this across any and all of our repos that store release bundles on GitHub. I can take a look on how to do it myself.

A separate question is, do we even need these releases on GitHub? Is that adding any value?

brianhelba:
> A separate question is, do we even need these releases on GitHub? Is that adding any value?

@zachmullen Fair question. In the case of this repo, since we've created "GitHub releases" in the past, GitHub is now rendering on the side of the https://github.com/girder/girder landing page:
image

Prior to me creating a manual release today, this prominently showed the "Latest" release as v2.5.0, despite the fact that we had created subsequent tags.

If we stop issuing GitHub releases, we'd have to be comfortable with the fact that the landing page would prominently show an out-of-date version as the "Latest" release.

zachmullen:
> If we stop issuing GitHub releases, we'd have to be comfortable with the fact that the landing page would prominently show an out-of-date version as the "Latest" release.

Or, we can simply delete all the existing releases. I'm fine with that, personally. All the package bundles are on pypi, and all the source code is tagged in git already.

zachmullen:
I'm going to close this. My current plan, unless I hear otherwise, is to delete all releases from github, and use the more sanctioned release channels: pypi for package distributions, and git for source distributions. Someone brought up that release metadata on GH can be useful for patch notes, however we don't maintain release notes, and if we did, I'd prefer to keep them in source rather than coupling them to GH anyway. Another issue that was brought up is that it would be nice to have "new release" notifications from GitHub. I agree that it's a nice feature, however we can direct interested users to https://libraries.io/pypi/girder where they can subscribe to releases on pypi, including filtering out prereleases.

If you feel strongly that we should keep the existing releases on GH, speak now or forever hold your peace @girder/core  @girder/developers

jcfr:
This makes sense. Thanks for cleaning things up 🙏

PR #3291, opened 2020-07-10
doc: Fix dockerhub link in deployment-alternatives.rst

jcfr:
This commit changed invalid link `https://registry.hub.docker.com/u/girder/girder/` to https://hub.docker.com/r/girder/girder


zachmullen:
Good catch, thanks!

PR #3297, opened 2020-07-17
cgi.escape to html.escape based on Python Deprecation

tony-goss:
https://github.com/xhtml2pdf/xhtml2pdf/issues/443

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-import cgi
+import html
 import cherrypy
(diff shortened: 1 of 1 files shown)


zachmullen:
We only support python 3.6+, so there's no need for six. We can use `html.escape` unconditionally.

zachmullen:
LGTM. Please squash all commits into one and then we will merge.

Thanks!

tony-goss:
> LGTM. Please squash all commits into one and then we will merge.
> 
> Thanks!

Done!

PR #3299, opened 2020-08-06
Clarify example plugins in the deployment playbook template

brianhelba:


mgrauer (dismissed):

mgrauer:
Another idea is: put invalid yaml in, in order to force people to set the settings you want them to modify.

brianhelba:
@mgrauer Great idea, I will do that.

brianhelba:
@mgrauer I'm actually having trouble with your suggestion. YAML is too tolerant of most of the obvious choices to make it invalid. Trying to add Ansible checks for unset variables would significantly complicate the code.

For now, I think we should merge this as-is. There's only so much we can do to alert people who refuse to read the docs.

mgrauer:
Another idea is to put in a broken plugin, e.g. `girder-this-plugin-does-not-exist` so when run, it would break and they'd have to read the comments more carefully. I'll approve as is, if you want to pursue the intentionally broken plugin that's fine too.

mgrauer (approved):

mgrauer (approved):

brianhelba:
I'll look into the invalid plugin idea. I'll have to examine what the error message looks like.

The worse offender is actually the `nginx_hostname` variable, since if you try to change that on an existing machine, the `girder.nginx` role isn't always able to handle it cleanly. I might actually pursue making `girder.nginx` more robust, so it can handle such changes, since (for a variety of reasons) it's proven to be changed more frequently than I initially expected.

PR #3300, opened 2020-08-06
Use `fnmatch` for CORS Allowed Origins

AlmightyYakob:
This would allow, for example, allowing requests from specific subdomains (my motivation for this was Netlify deploy previews). Regarding deploy previews, that would be accomplished through a CORS allowed origin entry of something like `deploy-preview-*--my-app.netlify.app`.

### DIFF
--- a/test/test_rest_utils.py
+++ b/test/test_rest_utils.py
@@ -6,6 +6,8 @@ import pytz
 
 from girder.api import rest
 from girder.exceptions import GirderException
+from girder.settings import SettingKey
+from girder.models.setting import Setting
 import girder.events
 
 date = datetime.datetime.now()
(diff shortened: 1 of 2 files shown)


zachmullen:
Could you interactive rebase and fixup your last commit into the second commit?

test/test_rest_utils.py
```diff
@@ -132,3 +134,26 @@ def testSetContentDisposition(name, disp, expected):
...
+    ('thing.com', 'thing.com', 'thing.com'),
+    ('thing.com', 'notthing.com', None),
+    ('thing.com', 'subdomain.thing.com', None),
+    ('*.thing.com', 'subdomain.thing.com', 'subdomain.thing.com'),
+    (
+        'fixed*more-fixed.thing.com',
+        'fixed-something-more-fixed.thing.com',
+        'fixed-something-more-fixed.thing.com'
+    ),
+    ('fixed*more-fixed.thing.com', 'other-fixed*other.thing.com', None),
```
brianhelba, line 151: Not sure what the point of this particular case is. An actual origin shouldn't typically have a wildcard. If you want to test what happens if the origin did have a `*`,  it seems like the case could be simpler.

AlmightyYakob: I believe the two tests above your comment are testing against the case I described in the PR description. Specifically, putting a wildcard in the middle of the subdomain of the origin. I know this test case is technically identical to the one on line 145, but I figured I'd include it anyway.

AlmightyYakob:
> LGTM, I particularly like the test. It's very elegant.
> 
> See @zachmullen 's request before merging.

I believe I've done what he asked, have I not @zachmullen?

PR #3308, opened 2020-09-24
Update folder size when deleting from the client.

manthey:
We update the number of items and folders, but not the total size.

The bug appears by going to a folder that contains items, clicking the **i** info button, then deleting at least one item from via the checkboxes and checkbox menu and checking the **i** dialog again.


manthey:
There seems to be a reporting issue with codecov.

zachmullen (approved):

PR #3309, opened 2020-09-29
Accept limit=0 for listResources for python client

JKitok:
limit=0 is a valid argument when listing resources on the REST-api, which,
if given, will list all items. The python client does not accept 0 as a valid
limit due to its "falsyness", and will override it with the default page
size.

This commit fixes the python-client so that limit=0 can be sent to
functions that internally call listResource.


brianhelba (approved):
LGTM, thanks!

brianhelba:
Coverage submission won't work on a fork. Other tests are passing, so I'll merge.

PR #3311, opened 2020-10-01
Add query filter to LDAP plugin

djromberg:
An optional query filter is helpful if, for example, only users of a specific LDAP group should be able to authenticate. This way, the administrator can specify additional criteria for the user search without the need to put them into a special root, e.g., an own OU that is configured for the Base DN.
Whenever the field 'queryFilter' is set to a non-empty string, the search string is ANDed: `(&(<searchField>=<login>)(<queryFilter>))`. If the queryFilter is not set, the search string is `<searchField>=<login>` as it was before in order to stay backwards compatible.

### DIFF
--- a/plugins/ldap/girder_ldap/__init__.py
+++ b/plugins/ldap/girder_ldap/__init__.py
@@ -91,6 +91,9 @@ def _ldapAuth(event):
             conn.bind_s(server['bindName'], server['password'], ldap.AUTH_SIMPLE)
 
             searchStr = '%s=%s' % (server['searchField'], login)
+            if server['queryFilter']:
+                searchStr = '(&(%s)(%s))' % (searchStr, server['queryFilter'])
+
             # Add the searchStr to the attributes, keep local scope.
             lattr = _LDAP_ATTRS + (server['searchField'],)
             results = conn.search_s(server['baseDn'], ldap.SCOPE_SUBTREE, searchStr, lattr)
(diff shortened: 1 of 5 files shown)


zachmullen (changes requested):
Thanks @djromberg ! The code LGTM, but could you fix the linter error, and also squash your last 3 commits into one?

plugins/ldap/girder_ldap/__init__.py
```diff
@@ -91,6 +91,9 @@ def _ldapAuth(event):
             conn.bind_s(server['bindName'], server['password'], ldap.AUTH_SIMPLE)
 
             searchStr = '%s=%s' % (server['searchField'], login)
+            if server['queryFilter']:
```
zachmullen, line 94: Sorry, one more change needed here. This will fail on existing LDAP settings that don't have the `queryFilter` key in them. So in order not to break existing users, let's change this to `if server.get('queryFilter'):`

zachmullen (changes requested):

plugins/ldap/girder_ldap/__init__.py
djromberg: I see, I thought that my change in settings.py added a default but that was just the validation, thanks for the hint, I'll fix it.

zachmullen (approved):

PR #3312, opened 2020-10-02
Add environment variables to upload codecov results without a token

jbeezley:
This worked in https://app.circleci.com/pipelines/github/girder/girder/1036/workflows/de720239-3ce3-4d19-a16b-dee390870e26/jobs/28677.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -73,6 +73,9 @@ commands =
 [testenv:codecov_upload]
 skip_install = true
 skipsdist = true
+passenv =
+    CI
+    CIRCLE*
 deps =
     codecov
 commands =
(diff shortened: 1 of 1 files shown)


zachmullen:
Looks like this doesn't work for codecov

zachmullen:
Actually coverage uploaded just fine... no idea why the codecov status checks are failing

zachmullen (approved):

tox.ini
```diff
@@ -73,14 +73,15 @@ commands =
...
+passenv =
+    CI
+    CIRCLE*
 deps =
     codecov
 commands =
     codecov \
         --disable search pycov gcov \
         --root girder \
-        --required \
```
zachmullen, line 82: I think we should keep `--required`, right? Otherwise it fails silently

zachmullen (changes requested):

tox.ini
jbeezley: Right.  Added back.

zachmullen (approved):

PR #3321, opened 2020-12-30
Fix how tox installs pip packages

manthey:
We explicitly stated an install_command that specified the local directory.  This is part of how tox does installations in any case.  With the introduction of the new resolver in pip, this now fails.

This also adjusts some installation requirements for the pydicom package, as the most recent version of that package has dropped support for Python 3.6 without properly specifying so.

This also specifies the version pyjwt as the oauth plugin does not work with the most recent release.


manthey:
I'm not sure why codecov is complaining; no lines of code were altered.

BryonLewis (approved):
pulled and ran tests, they seem to complete now.

PR #3328, opened 2021-02-08
Upgrade pip in the ansible virtualenv.

manthey:
The CI on this will fail due to some doc issues and that it uses the ansible module BEFORE these changes.

Upgrading pip is necessary, as some dependency pulls in cryptography which now requires a rust executable which is only bundled starting with pip 21.0.1.

### DIFF
--- a/devops/ansible-role-girder/tasks/main.yml
+++ b/devops/ansible-role-girder/tasks/main.yml
@@ -23,6 +23,14 @@
 - import_tasks: nodejs.yml
   when: girder_web|bool
 
+- name: Upgrade pip in the virtualenv
+  pip:
+    name: pip
+    extra_args: "--update"
+    virtualenv: "{{ girder_virtualenv }}"
+    # Implicitly create a Python 3 virtualenv if it doesn't exist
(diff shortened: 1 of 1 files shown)


zachmullen (approved):

devops/ansible-role-girder/tasks/main.yml
```diff
@@ -23,6 +23,14 @@
 - import_tasks: nodejs.yml
   when: girder_web|bool
 
+- name: Upgrade pip in the virtualenv
+  pip:
+    name: pip
+    extra_args: "--update"
```
brianhelba, line 29: This should use `state:`.

brianhelba: Is there a particular minimum version of pip that's necessary? I'd rather ensure we have `>=` that version, rather than the latest, as that impacts idempotency.

manthey: `cryptography` appears to need pip 21.0.1 because it needs rust.  However, it is probably dependent on the version of Python, as if `cryptography` publishes a wheel for that version, then pip 9 might be sufficient.

PR #3330, opened 2021-02-08
Change how pip is upgraded in ansible.

manthey:


devops/ansible-role-girder/tasks/main.yml
```diff
@@ -25,8 +25,8 @@
 
 - name: Upgrade pip in the virtualenv
   pip:
-    name: pip
-    extra_args: "--update"
+    name: "pip>=21.0.1"
+    state: present
```
brianhelba, line 29:
```suggestion
    name: pip
    version: ">=21.0.1"
```

devops/ansible-role-girder/tasks/main.yml
```diff
@@ -25,8 +25,8 @@
 
 - name: Upgrade pip in the virtualenv
```
brianhelba, line 26:
```suggestion
- name: Install a recent pip version
```

devops/ansible-role-girder/tasks/main.yml
```diff
@@ -26,7 +26,8 @@
...
   pip:
     name: pip
-    extra_args: "--update"
+    version: ">=21.0.1"
```
brianhelba, line 29: Is this requirement documented anywhere else?

devops/ansible-role-girder/tasks/main.yml
```diff
@@ -23,10 +23,11 @@
 - import_tasks: nodejs.yml
   when: girder_web|bool
 
-- name: Upgrade pip in the virtualenv
+- name: Install a recent pip version
   pip:
     name: pip
-    extra_args: "--update"
+    version: ">=21.0.1"
+    state: present
```
brianhelba, line 30:
```suggestion
```

This is the default.

manthey: Yep.  You asked for it, though.

devops/ansible-role-girder/tasks/main.yml
manthey: We don't document that we pull in cryptography, and it is a requirement cause by the cryptography module.  How do you recommend we document that an upstream package requires a new version?  We didn't document that we needed pip 9 before this, either (nor check for it).

devops/ansible-role-girder/tasks/main.yml
brianhelba: Yeah, if you're using `state: upgrade`, it's necessary, but the minimum version is better for idempotency. Sorry for the confusion from my lack of specificity.

I have a number of CD pipelines using this role, and minimizing unnecessary Ansible changes is an important requirement.

devops/ansible-role-girder/tasks/main.yml
brianhelba: How about here: https://github.com/girder/girder/blob/master/docs/dependencies.rst ?

brianhelba (approved):
Great, thanks!

PR #3331, opened 2021-02-09
Remove mongo intersphinx mapping.

manthey:
It is broken and prevents the docs from being generated.

### DIFF
--- a/docs/deployment-alternatives.rst
+++ b/docs/deployment-alternatives.rst
@@ -80,7 +80,7 @@ And under the containing ``server`` block, make sure to add the following rule:
 
 Ansible
 -------
-:doc:`installation` provides a complete installation workflow, which depends primarily on the
+The Ansible deployment provides a complete installation workflow which depends primarily on the
 Ansible roles:
 
 * `girder.girder <https://galaxy.ansible.com/girder/girder>`_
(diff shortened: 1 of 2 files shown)


manthey:
Our codecov reporting is still broken, but this fixes the docs build.

zachmullen (approved):

docs/deployment-alternatives.rst
```diff
@@ -80,7 +80,7 @@ And under the containing ``server`` block, make sure to add the following rule:
 
 Ansible
 -------
-:doc:`installation` provides a complete installation workflow, which depends primarily on the
+The Ansible deployment provides a complete installation workflow which depends primarily on the
```
brianhelba, line 83: Is it possible to use another type of intra-doc link here? I'd like to funnel people to the installation page as much as possible.

manthey: @brianhelba Before it had an erroneous link to mongo's documentation.  I think wrong links are worse than no links.  I asked where this should have pointed to on slack, but no one responded. It isn't clear to me where it was supposed to reference.

brianhelba: Ah. This should reference https://github.com/girder/girder/blob/master/docs/deployment.rst

PR #3339, opened 2021-03-09
Use best practices from newer versions of Ansible Molecule testing

brianhelba:


zachmullen:
I don't think I'm in a position to give a stamp of approval that these represent ansible best practices. Is there someone else who knows more about these changes? Or could you provide a description / links about these changes?

brianhelba:
* Using `docker_container` requires that `network_mode` be set to avoid warnings:
https://docs.ansible.com/ansible/latest/collections/community/docker/docker_container_module.html
* Ansible facts are now namespaced to `ansible_facts`. The old behavior of putting them as top-level variables is controlled by `inject_facts_as_vars` and will be disabled in an unspecified future release: https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.5.html#ansible-fact-namespacing
* The pip package `testinfra` has been renamed to `pytest-testinfra`: https://pypi.org/project/testinfra/
* Molecule can now install Ansible as an extra: https://molecule.readthedocs.io/en/latest/installation.html#pip

zachmullen (approved):

PR #3345, opened 2021-03-12
Fix #3344

### LINKED ISSUE #3344: jsonParam array default handled improperly
``` python
jsonParam(
    "arr",
    "A required array",
    paramType='body',
    required=True,
    requireArray=True,
    default=[],
)
```

Should produce a swagger schema where the default is `[]`, not the empty string.

https://github.com/girder/girder/blob/e5b531c7cb6797659341bb56e9d4e8f246ef1743/girder/api/describe.py#L330

subdavis:
Note that this changes the behavior of other falsey json serializable types, such as `0` and `""`.

The current behavior for those is also incorrect, but there may be places where the incorrect behavior is relied upon.

fixes #3344

### DIFF
--- a/girder/api/describe.py
+++ b/girder/api/describe.py
@@ -327,7 +327,7 @@ class Description:
             that were passed.
         :type schema: dict
         """
-        if default:
+        if default is not None:
             default = bson.json_util.dumps(default)
 
         self.param(


zachmullen (approved):

manthey:
I think if you merge or rebase from master the status checks will now pass.

PR #3346, opened 2021-03-16
Clarify explicit support for Ubuntu 18.04 + Python 3.6

brianhelba:


manthey:
I'm against changing all of the setup.py files' classification strings.

brianhelba:
> Pypi lists classification strings are for the purpose of listing "what systems it can run on". Removing these values implies we now think it won't run on those versions (but not so strongly as to change the python_requires field).

We have no CI testing to ensure that other versions actually do work properly. Also, we know that `_hash_state.py` (an essential part of file upload) is particularly sensitive to both the Python version and the OpenSSL version which Python is linked against.

If we're assuming that Girder works properly on other versions of Python based on anecdotal developer reports and downstream usages only:
1. That's a less rigorous standard of software reliability than I think we should tolerate.
2. Why don't we list Python 3.9 in the classifiers too?

---
That said, I'm fine to drop the classifier changes as out of scope of this PR. I suspect most people will be making compatibility decisions based on the Girder docs, rather than PyPI classifiers.

manthey (approved):

PR #3354, opened 2021-04-06
Support cookie lifetime less than a full day

scottwittenburg:
By allowing COOKIE_LIFETIME to be a floating point value, provides more granularity on session timeout.


scottwittenburg:
Cc: @girder/developers

scottwittenburg:
Thanks for the suggestion @zachmullen 👍

zachmullen (approved):

PR #3355, opened 2021-04-15
S3 Assetstore emit events, fix prefix behavior

subdavis:
* Allow blocking of recursive import to target a single directory (as defined by prefix)
* Emit events on import similar to the filesystem assetstore import behavior (items and folders only, no file events)
* ~~Upgrade to `list_objects_v2` (no behavior change, pagination hides the difference)~~
* Bugfix behavior of `importPath`:
  * Example: a folder where you wanted to import all the objects starting with 'foo' by passing prefix `folder/foo` would have a delimiter added to the end and result in no imports
  * Example: a folder where you wanted to import an object exactly matching prefix would also be ignored if you pass prefix `folder/myfile.csv`.
  * If the user supplies a prefix without a trailing slash intending it to be a directory name, everything will still work, you'll just get an extra level of nesting, which I believe is correct behavior.  I expect this behavior would also teach the user how to use delimiters properly.

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import datetime
 import json
 import re
 import urllib.parse
@@ -408,9 +409,10 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
                         yield chunk
             return stream
 
-    def importData(self, parent, parentType, params, progress, user, recursive=True, **kwargs):
+    def importData(self, parent, parentType, params, progress, user, force_recursive=True, **kwargs):
         importPath = params.get('importPath', '').strip().lstrip('/')
         bucket = self.assetstore['bucket']
+        now = datetime.datetime.utcnow()
         paginator = self.client.get_paginator('list_objects')
         pageIterator = paginator.paginate(Bucket=bucket, Prefix=importPath, Delimiter='/')
         for resp in pageIterator:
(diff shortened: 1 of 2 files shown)


girder/utility/s3_assetstore_adapter.py
```diff
@@ -406,14 +409,10 @@ def stream():
...
-    def importData(self, parent, parentType, params, progress, user, **kwargs):
+    def importData(self, parent, parentType, params, progress, user, recursive=True, **kwargs):
         importPath = params.get('importPath', '').strip().lstrip('/')
-
-        if importPath and not importPath.endswith('/'):
-            importPath += '/'
-
         bucket = self.assetstore['bucket']
-        paginator = self.client.get_paginator('list_objects')
+        paginator = self.client.get_paginator('list_objects_v2')
```
zachmullen, line 415: This will not work if using Google Cloud Storage. Is it strictly necessary to switch to v2 for what you need?

subdavis: I've tested it using google cloud storage and it appears to work.  Is there something saying it doesn't?

subdavis: Weird, there appears to be more nuance here.  These lines work, but https://issuetracker.google.com/issues/155109631 does not work.  I don't know what the difference is, but I'll just revert to list_objects.

zachmullen: If it's as simple as just using `list_objects` instead of `list_objects_v2`, let's do that. I'm not sure of all the gory details, I just know that last week when I wrote a very basic script to list objects in a GCS bucket using this same paginator API, it didn't work with `v2` but worked fine with `list_objects`. There are some slight configuration differences that could account for it. 🤷

manthey (approved):

subdavis:
@manthey @brianhelba please merge if you're satisfied with this PR.

PR #3361, opened 2021-06-25
Raise 413 for DocumentTooLarge on job log update

subdavis:
I've been dealing with a 500 when the job log becomes too large. 

An argument could be made that most database operations could cause a `DocumentTooLarge` and the error should happen inside `model_base.py` for the `update()` method.

I'm also not 100% sure a 413 is appropriate since it isn't strictly the request that was too large, but the straw that broke the camel's back.  This PR will be paired with a change to `_flush()` in girder_worker's JobManager to catch and truncate the logs.

An alternative is to catch this error and truncate server-side without letting the client decide what to do, but I think it's better to give the client control.  Mostly, I don't think worker process that uses jobManager should ever fail because it ran out of room to store logs.  Especially when you have jobs that take hours or days.

### DIFF
--- a/plugins/jobs/girder_jobs/models/job.py
+++ b/plugins/jobs/girder_jobs/models/job.py
@@ -1,16 +1,18 @@
 # -*- coding: utf-8 -*-
 import datetime
+
 from bson import json_util
+from pymongo.errors import DocumentTooLarge
 
 from girder import events
 from girder.constants import AccessType, SortDir
-from girder.exceptions import ValidationException
+from girder.exceptions import RestException, ValidationException
 from girder.models.model_base import AccessControlledModel
 from girder.models.notification import Notification
 from girder.models.token import Token
 from girder.models.user import User
 
-from ..constants import JobStatus, JOB_HANDLER_LOCAL
+from ..constants import JOB_HANDLER_LOCAL, JobStatus
 
 
 class Job(AccessControlledModel):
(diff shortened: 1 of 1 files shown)


plugins/jobs/girder_jobs/models/job.py
```diff
@@ -397,7 +399,12 @@ def updateJob(self, job, log=None, overwrite=False, status=None,
             job['updated'] = now
             updates['$set']['updated'] = now
 
-            updateResult = self.update(query, update=updates, multi=False)
+            try:
+                updateResult = self.update(query, update=updates, multi=False)
+            except DocumentTooLarge:
+                # It's plausible that the cause of this error is that the log has become too large
+                raise RestException('Document size exceeded', code=413)
```
zachmullen, line 406: Typically models shouldn't throw RestExceptions. Let's handle this in the REST layer.

zachmullen (changes requested):

plugins/jobs/girder_jobs/models/job.py
subdavis: See https://github.com/girder/girder/pull/3361#issue-678099222

Any `BaseModel.update()` could actually throw a DocumentTooLarge that I think should throw a 413.  Moving the RestException into BaseModel might be a better move.

I agree that models shouldn't really throw RestException, but since this is an error coming from pymongo and you'd have to catch it in hundreds of places, this seems like a reasonable exception to that rule.

What do you think?

zachmullen: If you wanted to handle this more globally, I would suggest adding a `DocumentTooLarge` exception handler [here](https://github.com/girder/girder/blob/master/girder/api/rest.py#L682), and have that respond with 413. That keeps the model layer and HTTP layer separate, and if people want to handle `DocumentTooLarge` more ad hoc, they can still do so -- this would just act as a fallback for any unhandled ones in the REST layer.

subdavis: I'll do that, thanks.

subdavis:
Closing out remaining PRs.

PR #3363, opened 2021-07-09
Updated changelogs with redirect to releases page

### LINKED ISSUE #3362: ChangeLog appears to be outdated and gives wrong impression
Looking at https://girder.readthedocs.io/en/latest/changelog.html brings up a major question: Apparently, there were releases for 2.3.0, 2.4.0, 2.5.0, but nothing after that. Even though one can clearly see tags that seem to be releases (up until 3.1.5 ATM), the heading "Unreleased" after 2.5.0 makes the reader of the current documentation wonder if all subsequent versions visible as github tags, dockerhub tags, RTD versions, etc. should be considered "unreleased".

On Gitter, I was told that this impression is indeed wrong, so I herewith suggest to adapt the ChangeLog to either
* list current versions, or
* to indicate that the changelog is no longer maintained, but that recent releases can be found somewhere else (e.g., linking to GitHub).

f4str:
resolves #3362 

Updated ChangeLog to show it is no longer being maintained and provided a link to the releases page.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,8 +2,13 @@
 Release Notes
 =============
 
-This is the summary list of changes to Girder between each release. For full
-details, see the commit logs at https://github.com/girder/girder
+**This page is no longer being maintained.**
+
+This is the summary list of changes to Girder between each release.
+To find information about the latest release, see the releases page at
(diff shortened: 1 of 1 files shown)


zachmullen (approved):

brianhelba:
#3364 should fix this failing test, then this can be merged.

PR #3367, opened 2021-11-25
Implement the seekable function for the FileHandle class

pchoisel:
This method is required for the Python's IOBase implementation and
implementing it fixes a problem with the Python 3.8 implementation
of ZipFile.


finetjul:
LGTM

@zachmullen, WDYT ?

zachmullen (approved):
LGTM!

PR #3368, opened 2021-11-29
Pin pymongo version.

manthey:
Pin pymongo to version < 4 until we can verify Girder works with pymongo 4.

Remove a reference to bson.py3compat.  Since we require Python 3, "abc" can just be imported from collections.


zachmullen (approved):

manthey:
Merging without the ansible test passing, as the ansible test pulls girder from pypi (which is failing), not from the repo.

PR #3370, opened 2021-11-30
Update package-lock with newer versions.

manthey:
This should satisfy the dependabot alerts.


zachmullen (approved):
Assuming the front-end still renders fine, LGTM.

PR #3386, opened 2022-03-22
Remove eslint-plugin-underscore.

manthey:
This removes some lint checks.  The module eslint-plugin-underscore hasn't been updated since 2016 and relies on a version of lodash that is considered to have a critical security issue.  This stops using it; it really should be replaced with a better module, but a superficial search doesn't turn up a newer module.

### DIFF
--- a/girder/web_client/eslint-config/README.md
+++ b/girder/web_client/eslint-config/README.md
@@ -7,7 +7,7 @@ Backbone-based web clients and plugins.
 Typically, users of this package should depend on
 `@girder/eslint-config` and its peerDependencies via:
 ```bash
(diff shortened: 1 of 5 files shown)


subdavis (approved):
I looked in js files for the pattern _\.[a-zA-Z]+ and I don't think a migration from underscore to lodash would be very hard.  There are 314 usages of underscore, but maybe only 15 different functions are used.
https://github.com/lodash/lodash/wiki/Migrating is very helpful.
[9:59](https://kitware.slack.com/archives/C012L7G09JQ/p1647957554783369)
Realistically though, I don't think G3 client code is going to be changing much ever again, so linting rules aren't all that important when code isn't having regular revisions.  Risk of just disabling seems low.

I've approved if you want to merge, but if you'd rather have a review from a maintainer, disregard.

PR #3403, opened 2022-04-01
Wrap route handler and dispatch `failed` event

AlmightyYakob:
This PR adds a `failed` rest event, in the form of `rest.method.group.failed`. This will fire if an exception is raised during the request, re-raising the exception as to not modify behavior.

### DIFF
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -919,12 +919,17 @@ class Resource:
 
             ``rest.get.item/:id.after``
 
-        would be fired after the route handler returns. The query params are
-        passed in the info of the before and after event handlers as
-        event.info['params'], and the matched route tokens are passed in
-        as dict items of event.info, so in the previous example event.info would
-        also contain an 'id' key with the value of 123. For endpoints with empty
-        sub-routes, the trailing slash is omitted from the event name, e.g.:
+        would be fired after the route handler returns. If the request fails,
+        instead of firing the after event,
+
+            ``rest.get.item/:id.failed``
+
+        would be fired. The query params are passed in the info of the before,
+        after, and failed event handlers as event.info['params'], and the matched
+        route tokens are passed in as dict items of event.info, so in the previous
+        example event.info would also contain an 'id' key with the value of 123.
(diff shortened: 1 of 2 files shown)


girder/api/rest.py
```diff
@@ -965,7 +970,14 @@ def handleRoute(self, method, path, params):
...
             self._defaultAccess(handler)
-            val = handler(**kwargs)
+
+            # Attempt to handle and trigger failed event if an
+            # exception is raised.
+            try:
+                val = handler(**kwargs)
+            except Exception as e:
+                events.trigger('.'.join((eventPrefix, 'failed')), kwargs)
+                raise e
```
zachmullen, line 980:
```suggestion
            except Exception:
                events.trigger('.'.join((eventPrefix, 'failed')), {'kwargs': kwargs})
                raise
```

zachmullen (changes requested):

manthey:
We should add this event to the docs in docs/plugin-development.rst.

girder/api/rest.py
AlmightyYakob: I didn't know you could re-raise like that, thanks!

What's the reason for wrapping `kwargs` in a dict? It seems to differ from how the `before` and `after` events include these.

zachmullen: I don't have strong feelings, but passing it wrapped in an outer dict allows easily adding more info the event in the future, without breaking compatibility for anyone.

AlmightyYakob: I suppose that's true (I could image passing the exception itself as well) but I'd think consistency with the before and after events would be more important. Maybe not, @manthey any thoughts?

manthey: I'd rather have the consistency.  To my knowledge, we haven't wanted to extend the before and after events.

zachmullen (approved):

manthey:
For some reason, the ansible test is brittle and can fail without indicating an actual problem.  This can be merged.

manthey (approved):

AlmightyYakob:
@manthey @zachmullen I'm not authorized to merge so one of you will need to actually hit merge.

PR #3431, opened 2022-11-16
Make the hash_state library work with multiple OpenSSL versions

manthey:
CPython can be linked with OpenSSL 1.x or OpenSSL 3.x.  Before this assumed the structure that OpenSSL 1.x was using. 

One place where OpenSSL 3.x is used with CPython in Circle-CI's cimg/python images (but not the circleci/python images).

### DIFF
--- a/girder/utility/_hash_state.py
+++ b/girder/utility/_hash_state.py
@@ -4,9 +4,10 @@
 # Inspired by http://code.activestate.com/recipes/
 # 578479-save-and-restore-sha-512-internal-state/
 
-import hashlib
-import ctypes
 import binascii
+import ctypes
+import hashlib
+import ssl
 import sys
 
 _ver = sys.version_info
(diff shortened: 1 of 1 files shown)


zachmullen (commented):
It may make sense to keep this limping along, but this seems like a good time to evaluate switching to https://github.com/zachmullen/hashstate , which has the benefit of ensuring this will work at compile time rather than runtime.

manthey:
> It may make sense to keep this limping along, but this seems like a good time to evaluate switching to https://github.com/zachmullen/hashstate , which has the benefit of ensuring this will work at compile time rather than runtime.

Looking at that library, the CI tests never passed and it hasn't been built for anything after Python 3.6.  Does  it work with both openssl 1.x and openssl 3.x?

zachmullen:
No, the point is we would need to revive it and expand the build matrix. It _might_ work with OpenSSL 3 if the only relevant change was the position of a struct member, rather than changing its name or structure or moving it elsewhere.

manthey:
We keep using Girder 3 in projects where it makes sense.  I'd love to gradually bring the tooling up to date.  I view this as relatively low priority in that scheme as it works currently.  I'd like to merge this as it stands and revisit doing it via your repo later once we bring harder tooling into line.

zachmullen:
Fine to merge this.

PR #3438, opened 2023-02-22
Remove dependence on global cherrypy.tree

zachmullen:
I created (downstream for now) a WSGI script to run a Girder 3 app + plugins using alternate WSGI servers. In that script I make use of my own cherrypy `Tree` (which is the WSGI app object) rather than relying on the globally shared `cherrypy.tree`. The only place I found so far that broke in that case was in the swagger description generation, where we were referencing `cherrypy.tree`. This alternative method uses the `cherrypy.request.app` object instead, which is a pattern more in line with how e.g. Flask uses the WSGI app object.

As part of this change I simply deleted the `utility/resource.py` module, as it wasn't being used anywhere else, and only contained code that was terrible and shouldn't be used anywhere else.


girder/api/describe.py
```diff
@@ -478,11 +477,28 @@ def listResources(self, params):
         # Definitions Object
         definitions = dict(**docs.models[None])
 
+        def _walkTree(node, path=()):
```
manthey, line 480: Inserting this here makes the cyclometric complexity check fail.  Maybe hoist it out of the methd?

zachmullen: Done :)

manthey:
We are getting this error:
```
    routeMap = _walkTree(cherrypy.request.app.root.api.v1)
AttributeError: 'ApiDocs' object has no attribute 'api'
```

zachmullen:
I think there is undefined behavior because we mount multiple apps in our server.py code, one under `/api` explicitly, and the other under `/` with a `.api` attribute, and I think which ones gets chosen (seemingly undefined) affects what the value of `request.app.root` is. I'm digging into it more; I may try and only conditionally mount under `/api` if we it's not already mounted there. There's a lot of mess here 😄

manthey:
> I think there is undefined behavior because we mount multiple apps in our server.py code, one under `/api` explicitly, and the other under `/` with a `.api` attribute, and I think which ones gets chosen (seemingly undefined) affects what the value of `request.app.root` is. I'm digging into it more; I may try and only conditionally mount under `/api` if we it's not already mounted there. There's a lot of mess here 😄

Yes.  In my local instance, if I change `routeMap = _walkTree(cherrypy.request.app.root.api.v1)` to `routeMap = _walkTree(cherrypy.request.app.root.v1)`, swagger works.  I haven't thought through what that means.

zachmullen:
This solution accommodates the API being potentially mounted at multiple places, and also makes it so that only happens when necessary. I'd love to rip out the double mounting logic entirely, but I'm sure someone out there is depending on it.

manthey (approved):

PR #3439, opened 2023-03-02
Change how we test if we can access s3 readonly assetstores

manthey:
Before, if a prefix was used and access to the root bucket was limited but access to a folder was allowed, this would always fail.

Also, ensure imports are from the prefix used in the assetstore.

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -417,10 +417,6 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
     def importData(self, parent, parentType, params, progress,
                    user, force_recursive=True, **kwargs):
         importPath = params.get('importPath', '').strip().lstrip('/')
-        if (self.assetstore.get('prefix', '')
-                and importPath != self.assetstore['prefix']
-                and not importPath.startswith(self.assetstore['prefix'] + '/')):
-            importPath = (self.assetstore['prefix'] + '/' + importPath).rstrip('/')
         bucket = self.assetstore['bucket']
         now = datetime.datetime.utcnow()
         paginator = self.client.get_paginator('list_objects')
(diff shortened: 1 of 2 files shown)


manthey:
To test this, I created a bucket with folders "limited" and "open" and added the following bucket policy:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "PublicRead",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::girder-folder-test-1/second/*"
        },
        {
            "Sid": "PublicList",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:ListBucket",
            "Resource": "arn:aws:s3:::girder-folder-test-1",
            "Condition": {
                "StringLike": {
                    "s3:prefix": "open*"
                }
            }
        }
    ]
}
```
Probably that prefix should be a list of `["open", "open/*"]`.

On the primary permissions on the bucket, I unchecked "Block public and cross-account access to buckets and objects through any public bucket or access point policies".

girder/utility/s3_assetstore_adapter.py
```diff
@@ -73,7 +73,12 @@ def validateInfo(doc):
         client = S3AssetstoreAdapter._s3Client(params)
         if doc.get('readOnly'):
             try:
-                client.head_bucket(Bucket=doc['bucket'])
+                paginator = client.get_paginator('list_objects')
+                pageIterator = paginator.paginate(
+                    Bucket=doc['bucket'], Prefix=doc['prefix'], Delimiter='/')
+                for resp in pageIterator:
+                    resp.get('Contents', [])
```
zachmullen, line 80: Why is this line here? Is this not sufficient:

```python
client.list_objects(Bucket=doc['bucket'], Prefix=doc['prefix'], Delimiter='/', MaxKeys=1)
```

?

manthey: It probably is.  Sadly, I copied what we do when we enumerate items in a bucket/prefix, rather than figuring out the best way to do it.

manthey: See https://github.com/girder/girder/pull/3440

PR #3440, opened 2023-03-03
Use a simpler method to check if a bucket and prefix are listable.

manthey:


zachmullen (approved):

manthey:
@zachmullen As a minor follow up to importing S3 read-only data from a restricted bucket where there is only access to a prefix: The change we made to Girder works but requires a mildly eccentric "Path Prefix" in the assetstore configuration for certain AWS ACL choices.  Specifically, if the prefix is exposed via:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "PublicRead",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::girder-folder-test-1/open/*"
        },
        {
            "Sid": "PublicList",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "s3:ListBucket",
            "Resource": "arn:aws:s3:::girder-folder-test-1",
            "Condition": {
                "StringLike": {
                    "s3:prefix": "open/*"
                }
            }
        }
    ]
}
```
You have to specify the prefix in Girder as `open/*`, not `open`.  If the `ListBucket` condition is `"s3:prefix": "open*"`, then a prefix in Girder of `open` is sufficient.

Maybe we should change the logic in Girder so if `client.list_objects(Bucket=doc['bucket'], Prefix=doc['prefix'], Delimiter='/', MaxKeys=1)` fails we retry with a prefix of `doc['prefix'] + '/'`.

zachmullen:
Interesting. Your proposed change is fine by me, and we should add a comment to that effect.

PR #3445, opened 2023-05-12
Package updates to support newer node

manthey:
master builds and tests on node 14 / npm 6.  With these changes, we can on npm 8 and run on node 16, but doesn't yet pass tests  under node 16.

To install on npm >= 7, it is necessary to add `--install-links` to the `npm install` command.  Otherwise, plugins' node modules aren't installed; with it, they are installed alongside the girder node modules (whereas with npm < 7, they are installed next to each plugin).

Upgrading to webpack 3 doesn't change much, but does show off that we failed to properly ask for plugins during one of the client tests.

Currently, node 16 reports build success, but then fails to run because `webpack.DllReferencePlugin` seems to not locate packages on start.

### DIFF
--- a/girder/cli/build.py
+++ b/girder/cli/build.py
@@ -77,7 +77,7 @@ def main(dev, mode, watch, watch_plugin, npm, reinstall):
             shutil.rmtree(path, ignore_errors=True)
 
         # Run npm install
-        installCommand = [npm, 'install']
+        installCommand = [npm, 'install', '--install-links']
         if mode == ServerMode.PRODUCTION:
             installCommand.append('--production')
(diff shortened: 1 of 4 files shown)


zachmullen (approved):

manthey:
To go to node >=18, we'll have to upgrade webpack -- webpack 3 uses a deprecated call in node.  We probably should get the tests working on Node 16 first (probably by switching away from PhantomJS).

zachmullen:
> Currently, node 16 reports build success, but then fails to run because webpack.DllReferencePlugin seems to not locate packages on start.

Interesting, I would think this breakage would be more probable based on webpack version than node version.

PR #3449, opened 2023-05-24
Web client build system replatforming

jeffbaumes:
Very very WIP. Just putting this out there for people to kick the tires and tell me what I did wrong in this experiment.

See `girder/web_client/README.md` for details.


girder/api/v1/system.py
```diff
@@ -121,6 +122,21 @@ def getSetting(self, key, list):
...
+    def getPublicSettings(self):
+        publicSettings = [
+            SettingKey.BRAND_NAME,
+            SettingKey.CONTACT_EMAIL_ADDRESS,
+            SettingKey.PRIVACY_NOTICE,
+            SettingKey.BANNER_COLOR,
+            SettingKey.REGISTRATION_POLICY,
+            SettingKey.ENABLE_PASSWORD_LOGIN,
+        ]
+        return {k: Setting().get(k) for k in publicSettings}
```
zachmullen, line 138: It would be good to add a batch lookup function in the setting model to avoid this many RTTs to mongo.

jeffbaumes: Agree. Also the current admin settings page queries a much longer list of settings from the `getSetting` method above which has a similar structure to this.

jeffbaumes:
@zachmullen @manthey @marySalvi FYI, comments welcome.

There is now a concept plugin here that loads properly in a new `web/client` directory (original code from `web_client` is now in a `web/core` library package):

https://github.com/girder/girder/tree/2726dab8b4e502e24ba276d903eb5d5fa65a7c4c/plugins/homepage/girder_homepage/web_client

Here is the main conceptual difference: Instead of getting all the source files and build config from core and plugins lined up and doing a massive build, each plugin is built independently. This means each plugin can bundle itself and its resources however it wants without affecting the main build. The largest downside is that it's a major change, especially in how plugins interact with core Girder JS. Instead of importing classes, etc. from Girder, the plugin exports an `init()` function that takes the global `girder` instance as an argument. This object should have all the information needed to monkeypatch / change / add to Girder as needed for the plugin to work. If you think of all plugins as monkeypatching a single `girder` object in turn, that's pretty much what now would happen in this mode. I believe that's pretty much what all plugins do?

My thinking along these lines came in part by looking at [this article](https://javascript.plainenglish.io/how-to-build-a-plugin-system-with-node-js-68c097eb3a2e) on building a plugin system, where there is a `load(app)` function exported from each plugin module, which are loaded dynamically (this is code from the article, not my code):
```
const module = require(path);
this.plugins[plugin] = module;
await this.plugins[plugin].load(this.app);
```

The workflow I was shooting for in building a new app instance was:

For the server (same as before):
```
pip install girder-plugin-homepage
girder serve
```

For the client:
```
cookiecutter girder/girder-vite-client myclient # Note: This cookiecutter does not exist yet
cd myclient
npm install @girder/girder-plugin-homepage
npm run dev # or npm run build for production
```

So a plugin installation requires both a `pip install` (if there is a server component) and an `npm install` (if there is a client component). I have given no thought yet to how to ensure pip and npm package versions match.

A [`loadPlugins()`](https://github.com/girder/girder/blob/2726dab8b4e502e24ba276d903eb5d5fa65a7c4c/girder/web/core/src/index.ts#L23) function would inspect the `node_modules` for any packages matching the pattern `*girder-plugin-*` and load them in order based on dependency as exported as a `dependencies` array of strings in the plugin's module. This means there can be a standard [`main.ts`](https://github.com/girder/girder/blob/2726dab8b4e502e24ba276d903eb5d5fa65a7c4c/girder/web/client/src/main.ts) which does not need to be updated when installing new client plugins.

The cookiecutter would look something like this [`client`](https://github.com/girder/girder/tree/2726dab8b4e502e24ba276d903eb5d5fa65a7c4c/girder/web/client) directory, which is a small delta from stock `npm init vue@latest` with a new `main.ts` and with `@girder/core` and `@girder/girder-plugin-homepage` added to `package.json` dependencies.

Now that I've mangled a good part of Girder's client code to get this working we'd need to decide what to keep (if anything) and what to toss. Actually almost all of the core JS remains unchanged. The plugin code needed a bigger change but is mostly just indented so init code runs in a function instead of running on module import.

zachmullen:
> The largest downside is that it's a major change

girder/api/rest.py
```diff
@@ -740,7 +741,9 @@ def _setCommonCORSHeaders():
         # If there is no origin header, this is not a cross origin request
         return
 
-    allowed = Setting().get(SettingKey.CORS_ALLOW_ORIGIN)
+    allowed = os.getenv('GIRDER_CORS_ALLOW_ORIGIN')
```
zachmullen, line 744: This logic is now obviated by #3451

jeffbaumes: I've rebased to now use your general code path to set this setting and backed out my change for this one setting.

jeffbaumes:
After an attempt to update the `jobs` plugin which is much larger, I've already backtracked on my `init` function idea to something much closer to being backwards compatible. The idea is still the same, to call an initialization for each plugin by passing the global `girder` object. But instead of an actual `init()` function and literally passing `girder` as an argument, we can have the same effect by setting a global `girder` object on the `window` and then loading the module. This means we can keep plugin code unindented and much more "as-is" without wrapping things in initialization functions. The main difference is that we are pulling Girder things out of the `girder` object instead of by importing. So for example, this import code from a plugin source file:

```
import FrontPageView from '@girder/core/views/body/FrontPageView';
import { renderMarkdown } from '@girder/core/misc';
import { restRequest, getApiRoot } from '@girder/core/rest';
import { wrap } from '@girder/core/utilities/PluginUtils';
```

translates fairly mechanically to this code:

```
const FrontPageView = girder.views.body.FrontPageView;
const { renderMarkdown } = girder.misc;
const { restRequest, getApiRoot } = girder.rest;
const { wrap } = girder.utilities.PluginUtils;
```

This only affects imports from plugin code. Core Girder code can import as normal.

The other issue is dependencies. I had been exporting a `dependencies` list from the module but now it needs to be available outside the JavaScript code because they need to be known before loading the plugin module. This can be accomplished similar to how it was in pre-vite Girder, by using a `girderPlugin` config section in `package.json`. The plugin module `package.json` files can be loaded dynamically just like we find the modules themselves.

I think this is a better transition path that should hopefully still work.

I've also determined that by adding the right `.d.ts` and `declare global` statements for the `girder` global I'm able to get reasonable code completion in VSCode on the Girder core library from plugins, even in `.js` files.

zachmullen:
> const FrontPageView = girder.views.body.FrontPageView;
const { renderMarkdown } = girder.misc;
const { restRequest, getApiRoot } = girder.rest;
const { wrap } = girder.utilities.PluginUtils;

That is basically what our existing `DllReferencePlugin` was doing; deferring linkage to a singleton girder core library at runtime.

jeffbaumes:
> That is basically what our existing DllReferencePlugin was doing; deferring linkage to a singleton girder core library at runtime.

Makes sense, it seemed that's how it needed to have worked. Is it worth trying to get vite to do more precisely the same thing? That is, have plugins depend on `@girder/core` and do imports as normal but at runtime (when pulling the app bundle together) ensure it all works off the same Girder instance? Or does this "faux-importing off a girder object" approach seem fine to you?

There is an [unanswered question](https://github.com/vitejs/vite/discussions/6563) about this so it seems there may not be an analogous solution in vite at this point.

jeffbaumes:
Added the jobs plugin though I am not sure if it is actually fully working since plugin tests are not migrated.

Also added documentation on the [plugin conversion process](https://github.com/girder/girder/pull/3449/files#diff-50a9a6a39c248472c544ed77ad8204d0ebf9c2e2c66889eb525969763bd44a0dR28).

girder/web/client/package.json
```diff
@@ -11,8 +11,6 @@
   },
   "dependencies": {
     "@girder/core": "file:../core",
-    "@girder/girder-plugin-homepage": "file:../../../plugins/homepage/girder_homepage/web_client",
-    "@girder/girder-plugin-jobs": "file:../../../plugins/jobs/girder_jobs/web_client",
```
zachmullen: This is a key change: the `client` package must avoid having to statically include all plugin dependencies, in order to support plugins outside of the `girder/girder` repository.

girder/web/core/tests/server.ts
```diff
@@ -43,7 +43,7 @@ export const setupServer = () => {
 
   test.beforeEach(async ({ page }) => {
     await startCoverage(page);
-    await page.goto(`/?apiRoot=${encodeURIComponent(`http://localhost:${port}/api/v1`)}`);
+    await page.goto(`http://localhost:${port}/`);
```
zachmullen, line 46: This will allow us to remove the `index.html` within `core` altogether if we want, and instead build it as a library. I didn't do that yet since I wasn't sure how it would impact all the testing infrastructure.

girder/web/client/package.json
jeffbaumes: @zachmullen Ah, if I understand it well enough, this is clever. By adding the API endpoint to retrieve plugin source URLs, you are essentially moving plugin code inclusion from being a build time concern to being a runtime concern, obviating the need for updating the `package.json` for a client build for each package.

Essentially this is taking the "remove webroot.mako" that I had done to its logical conclusion. I had used JS and the build system to include (topo-sorted) assets, but yours method is cleaner. Just like I moved getting settings from mako to an endpoint queried at runtime, you also move plugin inclusion (which had been in mako), to an endpoint. This is clean and consistent IMO.

Also I like that there's an explicit list of objects to include in the frontend instead of using glob expressions for js, css.

zachmullen:
Example coverage report https://output.circle-artifacts.com/output/job/28d9e10a-5146-4bf0-8271-31b290fe3eb3/artifacts/0/girder/girder/web/coverage/report/index.html

zachmullen:
### Remaining tasks

- [ ] Swagger page
- [ ] Remove `girder build`
- [ ] Fix Dockerfile
- [ ] Move plugin web tests alongside core web tests
- [ ] Get coverage working on plugin web client files
- [ ] Documentation changes
- [ ] Verify dev server works for core dev
- [ ] Look into use of dev server for plugin development (may not be feasible)

### Remaining menial tasks

- [ ] Convert remaining plugins
- [ ] Convert remaining test specs

jeffbaumes:
>  Look into use of dev server for plugin development (may not be feasible)

@zachmullen I added a small ++ on this task. I don't feel it's feasible to integrate plugin watching with the vite hot-reloading, but I added a command to the homepage and jobs plugins so you can do a `npm run dev` that will perform `vite build --watch` which will rebuild the plugin each time you change a file. This will still require a browser refresh on plugin changes but it's a bit better than needing to perform the rebuild manually each time.

zachmullen:
> This will still require a browser refresh on plugin changes but it's a bit better than needing to perform the rebuild manually each time.

Great! That's what I figured would be the reasonable limit of what we could get.

zachmullen:
We could now consider the tasks in [my previous comment](https://github.com/girder/girder/pull/3449#issuecomment-1756469824) complete.

Note that the test suite has been dramatically slimmed down, in particular the core test suite. That's something we can add to whenever we see fit, and I don't think we need to hold this branch up any longer awaiting more web client test coverage, so long as people agree on the general patterns and tools being used.

As such, I'm marking his for review. We won't merge this into master, but rather an integration branch off of current master.

zachmullen:
For reference (now includes plugins): https://output.circle-artifacts.com/output/job/9d2f8c9d-7a63-46d2-8538-02211ffdbcc6/artifacts/0/girder/girder/web/coverage/report/index.html

PR #3451, opened 2023-06-09
Allow any Setting value to be set via the environment

zachmullen:


girder/models/setting.py
```diff
@@ -83,6 +86,18 @@ def _get(self, key):
         This method is so built in caching decorators can be used without specifying
         custom logic for dealing with the default kwarg of self.get.
         """
+        envKey = f'GIRDER_SETTING_{key.replace(".", "_").upper()}'
+        if envKey in os.environ:
+            value = os.environ[envKey]
+            try:
+                value = json.loads(value)
```
manthey, line 93: I don't know if we care, or we just need to handle it in documentation, but here we want things to be json and in the girder.conf file they need to be pythonic.  This means that via the environment, one sets numbers and booleans via short strings like `6` or `true` (but not `True` or `TRUE`) and strings MUST be double quoted (`"value"`).  A petty alternative, but probably too confusing, would be to use the environment variable's value as a string if it cannot be parsed as json.

Either way, we need to add what we are doing to the docs somewhere.

zachmullen: > A petty alternative, but probably too confusing, would be to use the environment variable's value as a string if it cannot be parsed as json.

That's what this does, by virtue of catching `ValueError` from JSON parsing.

manthey: > That's what this does, by virtue of catching `ValueError` from JSON parsing.

Yes -- I wasn't paying enough attention.

zachmullen:
@manthey documentation added, PTAL.

PR #3454, opened 2023-06-27
Fix an issue with LDAP.

### LINKED ISSUE #3412: LDAP plugin can't find the attribute 'distinguishedName' in search results
```
 [2022-07-13 13:19:36,773] ERROR: 500 Error  
  File "/girder/girder/api/rest.py", line 649, in endpointDecorator
    val = fun(self, path, params)
  File "/girder/girder/api/rest.py", line 1212, in GET
    return self.handleRoute('GET', path, params)
  File "/girder/girder/api/rest.py", line 968, in handleRoute
    val = handler(**kwargs)
  File "/girder/girder/api/describe.py", line 676, in wrapped
    return fun(*args, **kwargs)
  File "/girder/girder/api/v1/user.py", line 116, in login
    user = self._model.authenticate(login, password, otpToken)
  File "/girder/girder/models/user.py", line 159, in authenticate
    'password': password
  File "/girder/girder/events.py", line 292, in trigger
    handler(e)
  File "/usr/local/lib/python3.7/dist-packages/girder_ldap/__init__.py", line 102, in _ldapAuth
    dn = attrs['distinguishedName'][0].decode('utf8')
KeyError: 'distinguishedName'
Additional info:
  Request URL: GET http://localhost:8080/api/v1/user/authentication
  Query string: 
  Remote IP: 172.28.0.1
  Request UID: 2e4af87b-b511-402d-8919-cb330f5fe449 
```

The python-ldap documentations says that for a search_s() meathod it returns the dn as the first item in the tuple and the attributes in the second item. However, the plugin tries to find the dn(distinguishedName) in the attributes. Even though it is not an attribute.

See documentation https://www.python-ldap.org/en/python-ldap-3.3.0/reference/ldap.html under the search methods.

Instead of searching the attributes for the dn. It should just use the first item in the tuple. So instead using the variable entry which is the dn.

I am using: 
pip install girder-ldap==3.1.14
and the docker container girder/girder:3.1.5-py3

Thanks!

manthey:
Responses from an LDAP server via python-ldap don't include disinguishedName.  Rather, the entry that is returned should be used.

Fixes #3412

This was tested locally with the rroemhild/test-openldap docker image acting as an LDAP server.

### DIFF
--- a/plugins/ldap/girder_ldap/__init__.py
+++ b/plugins/ldap/girder_ldap/__init__.py
@@ -99,7 +99,10 @@ def _ldapAuth(event):
             results = conn.search_s(server['baseDn'], ldap.SCOPE_SUBTREE, searchStr, lattr)
             if results:
                 entry, attrs = results[0]
-                dn = attrs['distinguishedName'][0].decode('utf8')
+                try:
+                    dn = attrs['distinguishedName'][0].decode('utf8')
+                except KeyError:
+                    dn = entry
                 try:
                     conn.bind_s(dn, password, ldap.AUTH_SIMPLE)
                 except ldap.LDAPError:


plugins/ldap/girder_ldap/__init__.py
```diff
@@ -99,7 +99,10 @@ def _ldapAuth(event):
             results = conn.search_s(server['baseDn'], ldap.SCOPE_SUBTREE, searchStr, lattr)
             if results:
                 entry, attrs = results[0]
-                dn = attrs['distinguishedName'][0].decode('utf8')
+                try:
+                    dn = attrs['distinguishedName'][0].decode('utf8')
+                except KeyError:
+                    dn = entry
```
zachmullen, line 105: I think a code comment is warranted here to explain that the returned structure depends on the specific LDAP server implementation. Otherwise LGTM.

manthey: Added.

PR #3459, opened 2023-08-29
Add global search feature

bruyeret:
This is the PR for global search feature for [this issue](https://github.com/Kitware/UPennContrast/issues/501) @manthey 
What do you think?


scripts/publicNames.txt
zachmullen, line 1: What's going on with this file?

zachmullen:
If you want to just match any part of a string that's fine, but I'd prefer that that didn't go into core either, since it would require a full table scan.

PR #3463, opened 2023-09-11
Frontend: Add support for unicode in passwords

AlexanderKoehn:
The backend already decodes the auth string with utf8, but the frontend uses btoa, which encodes with latin1. This fixes the mismatch between back- and frontend and allows any unicode chars in the pw.


zachmullen:
The reason the web client tests are failing is that our test harness browser (phantomjs) is unfortunately missing `TextEncoder`. I think it's easy enough to polyfill, I will look into it. Thanks for the PR!

zachmullen:
That worked, thank you for the fix @AlexanderKoehn !

PR #3464, opened 2023-09-12
Creation date in FileListWidget.  Creator on FileInfoWidget.

PaulHax:
VolView's [grider plugin](https://github.com/PaulHax/girder_volview) saves annotations as files under Items.  Adding more information to the file list helps track mutliple annotation versions.

Adds creation date to file list:
image

Adds Creator field to file info modal:
image

### DIFF
--- a/girder/web_client/src/stylesheets/body/itemPage.styl
+++ b/girder/web_client/src/stylesheets/body/itemPage.styl
@@ -37,19 +37,25 @@
   >li
     padding 3px 5px
     border-bottom 1px solid #f2f2f2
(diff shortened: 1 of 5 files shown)


zachmullen:
@PaulHax could you add this change via a plugin, or else make it an off-by-default option that you can pass to the constructor? It's not uncommon for other plugins to add various things into this widget, and taking more of the horizontal space here would be disruptive to them.

PaulHax:
Keeping the common file list simple sounds good.  I'll put in plugin.

PR #3466, opened 2023-09-13
Expose content-disposition in CORS requests

### LINKED ISSUE #3465: Add "content-disposition" to default Access-Control-Expose-Headers
This is useful for CORS downloads and is not in the default allowed set.

zachmullen:
Fixes #3465 

This header is not in the standard set of exposed headers on CORS responses.


zachmullen:
Merging despite the failure from the ansible test as it is clearly unrelated.

Issue #3467, opened 2023-09-15
Move attached files between assetstores


manthey:
We have a function to move files between assetstores.  This doesn't support moving attached files, only files that are actually owned directly by an item.

Probably the correct behavior is to modify the upload record in `moveFileToAssetstore` to reflect the attached status.

Further, there is a major performance penalty in the movement of files because we build up the chunk to upload by repeatedly adding data blocks from the small downloaded chunks.  We should, instead, keep a list of these chunks and join them once for upload rather than iteratively adding binary string together.

willdunklin:
To quote my comment from the import-tracker plugin PR (https://github.com/DigitalSlideArchive/import-tracker/pull/19#issuecomment-1724046868)

> That looks good, I've been able to essentially recreate a version of `moveFileToAssetstore` for attached files. Using the lower level APIs there also expose control over the file metadata so maintaining the `created` field is automatically encapsulated (by the `upload.update(...)` line).
> 
> ```python
> def move_meta_file(file, assetstore):
>     parent = Item().findOne({'_id': file['attachedToId']})
> 
>     chunk = None
>     try:
>         for data in File().download(file, headers=False)():
>             if chunk is not None:
>                 chunk += data
>             else:
>                 chunk = data
>     except Exception as e:
>         return {'error': f'Exception downloading file: {e}'}
> 
>     upload = Upload().uploadFromFile(
>         obj=io.BytesIO(chunk), size=file['size'], name=file['name'],
>         parentType=file['attachedToType'], parent=parent,
>         mimeType=file['mimeType'], attachParent=True, assetstore=assetstore)
>     upload.update({k: v for k, v in file.items() if k != 'assetstoreId'})
>     upload = File().save(upload)
> 
>     return upload
> ```
> 
> This is modeled after the large image plugin's code for manipulating attached files https://github.com/girder/large_image/blob/master/girder/girder_large_image/rest/tiles.py#L1538-L1547

I believe this version addresses the chunking issue you mentioned, correct? Or am I misinterpreting the bottleneck at hand here?

manthey:
The chunking issue is the add in this loop:
```
    chunk = None
    try:
        for data in File().download(file, headers=False)():
            if chunk is not None:
                chunk += data
            else:
                chunk = data
```
[...]

PR #3469, opened 2023-09-18
Improve Upload assetstore move speed

willdunklin:
Addresses chunking slowdowns present when moving files between assetstores.
Related to: #3467


willdunklin:
On my local testing setup, moving ~700mb of files from one filesystem assetstore to another took ~27 seconds before the change and ~4.8 afterwards

manthey (approved):

PR #3470, opened 2023-09-21
Move attached files between assetstores

willdunklin:
Fixed edge cases where attached files wouldn't be transferred properly by `moveFileToAssetstore`.

Addresses #3467

### DIFF
--- a/girder/models/upload.py
+++ b/girder/models/upload.py
@@ -183,16 +183,20 @@ class Upload(Model):
         if 'fileId' in upload:  # Updating an existing file's contents
             file = File().load(upload['fileId'], force=True)
 
+            if upload.get('attachParent'):
+                item = Item().load(upload['parentId'], force=True)
+            else:
+                item = Item().load(file['itemId'], force=True)
+
+            File().propagateSizeChange(item, upload['size'] - file['size'])
+
             # Delete the previous file contents from the containing assetstore
             assetstore_utilities.getAssetstoreAdapter(
                 Assetstore().load(file['assetstoreId'])).deleteFile(file)
 
-            item = Item().load(file['itemId'], force=True)
-            File().propagateSizeChange(item, upload['size'] - file['size'])
-
             # Update file info
             file['creatorId'] = upload['userId']
-            file['created'] = datetime.datetime.utcnow()
+            file['updated'] = datetime.datetime.utcnow()
             file['assetstoreId'] = assetstore['_id']
             file['size'] = upload['size']
             # If the file was previously imported, it is no longer.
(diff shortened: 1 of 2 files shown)


girder/models/upload.py
```diff
@@ -183,16 +183,20 @@ def finalizeUpload(self, upload, assetstore=None):
         if 'fileId' in upload:  # Updating an existing file's contents
             file = File().load(upload['fileId'], force=True)
 
+            if upload.get('attachParent'):
+                item = Item().load(upload['parentId'], force=True)
```
manthey, line 187: The attached parent isn't necessarily an item.

girder/models/upload.py
```diff
@@ -183,16 +183,20 @@ def finalizeUpload(self, upload, assetstore=None):
...
+            else:
+                item = Item().load(file['itemId'], force=True)
+
+            File().propagateSizeChange(item, upload['size'] - file['size'])
```
manthey, line 191: Do attached files count toward item sizes?  I didn't think so (but don't know for sure).  In which case, this should only be in the non-attached half of the if/else statement.

plugins/user_quota/girder_user_quota/quota.py
```diff
@@ -258,7 +258,10 @@ def _getBaseResource(self, model, resource):
                 return None, None
         if model == 'file':
             model = 'item'
-            resource = Item().load(id=resource['itemId'], force=True)
+            if 'attachedToId' in resource:
+                resource = Item().load(id=resource['attachedToId'], force=True)
```
manthey, line 262: I don't think we track the size usage of the attached files (and this isn't necessarily an item).

girder/models/upload.py
manthey: I checked -- attached files don't count toward the parent's size (and therefore are not subject to the user quota either).

willdunklin: I was questioning this too, is there a simple way to check if they affect it?

plugins/user_quota/girder_user_quota/quota.py
willdunklin: Alright, I can try to exit gracefully when an attached item is logged here. I'm somewhat unfamiliar with the design behind the user quota plugin, but when I tried to exit similarly in the past it got unhappy. I'll give it a shot and see if there are errors

willdunklin: Turned out this wasn't the issue I was thinking of. The new code works and avoids quota logic for attached files entirely

girder/models/upload.py
```diff
@@ -291,22 +292,31 @@ def createUploadToFile(self, file, user, size, reference=None,
...
-            'userId': user['_id'],
-            'fileId': file['_id'],
-            'assetstoreId': assetstore['_id'],
-            'size': size,
-            'name': file['name'],
-            'mimeType': file['mimeType'],
-            'received': 0
-        }
+        if 'attachedToId' in file:
+            parent = Item().findOne({'_id': file['attachedToId']})
```
manthey, line 302: The `attachedToType` isn't necessarily an item, so this should probably be `parent = ModelImporter.model(file['attachedToType']).findOne({'_id': file['attachedToId']})`

willdunklin:
I tried to do a bit of digging into why the CI was failing (test 25 of server-legacy-test), but I got lost in the rabbit hole of replicating the error before I made any progress on understanding it. Is the error/it's cause intelligible to you @manthey ?

girder/models/upload.py
```diff
@@ -183,16 +184,17 @@ def finalizeUpload(self, upload, assetstore=None):
...
             assetstore_utilities.getAssetstoreAdapter(
                 Assetstore().load(file['assetstoreId'])).deleteFile(file)
 
-            item = Item().load(file['itemId'], force=True)
-            File().propagateSizeChange(item, upload['size'] - file['size'])
-
             # Update file info
             file['creatorId'] = upload['userId']
-            file['created'] = datetime.datetime.utcnow()
+            file['updated'] = datetime.datetime.utcnow()
```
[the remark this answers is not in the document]
willdunklin: Alright, that's too bad it didn't work. Thank you so much for helping track that down, I'll revert the change there

manthey (approved):

PR #3472, opened 2023-09-28
Update swagger-ui from version 2.x to 5.x.

manthey:
This has a few components:

- Swagger now publishes a bundle with no extra dependencies, greatly reducing the number of packages.  This means we import less stuff explicitly.

- Swagger 2 to swagger 3 required changing how the swagger ui was initiated and some of the default look and options.  I've tried to make some css adjustments and option picks to have as little difference with our swagger 2 deployment as reasonable.  For instance, how the "try it out" block appears is quite different, and the default sort was different.  Swagger 3 through 5 have no directly breaking changes as far as we are concerned.

- The OpenAPI spec requires specifying how security is handled (via our Girder-Token header).  This has been added to the describe response. The way that the token gets added to swagger requests changed slightly.

- The test for swagger is DISABLED.  Swagger 3 and above use a more modern version of React which requires certain browser features PhantomJS does not have.  Theoretically, we could use babel to transpile the swagger-ui-bundle code to work with PhantomJS, but it still misses a huge number of features.  Some of these can be polyfilled, but even adding all of the polyfills from polyfill.io I could not get it to work directly; PhantomJS doesn't identify internal features in the way React expects.  While we could possibly hack around this, this rapidly becomes a test of something we don't actually use.  Rather, when we update Girder's test harness, we should reenable this test in some form.

- This eliminates a critical security CVE and reduces some others.


girder/web_client/src/stylesheets/apidocs/apidocs.styl
```diff
@@ -47,3 +48,30 @@ body
...
+      padding 8px 8px 8px 0
+  .markdown p
+    margin 0.2em auto
+  input[type=email], input[type=file], input[type=password], input[type=search], input[type=text], textarea
+    margin 0
+  .scheme-container, .information-container.wrapper
+    display none
+  .parameter__name
+    font-size 13px
+    font-family monospace
```
zachmullen, line 77: I don't think we should go out of our way to preserve the old look and feel of the previous swagger-ui version, so long as all the capabilities are preserved.

zachmullen (changes requested):

girder/web_client/src/stylesheets/apidocs/apidocs.styl
manthey: Sure, I can remove this.  The main difference is that the newer style uses vastly more whitespace which makes it (in my opinion) harder to use since you end up scrolling more.  But that trend was all the rage the year swagger 3 came out.

zachmullen (approved):

PR #3480, opened 2023-10-26
Switch the stylus linter to avoid warnings

manthey:
The stylint package hasn't received updates in years and has a security warning on one of its dependencies; the stylelint with stylus plugin is actively maintained.

Update to testing on node 14 so we can use stylelint.

Use the default linting style for styl files.

### DIFF
--- a/girder/web_client/src/stylesheets/widgets/userOtpManagementWidget.styl
+++ b/girder/web_client/src/stylesheets/widgets/userOtpManagementWidget.styl
@@ -1,5 +1,5 @@
 #g-account-tab-otp
-// Included parent ID to avoid using !important for style overrides
+  // Included parent ID to avoid using !important for style overrides
   .g-user-settings-container
     background none
     border none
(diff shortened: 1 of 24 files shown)


manthey:
This will resolve a warning that dependabot identified.

girder/web_client/src/stylesheets/widgets/userOtpManagementWidget.styl
```diff
@@ -146,7 +146,7 @@
               &:before
                 display none
 
-@media screen and (max-width:960px)
+@media screen and (max-width960px)
```
zachmullen, line 149: Is this correct syntax?

zachmullen (approved):
Generally LGTM, assuming that media query syntax is actually valid

girder/web_client/src/stylesheets/widgets/userOtpManagementWidget.styl
manthey: Hmm  -- I just ran with the --fix option.  Let me check.

manthey: Oddly, it is valid.  But I made a push to make it more consistent anyway.

PR #3484, opened 2023-12-05
Mark auth cookie as httponly

manthey:


zachmullen:
I thought we read this value in JavaScript and use it to set the `Girder-Token` header?

manthey:
> I thought we read this value in JavaScript and use it to set the `Girder-Token` header?

You are right.  We were asked for security purposes if we could mark cookies as HttpOnly.

zachmullen:
@manthey now just set `GIRDER_SETTING_CORE_HTTP_ONLY_COOKIES=false` in your environment.

manthey:
This works for me, but I think we should add something to the docs.

zachmullen:
What doc section were you thinking? I can't find anywhere that we enumerate all the system settings, nor any sections specifically devoted to security configuration.

manthey:
> What doc section were you thinking? I can't find anywhere that we enumerate all the system settings, nor any sections specifically devoted to security configuration.

I was imagining in docs/configuration.rst.  Most of the other settings are discoverable because we expose them in the UI.  Maybe instead we have a link to the settings.py file and just say that many settings are enumerated there (but it doesn't explain them).

PR #3485, opened 2023-12-05
Cache the parsed JSON body on cherrypy.request

zachmullen:
This will make multiple calls to `getBodyJson` work, where before they would throw an error due to the request stream already being consumed.


bruyeret:
Maybe `allowConstants` being ignored if `getBodyJson` is called twice could be documented?

manthey (approved):

PR #3495, opened 2024-01-24
Add an assetstore method to get file size

psavery:
This adds a `getFileSize()` method on the assetstore, which allows assetstores to update a file size, if needed. If the file size changes, the `File` model will update it.

We plan to use this with a DICOMweb assetstore adapter so that we do not need to compute the file size right away, but only when it is needed.

### DIFF
--- a/girder/models/file.py
+++ b/girder/models/file.py
@@ -450,8 +450,15 @@ class File(acl_mixin.AccessControlMixin, Model):
         :param file: The file.
         :type file: dict
         """
-        # TODO: check underlying assetstore for size?
-        return file.get('size', 0), 0
+        fixes = 0
+        if file.get('assetstoreId'):
+            size = self.getAssetstoreAdapter(file).getFileSize(file)
+            if size != file.get('size', 0):
+                file['size'] = size
+                self.update({'_id': file['_id']}, {'$set': {'size': size}})
+                fixes += 1
+
+        return file.get('size', 0), fixes
 
     def open(self, file):
         """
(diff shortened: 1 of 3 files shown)


zachmullen (approved):

zachmullen:
Should we use this method everywhere that we reference `file['size']`?

psavery:
> Should we use this method everywhere that we reference `file['size']`?

Well, right now, for the DICOMweb assetstore adapter, we are keeping `file['size']` to be `None` at first, indicating that it hasn't been determined/computed yet. But when it is needed (for range requests or fuse mounting), the `File().updateSize()` method is called (which subsequently calls `getFileSize()`), and it is then computed.

So I think we should avoid calling `getFileSize()` unless we want to compute it. Maybe another name would be clearer, such as `computeFileSize()`?

zachmullen:
Hm, if `file['size']` is `None`, then `file.get('size', 0)` will yield `None` rather than `0`. Is that OK?

psavery:
We are using `None` to indicate that it hasn't been computed yet. We were previously using `0`, but some files might actually have a size of `0`, and in the Girder UI, it displays the size as `0 bytes`.

If it is `None`, then the file size isn't displayed at all in the Girder UI. So far, `None` has been working fine for us, except when it is needed (for range requests or fuse mounting), but we plan to compute it in those cases.

manthey:
> I can't remember right this second, but it seems to me that all `0` size files should not have an assetstore.

I think that is only true on _uploaded_ files, not _imported_ files.

PR #3502, opened 2024-02-05
Delete ansible role

zachmullen:


brianhelba:
Once this is merged, https://github.com/girder/ansible-role-girder can be deleted (an archive shouldn't be necessary, since it's just a mirror).

Perhaps find a way to clean up https://galaxy.ansible.com/ui/standalone/roles/girder/girder/install/ too.

PR #3505, opened 2024-02-12
Fix oauth bug

mitchklusty:
Add decorator to the getMe function which is used with OAuth to sign in. Cookie authentication is needed, but the correct attribute is not being set without this fix.

### DIFF
--- a/girder/api/v1/user.py
+++ b/girder/api/v1/user.py
@@ -74,6 +74,7 @@ class User(Resource):
         Description('Retrieve the currently logged-in user information.')
         .responseClass('User')
     )
+    @access.user(cookie=True)
     def getMe(self):
         return self.getCurrentUser()
 
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -182,7 +182,7 @@ def getCurrentUser(returnToken=False):
     if event.defaultPrevented and len(event.responses) > 0:
         return event.responses[0]
 
-    token = getCurrentToken()
+    token = getCurrentToken(True)
 
     def retVal(user, token):
         setCurrentUser(user)
--- a/girder/api/rest.py
+++ b/girder/api/rest.py
@@ -182,7 +182,7 @@ def getCurrentUser(returnToken=False):
     if event.defaultPrevented and len(event.responses) > 0:
         return event.responses[0]
 
-    token = getCurrentToken(True)
+    token = getCurrentToken()
 
     def retVal(user, token):
         setCurrentUser(user)


girder/api/v1/user.py
```diff
@@ -74,6 +74,7 @@ def getUser(self, user):
         Description('Retrieve the currently logged-in user information.')
         .responseClass('User')
     )
+    @access.user(cookie=True)
```
zachmullen, line 77: Rather than put this as a new decorator, please add it into the existing `access.public` decorator above.

zachmullen (changes requested):

girder/api/v1/user.py
mitchklusty: The `/me` route that is used when signing in with OAuth returns null as is. This is because in the `getCurrentToken()` function of [/api/rest.py]( https://github.com/girder/girder/blob/master/girder/api/rest.py), the `'girderToken'` field is set in `cherrypy.request.cookie`, but without this decorator, the `allowCookie` variable is False, so the elif statement:
```
elif allowCookie and 'girderToken' in cherrypy.request.cookie:
    tokenStr = cherrypy.request.cookie['girderToken'].value
```
does not allow `tokenStr` to be set, returning `None` as the token which in turn causes the `getCurrentUser()` function called by the `/me` route. 

I am not sure if there is a problem with making this route use cookie authentication, but it currently makes use of cookies in the case of using Microsoft OAuth at least (and I would assume other OAuth services as well), but it does not have the decorator to be allowed to use them. If there is another approach that should be taken to resolve this can you explain it? This was the most straightforward solution I could find by looking through the code.

zachmullen: Ah, are you by chance using the new `HttpOnly` cookie option? It's possible that using that has broken the OAuth plugin's behavior.

mitchklusty: As far as I can tell this is not set anywhere. Should the OAuth plugin not be using the `girderToken` cookie for authentication?

zachmullen: It does use the cookie, but not in subsequent HTTP requests. What happens is that we read the cookie in JavaScript at the time of application startup, and then use that cookie value to set the `Girder-Token` header in subsequent requests.

Looking through the code, I realize exactly what broke the OAuth plugin (this is a gap in our testing, as it's difficult to end-to-end test the OAuth flow): https://github.com/girder/girder/commit/ee433279e4c5d316b2576505be0ff3c063f9d1ee#diff-cdd08baf1bebeef9dccadb5544ed935ce2520b46185f7b7e2d0b37c51e5daf51R47

I'll put in a fix for the OAuth plugin that won't depend on JS-readable cookies, since we are setting all cookies to `HttpOnly` in the next major version of Girder. Thanks for identifying the bug!

zachmullen:
@mitchklusty hopefully #3508 has fixed the issue. Thanks again.

PR #3508, opened 2024-02-13
Do not rely on JavaScript-readable cookies for OAuth login

zachmullen:
Instead, we send a query parameter on successful OAuth login completion, and the client reads and removes it from the URL.

Refs #3505


manthey (approved):
This looks fine and I have run the tests locally.

PR #3511, opened 2024-02-14
Fix readthedocs search

manthey:


zachmullen:
That integration failure is what I also had to fix on my branch yesterday. It just needs a cache key bump.

manthey:
> That integration test failure is what I also had to fix on my branch yesterday. It just needs a cache key bump.

Does that mean we should be including additional files in .circleci/generatePyEnvChecksum.sh?

zachmullen:
I'm not sure, I didn't dig into the underlying cause.

PR #3521, opened 2024-03-14
Fix passing integer values to diskcache options

manthey:
#3520 fixed passing integers to diskcache, but broke passing booleans.  This fixes that.

### DIFF
--- a/girder/cli/mount.py
+++ b/girder/cli/mount.py
@@ -96,7 +96,7 @@ class ServerFuse(fuse.Operations):
         for key in list((cacheopts or {}).keys()):
             if key.startswith('diskcache'):
                 value = cacheopts.pop(key)
-                if value.strip().isdigit():
+                if str(value).strip().isdigit():
                     value = int(value)
                 key = key[len('diskcache'):].lstrip('_')
                 use = True if use is None else use


girder/cli/mount.py
```diff
@@ -96,7 +96,7 @@ def _configure_disk_cache(self, cacheopts):
         for key in list((cacheopts or {}).keys()):
             if key.startswith('diskcache'):
                 value = cacheopts.pop(key)
-                if value.strip().isdigit():
+                if str(value).strip().isdigit():
```
zachmullen, line 99: It'd be nice to use a `dataclass` for the `cacheopts` argument, and if any of the options are algebraic sum types, we could represent them with their own special types.

manthey: I didn't want to re-enumerate the possible options for diskcache.  If diskcache had such a dataclass, that would be great.

manthey: Looking at diskcache's source, it looks like `diskcache.core.DEFAULT_SETTINGS` plus `directory` and `timeout` is the list of parameters we could possible deal with.  But they aren't explicitly typed, so you'd be doing type inspection for those to determine what we should cast to...

manthey: I'll merge this for now; I may try to refactor to be smarter about this on a future PR.

zachmullen: Sure thing, don't mean to hold this up, was just musing :)

PR #3527, opened 2024-04-01
Allow passage of arbitrary parameters for assetstore/{id}/import route

willdunklin:
Use girder route `param` kwarg handling to pass arbitrary assetstore `importData` parameters. This allows plugins that define custom assetstores to reuse the `/assetstore/{id}/import` route by allowing for the passage of plugin-specific parameters to an assetstore adapter.

### DIFF
--- a/girder/api/v1/assetstore.py
+++ b/girder/api/v1/assetstore.py
@@ -132,10 +132,13 @@ class Assetstore(Resource):
         .errorResponse('You are not an administrator.', 403)
     )
     def importData(self, assetstore, importPath, destinationId, destinationType, progress,
-                   leafFoldersAsItems, fileIncludeRegex, fileExcludeRegex):
+                   leafFoldersAsItems, fileIncludeRegex, fileExcludeRegex, **kwargs):
         user = self.getCurrentUser()
         parent = ModelImporter.model(destinationType).load(
             destinationId, user=user, level=AccessType.ADMIN, exc=True)
+
+        # Capture any additional parameters passed to route
+        extraParams = kwargs.get('params', {})
 
         with ProgressContext(progress, user=user, title='Importing data') as ctx:
             return self._model.importData(
@@ -143,6 +146,7 @@ class Assetstore(Resource):
                     'fileIncludeRegex': fileIncludeRegex,
                     'fileExcludeRegex': fileExcludeRegex,
                     'importPath': importPath,
+                    **extraParams
                 }, progress=ctx, user=user, leafFoldersAsItems=leafFoldersAsItems)
 
     @access.admin


zachmullen:
This should be safe since it's an admin-only endpoint.

zachmullen (approved):

Xarthisius:
FWIW I do the same thing in a plugin by binding to `"rest.post.assetstore/:id/import.before"`. It works great with vanilla girder.

```
@rest.boundHandler
def import_sem_data(self, event):
    params = event.info["params"]
    if params.get("dataType") not in ("sem", "pdv"):
        logger.warning("Importing using default importer")
        return
    ...

def load(info):
    events.bind("rest.post.assetstore/:id/import.before", "sem_viewer", import_sem_data)
```

willdunklin:
@Xarthisius I see what you're doing but I believe this PR is addressing something slightly different. This functionality allows for plugin-defined assetstores to reuse the existing polymorphism we expect from the `/assetstore/{id}/import` endpoint, which dynamically chooses which assetstore adapter to use based on the assetstore associated with the `id`. Extending the functionality here allows us to reuse the same general import patterns the code already supports.
Are there any benefits to using the event binding method you mentioned though?

Xarthisius:
> @Xarthisius I see what you're doing but I believe this PR is addressing something slightly different. This functionality allows for plugin-defined assetstores to reuse the existing polymorphism we expect from the `/assetstore/{id}/import` endpoint, which dynamically chooses which assetstore adapter to use based on the assetstore associated with the `id`. Extending the functionality here allows us to reuse the same general import patterns the code already supports.

Consider this example, which I think preserves polymorphism you're talking about:

```
@rest.boundHandler    
def import_with_extra_kwargs(self, event):
    if some_exit_condition:
        return
    params = event.info["params"]
    progress = toBool(params.get("progress", "false"))
    user = self.getCurrentUser()
    assetstore = self._model.load(event.info["id"])
    parent = self.model(params["destinationType"]).load(
        params["destinationId"], user=user, level=AccessType.ADMIN, exc=True)

    with ProgressContext(progress, user=user, title='Importing data') as ctx:
        self._model.importData(
        assetstore, parent=parent, parentType=destinationType, params={
            'fileIncludeRegex': params.get("fileIncludeRegex"),
            'fileExcludeRegex': params.get("fileExcludeRegex"),
            'importPath': params["importPath"],
            'myExtraKwarg1': ...
            'myExtraKwarg2': ...
        }, progress=ctx, user=user, leafFoldersAsItems=params["leafFoldersAsItems"])

    event.preventDefault().addResponse(None)
```

>  Are there any benefits to using the event binding method you mentioned though?

In my case I'm doing a bit more in there than just passing through extra kwargs. I guess that's the major benefit: more control.

willdunklin:
Right, I see... So this method would require plugins to explicitly define their own parameter extensions. The polymorphism I'm talking about is in that `self._model.importData` where we choose which assetstore adapter to use (the vanilla girder adapters being s3, filesystem, etc. which happen to share the same parameter signatures):
```python
# models/assetstore.py
def importData(self, assetstore, parent, parentType, params, progress, user, **kwargs):
        """
        Calls the importData method of the underlying assetstore adapter.
        """
        adapter = assetstore_utilities.getAssetstoreAdapter(assetstore)
        return adapter.importData(
            parent=parent, parentType=parentType, params=params,
            progress=progress, user=user, **kwargs)
```
We expect plugin-defined assetstores to define their own adapter, but not to necessarily adhere to the hard-coded params that s3/filesystem accept. Making the girder API more general allows plugin-defined assetstores to tap into the same simple generality afforded to s3/filesystem ones. The point of the changes I made is to not have to redefine the API behavior for each assetstore plugin

willdunklin:
@zachmullen Would we be good to merge this?

zachmullen:
Yes, merge away

zachmullen:
Sorry, I thought everyone who could make branches in this repo had merge permission, but I guess not. All set.

PR #3530, opened 2024-04-10
Update Google Analytics plugin to use up to date API

willdunklin:
Removed old google analytics (analytics.js) API which has been [deprecated](https://support.google.com/analytics/answer/11583528).


plugins/google_analytics/girder_google_analytics/web_client/views/ConfigView.js
```diff
@@ -56,6 +56,8 @@ var ConfigView = View.extend({
             },
             error: null
         }).done(() => {
+            location.reload();
```
zachmullen, line 59: What's this for?

willdunklin: This clears the previous js bundle from google. If we change the tag we're tracking, we need to request a fresh js bundle from `main.js:initGoogleAnalytics`

willdunklin: @zachmullen is this an issue?

zachmullen: I'd say it's generally an anti-pattern to do a hard refresh in a single-page app. Is it not sufficient to re-invoke `initGoogleAnalytics`? Or would the presence of the previous state corrupt the page in some way?

zachmullen (approved):

plugins/google_analytics/girder_google_analytics/web_client/views/ConfigView.js
willdunklin: I was initially worried about the prospect of having multiple competing results from `initGoogleAnalytics`'s `$.getScript`, but after testing it further seems like it won't be an issue for usage.

willdunklin:
@zachmullen seems like the CI errors here are outside the scope of this PR's changes (unless I'm mistaken). Are you able to merge this? We have a customer waiting on these changes

PR #3540, opened 2024-05-22
Add saveMany function to the model

bruyeret:
Edit one of the tests to use saveMany() instead of several save()

This is the endpoint to optimize calls to `save` in a loop with validation and events @manthey

### DIFF
--- a/girder/models/model_base.py
+++ b/girder/models/model_base.py
@@ -7,7 +7,7 @@ import re
 
 from bson.objectid import ObjectId
 from bson.errors import InvalidId
-from pymongo.errors import WriteError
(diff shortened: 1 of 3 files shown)


zachmullen:
I'd rather not introduce this into core, because if it starts being used throughout Girder core, then any downstreams who were previously relying on receiving model event signals for their model would no longer receive them unless they updated their code to also listen on these new events.

I think when optimizations like this are needed downstream, it's best to reach down a layer directly to the pymongo API for the model in your downstream code, e.g. it's totally fine and supported to do:

```python
SomeModel().collection.insert_many(...)
```

bruyeret:
Oh ok I get it

PR #3545, opened 2024-06-23
Improve uploads

finetjul:


zachmullen (approved):

zachmullen (changes requested):
The change LGTM; it looks like we'll need to modify our test assertion to match the new behavior.

girder/utility/filesystem_assetstore_adapter.py
```diff
@@ -103,6 +108,15 @@ def __init__(self, assetstore):
...
             logger.error('Could not write to assetstore root: %s', self.assetstore['root'])
 
+        unavailable = False
+        try:
+            unavailable = getattr(self, 'unavailable')
+        except AttributeError:
+            pass
+        finally:
+            if not unavailable:
+                FilesystemAssetstoreAdapter.initialized = True
```
zachmullen, line 118: @finetjul  Rather than adding this logic, I think we should just move the `mkdir` call to only be where it's actually needed. Additionally, we can make `tempDir` a `property` of the class, and thereby remove this `__init__` method altogether.

PR #3554, opened 2024-07-12
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

*Inspired* by https://github.com/dandi/dandi-archive/pull/1882#issuecomment-2224252753 - no hurt feelings if not interested.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within setup.cfg
+---
+name: Codespell
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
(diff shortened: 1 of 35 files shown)


zachmullen (changes requested):
Thanks for your contribution @yarikoptic . Could you please make it so that this only runs on `workflow_dispatch` rather than pushes and PRs? Then I'd be fine to merge this.

yarikoptic:
Done, although IMHO that diminishes its value considerably.

zachmullen (approved):

PR #3555, opened 2024-07-15
Include timezone info in datetime objects

Xarthisius:
Should be a noop, but it fixes a lot of warnings I get during tests with py3.12. Nice write-up about why: https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated

### DIFF
--- a/girder/api/v1/notification.py
+++ b/girder/api/v1/notification.py
@@ -1,7 +1,7 @@
 import cherrypy
 import json
 import time
-from datetime import datetime
+from datetime import datetime, timezone
 
 from ..describe import Description, autoDescribeRoute
 from ..rest import Resource, disableAuditLog, setResponseHeader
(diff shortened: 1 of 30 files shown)


girder/api/v1/notification.py
```diff
@@ -70,7 +70,7 @@ def stream(self, timeout, params):
         setResponseHeader('Cache-Control', 'no-cache')
         since = params.get('since')
         if since is not None:
-            since = datetime.utcfromtimestamp(since)
+            since = datetime.fromtimestamp(since, timezone.utc)
```
zachmullen, line 73: This changes `since` from being a naive to an aware timestamp. Is that a behavioral change when we pass it into mongodb queries below?

Xarthisius: By my understanding of https://pymongo.readthedocs.io/en/stable/examples/datetimes.html the answer would be "no, there shouldn't be any behavioral change". I missed conversion to tz_aware during read from database. I will fix it shortly.

zachmullen:
Just to be safe, could you target this against `v4-integration` rather than `master`?

Xarthisius:
@zachmullen anything else I should do before merging?

zachmullen (approved):

zachmullen:
LGTM

PR #3273, opened 2020-06-16
Proof-of-concept implementation of CLI list command

Erotemic:
There have been a few times where I've wished I could use the CLI to list or describe the data in a folder / collection / item / file. 

I have a basic implementation of this feature. It works well enough, but it could be better. 

I''ve currently tested with the following commands to list items in the VIAME girder collection: 

```python
    girder-client --api-url https://data.kitware.com/api/v1 list 58b747ec8d777f0aef5d0f6a

    girder-client --api-url https://data.kitware.com/api/v1 list 58c49f668d777f0aef5d7960

    girder-client --api-url https://data.kitware.com/api/v1 list 5c423a5f8d777f072b0ba58f
    
    girder-client --api-url https://data.kitware.com/api/v1 list 5dd3181eaf2e2eed3505827c

    girder-client --api-url https://data.kitware.com/api/v1 list 5aac22638d777f068578d53c
```

This results in the following output:

```
(py38) joncrall@Ooo:~/code/girder$     girder-client --api-url https://data.kitware.com/api/v1 list 58b747ec8d777f0aef5d0f6a
Listing collection: 58b747ec8d777f0aef5d0f6a - VIAME
=== folder ===
ID                       TYPE   NAME                    
5dfcf2efaf2e2eed35a52a19 folder 3d_models
5c82ca1c8d777f072b8a85f9 folder annnotations
58d4c1178d777f0aef5d8920 folder binary_releases
5ba3d4348d777f06b9fee5b3 folder dependencies
58c49f668d777f0aef5d7960 folder detection_models
58e5149b8d777f0aef5d8d76 folder documentation
58c49f708d777f0aef5d7961 folder example_images
5b6efbb18d777f06857c32a7 folder hidden
5d362147877dfcc902244b0b folder jenkins_status
5aac22638d777f068578d53c folder other
595d10518d777f16d01e99cb folder videos
(py38) joncrall@Ooo:~/code/girder$ 
(py38) joncrall@Ooo:~/code/girder$     girder-client --api-url https://data.kitware.com/api/v1 list 58c49f668d777f0aef5d7960
Parent collection: 58b747ec8d777f0aef5d0f6a - VIAME
Listing folder: 58c49f668d777f0aef5d7960 - detection_models
=== folder ===
ID                       TYPE   NAME                    
5c82be098d777f072b8a66d8 folder microsoft
5dd3181eaf2e2eed3505827c folder netharn
5c82be028d777f072b8a66ae folder xnor
=== item ===
[...]

PR #3556, opened 2024-08-10
CLI list command (v2)

Erotemic:
Even though I pushed to the same branch, it looks like my old PR didn't update? Github seems to want me to open a new one. 

This is a new version of https://github.com/girder/girder/pull/3273

### DIFF
--- a/clients/python/girder_client/_jsonemitter.py
+++ b/clients/python/girder_client/_jsonemitter.py
@@ -1,0 +1,252 @@
+"""
+Helper functions used by the CLI. Currently this only defines
+:class:`_JSONEmitter`, which we test here.
+
+Example:
+    >>> from girder_client.util import _JSONEmitter
+    >>> import io
+    >>> self = _JSONEmitter(stream=io.StringIO())
+    >>> self.start_dict()
+    >>> self.setitem('version', '1.0.0')
+    >>> self.start_subcontainer('info')
+    >>> self.start_list()
+    >>> self.append({})
+    >>> self.append([{}])
+    >>> self.append({})
+    >>> self.end_list()
+    >>> self.end_dict()
+    >>> text = self.stream.getvalue()
+    >>> # Test that we decode correctly
+    >>> import json as json
+    >>> print(json.loads(text))
+    {'version': '1.0.0', 'info': [{}, [{}], {}]}
+
+Example:
+    >>> from girder_client.util import _JSONEmitter
+    >>> import io
+    >>> self = _JSONEmitter(stream=io.StringIO())
+    >>> self.start_dict()
+    >>> self.setitem('key1', 'value1')
+    >>> self.start_subcontainer('subdict1')
+    >>> self.start_dict()
+    >>> self.end_dict()
+    >>> #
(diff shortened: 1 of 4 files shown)


zachmullen:
@Erotemic given that this repo is in legacy / maintenance mode, I don't think this will be merged in this repo. It contains quite a lot of code and I can't really justify the cost of maintaining it.

The nice thing about `girder-client` is that it's also a stable python library, so downstreams can write their own scripts for things like this. If this functionality is useful to a wider audience beyond just your use case, it could be published as its own package.

One thing that is missing from this repository is the ability to extend the Girder CLI via package entry points. Adding support for the ability to add new subcommands to the `girder-client` command via an entry point is something I would consider adding to this repository.

Erotemic:
Is girder itself in legacy / maintenance mode or just the Python client? I'm confused about the status here. 
 
In either case the README for the particular component should be updated to make this clear. Now that you mention it, I recall having some conversation where someone mentioned something about legacy / maintenance only code.

If it is just the second case, then having some command line way to interact with the API is important. It would make sense for such a client to exist in a separate repo outside of girder itself.

What I need is a clear, concise, and programmatic way to demonstrate exploring and interacting with a dataset. This means I need to be able to tell a user to `pip install <something>` and then run these series of bash commands. I could just maintain a such a package myself, but it seems silly given that girder is a Kitware thing and the existing CLI is lacking the ability to explore contents of a folder. This is moot if girder itself is being deprecated.

zachmullen:
> Is girder itself in legacy / maintenance mode or just the Python client?

Girder itself. We are not recommending starting any new projects with this framework. It underpins Histomics, DIVE, and other important applications, so we are continuing to maintain it, but are not looking to add new features unless truly necessary or else very low maintenance cost.

It probably makes sense to put this status in the README, but we haven't gotten around to doing that and messaging it properly.

Erotemic:
Oh, that is news to me. I'm curious as to the reason, or if there is a new alternative we are recommending instead?

Does this mean data.kitware.com is also not recommended anymore as well?

zachmullen:
> or if there is a new alternative we are recommending instead?

In terms of frameworks that we use for the sorts of tasks Girder was built for, yes, we now choose [Resonant](https://github.com/kitware-resonant/) by default, though my real recommendation would be speaking to us if you are wanting to spin up a project in this space and we can try to steer you toward the most appropriate technology.

> Does this mean data.kitware.com is also not recommended anymore as well?

I'm happy to talk offline about this specific instance.

Erotemic:
@manthey even though this repo is in mostly maintenance mode, I see we are still making releases on pypi, and on the VIAME project we are still making heavy use of girder and having some way to navigate around via the CLI would help me out a bit.

I could keep porting this branch around every time I go back to the project, but having this in the pypi client package for the next release would save me a good amount of time and effort.

manthey:
@Erotemic Girder very much is still the core behind HistomicsTK. I will review and we can merge this soon. Thank you.

manthey (approved):

manthey:
@Erotemic Thanks for adding tests.

PR #3562, opened 2024-08-19
Handle importing files and folders whose names end in whitespace

### LINKED ISSUE #3561: Cannot import folder from S3 when the folder name ends in a space

manthey:
Fixes #3561

### DIFF
--- a/girder/utility/abstract_assetstore_adapter.py
+++ b/girder/utility/abstract_assetstore_adapter.py
@@ -456,3 +456,16 @@ class AbstractAssetstoreAdapter:
         :rtype: str
         """
         raise FilePathException('This assetstore does not expose file paths')
+
+    def safeName(self, name):
+        """
+        Girder strips item and folder names of whitespace.  If we have a name
+        that will be affected, repalce the leading and trailing whitespace with
+        underscores.
+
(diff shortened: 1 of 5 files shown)


manthey:
I added a test for the filesystem assetstore but not the s3 assetstore.

zachmullen:
Did we previously have a bijective mapping from Girder items to import source objects? Do we care about what happens if multiple source objects for an import map to the same name in Girder?

manthey:
> Did we previously have a bijective mapping from Girder items to import source objects? Do we care about what happens if multiple source objects for an import map to the same name in Girder?

This doesn't make the situation any worse than it was.  Before, if you had two files called `foo` and `foo ` and were importing where we can reuse existing entries, you would end up with only `foo` and the actual contents would depend on the walking order.  But, if you only had a folder called `foo `, the import would fail because it couldn't determine where the items in the folder should go.

This doesn't change the collisions -- just changes them from collapsed whitespace to replaced underscores, which allows imports to succeed when folders and files start or end with whitespace.  Stripping would work, too, but then additional checks are needed to make sure that a file name isn't _only_ whitespace.

Actually forcing the bijective would be its own issue.

zachmullen (approved):

PR #3563, opened 2024-08-21
Support pymongo 4

manthey:
This is done by re-adding the count method to cursors based on code in pymongo 3.  It still shows a deprecation warning, as it'd be better for future support to refactor code to use the count_documents method.

Although this could be merged as is, it is intended as, at last, a point for discussing the approach that should be taken.

There are a few points in Girder (and probably several in plugins as well) where counts are used adjacent to the query.  These places are relatively straightforward to update, and, for each, one could choose to use the count_documents or estimated_document_count as appropriate.  The sticking point is that when we return a cursor from an endpoint, we serialize the requested records and add a Girder-Total-Count header with the document count.  If we don't have a cursor count method or a workaround equivalent to it, then any endpoint that returns a cursor would instead need to also store the count in some out-of-band part of the response (adding the header itself, adding the count to a property of some class, or something similar).  This seems like it is more burdensome and more prove to failure than the cursor count method, but has the virtue of less surprise in the future.

### DIFF
--- a/girder/models/__init__.py
+++ b/girder/models/__init__.py
@@ -1,10 +1,30 @@
 import pymongo
+import pymongo.cursor
 import urllib.parse
 
 from girder import logprint
 from girder.utility import config
 
 _dbClients = {}
+
+
+if not hasattr(pymongo.cursor.Cursor, 'count'):
+    import warnings
+
+    def _cursorCount(self, with_limit_and_skip=False):
+        warnings.warn(
+            'count is deprecated. Use Collection.count_documents instead.',
+            DeprecationWarning,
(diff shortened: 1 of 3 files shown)


manthey:
For what it is worth, I've been trying this out in a local deployment with a few million items and lots of metadata for the last few weeks.  Anecdotally, pymongo 4 is nearly twice as fast as pymongo 3 for bulk inserts and also faster (but to a much lesser degree) for most other actions.

zachmullen:
This LGTM. I think we should merge this, merge master into v4-integration, then remove this polyfill in Girder 5.

PR #3568, opened 2024-09-18
pymongo 4.9 requires explicit import of command_cursor.

manthey:
Also, the internals of pymongo cursor have changed.

This is backwards compatible with all of pymongo 4.x (and probably 3.x)


manthey:
pymongo has introduced synchronous and asynchronous cursors and command cursors.  Their refactor changes quite a bit under the hood.

zachmullen (approved):

PR #3570, opened 2024-09-26
Style update (experimental)

jeffbaumes:
Adds tailwind and does some restyling of the core app.

Also enables a `extras` directory which can host custom files and includes `extras.css` (initially empty) which can be replaced with custom CSS at deployment time.

Known issues:
* Quick search is currently not there but I would like to have that integrated back
* Not tested against arbitrary plugins, is intended to work mainly with `large_image` and `histomicsui` plugins at the current time

This branch will be substantially upgraded for Girder 5 and is only being proposed to be merged to an experimental integration branch in Girder 3.

### DIFF
--- a/girder/web_client/src/views/App.js
+++ b/girder/web_client/src/views/App.js
@@ -2,6 +2,7 @@ import $ from 'jquery';
 import _ from 'underscore';
 import Backbone from 'backbone';
 import 'typeface-open-sans';
+import 'remixicon/fonts/remixicon.css';
 import 'bootstrap/dist/css/bootstrap.css';
 import 'bootstrap/js/alert';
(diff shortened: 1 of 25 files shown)


girder/web_client/src/views/App.js
```diff
@@ -188,7 +189,11 @@ var App = View.extend({
         }
         this.$el.html(LayoutTemplate());
 
-        this.globalNavView.setElement(this.$('#g-global-nav-container')).render();
+        // Only show side navigation if logged in as an admin
+        if (getCurrentUser() && getCurrentUser().get('admin')) {
```
manthey, line 193: @jeffbaumes We should remove the check for admin -- regular users need to be able to navigate the collections (and probably groups and users, too).

PR #3576, opened 2024-10-24
Remove references to "Resonant"

brianhelba:
Resonant has become a distinct product. Any benefits of cross-marketing are outweighed by the likelihood of ongoing confusion for consumers, especially given the historical usage of "Resonant" to refer to direct parts of the Girder stack.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -27,9 +27,8 @@ Girder: a data management platform
 What is Girder?
 ---------------
 
-Girder is a free and open source web-based **data management platform** developed by
-`Kitware <http://www.kitware.com/>`_ as part of the
-`Resonant data and analytics ecosystem <http://resonant.kitware.com/>`_. What does
+Girder is a free and open source web-based **data management platform**, developed by
+`Kitware <http://www.kitware.com/>`_. What does
 that mean? Girder is both a standalone application and a platform for building new web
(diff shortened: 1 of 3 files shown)


docs/index.rst
```diff
@@ -27,9 +27,8 @@ Girder: a data management platform
...
 ---------------
 
-Girder is a free and open source web-based **data management platform** developed by
-`Kitware <http://www.kitware.com/>`_ as part of the
-`Resonant data and analytics ecosystem <http://resonant.kitware.com/>`_. What does
+Girder is a free and open source web-based **data management platform**, developed by
+`Kitware <http://www.kitware.com/>`_. What does
 that mean? Girder is both a standalone application and a platform for building new web
 services. It's meant to enable quick and easy construction of web applications
 that have some or all of the following requirements:
```
marySalvi, line 34: I think this sentence is awkward after removing the 'part of Resonant'. Also I am not sure we want to suggest using it for new applications here.

zachmullen (approved):

docs/index.rst
brianhelba: @marySalvi That's a reasonable point, but I think it's out of scope for this change.

I'd be fine if you want to change the language in another PR with a different scope.

PR #3577, opened 2024-10-24
Change Discourse references to github Discussions

marySalvi:


brianhelba:
@marySalvi We should verify that the docs are correctly building this change, now that it's merged.

PR #3578, opened 2024-10-31
Pin setuptools-scm version

zachmullen:
The latest release breaks if a pyproject.toml file isn't present.

At some point we should support the new behavior instead of the current way, but that's out of scope of v4-integration.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(
     name='girder',
     use_scm_version={'local_scheme': prerelease_local_scheme},
     setup_requires=[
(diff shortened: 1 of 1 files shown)


manthey:
You could also just add a basic pyproject.toml file.  I think
```
[build-system]
requires = ["setuptools", "setuptools-scm"]
build-backend = "setuptools.build_meta"
```
will work.

zachmullen:
I think we should do that, but it makes more sense to do it on master and make sure it's working properly there.

PR #3581, opened 2024-11-01
Harden S3 imports.

manthey:
S3 buckets can have folders and items with the same name.  We want to reuse existing folders so when we import a second time we don't duplicate data.  In an item already exists with the name of the folder, append a / (this is what S3 does internally) and use that as a name.

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -468,6 +468,14 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
                     progress.update(message=obj['Prefix'])
 
                 name = obj['Prefix'].rstrip('/').rsplit('/', 1)[-1]
+                # If there is already an item with the folder's name, append
+                # '/'.  This is what S3 does internally, allowing a folder and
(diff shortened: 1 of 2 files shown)


girder/utility/s3_assetstore_adapter.py
```diff
@@ -468,6 +468,14 @@ def importData(self, parent, parentType, params, progress,
                     progress.update(message=obj['Prefix'])
 
                 name = obj['Prefix'].rstrip('/').rsplit('/', 1)[-1]
+                # If there is already an item with the folder's name, append
+                # '/'.  This is what S3 does internally, allowing a folder and
+                # file to have the same name once stripped of the right /.
```
zachmullen, line 473: To be a bit pedantic, the internal reason is because objects in S3 are not actually hierarchical, they're a flat list a la materialized paths. The presentation of things in S3 as folders is a bit of an illusion; objects don't have the same name, they simply share a common prefix.

I think this is a fine accommodation of that behavior, but maybe we could word the comment more clearly.

zachmullen (approved):

PR #3586, opened 2024-11-09
Fixing EventStream errors when logged in with OAuth

mitchklusty:
Hi. I have been working on a DSA instance that uses OAuth for login. I have made fixed some login issues on my local instance that are not ready for a pull because they do not yet work generally, but this fix to the EventStream behavior is ready. It addresses an issue of the girderToken cookie not being properly set when using OAuth, causing breaks in the EventStream that prevent viewing images properly. If something needs to be done to make my changes more "Girder-like," I am open to critique on them, but this addresses a large issue of using Girder with OAuth.


zachmullen:
Hi @mitchklusty , I believe this issue was resolved already with the following commit: https://github.com/girder/girder/commit/f6c95beec0585937778aa68d8ee7985164251853

That is contained in release 3.2.6, have you tried that version?

It's worth nothing that we are moving Girder to use `HttpOnly` cookies, so the approach here wouldn't work with that.

PR #3587, opened 2024-11-12
Move slicer_cli_web plugin into this repository

zachmullen:
- [ ] Re-enable web client tests for this plugin
- [ ] Convert `_createPutImageJob` to a celery task rather than an async local job (which are no longer supported in Girder 5)
- [x] Enable batch mode on generic runCli endpoint
- [x] Enable batch mode on rerunCli endpoint


zachmullen:
@manthey I will probably need some help understanding the intended behavior of the batch job integration tests. Switching from the dynamically named run endpoints to the UUID run endpoint seems to have introduced a behavioral change that breaks the tests.

PR #3591, opened 2024-11-18
Allow to set a domain attribute for the girder cookie

Xarthisius:
### Rationale

We allow users to spin up 3rd party webapps along side girder (basically enabling access to data via things like Jupyter Lab, RStudio, VSCode etc). Each instance of the webapp must only be accessible by the user that ran it. Since we cannot control apps themselves and make them respect girderToken to set their own cookies, access control is enforced on a reverse proxy level e.g. [Traefik's Forward Auth](https://doc.traefik.io/traefik/middlewares/http/forwardauth/) . In order for that to work we need to set girderToken cookie with a domain.

### Proposed solution
Allow to set a domain attribute for the girder cookie via a setting.


zachmullen (changes requested):
LGTM other than removing the front-end changes

Xarthisius (commented):
Dropped frontend changes and rebased.

girder/api/rest.py
```diff
@@ -1198,6 +1198,9 @@ def sendAuthTokenCookie(self, user=None, scope=None, token=None, days=None):
         if cherrypy.request.scheme == 'https' or cherrypy.request.base.startswith('https'):
             cookie['girderToken']['secure'] = True
 
+        if domain := Setting().get(SettingKey.COOKIE_DOMAIN):
```
zachmullen, line 1201: I think this is why the integration test is failing; this language feature isn't supported on our minimum supported python version (sadly).

Xarthisius: `setup.py` has `python_requires='>=3.8'` that is 3.8 feature

Xarthisius: `'bad'` was passing the validator for COOKIE_DOMAIN. I fixed it.

PR #3595, opened 2024-12-16
Add a show-download setting that defaults to all

manthey:
This setting is core.download_shown and can be "all" (default), "user", "admin", or "none".  If not all, download controls are only shown for a subset of users -- any logged in user, only those who are admins, or no one.

This doesn't effect endpoints, just what is shown in the UI.


manthey:
@jeffbaumes You have to restart girder if you change this setting to see it take effect.  You can use the usual techniques: add it to the setting in the provisioning file, use an environment variable, or hit the settings endpoint:
image

manthey:
Also, I have a branch that would merge into girder master if we want to avoid feature separation (but it would also then need to be pushed into the v4 integration branch).

jeffbaumes (approved):

PR #3599, opened 2025-01-06
Add CILogon oauth provider

### LINKED ISSUE #3594: CILogon OAuth Implementation
I'm working on adding CILogon as a provider to the OAuth plugin. I haven't been able to find all of the files I need to edit to make a new configuration available. Is there documentation somewhere for adding a new provider? Thanks.

mitchklusty:
Resolves #3594 

Issues with this PR that I'd like comments on:
1. I have not written a test for this
2. I am using the Box icon for CILogon currently. Not sure how to add the CILogon logo.
3. I don't know if this is a bug or not, when running Girder through the DSA docker container, the `getApiUrl()` function from `girder/api/rest.py` is returning `localhost:8080` when it should return the site URL as far as I can tell. This is causing some issues for my site, but I am not sure how to fix it. I copied how the other plugins were written, so I think this is just an issue with the function. I have also seen behavior before where the most recently added callback URL for the OAuth service is the only one it redirects to. I'm not sure how that affects this function though, so they may be unrelated issues.

### DIFF
--- a/plugins/oauth/girder_oauth/providers/cilogon.py
+++ b/plugins/oauth/girder_oauth/providers/cilogon.py
@@ -1,0 +1,93 @@
+import warnings
+
+from girder.api.rest import getApiUrl
+from girder.exceptions import RestException
+from girder.models.setting import Setting
+
+from ..settings import PluginSettings
+from .base import ProviderBase
+import requests
+
+
+class CILogon(ProviderBase):
+    _AUTH_SCOPES = ['openid', 'email', 'profile']
+    _API_USER_URL = 'https://cilogon.org/oauth2/userinfo'
+
+    @classmethod
+    def _authority(cls):
+        return 'https://cilogon.org'
+
+    def getClientIdSetting(self):
+        return Setting().get(PluginSettings.CILOGON_CLIENT_ID)
+
+    def getClientSecretSetting(self):
+        return Setting().get(PluginSettings.CILOGON_CLIENT_SECRET)
+
+    @classmethod
+    def getUrl(cls, state):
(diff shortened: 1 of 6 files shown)


scripts/publicNames.txt
```diff
@@ -1214,6 +1214,13 @@ plugins
                         getToken
                         getUrl
                         getUser
+                cilogon
```
manthey, line 1217: These lines just need to be moved up so that cilogon is alphabetical.  The needed diff is attached 
[diff.txt](https://github.com/user-attachments/files/18457495/diff.txt)

manthey:
@mitchklusty For an icon, if we can find an svg of the preferred icon, we can add it to the icon set.

plugins/oauth/girder_oauth/providers/cilogon.py
```diff
@@ -0,0 +1,94 @@
...
+import requests
+
+
+class CILogon(ProviderBase):
+    _AUTH_SCOPES = ['openid', 'email', 'profile']
+    _API_USER_URL = 'https://cilogon.org/oauth2/userinfo'
+
+    @classmethod
+    def _authority(cls):
+        return 'https://cilogon.org'
```
zachmullen, line 18: Given that this is just a static value, if you don't want to repeat it, let's just do:

```suggestion
    _AUTHORITY = 'https://cilogon.org'
```

(Note that it's already repeated in the above `_API_USER_URL`.

zachmullen (changes requested):
LGTM other than the one small suggestion. Thanks!

zachmullen (approved):

PR #3617, opened 2025-02-14
Expose assetstoreImportViewMap as a property of AssetsoresView

willdunklin:
This is needed for girder_assetstore because of the `import @girder/core` -> `window.girder` changes:
girder_assetstore can no longer import `assetstoreImportViewMap` from AssetsoresView.js.

### DIFF
--- a/girder/web/src/views/body/AssetstoresView.js
+++ b/girder/web/src/views/body/AssetstoresView.js
@@ -214,7 +214,9 @@ var AssetstoresView = View.extend({
             this.render();
         }, this);
         editAssetstoreWidget.render();
-    }
+    },
+
+    assetstoreImportViewMap,
 }, {
     import: function (assetstoreId) {
         var assetstore = new AssetstoreModel({ _id: assetstoreId });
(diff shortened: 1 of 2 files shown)


zachmullen:
Please coordinate with @naglepuff on this fix

naglepuff:
@willdunklin see https://github.com/girder/girder/pull/3613 for an alternate approach, which was merged earlier today.

If you think the change in this PR is better, I don't have a preference, so you can continue with this if you want, but might need to rebase and make a small change in views/body/index.js

willdunklin:
@naglepuff I prefer the method in this PR if you don't mind, I think it makes more sense than "importing" the map from `girder.views.body`.
The rebase should be all set though!

zachmullen:
I think I slightly prefer just keeping it as a module-scoped variable. I appreciate the convenience of the view objects keeping a reference to it, but that kind of obfuscates its purpose as a global registry -- it's exposed from the module because it's intended to be mutated by downstreams. Mutating it through a View instance feels odd to me.

willdunklin:
That's fair, I see what you mean. I don't have a strong preference either way so we can keep the existing code from #3613

PR #3618, opened 2025-02-17
Redesign S3 assetstore finalization process

zachmullen:
This moves the multipart upload finalization step from the web client to the server.

While working in this code, I also tuned the part size from 32MB to 64MB to have fewer parts, to reduce the overhead in the multipart upload process while still retaining a reasonable amount of resumability for modern network performance.

This ended up being simpler than my initial attempts. It's still a breaking change, as any clients still attempting to finalize their own multipart uploads would break (they are no longer even provided a signed URL to attempt it).

I tested this with both single-shot and multipart uploads and everything seems to be working.

FYI @pchoisel


pchoisel:
That looks great, thanks

PR #3621, opened 2025-02-19
Switch from as-jqplot to apexcharts for pie charts.

manthey:
We show pie charts in the Assetstore list and in the user quota. as-jqplot hasn't been updated in ages; apexcharts is currently popular and well maintained.

These look slightly different:
Girder 3:
image
This PR:
image

### DIFF
--- a/girder/web/package-lock.json
+++ b/girder/web/package-lock.json
@@ -1,16 +1,16 @@
 {
   "name": "@girder/core",
-  "version": "5.0.0-beta.0",
+  "version": "5.0.0-beta.1",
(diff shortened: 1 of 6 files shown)


manthey:
In the user quota plugin, I require apexcharts as a dependency.  We don't seem to do this much in other plugins.  @zachmullen Is this preferred, or should the girder package expose this in some manner and the plugin use it from there?

manthey:
@naglepuff With this, the assetstore page loads cleanly.

naglepuff:
looks great

PR #3622, opened 2025-02-19
Move import tracker to girder repo under plugins/

naglepuff:
Depends on #3621


manthey:
I realize this is still in draft status, but currently the edit a previous import doesn't work.

naglepuff:
thanks for testing it out. i will look into that

naglepuff:
@manthey I believe this is working now, fix was commit a68609a

manthey (approved):

PR #3626, opened 2025-02-26
Always configure girder worker app via the environment

zachmullen:
This removes the previous server-side settings for configuring the broker and backend. Because configuration should be identical between worker and server side, we cannot use the database for configuration.


manthey:
We should remove the settings from plugins/worker/girder_plugin_worker/web_client/views/ConfigView.js and plugins/worker/girder_plugin_worker/web_client/templates/configView.pug.

zachmullen:
Thanks, I will do that and also update the migration guide before merging this.

PR #3630, opened 2025-03-12
Only use environment variables for configuring girder worker

manthey:
I still need to update the documentation.


zachmullen (approved):
This LGTM -- let's also add something to the migration guide for this, given that Girder Worker is now part of Girder itself.

PR #3631, opened 2025-03-13
Guess mimeType during import if none was provided

### LINKED ISSUE #1645: girder_client and FS assetstore import should attempt to guess MIME type

Xarthisius:
Fixes #1645

I run into a weird issue where a) `item_previews` were not rendering images and b) clicking on "eye" icon in Item/File view resulted in images being downloaded instead of opened in a browser. Turns out it's all due to mimeType and those file had mimeType undefined / octet-stream because they were imported.

### DIFF
--- a/girder/utility/filesystem_assetstore_adapter.py
+++ b/girder/utility/filesystem_assetstore_adapter.py
@@ -1,6 +1,7 @@
 import filelock
 from hashlib import sha512
 import io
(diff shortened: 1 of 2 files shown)


zachmullen:
Thanks @Xarthisius ! I've tagged @manthey for review since he has dealt with importing the most.

PR #3632, opened 2025-03-17
Remove the `girder-worker` command

zachmullen:


jeffbaumes:
@zachmullen is there actual cli code to remove as well, beyond just removing the console script in setup.py?

jeffbaumes:
Didn't know if this file (which the girder-worker command points to) is dead and can be removed: https://github.com/girder/girder/blob/v4-integration/worker/girder_worker/__main__.py

PR #3634, opened 2025-03-17
Adjust swagger to include brandName and a description

manthey:


girder/api/describe.py
```diff
@@ -548,12 +548,18 @@ def listResources(self, params):
...
             'info': {
-                'title': 'Girder REST API',
-                'version': VERSION['release']
+                'title': f'{brandName} REST API',
+                'version': VERSION['release'],
+                'description':
+                    'Below you will find the list of all of the resource '
+                    f'types exposed by the {brandName} RESTful Web API. '
+                    'Click any of the resource links to open up a list of '
+                    'all available endpoints related to each resource type.',
```
zachmullen, line 562: For authentication docs, we can mention that users will be required to input an auth token into the swagger-ui Authorize dialog. We should document that the two ways to obtain such a token via that page are to use the `POST /api_key/token` if they are using an API key (recommended) and grab the token from that response, or else use the `GET /user/authentication` with Basic Auth. Unfortunately for the latter, there's no easy way to do that via the swagger-ui page itself; if we just exposed the header naively, the user would be required to base-64 encode the credentials themselves. If we expose it as a top-level auth method on the whole API, it'd be confusing since it only applies to that one endpoint. Is there a way with swagger to expose Basic Auth on just one endpoint?

manthey:
@zachmullen With these changes, the swagger lock icon per endpoint shows the correct auth (token for everything except GET user/authentication).  The global lock icon offers both.

zachmullen (approved):

PR #3643, opened 2025-03-21
Allow to set cherrypy thread_pool and db.replica_set via env vars

Xarthisius:
After the config removal cherrypy settings default to 10 threads and there's no way to change that. This PR introduces support for setting cherrypy thread pool size vie env var `GIRDER_THREAD_POOL`. Additionally I added support for `database.replica_set` (via `GIRDER_MONGO_REPLICA_SET`) for full backward compatibility.

### DIFF
--- a/docs/migration-guide.rst
+++ b/docs/migration-guide.rst
@@ -202,6 +202,15 @@ via the config file alone. Now, the caching modules are configured via environme
 Config keys prefixed by ``cache.global.`` are used to configure the global dogpile cache, and
 keys prefixed by ``cache.request.`` are used to configure the request cache.
 
+CherryPy specific settings are now passed via environment variables as well. List of settings that
+can be configured:
+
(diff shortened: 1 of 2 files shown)


zachmullen:
@Xarthisius for what it's worth, the use of cherrypy as the WSGI server is no longer supported and is recommended against for production use cases. `girder serve` is only to be used in development, and we recommend using something like `gunicorn` to server Girder in production now. We can merge this change, but just wanted to make you aware. The Replica Set support is still good to merge regardless.

Xarthisius:
Development is what prompted this PR. When you refresh girder's UI to e.g. reload plugin it's easy to bog down the server with `GET /notification/stream` when there are only 10 threads. I haven't checked if that also the case when running with `guvicorn`

zachmullen:
I'm going to merge this and fix the failing test on another branch. Thanks!

PR #3644, opened 2025-03-21
Fix pytest plugin mocking to work with setuptools >= 77

zachmullen:


Xarthisius:
It's `setuptools==77.0.3` that's causing issues. I think they might have fixed it upstream, but haven't released yet.

PR #3646, opened 2025-03-25
Serve js with a proper mimeType. Fixes #3645

### LINKED ISSUE #3645: Plugins js code is served with a wrong content type
Firefox complains a lot about it in the dev console:

```
The script from “https://girder.local.xarthisius.xyz/plugin_static/jobs/girder-plugin-jobs.umd.cjs” was loaded even though its MIME type (“text/html”) is not a valid JavaScript MIME type.
```

and it's repeated for every plugin.

Xarthisius:
Makes firefox happy.

### DIFF
--- a/girder/utility/server.py
+++ b/girder/utility/server.py
@@ -40,6 +40,7 @@ def create_app(mode: str) -> dict:
     mimetypes.add_type('application/vnd.ms-fontobject', '.eot')
     mimetypes.add_type('application/x-font-ttf', '.ttf')
     mimetypes.add_type('application/font-woff', '.woff')
+    mimetypes.add_type('text/javascript', '.cjs')
 
     curConfig.update(appconf)
     curConfig['server'] = {'mode': mode}


girder/plugin.py
```diff
@@ -42,6 +42,7 @@ def registerPluginStaticContent(plugin: str, css: List[str], js: List[str], stat
             '/': {
                 'tools.staticdir.on': True,
                 'tools.staticdir.dir': str(staticDir),
+                'tools.staticdir.content_types': {'cjs': 'application/javascript'},
```
zachmullen, line 45: Let's also add this to Girder core here: https://github.com/girder/girder/blob/v4-integration/girder/utility/server.py#L29-L42

If we add it using the `mimetypes` mechanism, I think it will automatically apply to all plugins too.

zachmullen (changes requested):

girder/plugin.py
Xarthisius: That works too!

zachmullen (approved):

zachmullen:
Thanks!

PR #3647, opened 2025-03-25
Introduce "local" celery task queue for long-running async tasks that need database and/or local filesystem access

zachmullen:
This is the Girder 5 replacement for the now-removed events daemon. All tasks that might take longer than 1-2 seconds should eventually be converted to use this mechanism.

This introduces the need to run an additional process on a machine that can connect to the mongo database, namely `celery -A girder_worker.app worker -Q local`.

I still need to document these changes.

### DIFF
--- a/girder/tasks.py
+++ b/girder/tasks.py
@@ -1,0 +1,20 @@
+from girder.models.assetstore import Assetstore
+from girder.utility.progress import ProgressContext
+from girder_worker.app import app
+
+
+@app.task(queue='local')
+def importDataTask(
+    assetstore: dict,
+    parent: dict,
+    parentType: str,
+    params: dict,
+    progress: bool,
+    user: dict,
+    leafFoldersAsItems: bool
+):
+    with ProgressContext(progress, user=user, title='Importing data') as ctx:
+        return Assetstore().importData(
+            assetstore, parent=parent, parentType=parentType, params=params,
+            progress=ctx, user=user, leafFoldersAsItems=leafFoldersAsItems,
+        )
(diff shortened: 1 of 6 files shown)


zachmullen:
@manthey PTAL. We'll need to do some impact analysis to discover all the places we need to add this new process to deployments.

girder/tasks.py
```diff
@@ -0,0 +1,20 @@
...
+def importDataTask(
+    assetstore: dict,
+    parent: dict,
+    parentType: str,
+    params: dict,
+    progress: bool,
+    user: dict,
+    leafFoldersAsItems: bool
+):
+    with ProgressContext(progress, user=user, title='Importing data') as ctx:
```
zachmullen, line 16: All `queue='local'` tasks can use any existing mechanisms that require database access, such as `ProgressContext`. Because these are using the `girder_worker` app, they'll also create and manage Job records automatically.

tests/cases/assetstore_test.py
```diff
@@ -22,9 +22,12 @@
 from girder.utility.progress import ProgressContext
 from girder.utility.s3_assetstore_adapter import makeBotoConnectParams
 from girder.utility import path as path_util
+from girder_worker.app import app
 
 
 def setUpModule():
+    app.conf.task_always_eager = True
+    app.conf.task_eager_propagates = True
```
zachmullen, line 30: These are the two config options that will need to be used in testing to have tasks run eagerly. If set, we shouldn't need to change (much) testing code, as it should behave the same as if it was running the code in the request thread.

PR #3653, opened 2025-04-11
Add girder-client dowload option to skip file downloads conditionally

willdunklin:


brianhelba (changes requested):
Hey @willdunklin , it would be nice for Git hygiene if you could [squash the commits which fix linting errors](https://dev.to/the_real_stacie/git-are-you-an-over-committer-squash-those-commits-2klk).

PR #3667, opened 2025-05-05
Update the girder/fontello package requested

manthey:
Also update some packages for vue and typescript.

@zachmullen I'm not sure why some of the vue packages were set to older versions.

@bnmajor This should resolve getting the most recent fontello icons


zachmullen:
No particular reason.

zachmullen (approved):

PR #3671, opened 2025-05-07
Make girder-side requirements for slicer_cli_web unconditional

zachmullen:
These packages get imported on both the worker side and the girder side.

### DIFF
--- a/plugins/slicer_cli_web/setup.py
+++ b/plugins/slicer_cli_web/setup.py
@@ -59,24 +59,22 @@ setup(
     packages=find_packages(exclude=['tests', 'test.*']),
     zip_safe=False,
     install_requires=[
+        'girder>=5.0.0a2',
+        'girder-jobs>=5.0.0a2',
+        'girder-worker>=5.0.0a4',
+        'girder-client',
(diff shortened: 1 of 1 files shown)


manthey:
There are packages that use the slicer_cli_web modifications to ctk_cli that DON'T use girder. (e.g. they use slicer_cli_web.CLIArgumentParser).  Doesn't this change make it so those packages now will end up with girder as a dependency?

zachmullen:
The motivation for this was that I was installing `slicer_cli_web[worker]` into an EC2 worker for a Histomics deployment, and that was insufficient to get it working due to the module-level imports that are used.

zachmullen:
That's fine, I am going to merge this as a temporary fix on v4-integration so I can get this deployment working, and we can come up with a better long term solution.

manthey:
I will also note that without this, the docker-compose setup DSA uses has been working (probably because we are installing more than necessary and using the same docker image for both girder and the workers).

zachmullen:
That checks out; I think if you install `girder` extra on both sides it will work fine.

PR #3674, opened 2025-05-08
Internally implement cache-busting for all plugin static content

### LINKED ISSUE #3669: Investigate cache-busting for plugin static content
When plugin javascript and styles are rebuilt, the browser over-aggressively caches them because they are not renamed. The accepted way to solve this is to add a file hash to the filename during build time. This will require globbing in plugin code when calling `registerPluginStaticContent`, and it will also make our `watch` for plugin frontend code no longer work with `registerPluginStaticContent` without a more sophisticated setup for development.

zachmullen:
Fixes #3669.

I chose to use the file hash as the key because it seems more stable than file mtimes across execution environments. The hashing only happens once at server startup.

In local development, developers should use the "Disable cache" developer setting in their dev tools when using the watch mode for plugin front-end development.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,6 +69,11 @@ commands:
             --editable ./worker
             --requirement requirements-dev.txt
           working_directory: girder
+  build-plugins:
+    steps:
+      - run:
+          name: Install and build all plugins
+          command: python girder/.circleci/build_plugins.py girder/plugins
 
 jobs:
   server-lint-test:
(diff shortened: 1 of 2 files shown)


zachmullen:
The test failures are because the static files aren't built in CI testing. I'm actually considering this a feature rather than a bug after thinking about it -- the server should fail to start if static content is missing, because the web client won't start up if that's the case. I'll modify CI to make sure the front-end code is built for all plugins.

PR #3675, opened 2025-05-26
Add option to use S3 accelerated endpoint for faster transfers

pchoisel:


manthey:
In a quick read about the use_accelerate_endpoint, it sounds like it only has benefit when the client is in a different region from the bucket and the files are largish.  Should there be any sort of check for either of these conditions, since using use_accelerate_endpoint incurs higher transfer costs?

pchoisel:
> In a quick read about the use_accelerate_endpoint, it sounds like it only has benefit when the client is in a different region from the bucket and the files are largish. Should there be any sort of check for either of these conditions, since using use_accelerate_endpoint incurs higher transfer costs?

That sounds good, but I'm not sure how to implement this.
I can easily know the location of the S3 bucket, but it's harder to get the location of the client. Maybe using its IP address ? But that would work if Girder is reverse proxied.

pchoisel:
@zachmullen Thanks.  
A customer wants to enable [S3 transfer acceleration](https://aws.amazon.com/s3/transfer-acceleration/) for his users that are not in the US to speed up their uploads/downloads to a Girder-based application using an S3 assetstore.  
You can enable that by changing the domain name used to connect to S3, but Boto does that automatically if you set `use_accelerate_endpoint` in the config.

To make this more optimized, I could perhaps use this accelerated endpoint only when Girder sends a transfer link to a user so their client can transfer data to S3, and not when Girder directly transfers data to S3 ?

pchoisel:
I agree that option 2 is the best. However, there are some clients that I cannot change easily #girderwebcomponents  
But that's fine, I'll just patch them a bit more. Inferring the location of the client using its IP address seems really shaky.  

Do you think I should still add an assetstore option to enable accelerated transfer ? That would make the endpoints return an error if it's disabled but the client requested a accelerated transfer.

zachmullen:
Yes, we should probably only allow accelerated transfer if the assetstore explicitly allows it.

girder/api/v1/file.py
```diff
@@ -254,8 +262,8 @@ def readChunk(self, upload, offset, params):
         .param('contentDisposition', 'Specify the Content-Disposition response '
                'header disposition-type value.', required=False,
                enum=['inline', 'attachment'], default='attachment')
-        .param('extraParameters', 'Arbitrary data to send along with the download request.',
-               required=False)
+        .jsonParam('extraParameters', 'Arbitrary data to send along with the download request.',
```
manthey, line 265: I'm not sure what the ramifications are of changing from `param` to `jsonParam` will be.  This switching the extra parameters from a string to a parsed json object.  The only place I see it used before this is in downloads.

Issue #3657, opened 2025-04-17
Fix task in import tracker


zachmullen:
Import tracker has its own version of the import endpoint that replaces the core version. We can probably remove this and make the core endpoint sufficient.

Additionally, make it possible when invoking tasks to specify that no girder Job object should be created, via some special kwarg.

PR #3676, opened 2025-05-28
Redesign import tracker core behavior for new local queue architecture

zachmullen:
With assetstore import moving off the request thread and into celery workers, it was necessary to redesign the import_tracker plugin behavior from the ground up.


zachmullen:
This requires a new and unprecedented type of behavioral augmentation in plugins: `girder.events` bindings within celery worker-side code. Fortunately, the existing mechanism of girder worker plugins made this very straightforward. They're usually used just for task discovery, but they can also be used to bind event handlers and everything "just works" if the events are being triggered within celery tasks on the `local` queue.

zachmullen:
Fixes #3657

PR #3680, opened 2025-06-06
Switch to ASGI and overhaul Girder's notification system

### LINKED ISSUE #3665: Overhaul notifications system

zachmullen:
Fixes #3665 

It's been a longstanding problem with Girder that long-polling using the browser's `EventSource` API would quickly exhaust Girder's very limited connection pool under WSGI despite very low resource utilization incurred by those requests.

This PR moves away from `EventSource` to instead use websockets, which necessitates moving Girder's server to run under ASGI rather than WSGI. Redis is used as the pub/sub mechanism to facilitate message passing to the client websockets. Starlette is used as the ASGI framework -- I initially attempted to use `asgiref`'s WSGI to ASGI adapter, but it was flaky and requests would intermittently fail in single-threaded mode due to deadlock on the event loop. Starlette's adapter is intentionally designed to avoid that problem.

TODO:

* [x] Make `girder serve` serve the ASGI app in dev mode
* [x] Document all the API changes around the notification model
* [x] Figure out how to best handle errors when publishing messages on the notification channel (failure to write these transient messages should not cause the entire task to fail, but should probably be logged)
* [x] Fix all failing tests
* [x] Reintroduce time estimation feature in side progress context
* [x] Move notification code out of the models directory


zachmullen:
Note that if you want to test this with the large image plugins, you'll need to use the `updated-notifications-api` in the large_image repo. https://github.com/girder/large_image/pull/1938

manthey (approved):

Issue #3703, opened 2025-08-01
Progress notifications lack unique identifiers


Xarthisius:
### Description

New `progress` notifications lack `_id` field that's used by UI to manage notification pop ups (https://github.com/girder/girder/blob/v4-integration/girder/web/src/views/layout/ProgressListView.js#L35-L51). As a result only one toast with notification is present even if there are more things happening that generate independent progress streams.

### How to reproduce

```python

from girder.models.user import User
from girder.notification import Notification

user = User().findOne()
n1 = Notification.initProgress(user=user, title="foo", message="blah")
n2 = Notification.initProgress(user=user, title="bar", message="blah")
n1.flush()
n2.flush()
```

### Expected result
Two notification toasts are present in the UI

### Actual result
Only one toast present, first notification is replaced by the second.

### Note 
I depend on the fact that Notifications have `_id`, or rather on the fact that jobs have `notificationId` associated with them. https://github.com/girder/girder/pull/3702 breaks things for me too, but I assume it also breaks the updates for Girder proper.

zachmullen:
Good catch, thanks @Xarthisius . Will fix this today hopefully.

PR #3704, opened 2025-08-01
Restore the _id field of progress notifications

zachmullen:
@Xarthisius PTAL. Does this fix your issue (#3703 )?


Xarthisius:
> @Xarthisius PTAL. Does this fix your issue (#3703 )?

Yup! Thanks!

PR #3705, opened 2025-08-01
Omit uvicorn's Server header in `girder serve`

zachmullen:


girder/cli/serve.py
```diff
@@ -34,19 +34,22 @@ def main(mode: str, database: str, host: str, port: int, with_temp_assetstore: b
     config.getConfig()['server'] = {'mode': mode}
     config.getConfig()['database']['uri'] = database
 
+    def _run_app():
+        uvicorn.run('girder.asgi:app', host=host, port=port, reload=True, server_header=False)
```
manthey, line 38: Do we also want to add `date_header=False` to not duplicate cherrypy?

zachmullen: Done on #3706

PR #3707, opened 2025-08-01
Inline the favicon as part of the vite build

### LINKED ISSUE #3685: Add correct cache control headers to favicon serving
Whenever the hash state is updating in Histomics, Chrome for some reason attempts to re-fetch the favicon, and it's not using a cached value.

manthey:
The browser is requesting the favicon on every route change, even when we only change the fragment.  This is different from Girder-3, probably because of something that babel did in the webpack build.  I tried a variety of solutions to prevent constantly re-requesting the favicon, but this was the only one I found that actually worked reliably.

Closes #3685.

### DIFF
--- a/girder/web/vite.config.ts
+++ b/girder/web/vite.config.ts
@@ -1,3 +1,4 @@
+import fs from 'fs';
 import path, { resolve } from 'path';
 
 import { defineConfig } from 'vite';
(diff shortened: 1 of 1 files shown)


zachmullen:
Did you manage to reproduce this issue? I never was able to.

manthey:
Yes -- locally on http this reproduced every time.  On https connections I would get _occasional_ re-requests for favicon (and did even on Girder 3).  Reading into it, browsers vary when they decide they should check, and there doesn't seem to be any strict standard; if we mutate the url (even if only the fragment), different browsers could occasionally re-request the favicon.  It _seems_ to happen less when the favicon is in a subdirectory than in the root.  The two solutions seem to be to inline the favicon (as done here) or to _remove_ the favicon link node from the dom after loading.

zachmullen:
Strange. LGTM.

zachmullen (approved):

PR #3714, opened 2025-08-19
Retry s3 requests when fetching from the assetstore

manthey:
We find that occasionally the s3 request gets interrupted with a ChunkedEncodingError.  This is marked as an IOError, but doesn't have an errno set on it.  Both retry, in case the problem is recoverable, and set an errno so downstream handlers (such as the fuse.py library) break less.

### DIFF
--- a/girder/utility/s3_assetstore_adapter.py
+++ b/girder/utility/s3_assetstore_adapter.py
@@ -439,14 +439,12 @@ class S3AssetstoreAdapter(AbstractAssetstoreAdapter):
             raise cherrypy.HTTPRedirect(url)
         else:
             headers = {}
-            if offset or endByte is not None:
-                if endByte is None or endByte > file['size']:
-                    endByte = file['size']
+            offset = offset or 0
+            if endByte is None or endByte > file['size']:
+                endByte = file['size']
+            # only send the range header if we aren't asking for the whole file
+            if offset or endByte != file['size']:
                 headers = {'Range': 'bytes=%d-%d' % (offset, endByte - 1)}
-            else:
-                offset = 0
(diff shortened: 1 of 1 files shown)


girder/utility/s3_assetstore_adapter.py
```diff
@@ -441,12 +443,46 @@ def stream():
...
+                                except Exception:
+                                    # if the exception occurred because of the
+                                    # consumer, just stop
+                                    halt = True
+                                    raise
+                        halt = True
+                    except IOError as exc:
+                        retries += 1
+                        if halt or retries >= maxRetriesWithoutData:
+                            if not hasattr(exc, 'errno'):
```
pchoisel, line 480: I would just add a comment to let future us know that this is to accommodate fusepy

girder/utility/s3_assetstore_adapter.py
```diff
@@ -441,12 +443,46 @@ def stream():
                 if endByte is None or endByte > file['size']:
                     endByte = file['size']
                 headers = {'Range': 'bytes=%d-%d' % (offset, endByte - 1)}
+            else:
+                offset = 0
+                if endByte is None or endByte > file['size']:
+                    endByte = file['size']
```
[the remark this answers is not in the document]
manthey: See [b15fc04](https://github.com/girder/girder/commit/b15fc040278d828ffda923b2fc0bd28e62674f95)

girder/utility/s3_assetstore_adapter.py
manthey: See [b15fc04](https://github.com/girder/girder/commit/b15fc040278d828ffda923b2fc0bd28e62674f95)

PR #3723, opened 2025-09-22
Update pytest fixture to work with cheroot 11.x

manthey:
As of cheroot 11.0.0, the cheroot server creates a non-daemon thread that prevents pytest from shutting down in some instances.  Monkeypatch the code within that thread just return None so that it doesn't stop pytest from stopping.

### DIFF
--- a/pytest_girder/pytest_girder/utils.py
+++ b/pytest_girder/pytest_girder/utils.py
@@ -306,6 +306,17 @@ def serverContext(plugins=None, bindPort=False):
     from girder.utility.server import setup as setupServer
     from girder.constants import ServerMode
 
+    # As of cheroot 11.0.0, the cheroot server creates a non-daemon thread that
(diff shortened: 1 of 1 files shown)


zachmullen:
I assume that this bug would only be encountered in test code if `bindPort=True`, is that correct?

zachmullen:
Our teardown code for this fixture does the following actions on the cherrypy API:

```python
        cherrypy.engine.unsubscribe('start', girder.events.daemon.start)
        cherrypy.engine.unsubscribe('stop', girder.events.daemon.stop)
        cherrypy.engine.stop()
        cherrypy.engine.exit()
```

Are we just missing something extra that we need to be doing to stop the server? IMO if we are doing everything possible via the cherrypy API to stop the server in this fixture, and it still doesn't stop, that's definitely an upstream bug.

PR #3729, opened 2025-09-26
Stop using passlib

manthey:
bcrypt 5.x makes a change that explicitly warns about truncating long passwords.  passlib tests for a bug of some previous version by testing with a long password, which raises this exception.

We currently use passlib for bcrypt and for one-time-passwords (2fa). For the use of bcrypt, the change is straightforward.  We can get rid of our `_cryptContext` object and replace any uses of it (two in model/user.py and one in the rest user.py).  This will resolve the current issue with passlib, and won't affect current salts, etc.

The OTP requires a bit more of a refactor, but we can switch to pyotp. I haven't investigated whether this can be done without any major change.

### DIFF
--- a/girder/models/user.py
+++ b/girder/models/user.py
@@ -1,5 +1,4 @@
 import datetime
-import logging
 import os
 import re
 import time


manthey:
@zachmullen The changes to stop using bcrypt through passlib are pretty small.  I could do that instead of pinning bcrypt.  The work to refactor OTP is probably more, and would need a little investigation.  If you think it would be better to stop using bcrypt through passlib right now rather than pinning bcrypt, I'll update this PR.

PR #3731, opened 2025-09-29
Add an index to the file collection to speed up s3 deletions

manthey:
This changes the generalized collector of indices to not explicitly use sets, as we support compound indices


manthey:
> > to not explicitly use sets
> 
> Is this because we're using unhashable types to specify indices?

Yes.

PR #3738, opened 2025-11-04
Remove local girder folder mapping in docker compose

pchoisel:
This volume maps the content of the local girder folder in the container, shadowing what was previously there.
This prevents the front-end files that were built in the image to be used by the server, preventing people from accessing Girder's front-end.

While this volume is convenient for development purposes, I don't think it should be in the default docker-compose.yml file since it prevents users from easily setting up the project and have a working app.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3.8'
+# This docker compose setup is exclusively intended to be used as a development environment
 
 services:
   mongodb:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3.8'
```
zachmullen: Why did the version get removed?

zachmullen (approved):

PR #3798, opened 2026-04-08
Add a bridge between ASGi and WSGI for performance

manthey:
Closes #3795

There was a noticeable pause at the end of each chunk when uploading large files.  This pipes data between the layers for large responses.

Many thanks to @Xarthisius for hunting this down and generating an initial solution.

### DIFF
--- a/girder/asgi.py
+++ b/girder/asgi.py
@@ -1,12 +1,138 @@
+import asyncio
 import logging
+import socket
+import sys
+import threading
 from contextlib import asynccontextmanager
 
 from starlette.applications import Starlette
-from starlette.middleware.wsgi import WSGIMiddleware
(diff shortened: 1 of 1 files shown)


manthey:
@Xarthisius In my testing, this seems very close to the Girder 3 performance numbers; it is loosely based on the code you shared.

Xarthisius:
> @Xarthisius In my testing, this seems very close to the Girder 3 performance numbers; it is loosely based on the code you shared.

I think the next step would be to port download routes to native async to get the remaining difference. However, that'd be a significant undertaking and I don't think it's worth only a minor performance bump.

manthey:
@Xarthisius Agreed.  Unless you think something else is needed right now, I can merge this.

Xarthisius:
LGTM!

PR #3827, opened 2026-05-29
Add sphinx-llm

manthey:
This adds llms.txt and markdown files to the generated docs.

@marySalvi This doesn't address the issue you found but adds the markdown files we talked about.


marySalvi:
Great!

PR #3839, opened 2026-06-11
Propagate policy errors in OAuth2 flow

Xarthisius:
If girder user account is disabled or require an additional action before allowing user to login, inject meaningful 'error' in OAuth callback redirect.

I added a hacky way for showing an alert in Girder UI. I'm happy to improve (or remove)  it if you have any suggestions.


manthey:
@Xarthisius This is probably fine.  I feel there is some event-based method that would somehow be more correct, but cannot figure out how that would propagate where it needs to.  I'm going to accept this as it stands and merge it in.

Thank you.

manthey (approved):

PR #3845, opened 2026-06-18
Allow to overwrite emails sent due to various auth policies

Xarthisius:
This allows e.g. to customize templates and what goes into them.

### DIFF
--- a/tests/cases/user_test.py
+++ b/tests/cases/user_test.py
@@ -1,4 +1,7 @@
 import collections
+import contextlib
+import io
+import sys
 
 from girder import events
 from girder.constants import AccessType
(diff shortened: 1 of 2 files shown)


tests/cases/user_test.py
```diff
@@ -447,6 +489,27 @@ def testAccountApproval(self):
         self.assertStatus(resp, 401)
         self.assertEqual(resp.json['extra'], 'disabled')
 
+    def testAccountVerification(self):
+        User().createUser('admin', 'password', 'Admin', 'Admin', 'admin@girder.test')
+        Setting().set(SettingKey.EMAIL_VERIFICATION, 'required')
+        with captureOutput() as output:
```
manthey, line 495: Can we use the pytest `capsys` fixture instead of your proposed contextmanager?

Xarthisius: I'd love to, but these are not running with pytest AFAICT. Do you want me to move it from `tests/cases/` (which is legacy unittest) to `test/` (pytest)?

manthey: You are right.  I have a back-burner project to move all the legacy tests over to pytest, but that isn't for this month.

manthey (approved):

PR #3856, opened 2026-07-06
Update Playwright To Install Chromium Only

Underwaterr:
By default, `playwright install` will attempt install Chromium, Firefox and Webkit. Girder's Playwright config only tests with Chromium so I updated the `tox.ini` file to reflect that. This means tests run sooner and the test environment requires fewer dependencies. 

Additionally, any code in the Playwright config that was commented out has been removed.


manthey:
I'd rather test on firefox, too: https://github.com/girder/girder/pull/3865.  We could modify this to be just chromium and firefox, though.

PR #3858, opened 2026-07-08
Fix checked-resource count and picked-resource merge bugs

### LINKED ISSUE #3855: Selection counts in hierarchy viewer can be wrong
Check items from an item list.  Select "Pick items for move or copy".  The menu shows that the items are selected.  Click the select all button.  Select "Pick items for move or copy".  The menu now shows the total number of items plus the number first picked.

Example, I have a list of 5 items.  I pick two.  2 items are ready to be copied.  I select all and say pick them.  It now says 7 items.

In general, this should be maintained as a set (and collect across pages) and properly resolve to the correct number.

bnmajor:
Fixes #3855 and a related bug:
- `ItemListWidget`: guard against re-adding a `cid` already in `checked`on repeated change events (e.g. `shift-click` range selection), which previously caused the checked count to inflate.
- `HierarchyWidget.pickChecked`: only keep previously picked ids that are either still checked or not visible in the current folder/item list, so unchecking a picked resource actually removes it from the picked set instead of being silently re-added by the union merge.

- PREVIOUSLY
    - Select an item
    - Select `Pick checked resources for Move or Copy`
    - See that "selection" count is `1` and "picked" count is `1`
    - Screenshot from 2026-07-08 11-53-54
    - Shift+click to select a second item
    - Select `Pick checked resources for Move or Copy`
    - See that "selection" count is `3` [incorrect] and "picked" count is `2`
    -  Screenshot from 2026-07-08 11-54-19
    - Shift+click to de-select the second item
    - Select `Pick checked resources for Move or Copy`
    - See that "selection" count is `1` and "picked" count is `2` [incorrect]
    -  Screenshot from 2026-07-08 11-56-08
- CURRENTLY
    - Select an item
    - Select `Pick checked resources for Move or Copy`
    - See that "selection" count is `1` and "picked" count is `1`
    - Screenshot from 2026-07-08 11-59-04
    - Shift+click to select a second item
    - Select `Pick checked resources for Move or Copy`
    - See that "selection" count is `2` and "picked" count is `2`
    -  Screenshot from 2026-07-08 11-59-15
    - Shift+click to de-select the second item
    - Select `Pick checked resources for Move or Copy`
    - See that "selection" count is `1` and "picked" count is `1`
    -  Screenshot from 2026-07-08 11-59-26

### DIFF
--- a/girder/web/src/views/widgets/FolderListWidget.js
+++ b/girder/web/src/views/widgets/FolderListWidget.js
@@ -25,12 +25,14 @@ var FolderListWidget = View.extend({
         'change .g-list-checkbox': function (event) {
             const target = $(event.currentTarget);
             const cid = target.attr('g-folder-cid');
-            if (target.prop('checked')) {
-                this.checked.push(cid);
-            } else {
-                const idx = this.checked.indexOf(cid);
-                if (idx !== -1) {
-                    this.checked.splice(idx, 1);
+            if (!this.checked.includes(cid)) {
+                if (target.prop('checked')) {
+                    this.checked.push(cid);
+                } else {
+                    const idx = this.checked.indexOf(cid);
+                    if (idx !== -1) {
+                        this.checked.splice(idx, 1);
+                    }
                 }
             }
             this.trigger('g:checkboxesChanged');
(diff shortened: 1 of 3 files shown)


manthey:
I think we have the same issue in folders (besides items), and you'll need the equivalent change in `src/views/widgets/FolderListWidget.js`.

bnmajor:
> I think we have the same issue in folders (besides items), and you'll need the equivalent change in `src/views/widgets/FolderListWidget.js`.

Thanks, I completely overlooked this! Should be resolved now.

manthey (approved):

PR #3862, opened 2026-07-09
Remember picked selections even after they are de-selected

### LINKED ISSUE #3861: Picked selections are reset when all toggle is applied
Somehow, #3858 introduces a new bug.  On a hierarchy view with folders and items, select one of each and select "Pick checked resources for move or copy".  Now toggle the select all check box on and off so that nothing is selected.  Select either one item or one folder, and select "Pick checked resources for move or copy".  We would expect the picked items to still have the original folder and item, plus the new folder or item, but instead it only has the most recent folder or item.

@bnmajor

bnmajor:
Fixes #3861

### DIFF
--- a/girder/web/src/views/widgets/FolderListWidget.js
+++ b/girder/web/src/views/widgets/FolderListWidget.js
@@ -26,7 +26,9 @@ var FolderListWidget = View.extend({
             const target = $(event.currentTarget);
             const cid = target.attr('g-folder-cid');
             if (target.prop('checked')) {
-                this.checked.push(cid);
+                if (!this.checked.includes(cid)) {
+                    this.checked.push(cid);
+                }
             } else {
                 const idx = this.checked.indexOf(cid);
                 if (idx !== -1) {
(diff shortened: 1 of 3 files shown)


manthey:
I realize we introduced one other bug in girder #3858, I don't think this code: https://github.com/girder/girder/pull/3858/changes#diff-0c4be13591820c4300c8dedbe4492bcd692dc2690e955d2b43d7c17eb7c7b8dbR32-R35 is reachable, since the outer if will only be entered if cid isn't checked, and so that splice never happens.  This is the remove-the-checked item from the tally code. 

We originally had
```
if (target.prop('checked')) {
    this.checked.push(cid);
} else {
    const idx = this.checked.indexOf(cid);
    if (idx !== -1) {
        this.checked.splice(idx, 1);
    }
}
```
we changed it to
```
if (!this.checked.includes(cid)) {
    if (target.prop('checked')) {
        this.checked.push(cid);
    } else {
        const idx = this.checked.indexOf(cid);
        if (idx !== -1) {
            this.checked.splice(idx, 1);
        }
    }
}
```
but I think we should have changed it to
```
if (target.prop('checked')) {
    if (!this.checked.includes(cid)) {
        this.checked.push(cid);
    }
} else {
    const idx = this.checked.indexOf(cid);
    if (idx !== -1) {
        this.checked.splice(idx, 1);
    }
}
```

manthey (approved):

PR #3872, opened 2026-07-13
Reduce logs for "docs" test

Underwaterr:
When `tox -e docs` is run, both `pip` and `sphinx` output a lot of unhelpful information: long lists of dependencies being installed. By adding the `--quiet` flag to both commands those information-level logs are suppressed but any errors are still output. 

Additionally, `--disable-pip-version-check` means Pip won't check if there's a new version of Pip available because it's neither the time or place for that. 

I've also done some minor syntactical changes to make the commands more readable without changing their behavior.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -47,13 +47,18 @@ deps =
 setenv =
     GIRDER_SPHINX_BUILD = true
 commands =
-    pip install --editable . --editable clients/python --editable ./worker
+    pip install \
+        --quiet \
+        --disable-pip-version-check \
+        --editable . \
+        --editable clients/python \
+        --editable ./worker
     sphinx-build \
-        -W \
-        -b html \
-        -d {envtmpdir}/doctrees \
-        docs \
-        build/docs/html
+        --quiet \
+        --fail-on-warning \
+        --builder html \
+        --doctree-dir {envtmpdir}/doctrees \
+        docs build/docs/html
 
 [testenv:coverage]
 skip_install = true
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -47,13 +47,18 @@ deps =
 setenv =
     GIRDER_SPHINX_BUILD = true
 commands =
-    pip install --editable . --editable clients/python --editable ./worker
+    pip install \
+        --quiet \
```
manthey: I have a concern with adding `--quiet` to the `pip install`.  When some package updates and breaks CI, this is one of the most valuable pieces of forensic evidence about what changed, and frequently looking at the diff of packages before and after lets one see that some transitive package is the issue.

tox.ini
```diff
@@ -47,13 +47,18 @@ deps =
...
+        --disable-pip-version-check \
```
manthey: This is good -- is there an environment variable we could add to tox.ini to do this in general?  It is essentially never useful to see it in tox's environments.

Underwaterr: There sure is!

Underwaterr:
Changes:
- No longer using `--quiet` flag, instead just disabling the Pip progress bar
- Using environment variables set at a global level for Tox

Weird quirk of Tox: if a section defines its own `setenv` it doesn't inherit from the top-level one. Adding `{[tox]setenv}` brings back that inheritance.

manthey (approved):

PR #3874, opened 2026-07-20
Girder Worker Override API URL using Enviroment Variable

### LINKED ISSUE #3857: Override worker girder Api URL with new Environment Variable
`GIRDER_SETTING_WORKER_API_URL` can override `worker.api_url` in the Girder server process. When jobs are launched, that URL is added into the job/task payload (e.g. `girder_api_url`, `jobInfoSpec.url`, and girder_io `api_url` fields).
This breaks inside of mixed deployments where internal and external workers need different reachable API URLs. `GIRDER_SETTING_WORKER_API_URL` only affects the Girder server; setting it on worker containers has no effect (obvious in hindsight because it's modifying the settings).

Example:

- **Girder Server:** on an internal Docker network with service name `girder`, `worker.api_url` set to `http://girder:8080/api/v1`
- **Local Worker:** on the same Docker network — can reach `girder:8080`
- **External Worker:** outside the network — cannot reach `girder:8080`
Jobs created on the Girder server embed `http://girder:8080/api/v1`, so the external worker cannot call back to Girder correctly.

The jobs created within the Girder Server will attach the API_URL of `girder:8080` to the information of the job so the External Worker doesn't hit the proper location.

Suggestion:

Add a worker-process environment variable, e.g. `GIRDER_WORKER_API_URL`, following the same pattern as `GIRDER_WORKER_BROKER`.

Suggestion of some environment variable (`GIRDER_WORKER_API_URL`) that can be applied to individual workers to override the request api url:
https://github.com/girder/girder/blob/master/worker/girder_worker/app.py#L116

Probably needs to be overriden here as well: https://github.com/girder/girder/blob/master/plugins/worker/girder_plugin_worker/utils.py#L27

Then documentation on the environment variable and explination of usage somewhere in the list of environment variables: https://github.com/girder/girder/blob/master/docs/list_env_settings.yaml#L93

BryonLewis:
resolves #3857

Currently a Draft PR so I can easily access it in the PR review and test within DIVE.
The main functionality is to provide a way to override the girder_api_url in tasks that are initiated by the server but happen to live in a different network context.
Specifically when the server and localworker have a shared context that is not publicly accessible but workers may need to be remote.

It relates to an issue in DIVE where a deployment is behind a GCP VPC

__init__.py
- new `GIRDER_WORKER_API_URL` environment variable that can be used to override the tasks bound apiURL with a system environment variable.  Allowing for workers to have a different network path to the server compared to locations deploying the task
- `resolve_girder_api_url(api_url=None)` -  returns either the provided url or the override from the environment value if one exists
- `rewrite_url_pai_base(url, original_api_url, new_api_url)` - The `jobInfoSpec` in girder tasks has it's own URL for reporting back to the server the job.  This utility function is used to replace the default URL with a new one from the environment variable override.  Shouldn't happen but if no api_url is found it will create a warning in the logs and return the original.
- `apply_girder_api_url_override(request)`- takes a request inside of a task (`task.request`) and replaces the `girder_api_url` and  the `jobInfoSpec` url with the environment variable's URL
- within the job_manager initialization the override is applied if found to the jobSpec['url'] as well

nongirder_context.py
- Replaces the girder client api url and sets the header `girder_api_url` to the environment variable url so that any jobs launched from the worker have the proper girder api url used for communication

app.py
- during the prerun for a task this will apply the `GIRDER_WORKER_API_URL` settings for an updated URL to the `task.request` to override the server's girder_api_url setting as well as the job_manager through the `jobInfoSpec`

list_env_settings.yaml
- Added the newer `GIRDER_WORKER_API_URL` environment variable and description
- Added `GIRDER_SETTING_WORKER_API_URL` and explained how it sets the worker apiURL on job creation time

Testing:
- Plan to use this branch in DIVE and test with a remote worker (outside of the docker context in another compose stack) as well as having the local worker for girder 5
	- local worker: http://girder:8080/api/v1 - uses the internal docker network to point to the worker
	- external worker: http://host.docker.internal:8010/api/v1 - so it communicates outside of the docker network.  Need to make sure that the worker docker network will access the external system
	- Ensure both have access to the message broker (may need to expose in DIVE)
	- Could also try running it on separate LAN systems with a public api and broker

Results:
- Testing worked properly and was able to have two disconnected instances the Job updates/status and pushing/pulling of data worked properly.

### DIFF
--- a/docs/list_env_settings.yaml
+++ b/docs/list_env_settings.yaml
@@ -93,6 +93,13 @@ GIRDER_WORKER_BROKER: >-
 GIRDER_WORKER_BACKEND: >-
   URL for the Celery result backend (e.g., redis://localhost:6379/0).
 
+GIRDER_WORKER_API_URL: >-
+  Optional per-worker override for the Girder API base URL used when calling
+  back to the server (job status, logs, data transfer). When set, this takes
+  precedence over the girder_api_url header attached when the job was scheduled.
+  Useful when workers run in a different network context than the web server
+  (for example Docker hostname vs localhost).
+
 GIRDER_WORKER_JOB_GC_SCOPE: >-
   A comma-separated list of extra default scopes for worker jobs.  For instance to allow jobs to schedule and check on other jobs, set this to "jobs.rest.create_job,jobs.rest.list_job"
 
--- a/docs/list_env_settings.yaml
+++ b/docs/list_env_settings.yaml
@@ -72,6 +72,9 @@ GIRDER_SETTING_CORE_FILEHANDLE_MAX_SIZE: >-
 GIRDER_SETTING_CORE_UPLOAD_MINIMUM_CHUNK_SIZE: >-
   Minimum chunk size for file uploads in bytes. Default is 5MB (1024 * 1024 * 5).
 
+GIRDER_SETTING_WORKER_API_URL: >-
+  The base URL for the Girder API when creating worker jobs.  This is used to call back to the server (job status, logs, data transfer) from workers.  This setting makes all jobs that are launched by a the server utilize the URL set here.  If GIRDER_WORKER_API_URL is set on a worker container, that will override this setting on that specific worker.
+
 GIRDER_HOST: >-
   Hostname or IP address to bind the server to. Use '0.0.0.0' to bind to all interfaces.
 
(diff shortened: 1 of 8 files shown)


manthey (approved):
This has worked in my local testing environment.

PR #3875, opened 2026-07-20
Cache public settings in UI

Xarthisius:
This PR stores the response from `GET system/public_settings` that triggers upon up initialization.

### Rationale
This allows plugins to inject custom settings via simple girder events, e.g.:

```python
@access.public
@girderRest.boundHandler
def add_public_settings(self, event):
    settings = event.info["returnVal"]
    public_settings = [PluginSettings.AIMDL_COUNTS]
    settings.update({key: Setting().get(key) for key in public_settings})
```

and later use it in UI extensions without reinventing a lot of internal Girder's machinery.

### DIFF
--- a/girder/web/src/index.ts
+++ b/girder/web/src/index.ts
@@ -29,6 +29,7 @@ const initializeDefaultApp = async (apiRoot: string, el: string | HTMLElement = 
                 url: `system/public_settings`,
                 method: 'GET',
             }).done((resp: any) => {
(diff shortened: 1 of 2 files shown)


manthey (approved):
@Xarthisius Thank you.  Do you have any thoughts on where we should document this so it is easier for others to discover?

Xarthisius:
Probably `docs/plugin-development.rst`. I'll try to add something

PR #3907, opened 2026-08-24
Return partial paths for public resources

### LINKED ISSUE #2588: Client shows an error when browsing to public resources in unshared folders
Placing public folders inside private folders is a supported mechanism for creating "hidden folders".  I have created an example of this on data.kitware.com:

An item in a public folder in a private folder:
https://data.kitware.com/#item/5a6b2e6c8d777f0649e03ad6

When browsing to this as a regular user the page is displayed, but an access error message is displayed.  As an anonymous user, the page loads but a login dialog is shown.

A public folder in a private folder:
https://data.kitware.com/#user/55087c098d777f5ff55d5e3a/folder/5a6b2db08d777f0649e03ad1

In this case, the folder browser fails to appear at all.

naglepuff:
Fix #2588 

Updates the `/rootpath` endpoints. Previously, a `403` was returned if any folder or collection in the hierarchy was private. Now, the entire hierarchy is traversed first, and then the full hierarchy is filtered such that:

1) No private resources are included in the hierarchy list for a user that has no access to those resources
2) Only accessible hierarchy from the child element (inclusive) to the first private member of the hierarchy (exclusive) are returned

This allows users to share links to public folders/items that are nested within private collections or folders. If a shared link contains a private resource, the front end will replace the URL to just contain parent info:
`/#collection/<private_collection_id>/folder/<public_folder_id>` -> `/#folder/<public_folder_id>`

Once this PR, which targets the `3.x-maintenance` branch is merged, a follow-up PR will be opened against `master` to include these changes. 

Additionally, plugins which add additional rest requests that might return `40X` access codes might still see the login dialog on navigating to a public resource inside a private folder (in particular, the request to get the histomics config yaml file). Additional work will need to be done to fix the behavior in the respective plugins.

### DIFF
--- a/girder/api/v1/folder.py
+++ b/girder/api/v1/folder.py
@@ -429,4 +429,4 @@ class Folder(Resource):
         .errorResponse('Read access was denied for the folder.', 403)
     )
     def rootpath(self, folder, params):
-        return self._model.parentsToRoot(folder, user=self.getCurrentUser())
+        return self._model.parentsToRoot(folder, user=self.getCurrentUser(), hideInaccessible=True)
--- a/girder/api/v1/item.py
+++ b/girder/api/v1/item.py
@@ -313,7 +313,7 @@ class Item(Resource):
         .errorResponse('Read access was denied for the item.', 403)
     )
     def rootpath(self, item):
-        return self._model.parentsToRoot(item, self.getCurrentUser())
+        return self._model.parentsToRoot(item, self.getCurrentUser(), hideInaccessible=True)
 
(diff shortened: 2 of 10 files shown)


manthey:
When you get around to fixing the 40x from large_image config files, in that library we should (a) not surface that error to the UI and (b) I think if the _config_ file is readable to the user, then it should be applied, even if the route to get to it is not. This means internally walking up path components that the user doesn't have access to, so I could be convinced that is wrong.

manthey (approved):

PR #3919, opened 2026-08-31
Add a "Search here" option to top search

### LINKED ISSUE #3895: Add a search here option in the top search
When you search in girder it searches everything.  There should be an option to "search here" which would search in the current user/collection/folder only.

bnmajor:
Adds a "Search here" checkbox to the search field that restricts results to wherever the user currently is in the data hierarchy. It defaults to off, so search behavior is unchanged unless the option is checked.

Some routes, like the search results page, aren't hierarchy locations. Rather than disabling the option there, the last hierarchy location the user visited is remembered and reused.

[Screencast from 08-27-2026 04:39:20 PM.webm](https://github.com/user-attachments/assets/bbda1067-d84c-4cfc-93e3-c009895c4189)

Fixes #3895

### DIFF
--- a/girder/models/folder.py
+++ b/girder/models/folder.py
@@ -660,6 +660,38 @@ class Folder(AccessControlledModel):
 
         return folders.count()
 
+    def subtreeFolderIds(self, folder, includeRoot=True):
+        """
+        Return the set of ids of every folder in the subtree rooted at the given folder.
+
+        Folders record only their immediate parent, so this walks the tree one depth level at a
+        time, grouping each level into a single query.
+
+        :param folder: The root of the subtree.
+        :type folder: dict
+        :param includeRoot: Whether to include the root folder's own id.
+        :type includeRoot: bool
+        :returns: The set of folder ids.
+        :rtype: set
+        """
+        seen = {folder['_id']}
+        parentIds = [folder['_id']]
+
+        while parentIds:
+            children = [
+                child['_id']
+                for child in self.find(
+                    {'parentId': {'$in': parentIds}, 'parentCollection': 'folder'},
+                    fields=['_id'])
+                if child['_id'] not in seen
+            ]
+            seen.update(children)
+            parentIds = children
+
+        if not includeRoot:
+            seen.discard(folder['_id'])
+        return seen
+
     def subtreeCount(self, folder, includeItems=True, user=None, level=None):
         """
         Return the size of the subtree rooted at the given folder. Includes
--- a/girder/api/v1/resource.py
+++ b/girder/api/v1/resource.py
@@ -4,7 +4,7 @@ from girder.constants import AccessType, TokenScope
 from girder.exceptions import RestException
 from girder.api import access
 from girder.utility import parseTimestamp
-from girder.utility.search import getSearchModeHandler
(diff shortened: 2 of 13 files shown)


bnmajor:
@manthey Open question: what should "Search here" do for a search mode that can't honor it? Is that even a case we need to consider? If so, we could:
1. Hide the checkbox for modes that don't support it
2. Show it grayed out with a tool tip. Explains why instead of silently removing the option, and avoids the checkbox appearing/disappearing as you flip between modes
3. Leave it checked and search everywhere anyway - results would say "searching here" while showing matches from outside
4. Leave it and show an error

manthey:
> @manthey Open question: what should "Search here" do for a search mode that can't honor it? Is that even a case we need to consider? If so, we could:

Which search modes don't honor it? Plugins _can_ add arbitrary search modes, but I think we could make all of the existing modes honor it if they do not do so already.

bnmajor:
> > @manthey Open question: what should "Search here" do for a search mode that can't honor it? Is that even a case we need to consider? If so, we could:
> 
> Which search modes don't honor it? Plugins _can_ add arbitrary search modes, but I think we could make all of the existing modes honor it if they do not do so already.

So the one I was looking at was large image's annotation metadata search. Essentially the wall I feel like I'm hitting is that it's not clear to me how to to apply "search here" generically since plugins return results rather than queries, so there's not necessarily the required information for us to apply the restriction to. For a search like the annotation metadata one specifically, there is no folder id for us to use. Of course I'm happy to discuss further (here or offline if that would be easier) since I may just be misunderstanding things.

girder/models/folder.py
```diff
@@ -660,6 +660,38 @@ def countFolders(self, folder, user=None, level=None):
 
         return folders.count()
 
+    def subtreeFolderIds(self, folder, includeRoot=True):
+        """
+        Return the set of ids of every folder in the subtree rooted at the given folder.
```
manthey, line 665: I'm nervous about this -- a subtree could have millions of folders.

manthey:
> > > @manthey Open question: what should "Search here" do for a search mode that can't honor it? Is that even a case we need to consider? If so, we could:
> > 
> > 
> > Which search modes don't honor it? Plugins _can_ add arbitrary search modes, but I think we could make all of the existing modes honor it if they do not do so already.
> 
> So the one I was looking at was large image's annotation metadata search. Essentially the wall I feel like I'm hitting is that it's not clear to me how to to apply "search here" generically since plugins return results rather than queries, so there's not necessarily the required information for us to apply the restriction to. For a search like the annotation metadata one specifically, there is no folder id for us to use. Of course I'm happy to discuss further (here or offline if that would be easier) since I may just be misunderstanding things.

We need to use a pipeline rather than a list of subfolders to search. Such a pipeline passed to the search mode that "supports hierarchy" would be modifiable by the metadata search mode to find annotations belong to items that are in the selected folders. I think your basic expansion method is sound (other than switching from a filter of ids to a pipeline); if the search doesn't support hierarchy, we could hide it in the "here" mode.

And, I have found that LLMs often get me 90% of the way to a decent mongo pipeline (but have yet to get there completely).

PR #3927, opened 2026-09-04
Add a show-download setting that defaults to all

naglepuff:
@manthey I found a branch from ~2 years ago where it looked like you had started the work for this.

I made a couple of changes, including renaming `DOWNLOAD_SHOWN` -> `SHOW_DOWNLOAD` to fit the existing pattern, which was dominated by VERB_ACTION. I also deduplicated the hierarchy traversal to get the setting value from the top level `App` view into a utility.

Tests have been added for validation of the setting, unit test of the utility function, and some integration tests to verify the setting affects the UI.

### DIFF
--- a/girder/web_client/src/templates/body/itemPage.pug
+++ b/girder/web_client/src/templates/body/itemPage.pug
@@ -5,28 +5,30 @@
     if (accessLevel >= AccessType.WRITE)
       button.g-upload-into-item.btn.btn-sm.btn-success(title="Upload files into item")
         i.icon-upload
-    .btn-group
-      button.g-item-actions-button.btn.btn-sm.btn-default.dropdown-toggle(
-          data-toggle="dropdown", title="Item actions")
-        i.icon-doc-text-inv
-        |  Actions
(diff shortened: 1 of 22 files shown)


girder/web_client/src/templates/body/itemPage.pug
```diff
@@ -5,28 +5,30 @@
...
+          |  Actions
+          i.icon-down-dir
+        ul.g-item-actions-menu.dropdown-menu.pull-right(role="menu")
+          if (showDownload)
+            li(role="presentation")
+              a.g-download-item(role="menuitem", href=item.downloadUrl())
+                i.icon-download
+                | Download item
+          if (accessLevel >= AccessType.WRITE)
+            li.divider(role="presentation")
```
manthey, line 22: It's very petty, but can we only show this divider IF we are also showing "Download item"?

naglepuff: Fixed for the item and collection pages in ce17867

manthey (approved):

